PR #3014, opened 2019-05-27
Accept mmap objects in places where bytearray is expected.

jonathanslenders:
Fixes #1467.

PR #7158, opened 2022-02-07
Make regex methods accept all buffer types as inputs

### LINKED ISSUE #1467: Mypy does not accept `mmap` object in `re.match`
It seems that the `re.compile` types are to restrictive, the following code errors in mypy but running it is no problem:
```python
# _txt_fmt is a object gotten from `re.compile`
with open(file, 'r+') as f:
    with mmap.mmap(f.fileno(), 0) as data:
        match = _txt_fmt.match(data)
```

Mypy gives the error `Argument 1 to "match" of "Pattern" has incompatible type "mmap"; expected "bytes"`.

itaisteinherz:
Fixes #1467
Refs #3014

```
~/Development/PoCs/typing
❯ cat scratch.py
import mmap
import re

with open("scratch.py", 'r+') as f:
    with mmap.mmap(f.fileno(), 0) as data:
        match = re.compile(b"").match(data)

~/Development/PoCs/typing
❯ mypy scratch.py
scratch.py:6: error: Argument 1 to "match" of "Pattern" has incompatible type "mmap"; expected "bytes"

~/Development/PoCs/typing
❯ mypy --custom-typeshed-dir ../../OSS/typeshed scratch.py
```


stdlib/mmap.pyi
```diff
@@ -27,7 +27,7 @@ if sys.platform != "win32":
 
     PAGESIZE: int
 
-class mmap(AbstractContextManager[mmap], Iterable[int], Sized):
+class mmap(AbstractContextManager[mmap], bytearray, Iterable[int], Sized):
```
itaisteinherz, line 30: By the way, why does `bytearray` have to be in this position in the class signature? I initially wanted to place it after `Iterable[int]` and before `Sized`, but received this error:

```
~/Development/PoCs/typing
❯ mypy --custom-typeshed-dir ../../OSS/typeshed scratch.py
scratch.py:5: error: "mmap" has no attribute "__enter__"; maybe "__iter__"?
scratch.py:5: error: "mmap" has no attribute "__exit__"
```

srittau:
In the past we've had problems whenever we tried to "fudge" types and made them differ from runtime types. I think a better solution is python/typing#593.

JelleZijlstra:
I see that I approved the previous instance of this, but I no longer think it's a good idea to lie in the stubs.

The proper solution is python/typing#593. Until that is implemented, we can just use the aliases from https://github.com/python/typeshed/blob/master/stdlib/_typeshed/__init__.pyi#L185.

itaisteinherz:
@srittau @JelleZijlstra I'll update this PR to use the currently-implemented aliases then.

Regarding https://github.com/python/typing/issues/593 (and https://github.com/python/typing/issues/593#issuecomment-1031906706), since I've never participated in any Python maling list discussion nor worked on CPython, it'll probably take me a while until I'll be comfortable with leading the implementation of the better solution 😬

stdlib/mmap.pyi
```diff
@@ -27,7 +27,7 @@ if sys.platform != "win32":
 
     PAGESIZE: int
 
-class mmap(AbstractContextManager[mmap], Iterable[int], Sized):
+class mmap(AbstractContextManager[mmap], ReadableBuffer, Iterable[int], Sized):
```
itaisteinherz, line 30: Another thought: how does this even work? `ReadableBuffer` is a `Union` which includes `mmap.mmap`, which leads me to the conclusion that this should result in some sort of infinite type loop.

itaisteinherz: Update: this does lead to an error :) Will probably split up `WriteableBuffer` to fix this.

JelleZijlstra: We should add it to the union, not use the union as a base class.

itaisteinherz: What do you mean? `mmap.mmap` is already part of `WriteableBuffer`.

JelleZijlstra: Oh, right. Then we shouldn't change it at all.

The original problem was that mmap wasn't accepted in re.match. If re.match accepts all buffers, then we should change its signature, not anything about mmap.

itaisteinherz: Upon further research, it seems to me like a way to solve this would be to have `mmap.mmap` extend `ByteString`, and change the definition of `AnyStr` to:

```python
AnyStr = TypeVar("AnyStr", str, ByteString)
```

However, that would require a change in CPython. I could also create an `AnyStr` alternative, specific to `Pattern`, with this definition. One last option would be to have `mmap.mmap` extend `bytes` (aka `ReadOnlyBuffer`). What do you think?

JelleZijlstra: We should not change AnyStr. That would require CPython runtime changes. We could use a separate typevar specific to regexes, but I'd have to look more into it to figure out whether that's feasible.

JelleZijlstra (changes requested):
Changing the definition of AnyStr is not an option.

JelleZijlstra:
Nice, the last run found a mypy crash. If you're so inclined, feel free to report that on the mypy issue tracker.

itaisteinherz:
> Nice, the last run found a mypy crash. If you're so inclined, feel free to report that on the mypy issue tracker.

I didn't even notice 🧐 Will take a look. Also, defining `StrOrBuffer` as `TypeVar("StrOrBuffer", str, ReadableBuffer)` didn't work for some reason. I had to write it as `TypeVar("StrOrBuffer", str, ReadOnlyBuffer, WriteableBuffer)`.

itaisteinherz:
Note that the PR does not fix the original issue at the moment, as `re.compile(b"")` returns `Pattern[bytes]`, which means that running `.match(data)` will not work. That would probably be fixed if I were able to do:

```python
StrOrBuffer = TypeVar("StrOrBuffer", str, ReadableBuffer)
```

However that didn't work for me earlier, so I'll look into it.

stdlib/typing.pyi
```diff
@@ -587,69 +587,71 @@ class TextIO(IO[str]):
...
-    # group() returns "AnyStr" or "AnyStr | None", depending on the pattern.
+    re: Pattern[StrOrBuffer]
+    def expand(self, template: StrOrBuffer) -> StrOrBuffer: ...
+    # group() returns "StrOrBuffer" or "StrOrBuffer | None", depending on the pattern.
     @overload
-    def group(self, __group: _Literal[0] = ...) -> AnyStr: ...
+    def group(self, __group: _Literal[0] = ...) -> StrOrBuffer: ...
     @overload
-    def group(self, __group: str | int) -> AnyStr | Any: ...
+    def group(self, __group: str | int) -> StrOrBuffer | Any: ...
```
JelleZijlstra, line 605: This isn't right: if you match on a memoryview, you get bytes objects from .group(), not memoryviews.

stdlib/re.pyi
```diff
@@ -52,66 +53,90 @@ if sys.version_info < (3, 7):
     _pattern_type: type
 
 @overload
-def compile(pattern: AnyStr, flags: _FlagsType = ...) -> Pattern[AnyStr]: ...
+def compile(pattern: StrOrBuffer, flags: _FlagsType = ...) -> Pattern[StrOrBuffer]: ...
```
JelleZijlstra, line 56: re.compile actually works only on str or bytes

JelleZijlstra (changes requested):
This doesn't exactly describe re's real behavior in a few cases. I think Pattern and Match should continue to be generic over AnyStr, but additional overloads can be added so that e.g. Pattern[bytes].search() accepts other ReadableBuffers.

itaisteinherz:
I noticed that the type of `Match.string` depends on the type of the `string` provided to `Pattern.match`. It seems to me like a neat way to implement this would be to add an additional type variable to `Match`, however that would be a breaking change (I think).

itaisteinherz:
Had it been possible to provide a default value for `MatchString`, that would have been a non-breaking change. However, it seems like [the discussion on defaults for `Generic`](https://github.com/python/mypy/issues/4236) was never resolved.

JelleZijlstra:
I implemented an alternative in #7679. Thanks for your work here! You may be interested in the upcoming [PEP 688](https://github.com/python/peps/pull/2549).

PR #7161, opened 2022-02-08
Improve in-place BinOp methods for sets

AlexWaygood:
This PR proposes altering the annotations of four BinOp dunder methods for set classes, so that they return `Self`. The four methods in question are:

- `__ior__`
- `__ixor__`
- `__iand__`
- `__isub__`

I have several reasons for proposing this change:

1. The existing annotations for `__ior__` and `__ixor__` are unsafe. If I have a set `s = {1, 2}` it has type `set[int]`, but if I then do `s |= {'foo', 'bar'}`, the type of the variable has mutated from `set[int]` to `set[str | int]`.
2. Mypy [will actually complain](https://mypy-play.net/?mypy=latest&python=3.10&gist=4ec2288e35a2683117dbc48e3ac2efa6) about point (1) if you try to do it, but it gives an obscure error: `Result type of | incompatible in assignment`. If type checkers are going to balk at this kind of thing anyway, then we may as well improve the error message.
3. The proposed new annotations will be more accurate in subclasses of `set`: these methods do, in fact, return `self` at runtime.
4. This brings the signatures of these dunder methods in line with their semantically equivalent methods (see table below).
5. Consistency with the rest of typeshed: these methods are currently the only (or, if there are others, they're very rare) in-place BinOp methods that do _not_ return `Self`.
6. It just doesn't really make _sense_ for any subclass of `typing.AbstractSet` or `typing.MutableSet` to return anything except `Self` from any of these methods, given that they update the object in-place.

For those who (like me) haven't got their `set` operations memorized, here's a rundown I made for myself...

| Dunder method | Symbol | Equivalent method                        | What it does 
|-------------------|----------|--------------------------------------|----------------
| `__ior__`               | `\|=`       | `.update()`                                     | Add elements to the first set if they are in second set but not in the first set
| `__ixor__`            | `^=`       | `.symmetric_difference_update()` | Add any elements to the first set if they appear in the second set but not in the first set; remove any elements from the first set if they also appear in the second set.
| `__iand__`           | `&=`       | `.intersection_update()`                | Remove any elements from the first set if they do not appear in the second set
| `__isub__`           | `-=`         | `.difference_update()`                  | Remove any elements from the first set if they appear in the second set.


stdlib/_weakrefset.pyi
```diff
@@ -20,15 +20,15 @@ class WeakSet(MutableSet[_T], Generic[_T]):
     def __contains__(self, item: object) -> bool: ...
     def __len__(self) -> int: ...
     def __iter__(self) -> Iterator[_T]: ...
-    def __ior__(self, other: Iterable[_S]) -> WeakSet[_S | _T]: ...
+    def __ior__(self: Self, other: Iterable[_T]) -> Self: ...  # type: ignore[override,misc]
     def difference(self: Self, other: Iterable[_T]) -> Self: ...
-    def __sub__(self: Self, other: Iterable[_T]) -> Self: ...
-    def difference_update(self, other: Iterable[_T]) -> None: ...
-    def __isub__(self: Self, other: Iterable[_T]) -> Self: ...
+    def __sub__(self: Self, other: Iterable[Any]) -> Self: ...
```
srittau, line 25: Why did you change this (as well as `__and__`)? If nothing else, it seems a bit off-topic for this PR, but also inconsistent with the super class annotations.

srittau (approved):

PR #7165, opened 2022-02-08
Add locale.gettext and related stubs, fixes #7164

### LINKED ISSUE #7164: locale/pyi missing gettext related functions
https://docs.python.org/3/library/locale.html#access-to-message-catalogs

The functions in this section are missing from locale.pyi causing incorrect errors.

Ran into here https://github.com/sezanzeb/input-remapper/pull/288/files#diff-0fa1c2f5112b03576baf7c13cc5bc8145a4b86dc37ce877f5aa42f8b2fed676dR30

```
inputremapper/gui/gettext.py:30:1: error: Module has no attribute "bindtextdomain"
inputremapper/gui/gettext.py:31:1: error: Module has no attribute "textdomain"
```

LunNova:
As draft as quickly made with web editor and need to check locally later.

Fixes #7164


LunNova:
The CI setup here is really nice! ^_^

I still haven't checked it out and ran black/flake8 etc locally according to the guidelines but pre-commit.ci has done those for me so I guess there's no need.

stdlib/locale.pyi
```diff
@@ -103,6 +103,15 @@ def atof(string: _str, func: Callable[[_str], float] = ...) -> float: ...
 def atoi(string: _str) -> int: ...
 def str(val: float) -> _str: ...
 
+# native gettext functions
+# https://github.com/python/cpython/blob/f4c03484da59049eb62a9bf7777b963e2267d187/Modules/_localemodule.c#L626
+if sys.platform == "linux" or sys.platform == "darwin":
+    def gettext(__msg: _str) -> _str: ...
+    def dgettext(__domain: _str, __msg: _str) -> _str: ...
```
JelleZijlstra, line 110:
```suggestion
    def dgettext(__domain: _str | None, __msg: _str) -> _str: ...
```
The argument clinic code has `    domain: str(accept={str, NoneType})`.

stdlib/locale.pyi
```diff
@@ -103,6 +103,15 @@ def atof(string: _str, func: Callable[[_str], float] = ...) -> float: ...
...
+    def dcgettext(__domain: _str, __msg: _str, __category: _str) -> _str: ...
```
JelleZijlstra, line 111:
```suggestion
    def dcgettext(__domain: _str | None, __msg: _str, __category: int) -> _str: ...
```
Also from the C signature

stdlib/locale.pyi
```diff
@@ -103,6 +103,15 @@ def atof(string: _str, func: Callable[[_str], float] = ...) -> float: ...
...
+    def textdomain(__domain: _str | None) -> _str | None: ...
+    def bindtextdomain(__domain: _str, __dir: _str) -> _str | None: ...
```
JelleZijlstra, line 113: The dir argument can also be a path-like object or None. You probably want to use `_typeshed.StrPath` but I haven't checked if bytes paths work.

stdlib/locale.pyi
```diff
@@ -103,6 +103,15 @@ def atof(string: _str, func: Callable[[_str], float] = ...) -> float: ...
...
+
```
JelleZijlstra, line 114: Any reason not to add the `bind_textdomain_codeset` function?

JelleZijlstra (changes requested):

stdlib/locale.pyi
LunNova: Added with `StrOrBytesPath | None`

LunNova: While it does take byte paths as input, if you use bytes that aren't a valid string it will successfully set the path to that then crash trying to return it because it always returns a str.

stdlib/locale.pyi
LunNova: Added

stdlib/locale.pyi
srittau: In that case I think it's safer to just allow `StrPath` here.

LunNova: Changed to `StrPath`

Issue #6063, opened 2021-09-22
`random.Random.getvalue` and `random.Random.setvalue` are unduly opinionated about their types and ill-suited for subclassing


posita:
If i understand [the docs](https://docs.python.org/3/library/random.html) correctly, this …

> Class [`Random`](https://docs.python.org/3/library/random.html#random.Random) can also be subclassed if you want to use a different basic generator of your own devising: in that case, override the `random()`, `seed()`, `getstate()`, and `setstate()` methods. Optionally, a new generator can supply a `getrandbits()` method — this allows [`randrange()`](https://docs.python.org/3/library/random.html#random.randrange) to produce selections over an arbitrarily large range.
>
> …
>
> `random.`**`getstate`**`()`
> Return an object capturing the current internal state of the generator. This object can be passed to [`setstate()`](https://docs.python.org/3/library/random.html#random.setstate) to restore the state.
> 
> `random.`**`setstate`**`(`_`state`_`)`
_state_ should have been obtained from a previous call to [`getstate()`](https://docs.python.org/3/library/random.html#random.getstate), and [`setstate()`](https://docs.python.org/3/library/random.html#random.setstate) restores the internal state of the generator to what it was at the time [`getstate()`](https://docs.python.org/3/library/random.html#random.getstate) was called.

… strongly suggests that the return value of `getstate()` (and corresponding argument to `setstate()`) should be opaque and treated as implementation specific. However, typeshed [seems to want to enforce that they should be `Tuple`s](https://github.com/python/typeshed/blob/ee487304d76c671aa353a15e34bc1102bffa2362/stdlib/random.pyi#L12-L13).

This leads to warnings for subclasses who, e.g., capture state as `bytes` or a `dict`. E.G.:

```
…/foo.py:57: error: Return type "_RandState" of "getstate" incompatible with return type "Tuple[Any, ...]" in supertype "Random"
…/foo.py:57: error: Return type "_RandState" of "getstate" incompatible with return type "Tuple[int, ...]" in supertype "Random"
```

(Not sure what that second warning is or where it came from.)

hauntsaninja:
Seems like another use case for defaulted generics.

srittau:
For now, the best we could do is to use `Any` to prevent false positives? But in this particular case, for now, I believe we should keep the stubs as is, and rely on users overriding these methods to use `# type: ignore`.

posita:
[...]

PR #7168, opened 2022-02-08
Improve `Random.getstate` and `Random.setstate`

sobolevn:
Source: https://github.com/python/cpython/blob/243436f3779c1e7bed1fd4b23d5a8ec5eff40699/Lib/random.py#L168

Refs https://github.com/python/typeshed/blob/4b75ecbbe2baf600847b79c3e4c8ffd97ff6cae5/stdlib/_random.pyi#L2


hauntsaninja (commented):
Hmm, this feels potentially a little bit fragile, since the return type is versioned. That said, for the state versions that are currently supprted (2 & 3), the return type is about the same (the only difference is version 2 uses signed ints and version 3 uses unsigned). Ideally I think we'd have a `Tuple[int, Any, ...]` type, but that's not legal.

Also linking #6063, for some previous discussion of these methods, where the argument is made that these should be opaque types.

All in all I'm in favour of this, but figured I'd mention in case you see something worth changing or some clever solution.

sobolevn:
I thought about `VERSION` as `Literal[2]` and `Literal[3]`, but it feels too complex with almost no benefits.

I also think that concrete `tuple` is better than just `tuple[Any, ...]`. If someone wants to redefine this methos, `# type: ignore` is there to help.

stdlib/random.pyi
```diff
@@ -9,15 +9,16 @@ _T = TypeVar("_T")
 
 class Random(_random.Random):
     VERSION: ClassVar[int]
+    gauss_next: float | None  # undocumented
```
JelleZijlstra, line 12: Do you have a use for this? We have a policy to add undocumented things if users ask for them, but we shouldn't go out of our way to add every undocumented attribute.

JelleZijlstra:
I feel like this might be better as just `object`. The exact return type is clearly an implementation detail, and I don't see much point in describing the precise format in typeshed.

srittau:
Using `NewType` is also an option.

stdlib/random.pyi
sobolevn: Yes, it is quite widely used: https://cs.github.com/?q=.gauss_next%20language%3APython&scopeName=All%20repos

I personally remember using in it my fake-data generator project.

sobolevn:
> The exact return type is clearly an implementation detail, and I don't see much point in describing the precise format in typeshed.

Yes, I agree. We will just generate `overload` errors for users with alterantive implementations.

PR #7146, opened 2022-02-06
Improve `toml` typings

FaresAhmedb:
Add `toml`'s full API. Fixes #6484.

PR #7170, opened 2022-02-09
Improve `toml.pyi` by removing old versions

sobolevn:
No other modules use `>= (3, 6)`.
I am not even saying about `(3, 4)` 🙂


sobolevn:
Oh, I see. It support `python2`.

sobolevn:
What about `3.4` though?

Akuli:
This will conflict with #7146, which also cleans up the version checks.

PR #7176, opened 2022-02-11
protobuf doesn't support Python 2 anymore

srittau:
Remove Python-2-only package "futures"


srittau:
Cc @nipunn1313

nipunn1313 (approved):
Correct! Thanks!

PR #7177, opened 2022-02-11
Add stubs for `pynput` package

### LINKED ISSUE #4328: pynput
Is this where you request typing annotation?

FaresAhmedb:
Fixes #4328


stubs/pynput/@tests/stubtest_allowlist.txt
```diff
@@ -0,0 +1,9 @@
+pynput
+pynput._info
+pynput._util.RESOLUTIONS
+pynput.keyboard
+pynput.keyboard._base
+pynput.keyboard._dummy
+pynput.mouse
+pynput.mouse._base
+pynput.mouse._dummy
```
srittau, line 9: Why are all of those on the stubtest allowlist?

srittau:
@rchen152 Could you have a look at the pytype failure? I don't see anything obvious wrong.

FaresAhmedb:
I already did but there's no obvious solution to me.
```python
    @property  # type: ignore[misc]
    @contextlib.contextmanager
    def modifiers(self) -> Iterator[set[Key]]: ...
```
mypy already complains about it if we remove `type: ignore[misc]`
```sh
stubs/pynput/pynput/keyboard/_base.pyi:103: error: Decorated property not supported
```
but pytype doesn't ignore the line.

stubs/pynput/@tests/stubtest_allowlist.txt
FaresAhmedb: stubtest fails to import `pynput` since there's no display server running (e.g. x11)

srittau: I see. We should probably re-introduce the flag to disable stubtest in `METADATA.toml` we had for a short time. I will send a PR tomorrow.

Akuli: We could run stubtest under xvfb, which emulates an x11 display. For example: https://github.com/Akuli/porcupine/blob/1ad0d5fa8d944ef66a0dbb7e0100da3a13aebfda/.github/workflows/check.yml#L60-L62

srittau:
Depends on #7351. This adds support for the `stubtest = false` key back to `METADATA.toml`.

rchen152:
Looking into the pytype error.

rchen152:
This change seems to get rid of the pytype_test failure:
```
--- a/stubs/pynput/pynput/keyboard/_base.pyi
+++ b/stubs/pynput/pynput/keyboard/_base.pyi
@@ -101,8 +101,7 @@ class Controller:
     def pressed(self, *args: str | Key | KeyCode) -> Iterator[None]: ...
     def type(self, string: str) -> None: ...
     @property  # type: ignore[misc]
-    @contextlib.contextmanager
-    def modifiers(self) -> Iterator[set[Key]]: ...
+    def modifiers(self) -> contextlib.AbstractContextManager[Iterator[set[Key]]]: ...
     @property
     def alt_pressed(self) -> bool: ...
     @property
```

stubs/pynput/@tests/stubtest_allowlist.txt
hauntsaninja: Am I missing something? stubtest does not appear to fail to import pynput in CI. In fact, if it was failing, you wouldn't see any errors.

FaresAhmedb: @hauntsaninja Here is the failing CI test in the first commit:

https://github.com/python/typeshed/runs/5163416378
```
error: pynput failed to import: this platform is not supported: ('failed to acquire X connection: Bad display name ""', DisplayNameError(''))
```

I added the allowlist in the commit after it.

FaresAhmedb:
Thanks! Just added the change in the last commit. 😄

stubs/pynput/@tests/stubtest_allowlist.txt
hauntsaninja: Oh, I see how I confused myself. Sorry about that, thanks!

stubs/pynput/METADATA.toml
```diff
@@ -0,0 +1,2 @@
+version = "1.7.*"
+stubtest = false
```
srittau, line 2: Could you add a short comment why we need to disable stubtest?

stubs/pynput/pynput/_util.pyi
```diff
@@ -0,0 +1,69 @@
+import sys
+import threading
+from _typeshed import Self
+from queue import Queue
+from types import ModuleType, TracebackType
+from typing import Any, Callable, ClassVar, Generic, TypeVar
```
srittau, line 6: `Callable` should be imported from `collections.abc`. Same for the other files.

stubs/pynput/pynput/mouse/_base.pyi
```diff
@@ -0,0 +1,38 @@
...
+    left: int
+    middle: int
+    right: int
+
+class Controller:
+    def __init__(self) -> None: ...
+    @property
+    def position(self) -> tuple[int, int]: ...
+    @position.setter
+    def position(self, pos: tuple[int, int]) -> None: ...
```
srittau, line 19: Just as a heads up: This might not work and we might need to use just `position: tuple[int, int]`, but we can always change it later if that turns out to be problematic.

srittau (changes requested):
Thanks, looks good to me, although I didn't compare it to the implementation. A few nits below.

srittau (approved):

PR #7179, opened 2022-02-12
Fix `context` type in `pprint.PrettyPrinter.format`

sobolevn:
In this case `int` is a value, not a key: https://github.com/python/cpython/blob/b70690bb37cc4bac695051484734eede0c1f9ada/Lib/pprint.py#L180


stdlib/pprint.pyi
```diff
@@ -131,4 +131,4 @@ class PrettyPrinter:
     def pprint(self, object: object) -> None: ...
     def isreadable(self, object: object) -> bool: ...
     def isrecursive(self, object: object) -> bool: ...
-    def format(self, object: object, context: dict[int, Any], maxlevels: int, level: int) -> tuple[str, bool, bool]: ...
+    def format(self, object: object, context: dict[Any, int], maxlevels: int, level: int) -> tuple[str, bool, bool]: ...
```
JelleZijlstra, line 134:
```suggestion
    def format(self, object: object, context: dict[int, int], maxlevels: int, level: int) -> tuple[str, bool, bool]: ...
```
Looks like the keys are also ints, they come from `id()`.

sobolevn:
Thank you!

PR #7181, opened 2022-02-12
Improve `pathlib` stubs

AlexWaygood:
- `__class_getitem__` has been removed in 3.11
- Several methods always raise `NotImplementedError` on Windows, so it's better to just pretend they don't exist at all on Windows imo


stdlib/pathlib.pyi
```diff
@@ -147,7 +143,14 @@ class Path(PurePath):
     def open(
         self, mode: str, buffering: int = ..., encoding: str | None = ..., errors: str | None = ..., newline: str | None = ...
     ) -> IO[Any]: ...
-    def owner(self) -> str: ...
+    if sys.platform != "win32":
+        # These methods do "exist" on Windows, but they always raise NotImplementedError,
```
JelleZijlstra, line 147: Not sure about this, calling code might just catch the NotImplementedError. Let's see what mypy-primer says.

AlexWaygood: Primer seems happy!

PR #7185, opened 2022-02-13
`asyncio`: make `AbstractServer` abstract and remove unnecessary metaclass=ABCMeta

AlexWaygood:
I've been fiddling about with a script to identify functions that:

1. from an AST analysis of the runtime source code, appear to return `None` at runtime.
2. Are annotated as returning something other than `None` at runtime.

One thing the script has identified is that there seem to be a bunch of classes in `asyncio` that are not marked as being abstract in typeshed, but look awfully like they're intended to be abstract at runtime.


stdlib/asyncio/events.pyi
```diff
@@ -55,16 +55,22 @@ class TimerHandle(Handle):
     if sys.version_info >= (3, 7):
         def when(self) -> float: ...
 
-class AbstractServer:
+class AbstractServer(metaclass=ABCMeta):
```
Akuli, line 58: We shouldn't lie about the metaclass in typeshed. It's not necessary. Mypy will error if you try to instantiate a class that doesn't override methods decorated with `@abstractmethod`, even if the base class doesn't have `ABCMeta` as its metaclass.

stdlib/asyncio/protocols.pyi
```diff
@@ -1,28 +1,29 @@
 import sys
 from asyncio import transports
+from typing import Protocol as TypingProtocol
```
Akuli, line 3: Why should asyncio's protocols be typing protocols? Does that solve a concrete problem? Is it common to define classes that have compatible methods but don't inherit from asnycio's protocol classes?

Even if this is a good idea, I would prefer this to be a separate pull request, as it doesn't have anything to do with making classes abstract.

AlexWaygood: You're right, I'll move the changes to this file into a new PR, with an explanation as to why the changes are being made.

I originally made the methods on these protocols abstract too, but it wasn't a good approach; so those changes no longer belong in this PR.

stdlib/asyncio/events.pyi
```diff
@@ -56,15 +56,21 @@ class TimerHandle(Handle):
...
         def is_serving(self) -> bool: ...
+        @abstractmethod
         async def start_serving(self) -> None: ...
+        @abstractmethod
         async def serve_forever(self) -> None: ...
 
+    @abstractmethod
     async def wait_closed(self) -> None: ...
 
 class AbstractEventLoop(metaclass=ABCMeta):
```
Akuli, line 76: Maybe we should remove the metaclass here, and figure out why it was originally added to this class but not to `AbstractServer`.

AlexWaygood: `AbstractEventLoop` actually does have `ABCMeta` as the metaclass at runtime, though: https://github.com/python/typeshed/blob/5081f684c0ddce5dca9cb0678583c0793bfdef35/stdlib/asyncio/events.pyi#L70

AlexWaygood: Yes.

Akuli: `AbstractServer` was added without the metaclass in 0f5c38a2ce9e99b8bfdbffc5f22cbaa371a69873, when `AbstractEventLoop` already had the metaclass. This doesn't seem to be intentional.

AlexWaygood: It might be because of this line, where the metaclass lie is in fact necessary, or else mypy will complain that you either need to implement all abstractmethods from the base class or use `ABCMeta` as the metaclass to be explicit about the fact that you want this class that has no abstractmethods in it to be abstract:

https://github.com/python/typeshed/blob/7fa3d9017d54764694b6c48f25729eabfe02a514/stdlib/asyncio/events.pyi#L513

Akuli:
```
$ grep -r metaclass=ABCMeta stdlib/asyncio/
stdlib/asyncio/events.pyi:class BaseDefaultEventLoopPolicy(AbstractEventLoopPolicy, metaclass=ABCMeta):
stdlib/asyncio/base_events.pyi:class BaseEventLoop(AbstractEventLoop, metaclass=ABCMeta):
```

Neither of these seems to be necessary.

AlexWaygood:
> ```
> $ grep -r metaclass=ABCMeta stdlib/asyncio/
> stdlib/asyncio/events.pyi:class BaseDefaultEventLoopPolicy(AbstractEventLoopPolicy, metaclass=ABCMeta):
> stdlib/asyncio/base_events.pyi:class BaseEventLoop(AbstractEventLoop, metaclass=ABCMeta):
> ```
> 
> Neither of these seems to be necessary.

The metaclass lie is necessary for both of these classes, unfortunately. If we do not lie about the metaclass, then mypy will complain about not all `abstractmethod`s being implemented -- because there are no `abstractmethod`s defined in the body of either of these classes, mypy will assume that the classes are meant to be a concrete implementation of the abstract classes they inherit from.

The way to get mypy to accept that the subclass is also meant to be abstract, even though it defines no abstractmethods itself, is explicitly re-specify the metaclass in the subclass. Unfortunately, in this case that means lying about the metaclass.

Akuli:
`tests/mypy_test.py` passes to me with `metaclass=ABCMeta` removed from `BaseEventLoop`, although the one in `BaseDefaultEventLoopPolicy` seems to be necessary.

AlexWaygood:
> `tests/mypy_test.py` passes to me with `metaclass=ABCMeta` removed from `BaseEventLoop`, although the one in `BaseDefaultEventLoopPolicy` seems to be necessary.

Okay, let's try it

Akuli (approved):

PR #7186, opened 2022-02-13
Make `asyncio.protocols` classes protocols

AlexWaygood:
This PR proposes making the classes in `asyncio.protocols` typing `Protocol`s.

These classes [are clearly, in some sense, abstract](https://github.com/python/cpython/blob/main/Lib/asyncio/protocols.py): the bodies of all methods on all of these classes are completely empty. The intent is clearly to provide a set of abstract interfaces on which other modules can build (the module docstring is "Abstract Protocol base classes").

However, marking the methods on these classes as `abstractmethod`s would be the wrong approach. There are many classes that inherit from one of these protocols without providing concrete implementations of these quasi-abstract classes, meaning type-checkers would start raising unnecessary false-positive errors.

Moreover, these classes seem very similar to the concept of a PEP 544 protocol. The docstring for `asyncio.protocols.Protocol` states:

```
The user should implement this interface. They can inherit from
this class but don't need to.
```

In other words, these classes describe _abstract structures_, in exactly the same way subclasses of `typing.Protocol` do.


JelleZijlstra:
Not so sure about this; they're not in fact `typing.Protocol`s, and marking them as such may have confusing effects on type checkers.

AlexWaygood:
> Not so sure about this; they're not in fact `typing.Protocol`s, and marking them as such may have confusing effects on type checkers.

It's a weird one, I agree, but:

- I'd argue it's the same basic idea as a `typing.Protocol`
- A static type-checker currently won't recognise a class as satisfying one of these interfaces, even if it implements all the required methods. That's arguably a false-positive error, since the docstring specifically states that classes don't need to explicitly inherit from one of these classes in order to be considered to satisfy the protocol.
- We do a similar thing with `os.PathLike`, which is an ABC at runtime but a `Protocol` in typeshed.

srittau:
I agree with @JelleZijlstra. As a general rule, we should try to avoid these kinds of hacks, especially considering the large number of users typeshed has by now. A few selected hacks are generally know to work (e.g. using `@abstractmethod` in stubs only), but this is far too big a difference for me to be comfortable.

AlexWaygood:
> I agree with @JelleZijlstra.

Okay — thanks, both!

PR #7187, opened 2022-02-13
Make `asyncio.transports.BaseTransport` generic

AlexWaygood:
The return type of `asyncio.proactor_events._ProactorBasePipeTransport.get_protocol()` is currently always inferred by type-checkers to be `BaseProtocol`, but we can be much more specific than that if we make the classes in `asyncio.transports` generic.


AlexWaygood:
Okay, that looks far too disruptive :(

PR #7191, opened 2022-02-13
`poplib.POP3_SSL.stls` unconditionally raises an exception

AlexWaygood:
Source code here: https://github.com/python/cpython/blob/0ae40191793da1877a12d512f0116d99301b2c51/Lib/poplib.py#L460

We have to keep the method in the stub, as it's overriding a method in the superclass that doesn't unconditionally raise.

### DIFF
--- a/stdlib/poplib.pyi
+++ b/stdlib/poplib.pyi
@@ -1,6 +1,6 @@
 import socket
 import ssl
-from typing import Any, BinaryIO, Pattern, overload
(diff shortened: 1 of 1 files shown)


JelleZijlstra:
This would be better fixed with python/typing#1043, but I guess we can go with NoReturn for now.

AlexWaygood:
> This would be better fixed with [python/typing#1043](https://github.com/python/typing/issues/1043), but I guess we can go with NoReturn for now.

My thoughts precisely.

PR #7193, opened 2022-02-13
`cmd.Cmd.default` returns `None`, not `bool`

AlexWaygood:
Source code here: https://github.com/python/cpython/blob/b74b1f36993a4e1700869133f3be13dd60ef4e40/Lib/cmd.py#L229


JelleZijlstra:
The onecmd method returns the return value of self.default, so its return annotation of `bool` may also be wrong.

JelleZijlstra:
Maybe; I'm not familiar with this module so there might be some expected behavior we're missing.

AlexWaygood:
> Maybe; I'm not familiar with this module so there might be some expected behavior we're missing.

Can we leave `onecmd` and stick with just fixing `default` for now, then? That one seems more clear-cut :)

srittau (approved):

PR #7200, opened 2022-02-14
Make `pydoc.Doc` an abstract class

AlexWaygood:
The `pydoc.Doc` class is entirely undocumented, but it seems like it is intended to be an abstract base class for internal use within the module. All of the `doc*` methods raise exceptions, several are called freely by other methods in the `Doc` class and all are overridden in `Doc` subclasses in the `pydoc` module.


stdlib/pydoc.pyi
```diff
@@ -42,11 +43,17 @@ class Doc:
...
+    @abstractmethod
     def docclass(self, object: object, name: str | None = ..., *args: Any) -> str: ...
+    @abstractmethod
     def docroutine(self, object: object, name: str | None = ..., *args: Any) -> str: ...
+    @abstractmethod
     def docother(self, object: object, name: str | None = ..., *args: Any) -> str: ...
+    @abstractmethod
     def docproperty(self, object: object, name: str | None = ..., *args: Any) -> str: ...
+    @abstractmethod
     def docdata(self, object: object, name: str | None = ..., *args: Any) -> str: ...
```
srittau, line 57: At least in 3.9, these are just aliases of `fail`. Why are they treated differently?

AlexWaygood: The fact that these methods are called freely from `doc.Document`, despite the fact that they all raise `TypeError` (due to being aliases of `fail` in the `Doc` class), and without an `except TypeError` guard, indicates to me that the expectation is that they will be overridden in subclasses: https://github.com/python/cpython/blob/15ee55528e1cbc47ef7e9f64186393eee40a49d9/Lib/pydoc.py#L472-L474

And, indeed, all of these methods are overridden in the two subclasses within `pydoc`: `HTMLDoc` and `TextDoc`.

PR #7202, opened 2022-02-14
stdlib: Add several missing comparison methods

AlexWaygood:


stdlib/tracemalloc.pyi
```diff
@@ -41,6 +41,10 @@ class Frame:
     filename: str
     lineno: int
     def __init__(self, frame: _FrameTupleT) -> None: ...
+    def __lt__(self, other: Frame) -> bool: ...
+    def __gt__(self, other: Frame) -> bool: ...
```
JelleZijlstra, line 45:
```suggestion
    def __le__(self, other: Frame) -> bool: ...
```

stdlib/collections/__init__.pyi
```diff
@@ -266,6 +270,10 @@ class Counter(dict[_T, int], Generic[_T]):
     def __ior__(self: Self, other: Counter[_T]) -> Self: ...  # type: ignore[override]
     if sys.version_info >= (3, 10):
         def total(self) -> int: ...
+        def __le__(self, other: Counter[object]) -> bool: ...
```
JelleZijlstra, line 273: I think we should use Counter[Any] here. It doesn't work with *all* Counters, only with Counters of a type that's comparable to our type. That's something we can't express in the current type system, so we should use Any.

stdlib/tracemalloc.pyi
```diff
@@ -41,6 +41,15 @@ class Frame:
     filename: str
     lineno: int
     def __init__(self, frame: _FrameTupleT) -> None: ...
+    def __lt__(self, other: Frame) -> bool: ...
```
JelleZijlstra, line 44: Confirmed that `__lt__` for some reason does not have the NotImplemented trick.

stdlib/collections/__init__.pyi
AlexWaygood:
```python
>>> from collections import Counter
>>> Counter('abcde') < Counter((1, 2, 3))
False
```

What am I missing? :)

stdlib/tracemalloc.pyi
AlexWaygood: This is why I used the `sys.version_info` guards, FWIW: https://github.com/python/cpython/pull/30818

stdlib/collections/__init__.pyi
JelleZijlstra: Ah right, the type parameter is the keys, but it compares the values, which are always ints.

This does make for some fun behavior:
```
In [16]: Counter((1, 2, 3)) < Counter('abcde')
Out[16]: False

In [17]: Counter('abcde') < Counter((1, 2, 3))
Out[17]: False
```

Issue #3898, opened 2020-04-01
@singledispatchmethod fails mypy validation


cosenal:
I am trying to validate the following example straight [from the docs](https://docs.python.org/3/library/functools.html#functools.singledispatchmethod) -- Here I renamed the two methods `_` from the example to `_int` and `_bool` to obviate an issue already reported, i.e., https://github.com/python/mypy/issues/2904 
```python
from functools import singledispatchmethod

class Negator:
    @singledispatchmethod
    def neg(self, arg):
        raise NotImplementedError("Cannot negate a")

    @neg.register
    def _int(self, arg: int):
        return -arg

    @neg.register
    def _bool(self, arg: bool):
        return not arg

n = Negator()
n.neg(2)
n.neg(True)
```

When I run mypy checks on this file, I get the following errors:
```
negator.py:17: error: "singledispatchmethod[Any]" not callable
negator.py:18: error: "singledispatchmethod[Any]" not callable
Found 2 errors in 1 file (checked 1 source file)
```

Python 3.8.2
mypy 0.770

hauntsaninja:
Thanks for the report! Linked PR fixes your repro for me locally.

PR #7203, opened 2022-02-14
Add missing descriptor methods in `dataclasses` and `functools`

AlexWaygood:


AlexWaygood:
I have confirmed locally that the test case in #3898 still passes with this PR applied.

(The fix for #3898 was to add `singledispatchmethod.__call__`, but `__call__` doesn't actually exist at runtime; `__get__`, on the other hand, does.)

PR #7206, opened 2022-02-14
Add `EnumMeta.__bool__`

AlexWaygood:
Most Python objects evaluate as falsey if they have length 0, but an enum class is truthy even if it has length 0.

Source code: https://github.com/python/cpython/blob/841c77d802e9ee8845fa3152700474021efe03fd/Lib/enum.py#L353


JelleZijlstra:
I hope you'll contribute whatever script you are using to stubtest at some point :)

PR #7207, opened 2022-02-14
`collections`: Add missing reflected BinOp methods

AlexWaygood:
[Testing confirms](https://mypy-play.net/?mypy=latest&python=3.10&gist=a72f03e007c92a9eabe03a46b25f7df8) that mypy currently raises false positives for these operations, whereas they all succeed at runtime.


stdlib/collections/__init__.pyi
```diff
@@ -129,8 +131,11 @@ class UserString(Sequence[UserString]):
     def __iter__(self: Self) -> Iterator[Self]: ...
     def __reversed__(self: Self) -> Iterator[Self]: ...
     def __add__(self: Self, other: object) -> Self: ...
+    def __radd__(self: Self, other: object) -> Self: ...
     def __mul__(self: Self, n: int) -> Self: ...
+    def __rmul__(self: Self, n: int) -> Self: ...
     def __mod__(self: Self, args: Any) -> Self: ...
+    def __rmod__(self: Self, template: object) -> Self: ...
```
JelleZijlstra, line 138: This one is actually broken:

```
In [12]: [] % collections.UserString("x")

  File "/main_instance_shell/jelle/venv/lib/python3.6/site-packages/IPython/core/interactiveshell.py", line 3343, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-12-0ca5ac41c5a2>", line 1, in <module>
    [] % collections.UserString("x")
  File "/main_instance_shell/jelle/venv/lib/python3.6/collections/__init__.py", line 1162, in __rmod__
    return self.__class__(format % args)
NameError: name 'args' is not defined
```

It should error but not with a NameError.

JelleZijlstra: Go ahead! I think this method just shouldn't exist. It's presumably there to support %-formatting, but I don't think a reflected method makes sense for that.

AlexWaygood: Wait, I can't reproduce the `NameError`. This is what I get on my machine:

```python
>>> from collections import UserString
>>> [] % UserString('foo')
'[]'
>>> [] % UserString('x')
'[]'
>>>
```

JelleZijlstra: Ah, it got fixed in https://bugs.python.org/issue25652. That's what I get for trying things out on 3.6. I admit I don't understand the reasoning they ended up with for keeping the method.

stdlib/collections/__init__.pyi
```diff
@@ -129,8 +131,15 @@ class UserString(Sequence[UserString]):
...
     def __reversed__(self: Self) -> Iterator[Self]: ...
     def __add__(self: Self, other: object) -> Self: ...
+    def __radd__(self: Self, other: object) -> Self: ...
     def __mul__(self: Self, n: int) -> Self: ...
+    def __rmul__(self: Self, n: int) -> Self: ...
     def __mod__(self: Self, args: Any) -> Self: ...
+    if sys.version_info >= (3, 8):
+        def __rmod__(self: Self, template: object) -> Self: ...
+    else:
+        def __rmod__(self: Self, template: Any) -> Self: ...
```
JelleZijlstra, line 141:
```suggestion
        def __rmod__(self: Self, format: Any) -> Self: ...
```
Judging from commit 7abf8c60819d5749e6225b371df51a9c5f1ea8e9 in CPython

JelleZijlstra (approved):

JelleZijlstra:
The mypy-primer errors are here: https://github.com/sphinx-doc/sphinx/blob/4.x/sphinx/locale/__init__.py#L67. They're already type-ignoring similar issues with `__add__`, so it seems fine to produce the same error on `__radd__`.

stdlib/collections/__init__.pyi
JelleZijlstra:
```
In [17]: collections.UserString("x") + collections.UserList(["y"])
Out[17]: "x['y']"

In [18]: collections.UserList(["y"]) + collections.UserString("x")
Out[18]: ['y', 'x']
```

AlexWaygood: Truly a nightmare of what Python could have been

PR #7209, opened 2022-02-14
`optparse.check_choice` returns `value` type

sobolevn:
Source: https://github.com/python/cpython/blob/ad4e8d2b871c4f5bce27520627bbb2e0e544bc24/Lib/optparse.py#L435-L437

If the `value` type is correct 🙂


hauntsaninja (approved):
Thanks!

PR #6769, opened 2021-12-31
fix(stdlib/types): FrameType.f_lineno can be None

jd:
In very rare (error?) cases, f_lineno can actually be None.

PR #7210, opened 2022-02-14
Improve `traceback.FrameSummary`

AlexWaygood:
- Add missing `__len__` method.
- `lineno` [can be `None`](https://github.com/python/cpython/blob/b27195332e91e932501f16cf9877761b218a9c99/Lib/traceback.py#L304)
- `line` is a read-only property that can be `str` [or `None`](https://github.com/python/cpython/blob/b27195332e91e932501f16cf9877761b218a9c99/Lib/traceback.py#L305), not a `str` attribute.
- Improve `__getitem__` with overloads.
- Reflect changes to the class that have been made in 3.11 (new `__init__` signature, new instance attributes).

Source code here: https://github.com/python/cpython/blob/b27195332e91e932501f16cf9877761b218a9c99/Lib/traceback.py#L243


AlexWaygood:
> 
> ```diff
> + homeassistant/util/async_.py:158: error: Item "None" of "Optional[str]" has no attribute "strip"  [union-attr]
> + homeassistant/util/async_.py:164: error: Item "None" of "Optional[str]" has no attribute "strip"  [union-attr]
> + homeassistant/helpers/frame.py:112: error: Item "None" of "Optional[str]" has no attribute "strip"  [union-attr]
> ```

Arguably these are true positives: `line` [can clearly be `None`](https://github.com/python/cpython/blob/b27195332e91e932501f16cf9877761b218a9c99/Lib/traceback.py#L305)

JelleZijlstra (approved):

hauntsaninja:
@AlexWaygood any idea under what circumstances lineno can be None? It seems like there was a case of lineno being None and it was treated as a bug and fixed, https://github.com/python/cpython/issues/88575. There's a chance that what's in traceback.py is overly defensive programming.

hauntsaninja:
In practice it looks like lineno's in traceback come from https://github.com/python/cpython/blob/b27195332e91e932501f16cf9877761b218a9c99/Lib/traceback.py#L329 which we have as an int https://github.com/python/typeshed/blob/e3d4bdc91a792bcef5d8eae2c1abab3d4ba27541/stdlib/types.pyi#L520

AlexWaygood:
https://github.com/python/cpython/blob/b27195332e91e932501f16cf9877761b218a9c99/Lib/traceback.py#L305 _looks_ pretty reachable to me, but I'll confess I don't know what real-life code ends up with us getting there :)

hauntsaninja:
So should we be changing TracebackType.tb_lineno?

AlexWaygood:
I don't know. I'm not an exception-handling expert, and the [language reference](https://docs.python.org/3/reference/datamodel.html#traceback-objects) doesn't shed much light here. Feel free to revert the `int` -> `int | None` changes if you're confident I'm incorrect here :)

Akuli:
This might have something to do with #6769. Presumably traceback line numbers come from stack frames.

PR #7213, opened 2022-02-14
Add `@final` to several stdlib classes that cannot be subclassed at runtime

AlexWaygood:


hauntsaninja (commented):
It looks like this might be a subset of the ones found in https://github.com/python/mypy/pull/12091 ? Or am I missing something

AlexWaygood:
> It looks like this might be a subset of the ones found in [python/mypy#12091](https://github.com/python/mypy/pull/12091) ? Or am I missing something

Nope, you're exactly right -- I'm working on a patch for stubtest so I'm using a tweaked version of stubtest on the master branch :)

hauntsaninja:
Definitely worth fixing, I was just curious about the subset aspect, e.g. `http.HTTPStatus`, `xxlimited.Xxo` are missing

AlexWaygood:
> Definitely worth fixing, I was just curious about the subset aspect, e.g. `http.HTTPStatus`, `xxlimited.Xxo` are missing. So was wondering if there were false positives or something

- `http.HTTPStatus` didn't end up in the original diff on https://github.com/python/mypy/pull/12091, because it's an `Enum` (so implicitly `@final`; stubtest raising an error would be a false positive there).
- `xxlimited` isn't available on Windows, so stubtest won't pick it up on my machine :)

hauntsaninja:
Okay, sg, I can add the unix ones!

PR #7215, opened 2022-02-14
Add `@final` to several more stdlib classes

hauntsaninja:


stdlib/typing_extensions.pyi
```diff
@@ -150,6 +152,7 @@ NotRequired: _SpecialForm
 LiteralString: _SpecialForm
 Unpack: _SpecialForm
 
+@_final
```
AlexWaygood, line 155: Can we not use `typing_extensions.final` inside `typing_extensions`? We use `typing_extensions.runtime_checkable` in `typing_extensions`

hauntsaninja: The issue is I used `typing.final`, which we can't use in `typing_extensions` on 3.7 and below.
I think I could use `typing_extensions.final` inside `typing_extensions` and reorder the stubs, but that seemed less nice. E.g., we'd have to separate the definitions of `final` and `Final`.

AlexWaygood (approved):

PR #7219, opened 2022-02-15
Optional output for `simpledialog`

### LINKED ISSUE #7218: Types for ask{string, float, int} should be optional
When `simpledialog.askstring` is closed without entering the data, the returned value is `None`, so all the `ask*` functions should have `Optional` output.

The same applies for:
- `simpledialog.askfloat`
- `simpledialog.askinteger`

matiuszka:
Fixes #7218


Akuli (approved):
Thanks!

PR #7221, opened 2022-02-15
Fix positional-only differences in `sys`

AlexWaygood:


stdlib/sys.pyi
```diff
@@ -215,8 +215,8 @@ def _clear_type_cache() -> None: ...
 def _current_frames() -> dict[int, FrameType]: ...
 def _getframe(__depth: int = ...) -> FrameType: ...
 def _debugmallocstats() -> None: ...
-def __displayhook__(value: object) -> None: ...
-def __excepthook__(type_: type[BaseException], value: BaseException, traceback: TracebackType | None) -> None: ...
+def __displayhook__(__value: object) -> None: ...
+def __excepthook__(__type_: type[BaseException], __value: BaseException, __traceback: TracebackType | None) -> None: ...
```
JelleZijlstra, line 219:
```suggestion
def __excepthook__(__exctype_: type[BaseException], __value: BaseException, __traceback: TracebackType | None) -> None: ...
```
Doesn't really matter but this is what the Argument Clinic annotations call it

PR #7222, opened 2022-02-15
Fix positional-only differences in `sqlite3`

AlexWaygood:


stdlib/sqlite3/dbapi2.pyi
```diff
@@ -183,7 +183,7 @@ class Connection:
 
     def __call__(self, *args: Any, **kwargs: Any) -> Any: ...
     def __enter__(self: Self) -> Self: ...
-    def __exit__(self, t: type | None, exc: BaseException | None, tb: Any | None) -> None: ...
+    def __exit__(self, __t: type | None, __exc: BaseException | None, __tb: Any | None) -> None: ...
```
JelleZijlstra, line 186:
```suggestion
    def __exit__(self, __type: type | None, __value: BaseException | None, __traceback: Any | None) -> Literal[False]: ...
```
Match AC names, and it does `Py_RETURN_FALSE`.

stdlib/sqlite3/dbapi2.pyi
```diff
@@ -233,16 +233,16 @@ class ProgrammingError(DatabaseError): ...
 class Row:
     def __init__(self, *args: Any, **kwargs: Any) -> None: ...
     def keys(self): ...
-    def __eq__(self, other): ...
-    def __ge__(self, other): ...
-    def __getitem__(self, index): ...
-    def __gt__(self, other): ...
+    def __eq__(self, __other): ...
+    def __ge__(self, __other): ...
```
JelleZijlstra, line 237: Arguably this shouldn't exist. The type defines the tp_richcompare slot, which I believe automatically gets mapped to all six Python-level dunders, but it always returns NotImplemented for anything other than EQ and NE. As a result, `sorted(list[Row])` doesn't work, but a type checker will allow it. Anyway, that's a problem for another day.

JelleZijlstra (approved):

PR #7224, opened 2022-02-15
Fix positional-only differences in `_tkinter`

AlexWaygood:


stdlib/_tkinter.pyi
```diff
@@ -20,12 +20,12 @@ class Tcl_Obj:
     string: str | bytes
```
JelleZijlstra, line 20: Unrelatedly, these appear to be read-only in the code

JelleZijlstra (approved):

PR #7227, opened 2022-02-15
Fix positional-only differences in `threading.local`

AlexWaygood:
This is the last one of these, I think.


JelleZijlstra (approved):

JelleZijlstra:
There might be a few more, e.g. `MethodDescriptorType.__get__`.

PR #7230, opened 2022-02-15
Add overload annotation for stripe DeletableAPIResource delete classmethod

yejia-stripe:
Fixes stripe/stripe-python#770

`stripe`'s `DeletableAPIResource.delete` [instance method](https://github.com/stripe/stripe-python/blob/cf794375df6e1d1adaef5ff5d12b9bcb31a845c0/stripe/api_resources/abstract/deletable_api_resource.py#L14) has a [wrapper](https://github.com/stripe/stripe-python/blob/cf794375df6e1d1adaef5ff5d12b9bcb31a845c0/stripe/util.py#L213) that effectively overloads the method to add a class method version.

Both
```python
s = stripe.Subscription.create(...)
s.delete()
```
and
```python
s = stripe.Subscription.create(...)
stripe.Subscription.delete(s.id)
```
should work, but the latter raises a type error.

### DIFF
--- a/stubs/stripe/stripe/api_resources/abstract/deletable_api_resource.pyi
+++ b/stubs/stripe/stripe/api_resources/abstract/deletable_api_resource.pyi
@@ -4,9 +4,5 @@ from typing import overload
 from stripe.api_resources.abstract.api_resource import APIResource as APIResource
 
 class DeletableAPIResource(APIResource):
-    @overload
     @classmethod
-    def delete(cls: type[Self], **params) -> Self: ...
-    @overload
-    @classmethod
-    def delete(cls: type[Self], sid: str, **params) -> Self: ...
+    def delete(cls: type[Self], sid: str = ..., **params) -> Self: ...
--- a/stubs/stripe/stripe/api_resources/abstract/deletable_api_resource.pyi
+++ b/stubs/stripe/stripe/api_resources/abstract/deletable_api_resource.pyi
@@ -1,5 +1,4 @@
 from _typeshed import Self
-from typing import overload
 
 from stripe.api_resources.abstract.api_resource import APIResource as APIResource
 
--- a/stubs/stripe/stripe/api_resources/abstract/deletable_api_resource.pyi
+++ b/stubs/stripe/stripe/api_resources/abstract/deletable_api_resource.pyi
@@ -3,7 +3,7 @@ from typing import TypeVar, overload
 
 from stripe.api_resources.abstract.api_resource import APIResource as APIResource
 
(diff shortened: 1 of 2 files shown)


stubs/stripe/stripe/api_resources/abstract/deletable_api_resource.pyi
```diff
@@ -1,4 +1,13 @@
+from _typeshed import Self
+from typing import TypeVar, overload
+
 from stripe.api_resources.abstract.api_resource import APIResource as APIResource
 
+_T = TypeVar("_T", bound="DeletableAPIResource")
```
JelleZijlstra, line 6:
```suggestion
_T = TypeVar("_T", bound=DeletableAPIResource)
```

JelleZijlstra:
I fixed the lint error, but it looks like pytype is unhappy about having `@classmethod` on only one of the two overloads. Could you rewrite this so the decorators are consistent?

JelleZijlstra:
Actually there's another lint error too—you should use `_typeshed.Self`.

yejia-stripe:
> I fixed the lint error, but it looks like pytype is unhappy about having `@classmethod` on only one of the two overloads. Could you rewrite this so the decorators are consistent?

@JelleZijlstra I want to fake having a class method and an instance method with the same name, which is weird and it makes sense that `pytype` complains about it. I'm not sure how to both express that without using `@classmethod` for just one of the overloads. Do you have suggestions? Thanks!

JelleZijlstra:
Would annotating it as only a classmethod work? Classmethods can still be called on instances.

srittau:
I think it's easiest to add the following to `@tests/stripe/stubtest_allowlist.txt`: `stripe\..*\.delete` to fix CI.

stubs/stripe/stripe/api_resources/abstract/deletable_api_resource.pyi
```diff
@@ -1,4 +1,12 @@
...
 from stripe.api_resources.abstract.api_resource import APIResource as APIResource
 
 class DeletableAPIResource(APIResource):
-    def delete(self, **params): ...
+    @overload
+    @classmethod
+    def delete(cls: type[Self], **params) -> Self: ...
+    @overload
+    @classmethod
+    def delete(cls: type[Self], sid: str, **params) -> Self: ...
```
JelleZijlstra, line 12:
```suggestion
    @classmethod
    def delete(cls: type[Self], sid: str = ..., **params) -> Self: ...
```
I don't think you even need overloads.

PR #7231, opened 2022-02-16
stdlib: Add many missing dunder overrides

AlexWaygood:
This PR proposes adding many missing dunder overrides in the stdlib.

Most of these have been omitted because they have similar signatures to the same method in a superclass. However, for all of these methods, the subclass override means that the method can be called with positional-or-keyword arguments, whereas the same method in the superclass can only be called with positional arguments.

It's not a _particularly_ important difference, but, with a patch I'm writing for stubtest (which has other, better effects!), it starts complaining about all of these.

For the vast majority of these, the signature is the same as without this patch applied (modulo the change from positional-only parameters to positional-or-keyword). However, there are a few exceptions where I spotted that a change in signature was required:
- `collections.Counter.__delitem__` (has a more permissive signature than `MutableMapping.__delitem__`)
- `multiprocessing.pool`
- `multiprocessing.managers`
- `shelve.Shelf.__contains__` (signature is more restrictive than `MutableMapping.__contains__`, violating LSP)


srittau (approved):
This a more precise, so a win in my eyes. I also like that we got rid of some `AbstractContextManager` inheritance.

PR #7235, opened 2022-02-16
Add `--disallow-untyped-decorators` to `mypy_test`

sobolevn:
This can possibly catch if we apply `(*Any, **Any) -> Any` to some existing typed function.
Right now, we don't do that anywhere in the code base 🙂


srittau (approved):
stricter = better

hauntsaninja (approved):

PR #7240, opened 2022-02-16
Add asyncio.taskgroups and new Task methods

gvanrossum:
The new Task methods are .cancelling() and .uncancel().

Everything is conditional to 3.11 (but the taskgroups.py module
always exists -- it's just empty before 3.11).


gvanrossum:
Question for other typeshed maintainers. How do we handle modules that were added in a certain release? I am getting errors from the linters that seem to be complaining about this.

(@srittau @JelleZijlstra )

JelleZijlstra:
You need to add it to the VERSIONS file, https://github.com/python/typeshed/blob/master/stdlib/VERSIONS

stdlib/asyncio/taskgroups.pyi
```diff
@@ -0,0 +1,17 @@
...
+from .tasks import Task
+
+_T = TypeVar("_T")
+
+if sys.version_info >= (3, 11):
+    class TaskGroup:
+        def __init__(self, name: str | None = ...) -> None: ...
+        def get_name(self) -> str: ...
+        def __aenter__(self) -> TaskGroup: ...
+        def __aexit__(self, et: type | None, exc: BaseException | None, tb: TracebackType | None) -> None: ...
```
JelleZijlstra, line 14:
```suggestion
        async def __aenter__(self) -> TaskGroup: ...
        async def __aexit__(self, et: type | None, exc: BaseException | None, tb: TracebackType | None) -> None: ...
```
(Don't have time for a full review yet)

gvanrossum:
These test failures are excruciating if you haven't been following developments here. :-)

I now get "error: asyncio.taskgroups failed to find stubs" e.g. https://github.com/python/typeshed/runs/5222298476?check_suite_focus=true

I'm not sure what it's telling me I have to do. :-(

Akuli:
`stubtest` is a tool that imports the module and compares it to stubs. Because `asyncio.taskgroups` can't actually be imported before 3.11, you need to suppress the error by adding a line `asyncio.taskgroups` to `py36.txt`, `py37.txt` etc in `tests/stubtest_allowlists/`.

stdlib/asyncio/taskgroups.pyi
```diff
@@ -0,0 +1,17 @@
...
+from types import TracebackType
+from typing import Any, Coroutine, Generator, TypeVar
+
+from .tasks import Task
+
+_T = TypeVar("_T")
+
+if sys.version_info >= (3, 11):
+    class TaskGroup:
+        def __init__(self, name: str | None = ...) -> None: ...
```
Akuli, line 11: The name is keyword-only:

```suggestion
        def __init__(self, *, name: str | None = ...) -> None: ...
```

stdlib/asyncio/taskgroups.pyi
```diff
@@ -0,0 +1,17 @@
...
+        def get_name(self) -> str: ...
+        async def __aenter__(self) -> TaskGroup: ...
```
Akuli, line 13: When called on a subclass of `TaskGroup`, it will return an instance of the subclass. While PEP 673 isn't available for us yet, we use a typevar defined in `_typeshed` instead:

```suggestion
        async def __aenter__(self: Self) -> Self: ...
```

You need to also add `from _typeshed import Self`.

stdlib/asyncio/taskgroups.pyi
```diff
@@ -0,0 +1,17 @@
+import sys
+from types import TracebackType
+from typing import Any, Coroutine, Generator, TypeVar
+
+from .tasks import Task
+
+_T = TypeVar("_T")
+
+if sys.version_info >= (3, 11):
```
Akuli, line 9: I believe this is unnecessary, because of the `VERSIONS` file. Other asyncio submodules don't have if statements like this.

JelleZijlstra:
I suppose we should make stubtest understand the VERSIONS file.

stdlib/asyncio/taskgroups.pyi
```diff
@@ -0,0 +1,17 @@
...
+    class TaskGroup:
+        def __init__(self, name: str | None = ...) -> None: ...
+        def get_name(self) -> str: ...
+        async def __aenter__(self) -> TaskGroup: ...
+        async def __aexit__(self, et: type | None, exc: BaseException | None, tb: TracebackType | None) -> None: ...
```
Akuli, line 14:
```suggestion
        async def __aexit__(self, et: type[BaseException] | None, exc: BaseException | None, tb: TracebackType | None) -> None: ...
```

Akuli (commented):
A few comments below. Unfortunately our CI isn't very helpful, because we don't run the tests on Python 3.11 yet.

AlexWaygood:
> I suppose we should make stubtest understand the VERSIONS file.

I believe that should already be fixed in the next version of mypy (https://github.com/python/mypy/pull/12083)

stdlib/asyncio/taskgroups.pyi
gvanrossum: Heh. I copied that literally from create_taks() in task.pyi. :-)

stdlib/asyncio/taskgroups.pyi
```diff
@@ -0,0 +1,15 @@
...
+from .tasks import Task
+
+_T = TypeVar("_T")
+
+class TaskGroup:
+    def __init__(self, name: str | None = ...) -> None: ...
+    def get_name(self) -> str: ...
+    async def __aenter__(self) -> TaskGroup: ...
+    async def __aexit__(self, et: type[BaseException] | None, exc: BaseException | None, tb: TracebackType | None) -> None: ...
+    def create_task(self, coro: Generator[Any, None, _T] | Coroutine[Any, Any, _T], *, name: str | None = ...) -> Task[_T]: ...
```
Akuli, line 15: Just noticed that the `name` argument doesn't seem to exist on cpython's `main` branch:

```suggestion
    def create_task(self, coro: Generator[Any, None, _T] | Coroutine[Any, Any, _T]) -> Task[_T]: ...
```

gvanrossum: Good catch! Fixed.

stdlib/asyncio/taskgroups.pyi
gvanrossum: Whoops, didn't see this before. Fixed now!

Akuli (approved):

gvanrossum:
Thanks!

PR #7241, opened 2022-02-16
xml: use SupportsRead/SupportsWrite instead of IO

Rogdham:
The consensus regarding typing file-like objects seems to be (at least for typeshed) [to use protocols instead of `IO`](https://github.com/python/typing/discussions/829).

This PR targets the `xml` module in the stdlib.

Non-straightforward changes:
 - `pulldom` works with non-binary file-like objects as well
 - `etree`'s `ElementTree.write_c14n` does not work with non-binary file-like objects
 - `sax`'s `parse` function requires the protocol to have a `close` method as well, so I added a subclass to `SupportsRead`

I have created a [gist with tests](https://gist.github.com/Rogdham/d5133e91015d1eb5f798d7b580be0532) demonstrating that the functions actually works with minimal implementations of the protocls.


stdlib/_typeshed/__init__.pyi
```diff
@@ -170,6 +170,9 @@ FileDescriptorLike = Union[int, HasFileno]  # stable
 class SupportsRead(Protocol[_T_co]):
     def read(self, __length: int = ...) -> _T_co: ...
 
+class SupportsReadClose(SupportsRead[_T_co]):
+    def close(self) -> None: ...
+
```
AlexWaygood, line 175: If this is only used in one file, it would be better to define it in `xml.sax` rather than adding it to `_typeshed` -- you can get stubtest to stop complaining about it not being there at runtime by calling it `_SupportsReadClose` instead of `SupportsReadClose` (stubtest knows that names starting with a single underscore aren't necessarily meant to exist at runtime)

Rogdham: Thank you, I've pushed a separated commit to do so!

stdlib/xml/sax/__init__.pyi
```diff
@@ -1,8 +1,12 @@
 import sys
-from typing import IO, Any, Iterable, NoReturn
+from _typeshed import SupportsRead, _T_co
+from typing import Any, Iterable, NoReturn
```
AlexWaygood, line 3:
```suggestion
from typing import Any, Iterable, NoReturn, Protocol
```

stdlib/xml/sax/__init__.pyi
```diff
@@ -1,8 +1,12 @@
...
 from xml.sax.handler import ContentHandler, ErrorHandler
 from xml.sax.xmlreader import Locator, XMLReader
 
+class _SupportsReadClose(SupportsRead[_T_co]):
```
AlexWaygood, line 7:
```suggestion
class _SupportsReadClose(SupportsRead[_T_co], Protocol[_T_co]):
```

AlexWaygood (commented):
You need to mix in `Protocol` again, or the class definition is interpreted as a concrete class inheriting from a protocol, rather than a protocol inheriting from a protocol :)

stdlib/xml/sax/__init__.pyi
Rogdham: Damn, good catch :+1:

JelleZijlstra:
Thanks for the PR and thanks @AlexWaygood for the review!

PR #7243, opened 2022-02-16
Add `EnumMeta.__prepare__`

AlexWaygood:
I don't really understand why this needs the `# type: ignore`s, but mypy seems absolutely convinced that this method breaks LSP!

### DIFF
--- a/stdlib/enum.pyi
+++ b/stdlib/enum.pyi
@@ -49,6 +49,13 @@ class EnumMeta(ABCMeta):
     else:
         def __new__(metacls: type[Self], cls: str, bases: tuple[type, ...], classdict: _EnumDict) -> Self: ...  # type: ignore
 
+    if sys.version_info >= (3, 9):
+        @classmethod
+        def __prepare__(metacls, cls: str, bases: tuple[type, ...], **kwds: Any) -> _EnumDict: ...  # type: ignore[override]
+    else:
+        @classmethod
+        def __prepare__(metacls, cls: str, bases: tuple[type, ...]) -> _EnumDict: ...  # type: ignore[override]
+
     def __iter__(self: type[_T]) -> Iterator[_T]: ...
     def __reversed__(self: type[_T]) -> Iterator[_T]: ...
     def __contains__(self: type[Any], member: object) -> bool: ...


hauntsaninja (commented):
It breaks LSP on 3.8 by not accepting `**kwds`, not sure about on 3.9 (will check the PR out locally later and see)

hauntsaninja:
Looking into the LSP complaint on 3.9:
```
stdlib/enum.pyi:85: error: Signature of "__prepare__" incompatible with supertype "type"  [override]
stdlib/enum.pyi:85: note:      Superclass:
stdlib/enum.pyi:85: note:          def __prepare__(metacls, str, Tuple[type, ...], **kwds: Any) -> Mapping[str, object]
stdlib/enum.pyi:85: note:      Subclass:
stdlib/enum.pyi:85: note:          @classmethod
stdlib/enum.pyi:85: note:          def __prepare__(metacls, cls: str, bases: Tuple[type, ...], **kwds: Any) -> _EnumDict
```

It goes away if you mark the args as positional-only, which seems incorrect:
```
diff --git a/stdlib/enum.pyi b/stdlib/enum.pyi
index 8ca0a5f8..0c4a2943 100644
--- a/stdlib/enum.pyi
+++ b/stdlib/enum.pyi
@@ -82,7 +82,7 @@ class EnumMeta(ABCMeta):
 
         @classmethod
-        def __prepare__(metacls, cls: str, bases: tuple[type, ...], **kwds: Any) -> _EnumDict: ...  # type: ignore[override]
+        def __prepare__(metacls, __cls: str, __bases: tuple[type, ...], **kwds: Any) -> _EnumDict: ...
     else:
         @classmethod
         def __prepare__(metacls, cls: str, bases: tuple[type, ...]) -> _EnumDict: ...  # type: ignore[override]
```
(Also the error message seems to drop the `classmethod` on the superclass, but based on tests I think that might just be a reporting issue). So I'm happy to chalk this up to mypy bug.

JelleZijlstra:
Thanks for checking! Wasn't there some bug where the `@classmethod` decorator in this message sometimes gets eaten by the cache? Anyway, this does look like a mypy bug.

hauntsaninja:
I mean I guess technically mypy has a point in that: `x.__prepare__(metaclass, cls, bases, cls=100)` will get you incorrect behaviour if it turns out your `x` is an `EnumMeta` and not a `type`. But I'm guessing the complaint might not be intentional.

Issue #6920, opened 2022-01-15
tkinter.Canvas.coords doesn't accept item id like it should


Akuli:
https://github.com/python/typeshed/blob/5d206358907c419802cd9afcfcc83cce4f01bebb/stdlib/tkinter/__init__.pyi#L1070-L1075

Looking at the `canvas` manual page, the first argument seems to be `tagOrId` which is missing here.

Akuli:
Here's a minimal-ish example that displays a circle at runtime and creates a mypy error when type checking:

```
from tkinter import Canvas
c = Canvas()
c.pack()
circle_id = c.create_oval(0,0,0,0)
c.coords(circle_id, (50, 50, 100, 100))
c.mainloop()
```

AlexWaygood:
@Akuli, has this been fixed by #7246?

Akuli:
Yes, this has been fixed.

PR #7246, opened 2022-02-17
Add missing tagOrId parameter to tkinter.Canvas.coords()

TeamSpen210:
As described by #6920, Tk's `canvas`  widget [requires a tag (string) or item ID](https://tcl.tk/man/tcl8.6/TkCmd/canvas.htm#M42) for the `coords()` method. This fixes that and also swaps some other int types to the alias for clarity. `coords()` actually takes `*args`, so these must be positional-only.


Akuli (approved):
Thanks!

I don't like `_CanvasItemId`, and I think we should just use `int` instead everywhere: `tagOrId: str | int` is very clear and readable. But that should be a separate PR.

TeamSpen210:
It's simpler yeah, the only compelling reason I can see for keeping it is to upgrade it to a `NewType` so you can distinguish IDs from regular ints. But that might be a little awkward considering the type wouldn't be available at runtime for declarations.

Akuli:
Indeed. We aren't switching to `NewType` any time soon, and if it later becomes a more reasonable thing to do, it isn't too much work to just change each method.

PR #7247, opened 2022-02-17
Update to redis 4.1.4

srittau:
Closes: #7245


hauntsaninja (approved):
Yay for catching upstream bugs! :-)

JelleZijlstra:
They added a whole new file in a bugfix release? I have to wonder what their versioning scheme is...

PR #7249, opened 2022-02-17
Adds `bdist_wininst` definition

sobolevn:
Source: https://github.com/python/cpython/blob/3.9/Lib/distutils/command/bdist_wininst.py

Removed in `3.10`


stdlib/@python2/distutils/config.pyi
```diff
@@ -13,5 +13,3 @@ class PyPIRCCommand(Command):
     boolean_options: ClassVar[list[str]]
     def initialize_options(self) -> None: ...
     def finalize_options(self) -> None: ...
-    @abstractmethod
```
sobolevn, line 16: I've also removed several redefinitions of `run` abstract method. It is defined in `Command`

stdlib/distutils/command/bdist_wininst.pyi
```diff
@@ -0,0 +1,19 @@
+import sys
+from _typeshed import StrOrBytesPath
+from typing import Any, ClassVar
+
+if sys.version_info < (3, 10):
```
srittau, line 5: Instead of using this guard, the file should be added to `VERSIONS`.

sobolevn: It is already there: https://github.com/python/typeshed/blob/5c309e5fe4a8d4898fe9f2b8664bb66cff845878/stdlib/VERSIONS#L116

sobolevn:
Ok, looks like this commit finally passes: https://github.com/python/typeshed/pull/7249/commits/ebca8105aed047f09cd313c82aba5a2b1d213b1c

I pushed one more to test things without `if sys.version_info`: https://github.com/python/typeshed/pull/7249/commits/d2506c827d7dda8c9f137af36d8f2cadb8cdf21a

sobolevn:
Yes, it failed 🎉

stdlib/distutils/command/bdist_wininst.pyi
hauntsaninja: I'm not sure I see the comments you're referring to, but cf https://github.com/python/mypy/pull/12083

sobolevn: https://github.com/python/typeshed/pull/7249#issuecomment-1042909330

JelleZijlstra:
We now run stubtest from mypy master with the VERSIONS fix, could you merge master and try removing the version check again?

sobolevn:
Done!

PR #7250, opened 2022-02-17
Initialize FeedParser generic argument

srittau:


srittau:
Oops, it seems I had an old `cast` in my project, but for some reason `--warn-redundant-casts` was not picked up. No idea what was going on there, but removing the cast works fine with mypy 0.931.

PR #7251, opened 2022-02-17
Remove `_NNTPBase` from `nntplib`

sobolevn:
It was always a protected name, but it was removed from CPython 2 years ago.
See https://github.com/python/cpython/blob/2945f5a7c51200bfc5c493ccb626fc414e1385b1/Lib/nntplib.py#L490

### DIFF
--- a/stdlib/nntplib.pyi
+++ b/stdlib/nntplib.pyi
@@ -35,7 +35,7 @@ def decode_header(header_str: str) -> str: ...
 
 _list = list  # conflicts with a method named "list"
 
-class NNTP:     
+class NNTP:
     encoding: ClassVar[str]
     errors: ClassVar[str]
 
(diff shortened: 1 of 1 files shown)


stdlib/nntplib.pyi
```diff
@@ -34,11 +35,13 @@ def decode_header(header_str: str) -> str: ...
 
 _list = list  # conflicts with a method named "list"
 
-class _NNTPBase:
-    encoding: str
-    errors: str
+class NNTP:
+    encoding: ClassVar[str]
+    errors: ClassVar[str]
```
Akuli, line 40: These might make more sense if typed as instance attributes, so that you don't have to subclass `NNTP` or `NNTP_SSL` just to change them.

stdlib/nntplib.pyi
```diff
@@ -34,11 +35,13 @@ def decode_header(header_str: str) -> str: ...
...
 
     host: str
+    port: int
+    sock: socket.socket
```
Akuli, line 44: In `NNTP_SSL`, the `sock` should be `ssl.SSLSocket`.

Akuli (approved):
Thanks!

PR #7253, opened 2022-02-17
Tag.attrs are Mutable

rgant:
I keep getting errors like: `"Mapping[str, str]" has no attribute "__delitem__"; maybe "__getitem__"?`

Real type is a python `dict`.

```py
>>> type(item.attrs)
<class 'dict'>
```

So I need to add `isinstance(tag.attrs, dict)` to force the type (maybe there is a better way? I haven't written python mypy in about 8 months)

For example, without `and isinstance(tag.attrs, dict)` this would result in the above error:
```py
# No custom styles
if 'style' in tag.attrs and isinstance(tag.attrs, dict):
    del tag.attrs['style']
```

### DIFF
--- a/stubs/beautifulsoup4/bs4/element.pyi
+++ b/stubs/beautifulsoup4/bs4/element.pyi
@@ -1,6 +1,6 @@
 from _typeshed import Self
 from collections.abc import Iterator
-from typing import Any, Callable, Generic, Iterable, Mapping, Pattern, TypeVar, Union, overload
+from typing import Any, Callable, Generic, Iterable, Pattern, TypeVar, Union, overload
 
 from . import BeautifulSoup
 from .builder import TreeBuilder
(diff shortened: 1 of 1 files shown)


Akuli (approved):
Thanks! Mutating the dict seems to actually work, so banning it doesn't make sense:

```
>>> import bs4
>>> s=bs4.BeautifulSoup('<span style="foo:bar;">blah</span>')
>>> s
<html><body><span style="foo:bar;">blah</span></body></html>
>>> s=s.html.body.span
>>> s
<span style="foo:bar;">blah</span>
>>> s.attrs
{'style': 'foo:bar;'}
>>> s.attrs.clear()
>>> s
<span>blah</span>
```

Akuli:
> So I need to add isinstance(tag.attrs, dict) to force the type (maybe there is a better way? I haven't written python mypy in about 8 months)

`assert` is useful for this:

```
tag = ...
assert isinstance(tag.attrs, dict)
tag.attrs["foo"] = "bar"
```

PR #6081, opened 2021-09-27
Improves BeautifulSoup4 typing.

cmckain:
PageElement.find_all() can return both PageElement and any subclass of PageElement on the latest version of bs4 and ResultSet already accounts for this so find_all() should be typed with the same generic TypeVar.

PR #7254, opened 2022-02-17
find_all ResultSet return type

rgant:
`Tag.find_all` should return a ResultSet of something descended from `PageElement`.

A `Tag` or a `NavigableString`, or a `Comment` likely, but the parent class is used in other places so this should be close enough.

### DIFF
--- a/stubs/beautifulsoup4/bs4/element.pyi
+++ b/stubs/beautifulsoup4/bs4/element.pyi
@@ -319,7 +319,7 @@ class Tag(PageElement):
         text: _Strainable | None = ...,
         limit: int | None = ...,
         **kwargs: _Strainable,
-    ) -> ResultSet[Any]: ...
+    ) -> ResultSet[PageElement]: ...
     __call__ = find_all
     findAll = find_all
     findChildren = find_all


Akuli:
The `Any` was deliberately added in #6081. It seems to be a bit controversial.

rgant:
`Tag.find_all` doesn't return `Any`thing, it returns some element decedent from PageElement. After that, the user needs to figure out what end type each item is (From [documentation](https://www.crummy.com/software/BeautifulSoup/bs4/doc/#kinds-of-objects) it expects to be either a `Tag` or a `NavigableString`, or a `Comment`. But I've also seen a `Script` type.)

I suppose you could make it `: ResultSet[typing.Union[Tag, NavigableString, Comment, Script, PreformattedString, CData, ProcessingInstruction,  So So Many More subclasses...]]` but then if a new `PageElement` decedent is added you would need to update the list. But it will never return an int or a bool or a BeautifulSoup, and that should be known.

The complaint in the referenced pull:
> A type var needs to be used twice in a definition (or it has to be used in a generic class), so this typevar has no meaning in this context. In this case, just using Any is preferred.

I feel is wrong since `Any` will allow (on mypy strict) silent and hidden typing issues that are hard to diagnose. The terrible thing about BeautifulSoup is, it's not designed well for typing, so every `.find_all` needs several `isinstance` to figure out what you are dealing with. Just like the returns from `find_all_next` and `find_next_siblings` and the rest that are typed `-> ResultSet[PageElement]`

Please please don't change those, they are working. Inheritance is a terrible thing in software, but one we must deal with.

Akuli:
Yeah, the typing is going to be pretty bad either way:
- If we use `Any`, type checkers won't warn about actual problems and you will likely get worse autocompletions.
- If we use `PageElement` or a union, you need to use `assert isinstance(foo, Tag)` or similar every time you deal with the return value. You really need to. Type checkers don't think "this code works with a `Tag` so it works with `Union[Tag, other stuff]`". Instead, whatever you do with a `Union` must work with all members of the union.

It would be awesome if https://github.com/python/typing/issues/566 was a thing :)

JelleZijlstra:
Maybe we should reconsider the "false negatives over false positives" stance. There's definitely a userbase who want types that are as precise as possible.

srittau:
Maybe we could start by having an `AnyOf[]` subclass of `Any` in `_typeshed`. Type checkers could experiment with that.

srittau:
Our sub-class could be.

Akuli:
We could return an unbound `TypeVar` to make type checkers tell users to add a type annotation. This way you can annotate the result however you like, and explicitly show what you assume it to be.

Here's a toy example:

```
from typing import TypeVar

T = TypeVar("T")
def func() -> T: ...
x = func()  # error: Need type annotation for "x"
```

It does require users to be explicit, but I think that's better than just returning `Any` in this situation.

PR #6610, opened 2021-12-16
Use Literal instead of int for logging levels

DevilXD:
According to PEP 586:

> Literal may also be parameterized by other literal types, or type aliases to other literal types. For example, the following is legal:
> ```py
> ReadOnlyMode         = Literal["r", "r+"]
> WriteAndTruncateMode = Literal["w", "w+", "wt", "w+t"]
> WriteNoTruncateMode  = Literal["r+", "r+t"]
> AppendMode           = Literal["a", "a+", "at", "a+t"]
> 
> AllModes = Literal[ReadOnlyMode, WriteAndTruncateMode,
>                    WriteNoTruncateMode, AppendMode]
> ```
> This feature is again intended to help make using and reusing literal types more ergonomic.

Having the above in mind, I've ran into an issue with MyPy while trying to implement an argument parser with arguments setting different logging levels, and trying to type a variable with `Literal[logging.NOTSET, logging.DEBUG]`. Because they're constants and are typed as `int` here, MyPy has rejected such a type as valid. Specifying them as `Literal` types instead should fix this issue.

I did consider this introducing a potential inconsistency problem, however I'd imagine it'd be quite unlikely for a user to overwrite the default logging level's values, and rather treat them as constants only.

PR #7259, opened 2022-02-18
Use Final instead of Literal in logging

srittau:


AlexWaygood:
An encouraging primer output, however

hauntsaninja:
mypy_primer did tell us this was going to be a problem in https://github.com/python/typeshed/pull/6610 and we merged, acknowledging that we might want to revert later. Maybe this question isn't meaningful, but do we feel like "this is things going to plan: we tried a thing, it didn't work" or should we update in the direction of being less likely to merge past primer issues?

srittau:
Closing due to the discussion in #7349.

PR #7261, opened 2022-02-18
Accept None filename in linecache

JukkaL:
None is accepted as the file name:

```
>>> import linecache
>>> linecache.getline(None, 0)
''
>>> linecache.getlines(None)
[]
>>> linecache.updatecache(None)
[]
```


srittau:
This seems to be a side-effect of the `if not filename` check in `updatecache`. While passing `None` is accepted, this seems like a potential pitfall to me that a type checker could catch. I'm fine with merging this, if you think this really is a valid use case, but I want to make sure.

JukkaL:
I thought I had a valid use case, but actually the call site seems wrong, so I'm closing this.

PR #7264, opened 2022-02-19
Add `asyncio.base_events.__all__`

AlexWaygood:


stdlib/asyncio/base_events.pyi
```diff
@@ -14,6 +14,8 @@ from typing_extensions import Literal
 if sys.version_info >= (3, 7):
     from contextvars import Context
 
+__all__ = ["BaseEventLoop"]
```
JelleZijlstra, line 17:
```suggestion
__all__ = ("BaseEventLoop",)
```
stubtest being a pedant

AlexWaygood: Thanks!

JelleZijlstra:
I'd be open to making stubtest not care whether `__all__` is a list or a tuple.

JelleZijlstra (approved):

AlexWaygood:
> I'd be open to making stubtest not care whether `__all__` is a list or a tuple.

More problematic imo is the fact that stubtest doesn't check whether the elements within `__all__` are correct or not. I was trying to find a way today to patch it so that it would, but couldn't figure out how to retrieve that information from the `Var` node mypy generates.

PR #7266, opened 2022-02-19
Add `asyncio.events.__all__`

AlexWaygood:


JelleZijlstra:
Looks like stubtest does tell us if the item count is wrong.

AlexWaygood:
> Looks like stubtest does tell us if the item count is wrong.

But, I think only if `__all__` is a tuple(!)

hauntsaninja (approved):

PR #7275, opened 2022-02-19
Add `asyncio.streams.__all__`

AlexWaygood:


JelleZijlstra:
Do we need a sys.platform check around `__all__`? What do type checkers do if you `import *` from this module on windows?

AlexWaygood:
> Do we need a sys.platform check around `__all__`? What do type checkers do if you `import *` from this module on windows?

I was going off https://typing.readthedocs.io/en/latest/source/stubs.html#all, which states:

> If the runtime dynamically adds or removes elements (for example if certain functions are only available on some platforms), include all possible elements in the stubs.

But, it seems like stubtest is not a fan of those docs.

PR #7279, opened 2022-02-19
Add `asyncio.base_futures.__all__`

AlexWaygood:


JelleZijlstra:
Do you want to cut a new flake8-pyi release? I can do it too.

AlexWaygood:
> Do you want to cut a new flake8-pyi release? I can do it too.

No, it's fine, I can do it! (Do you think we need to bother with a release-candidate version this time, or can we skip it, do you think?)

JelleZijlstra:
Thanks! We can skip the RC. I did that last time because the release process automation was new and I wasn't sure it was going to work.

AlexWaygood:
Who knew adding an empty `__all__` to a stub could be this complicated 🤯

PR #7280, opened 2022-02-19
Add `asyncio.tasks.__all__`

AlexWaygood:


stdlib/asyncio/tasks.pyi
```diff
@@ -11,6 +11,45 @@ from .futures import Future
...
+    __all__ = [
+        "Task",
+        "FIRST_COMPLETED",
+        "FIRST_EXCEPTION",
+        "ALL_COMPLETED",
+        "wait",
+        "wait_for",
+        "as_completed",
+        "sleep",
+        "async",
```
JelleZijlstra, line 46: async exists but we may have to remove it. It's omitted from the stub because the name is a syntax error in later versions. Let's see what CI says.

PR #7283, opened 2022-02-19
create_baseline_stubs: improve package detection

Rogdham:
In current version, the `create_baseline_stubs.py` script defaults to the project name as the package name. This is not always the best choice (e.g. `PyYAML` and `yaml`), and the `--package` can be used in that case.

In this PR, I'm improving the default package detection when there is no `--package` flag. Because this script is run manually, I feel like it is ok to display a message saying to use `--package` flag if the detected package is not the wanted one.

I checked with the following projects:
 - `redis` defaults to `redis`
 - `python-nmap` defaults to `nmap`
 - `PyYAML` defaults to `yaml` (note that `PyYAML` also exports a `_yaml` module)

The improvement relies on `importlib.metadata` module, and fallbacks to current default (i.e. project name) if that fails or for Python < 3.8 (where that module does not exist).

The code for package detection is slightly inspired by [`packages_distributions`](https://github.com/python/cpython/blob/3.10/Lib/importlib/metadata/__init__.py#L1004) in Python >= 3.10.

### DIFF
--- a/scripts/create_baseline_stubs.py
+++ b/scripts/create_baseline_stubs.py
@@ -16,6 +16,9 @@ import subprocess
 import sys
 from typing import Optional, Tuple
 
+if sys.version_info >= (3, 8):
+    from importlib.metadata import distribution
+
 PYRIGHT_CONFIG = "pyrightconfig.stricter.json"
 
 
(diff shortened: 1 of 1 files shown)


hauntsaninja (approved):
Thanks!

PR #7306, opened 2022-02-19
Update stubtest for submodule versions, typeddict fix

### LINKED ISSUE #7305: Use latest stubtest from git in CI
There have been a number of great additions to stubtest recently. Several of these introduce new errors in typeshed. This means whenever we next update stubtest, there'll be a large backlog of allowlist adjustments or regression fixes — often duplicating work done when validating the change to stubtest. Lack of support for submodules in VERSIONS has also recently confused two contributors.

When stubtest was new, we used to install it from git (with a pinned hash) and this let us iterate quickly (and before that, for the longest time, we used to run all of mypy unpinned from latest git master). The only real downside of this is that stubtest is marginally (not too noticeably) slower. If we really cared about speed, we could install from https://github.com/mypyc/mypy_mypyc-wheels.

There've been about ten commits to stubtest since 0.930, we can upgrade commits one by one if necessary.

cc @AlexWaygood (if this is something you'd be interested in!)

hauntsaninja:
First step in #7305

### DIFF
--- a/.github/workflows/stubtest.yml
+++ b/.github/workflows/stubtest.yml
@@ -1,4 +1,4 @@
-name: Run stubtest on all stubs
+name: Daily test
 
 on:
   workflow_dispatch:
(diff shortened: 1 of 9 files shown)


AlexWaygood:
`platform.uname_result` is [a mess in CPython](https://github.com/python/cpython/blob/fc115c9bde52a58b4fb9be2b80c3d6be8a4d2454/Lib/platform.py#L772), so I think `platform.uname_result.__match_args__` should just be allowlisted!

hauntsaninja:
Yeah, I agree. I guess it's complaining because mypy learnt some `__match_args__` tricks on master

AlexWaygood (approved):
LGTM!

PR #7308, opened 2022-02-20
heapq: re-export, add __about__

hauntsaninja:


AlexWaygood:
Looks like `_heapq` should be added to the list here: https://github.com/python/typeshed/blob/879b6173662d77c0569854bc513f6ff0ee873926/pyproject.toml#L12

hauntsaninja:
Why does `from _heapq import *` lead to pyright import `_T`? `import *` doesn't import symbols that start with underscore. Or is there some special stub rule that I'm forgetting about.

JelleZijlstra:
Right, seems like a pyright bug.

AlexWaygood:
For now we can just rename one of the TypeVars though, right?

hauntsaninja:
It'd take me more time to rename the TypeVar than to tag @erictraut and have him make a pyright release :D

erictraut:
I'm on vacation this week, so replies may be slower than normal.

Pyright's current behavior is:
1. If a symbol is declared within a stub file and its name starts with an underscore, it is considered visible to other stubs or code files that import from it. For example, if `stub.pyi` includes the declaration `_T = TypeVar("_T")`, it is legal for a source file to do `from stub import _T`. In other words, names beginning with an underscore are visible outside of the stub.
2. If someone performs a wildcard import from a stub and the stub contains a dunder all definition, only the symbols within the dunder all are imported.
3. If someone performs a wildcard import from a stub and the stub does not contain a dunder all, all visible symbols (including those that begin with an underscore) are imported.

It sounds like you're suggesting that rule 3 is wrong? Is there a special-case exemption for wildcard imports from a stub file with regard to names that start with an underscore? If so, why? Is this special case documented somewhere?

JelleZijlstra:
This is not specific to stubs. If there is no `__all__`, `import *` excludes names that start with an underscore. This is documented in https://docs.python.org/3.10/reference/simple_stmts.html#the-import-statement (paragraph starting "The public names defined by a module").

JelleZijlstra:
(And enjoy your vacation! This bug can wait; it's easy to work around anyway.)

stdlib/heapq.pyi
```diff
@@ -1,14 +1,12 @@
...
 
-def heappush(__heap: list[_T], __item: _T) -> None: ...
-def heappop(__heap: list[_T]) -> _T: ...
-def heappushpop(__heap: list[_T], __item: _T) -> _T: ...
-def heapify(__heap: list[Any]) -> None: ...
-def heapreplace(__heap: list[_T], __item: _T) -> _T: ...
-def merge(*iterables: Iterable[_T], key: Callable[[_T], Any] | None = ..., reverse: bool = ...) -> Iterable[_T]: ...
-def nlargest(n: int, iterable: Iterable[_T], key: Callable[[_T], SupportsRichComparison] | None = ...) -> list[_T]: ...
-def nsmallest(n: int, iterable: Iterable[_T], key: Callable[[_T], SupportsRichComparison] | None = ...) -> list[_T]: ...
+__about__: str
```
Akuli, line 7: Is this needed? Usually `__foo__` names are treated as public, not private like `_foo` names.

hauntsaninja: I think it is, or at least when I checked earlier it appears dunders are not imported using `from x import *`

Akuli: Lol, cpython developers have decided to copy/paste the same text into both modules

```
>>> from _heapq import __about__ as a
>>> from heapq import __about__ as b
>>> a == b
True
>>> a is b
False
```

Akuli (approved):

erictraut:
Thanks for the clarification. Here's a bug report to track the issue in pyright: https://github.com/microsoft/pyright/issues/3087.

PR #7314, opened 2022-02-20
asyncio: fix for py311

hauntsaninja:
BaseEventLoop was updated, but incompatible with AbstractEventLoop

### DIFF
--- a/stdlib/asyncio/events.pyi
+++ b/stdlib/asyncio/events.pyi
@@ -451,21 +451,37 @@ class AbstractEventLoop:
             ssl: _SSLContext = ...,
         ) -> Server: ...
 
-    @abstractmethod
-    async def create_datagram_endpoint(
-        self,
-        protocol_factory: Callable[[], _ProtocolT],
-        local_addr: tuple[str, int] | None = ...,
-        remote_addr: tuple[str, int] | None = ...,
-        *,
-        family: int = ...,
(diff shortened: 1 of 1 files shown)


stdlib/asyncio/events.pyi
```diff
@@ -451,21 +451,37 @@ class AbstractEventLoop:
...
+        async def create_datagram_endpoint(
+            self,
+            protocol_factory: Callable[[], _ProtocolT],
+            local_addr: tuple[str, int] | None = ...,
+            remote_addr: tuple[str, int] | None = ...,
+            *,
+            family: int = ...,
+            proto: int = ...,
+            flags: int = ...,
+            reuse_port: bool | None = ...,
```
Akuli, line 465: On freshly pulled cpython's `main` branch, I get:

```
akuli@akuli-desktop:~/sourcestuff/cpython$ grep -A4 'def create_datagram_endpoint' Lib/asyncio/events.py 
    async def create_datagram_endpoint(self, protocol_factory,
                                       local_addr=None, remote_addr=None, *,
                                       family=0, proto=0, flags=0,
                                       reuse_address=None, reuse_port=None,
                                       allow_broadcast=None, sock=None):
```

So it seems like `reuse_address` was not removed. Am I missing something?

AlexWaygood:
I think this has been superseded by #7389

PR #7317, opened 2022-02-20
graphlib: update for py311

hauntsaninja:


stdlib/graphlib.pyi
```diff
@@ -1,4 +1,6 @@
+import sys
 from _typeshed import SupportsItems
+from types import GenericAlias
```
JelleZijlstra, line 3: This needs to be version-limited

JelleZijlstra:
It doesn't seem like we have any previous imports in class scopes. I'm not sure we should start doing that, even if CI is green right now.

hauntsaninja (commented):
I thought we loved class scoped imports :'(

PR #7318, opened 2022-02-20
typing: don't accidentally use typing.Self

hauntsaninja:
We can switch to it when all type checkers have support


AlexWaygood:
Lol, this broke flake8-pyi https://github.com/PyCQA/flake8-pyi/pull/186

Issue #7144, opened 2022-02-06
Tensorflow Type Stubs


hmc-cs-mdrissi:
I would be interested in contributing and working on typestubs for [tensorflow](https://www.tensorflow.org/api_docs/python/tf). It does have a fairly massive api though and am unsure whether it would be reasonable to have them start at typeshed and then eventually go back to tensorflow one day when they decide to resolve this [issue](https://github.com/tensorflow/tensorflow/issues/12345).

So first question is would working on them here be acceptable? If yes, what would be recommended way to start them? Start with a partial stub and just work on getting directory structure reasonable with a lot of `__getattr__(name: str) -> Any`? I think other main thing would be common type aliases that will be used for most functions. I already have been working on stubs for tensorflow for work usage and could start by cutting up that into small prs.

Akuli:
> have them start at typeshed and then eventually go back to tensorflow one day when they decide to resolve this issue.

This seems like a good idea to me. I think there are tools that automatically merge type annotations from `.pyi` files into `.py` files.

> So first question is would working on them here be acceptable?

Yes, it is acceptable :)

> If yes, what would be recommended way to start them?

We usually auto-generate the stubs (see CONTRIBUTING.md for instructions). They are initially very incomplete and full of `Any` and the like, but we then add types to the most commonly used parts, or whatever parts of the library the contributor happens to need.

> I already have been working on stubs for tensorflow for work usage and could start by cutting up that into small prs.

Awesome :)

hmc-cs-mdrissi:
Trying stubgen on tensorflow, you have to raise timeout as default timeout is too short given libraries size. With a high timeout stubgen crashes,

```
...
Critical error during semantic analysis: /Users/mdrissi/.virtual_envs/mypy/lib/python3.8/site-packages/tensorflow/python/ops/parallel_for/pfor.py:4657: error: invalid syntax 
```

I think that file is valid python given importing it works fine. So looks like a mypy bug (mypy also crashes on that file). I'll open a bug report for mypy later. 

After tweaking file a bit I was able to get stubs. I'd recommend not using generated stubs from past experience using generated tf stubs (in past pyright ones).  Tensorflow does a lot of import/method patching magic that even with Any used still produces many false positive errors.

[...]

PR #7319, opened 2022-02-20
Minimal tensorflow stub structure

hmc-cs-mdrissi:
# Summary
Minimal pr to start [tensorflow stubs](https://github.com/python/typeshed/issues/7144).

Due to tensorflow's heavy import magic using stubgen to make all files isn't really an option here. Defining the right files to even make stubs for is non-trivial as most classes in tensorflow are defined in private file and then dynamically exported elsewhere (sometimes in ways not describable by stubs). So I expect partial packages/modules to be norm for these stubs for a while.

I included basic stub for most common class Tensor, one example simple function in math.pyi, and commonly needed type aliases.

edit: I end up having two follow prs be added here as separate commits. I can move them to separate prs, I'm unsure if prs are allowed to point to a different base from a separate fork though.


stubs/tensorflow/tensorflow/__init__.pyi
```diff
@@ -0,0 +1,68 @@
+# Alias for bool is used because tensorflow name shadows bool with tf.bool.
+from builtins import bool as _bool
+from typing import Any, Iterable, Iterator, NoReturn, overload
+
+import numpy as np
```
hmc-cs-mdrissi, line 5: How do I handle numpy errors? I thought requires line in METADATA.toml would allow using numpy?

numpy is only library I think needed as most tensorflow functions accept numpy arrays too.

stubs/tensorflow/tensorflow/_aliases.pyi
```diff
@@ -0,0 +1,36 @@
...
+# and do not exist at runtime.
+
+from typing import Iterable, Mapping, Optional, Sequence, TypeVar, Union
+
+import numpy as np
+import tensorflow as tf
+
+# These aliases mostly ignore rank/shape/dtype information as that
+# will complicate the types heavily and can be a follow up problem.
+_FloatDataSequence = Union[Sequence[float], Sequence["_FloatDataSequence"]]
```
hmc-cs-mdrissi, line 11: Several of these type aliases are recursive. What would be recommended way to write them? Arbitrarily nested sequences or json like containers (nested maps/sequences) are very common types needed. Should I leave them like this, use a recursive protocol, avoid recursion and fallback to Any beyond given depth, or something else?

_ContainerGeneric and _DataSequence alises are the main recursive ones here.

hmc-cs-mdrissi:
Marking as draft, as it looks like step 1 is adding support for non-stub dependencies.

hmc-cs-mdrissi:
I will close this for now and reopen it later once I've gotten CI worked out for using numpy.

PR #7323, opened 2022-02-20
xml.dom.minidom: add Node.__bool__

hauntsaninja:


stdlib/xml/dom/minidom.pyi
```diff
@@ -22,6 +22,7 @@ class Node(xml.dom.Node):
     def lastChild(self) -> Node | None: ...
     @property
     def localName(self) -> str | None: ...
+    def __bool__(self) -> bool: ...
```
Akuli, line 25: Should this perhaps return `Literal[True]`, or is the expectation that subclasses can return `False`? The method seems quite pointless to me, as objects are truthy by default, and the base class `xml.dom.Node` doesn't define `__bool__`.

hauntsaninja: I assumed it was a subclass thing, but you're right that I don't see any subclasses in CPython.

JelleZijlstra (approved):
I'm fine with just returning `bool`.

PR #7324, opened 2022-02-20
builtins: add __build_class__

hauntsaninja:
TIL about __build_class__.

I'm a little unsure about the type of `func` here.
It sounds like it should take in `ns`, a dict, that it can modify:
https://github.com/python/cpython/blob/12a2e41e8a276ec3a68f3e5f94b02752185c66fb/Python/compile.c#L2538
https://github.com/python/cpython/blob/main/Python/bltinmodule.c#L201
https://eli.thegreenplace.net/2012/06/15/under-the-hood-of-python-class-definitions

But empirically doesn't seem to be the case? It doesn't seem to do anything :-/ I'm probably missing something dumb, maybe things will make sense in the morning.
```
In [1]: def f(*a, **kw):
   ...:     print(locals())
   ...:     set_something = 10
   ...:     return {"return_something": 20}
   ...: 

In [2]: __build_class__(f, "myclass")
{'a': (), 'kw': {}}
Out[2]: __main__.myclass
```


TeamSpen210:
The `ns` parameter is actually the function's locals namespace, not a parameter - it always takes nothing. The return value is used for implementing `super()` - if that's used, the namespace gets given `__classcell__ =` a cell object containing the class, which gets fed into the closures of the methods as `__class__`. It's then returned out of the class-function. [Here's the relevant compiler code](https://github.com/python/cpython/blob/3be1a443ca8e7d4ba85f95b78df5c4122cae9ede/Python/compile.c#L2593).

So the right annotation would be `Callable[[], types.CellType | None]` - except it's not publicly defined there before 3.8, so I guess just use the typeshed-private definition [`types._Cell`](https://github.com/python/typeshed/blob/6b29dca1284619cbc2a2ca79555c4a67eff08c1b/stdlib/types.pyi#L36)?

hauntsaninja:
Okay, good to know, but I still don't quite understand what's going on. I am modifying the locals namespace in my `f`, but `set_something` does not appear on the class.

hauntsaninja:
Okay, if I steal the code object from an existing class and make my own function it works:
```
In [1]: import types

In [2]: def f():
   ...:     class C:
   ...:         set_something = "value"
   ...: 

In [3]: __build_class__(types.FunctionType(f.__code__.co_consts[1], globals()), "myclass").__dict__
Out[3]: 
mappingproxy({'__module__': '__main__',
              'set_something': 'value',
              '__dict__': <attribute '__dict__' of 'myclass' objects>,
              '__weakref__': <attribute '__weakref__' of 'myclass' objects>,
              '__doc__': None})
```

hauntsaninja:
Looking at disassembly, maybe this is some difference arising from the use of `STORE_NAME` vs `STORE_FAST`?

hauntsaninja:
Okay, here's also why comments in code and some of the stuff online is misleading. In early Python 3, this was a function that took a dict as an argument: https://github.com/python/cpython/commit/e8e14591ebb729b4fa19626ce245fa0811cf6f32

The comment in compile.c should probably be updated, I'll make a CPython PR for that.

TeamSpen210:
It is yes, class "functions" are compiled to not use the fast locals array, and instead store to the locals dict every time, as well as fully dynamically lookup names from locals and then globals. STORE_FAST always stores to the fast locals array, which isn't then propagated back to the specified namespace. But doing `locals()["blah"] = 4` should work.

hauntsaninja:
I tried `locals()["blah"] = 4` earlier and it didn't work. Looks like it maps to `STORE_SUBSCR` and maybe that doesn't work either.

stdlib/builtins.pyi
```diff
@@ -1547,6 +1547,7 @@ def __import__(
     fromlist: Sequence[str] = ...,
     level: int = ...,
 ) -> types.ModuleType: ...
+def __build_class__(__func: Callable[[], Any], __name: str, *bases: type, metaclass: type = ..., **kwds: Any) -> type: ...
```
JelleZijlstra, line 1550:
```suggestion
def __build_class__(__func: Callable[[], Any], __name: str, *bases: Any, metaclass: Any = ..., **kwds: Any) -> Any: ...
```
While conventionally these are all types, there is no strict requirement that they are. For example, the bases could just be objects providing `__mro_entries__`. We could be more precise here, but any code using `__build_class__` directly probably isn't going to make type checkers very happy anyway.

JelleZijlstra:
No strong opinion on the return type of the function that's passed as an argument. As @TeamSpen210 said, in practice the compiler generates function that return either None or a cell object, but in fact the implementation just ignores anything that's not a cell object. (I am looking at CPython main in case it matters.)

```
>>> c=__build_class__(lambda: "whatever", "hello", int)
>>> c
<class '__main__.hello'>
>>> type(c)
<class 'type'>
>>> inst = c()
>>> type(inst)
<class '__main__.hello'>
```

JelleZijlstra (approved):

PR #7326, opened 2022-02-20
Improve `importlib.metadata`

AlexWaygood:
- Add `__all__`
- Remove `sys.version_info` guard (should be unneeded now that stubtest recognises submodules in `VERSIONS` 🤞)


AlexWaygood:
The diff isn't particularly helpful here; it's easier just to compare the before-and-after manually...

JelleZijlstra (approved):
GitHub actually did a pretty good job showing the diff

PR #7328, opened 2022-02-20
Add `__all__` to modules beginning with 'j', 'k' and 'l'

AlexWaygood:


stdlib/lib2to3/pgen2/tokenize.pyi
```diff
@@ -1,6 +1,40 @@
-from lib2to3.pgen2.token import *  # noqa
+import sys
+from lib2to3.pgen2.token import *
 from typing import Callable, Iterable, Iterator
 
+# fmt: off
```
AlexWaygood, line 5: Desperate times call for desperate measures.

JelleZijlstra: Just let Black format it.

hauntsaninja (approved):

PR #6826, opened 2022-01-05
Use `Literal` types in `argparse`

sobolevn:

PR #7329, opened 2022-02-20
argparse: Use Literal for action & nargs add_argument parameter types

not-my-profile:
As has previously been discussed in https://github.com/python/typeshed/pull/6826#issuecomment-1005654071 the `argparse` module sometimes compares `argparse.SUPPRESS` with `is` meaning `Literal["==SUPPRESS=="]` is not suited. This PR solves this using `NewType`.

As reported by [mypy_primer](https://github.com/hauntsaninja/mypy_primer) the following fails with mypy:

```python
import argparse
x = argparse.SUPPRESS
x = "test"
```

because of:

```
foo.py:4: error: Incompatible types in assignment (expression has type "str", variable has type "_SUPPRESS_T")
```

This however seems to be due to a limitation in mypy not understanding that a newtype is backwards compatible with the original type as can be observed in the following example:

```python
import typing

X = typing.NewType('X', int)

def foo(a: X) -> int:
    return a
```

Mypy complains whereas Pyright says it's ok. With that in mind I think that the advantages of this PR (type safety against typos) outweigh the caused inconvenience (when reassigning SUPPRESS with mypy).

### DIFF
--- a/stdlib/argparse.pyi
+++ b/stdlib/argparse.pyi
@@ -1,5 +1,20 @@
 import sys
-from typing import IO, Any, Callable, Generator, Generic, Iterable, NoReturn, Pattern, Protocol, Sequence, TypeVar, overload
+from typing import (
+    IO,
+    Any,
+    Callable,
+    Generator,
+    Generic,
+    Iterable,
+    NewType,
+    NoReturn,
+    Pattern,
+    Protocol,
+    Sequence,
+    TypeVar,
+    overload,
+)
+from typing_extensions import Literal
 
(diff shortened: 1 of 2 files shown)


JelleZijlstra (approved):
Looks good. The cwltool code found by mypy-primer can just add explicit annotations.

Akuli:
We could try `SUPPRESS: _SUPPRESS_T | str` to avoid the primer error. It's not really correct because `SUPPRESS` isn't just a string in the usual sense, but after `foo = argparse.SUPPRESS` you would at least be able to reassign a string.

PR #7330, opened 2022-02-20
Add `__all__` to modules beginning with 'm'

AlexWaygood:
Also:

- Improve `multiprocessing.connection._ConnectionBase.__init__`. (`SupportsIndex` can be used unconditionally if we import it from `typing_extensions` instead of `typing`).
- Add `multiprocessing.shared_memory` to `VERSIONS`; remove the module-level `sys.version_info` guard and the allowlist entries.


stdlib/multiprocessing/__init__.pyi
```diff
@@ -29,6 +29,33 @@ if sys.version_info >= (3, 8):
 if sys.platform != "win32":
     from multiprocessing.context import ForkContext, ForkServerContext
 
+# fmt: off
```
JelleZijlstra, line 32: I'd rather we just let Black format this.

AlexWaygood: Okay!

JelleZijlstra: Thanks! (pet peeve: it's "free rein", as in letting go of your horse's reins)

AlexWaygood: Lol, of course. Too much writing of code recently, not enough writing of words...

PR #6557, opened 2021-12-09
Update `importlib` to reflect recent changes

AlexWaygood:
The `importlib` API has evolved quite rapidly over the past 3 versions of Python.

The big change here is creating the subdirectory `importlib.metadata` (added to CPython in Python 3.10), and moving the code that was in `importlib.metadata.pyi` into `importlib.metadata.__init__.pyi`

PR #7331, opened 2022-02-20
Update `importlib.metadata` typings for 3.10

### LINKED ISSUE #7262: Stub for `importlib.metadata` on Python 3.10 is incomplete
- [ ] [`entry_points()`](https://docs.python.org/3/library/importlib.metadata.html#entry-points) was [changed](https://docs.python.org/3/library/importlib.metadata.html#:~:text=definition%2C%20and%20usage.-,Compatibility%20Note,-The%20%E2%80%9Cselectable%E2%80%9D%20entry) in 3.10
- [ ] [`metadata()`](https://docs.python.org/3/library/importlib.metadata.html#distribution-metadata) function returns `PackageMetadata` which got new attribute [`json`](https://docs.python.org/3/library/importlib.metadata.html#:~:text=have%20been%20removed.-,New%20in%20version%203.10,-%3A%20The%20json%20attribute)

itaisteinherz:
Fixes #7262 
Refs #6557 

// cc @Molkree @hauntsaninja


itaisteinherz:
At first I thought that I'll need to add `EntryPoints` as well, but after another look it seems like it isn't used by any public-facing method.

stdlib/importlib/metadata/__init__.pyi
```diff
@@ -35,6 +35,16 @@ if sys.version_info >= (3, 8):
             dist: ClassVar[Distribution | None]
             def matches(self, **params: Any) -> bool: ...  # undocumented
 
+    if sys.version_info >= (3, 10):
+        class SelectableGroups(dict):  # use as list is deprecated since 3.10
+            @classmethod
+            def load(cls, eps: SelectableGroups) -> SelectableGroups: ...
```
itaisteinherz, line 41: I'm not sure if `eps` should be `SelectableGroups` / `EntryPoints` / both.

itaisteinherz:
> At first I thought that I'll need to add `EntryPoints` as well, but after another look it seems like it isn't used by any public method.

It seems like `EntryPoints` will be required for the typing of `SelectableGroups`.

stdlib/importlib/metadata/__init__.pyi
```diff
@@ -10,7 +10,7 @@ from pathlib import Path
 from typing import Any, ClassVar, Iterable, NamedTuple, Pattern, overload
 
 if sys.version_info >= (3, 10):
-    from importlib.metadata._meta import PackageMetadata as PackageMetadata
+    from importlib.metadata._meta import PackageMetadata
```
AlexWaygood, line 13: Why make this change? It's present in the module's `__all__` (https://github.com/python/cpython/blob/be095f6c32188bba02079d086ac8639ea37cec3c/Lib/importlib/metadata/__init__.py#L33)

itaisteinherz: You're right, I forgot that this is a stubs convention 😅 Will fix.

AlexWaygood: (Imports in a stub are considered private to the stub unless you use the weird `import foo as foo` idiom — it's one of the ways stub files differ from `.py` files 🙂)

itaisteinherz: Fixed.

stdlib/importlib/metadata/__init__.pyi
itaisteinherz: Maybe `SelectableGroups` should extend `EntryPoints`?

itaisteinherz: > Maybe `SelectableGroups` should extend `EntryPoints`?

Never mind, that doesn't make sense given the current implementation in CPython.

itaisteinherz:
Not sure what caused this diff 🤔

stdlib/importlib/metadata/__init__.pyi
```diff
@@ -109,8 +134,14 @@ if sys.version_info >= (3, 8):
...
         *, context: None = ..., name: str | None = ..., path: list[str] = ..., **kwargs: Any
     ) -> Iterable[Distribution]: ...
-    def metadata(distribution_name: str) -> Message: ...
+
+    if sys.version_info >= (3, 10):
+        def metadata(distribution_name: str) -> PackageMetadata: ...
+        def entry_points() -> SelectableGroups: ...
+    else:
+        def metadata(distribution_name: str) -> Message: ...
+        def entry_points() -> dict[str, list[EntryPoint]]: ...
```
Molkree, line 143: I think this is incomplete, consider the example:
```py
(mypy,) = entry_points(group="console_scripts", name="mypy")
print(mypy)
```
This is valid and works:
```
EntryPoint(name='mypy', value='mypy.__main__:console_entry', group='console_scripts')
```

image

Molkree (changes requested):

stdlib/importlib/metadata/__init__.pyi
itaisteinherz: This is why I chose to use `list`:

https://github.com/python/cpython/blob/b77158b4da449ec5b8f682816a79d004fd65ed07/Lib/importlib/metadata/__init__.py#L325

Using `tuple` instead doesn't make much sense to me, as it would contradict the CPython implementation.

Molkree: My point was about 3.10 if branch, sorry for commenting on the wrong line.

https://github.com/python/typeshed/blob/a10cced8ea144af367f70349c65f475cd81707aa/stdlib/importlib/metadata/__init__.pyi#L140

This function should have `name` and `groups` arguments.

stdlib/importlib/metadata/__init__.pyi
```diff
@@ -35,6 +35,23 @@ if sys.version_info >= (3, 8):
...
+            def groups(self) -> set[str]: ...
+
+        class SelectableGroups(dict[str, EntryPoints]):  # use as list is deprecated since 3.10
+            @classmethod
+            def load(cls, eps: SelectableGroups) -> SelectableGroups: ...
+            @property
+            def groups(self) -> set[str]: ...
+            @property
+            def names(self) -> set[str]: ...
+            def select(self, **params: Any) -> SelectableGroups: ...
```
Molkree, line 53: I'm curious, shouldn't it be only `str` and not `Any`? I think it can only be `name`, `group`, `value`, `module`, `attr` and `extras` here, all are just strings as far as I understand?

Molkree, line 53: This should return `EntryPoints` instead.

```suggestion
            def select(self, **params: Any) -> EntryPoints: ...
```

Molkree (changes requested):

stdlib/importlib/metadata/__init__.pyi
```diff
@@ -35,6 +35,23 @@ if sys.version_info >= (3, 8):
...
+        class EntryPoints(list[EntryPoint]):  # use as list is deprecated since 3.10
+            def select(self, **params: Any) -> EntryPoints: ...
+            @property
+            def names(self) -> set[str]: ...
+            @property
+            def groups(self) -> set[str]: ...
+
+        class SelectableGroups(dict[str, EntryPoints]):  # use as list is deprecated since 3.10
+            @classmethod
+            def load(cls, eps: SelectableGroups) -> SelectableGroups: ...
```
JelleZijlstra, line 48: I think this is right:
```suggestion
            def load(cls: type[Self], eps: Iterable[EntryPoint]) -> Self: ...
```
Where `Self` is from `_typeshed`.

- It returns `cls()`, so it should return `Self` (this makes a difference for a subclass of `SelectableGroups`)
- The arguments get passed to the `EntryPoints` constructor, which wants an iterable of `EntryPoint` instances.

stdlib/importlib/metadata/__init__.pyi
JelleZijlstra: It looks like `extras` is a list of strings instead, so `str` doesn't seem right. However, we could just specify all of the allowed attributes manually to provide more precise types.

JelleZijlstra: Actually it can be either: it returns self if there are no params, but `EntryPoints` if there are any.

stdlib/importlib/metadata/__init__.pyi
```diff
@@ -109,8 +134,14 @@ if sys.version_info >= (3, 8):
     def distributions(
         *, context: None = ..., name: str | None = ..., path: list[str] = ..., **kwargs: Any
     ) -> Iterable[Distribution]: ...
-    def metadata(distribution_name: str) -> Message: ...
+
+    if sys.version_info >= (3, 10):
+        def metadata(distribution_name: str) -> PackageMetadata: ...
+        def entry_points() -> SelectableGroups: ...
```
JelleZijlstra, line 140: Like `SelectableGroups.entry_points`, this takes `**params` and can return either an `EntryPoints` (if there are any params) or `SelectableGroups` (if there aren't).

JelleZijlstra (changes requested):

stdlib/importlib/metadata/__init__.pyi
hauntsaninja: Technically type checkers use the `X as X` idiom when determining public symbols in .py code too: it's more a weird typed Python convention than a stubs convention

AlexWaygood: Huh, TIL.

stdlib/importlib/metadata/__init__.pyi
Molkree: Didn't realize empty `select()` is possible 😅

JelleZijlstra: Not how I would design this API :)

stdlib/importlib/metadata/__init__.pyi
```diff
@@ -35,6 +35,23 @@ if sys.version_info >= (3, 8):
...
 
+    if sys.version_info >= (3, 10):
+        class EntryPoints(list[EntryPoint]):  # use as list is deprecated since 3.10
+            def select(self, **params: Any) -> EntryPoints: ...
+            @property
+            def names(self) -> set[str]: ...
+            @property
+            def groups(self) -> set[str]: ...
+
+        class SelectableGroups(dict[str, EntryPoints]):  # use as list is deprecated since 3.10
```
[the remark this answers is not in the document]
itaisteinherz: Nice catch :)

stdlib/importlib/metadata/__init__.pyi
itaisteinherz: Note that this isn't done in the rest of the stub file.

stdlib/importlib/metadata/__init__.pyi
itaisteinherz: Added an appropriate `@overload`.

stdlib/importlib/metadata/__init__.pyi
itaisteinherz: Done.

stdlib/importlib/metadata/__init__.pyi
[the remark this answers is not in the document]
itaisteinherz: Done.

stdlib/importlib/metadata/__init__.pyi
itaisteinherz: Updated, let me know what you think.

stdlib/importlib/metadata/__init__.pyi
itaisteinherz: Updated, see the discussion below.

stdlib/importlib/metadata/__init__.pyi
```diff
@@ -35,6 +35,23 @@ if sys.version_info >= (3, 8):
             dist: ClassVar[Distribution | None]
             def matches(self, **params: Any) -> bool: ...  # undocumented
 
+    if sys.version_info >= (3, 10):
+        class EntryPoints(list[EntryPoint]):  # use as list is deprecated since 3.10
```
srittau, line 39: We also need to override `__getitem__`. `list.__getitem__` accepts `SupportsIndex`, while this accepts `int` (deprecated) and `str`.

stdlib/importlib/metadata/__init__.pyi
srittau: I believe this needs an overload:

```python
@overload
def select(self: Self) -> Self: ...
@overload
def select(self, **params: Any) -> EntryPoints: ...
```

This might need a `# type: ignore` on the second overload.

srittau (changes requested):

stdlib/importlib/metadata/__init__.pyi
Molkree: Now I get this:
```
stdlib\importlib\metadata\__init__.pyi:141: error: Overloaded function signatures 1 and 2 overlap with incompatible return types
```

stdlib/importlib/metadata/__init__.pyi
Molkree: > This might need a `# type: ignore` on the second overload.

To silence this `mypy` error?
```
stdlib\importlib\metadata\__init__.pyi:54: error: Overloaded function signatures 1 and 2 overlap with incompatible return types
```

`mypy` reports the error on the first overload so comment should be on the first one I think.

srittau: You are right, first overload it is.

itaisteinherz: By the way, why does mypy report this as an error? Dynamic return types aren't that uncommon.

stdlib/importlib/metadata/__init__.pyi
itaisteinherz: Wasn't sure if I should include `__getitem__` in the stubs at first :) Will add it.

itaisteinherz: Added.

stdlib/importlib/metadata/__init__.pyi
itaisteinherz: Added the overload.

stdlib/importlib/metadata/__init__.pyi
itaisteinherz: Error is now ignored.

stdlib/importlib/metadata/__init__.pyi
srittau: Good question. This needs to be commonly ignored in typeshed. I'd be happy if mypy didn't warn about this.

Akuli: The idea is probably that the return type of `select(**some_dict)` depends on whether the dict is empty or not. So the best that mypy can do is to return a union, which is annoying to deal with in practice.

For non-typeshed code, I think getting a warning about this is a good thing. because methods like this would be less confusing if written as two separate methods, and getting a warning will make you rethink about whether it is a good idea.

stdlib/importlib/metadata/__init__.pyi
```diff
@@ -61,7 +61,53 @@ class EntryPoint(_EntryPointBase):
...
+        def select(self: Self) -> Self: ...  # type: ignore[misc]
+        @overload
+        def select(
+            self,
+            __name: str = ...,
+            __value: str = ...,
+            __group: str = ...,
+            __module: str = ...,
+            __attr: str = ...,
+            __extras: list[str] = ...,
```
Molkree, line 109: I don't think these should start with dunder, they are being hidden this way.

Check this example:
```py
from importlib import metadata as importlib_metadata

(mypy,) = importlib_metadata.entry_points().select(name="mypy")
print(mypy)
```

Running `mypy` on it results in:
```
example.py:3: error: No overload variant of "select" of "SelectableGroups" matches argument type "str"
example.py:3: note: Possible overload variants:
example.py:3: note:     def select(self) -> SelectableGroups
example.py:3: note:     def select(self, str = ..., str = ..., str = ..., str = ..., str = ..., List[str] = ...) -> EntryPoints
```

pylance/pyright doesn't suggest arguments either:
image

But if I remove dunders for all arguments (I think this should apply to all functions/methods in this file) everything seems to work fine:
image

Molkree (changes requested):

stdlib/importlib/metadata/__init__.pyi
itaisteinherz: The reason I prefixed these arguments with a dunder is because from what I could recall, according to the conventions, that's how you annotate that an argument is keyword-only. However, I now see that this is actually the convention for positional-only arguments 😅

itaisteinherz: Fixed.

stdlib/importlib/metadata/__init__.pyi
```diff
@@ -61,7 +61,56 @@ class EntryPoint(_EntryPointBase):
         def attr(self) -> str: ...
     if sys.version_info >= (3, 10):
         dist: ClassVar[Distribution | None]
-        def matches(self, **params: Any) -> bool: ...  # undocumented
+        def matches(
```
itaisteinherz, line 64: By the way, I don't really like the fact that I manually copy-and-pasted the typings of all methods which are based on `matches()`. I thought about maybe using a `TypedDict` for defining a general interface for `params`. Is there some convention for how to do this?

stdlib/importlib/metadata/__init__.pyi
itaisteinherz: Added.

stdlib/importlib/metadata/__init__.pyi
AlexWaygood: > Is there some convention for what to do in such cases?

No, the way you've done it is the best way possible at the moment -- but there are discussions on the typing-sig mailing list right now about a possible PEP proposing being able to use a `TypedDict` for kwargs in a function signature :)

itaisteinherz: Cool :) I also feel like a solution in the fashion of `ParamSpec` could be interesting (that is, some way to type-annotate that a set of keyword arguments are literally forwarded as-is to some other method).

stdlib/importlib/metadata/__init__.pyi
```diff
@@ -61,7 +61,56 @@ class EntryPoint(_EntryPointBase):
...
+            self,
+            *,
+            name: str = ...,
+            value: str = ...,
+            group: str = ...,
+            module: str = ...,
+            attr: str = ...,
+            extras: list[str] = ...,
```
srittau, line 72: Does it make sense to include `extras`? The property always returns a newly constructed list, so it can never match as `matches()` only does a shallow `==`. (The same is true for other methods below.)

srittau (commented):
Just one more thing I noticed, looks good to me otherwise.

stdlib/importlib/metadata/__init__.pyi
itaisteinherz: Why? Isn't `[1, 2, 3] == [1, 2, 3]` (for example) true in Python?

srittau: You are right, I was in JavaScript-mode.

srittau (approved):

JelleZijlstra (approved):

itaisteinherz:
@AlexWaygood should I also have added `SelectableGroups` and `EntryPoints` to `__all__`? I just noticed your PR #7326, which initially added `__all__`, and noticed I forgot to add them.
(I also noticed that you didn't include `EntryPoint` in `__all__`?)

AlexWaygood:
> 



> @AlexWaygood should I have also added `SelectableGroups` and `EntryPoints` to `__all__`? I just noticed your PR #7326, which initially added `__all__`, and noticed I forgot to add them. (I also noticed that you didn't include `EntryPoint` in `__all__`?)

@itaisteinherz, none of them are included in `__all__` at runtime, so I don't see any reason to include them in `__all__` in the stub https://github.com/python/cpython/blob/2c228a7b8f89e9ed8d390370abd771d4993b79d8/Lib/importlib/metadata/__init__.py#L30

PR #7333, opened 2022-02-21
Update stubtest for async and dunder pos only checking

hauntsaninja:


stdlib/types.pyi
```diff
@@ -236,15 +236,15 @@ class AsyncGeneratorType(AsyncGenerator[_T_co, _T_contra]):
     __name__: str
     __qualname__: str
     def __aiter__(self) -> AsyncGeneratorType[_T_co, _T_contra]: ...
-    async def __anext__(self) -> _T_co: ...
-    async def asend(self, __val: _T_contra) -> _T_co: ...
+    def __anext__(self) -> Coroutine[Any, Any, _T_co]: ...
```
JelleZijlstra, line 239: Out of curiosity, what is the reason for these changes?

hauntsaninja: Well,
```
>>> inspect.iscoroutinefunction(types.AsyncGeneratorType.aclose)
False
```
I'm not sure if this matters at all, cc @AlexWaygood who added this to stubtest.

AlexWaygood: cc. @AlexWaygood who made them async only a few weeks ago as well...

AlexWaygood: "Classes" in the `types` module are weird, I'm happy to go with whatever stubtest thinks best personally.

JelleZijlstra:
There are also a ton of new stubtest failures on 3.6 and 3.7, probably because they were using `@asyncio.coroutine` instead of `async def` at the time. I think we should just ignore those, similar to how we ignore positional-only discrepancies in old versions.

stdlib/types.pyi
hauntsaninja: (the other big change we've made recently is switching a lot of `(...) -> Awaitable` into `async (...) -> ` which actually does make a difference to type checking)

AlexWaygood:
> There are also a ton of new stubtest failures on 3.6 and 3.7, probably because they were using `@asyncio.coroutine` instead of `async def` at the time. I think we should just ignore those, similar to how we ignore positional-only discrepancies in old versions.

The 3.7 ones look like true positives to me. But yeah, it would maybe be good to have a command-line option to switch the `async` check off, so that we don't have a bajillion false-positives when checking 3.6.

tests/stubtest_allowlists/py36.txt
```diff
@@ -103,6 +103,103 @@ collections.Generator.gi_yieldfrom
 collections.Mapping.get  # Adding None to the Union messed up mypy
 collections.Sequence.index  # Supporting None in end is not mandatory
 
+# Not "async defs" at runtime
```
hauntsaninja, line 106:
```suggestion
# Not "async def"s at runtime
```

tests/stubtest_allowlists/py37.txt
```diff
@@ -3,6 +3,8 @@ _collections_abc.AsyncGenerator.ag_code
 _collections_abc.AsyncGenerator.ag_frame
 _collections_abc.AsyncGenerator.ag_running
 _dummy_threading
+asyncio.AbstractEventLoop.run_in_executor
+asyncio.events.AbstractEventLoop.run_in_executor
```
AlexWaygood, line 7: This is a true positive -- it's `async` in the 3.7 branch of CPython; it was change to be a synchronous method in 3.8. https://github.com/python/cpython/blob/811f65ba263140b6ba28151246b52efe149a6382/Lib/asyncio/events.py#L289

hauntsaninja: I think we should still allowlist it. If you've implemented the AbstractEventLoop interface, you probably don't want some type checker yelling at you for returning a Future (if you check wit 3.7).

AlexWaygood: Okay, fair enough!

AlexWaygood: Maybe we could add a comment to that effect?

AlexWaygood (approved):

hauntsaninja:
Hopefully just in time for the daily test to tell us if this causes issues with third party stubs :-)

AlexWaygood:
@hauntsaninja, would you be up for trying this again now that https://github.com/python/typeshed/commit/64af11c78f2a07c18b8e3b7e6055dc97ee91955e is fixed? The `aiofiles` hits would have to be allowlisted for now until https://github.com/python/mypy/pull/12234 is in, but the other hits should  be resolved now.

stdlib/types.pyi
AlexWaygood: Interestingly, it looks like there are already at least 2 open BPO issues about the fact that `inspect.iscoroutinefunction` returns `False` for `types.AsyncGeneratorType.__anext__`, `.asend` and `.aclose`. https://bugs.python.org/issue42760.

I'm not really sure what we should make of that, but it's interesting!

PR #7335, opened 2022-02-21
flake8: Turn off Y034 in typing.pyi

JelleZijlstra:
See PyCQA/flake8-pyi#186. flake8-pyi doesn't like us renaming `_typeshed.Self` to `_TypeshedSelf`.

cc @AlexWaygood


JelleZijlstra:
#7336 is better

PR #7337, opened 2022-02-21
select: mark positional-only args

hauntsaninja:


JelleZijlstra (approved):
Seems like it actually just takes `*args` (on 3.6 at least), but I don't think it matters.

PR #7340, opened 2022-02-21
Fix unittest.TextTestResult showAll name

adamchainz:
`showall` -> `showAll` - it has always been camel-cased and didn't change - [python 2.7 source](https://github.com/python/cpython/blob/2.7/Lib/unittest/runner.py) , [main](https://github.com/python/cpython/blob/main/Lib/unittest/runner.py)

### DIFF
--- a/stdlib/unittest/runner.pyi
+++ b/stdlib/unittest/runner.pyi
@@ -11,7 +11,6 @@ class TextTestResult(unittest.result.TestResult):
     separator1: str
     separator2: str
     showAll: bool  # undocumented
-    dots: bool  # undocumented
     stream: TextIO  # undocumented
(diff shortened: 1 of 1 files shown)


stdlib/unittest/runner.pyi
```diff
@@ -10,7 +10,8 @@ class TextTestResult(unittest.result.TestResult):
     dots: bool  # undocumented
     separator1: str
     separator2: str
-    showall: bool  # undocumented
+    showAll: bool  # undocumented
+    dots: bool  # undocumented
```
Akuli, line 14: `dots` is already defined above:

```suggestion
```

adamchainz: My bad!

Akuli (approved):

Akuli:
Thanks!

PR #7341, opened 2022-02-21
Fix unittest.TextTestResult.printErrorList()

adamchainz:
This function is passed either `self.errors` or `self.failures` which are defined as `list`s of such tuples: https://github.com/python/typeshed/blob/4d169eaa83bfde68e0abc55670b73c1f0f408043/stdlib/unittest/result.pyi#L16-L17

This can be seen from [Python 2.7 source](https://github.com/python/cpython/blob/2.7/Lib/unittest/runner.py#L105) and up.

(Sorry for not merging in #7340 - I was still untangling some type check errors 😅)

### DIFF
--- a/stdlib/unittest/runner.pyi
+++ b/stdlib/unittest/runner.pyi
@@ -1,7 +1,7 @@
 import unittest.case
 import unittest.result
 import unittest.suite
-from typing import Callable, TextIO
+from typing import Callable, Iterable, TextIO
 
 _ResultClassType = Callable[[TextIO, bool, int], unittest.result.TestResult]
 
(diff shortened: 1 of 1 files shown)


stdlib/unittest/runner.pyi
```diff
@@ -15,7 +15,7 @@ class TextTestResult(unittest.result.TestResult):
     def __init__(self, stream: TextIO, descriptions: bool, verbosity: int) -> None: ...
     def getDescription(self, test: unittest.case.TestCase) -> str: ...
     def printErrors(self) -> None: ...
-    def printErrorList(self, flavour: str, errors: tuple[unittest.case.TestCase, str]) -> None: ...
+    def printErrorList(self, flavour: str, errors: list[tuple[unittest.case.TestCase, str]]) -> None: ...
```
Akuli, line 18:
```suggestion
    def printErrorList(self, flavour: str, errors: Iterable[tuple[unittest.case.TestCase, str]]) -> None: ...
```

Akuli (approved):
Nit for future reference: It's better to not squash and force-push, as it makes it harder for us to see what has changed since the previous review. When merging the PR, we will use GitHub's "Squash and merge" button, which combines the whole PR into a single commit anyway.

adamchainz:
Okay fair enough. I am used to force-pushing since we do it on Django all the time.

You *can* see the diff but it's a kinda hidden feature - click the word "force-pushed" in the PR log message:

> [](https://github.com/adamchainz)@[adamchainz](https://github.com/adamchainz) adamchainz [force-pushed](https://github.com/python/typeshed/compare/43de3937be9d7c41afbe29ed51b750e7c97665d8..3a1e1a7134d9c00164672c3dc2422e2eb9742839) the patch-1 branch from 43de393 to 3a1e1a7 34 minutes ago

srittau:
@adamchainz That's a useful hint. I didn't know you can still see the diff.

PR #7345, opened 2022-02-22
Add `__all__` to most modules beginning with 'n', 'o' and 'p'

AlexWaygood:
I'm leaving `os/__init__.pyi` and `pickle.pyi` to do another time (if at all).


stdlib/plistlib.pyi
```diff
@@ -4,6 +4,57 @@ from datetime import datetime
...
+        "InvalidFileException",
+        "FMT_XML",
+        "FMT_BINARY",
+        "load",
+        "dump",
+        "loads",
+        "dumps",
+        "UID",
+    ]
+elif sys.version_Info >= (3, 7):
```
JelleZijlstra, line 25:
```suggestion
elif sys.version_info >= (3, 7):
```

AlexWaygood: Thanks!

PR #7347, opened 2022-02-22
uuid: mark setattr

hauntsaninja:


AlexWaygood:
What does this do for us? mypy will already raise an error if a user tries to set an attribute on an object that hasn't been formally declared as an attribute of that object. And we already have all instance attributes on this class as read-only properties.

hauntsaninja:
It doesn't do much, beyond mypy understanding what happens if you do `UUID.__setattr__("attr", 5)`.
I was looking at this because it's one of the new errors from https://github.com/python/mypy/pull/12203.

hauntsaninja:
Actually no, this makes things worse, since mypy will give you worse (and depending on settings, sometimes no) diagnostics if you do `u.doesnt_exist = 3`

PR #6782, opened 2022-01-02
Simplify `bool` stub

AlexWaygood:
The semantics of this proposed PR should be identical to the existing stub. I'm curious, though, to see whether primer shows false-positives like those we saw in #6762

PR #6762, opened 2021-12-30
Remove unnecessary overrides from re

jpy-git:
This PR follows on from the discussion [here](https://github.com/python/typeshed/pull/6713#discussion_r776677437) where @AlexWaygood highlighted that overloads are unnecessary when a function has the same return type and therefore simple union-types should be used instead. This PR combines these unnecessary overloads in `re.pyi` and for `Pattern` in `typing.pyi`.

Additionally I spotted that `Pattern.split` should have a return type of `list[AnyStr | Any]` rather than `list[AnyStr]`:
image
(`re.split` is already correct)

PR #7349, opened 2022-02-22
re: Merge overloaded functions by using unions

not-my-profile:
The signatures of these functions can also be modeled
without overloads using unions as parameter types.

While this shouldn't make any difference for type checking,
it improves the UX for people using IDEs because having to
cycle between subtly different function signatures isn't fun.


AlexWaygood:
See also:

- https://github.com/python/typeshed/pull/6762
- https://github.com/python/typing/discussions/1080
- https://github.com/python/typeshed/pull/6782

not-my-profile:
Ah, well ... how unfortunate that mypy has such a limitation. I guess the tracking issue for that is https://github.com/python/mypy/issues/11880?

AlexWaygood:
> Ah, well ... how unfortunate that mypy has such a limitation. I guess the tracking issue for that is [python/mypy#11880](https://github.com/python/mypy/issues/11880)?

Yup 🙂 it might be good to add a comment to the `re` module functions and `bool`, to explain why they're typed the way they currently are — it's obviously pretty unintuitive for contributors.

PR #7356, opened 2022-02-22
Increase pyright strictness when checking `__all__`

AlexWaygood:
And also fix a typo in `pyrightconfig.stricter.json` that was already fixed in `pyrightconfig.json` in https://github.com/python/typeshed/pull/7355

PR #6398, opened 2021-11-27
Clarify why `csv` stub has no `__version__` attribute

AlexWaygood:

PR #7357, opened 2022-02-22
Add `csv.__version__`

AlexWaygood:
We previously discussed this in #6398, and decided not to include it in the stub. However, it is declared as a publicly exported object in the module's `__all__` at runtime, so not including it in the stub creates an inconsistency in the stub's `__all__` (see https://github.com/python/typeshed/pull/7356). Moreover, there doesn't really seem to be any harm in adding it to the stub. So, I think we should add it.

### DIFF
--- a/stdlib/_csv.pyi
+++ b/stdlib/_csv.pyi
@@ -1,6 +1,8 @@
 from typing import Any, Iterable, Iterator, Protocol, Union
 from typing_extensions import Literal
 
+__version__: str
+
 QUOTE_ALL: Literal[1]
 QUOTE_MINIMAL: Literal[0]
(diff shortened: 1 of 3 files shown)


JelleZijlstra (approved):

Akuli:
I still would prefer if `csv.VERSION` didn't show up in autocompletions.

AlexWaygood:
> I still would prefer if `csv.VERSION` didn't show up in autocompletions.

I get that, but I feel like it's a CPython bug rather than a typeshed bug if the `__version__` attribute is really out of date.

Akuli:
Given that it starts with `_`, it will likely go to the end of a long autocompletion list anyway.

PR #7358, opened 2022-02-22
Add `enum` module features new in 3.11

AlexWaygood:


stdlib/enum.pyi
```diff
@@ -221,12 +227,24 @@ class IntFlag(int, Flag):
...
+        CONTINUOUS: str
+        NAMED_FLAGS: str
+        UNIQUE: str
+    CONTINUOUS = EnumCheck.CONTINUOUS
+    NAMED_FLAGS = EnumCheck.NAMED_FLAGS
+    UNIQUE = EnumCheck.UNIQUE
+
+    class verify:
+        def __init__(self, *checks: EnumCheck) -> None: ...
+        def __call__(self, enumeration: _S) -> _S: ...
```
JelleZijlstra, line 246: This TypeVar is bound to Type[Enum], would be good to give it a more descriptive name.

stdlib/enum.pyi
```diff
@@ -163,16 +163,22 @@ class Enum(metaclass=EnumMeta):
...
+if sys.version_info >= (3, 11):
+    class ReprEnum(Enum): ...
+
+if sys.version_info >= (3, 11):
+    class IntEnum(int, ReprEnum):
+        _value_: int
         @property
         def value(self) -> int: ...
-    else:
+        def __new__(cls: type[Self], value: int | Self) -> Self: ...
```
JelleZijlstra, line 174: Self here is a subclass of int, so `int | Self` seems redundant.

JelleZijlstra (approved):
Noticed two things that aren't new to this diff. The changes look good.

PR #7364, opened 2022-02-22
Add `__all__` to most modules beginning with 'q', 'r' and 's'

AlexWaygood:


stdlib/quopri.pyi
```diff
@@ -1,5 +1,7 @@
 from typing import BinaryIO
 
+__all__ = ["encode", "decode", "encodestring", "decodestring"]
```
JelleZijlstra, line 3: It has a comment `# (Dec 1991 version).`, that module is older than me

stdlib/smtpd.pyi
```diff
@@ -4,6 +4,8 @@ import socket
 from collections import defaultdict
 from typing import Any
 
+__all__ = ["SMTPChannel", "SMTPServer", "DebuggingServer", "PureProxy", "MailmanProxy"]
```
JelleZijlstra, line 7: MailmanProxy is gone in 3.11 (python/cpython#26617). I'll send a separate PR for that.

JelleZijlstra (approved):

PR #7366, opened 2022-02-23
Address parameter of datagram_received can be of type tuple[int, int]

thomascellerier:
I am writing a library that implements a small subset of the pyroute2 functionality but purely using asyncio.
It uses the RTNETLINK protocol to interract with the linux kernel for functionality similar to what the `ip` command provides.
I noticed that type annotation for `datagram_received` is wrong in that case.

I'm not sure if my simple fix here is the right approach as it might cause some typecheck errors for more standard usages.
Maybe there is a way using literals and overload to return a different typestub for `DatagramProtocol` if the family is set to `socket.AF_NETLINK` in `create_datagram_endpoint`.

As can be seen by running the following reproducer script:
```sh
$ python3 test.py
datagram received
data <class 'bytes'> b'4\x00\x00\x00\x02\x00\x00\x00\x00\x00\x00\x00w(\x00\x00\xff\xff\xff\xff \x00\x00\x00\x10\x00\x01\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x01\x00\x00\x00\x00\x00\x00\x00\xff\xff\xff\xff'
addr <class 'tuple'> (0, 0)
```

```Python
import asyncio
import socket
import struct


# See https://man7.org/linux/man-pages/man7/rtnetlink.7.html
NL_MSG_HEADER_FORMAT = b"IHHII"
NL_MSG_HEADER_SIZE = struct.calcsize(NL_MSG_HEADER_FORMAT)

IF_INFO_MSG_FORMAT = b"BxHiII"
IF_INFO_MSG_SIZE = struct.calcsize(IF_INFO_MSG_FORMAT)


# See <linux/netlink.h>
NETLINK_ROUTE = 0
NLM_F_REQUEST = 1


# See <linux/rtnetlink.h>
RTM_NEWLINK = 16


class NetlinkProtocol(asyncio.DatagramProtocol):
    def __init__(self, event):
        self._event = event

    def connection_made(self, transport: asyncio.BaseTransport) -> None:
        sock = transport.get_extra_info("socket")
        sock.bind((0, 0))

    def datagram_received(self, data, addr):
        print("datagram received")
        print("data", type(data), data)
        print("addr", type(addr), addr)
        self._event.set()


async def main():
    event = asyncio.Event()
    transport, protocol = await asyncio.get_running_loop().create_datagram_endpoint(
        lambda: NetlinkProtocol(event), family=socket.AF_NETLINK, proto=NETLINK_ROUTE
    )

    nlmsg_len = NL_MSG_HEADER_SIZE + IF_INFO_MSG_SIZE
    nlmsg_header = struct.pack(NL_MSG_HEADER_FORMAT, nlmsg_len, RTM_NEWLINK, NLM_F_REQUEST, 0, 0)
    nlmsg_data = struct.pack(IF_INFO_MSG_FORMAT, socket.AF_UNSPEC, 0, 1, 0, 0xFFFFFFFF)
    nlmsg = nlmsg_header + nlmsg_data

    transport.sendto(nlmsg)
    await event.wait()


asyncio.run(main())
```


srittau:
If I understand it correctly, the first element of `addr` depends on the protocol. It might be better to just use `Any` in that case so that asserting the correct type is not necessary. One more case, where python/typing#566 would come in handy.

thomascellerier:
Yes that seems to match exactly the issues in https://github.com/python/typing/issues/566

For the time being should I change it to `Any` or drop the PR altogether. `Any` makes the checks worse for common cases even though its more correct.

Using complicated overloads on `create_datagram_endpoint` and making `DatagramProtocol` generic could also work but might not be worth the effort.

JukkaL:
It's also possible to use `<common type> | Any`, if one item of the union is vastly more common than the other one(s). In this case it would mean `tuple[str | Any, int]`. You can add a comment that explains why you are using `| Any`. This may allow type checkers to do better reachability analysis and avoid false negatives in the rare case while still catching errors in the common case.

srittau (approved):

PR #7373, opened 2022-02-23
Add `__all__` for most modules beginning with 't'

AlexWaygood:


srittau:
primer seems to be correct. At least `EXACT_TOKEN_TYPES` is missing from `token.__all__`.

AlexWaygood:
> primer seems to be correct. At least `EXACT_TOKEN_TYPES` is missing from `token.__all__`.

Thanks, I'll take another look at that one :/

AlexWaygood:
> primer seems to be correct. At least `EXACT_TOKEN_TYPES` is missing from `token.__all__`.

primer is correct, but so is `__all__` as it is proposed in this PR:

```python
C:\Users\Alex\Desktop\Code dump>python
Python 3.10.2 (tags/v3.10.2:a58ebcc, Jan 17 2022, 14:12:15) [MSC v.1929 64 bit (AMD64)] on win32
>>> import token
>>> 'EXACT_TOKEN_TYPES' in token.__all__
False
>>> exit()

C:\Users\Alex\Desktop\Code dump>conda activate py36

(py36) C:\Users\Alex\Desktop\Code dump>python
Python 3.6.12 |Anaconda, Inc.| (default, Sep  9 2020, 00:29:25) [MSC v.1916 64 bit (AMD64)] on win32
>>> import token
>>> 'EXACT_TOKEN_TYPES' in token.__all__
False
```

Although `EXACT_TOKEN_TYPES` is present in the module, it is _not_ exposed as a public attribute.

hauntsaninja:
Hmm, seems potentially unfortunate.
Does adding `__all__` remove all value of definitions not in `__all__`? I guess there was general consensus on typing-sig that `__all__` should be one of a couple symbols to determine (public) contents of a module, so the initial error is fine by me. But my question is what happens next, after the initial error? Do type checkers use the private type or do they use Any?

JelleZijlstra:
There's actually bug in https://github.com/python/typeshed/blob/master/stdlib/tokenize.pyi: `tokenize.EXACT_TOKEN_TYPES` exists at runtime but not in our stub. Alex, could you add it to tokenize.pyi in this PR?

@hauntsaninja Adding `__all__` should only directly affect the behavior of `import *`. `__all__` can be used as a proxy to decide whether some value is meant to be public and should be included in a stub, but type checkers should let you refer to all symbols in the stub, including ones that are not in `__all__`.

AlexWaygood:
> There's actually bug in https://github.com/python/typeshed/blob/master/stdlib/tokenize.pyi: `tokenize.EXACT_TOKEN_TYPES` exists at runtime but not in our stub.

Aah, I see, at runtime this happens:

```python 
from token import *
from token import EXACT_TOKEN_TYPES
```

In the stub we've previously only had `from token import *`, and we got away with it, because we didn't define `__all__`. But now that we _are_  defining `__all__`, `EXACT_TOKEN_TYPES` is no longer included in the wildcard import in `tokenize.pyi`.

So we just need to match the runtime imports.

JelleZijlstra (approved):

hauntsaninja:
@JelleZijlstra Oops, you're right. I confused "type checkers determining what is public" with mypy's `--no-implicit-reexport` option (which is part of `--strict`). mypy's `--no-implicit-reexport` does prevent you from referring to a given symbol (that exists in the stub), but it only applies to imports. Anyway, this is a reason why mypy's `module_hidden` is not `not module_public`, which you'd think I'd know, given how many times I've fixed mypy code that assumes that.

PR #7377, opened 2022-02-24
docutils: Add missing dunders to nodes.Element

not-my-profile:
Adding the dunders is important since
they are not covered by __getattr__.


stubs/docutils/docutils/nodes.pyi
```diff
@@ -10,8 +11,27 @@ class Node:
     def __getattr__(self, __name: str) -> Any: ...  # incomplete
 
 class Element(Node):
+    children: list[Node]
     def __init__(self, rawsource: str = ..., *children: Node, **attributes): ...
     def __getattr__(self, __name: str) -> Any: ...  # incomplete
+    def __len__(self) -> int: ...
+    def __contains__(self, key: str | Node) -> bool: ...
+    @overload
+    def __getitem__(self, key: str) -> Any: ...
```
srittau, line 20: Why does this return `Any`? What's an attribute in docutils?

stubs/docutils/docutils/nodes.pyi
```diff
@@ -10,8 +11,27 @@ class Node:
     def __getattr__(self, __name: str) -> Any: ...  # incomplete
 
 class Element(Node):
+    children: list[Node]
     def __init__(self, rawsource: str = ..., *children: Node, **attributes): ...
     def __getattr__(self, __name: str) -> Any: ...  # incomplete
```
srittau, line 16: Nit: I think it would be clearer to move this to the bottom, as its basically an "... etc.".

stubs/docutils/docutils/nodes.pyi
```diff
@@ -1,4 +1,5 @@
-from typing import Any
+from _typeshed import Self
+from typing import Any, Iterable, overload
```
srittau, line 2: `Iterable` should be imported from `collections.abc`.

stubs/docutils/docutils/nodes.pyi
```diff
@@ -10,8 +11,27 @@ class Node:
...
+    @overload
+    def __getitem__(self, key: int) -> Node: ...
+    @overload
+    def __getitem__(self, key: slice) -> list[Node]: ...
+    @overload
+    def __setitem__(self, key: str, item: Any) -> None: ...
+    @overload
+    def __setitem__(self, key: int, item: Node) -> None: ...
+    @overload
+    def __setitem__(self, key: slice, item: list[Node]) -> None: ...
```
srittau, line 30:
```suggestion
    def __setitem__(self, key: slice, item: Iterable[Node]) -> None: ...
```

srittau (commented):
Thanks, a few nits and a question below.

stubs/docutils/docutils/nodes.pyi
not-my-profile: The attributes are strings mapped to arbitrary objects.

srittau (approved):

PR #7380, opened 2022-02-25
docutils: complete nodes.Node & bump version to 0.18.*

not-my-profile:
The version bump is necessary because
Node.findall() was introduced in 0.18.1.


not-my-profile:
Passing a type as the `condition` to `Node.findall` is equivalent to `lambda n: isinstance(n, type)` ... hence the overload:

```python
@overload
def findall(self, condition: type[_N], ...) -> Generator[_N, None, None]: ...
@overload
def findall(self, condition: Callable[[Node], bool] = ..., ...) -> Generator[Node, None, None]: ...
```

I intentionally omitted the `= ...` for `condition` in the first overload because otherwise, Pyright picks the first overload for `node.findall()` and infers ` _N@findall` as a return type, which is obviously wrong.

Omitting the default value for the first overload however results in `mypy.stubtest` failing:

>error: docutils.nodes.Node.findall is inconsistent, runtime argument "condition" has a default value of type None, which is incompatible with stub argument type Union[Type[_N`-1], def (docutils.nodes.Node) -> builtins.bool]. This is often caused by overloads failing to account for explicitly passing in the default value.

What do you suggest?

Inferring `_N@findall` for the return type shouldn't lead to type checking errors since `_N` is bounded by `Node` but I'd still prefer type checkers to infer `Node` as a return type for `node.findall()` because `_N@findall` might confuse users looking at the type tooltips in their IDE.

stubs/docutils/docutils/nodes.pyi
```diff
@@ -1,15 +1,73 @@
...
+    def walk(self, visitor: NodeVisitor) -> bool: ...
+    def walkabout(self, visitor: NodeVisitor) -> bool: ...
+    @overload
+    def findall(
+        self, condition: type[_N], include_self: bool = ..., descend: bool = ..., siblings: bool = ..., ascend: bool = ...
+    ) -> Generator[_N, None, None]: ...
+    @overload
+    def findall(
+        self,
+        condition: Callable[[Node], bool] = ...,
```
Akuli, line 38: As stubtest's error message says, you forgot that the default value `None` can be also passed explicitly :) So `node.findall(condition=None)` should work. It's a bit weird, but shouldn't be banned.

```suggestion
        condition: Callable[[Node], bool] | None = ...,
```

Akuli (commented):
Didn't read most of the PR yet, but this should answer your question.

Akuli:
> Sidenote: I unfortunately cannot check if mypy infers the same because of https://github.com/python/mypy/issues/12241.

You can set the MYPYPATH environment variable:

```
(env) akuli@akuli-desktop:~/typeshed$ cat asd.py
from docutils.nodes import Element
e: Element
reveal_type(e.findall())

(env) akuli@akuli-desktop:~/typeshed$ MYPYPATH=stubs/docutils/ mypy asd.py
asd.py:3: note: Revealed type is "typing.Generator[docutils.nodes.Node, None, None]"
```

stubs/docutils/docutils/nodes.pyi
not-my-profile: Ah thanks! I didn't read the error message carefully enough.

stubs/docutils/docutils/nodes.pyi
```diff
@@ -1,15 +1,73 @@
...
+
 class Node:
     parent: Node | None
     source: str | None
     line: int | None
     document: document | None
-    def __getattr__(self, __name: str) -> Any: ...  # incomplete
+    def __bool__(self) -> Literal[True]: ...
+    def asdom(self, dom: ModuleType | None = ...) -> xml.dom.minidom.Element: ...
+    # ModuleType is inaccurate, you cannot pass any module, but it's the best we can express
```
JelleZijlstra, line 23:
```suggestion
    # ModuleType is inaccurate, you cannot pass any module, but it's the best we can express
    def asdom(self, dom: ModuleType | None = ...) -> xml.dom.minidom.Element: ...
```
Put the comment before the function it refers to

JelleZijlstra: Also, this could be typed precisely with a Protocol if you want.

not-my-profile: Thanks, I changed it use a Protocol.

stubs/docutils/docutils/nodes.pyi
```diff
@@ -1,15 +1,75 @@
+import xml.dom.minidom
 from _typeshed import Self
-from collections.abc import Iterable
-from typing import Any, overload
+from collections.abc import Iterable, Sequence
+from typing import Any, Callable, Generator, Protocol, TypeVar, overload
```
srittau, line 4:
```suggestion
from collections.abc import Callable, Generator, Iterable, Sequence
from typing import Any, Protocol, TypeVar, overload
```

stubs/docutils/docutils/nodes.pyi
```diff
@@ -1,15 +1,75 @@
...
+from typing_extensions import Literal
 
 from docutils.transforms import Transformer
 
+class NodeVisitor:
```
srittau, line 9: This class should be moved below the "type stuff", ideally about the same position where it resides in the implementation.

stubs/docutils/docutils/nodes.pyi
```diff
@@ -1,15 +1,75 @@
...
+    def __init__(self, document: document): ...
+    def __getattr__(self, __name: str) -> Any: ...  # incomplete
+
+_N = TypeVar("_N", bound=Node)
+
+class _DomModule(Protocol):
+    Document: type[xml.dom.minidom.Document]
+
 class Node:
+    children: Sequence[Node]
```
srittau, line 19: We should add a note that this is never initialized on `Node` itself, but by its subclasses.

stubs/docutils/docutils/nodes.pyi
```diff
@@ -1,15 +1,75 @@
...
     source: str | None
     line: int | None
     document: document | None
-    def __getattr__(self, __name: str) -> Any: ...  # incomplete
+    def __bool__(self) -> Literal[True]: ...
+    def asdom(self, dom: _DomModule | None = ...) -> xml.dom.minidom.Element: ...
+    def pformat(self, indent: str = ..., level: int = ...) -> str: ...
+    def copy(self: Self) -> Self: ...
+    def deepcopy(self: Self) -> Self: ...
+    def astext(self) -> str: ...
```
srittau, line 29: As these are abstract, they should be annotated using `@abc.abstractmethod`, even if they are not annotated in the implementation. This also means that they needs to be copied to `Text` (as non-abstract).

stubs/docutils/docutils/nodes.pyi
```diff
@@ -34,6 +94,17 @@ class Element(Node):
     def __iadd__(self: Self, other: Node | Iterable[Node]) -> Self: ...
     def __getattr__(self, __name: str) -> Any: ...  # incomplete
 
+class Text(Node, str):
+    tagname: str
+    children: tuple[()]
```
srittau, line 99: Both of these should probably marked as `ClassVar`.

srittau (commented):
Thanks, a few notes below.

stubs/docutils/docutils/nodes.pyi
not-my-profile: What is the benefit of adding `@abc.abstractmethod`?

stubs/docutils/docutils/nodes.pyi
not-my-profile: I added it for `tagname`. For `children` I cannot because then mypy.stubtest fails with:

> Cannot override instance variable (previously declared on base class "Node") with class variable

stubs/docutils/docutils/nodes.pyi
srittau: Documentation and tooling can - in theory - flag improper use of these methods.

not-my-profile: What would improper use be?

AlexWaygood: It's a bit of a weird example, given that it's a deprecated module, but it's a good example: `asynchat.async_chat` doesn't have any abstract methods at runtime, and doesn't have `ABCMeta` as its metaclass. But its docstring nonetheless states that it is an "abstract class", and so we decorate the abstract methods with `@abstractmethod` in typeshed. As a result, mypy will raise an error if you try to directly instantiate the class (the proper way to use the class is to create a concrete subclass that overrides the abstract methods, and then instantiate an instance of the subclass).

- Runtime: https://github.com/python/cpython/blob/4558af5a8f8e56a9b0dc11f6e834c47e0fd05f9e/Lib/asynchat.py#L60-L62
- typeshed stub: https://github.com/python/typeshed/blob/64133df6eb3f617c2e2b055a66943fc4503d82ce/stdlib/asynchat.pyi#L9-L16
- mypy treatment of this: https://mypy-play.net/?mypy=latest&python=3.10&gist=947e63ddae68c9ab2269ac7ce42da22d

I'm not aware whether pyright also has this feature, but the point is that it _could_ implement this feature if it wanted to.

not-my-profile: Thanks for the thorough explanation and the additional example! I asked Eric about adding that check to pyright but declined since the check is expensive.

Anyway I added the `@abstactmedhod` decorators but now mypy.stubtest is failing:

> If it is meant to be abstract, add 'abc.ABCMeta' as an explicit metaclass

srittau: Thanks. `pformat` and `astext` should also be abstract. Stubtest is failing, because these four methods now need to be copied to `Element` and `document` (non-abstract). Sorry for the hassle.

srittau (approved):

PR #7386, opened 2022-02-25
Add stripe.util.convert_to_stripe_object.

rchen152:
Adding this undocumented method because at least one pytype user wants to use it:
https://github.com/google/pytype/issues/1139.

I did my best to guess the type signature from reading the source code:
https://github.com/stripe/stripe-python/blob/713015ebc937a198e3f8df5af13a839fedd37925/stripe/util.py#L139.

Are recursive types allowed in typeshed? mypy_test seems to pass
locally, but I can just change _ListRespType to Any if it seems like a
bad idea.


rchen152:
Ah, ok, stubtest failed with:
```
error: not checking stubs due to mypy build errors:
stubs/stripe/stripe/util.pyi:25: error: Cannot resolve name "_ListRespType" (possible cyclic definition)
```
I'll remove the recursive type.

srittau (approved):

PR #7392, opened 2022-02-27
Improve `types.pyi`

AlexWaygood:
- Add `types.FunctionType.__builtins__` (and `types.LambdaType.__builtins__`), currently missing from the stub.
- Make many fields on many classes read-only, as they are at runtime.
- All parameters in `types.MethodType.__init__` are positional-only.


srittau (approved):
Looks reasonable, although I didn't verify.

PR #7396, opened 2022-02-28
Add asyncio.windows_utils.Popen

JelleZijlstra:
As requested in #7356, cc @AlexWaygood.﻿

### DIFF
--- a/stdlib/asyncio/windows_utils.pyi
+++ b/stdlib/asyncio/windows_utils.pyi
@@ -41,9 +41,9 @@ if sys.platform == "win32":
         # The docstring says "The stdin, stdout, stderr are None or instances of PipeHandle."
         # but that is incorrect: if you pass in a file object, that will still be
         # put in the stdin/stdout/stderr attributes.
-        stdin: PipeHandle | IO[AnyStr] | None  # type: ignore[override]
-        stdout: PipeHandle | IO[AnyStr] | None  # type: ignore[override]
-        stderr: PipeHandle | IO[AnyStr] | None  # type: ignore[override]
+        stdin: PipeHandle | IO[AnyStr] | None  # type: ignore[assignment]
+        stdout: PipeHandle | IO[AnyStr] | None  # type: ignore[assignment]
+        stderr: PipeHandle | IO[AnyStr] | None  # type: ignore[assignment]
         # For simplicity we omit the full overloaded __new__ signature of
         # subprocess.Popen. The arguments are mostly the same, but
         # subprocess.Popen takes other positional-or-keyword arguments before
--- a/stdlib/asyncio/windows_utils.pyi
+++ b/stdlib/asyncio/windows_utils.pyi
@@ -56,3 +56,11 @@ if sys.platform == "win32":
             stderr: subprocess._FILE | None = ...,
             **kwds: Any,
         ) -> Self: ...
+        def __init__(
+            self,
+            args: subprocess._CMD,
+            stdin: subprocess._FILE | None = ...,
+            stdout: subprocess._FILE | None = ...,
+            stderr: subprocess._FILE | None = ...,
+            **kwds: Any,
+        ) -> None: ...
--- a/stdlib/asyncio/windows_utils.pyi
+++ b/stdlib/asyncio/windows_utils.pyi
@@ -2,7 +2,7 @@ import subprocess
 import sys
 from _typeshed import Self
 from types import TracebackType
-from typing import Callable, Protocol
+from typing import Any, AnyStr, Callable, Protocol
 from typing_extensions import Literal
 
 if sys.platform == "win32":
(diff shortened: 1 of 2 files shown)


AlexWaygood:
Thank you!!

stdlib/asyncio/windows_utils.pyi
```diff
@@ -36,3 +36,23 @@ if sys.platform == "win32":
...
         def fileno(self) -> int: ...
         def close(self, *, CloseHandle: Callable[[int], None] = ...) -> None: ...
+
+    class Popen(subprocess.Popen[AnyStr]):
+        # The docstring says "The stdin, stdout, stderr are None or instances of PipeHandle."
+        # but that is incorrect: if you pass in a file object, that will still be
+        # put in the stdin/stdout/stderr attributes.
+        stdin: PipeHandle | IO[AnyStr] | None
+        stdout: PipeHandle | IO[AnyStr] | None
+        stderr: PipeHandle | IO[AnyStr] | None
```
JelleZijlstra, line 46:
```suggestion
        stdin: PipeHandle | IO[AnyStr] | None  # type: ignore[override]
        stdout: PipeHandle | IO[AnyStr] | None  # type: ignore[override]
        stderr: PipeHandle | IO[AnyStr] | None  # type: ignore[override]
```

AlexWaygood: Needs to be `[assignment]` rather than `[override]`

JelleZijlstra: Huh, thanks. That's a weird code to use there; I'm not even assigning anything.

AlexWaygood (approved):

stdlib/asyncio/windows_utils.pyi
```diff
@@ -36,3 +36,31 @@ if sys.platform == "win32":
         def handle(self) -> int: ...
         def fileno(self) -> int: ...
         def close(self, *, CloseHandle: Callable[[int], None] = ...) -> None: ...
+
+    class Popen(subprocess.Popen[AnyStr]):
+        # The docstring says "The stdin, stdout, stderr are None or instances of PipeHandle."
+        # but that is incorrect: if you pass in a file object, that will still be
+        # put in the stdin/stdout/stderr attributes.
```
Akuli, line 43: Are you sure this is correct? The attributes are set to file objects in `super().__init__()`, but later replaced in the `else:` below.

JelleZijlstra: I might be misreading the code; I don't have access to Windows myself. But I think it sets these attrs to a PipeHandle only if stdin == PIPE, so if you pass stdin=<some file object>, that file object ends up in self.stdin.

AlexWaygood: Looks like @Akuli is correct. If I apply this patch(!) to `asyncio/windows_utils.py` in my local clone of CPython:

```diff
diff --git a/Lib/asyncio/windows_utils.py b/Lib/asyncio/windows_utils.py
index ef277fac3e..24dda3263d 100644
--- a/Lib/asyncio/windows_utils.py
+++ b/Lib/asyncio/windows_utils.py
@@ -171,3 +171,4 @@ def __init__(self, args, stdin=None, stdout=None, stderr=None, **kwds):
                 os.close(stdout_wfd)
             if stderr == PIPE:
                 os.close(stderr_wfd)
+        print(f'{type(self.stdin)=}')
```

... then this is what I get in an interactive shell:

```python
Python 3.11.0a5+ (main, Feb 11 2022, 10:54:07) [MSC v.1929 64 bit (AMD64)] on win32
>>> import sys
>>> from asyncio.windows_utils import Popen
>>> with open('foo.txt', 'w') as f:
...     p = Popen(sys.executable, stdin=f)
...
type(self.stdin)=<class 'NoneType'>
```

Akuli: I'm starting to think that maybe you aren't supposed to pass in file objects at all, only `PipeHandle`s or `None` or `PIPE`.

For testing this on windows, maybe print each of `self.stdin`, `self.stdout` and `self.stderr`, and then run a subprocess with some high level function? As the file name suggests, this class seems to be mostly a platform-specific lower level implementation detail.

JelleZijlstra: @AlexWaygood Interesting, I'll have to look at the code to figure out how that's possible.

@Akuli Interestingly, `Popen` is not actually used or documented anywhere else in asyncio. It looks more like they kept it in by accident.

AlexWaygood: We could always just cop out and use `Any`...

Akuli: It is used, at least in Python 3.9 where `windows_events.py` has this code:
```
class _WindowsSubprocessTransport(base_subprocess.BaseSubprocessTransport):

    def _start(self, args, shell, stdin, stdout, stderr, bufsize, **kwargs):
        self._proc = windows_utils.Popen(
            args, shell=shell, stdin=stdin, stdout=stdout, stderr=stderr,
            bufsize=bufsize, **kwargs)
```

AlexWaygood: Yes, the use is still there in 3.11 https://github.com/python/cpython/blob/df9f7597559b6256924fcd3a1c3dc24cd5c5edaf/Lib/asyncio/windows_events.py#L890

Akuli (approved):
This matches the docstring, which is the best we can make things in asyncio as the source code is complicated.

PR #7399, opened 2022-02-28
Add missing constant from urllib3.util.ssl_

WhyNotHugo:
This constant is missing from type stubs but present in the source.
Importing it works, and it's used in various different places.


srittau (approved):
Thanks!

PR #4817, opened 2020-12-13
Fix pkg_resource functions

Dreamsorcerer:
These functions are actually bound methods from working_set.

Any differences in the signatures (e.g. `Sequence[str]` in `require()` are bugs.

PR #7400, opened 2022-02-28
random: add new default args in py311

hauntsaninja:
Added in bpo-46737


stdlib/random.pyi
```diff
@@ -154,9 +159,16 @@ def triangular(low: float = ..., high: float = ..., mode: float | None = ...) ->
...
 def gammavariate(alpha: float, beta: float) -> float: ...
-def gauss(mu: float, sigma: float) -> float: ...
+
+if sys.version_info >= (3, 11):
+    def gauss(mu: float = ..., sigma: float = ...) -> float: ...
+    def normalvariate(mu: float = ..., sigma: float = ...) -> float: ...
+
+else:
+    def gauss(mu: float, sigma: float) -> float: ...
+    def normalvariate(mu: float, sigma: float) -> float: ...
```
Akuli, line 169: I wonder if replacing these with `pkg_resources.WorkingSet` style aliases would be possible, to better match to runtime.

Akuli (approved):

stdlib/random.pyi
hauntsaninja: Good point, we should do that. I think for a long time pytype did not support bound methods, but now it does.

hauntsaninja: Yup, seems like we should be able to do this now: https://github.com/python/typeshed/pull/4817#issuecomment-744091994

PR #7408, opened 2022-03-01
random: use bound methods

hauntsaninja:
Suggested by Akuli in #7400

Matches runtime better and minimises duplicated definitions.

We use bound methods in a couple other places, e.g. multiprocessing and pkg_resources.
pytype only added support for bound methods about a year ago, which is presumably
why the stubs weren't already this way.


stdlib/random.pyi
```diff
@@ -122,53 +122,29 @@ class SystemRandom(Random):
...
-def choices(
-    population: SupportsLenAndGetItem[_T],
-    weights: Sequence[float] | None = ...,
-    *,
-    cum_weights: Sequence[float] | None = ...,
-    k: int = ...,
-) -> list[_T]: ...
-def shuffle(x: MutableSequence[Any], random: Callable[[], float] | None = ...) -> None: ...
 
+_inst: Random
```
JelleZijlstra, line 126:
```suggestion
_inst: Random = ...
```
To placate flake8

srittau (approved):

PR #7411, opened 2022-03-01
6-12 months -> six months

srittau:
"At least 6-12 months" is equivalent to "at least six months" and we
usually remove after six month, unless there are issues.


hauntsaninja (approved):
Good by me!

Akuli (approved):

PR #7412, opened 2022-03-01
`asyncio.unix_events`: Make `AbstractChildWatcher` abstract, add `PidfdChildWatcher`

AlexWaygood:
Refs #7356


stdlib/asyncio/unix_events.pyi
```diff
@@ -48,14 +57,25 @@ if sys.platform != "win32":
     else:
         __all__ = ["SelectorEventLoop", "AbstractChildWatcher", "SafeChildWatcher", "FastChildWatcher", "DefaultEventLoopPolicy"]
 
-    class BaseChildWatcher(AbstractChildWatcher):
+    class BaseChildWatcher(AbstractChildWatcher, metaclass=ABCMeta):
```
srittau, line 60: Why do we need the metaclass? At least in 3.9, this does not match implementation.

AlexWaygood: I believe https://github.com/python/typeshed/pull/7185#issuecomment-1038437825 will bite us if we don't lie about the metaclass, but I'm happy to try it.

AlexWaygood: I was correct: https://github.com/python/typeshed/runs/5375362796?check_suite_focus=true

srittau: I'd prefer to `# type: ignore[misc]` this.

AlexWaygood: Neither `# type: ignore[misc]` nor a blanket `# type: ignore` will make the mypy error go away, it seems.

AlexWaygood: Seems like this mypy feature actually exists specifically to help spot typos in stubs https://github.com/python/mypy/blob/187f39461df068766a3129fa99ac376fc5a7143b/mypy/semanal_classprop.py#L95-L111

srittau: "Feature"

srittau: Could you add a comment that the implementation doesn't have that metaclass?

AlexWaygood:
```suggestion
    # Doesn't actually have ABCMeta metaclass at runtime, but mypy complains if we don't have it in the stub.
    # See discussion in #7412
    class BaseChildWatcher(AbstractChildWatcher, metaclass=ABCMeta):
```

srittau (approved):

AlexWaygood:
Thanks!

PR #7416, opened 2022-03-01
[just testing mypy-primer] union return types

not-my-profile:


JelleZijlstra:
I think mypy 0.931 still has a custom plugin for `open()`, so the typeshed signature doesn't actually affect it. But the new mypy release should drop that plugin. I don't know why your changes makes the plugin crash, though.

PR #7418, opened 2022-03-02
Let `build_ext` inherit from `setuptools.Command`

BvB93:
The `build_ext` class was previously missing inheritance from `setuptools.Command`, and hence subtypes were considered invalid mapping values for the `cmdclass` parameter back in `setuptools.setup`.

Examples
----------
``` python
In [1]: from setuptools.command.build_ext import build_ext

In [2]: build_ext.mro()
Out[2]: 
[setuptools.command.build_ext.build_ext,
 distutils.command.build_ext.build_ext,
 setuptools.Command,
 distutils.cmd.Command,
 object]
```


stubs/setuptools/setuptools/command/build_ext.pyi
```diff
@@ -1,14 +1,16 @@
...
+
 have_rtld: bool
 use_stubs: bool
 libtype: str
 
 def if_dl(s): ...
 def get_abi3_suffix(): ...
 
-class build_ext(_build_ext):
+class build_ext(_build_ext, Command):
```
JelleZijlstra, line 13: This might need to be type ignored.

JelleZijlstra (commented):
How does this work exactly? At runtime I just see `class build_ext(_build_ext):`; is there some monkeypatching going on?

stubs/setuptools/setuptools/command/build_ext.pyi
BvB93: Done as of How does this work exactly? At runtime I just see class build_ext(_build_ext):; is there some monkeypatching going on?.

BvB93:
> How does this work exactly? At runtime I just see `class build_ext(_build_ext):`; is there some monkeypatching going on?

It's not exactly my area of expertise, but this seems to be the case. Based on the at the mro output `setuptools.Command` is injected as a baseclass for the distutils `build_ext` class.

Akuli:
The mro depends on whether `distutils.command.build_ext` is imported before `setuptools`:

```python3
akuli@akuli-desktop:~$ python3
>>> import distutils.command.build_ext
>>> import setuptools.command.build_ext
>>> setuptools.command.build_ext.build_ext.mro()
[<class 'setuptools.command.build_ext.build_ext'>, <class 'distutils.command.build_ext.build_ext'>, <class 'distutils.cmd.Command'>, <class 'object'>]
>>> 
akuli@akuli-desktop:~$ python3
>>> import setuptools.command.build_ext
>>> setuptools.command.build_ext.build_ext.mro()
[<class 'setuptools.command.build_ext.build_ext'>, <class 'distutils.command.build_ext.build_ext'>, <class 'setuptools.Command'>, <class 'distutils.cmd.Command'>, <class 'object'>]
>>> 
```

What practical problem do specific `build_ext` typings solve? Do you need to import `build_ext` directly into your program and use some `setuptools.Command` functionality?

I think we should do the typing like it is if you don't import `distutils.command.build_ext`, because distutils is deprecated.

BvB93:
> What practical problem do specific `build_ext` typings solve? Do you need to import `build_ext` directly into your program and use some `setuptools.Command` functionality?
> 
> I think we should do the typing like it is if you don't import `distutils.command.build_ext`, because distutils is deprecated.

The problem that prompted this PR was a false positive when passing custom a `cmdclass`, the latter expecting a `Mapping[str, type[setuptools.Command]]` subtype. As `setuptools.command.build_ext.build_ext` currently does not inherit from `setuptools.Command` in the stub file, my first reaction was to explicitly add it as a baseclass (which apparently may or may not be true during runtime as you pointed out). 

Another option here would be to widen the type of `cmdclass` from `Mapping[str, type[setuptools.Command]]` to `Mapping[str, type[distutils.core.Command]]`.

``` python
from setuptools import setup
from setuptools.command.build_ext import build_ext

class build_test_ext(build_ext):...

setup(
    name="test",
    cmdclass={"build_ext": build_test_ext},
)
```

BvB93:
> Another option here would be to widen the type of `cmdclass` from `Mapping[str, type[setuptools.Command]]` to `Mapping[str, type[distutils.core.Command]]`.

For now I've created a PR with this alternative approach at https://github.com/python/typeshed/pull/7458. I don't have a particularly strong opinion about which PR I found more desirable, as long as the problem gets fixed.

Akuli:
I merged the alternative PR.

PR #7421, opened 2022-03-02
stdlib: Improve shutil stub

not-my-profile:
`shutil.copy('foo', b'bar')` and `shutil.copy2('foo', b'bar')` both work (same if the arguments are reversed).

And several other functions do not need to return `Any` since they just return the passed `dst` argument, we can use a `TypeVar`.


stdlib/shutil.pyi
```diff
@@ -49,36 +48,30 @@ def copyfileobj(fsrc: SupportsRead[AnyStr], fdst: SupportsWrite[AnyStr], length:
...
 def copystat(src: StrOrBytesPath, dst: StrOrBytesPath, *, follow_symlinks: bool = ...) -> None: ...
-@overload
-def copy(src: StrPath, dst: StrPath, *, follow_symlinks: bool = ...) -> _PathReturn: ...
-@overload
-def copy(src: BytesPath, dst: BytesPath, *, follow_symlinks: bool = ...) -> _PathReturn: ...
-@overload
-def copy2(src: StrPath, dst: StrPath, *, follow_symlinks: bool = ...) -> _PathReturn: ...
-@overload
-def copy2(src: BytesPath, dst: BytesPath, *, follow_symlinks: bool = ...) -> _PathReturn: ...
+def copy(src: StrOrBytesPath, dst: _StrOrBytesPathT, *, follow_symlinks: bool = ...) -> _StrOrBytesPathT: ...
```
JelleZijlstra, line 51: This is incorrect. If dst is a directory, it returns the result of an `os.path.join` call, which is always a str/bytes, not an arbitrary PathLike. Similarly, in that case it is not possible to mix bytes and str. See https://bugs.python.org/issue46727 for some context on bytes paths in shutil.

stdlib/shutil.pyi
```diff
@@ -49,36 +48,30 @@ def copyfileobj(fsrc: SupportsRead[AnyStr], fdst: SupportsWrite[AnyStr], length:
...
         src: StrPath,
-        dst: StrPath,
+        dst: _StrPathT,
         symlinks: bool = ...,
         ignore: None | Callable[[str, list[str]], Iterable[str]] | Callable[[StrPath, list[str]], Iterable[str]] = ...,
         copy_function: Callable[[str, str], None] = ...,
         ignore_dangling_symlinks: bool = ...,
         dirs_exist_ok: bool = ...,
-    ) -> _PathReturn: ...
+    ) -> _StrPathT: ...
```
JelleZijlstra, line 64: This one is right, it always returns `dst` unchanged.

stdlib/shutil.pyi
```diff
@@ -88,11 +81,11 @@ _CopyFn = Union[Callable[[str, str], None], Callable[[StrPath, StrPath], None]]
 # this does not work when dst is (or is within) an existing directory.
 # (#6832)
 if sys.version_info >= (3, 9):
-    def move(src: StrPath, dst: StrPath, copy_function: _CopyFn = ...) -> _PathReturn: ...
+    def move(src: StrPath, dst: _StrPathT, copy_function: _CopyFn = ...) -> _StrPathT: ...
```
JelleZijlstra, line 84: This is incorrect; it works similarly to copy.

JelleZijlstra (changes requested):

PR #7425, opened 2022-03-02
Give precise return type for asyncio.gather(*aws)

JukkaL:
Previously this used Any types. Caller `*args` looks like a fairly
common use case, so it would be nice to give a precise type in that
case. This could give a too general return type if `gather` is called
with a large number of positional arguments with non-uniform types,
but I'd expect this to be fairly rare.


JukkaL:
Adding extra overload variants should help with the extra errors (but there would have to a limit).

JukkaL:
After some more testing, this only seems to help in a fraction of use cases. Maybe we can use variadic generics (PEP 646) once all type checkers support them.

Issue #7424, opened 2022-03-02
Missing private global variables of `threading`


eggplants:
ref: https://github.com/python/typeshed/blob/master/stdlib/threading.pyi

ref: https://github.com/python/cpython/blob/main/Lib/threading.py#L55

eggplants:
- `_profile_hook`
- `_trace_hook`
- `_SHUTTING_DOWN`
- `_active_limbo_lock`
- `_main_thread`
- `_shutdown_locks_lock`
- `_shutdown_locks`

Akuli:
If you need to access private variables, just make a pull request that adds what you need. We're not going to add them all just for the sake of completeness.

eggplants:
@Akuli So if I create a PR to provide the above, will it be incorporated?

JelleZijlstra:
Our general policy is to add private objects if someone asks for them. However, that's meant to happen when you're actually using them, not just to fill out the stub.

eggplants:
Now, I try to provide type hints for [agronholm/typeguard](https://github.com/agronholm/typeguard).

typeguard uses `threading._profile_hook` in [src/typeguard/__init__.py](https://github.com/agronholm/typeguard/blob/de6ab051309ba74a0a27840f8172697c8778ae4f/src/typeguard/__init__.py#L1165-L1194). But it raises an `attr-defined` error when `mypy --strict` due to lacking type information.

Akuli:
Alright, let's add only `_profile_hook` then :)

PR #7427, opened 2022-03-02
Add `threading._profile_hook` to `threading.pyi`

eggplants:
Ref: #7424

### DIFF
--- a/stdlib/threading.pyi
+++ b/stdlib/threading.pyi
@@ -8,8 +8,6 @@ _TF = Callable[[FrameType, str, Any], Optional[Callable[..., Any]]]
 _PF = Callable[[FrameType, str, Any], None]
 _T = TypeVar("_T")
 
-_profile_hook: _PF | None
-
 if sys.version_info >= (3, 10):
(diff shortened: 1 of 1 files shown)


stdlib/threading.pyi
```diff
@@ -8,6 +8,8 @@ _TF = Callable[[FrameType, str, Any], Optional[Callable[..., Any]]]
 _PF = Callable[[FrameType, str, Any], None]
 _T = TypeVar("_T")
 
+_profile_hook = Optional[_TF]
```
AlexWaygood, line 11:
```suggestion
_profile_hook: _TF | None
```

hauntsaninja (commented):
This should match the type of `setprofile`, right? So `_PF`?

Also nit: might be nice to move this below the `__all__` definition

eggplants:
> This should match the type of `setprofile`, right? So `_PF`?

You're right, sorry for my mistake...

hauntsaninja (approved):
Looks great, thanks!

PR #7428, opened 2022-03-02
Improve open overloads when mode is a literal union

hauntsaninja:
As pointed out by @gvanrossum in https://github.com/python/typing/issues/1096

Improves type inference in cases when we know that mode is
OpenBinaryMode, but don't know anything more specific:
```
def my_open(name: str, write: bool):
    mode: Literal['rb', 'wb'] = 'wb' if write else 'rb'
    with open(name, mode) as f:
        reveal_type(f)  # previously typing.IO[Any], now typing.BinaryIO
```

You may be tempted into thinking this is some limitation of type
checkers. mypy does in fact have logic for detecting if we match
multiple overloads and union-ing up the return types of matched
overloads. The problem is the last overload interferes with this logic.
That is, if you remove the fallback overload (prior to this PR), you'd get
`Union[io.BufferedReader, io.BufferedWriter]` in the above example.


gvanrossum:
Oh, the previous overloads try to do clever things based on specific buffering values. I hadn't noticed that when I reported this. :-(

hauntsaninja:
Most of the previous overloads still take default values for buffering, so their cleverness can mostly be ignored in the case where you don't pass anything in. The fix is still correct (afaict); the tricky part for me was realising why mypy's usual "union up the matching overloads" logic wasn't working.

gvanrossum:
I guess there are several cases:
- No buffering specified
- buffering is one of the literals in one of the previous overloads
- buffering is an int

To validate correctness of the fix I recommend manually creating calls that exercise all of the above, combined with various values for mode, and make sure that the inferred result type is still the same.

Then repeat with various unions.

JelleZijlstra (approved):
Good catch

PR #2773, opened 2019-01-30
contextmanager decorators require Generators

asottile:
Resolves #2772

Issue #2772, opened 2019-01-30
contextlib.contextmanager allows `Iterator[T]` -- should probably be `Generator[T, None, None]`?


asottile:
For instance, this script passes mypy:

```python
import contextlib
from typing import Iterator

@contextlib.contextmanager
def f() -> Iterator[int]:
    return iter([1])


with f():
    pass


with f():
    raise TypeError('wat')
```

but fails at runtime (and not in the expected way):

```console
$ python t.py 
Traceback (most recent call last):
  File "t.py", line 14, in <module>
    raise TypeError('wat')
TypeError: wat

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "t.py", line 14, in <module>
    raise TypeError('wat')
  File "/usr/lib/python3.6/contextlib.py", line 99, in __exit__
    self.gen.throw(type, value, traceback)
AttributeError: 'list_iterator' object has no attribute 'throw'
```

graingert:
contextlib.contextmanager doesn't need to be as specific as `Generator[T, None, None]`, it could take the slightly wider type:

```python
class ContextManagerArgReturnType(Protocol, Generic[T]):
    def __next__(self) -> T: ...
    def throw(\, self, exctype: Optional[Type[BaseException]], excinst: Optional[BaseException], exctb: Optional[TracebackType]) -> Any: ...

ContextManagerArgType = Callable[[], ContextManagerArgReturnType[T]]
```

asottile:
this (unfortunately) caused a production incident today -- any chance this can be reconsidered (despite the backward incompatibility)?

HacKanCuBa:
So, this is an interesting pitfall of using `Iterator[]` instead of `Generator[]` as [the docs suggest](https://docs.python.org/3/library/typing.html#typing.Generator):
> A generator can be annotated by the generic type Generator[YieldType, SendType, ReturnType]
> ...
> Alternatively, annotate your generator as having a return type of either Iterable[YieldType] or Iterator[YieldType]

I'm just bumping this up.

AlexWaygood:
Marking as deferred unless/until PEP 696 is accepted, as per https://github.com/python/typeshed/pull/7430#issuecomment-1238711970.

srittau:
See #11422 for the type var generics feature tracker.

hauntsaninja:
[...]

PR #7430, opened 2022-03-03
require Generator for `contextmanager`

asottile:
re-try of #2772 / #2773


asottile:
just to gauge impact here, while there's a lot of repositories with this mistake, the fix is pretty easy.  here's black for instance:

```diff
$ git diff | cat
diff --git a/src/black/__init__.py b/src/black/__init__.py
index c4ec99b..e6f4e80 100644
--- a/src/black/__init__.py
+++ b/src/black/__init__.py
@@ -19,7 +19,6 @@ from typing import (
     Any,
     Dict,
     Generator,
-    Iterator,
     List,
     MutableMapping,
     Optional,
@@ -1405,7 +1404,7 @@ def assert_stable(src: str, dst: str, mode: Mode) -> None:
 
 
 @contextmanager
-def nullcontext() -> Iterator[None]:
+def nullcontext() -> Generator[None, None, None]:
     """Return an empty context manager.
 
     To be used like `nullcontext` in Python 3.7.
diff --git a/src/blib2to3/pgen2/driver.py b/src/blib2to3/pgen2/driver.py
index 8fe8206..5f8a905 100644
--- a/src/blib2to3/pgen2/driver.py
+++ b/src/blib2to3/pgen2/driver.py
@@ -24,12 +24,12 @@ import sys
 from typing import (
     Any,
     cast,
+    Generator,
     IO,
     Iterable,
     List,
     Optional,
     Text,
-    Iterator,
     Tuple,
     TypeVar,
     Generic,
@@ -66,7 +66,7 @@ class TokenProxy:
         self._release_ranges: List[ReleaseRange] = []
 
     @contextmanager
-    def release(self) -> Iterator["TokenProxy"]:
+    def release(self) -> Generator["TokenProxy", None, None]:
         release_range = ReleaseRange(self._counter)
         self._release_ranges.append(release_range)
         try:
diff --git a/src/blib2to3/pgen2/parse.py b/src/blib2to3/pgen2/parse.py
index a9dc11f..f7c31bf 100644
--- a/src/blib2to3/pgen2/parse.py
+++ b/src/blib2to3/pgen2/parse.py
@@ -17,13 +17,13 @@ from . import grammar, token, tokenize
 from typing import (
     cast,
     Any,
+    Generator,
     Optional,
     Text,
     Union,
     Tuple,
     Dict,
     List,
-    Iterator,
     Callable,
     Set,
     TYPE_CHECKING,
@@ -72,7 +72,7 @@ class Recorder:
         return self._dead_ilabels.symmetric_difference(self._ilabels)
 
     @contextmanager
-    def switch_to(self, ilabel: int) -> Iterator[None]:
+    def switch_to(self, ilabel: int) -> Generator[None, None, None]:
         with self.backtrack():
             self.parser.stack = self._points[ilabel]
             try:
@@ -83,7 +83,7 @@ class Recorder:
                 self.parser.stack = self._start_point
 
     @contextmanager
-    def backtrack(self) -> Iterator[None]:
+    def backtrack(self) -> Generator[None, None, None]:
         """
         Use the node-level invariant ones for basic parsing operations (push/pop/shift).
         These still will operate on the stack; but they won't create any new nodes, or
```

AlexWaygood:
Arguably the [mypy docs](https://mypy.readthedocs.io/en/stable/kinds_of_types.html#generators) would also have to be updated:

> A basic generator that only yields values can be succinctly annotated as having a return type of either `Iterator[YieldType]` or `Iterable[YieldType]`.

There's also similar guidance in the CPython docs for [`typing.Generator`](https://docs.python.org/3/library/typing.html#typing.Generator).

srittau:
I know this is quite controversial, but I would like to see this go in rather sooner than later. The longer we wait the worse it gets, and our stubs are arguably wrong.

JelleZijlstra:
I also agree we should do this, but it's definitely going to be quite disruptive. mypy-primer is only going to show a fraction of the real-world code that needs to be changed.

I agree with @AlexWaygood that we should call this out explicitly in the mypy and typing docs; `@contextmanager` is one of the most common uses of generators and people will get confused by this error. I can help get these docs updates merged.

@asottile do you think it's possible to provide a codemod that fixes these annotations?

AlexWaygood:
^I started typing out a response, but (as is so often the case) @JelleZijlstra beat me to it, and I agree with everything he said :)

asottile:
> I also agree we should do this, but it's definitely going to be quite disruptive. mypy-primer is only going to show a fraction of the real-world code that needs to be changed.
> 
> I agree with @AlexWaygood that we should call this out explicitly in the mypy and typing docs; `@contextmanager` is one of the most common uses of generators and people will get confused by this error. I can help get these docs updates merged.
> 
> @asottile do you think it's possible to provide a codemod that fixes these annotations?

here's the start of one, based on the [approach outlined here](https://youtu.be/G1omxo5pphw) (same as `pyupgrade` / `add-trailing-comma` / `yesqa` / etc. work):

```python

import argparse
import ast
import sys
from typing import Sequence

import tokenize_rt


def is_ctxmanager(node: ast.AST, name: str) -> bool:
    return (
        isinstance(node, ast.Name) and node.id == name
    ) or (
        isinstance(node, ast.Attribute) and
        node.attr == name and
        # simple name, might be an aliased import :shrugs:
        isinstance(node.value, ast.Name)
    )


def is_iterator(node: ast.AST, name: str) -> bool:
    return (
        isinstance(node, ast.Subscript) and (
            (
                isinstance(node.value, ast.Name) and
                node.value.id == name
            ) or (
                isinstance(node.value, ast.Attribute) and
                node.value.attr == name and
                # simple name, might be an aliased import :shrugs:
                isinstance(node.value.value, ast.Name)
            )
        )
    )


class Visitor(ast.NodeVisitor):
    def __init__(self) -> None:
        self.iter_to_generator: set[tokenize_rt.Offset] = set()
        self.aiter_to_agenerator: set[tokenize_rt.Offset] = set()

    def visit_FunctionDef(self, node: ast.FunctionDef) -> None:
        if (
                len(node.decorator_list) == 1 and
                is_ctxmanager(node.decorator_list[0], 'contextmanager') and
                node.returns is not None and
                is_iterator(node.returns, 'Iterator')
        ):
            offset = tokenize_rt.Offset(
                node.returns.lineno,
                node.returns.col_offset,
            )
            self.iter_to_generator.add(offset)
        self.generic_visit(node)

    def visit_AsyncFunctionDef(self, node: ast.AsyncFunctionDef) -> None:
        if (
                len(node.decorator_list) == 1 and
                is_ctxmanager(node.decorator_list[0], 'asynccontextmanager') and
                node.returns is not None and
                is_iterator(node.returns, 'AsyncIterator')
        ):
            offset = tokenize_rt.Offset(
                node.returns.lineno,
                node.returns.col_offset,
            )
            self.aiter_to_agenerator.add(offset)
        self.generic_visit(node)


def _fix(
        i: int,
        tokens: list[tokenize_rt.Token],
        n: int,
        orig: str,
        dest: str,
) -> None:
    j = i
    try:
        while tokens[i].src != orig:
            i += 1
    except Exception:
        breakpoint()
        raise

    tokens[i] = tokens[i]._replace(src=dest)

    while tokens[i].src != '[':
        i += 1

    i += 1
    depth = 1
    while depth:
        if tokens[i].src in '([{':
            depth += 1
        elif tokens[i].src in ')]}':
            depth -= 1

        i += 1
    tokens[i - 1] = tokens[i - 1]._replace(src=f'{", None" * n}]')


def _fix_file(filename: str) -> int:
    with open(filename, 'rb') as fb:
        contents_bytes = fb.read()

    try:
        contents_text_orig = contents_text = contents_bytes.decode()
    except UnicodeDecodeError:
        print(f'{filename} is non-utf-8 (not supported)')
        return 1

    tree = ast.parse(contents_text, filename=filename)
    visitor = Visitor()
    visitor.visit(tree)

    if not visitor.iter_to_generator and not visitor.aiter_to_agenerator:
        return 0

    tokens = tokenize_rt.src_to_tokens(contents_text)
    for i, token in tokenize_rt.reversed_enumerate(tokens):
        if token.offset in visitor.iter_to_generator:
            _fix(i, tokens, n=2, orig='Iterator', dest='Generator')
        elif token.offset in visitor.aiter_to_agenerator:
            _fix(i, tokens, n=1, orig='AsyncIterator', dest='AsyncGenerator')

    contents_text = tokenize_rt.tokens_to_src(tokens)
    if contents_text == contents_text_orig:
        return 0

    print(f'Rewriting {filename}')
    with open(filename, 'w', encoding='UTF-8', newline='') as f:
        f.write(contents_text)
    return 1


def main(argv: Sequence[str] | None = None) -> int:
    parser = argparse.ArgumentParser()
    parser.add_argument('filenames', nargs='*')
    args = parser.parse_args(argv)

    if sys.version_info < (3, 10):
        raise SystemExit(
            'this script makes some assumptions about ast.Subscript and '
            'needs python 3.10+',
        )

    ret = 0
    for filename in args.filenames:
        ret |= _fix_file(filename)
    return ret


if __name__ == '__main__':
    raise SystemExit(main())
```

it doesn't fix the imports, but that's tricky to do accurately and with respecting the user's style

```python
import contextlib
from contextlib import contextmanager
from contextlib import asynccontextmanager
from typing import AsyncIterator, Iterator
import typing
import typing as t


def fn_unrelated() -> Iterator[None]:
    yield

@contextlib.contextmanager
def ctx1() -> Iterator[None]:
    yield

@contextmanager
def ctx2() -> Iterator[None]:
    yield

@contextlib.asynccontextmanager
async def ctx3() -> AsyncIterator[None]:
    yield

@asynccontextmanager
async def ctx4() -> AsyncIterator[None]:
    yield

@contextlib.contextmanager
def ctx5() -> typing.Iterator[None]:
    yield

@contextlib.contextmanager
def ctx6() -> t.Iterator[None]:
    yield
```

```diff
$ diff -u t2.py.bak t2.py
--- t2.py.bak	2022-03-03 14:24:22.591697064 -0500
+++ t2.py	2022-03-03 14:42:18.439722185 -0500
@@ -10,25 +10,25 @@
     yield
 
 @contextlib.contextmanager
-def ctx1() -> Iterator[None]:
+def ctx1() -> Generator[None, None, None]:
     yield
 
 @contextmanager
-def ctx2() -> Iterator[None]:
+def ctx2() -> Generator[None, None, None]:
     yield
 
 @contextlib.asynccontextmanager
-async def ctx3() -> AsyncIterator[None]:
+async def ctx3() -> AsyncGenerator[None, None]:
     yield
 
 @asynccontextmanager
-async def ctx4() -> AsyncIterator[None]:
+async def ctx4() -> AsyncGenerator[None, None]:
     yield
 
 @contextlib.contextmanager
-def ctx5() -> typing.Iterator[None]:
+def ctx5() -> typing.Generator[None, None, None]:
     yield
 
 @contextlib.contextmanager
-def ctx6() -> t.Iterator[None]:
+def ctx6() -> t.Generator[None, None, None]:
     yield
```

hauntsaninja:
Seems like I'm in the minority, but I'm -1 on this.

Breaking almost every use case of contextmanager for something pretty niche is too much for me.

Other than your reports, I haven't ever encountered instances of people running into this. I'm struggling to think of the real world code that motivates someone to return an iterator here too.

On the other hand, it's very common for users with correct code to get confused when adding types for Iterable, Iterator, Generator. By steering users to use the scary and not ergonomic Generator, this is a real loss of usability in the 99.99% use case going forward as well.

AlexWaygood:
It's a shame we don't have something like the following construct in `typing.py`:

```python

class SimpleGenerator(Generator[_T_co, None, None]): ...
```

Returning `SimpleGenerator[_T_co]` from these functions would be much more correct than returning `Iterator`, and much more user-friendly than having to type out `Generator[_T_co, None, None]` every time.

JelleZijlstra:
it would be a bit nicer if we had TypeVar defaults so we could just write `Generator[T]`.

hauntsaninja:
Also note that a linter could warn for contextmanager decorated functions that don't yield. Could be a good check to add to pylint or flake8-bugbear. Since this wouldn't look at annotations, it'd side step the issue of forcing user's to change their annotations.

asottile:
the case I had would not have been detected by that -- it returned an iterable from another function call.  I think this really needs type checkers to do the proper thing

AlexWaygood:
> the case I had would not have been detected by that -- it returned an iterable from another function call. I think this really needs type checkers to do the proper thing

@asottile I'm not sure I understand. If you have a function like this:

```python
@contextmanager
def func():
    return other_generator_function()
```

Then a flake8 plugin could easily require that you do this instead, ensuring type safety:

```python
@contextmanager
def func():
    yield from other_generator_function()
```

It doesn't need to do any fancy guesswork about whether `other_generator_function` is _actually_ a generator function, it just needs to make sure that `func` yields somewhere, no?

Akuli:
`Generator[T]` isn't great, because `typing.py` in standard library of released Pythons doesn't support it at runtime, and most people don't use `from __future__ import annotations`. This means we're stuck with `Iterator[T]` or `Generator[T, None, None]`.

I prefer `Iterator[T]` because it's simple. You should not have to know about sending anything into a generator just to make a context manager, and having to know about iterators is already confusing enough. To me, `Iterator[T]` satisfies "practicality beats purity", "simple is better than complex", and also "prefer false negatives".

Another possible solution would be mypy automatically inferring the correct return type for functions that `yield`, but I don't think that's ever going to happen because mypy has a concept of typed and untyped functions, and `def foo():` is untyped, regardless of whether or not it has decorators.

srittau:
> I prefer `Iterator[T]` because it's simple.

But it's also confusing. A generator function return a generator after all (which happens to be an iterator as well).

Akuli:
You really shouldn't have to care. The type annotation in this case should IMO be something simple that doesn't get in the way, not something full of `None, None` that makes you stop and think about it carefully.

srittau:
I'm still in favor of allowing `Generator[X]` for cases like this. `Iterator[X]` is just the wrong annotation and as this discussion shows problematic in some cases.

hauntsaninja:
@asottile could you share more thoughts on why your case couldn't be caught by a simple and much less disruptive new linter check? It's unclear to at least Alex and I, see https://github.com/python/typeshed/pull/7430#issuecomment-1059053121

asottile:
@hauntsaninja I addressed that here: https://github.com/python/typeshed/pull/7430#issuecomment-1058622386

asottile:
additionally, the recommendation also does not work for async where there's no `yield from` or when there's branching logic (both of which are needed for the use case at work which caused an outage)

Akuli:
So if I understood correctly, your function looks like this:

```
@contextlib.contextmanager
def foo() -> Iterator[int]:
    return bar()
```

and it broke when `bar()` was changed to return an iterator that wasn't a generator, and using `yield from` instead of `return` doesn't work in your use case.

I understand that this is hard to detect correctly with linters, but I don't understand why it means linters won't help. If your code is correct and the linter still warns about it, wouldn't a linter's ignore comment at least make you pay more attention to it while refactoring?

Now consider what happens if we change the generator's yield type to `Any` (and also use `Generator`, as you suggest):

```
@contextlib.contextmanager
def foo() -> Generator[Any, None, None]:
    return bar()
```

It would be nice if this was a linter error, even if `bar()` happens to return a generator. To me it looks like someone mistakenly wrote `return` instead of `yield`, and I expect that to be more common than actually wanting to `return` inside  `@contextlib.contextmanager`.

asottile:
it's an `@asynccontextmanager` and it has branching and defers to several generators but only on particular branches -- all of which aren't lintable without a type checker

the type checker is wrong and should be fixed, I'm not sure why there's so much resistance to keeping the type checker broken

Akuli:
It's technically not wrong:

```
>>> import contextlib
>>> @contextlib.contextmanager
... def foo():
...     return iter([1])
... 
>>> with foo():
...     print("Hi")
... 
Hi
```
but this is very unlikely to be what you wanted the code to do, and it would be good if type checkers warned about this.

> I'm not sure why there's so much resistance and insistence

It's because people use `Iterator[T]` a **lot**, and requiring everyone to change those to more complicated (and IMO more confusing) `Generator[T, None, None]` just doesn't feel like a good idea. `Generator[T]` would help, but it would take a few years until people's minimum supported Python versions have that.

A couple ideas:
- Keep `Iterator` for now, and reconsider this when `Generator[T]` becomes a thing, either with defaulted generics or with special-casing in type checkers?
- Just merge this and revert if users complain? In the past, we have done this with other pull requests that have a big `mypy_primer` diff.

JukkaL:
FWIW, I'm strongly against this change, mostly because of the confusing `Generator[T, None, None]` type, and in part because of the huge amount of changes required for arguably a pretty minor benefit. 

@Akuli 
> Keep Iterator for now, and reconsider this when Generator[T] becomes a thing, either with defaulted generics or with special-casing in type checkers?

In my opinion, this would a requirement to (potentially) make this a net improvement. If we could write `Generator[T]`, the new state would be clearly better -- it's more precise, and not (significantly) more difficult to use than what we currently have. Then the main tradeoff would be the cost of the migration vs. improved type checking effectiveness.

> Just merge this and revert if users complain? In the past, we have done this with other pull requests that have a big mypy_primer diff.

To be honest, I think that we should be more principled than this in typeshed, since the number of impacted users can be very large. If 1000 projects that use type checking each need to spend 15 minutes to fix their codebase, that's 250 hours lost.

asottile:
> If 1000 projects that use type checking each need to spend 15 minutes to fix their codebase, that's 250 hours lost.

fwiw, one outage easily costs more than that

JukkaL:
> fwiw, one outage easily costs more than that

Yeah, it *can* cost more than than, but not every outage will be this expensive? It seems hard to estimate the likelihood of a change like this preventing an outage, though -- or is there an actual outage you're familiar with? The likelihood of many users having to update code is very high, since `contextlib.contextmanager` is very popular, and `Generator` doesn't seem to be often used as the return type of the decorated function.

I have an alternative proposal that would mostly solve the original issue and wouldn't require a lot of work from end users: special case the `contextmanager` decorator in type checkers, and complain only if it's decorating a non-generator function that doesn't return `Generator[...]`. This should still catch most real errors. Implementing this check in mypy would take maybe 3 hours of effort, and even if 3 type checkers would implement this, it could still be only ~10 hours of effort.

Type checkers would allow these:

```py
@contextmanager
def f() -> Iterator[None]:
    yield None

@contextmanager
def f() -> Generator[None, None, None]:
    return <some generator>
```

But this wouldn't pass, since the decorated function isn't a generator function:

```py
@contextmanager
def f() -> Iterator[None]:
    return iter(...)
```

asottile:
> > fwiw, one outage easily costs more than that
> 
> Yeah, it can cost more than than, but not every outage will be this expensive? 

this particular typing wart has already cost 3x more than your estimates above just from my experiences, and I'm sure I'm not alone in this.  I also believe you're over-estimating the fix cost here, it is a simple find and replace to resolve the issue for the easiest case (and points out **real bugs** in the not-easiest case!).  the typeshed stubs-no-longer-included migration easily dwarfs this in toil to mypy users.  this is an exceptionally easy mistake to make [even above while arguing against the change!](https://github.com/python/typeshed/pull/7430#issuecomment-1075110200).

and again, your suggestion won't help anything except for the most basic situation (which could already be detected by a linter)

my situation looks roughly like:

```python

async def other_funtion_which_is_async_generator() -> AsyncIterator[...]:
    if some_precondition:
        return [...]  # bug was here
    elif ...:
        yield ...
        return ...

    yield ...
    return ...

@asynccontextmanger
async def ctx() -> AsyncIterator[...]:
   if some_complicated_condition:
        return other_function_which_is_async_generator()
   elif other_condition:
        async with ..."
            yield ...
            return

   yield ...
   return
```

unless your suggestion is to transparently change all "returns iterator but is a generator" functions from:

```python
def f() -> Iterator[T]:
    yield ...
```

to 

```python
def f() -> Generator[T, None, None]:
    yield ...
```

then I don't think it will help (and even then I'm not sure it helps my case given it's not yielding on all branches but returning another function)

JukkaL:
@asottile Thanks for the additional context! This change now makes more sense to me. 

>  I also believe you're over-estimating the fix cost here, it is a simple find and replace to resolve the issue for the easiest case 

Generally imports need to be adjusted as well to update code, which is somewhat hard to to do using find/replace tools.

I agree that the fix is quite easy, but figuring out what needs to be done, and investigating whether this is a false positive or a mypy issue + the required context switch, creating a PR/change request, possibly getting it reviewed, waiting for CI, etc. is included in my 15 minute estimate (which is still pretty arbitrary).

To reduce the toil, we could add a dedicated error message about the non-Generator return type with clear instructions about what needs to be done. Providing a tool to automatically upgrade simple cases in affected code would further help, if it could also update imports. 

>  the typeshed stubs-no-longer-included migration easily dwarfs this in toil to mypy users. 

Yes, we knew that to have a big cost. We decided to move forward with it since not doing it risked making every future mypy upgrade painful for larger projects (mainly due to requiring all third-party stubs to be upgraded in lockstep with mypy). We try to avoid similar big breakages whenever there is a good alternative. That's why I'm trying to figure out if there is something less disruptive we could do here.

jonathanslenders:
Thanks for considering this change.

I came here after learning that I have to wrap async generators in an ``contextlib.aclosing`` block to ensure that finally blocks get properly executed, even if it contains an ``await``.  See: https://trio.readthedocs.io/en/stable/reference-core.html#finalization

Unfortunately, ``aclosing()`` doesn't work for ``AsyncIterable[T]`` but it does for ``AsyncGenerator[T, None]``, so I ended up replacing all occurances, and at the same time wondered about contextmanagers.

``aclosing`` is here somewhat similar to ``asynccontextmanager``. Being more precise in the return type seems to be always better, unfortunately it's more typing in this case.

JelleZijlstra:
I think this is too disruptive until we have PEP-696 (TypeVar defaults).

AlexWaygood:
I also think this is just too disruptive, I'm afraid :( I'd much prefer it if type checkers could add some special-casing for this, as @JukkaL suggested.

This is one of those occasions where it really *should* be solved in the stubs... but I just don't think the backwards-compatibility break is worth the cost (and I don't want to have to tell people they need to make their type hints much uglier every time they use `@contextmanager`, either).

twoertwein:
The mypy-primer output looks disruptive but these cases are all easy to fix (for example in pandas https://github.com/pandas-dev/pandas/pull/48383). Importantly, it is only necessary to fix the few(er) `Argument 1 to "contextmanager" has incompatible type` errors, as that should take care of  all the `Need type annotation for` errors.

hauntsaninja:
There are currently three maintainers against and one maintainer for, so I'm closing this PR. We can revisit if PEP 696 is accepted.

I do think the pros and cons are slightly different for contextmanager vs asynccontextmanager, so I'm willing to consider the asynccontextmanager change on its own. This is https://github.com/python/typeshed/pull/8698

I'm also interested in hearing about non-hypothetical cases where you actually have run into issues preventable by this; please comment on issue #2772 if so.

---

As is mentioned in this PR, if you're looking for safety in the sync case, a linter can enforce use of `yield from` in `@contextmanager`.

```python
@contextmanager
def func():
    return other_generator_function()
```

A linter could ask you to change to:
```python
@contextmanager
def func():
    yield from other_generator_function()
```

You may need an empty return if you have something that does branching and you need to short circuit.

PR #7431, opened 2022-03-03
Add missing typings to `unittest.mock`

itaisteinherz:
(I found these by adding the `--disallow-incomplete-defs` flag to the mypy tests.)


stdlib/unittest/mock.pyi
```diff
@@ -428,14 +428,14 @@ if sys.version_info >= (3, 8):
     class AsyncMock(AsyncMockMixin, AsyncMagicMixin, Mock): ...  # type: ignore # argument disparities between base classes
 
 class MagicProxy:
-    name: Any
+    name: str
```
itaisteinherz, line 431: I'm not sure if I should change this, however it doesn't really make much sense to me that `name` is typed as `Any` in the module's stubs, when it's clear that it's supposed to be a string (at least from what I could tell from the CPython source).

itaisteinherz:
Note that I doubt that mypy_primer's output will be of use here, as tests are usually not type-checked, and the unittest module is usually not used in regular (non-test) code.

AlexWaygood:
If these are the only missing annotations in this file, you could also delete this line as part of this PR, increasing pyright's strictness when checking this file as part of CI:

https://github.com/python/typeshed/blob/c50cb066e02d6b53d4a261c147086ce425f81289/pyrightconfig.stricter.json#L16

itaisteinherz:
@AlexWaygood will do!

itaisteinherz:
Seems like there are some missing typings which mypy didn't report. I'll look into pyright's output.

stdlib/unittest/mock.pyi
```diff
@@ -74,21 +74,40 @@ class _Sentinel:
...
         cls: type[Self],
-        value: Any = ...,
-        name: Any | None = ...,
+        value: _ValueArgument = ...,
+        name: str | None = ...,
         parent: Any | None = ...,
-        two: bool = ...,
+        two: Literal[True] = ...,
+        from_kall: bool = ...,
+    ) -> Self[_ArgsKwargs]: ...
```
itaisteinherz, line 91: Mypy doesn't like this for some reason. Any ideas?

AlexWaygood: TypeVars currently aren't subscriptable. We'd need [higher-kinded TypeVars](https://github.com/python/typing/issues/548) for this kind of syntax.

hauntsaninja (commented):
Skimmed, the simple annotation additions look good to me. If we do want to try making `_Call` generic over something, I'd move that into its own PR. We've had at least a couple bad regressions involving `unittest.mock`, so it's something I'm wary of :-)

stdlib/unittest/mock.pyi
itaisteinherz: Right, I forgot that `Self` is a TypeVar :) I'm thinking about just removing `_CallValues` to reduce complexity.

itaisteinherz:
@hauntsaninja I agree :) I'll revert some of my newer additions.

stdlib/unittest/mock.pyi
itaisteinherz: Removed.

stdlib/unittest/mock.pyi
```diff
@@ -1,5 +1,6 @@
 import sys
 from _typeshed import Self
+from contextlib import AbstractContextManager
```
AlexWaygood, line 3:
```suggestion
from contextlib import _GeneratorContextManager
```

It's best to be as precise as possible when annotating return types, and this is the actual return type of functions decorated with `@contextmanager` :)

itaisteinherz: Cool, didn't know about it. However, I don't see it being used anywhere in the stubs of stdlib. Is that on purpose?

AlexWaygood: I don't _think_ it's on purpose 🤷

itaisteinherz: OK, done :)

AlexWaygood: Thanks!

JelleZijlstra (approved):

PR #7432, opened 2022-03-03
Add type hints for tomllib

hukkin:
The `tomllib` module will be added in Python 3.11 (but not yet merged to CPython, see https://github.com/python/cpython/pull/31498). Here's the type hints already.


AlexWaygood:
You'll want to add tomllib to https://github.com/python/typeshed/blob/master/stdlib/VERSIONS (congrats on the PEP acceptance btw!)

hauntsaninja (approved):
Looks good (once the VERSIONS change is in)!

PR #7433, opened 2022-03-03
Add tomllib.__all__

AlexWaygood:


hauntsaninja (approved):
Right, gotta catch em `__all__` :-)

hukkin:
As someone who's not that knowledgeable in typeshed things, I'm curious, does this change type checker behavior somehow?

AlexWaygood:
> As someone who's not that knowledgeable in typeshed things, I'm curious, does this change type checker behavior somehow?
> 
> EDIT: Ok I thought about it myself 😄 Perhaps this adds typing support for `from tomllib import *`?

Yup! It doesn't really make much difference for tomllib, but it makes a big difference for modules like e.g. `tokenize` that have lots of symbols that don't start with leading underscores, but aren't included in `__all__`.

Also, pyright calls us out in our CI if there are names declared in `__all__` which aren't actually defined in the stub, so it's a useful guard against typos.

PR #6819, opened 2022-01-04
Pygments: Add return types for format/highlight.

Dreamsorcerer:

PR #5652, opened 2021-06-16
Pymysql cursor types and fetching return types

Akuli:
Fixes #5567 so that `Cursor` and `DictCursor` no longer return annoying unions. I also got rid of several `Optional`s. In those methods, I looked at the source code and I tried `select * from foo` on a local test database, and couldn't get it to return `None`.

I also made `Connection` generic, so that when you create a connection, it remembers which cursor class you want to use. You can still customize this by passing a custom class to `.cursor()`.

PR #7435, opened 2022-03-04
pygments: Correct Formatter generics

not-my-profile:
933fec014c2410d9b1fb67926dbd832a87508d82 made
pygments.formatter.Formatter generic to improve the type-safety
of pygments.format() and pygments.highlight().

The actual pygments.formatter.Formatter however is not generic,
meaning that pygments.formatter.Formatter[str] should not type check
(since it leads to a runtime type error).

Furthermore the pygments ImageFormatter always expects BinaryIO,
no matter which constructor arguments are passed.

This commit addresses both problems.


stubs/Pygments/pygments/__init__.pyi
```diff
@@ -1,16 +1,16 @@
...
 
-from pygments.formatter import Formatter
+from pygments.formatter import _Formatter
 
 _T = TypeVar("_T", str, bytes)
 
 def lex(code, lexer): ...
 @overload
-def format(tokens, formatter: Formatter[_T], outfile: SupportsWrite[_T]) -> None: ...
+def format(tokens, formatter: _Formatter[_T], outfile: SupportsWrite[_T]) -> None: ...
```
Akuli, line 10: How am I supposed to call this function now? It should accept any `Formatter`, but it does not. I think hiding the generic-ness in a non-public class `_Formatter` only moves the problem.

If you get a runtime error for `Formatter[str]`, you should consider using `from __future__ import annotations`. Many other stubs have the same situation, and we are used to telling this to users. See e.g. #5652.

not-my-profile: Right good point, I didn't consider that. In that case I think we should revert #6819 because e.g. the following code:

```python
class Foo(pygments.formatter.Formatter):
    pass
```

should not yield errors when using pyright or mypy with strict settings. But currently it does because of `Missing type parameters for generic type "Formatter"`.

Akuli: Yeah, subclassing is quite annoying. One possible workaround is:

```python3
from typing import TYPE_CHECKING

if TYPE_CHECKING:
    _Base = pygments.formatter.Formatter[str]
else:
    _Base = pygments.formatter.Formatter

class Foo(_Base):
    pass
```

but it's a mess. Apart from subclassing, things should just work most of the time, because the type of e.g. `foo = Formatter(encoding="utf-8")` is `Formatter[bytes]`.

Making the actual `Formatter` inherit from `Generic` would probably be the best. In my experience, pygments maintainers merge pull requests eventually, but it usually takes a few weeks.

AlexWaygood: As @Akuli says, the best solution here is probably to get a fix merged upstream with pygments. Longer term, a general solution would probably be something like https://github.com/python/typing/issues/956 or https://github.com/python/typing/issues/307

not-my-profile: Right I opened a PR for pygments: https://github.com/pygments/pygments/pull/2081. I'll update this PR accordingly so that it only fixes the `ImageFormatter` situation.

not-my-profile:
Ok I am confused the current `Formatter.__init__` overloads do not appear to work at all:

```python
from pygments.formatters.html import HtmlFormatter

reveal_type(HtmlFormatter(encoding='utf-8'))
reveal_type(HtmlFormatter())
```

```
foo.py:3: note: Revealed type is "pygments.formatters.html.HtmlFormatter[builtins.str*]"
foo.py:4: note: Revealed type is "pygments.formatters.html.HtmlFormatter[builtins.str*]"
```

PR #7437, opened 2022-03-04
Introduce _typeshed.WeakUnion

not-my-profile:


stdlib/_typeshed/__init__.pyi
```diff
@@ -10,6 +10,25 @@ from os import PathLike
 from typing import AbstractSet, Any, Container, Generic, Iterable, Protocol, TypeVar, Union
 from typing_extensions import Final, Literal, final
 
+WeakUnion = Any
```
JelleZijlstra, line 13: what if we do this:

```suggestion
class WeakUnion(Any):
    def __class_getitem__(cls, args: Any) -> Any: ...
```

AlexWaygood: mypy doesn't (yet) have a generalised understanding of `__class_getitem__`. It only understands a class as being generic if the class inherits from `typing.Generic`. https://mypy-play.net/?mypy=latest&python=3.10&gist=1a3ed0af028c24c591ea1e63d80954f6

JelleZijlstra: That's too bad. Putting `__getitem__` on the metaclass doesn't work either. This might be easier to fix than full PEP 646 support though.

AlexWaygood: I have reservations about teaching mypy about `__class_getitem__`, as I think it should properly be seen as an implementation detail of the typing framework. But I definitely agree that it should have a stronger understanding of metaclass `__getitem__`.

not-my-profile: Can you give an example of what `__getitem__` with a metaclass would look like? And is there any type type checker that currently handles that? I think we should open issues with mypy, pyright, pytype & pyre if they don't support that currently.

JelleZijlstra:
```suggestion
class _WeakUnionMeta(type):
    def __getitem__(self, arg: Any) -> Any: ...

class WeakUnion(Any, metaclass=_WeakUnionMeta): ...
```

AlexWaygood:
```python
from typing import Any

class WeakUnionMeta(type):
    def __getitem__(self, *args: Any) -> Any: ...

class WeakUnion(metaclass=WeakUnionMeta): ...
```

JelleZijlstra: `*args` is wrong (though harmless), `__getitem__` only ever gets a single argument. `WeakUnion[A, B]` would call `_WeakUnionMeta.__getitem((A, B))`.

hauntsaninja: If we just want a dumb hack to get started, `WeakUnion: Any` (not =) should just work right?

AlexWaygood: > If we just want a dumb hack to get started, `WeakUnion: Any` (not =) should just work right?

...I feel like that shouldn't work, but,  bafflingly, it seems like it does!

https://mypy-play.net/?mypy=latest&python=3.10&flags=strict&gist=005bb4e14972e556125d0141e770cd90

hauntsaninja:
By the way, if it's appropriate to bikeshed here, I have a big preference for a name with "Any" (or possible "Unsafe") in it. The principle being that it should be obvious to users when a type involves unsafety. So candidates include:
- AnyOf
- AnyUnion
- UnsafeUnion

AlexWaygood:
> By the way, if it's appropriate to bikeshed here, I have a big preference for a name with "Any" (or possible "Unsafe") in it. The principle being that it should be obvious to users when a type involves unsafety. So candidates include:
> 
> * AnyOf
> * AnyUnion
> * UnsafeUnion

I like `UnsafeUnion` or `WeakUnion`. I agree that emphasising the unsafety is good, but I think emphasising the fact that it's _like_ a union is also good.

stdlib/_typeshed/__init__.pyi
not-my-profile: It doesn't work with pyright:

> variable not allowed unless it is a type alias

hauntsaninja: Hmm. I guess that's part of the "Unknown" vs "Any" distinction pyright has. The even dumber hack is `from _typeshed.does_not_exist import WeakUnion as WeakUnion  # type: ignore`

not-my-profile: That doesn't help pyright still treats such types it cannot import as `Unknown`.

AlexWaygood: @erictraut, how hard would it be for you to implement special-casing in pyright for a `WeakUnion`/`UnsafeUnion` symbol imported from `_typeshed`? It wouldn't need to be the full implementation of this proposed feature just yet; it would just need to be a special form semantically identical to `Any` that could be provided with arbitrary subscriptions.

erictraut: I'd prefer not to do it that way. This approach looks very hacky to me, and I don't think it should go into typeshed this way.

If it's going to be a special form, then I think we should go through a PEP process and define it formally. If we're going to use PEP 646 as I suggested, then let's do that — although we'd need to wait for some minimal support in mypy first.

If this is just going to be an `Any` for now, then there's no benefit in the short term. I don't see the need to rush it. Let's pick a sound approach and get this right.

stdlib/_typeshed/__init__.pyi
```diff
@@ -10,6 +10,31 @@ from os import PathLike
 from typing import AbstractSet, Any, Container, Generic, Iterable, Protocol, TypeVar, Union
 from typing_extensions import Final, Literal, final
 
+_WU = TypeVar("_WU")
+
+class WeakUnion(Generic[_WU], Any):
```
AlexWaygood, line 15:
```suggestion
class WeakUnion(Any, Generic[_WU]):
```

I'm curious to see if it makes any difference if we have `Any` first instead of second

AlexWaygood:
The test failures indicate that this would be very difficult to do without adding special-casing to type checkers to account for this feature. As such, I'm going to close this for now, but we can maybe revisit it after further discussion in the python/typing repo or the typing-sig mailing list.

PR #7438, opened 2022-03-04
Use latest stubtest in CI

AlexWaygood:


AlexWaygood:
cc. @hauntsaninja.

I think it's worth having another go at this now that the following two PRs have been merged:

- https://github.com/python/typeshed/pull/7352
- https://github.com/python/mypy/pull/12234

srittau (approved):

hauntsaninja (commented):
Great!

PR #7439, opened 2022-03-04
Make more miscellaneous fields read-only, annotate `_json.make_encoder`

AlexWaygood:
These are (nearly) all read-only get-set descriptors at runtime.

There are probably many more; these are only the ones that I could easily verify were read-only (not all get-set descriptors are read-only, and it's not as easy to verify which ones are read-only as it is with properties).


stdlib/_json.pyi
```diff
@@ -3,14 +3,22 @@ from typing_extensions import final
...
+    @property
+    def indent(self) -> Any: ...
+    @property
+    def markers(self) -> Any: ...
+    @property
+    def default(self) -> Any: ...
+    @property
+    def encoder(self) -> Any: ...
+    @property
+    def item_separator(self) -> Any: ...
```
srittau, line 21:
```suggestion
    @property
    def sort_keys(self): ...
    @property
    def skipkeys(self): ...
    @property
    def key_separator(self): ...
    @property
    def indent(self): ...
    @property
    def markers(self): ...
    @property
    def default(self): ...
    @property
    def encoder(self): ...
    @property
    def item_separator(self): ...
```

AlexWaygood: Actually, why don't I just add the annotations as part of this PR, I believe they're just the same as in the `__init__` function.

stdlib/asyncio/futures.pyi
```diff
@@ -42,9 +42,13 @@ if sys.version_info < (3, 7):
 
 class Future(Awaitable[_T], Iterable[_T]):
     _state: str
-    _exception: BaseException
+    @property
+    def _exception(self) -> BaseException: ...
```
JelleZijlstra, line 46: This confused me because in the Python code for futures.py, it is settable. However, the C version in `_asynciomodule.c` uses a read-only descriptor. (No change requested)

stdlib/functools.pyi
```diff
@@ -99,9 +99,12 @@ def total_ordering(cls: type[_T]) -> type[_T]: ...
 def cmp_to_key(mycmp: Callable[[_T, _T], int]) -> Callable[[_T], SupportsAllComparisons]: ...
 
 class partial(Generic[_T]):
-    func: Callable[..., _T]
-    args: tuple[Any, ...]
-    keywords: dict[str, Any]
+    @property
```
JelleZijlstra, line 102: Same here, these are read-write in Python but readonly in C `_functools.partial`. We should follow the C version as that's what people will be using virtually all of the time.

JelleZijlstra (approved):

stdlib/functools.pyi
AlexWaygood: Also: if it's only read-only in very specific circumstances, it arguably won't be type-safe for end users to try to write to that attribute in any circumstances

PR #7442, opened 2022-03-05
Upgrade to stubtest with dunder pos only reverted

hauntsaninja:


stubs/braintree/braintree/attribute_getter.pyi
```diff
@@ -2,3 +2,4 @@ from typing import Any
 
 class AttributeGetter:
     def __init__(self, attributes: Any | None = ...) -> None: ...
+    def __repr__(self, detail_list: list[str] | None = ...) -> None: ...
```
AlexWaygood, line 5:
```suggestion
    def __repr__(self, detail_list: list[str] | None = ...) -> None: ...  # type: ignore[override]
```

AlexWaygood (approved):
LGTM!

I assume we're commenting out various allowlist entries,rather than simply deleting them, because we "hope" we'll add them back soon, if we find a better way of doing this in stubtest?

hauntsaninja:
Yeah, I was thinking something along those lines. The comments are still true and still describe typeshed falsehoods, just ones no longer checked by stubtest. But happy to clean them up too, if people have a preference one way or the other

AlexWaygood:
> Yeah, I was thinking something along those lines. The comments are still true and still describe typeshed falsehoods, just ones no longer checked by stubtest. But happy to clean them up too, if people have a preference one way or the other

Why don't we delete them for now? There must be thousands of other typeshed falsehoods that aren't enumerated in the allowlists. I feel like we'll be able to find and add back the inline explanatory comments fairly easily if we ever do find a good halfway-house solution in stubtest.

PR #7384, opened 2022-02-25
stdlib: Add missing conditionals for Unix-only API

not-my-profile:
I'd prefer these commits to be merged without squashing to produce a more readable git log.

PR #7443, opened 2022-03-05
stdlib: Add several missing `@abstractmethod` decorators

AlexWaygood:
The following methods are all abstract at runtime, but not in the stub:

- `typing.ContextManager.__exit__` (and therefore also `contextlib.AbstractContextManager.__exit__`)
- `typing.AsyncContextManager.__aexit__` (and therefore also `contextlib.AbstractAsyncContextManager.__aexit__`)
- `numbers.Complex.__eq__`
- `numbers.Complex.__abs__`
- `numbers.Complex.__conjugate__`
- `os.PathLike.__fspath__`

This means that inheriting from any of these classes at runtime without overriding these methods would cause an exception to be raised. Type checkers, however, would not have flagged the possibility of an exception being raised, believing these methods to be concrete.


stdlib/multiprocessing/synchronize.pyi
```diff
@@ -28,8 +29,11 @@ class Condition(AbstractContextManager[bool]):
     def wait_for(self, predicate: Callable[[], bool], timeout: float | None = ...) -> bool: ...
     def acquire(self, block: bool = ..., timeout: float | None = ...) -> bool: ...
     def release(self) -> None: ...
+    def __exit__(
+        self, exc_type: type[BaseException] | None, exc_val: BaseException | None, exc_tb: TracebackType | None
+    ) -> bool | None: ...
 
-class Event(AbstractContextManager[bool]):
```
AlexWaygood, line 32: This class doesn't seem to be a context manager at runtime, and doesn't seem to have ever been a context manager at runtime.

stdlib/contextlib.pyi
```diff
@@ -137,6 +137,9 @@ elif sys.version_info >= (3, 7):
         func: Callable[..., AsyncGenerator[_T_co, Any]]
         args: tuple[Any, ...]
         kwds: dict[str, Any]
+        async def __aexit__(
+            self, typ: type[BaseException] | None, value: BaseException | None, traceback: TracebackType | None
+        ) -> bool | None: ...
```
JelleZijlstra, line 142: It actually always returns bool (on main at least), but I don't think that really matters. If we change this we have to change it above on line 131 too.

stdlib/mmap.pyi
```diff
@@ -70,6 +71,9 @@ class mmap(AbstractContextManager[mmap], Iterable[int], Sized):
     # Doesn't actually exist, but the object is actually iterable because it has __getitem__ and
     # __len__, so we claim that there is also an __iter__ to help type checkers.
     def __iter__(self) -> Iterator[int]: ...
+    def __exit__(
+        self, exc_type: type[BaseException] | None, exc_val: BaseException | None, exc_tb: TracebackType | None
```
JelleZijlstra, line 75:
```suggestion
        self, *args: object
```
It just ignores its arguments.

stdlib/mmap.pyi
```diff
@@ -70,6 +71,9 @@ class mmap(AbstractContextManager[mmap], Iterable[int], Sized):
...
+    ) -> bool | None: ...
```
JelleZijlstra, line 76:
```suggestion
    ) -> None: ...
```
This can matter because it means the CM will never swallow exceptions.

stdlib/multiprocessing/synchronize.pyi
```diff
@@ -28,8 +29,11 @@ class Condition(AbstractContextManager[bool]):
     def wait_for(self, predicate: Callable[[], bool], timeout: float | None = ...) -> bool: ...
     def acquire(self, block: bool = ..., timeout: float | None = ...) -> bool: ...
     def release(self) -> None: ...
+    def __exit__(
+        self, exc_type: type[BaseException] | None, exc_val: BaseException | None, exc_tb: TracebackType | None
+    ) -> bool | None: ...
```
JelleZijlstra, line 34:
```suggestion
    def __exit__(
        self, __exc_type: type[BaseException] | None, __exc_val: BaseException | None, __exc_tb: TracebackType | None
    ) -> None: ...
```
It takes *args at runtime, but those ultimately get forwarded to `_multiprocessing.SemLock.__exit__` (defined in semaphore.c), which does take exactly three arguments.

It always returns None unless you pass in a very exotic kind of lock.

stdlib/multiprocessing/synchronize.pyi
```diff
@@ -49,3 +53,6 @@ class Semaphore(SemLock):
 class SemLock(AbstractContextManager[bool]):
     def acquire(self, block: bool = ..., timeout: float | None = ...) -> bool: ...
     def release(self) -> None: ...
+    def __exit__(
+        self, exc_type: type[BaseException] | None, exc_val: BaseException | None, exc_tb: TracebackType | None
```
JelleZijlstra, line 57:
```suggestion
        self, __exc_type: type[BaseException] | None, __exc_val: BaseException | None, __exc_tb: TracebackType | None
```
As above

stdlib/multiprocessing/synchronize.pyi
```diff
@@ -49,3 +53,6 @@ class Semaphore(SemLock):
...
+    ) -> bool | None: ...
```
JelleZijlstra, line 58:
```suggestion
    ) -> None: ...
```

stdlib/os/__init__.pyi
```diff
@@ -722,6 +725,9 @@ def rmdir(path: StrOrBytesPath, *, dir_fd: int | None = ...) -> None: ...
 
 class _ScandirIterator(Iterator[DirEntry[AnyStr]], AbstractContextManager[_ScandirIterator[AnyStr]]):
     def __next__(self) -> DirEntry[AnyStr]: ...
+    def __exit__(
+        self, exc_type: type[BaseException] | None, exc_val: BaseException | None, exc_tb: TracebackType | None
+    ) -> bool | None: ...
```
JelleZijlstra, line 730:
```suggestion
    def __exit__(
        self, *args: object
    ) -> None: ...
```
This is `ScandirIterator_exit` in posixmodule.c. It ignores its args and always returns None.

JelleZijlstra (changes requested):

AlexWaygood:
> 
> ```diff
> pandera (https://github.com/pandera-dev/pandera)
> - pandera/checks.py:74: error: Argument 1 to "ChainMap" has incompatible type "MappingProxyType[str, Any]"; expected "MutableMapping[str, Callable[..., Any]]"  [arg-type]
> - pandera/model.py:362: error: Need type annotation for "attrs" (hint: "attrs: Dict[<type>, <type>] = ...")  [var-annotated]
> 
> pydantic (https://github.com/samuelcolvin/pydantic)
> - pydantic/class_validators.py:332: error: Need type annotation for "all_attributes"  [var-annotated]
> - pydantic/class_validators.py:332: error: Argument 1 to "ChainMap" has incompatible type "*List[MappingProxyType[str, Any]]"; expected "MutableMapping[<nothing>, <nothing>]"  [arg-type]
```

These are weird -- looks like this crossed the streams with #7423 being merged, maybe?

JelleZijlstra:
The boostedblob errors are unfortunate. They're in places like https://github.com/hauntsaninja/boostedblob/blob/eb6e95b41ba55c4c595cbe754d05a48f069ce42e/boostedblob/path.py#L328, where `request.execute` is an async CM. The new annotations make mypy catch on to the fact that async CMs can eat exceptions.

And yeah, not sure what caused the ChainMap changes.

JelleZijlstra:
Oh wait, I bet mypy special cases `-> bool` as "may swallow" and assumes that "bool | None" means "won't swallow". So we should probably just change `_AsyncGeneratorContextManager.__exit__` back; sorry for the thrash.

JelleZijlstra:
And now pyright in CI is complaining about shutil, which you didn't change.

JelleZijlstra (approved):

AlexWaygood:
> And now pyright in CI is complaining about shutil, which you didn't change.

Yeah, it's because https://github.com/python/typeshed/pull/7384 made `chown` unavailable on Windows, but it's still declared in `__all__` on Windows. It wasn't picked up by CI running on the PR, because I increased pyright strictness for checking `__all__` after the PR was filed.

JelleZijlstra:
Ah good catch

PR #7446, opened 2022-03-07
Put shutil.chown back on Windows

JelleZijlstra:
See #7384 and https://github.com/python/typeshed/pull/7443#issuecomment-1060071944 (thanks @AlexWaygood for diagnosing).


JelleZijlstra:
Choosing to put it back because the alternative would be to lie about `__all__`, which seems like it's just digging ourselves a deeper hole. The function does exist on Windows at runtime; it just always fails.

AlexWaygood:
> Choosing to put it back because the alternative would be to lie about `__all__`, which seems like it's just digging ourselves a deeper hole. The function does exist on Windows at runtime; it just always fails.

Agree. All options here feel like bad options, but I think this is the least bad option.

AlexWaygood (approved):

PR #7447, opened 2022-03-07
Add some missing things that stubtest found

JelleZijlstra:


stdlib/sys.pyi
```diff
@@ -292,6 +305,10 @@ if sys.version_info < (3, 9):
     def getcheckinterval() -> int: ...  # deprecated
     def setcheckinterval(__n: int) -> None: ...  # deprecated
 
+if sys.version_info < (3, 9):
+    # An 11-tuple or None
+    def callstats() -> tuple[int, int, int, int, int, int, int, int, int, int, int] | None: ...
```
JelleZijlstra, line 310: In 3.7 and 3.8 it actually always returned None. This was never documented as far as I can tell.

srittau: If it always returned `None`, why the tuple return type?

srittau: Because we still support Python 3.6.

srittau (approved):

PR #7448, opened 2022-03-07
Finish renaming stubtest.yml to daily.yml

Akuli:
Related: #7445 

I noticed that the URLs in auto-generated issues (e.g. #7444) were wrong, and `git grep` showed a few other places  too.


hauntsaninja (approved):
Thanks for doing this right!

PR #7453, opened 2022-03-07
rlcompleter: add missing *_matches

henryiii:
These two methods are missing, and seem to be in a todo-like list already.


stdlib/@python2/rlcompleter.pyi
```diff
@@ -1,7 +1,9 @@
-from typing import Any, Union
+from typing import Any, TypeVar, Union
```
AlexWaygood, line 1:
```suggestion
from typing import Any, Union
```

Doesn't look like this is used anywhere (not picked up by CI because only a subset of our CI runs on the Python 2 stubs)

AlexWaygood (approved):
Thanks, looks good! Hope this makes typing pyodide a little less painful :)

PR #7454, opened 2022-03-07
add overload to tuple.__new__ to better express an empty tuple

xavdid:
This PR adds an overload to the `tuple()` constructor to specify that an _empty_ tuple is returned, not `Tuple[Any]`. It covers the following code snippet (using the Pylance type checker):

```py
from typing import Tuple
Point3D = Tuple[int, int, int]

t1: Point3D = 1, 2, 3  # correct no error
t2: Point3D = 1, 2, 3, 4  # correct error; Element size mismatch; expected 3 but received 4
t3: Point3D = () # correct error; Element size mismatch; expected 3 but received 0
t4: Point3D = tuple() # expected error, none thrown
```

The docs specify that:

> If no argument is given, the constructor creates a new empty tuple, `()`.

This is my first typeshed PR, so please feel free to correct any mistakes I've made!

---

For more context, see: https://github.com/microsoft/pylance-release/issues/2445#issuecomment-1059993607


xavdid:
Ah, looks like CI is unhappy:

> `stdlib/builtins.pyi:788: error: Overloaded function signatures 1 and 2 overlap with incompatible return types  [misc]`

I see that most `__new__` functions take `cls: type[Self]` as their first arg, but `Self` isn't subscriptable, so I wasn't sure how to specify that the return was `Tuple[()]` and not the default `Tuple[Any]`. Guidance here is appreciated 😅

stdlib/builtins.pyi
```diff
@@ -784,6 +784,9 @@ class slice:
     def indices(self, __len: SupportsIndex) -> tuple[int, int, int]: ...
 
 class tuple(Sequence[_T_co], Generic[_T_co]):
+    @overload
+    def __new__(cls) -> tuple[()]: ...
+    @overload
     def __new__(cls: type[Self], __iterable: Iterable[_T_co] = ...) -> Self: ...
```
JelleZijlstra, line 790:
```suggestion
    def __new__(cls: type[Self], __iterable: Iterable[_T_co]) -> Self: ...
```
This overload should be used only if `__iterable` is provided, so there should be no default.

xavdid: Ah, because now there's 2 call signatures- one for when `__iterable` is provided, and one without. Makes sense! i'll update.

srittau:
It's unclear to me way mypy primer  explodes, but I think it's related to `isinstance` checks, i.e. mypy thinks that `isinstance(x, tuple)` means that `x` is an empty tuple. Could you try reordering the overloads (and add a comment)?

xavdid:
Sure thing! added in d556d1701714f615134de20d7f5ee8bebf7db884; hope that comment is descriptive enough- I don't quite follow the issue.

xavdid:
Ah, I think I'm following now. Using `tuple` as a class not the result of an instantiation is the issue. 

`tuple` should return `Self`, but `tuple()` should return `Tuple[()]`. I'm not sure how that's expressed in the interfaces correctly.

srittau:
I'm not totally sure what's going on here myself. primer output looks much better, although it's weird that the remaining problems are in mypy itself. Maybe a mypy developer could look over this? Cc @JukkaL @hauntsaninja

hauntsaninja:
I'm not sure either, will look into it soon.

I think the first error on `config_files = tuple(map(os.path.expanduser, defaults.CONFIG_FILES))` is not specific to mypy's codebase. The other two errors involve a custom mypy plugin that mypy uses on the mypy codebase; it's possible there's a bug in that plugin, but I'd need to investigate more to know what's going on.

xavdid:
Thanks for taking a look! Let me know if there's anything else I can do to be helpful.

hauntsaninja:
Here, https://github.com/python/mypy/pull/12319 will fix the errors from the plugin.

I unfortunately don't have a fix for the errors from `config_files = tuple(map(os.path.expanduser, defaults.CONFIG_FILES))`, but that's life.

JelleZijlstra:
I'm not sure this change is worth it if we lose correct typing for subclasses of `tuple`. After `class X(tuple)`, what does `reveal_type(X())` do?

AlexWaygood:
> I'm not sure this change is worth it if we lose correct typing for subclasses of `tuple`. After `class X(tuple)`, what does `reveal_type(X())` do?

Empty tuples and tuple subclasses are both reasonably common, but I can't say I've ever seen an example of an empty tuple subclass instance. And for non-empty tuple subclass instances, we'll still have correct type inference, as that overload still returns `Self`. So I think it will probably be okay?

AlexWaygood:
Seems like this change might also fix https://github.com/python/mypy/issues/6283

AlexWaygood:
> I'm not sure this change is worth it if we lose correct typing for subclasses of `tuple`. After `class X(tuple)`, what does `reveal_type(X())` do?

For both `master` and this PR, if I do `mypy --custom-typeshed-dir` on the following snippet, mypy yields the following result:

```python
# test.py
class X(tuple): ...
reveal_type(X())  # Revealed type is "test.X"
```

So this doesn't seem to cause any regressions in that regard.

JelleZijlstra (approved):

PR #7455, opened 2022-03-07
Update cachetools to 5.0.0

Akuli:


stubs/cachetools/@tests/stubtest_allowlist.txt
```diff
@@ -1,7 +1,22 @@
 cachetools.Cache.get
-cachetools.cache.Cache.get
 
-# it looks like cachetools stubs are really out of date, so none of the modules
-# referenced exist anymore
-cachetools\..*\.__(get|set|del)item__  # has extra argument
-cachetools\..*TTLCache.__repr__
+# stubs omit defaulted arguments that are meant to be optimizations, not provided by user
```
AlexWaygood, line 3: Does it hurt that much to include the weird extra arguments in these signatures? I worry that if we allowlist all these functions, stubtest won't warn us if there are any other problems with any of them

Akuli: And after thinking about this a bit more: Including the weird extra arguments also would mean including a new method definition for each of the lines I whitelisted. That would clutter the stubs and not help users in any way. CI checks should catch problems, not force you to do something weird and unnecessary IMO.

AlexWaygood: Okay, sounds good! Will try to do a full review of the rest of the PR soon.

stubs/cachetools/cachetools/__init__.pyi
```diff
@@ -45,18 +49,35 @@ class RRCache(Cache[_KT, _VT]):
...
 
-class TTLCache(Cache[_KT, _VT]):
+class _TimedCache(Cache[_KT, _VT]):
     def __init__(
-        self, maxsize: float, ttl: float, timer: Callable[[], float] = ..., getsizeof: Callable[[_VT], float] | None = ...
+        self, maxsize: float, timer: Callable[[], float] = ..., getsizeof: Callable[[_VT], float] | None = ...
     ) -> None: ...
     @property
     def currsize(self) -> float: ...
     @property
```
AlexWaygood, line 58:
```suggestion
    class _Timer:
        def __init__(self, timer: Callable[[], float]) -> None: ...
        def __call__(self) -> float: ...
        def __enter__(self) -> float: ...
        def __exit__(self, *exc: object) -> None: ...
    @property
```

And then just return `_Timer` from the `timer()` method below?

AlexWaygood (commented):
LGTM, other than the point I've raised about the TODO comment.

stubs/cachetools/@tests/stubtest_allowlist.txt
AlexWaygood: Side point: I'm amazed a library that's concerned with such micro-optimisations is also using `functools.total_ordering` in the same module.

stubs/cachetools/cachetools/__init__.pyi
Akuli: Let's see what CI says about nested classes.

AlexWaygood (approved):

PR #7458, opened 2022-03-08
Widen the `cmdclass` parameter of `setuptools.setup`

BvB93:
Alternative to https://github.com/python/typeshed/pull/7418.

As pointed out in abovementioned PR, classes that one might pass to the `cmdclass` parameters (such as the setuptools `build_ext` class) are not guaranteed to inherit from `setuptools.Command`, resulting in type check errors when one attempts to do so. This PR fixes the relevant false positives by relaxing value-type from `setuptools.Command` to the `distutils.core.Command` superclass.

``` python
from setuptools import setup, Extension
from setuptools.command.build_ext import build_ext

class build_test_ext(build_ext):
    def build_extension(self, ext: Extension) -> None: ...

setup(
    # error: dict entry 0 has incompatible type "str": "Type[build_test_ext]"; expected "str": "Type[Command]"
    cmdclass={"build_ext": build_test_ext},
)
```

### DIFF
--- a/stubs/setuptools/setuptools/__init__.pyi
+++ b/stubs/setuptools/setuptools/__init__.pyi
@@ -43,7 +43,7 @@ def setup(
     license: str = ...,
     keywords: list[str] | str = ...,
     platforms: list[str] | str = ...,
-    cmdclass: Mapping[str, type[Command]] = ...,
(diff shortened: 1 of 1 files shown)


Akuli (approved):
This makes the example code type-check, and widening types seems like a much better idea than relying on an inconsistently added base class.

PR #7461, opened 2022-03-08
Improve `mmap.mmap.__enter__`

AlexWaygood:
It returns `Self` at runtime, not `mmap.mmap`, so inheriting from `AbstractContextManager` doesn't really do anything for us.


JelleZijlstra (approved):
A bit surprised that they let you subclass mmap

Issue #7258, opened 2022-02-18
Temporary moratorium on literal constants (until the end of 2022)


JukkaL:
I've encountered cases where the changes from `int` to a literal type in stubs causes regressions for mypy users. Here's one (simplified) example:

```py
import logging

class C:
    def __init__(self) -> None:
        self.level = logging.INFO

    def enable_debug(self) -> None:
        # Error: expression has type "Literal[10]", variable has type "Literal[20]"
        self.level = logging.DEBUG
```

Here's another example:

```py
import logging

class Base:
    level = logging.INFO

class Derived(Base):
    level = logging.DEBUG  # Error -- incompatible with base class
```

Previously the types were `int`, so no errors were generated. The code in the above examples doesn't look buggy to me, and the behavior of mypy also seems as expected.

There could well be other problematic cases like this. Having more precise types is generally useful, but in the case of literal types, the regressions such as the above may make them not worth the potential false positives (which are hard to predict, since in most use cases literal types are fine). I wonder if there are concrete examples where the literal types bring concrete benefits that could offset the false positives.

There were several false positives like the above in a big internal codebase when using a recent typeshed. Working around the issues is simple by annotating the attributes as `int`, but this could be quite non-obvious for non-expert users who aren't familiar with literal types (a somewhat advanced typing feature).

Thoughts? I'm leaning towards preferring to only use literal types when there is a concrete use cases where they help, since the fallout from using them seems difficult to predict.

srittau:
Actually, at least in mypy the following already works:

```python
FOO: Final = 1
BAR: Final = 2

x = FOO

reveal_type(FOO)  # type: Literal[1]
reveal_type(x)  # type: int
```

So maybe we should allow `X: Final = 1` in stubs?

srittau:
Cc @rchen152 @erictraut @pradeep90 for feedback on using `X: Final = 3` in stubs.

AlexWaygood:
> So maybe we should allow `X: Final = 1` in stubs?

This to me feels much more expressive than writing `X: Literal[1]`, so I'd be in favour of allowing this in stubs whatever the case. I worry, however, that there's not much point in annotating a variable as `Literal` or `Final` if it's automatically inferred as something less precise as soon as it's assigned to a different variable.

JukkaL:
> I worry, however, that there's not much point in annotating a variable as Literal or Final if it's automatically inferred as something less precise as soon as it's assigned to a different variable.

[...]

PR #7463, opened 2022-03-08
Use Literal for statistics.quantiles()

### LINKED ISSUE #7462: Use Literal for statistics.quantiles()
Replace `method: str` with `method: Literal["inclusive", "exclusive"]`.

See:  https://github.com/python/typeshed/blob/6c967d4202bb3bca716ded414f3c4ebd56f4cd21/stdlib/statistics.pyi#L105

Akuli:
Fixes #7462


JelleZijlstra (approved):
Looks good. I don't think #7258 needs to block this.

hauntsaninja:
"Prefer false negatives over false positives" would imply we shouldn't do this without a fallback overload right? Esp given that there are false positives in practice^

AlexWaygood (approved):
I think this is a good change to make. The one false positive is very easily fixed. https://github.com/python/typeshed/issues/7258 is a temporary moratorium on literal constants, not on literal argument types.

AlexWaygood:
Shoot, I screwed up the commit description a bit there. Apologies :)

Akuli:
It happens :) It's not super critical with stdlib stubs anyway. With third-party stubs, this matters more because commit messages turn into an auto-generated changelog.

PR #7464, opened 2022-03-08
Improve `collections.Counter` stub

AlexWaygood:
~~Two~~Three changes:

1. ~~Remove~~Change the bad overload which the comment states is bad, and use a `# type: ignore` instead.
2. Improve `Counter` arithmetic dunders. `Counter` is halfway between a `set` and a `dict`, and the arithmetic dunders come from the `set`-like side of its personality. I think it makes sense to harmonise `Counter`'s `__sub__`, `__and__`, `__or__` and `__add__` methods with the equivalent ones on `typing.AbstractSet`.
3. Use `Counter[Any]` instead of `Counter[object]` in various `Counter` dunders, due to `dict` invariance.


stdlib/collections/__init__.pyi
```diff
@@ -266,15 +266,14 @@ class Counter(dict[_T, int], Generic[_T]):
...
-    # it's included so that the signature is compatible with
-    # Dict.update. Not sure if we should use '# type: ignore' instead
-    # and omit the type from the union.
-    @overload
+    # The second overload is deliberately different to dict.update()
+    # (if it were `Iterable[_T] | Iterable[tuple[_T, int]]`,
+    # the tuples would be added as keys, breaking type safety),
+    # hence the type: ignore[override]
+    @overload  # type: ignore[override]
     def update(self, __m: SupportsKeysAndGetItem[_T, int], **kwargs: int) -> None: ...
```
Akuli, line 274: `SupportsKeysAndGetItem` doesn't seem correct. The implementation does `if isinstance(iterable, _collections_abc.Mapping)` and then calls `.items()`.

I initially added `SupportsKeysAndGetItem` to `dict.__init__` to get a tkinter thing to work, and apparently it has now found its way here somehow :)

stdlib/collections/__init__.pyi
```diff
@@ -283,22 +282,23 @@ class Counter(dict[_T, int], Generic[_T]):
         def __eq__(self, other: object) -> bool: ...
         def __ne__(self, other: object) -> bool: ...
 
-    def __add__(self, other: Counter[_T]) -> Counter[_T]: ...
-    def __sub__(self, other: Counter[_T]) -> Counter[_T]: ...
-    def __and__(self, other: Counter[_T]) -> Counter[_T]: ...
-    def __or__(self, other: Counter[_T]) -> Counter[_T]: ...  # type: ignore[override]
+    def __add__(self, other: Counter[_S]) -> Counter[_T | _S]: ...
+    def __sub__(self, other: Counter[Any]) -> Counter[_T]: ...
+    def __and__(self, other: Counter[Any]) -> Counter[_T]: ...
```
Akuli, line 287: I don't think we should change to `Any`, unless users complain. With `Any`, you won't get an error for `counter_of_ints - counter_of_strings`, which is bad IMO. With `_T`, you will get an error for `counter_of_strings - counter_of_strings_or_nones`, which apparently doesn't come up in practice because nobody has complained about it.

stdlib/collections/__init__.pyi
AlexWaygood: Oops, I think this one was me. I'll change it back to `Mapping`.

stdlib/collections/__init__.pyi
AlexWaygood: The same argument surely applies for `Abstract Set.__sub__`:
https://github.com/python/typeshed/blob/c987c78077dc21ef2b1fdb11b83eeb947c3b4276/stdlib/typing.pyi#L841

Are we saying that in an ideal world, that also wouldn't be `AbstractSet[Any]`, but it would probably be too disruptive to change it now?

Akuli: Changing `AbstractSet` would be more disruptive, but the argument doesn't apply to `AbstractSet` as much as it does to `Counter`s:
- It's common to have sets that contain `None`s and subtract between them, which I think is the most common use for subtracting sets of different types. I use sets containing `None`s all the time.
- A counter that contains `None`s would be something like `{None: 2, "foo": 1, "bar": 6}`, which seems like it could useful in some case but not very often. I don't remember ever using a counter like this, for example.

AlexWaygood: I'm still sorta unconvinced. I'll revert for now, and maybe open an issue to discuss in more depth later :)

Akuli (approved):

PR #7465, opened 2022-03-09
ssl: mark ssl{object,socket}_class as Instance-Var

### LINKED ISSUE #7459: `SSLContext` in ssl-library: remove `ClassVar` from `sslsocket_class` and `sslobject_class`
According to https://docs.python.org/3/library/ssl.html#ssl.SSLContext.sslsocket_class and https://docs.python.org/3/library/ssl.html#ssl.SSLContext.sslobject_class, both variables (`sslsocket_class` and `sslobject_class`) should be changeable on instance, so `ClassVar` should be removed from both files in https://github.com/python/typeshed/blob/master/stdlib/ssl.pyi#L396

Thanks to @AlexWaygood for pointing this out in https://github.com/python/mypy/issues/12308

prauscher:
See #7459 and https://docs.python.org/3/library/ssl.html#ssl.SSLContext.sslsocket_class

This change will mark `SSLContext.sslobject_class` and `SSLContext.sslsocket_class` as Instance-Variable, which is correct according to its documentation.

Fixes #7459

### DIFF
--- a/stdlib/ssl.pyi
+++ b/stdlib/ssl.pyi
@@ -2,7 +2,7 @@ import enum
 import socket
 import sys
 from _typeshed import ReadableBuffer, Self, StrOrBytesPath, WriteableBuffer
-from typing import Any, Callable, ClassVar, Iterable, NamedTuple, Optional, Union, overload
+from typing import Any, Callable, Iterable, NamedTuple, Optional, Union, overload
 from typing_extensions import Literal, TypedDict, final
 
 _PCTRTT = tuple[tuple[str, str], ...]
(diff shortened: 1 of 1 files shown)


stdlib/ssl.pyi
```diff
@@ -393,8 +393,8 @@ class SSLContext:
         maximum_version: TLSVersion
         minimum_version: TLSVersion
         sni_callback: Callable[[SSLObject, str, SSLContext], None | int] | None
-        sslobject_class: ClassVar[type[SSLObject]]
-        sslsocket_class: ClassVar[type[SSLSocket]]
+        sslobject_class: type[SSLObject]
```
AlexWaygood, line 396:
```suggestion
        # The following two attributes have class-level defaults.
        # However, the docs explicitly state that it's OK to override these attributes on instances,
        # so making these ClassVars wouldn't be appropriate
        sslobject_class: type[SSLObject]
```

AlexWaygood (approved):
Thanks!

Issue #7367, opened 2022-02-23
Dropping Python 2 support


srittau:
As far as I'm aware, mypy is the last user of our Python 2 stubs. mypy plans to eventually drop Python 2 support, although no fixed date has been set. See python/mypy#12237 for details. Once mypy has dropped support, we can also consider dropping it. This would ease maintenance and we could simplify our release process quite a bit.

I'm opening and pinning this ticket for discussion and objections.

srittau:
And as a side note: Python 2 support in third-party packages is being slowly phased out anyway. We dropped all Python 2-only packages that are not required by other packages (with only enum34 and ipaddress left) and we stopped supporting Python 2 in many packages where upstream stopped support Python 2.

srittau:
The first step was to drop most Python-2-only third party packages. As a second step I suggest to drop the remaining Python-2-only packages and remove Python 2 support of third-party packages that have separate Python 2 and Python 3 versions. This affects:

* `six`
* `ipaddress`
* `enum34`

This also means dropping Python 2 support for the following packages as they depend on the packages above:

* `cryptography` (which is replaced by inline stubs anyway)
* `pyvmomi`
* `boto`
* `paramiko`
* `pysftp`

Python 2 support always remains in the current version as published on PyPI, but I think not offering any updates is a reasonable compromise with lowering our maintenance effort.

Akuli:
I think Python 2 support for `six` stubs should be dropped last, as `six` users typically care about Python 2.

srittau:
We had one non-maintenance commit since the re-org more than a year ago, but two maintenance commits. types-six 1.16.10 is not going away, but the maintenance burden remains. I believe that six is nowadays often used, just because removing it from Python 3 code is effort.

hauntsaninja:
Would it make sense to remove six altogether? Seems unlikely that six's API would change much and it feels a little weird and potentially confusing for users to have `types-six` not support Python 2.

srittau:
I have no strong feelings about six. Last release was less than a year ago, though.

srittau:
The next step would be to drop Python 2 support from the remaining third-party stubs. This affects:

* atomicwrites
* backports.ssl_match_hostname
* certifi
* characteristic
* chardet
* click-spinner
* colorama
* croniter
* Deprecated
* docopt
* emoji
* entrypoints
* first
* mypy-extensions
* polib
* pyaudio
* python-dateutil
* python-gflags
* python-nmap
* pytz
* retry
[...]

PR #7466, opened 2022-03-09
Remove Python 2 support from some third-party distributions

srittau:
* Remove all Python 2 only third-party distributions.
* Remove Python 2 support from six as the Python 2 stubs diverged from
  the Python 3 stubs.
* Recursively remove Python 2 support from all distributions that depend
  on the distributions above.

Affects:

* boto
* cryptography
* enum34 (removed)
* ipaddress (removed)
* paramiko
* pysftp
* pyvmomi
* six

Part of #7367


tests/check_new_syntax.py
```diff
@@ -46,11 +46,13 @@ def visit_AsyncFunctionDef(self, node: ast.AsyncFunctionDef) -> None:
 
     class ObjectClassdefFinder(ast.NodeVisitor):
         def visit_ClassDef(self, node: ast.ClassDef) -> None:
-            if any(isinstance(base, ast.Name) and base.id == "object" for base in node.bases):
-                errors.append(
-                    f"{path}:{node.lineno}: Do not inherit from `object` explicitly, "
-                    f"as all classes implicitly inherit from `object` in Python 3"
-                )
+            # Temporarily disable this check for two ex-Python 2 stubs.
+            if "paramiko" not in str(path) and "cryptography" not in str(path):
```
srittau, line 50: I'll send a separate PR once this is merged to fix those problems and re-enable this check.

tests/check_consistent.py
```diff
@@ -68,17 +68,9 @@ def check_stubs():
                 else:
                     assert name.isidentifier(), f"Bad file name '{entry}' in stubs"
             else:
-                if entry in ("@python2", "@tests"):
+                if entry in ("@tests",):
```
AlexWaygood, line 71: Any reason not to do this?

```suggestion
                if entry == "@tests":
```

AlexWaygood (approved):

Akuli (approved):

Akuli:
These stubs are now Python 3 only, and can potentially be cleaned up:
- boto
- cryptography
- paramiko
- pysftp
- pyvmomi
- six

srittau:
`six` already had separate Python 3 stubs, so I think it's already more or less clean.

PR #7470, opened 2022-03-10
change TemporaryFileWrapper.__exit__ return type to None

nbdaaron:


JelleZijlstra:
For context we found this from quora/pyanalyze#496, which makes pyanalyze stricter about context managers that may swallow exceptions. `tempfile._TemporaryFileWrapper.__exit__` returns whatever its underlying file returns (https://github.com/python/cpython/blob/3.6/Lib/tempfile.py#L502), and file objects don't tend to return anything from `__exit__`.

The new pip error is at https://github.com/pypa/pip/blob/bc45b93eb679963d23f100be4ebb5d0f1568ceee/src/pip/_internal/network/lazy_wheel.py#L138. I think pip will just need to change there.

JelleZijlstra (approved):

PR #7471, opened 2022-03-10
change Change IO/IOBase.__exit__ return type to None

nbdaaron:
similar to #7470, this was caught by quora/pyanalyze#496.

I don't believe `__exit__` in `IOBase` returns anything: https://github.com/python/cpython/blob/main/Lib/_pyio.py#L513 and I think `typing.IO` just represents I/O streams returned by `open()` which I don't think return anything from `__exit__`


srittau:
While I think it's okay to change the return type for `IOBase` if the subclasses all return `None`, I'm not sure we should do the same for `IO` as it is a semi-protocol and sub-classes could decide to return `True`.

JelleZijlstra:
I feel like the contract for `IO.__exit__` is that it closes the file. A subclass that also suppresses exceptions would be unexpected enough that it's good for a type checker to warn about it.

PR #7472, opened 2022-03-10
Add xmltodict

andrewshadura:
Add stubs for xmltodict: https://pypi.org/project/xmltodict/


AlexWaygood (changes requested):
Thanks! In addition to the linting errors picked up by CI, please import `OrderedDict` from `collections` rather than `typing` (`from __future__ import annotations` is the default behavior in a stub file)

andrewshadura:
@AlexWaygood, thanks for the comments, done!

pyrightconfig.stricter.json
```diff
@@ -71,7 +71,8 @@
         "stubs/stripe",
         "stubs/ttkthemes",
         "stubs/urllib3",
-        "stubs/vobject"
+        "stubs/vobject",
+        "stubs/xmltodict"
```
srittau, line 75: Is this necessary? As far as I can see, you fully typed the stubs, and if I remember correctly, incompletely typed stubs is what the stricter config will warn about.

andrewshadura: No idea, this was added by the script automatically. Should I remove it?

andrewshadura: Oh, it’s an exclusion list. Removed.

stubs/xmltodict/xmltodict.pyi
```diff
@@ -0,0 +1,34 @@
...
+    expat: Any = ...,
+    process_namespaces: bool = ...,
+    namespace_separator: str = ...,
+    disable_entities: bool = ...,
+    **kwargs: Any,
+) -> OrderedDict[str, Any]: ...
+@overload
+def unparse(
+    input_dict: Mapping[str, Any],
+    output: IO[AnyStr] | StreamWriter | StreamReaderWriter | SupportsWrite[str],
```
JelleZijlstra, line 20: This is a strange combination of types. The `AnyStr` doesn't make much sense because there's no other `AnyStr` in the signature. The codecs classes are compatible with both `IO[str]` and `SupportsWrite[str]`, so they're redundant. And `IO[str]` is itself redundant with `SupportsWrite[str]`, so you probably don't need that either.

In general we're trying to avoid using `IO` for annotations, in favor of protocols like `SupportsWrite`. I haven't looked at the code for xmltodict, but if it uses this parameter just for `.write()`, then `SupportsWrite[str]` by itself is the right annotation.

andrewshadura: I copied this from SAX and changed the first two parts of the union type to `IO[AnyStr]` because originally it was either binary or text stream.

andrewshadura: Here’s the original annotation:
```python
class XMLGenerator(handler.ContentHandler):
    def __init__(
        self,
        out: TextIOBase | RawIOBase | StreamWriter | StreamReaderWriter | SupportsWrite[str] | None = ...,
        encoding: str = ...,
        short_empty_elements: bool = ...,
    ) -> None: ...
```

andrewshadura: @JelleZijlstra, I’m not aware about the distinction between `IO` and `*IOBase` from the point of view of type annotations; in my code I usually use `IO[…]` with either `bytes` or `str` depending on the context.

Is there any documentation I can check where these conventions are described?

JelleZijlstra: https://github.com/python/typing/discussions/829 has some good discussion. I think the SAX annotation is also wrong.

stubs/xmltodict/xmltodict.pyi
```diff
@@ -0,0 +1,34 @@
+from _typeshed import SupportsRead, SupportsWrite
+from codecs import StreamReaderWriter, StreamWriter
+from collections import OrderedDict
+from typing import IO, Any, AnyStr, Mapping, overload
+
+class ParsingInterrupted(Exception): ...
+
+def parse(
+    xml_input: AnyStr | SupportsRead[bytes],
```
AlexWaygood, line 9: The docstring states:

> `xml_input` can be either a `string`, a file-like object.

But the source code also looks like it works fine with `bytes` inputs.

So based on all that, I would think the annotation should be something along the lines of:

```suggestion
    xml_input: str | bytes | SupportsRead[bytes],
```

(`AnyStr` is not appropriate here because there is no type dependency in this function.)

stubs/xmltodict/xmltodict.pyi
AlexWaygood: I agree with @JelleZijlstra that the SAX annotation looks wrong. (Sorry about that -- it was a very reasonable assumption to think that you could copy this annotation from that annotation!!)

andrewshadura: So what should I choose? Just `StreamWriter | SupportsWrite[str] | SupportsWrite[bytes]`? It supports writing to text and binary streams, and *also* to objects that have `.write()`.

AlexWaygood (approved):
Thanks! LGTM

stubs/xmltodict/xmltodict.pyi
```diff
@@ -0,0 +1,33 @@
+from _typeshed import SupportsRead, SupportsWrite
+from collections import OrderedDict
+from typing import Any, Mapping, overload
```
srittau, line 3: Nit for future reference: We prefer to import `Mapping` (and other ABCs) from `collections.abc` now rather than `typing`.

srittau (approved):
Thanks!

PR #7105, opened 2022-02-02
Use `async def` instead of `def ... -> Awaitable` in `typing`

sobolevn:
Source: https://github.com/python/cpython/blob/187930f74c44e460ba09c60ba5d9bb4fac543d8f/Lib/_collections_abc.py#L206-L240

Why is that important? Because `async def` is a `Coroutine`, which is much more than `Awaitable`. For example, this backfires in https://github.com/python/typeshed/pull/6779

PR #7475, opened 2022-03-11
AsyncIterator.__anext__ can be any Awaitable object

graingert:
https://peps.python.org/pep-0525/#support-for-asynchronous-iteration-protocol


graingert:
I think this also applies to AsyncGenerator

graingert:
Yep they're all specified as returning Awaitable:

https://peps.python.org/pep-0525/#asynchronous-generator-object

I think this PR needs reverting https://github.com/python/typeshed/pull/7105

AlexWaygood:
> I think this PR needs reverting #7105

Yes, I think that's probably right. stubtest is complaining because these are all `async def` functions at runtime (but wouldn't be in the stub with this PR merged), but the return type for abstract interfaces should be broader. I've filed https://github.com/python/mypy/pull/12343 to relax that stubtest check.

graingert:
It would be nice if "real" `async def agenfn(...): ... yield` had a stricter type which did have Coroutine `asend/aclose`

AlexWaygood:
> It would be nice if "real" `async def agenfn(...): ... yield` had a stricter type which did have Coroutine `asend/aclose`

Not sure I understand, sorry

graingert:
> > It would be nice if "real" `async def agenfn(...): ... yield` had a stricter type which did have Coroutine `asend/aclose`
> 
> Not sure I understand, sorry

Oh this already exists as `types.AsyncGeneratorType`

PR #7477, opened 2022-03-12
Parsimonious stubs

jpy-git:
[Parsimonious](https://github.com/erikrose/parsimonious) is a PEG parser package.

PyPI link: https://pypi.org/project/parsimonious/

I'm making use of this package in a couple projects and there's also an open [issue](https://github.com/erikrose/parsimonious/issues/183) about adding type hints so figured I'd pick this up.


stubs/parsimonious/parsimonious/exceptions.pyi
```diff
@@ -0,0 +1,24 @@
+from parsimonious.expressions import Expression as Expression
```
JelleZijlstra, line 1: Most of these (outside `__init__.pyi`) probably shouldn't be using `as`; that indicates that the file is re-exporting the name.

stubs/parsimonious/parsimonious/__init__.pyi
```diff
@@ -0,0 +1,8 @@
+from parsimonious.exceptions import (
```
JelleZijlstra, line 1: I'm also seeing `VisitationError` at runtime

stubs/parsimonious/parsimonious/expressions.pyi
```diff
@@ -0,0 +1,68 @@
...
+
+MARKER: Any
+
+def is_callable(value: Any) -> bool: ...
+
+_CALLABLE_RETURN_TYPE = int | tuple[int, list[Node]] | Node | None
+_CALLABLE_TYPE = (
+    Callable[[str, int], _CALLABLE_RETURN_TYPE]
+    | Callable[[str, int, dict[tuple[int, int], Node], ParseError, Grammar], _CALLABLE_RETURN_TYPE]
+    | Callable[[Any, str, int], _CALLABLE_RETURN_TYPE]
```
JelleZijlstra, line 17: At runtime `expression()` raises an error if it doesn't take either 2 or 5 arguments, so do we need the 3-arg and 6-arg variants?

stubs/parsimonious/parsimonious/expressions.pyi
```diff
@@ -0,0 +1,68 @@
...
+        multiline: bool = ...,
+        dot_all: bool = ...,
+        unicode: bool = ...,
+        verbose: bool = ...,
+        ascii: bool = ...,
+    ) -> None: ...
+
+class Compound(Expression):
+    members: typing.Sequence[Expression]
+    def __init__(self, *members: typing.Sequence[Expression], **kwargs: Any) -> None: ...
```
JelleZijlstra, line 57:
```suggestion
    def __init__(self, *members: Expression, **kwargs: Any) -> None: ...
```
The `*args` annotation is for the type of each arg, not the whole of args.

stubs/parsimonious/parsimonious/grammar.pyi
```diff
@@ -0,0 +1,69 @@
...
+    Regex as Regex,
+    Sequence as Sequence,
+    TokenMatcher as TokenMatcher,
+    ZeroOrMore as ZeroOrMore,
+    expression as expression,
+    is_callable as is_callable,
+)
+from parsimonious.nodes import Node as Node, NodeVisitor as NodeVisitor
+from parsimonious.utils import evaluate_string as evaluate_string
+from six import text_type
```
JelleZijlstra, line 24: Can just use `str`, we don't need to care about Python 2

stubs/parsimonious/parsimonious/grammar.pyi
```diff
@@ -0,0 +1,69 @@
...
+class TokenGrammar(Grammar): ...
+class BootstrappingGrammar(Grammar): ...
+
+rule_syntax: str
+
+class LazyReference(text_type):
+    name: str
+
+class RuleVisitor(NodeVisitor):
+    quantifier_classes: dict[str, Expression]
```
JelleZijlstra, line 42:
```suggestion
    quantifier_classes: dict[str, type[Expression]]
```

stubs/parsimonious/parsimonious/grammar.pyi
```diff
@@ -0,0 +1,69 @@
...
+    visit_expression: Callable[[Self, Node, list[Any]], Any]
+    visit_term: Callable[[Self, Node, list[Any]], Any]
+    visit_atom: Callable[[Self, Node, list[Any]], Any]
+    custom_rules: dict[str, Expression]
```
JelleZijlstra, line 46: Also maybe should have `type[]`?

stubs/parsimonious/parsimonious/nodes.pyi
```diff
@@ -0,0 +1,40 @@
...
+    def __iter__(self) -> Iterator[Node]: ...
+    @property
+    def text(self) -> str: ...
+    def prettily(self, error: Node | None = ...) -> str: ...
+
+class RegexNode(Node):
+    match: Match[str]
+
+class RuleDecoratorMeta(type):
+    def __new__(metaclass: type[Self], name: str, bases: tuple[type, ...], namespace: dict[str, Any]) -> Self: ...
```
jpy-git, line 27: Pyright is complaining about this line:
`error: __new__ override should take a "cls" parameter (reportSelfClsParameterName)`

The error makes sense, but this is how it's written in the source code. `metaclass` gets passed to the `cls` parameter of `type.__new__`.

I tried adding this method to the allowlist but doesn't seem to have suppressed the error?

stubs/parsimonious/parsimonious/nodes.pyi
```diff
@@ -0,0 +1,40 @@
...
+from parsimonious.expressions import Expression as Expression
+from parsimonious.grammar import Grammar as Grammar
+
+class Node:
+    expr: Expression
+    full_text: str
+    start: int
+    end: int
+    children: list[Node]
+    def __init__(self, expr: Expression, full_text: str, start: int, end: int, children: list[Node] | None = ...) -> None: ...
```
JelleZijlstra, line 15: Does it have to be a list? Would be preferable to write `Sequence` so we don't run into list invariance.

stubs/parsimonious/parsimonious/nodes.pyi
```diff
@@ -0,0 +1,40 @@
+from re import Match
+from typing import Any, Callable, Iterator, NoReturn, TypeVar
+from typing_extensions import Self
```
JelleZijlstra, line 3: We can't use PEP 673 yet because mypy doesn't support it

stubs/parsimonious/parsimonious/nodes.pyi
JelleZijlstra, line 27: I think we can omit this method, since the signature is the same as the base class.

stubs/parsimonious/parsimonious/nodes.pyi
```diff
@@ -0,0 +1,40 @@
...
+    def text(self) -> str: ...
+    def prettily(self, error: Node | None = ...) -> str: ...
+
+class RegexNode(Node):
+    match: Match[str]
+
+class RuleDecoratorMeta(type):
+    def __new__(metaclass: type[Self], name: str, bases: tuple[type, ...], namespace: dict[str, Any]) -> Self: ...
+
+class NodeVisitor(type):
```
JelleZijlstra, line 29:
```suggestion
class NodeVisitor(metaclass=RuleDecoratorMeta):
```

stubs/parsimonious/parsimonious/nodes.pyi
```diff
@@ -0,0 +1,40 @@
...
+    grammar: Grammar | Any
+    unwrapped_exceptions: tuple[Exception]
```
JelleZijlstra, line 31:
```suggestion
    unwrapped_exceptions: tuple[type[Exception], ...]
```
`tuple[T]` means a one-tuple. And I suspect this holds exception classes, not instances.

stubs/parsimonious/parsimonious/nodes.pyi
```diff
@@ -0,0 +1,40 @@
...
+    def visit(self, node: Node) -> Any: ...
+    def generic_visit(self, node: Node, visited_children: list[Any]) -> NoReturn: ...
+    def parse(self, text: str, pos: int = ...) -> Node: ...
+    def match(self, text: str, pos: int = ...) -> Node: ...
+    def lift_child(self, node: Node, children: list[Any]) -> Any: ...
+
+_T = TypeVar("_T")
+
+def rule(rule_string: str) -> Callable[[Callable[..., _T]], Callable[..., _T]]: ...
```
JelleZijlstra, line 40:
```suggestion
_CallableT = TypeVar("_CallableT", bound=Callable[..., Any])

def rule(rule_string: str) -> Callable[[_CallableT], _CallableT]: ...
```

stubs/parsimonious/parsimonious/nodes.pyi
jpy-git: You're correct thanks!

stubs/parsimonious/parsimonious/nodes.pyi
jpy-git: fine by me thanks 👍

JelleZijlstra: That allowlist is for stubtest, a different tool than pyright. To shut up pyright here you'd have to use `# type: ignore`.

stubs/parsimonious/parsimonious/nodes.pyi
jpy-git: good shout, I've removed the function it was used in anyway 👍

stubs/parsimonious/parsimonious/expressions.pyi
jpy-git: ahh yep my bad!

stubs/parsimonious/parsimonious/nodes.pyi
jpy-git: Perfect, good to know!

stubs/parsimonious/parsimonious/nodes.pyi
jpy-git: Makes sense, will update the various visit methods similarly as well

stubs/parsimonious/parsimonious/grammar.pyi
jpy-git: So this is used to updated the rule_map which contains instantiated Expressions so think this is correct as is?

https://github.com/erikrose/parsimonious/blob/3da7e804c07d4e495873be208701b5c955247c58/parsimonious/grammar.py#L438-L447

stubs/parsimonious/parsimonious/expressions.pyi
jpy-git: So these account for this case: https://github.com/erikrose/parsimonious/blob/3da7e804c07d4e495873be208701b5c955247c58/parsimonious/expressions.py#L72-L74

Essentially it's meant to cover when methods are supplied as the custom callable e.g. `callable=some_instance.some_method`. The code just assumes the first argument is `self` in that case and removes it and then checks the length is 2 or 5.

I put `Any` to represent this `self` param since it could be for any custom class. lmk if there's a better way to type it though 😄

jpy-git:
@JelleZijlstra Thanks for the thorough review 😄 

I believe I've addressed the comments. I'm having some trouble with getting [Check third party stubs with stubtest](https://github.com/python/typeshed/runs/5525076052?check_suite_focus=true) to pass since it's complaining about some of the subclass types not agreeing with the superclass types, which is what's happening in the code, and stubtest_allowlist doesn't seem to fix things. Any ideas what I'm missing?

hauntsaninja:
stubtest runs a mypy build first to understand what's in the stubs. The mypy build is failing, so we haven't even gotten to the actual stubtest logic, which is what stubtest allowlists affect.

Assuming your subclass types are actually correct, adding type ignore will allow the build to proceed. For the "Type application has too many types", it's probably some mypy bug, try using an old style typing.Union + typing.Tuple

JelleZijlstra:
The errors are from mypy itself, not stubtest; they'll need `# type: ignore`. I'll try to figure out why the regular mypy test didn't catch this.

JelleZijlstra:
@hauntsaninja it's not expected that the mypy test in CI is passing though, right? Looks like it runs only on the stdlib.

stubs/parsimonious/parsimonious/expressions.pyi
jpy-git: Actually apologies, just checked that code for the latest release of the package and these method arguments weren't available at that point so will remove them now: https://github.com/erikrose/parsimonious/blob/f4496cc0b6bae44c09eaaf4233e94a8d70e396a9/parsimonious/expressions.py#L60-L67

hauntsaninja:
Yeah, looks like it's only running against the stdlib. It's not the first time I've seen stubtest third party catch mypy errors that mypy hasn't caught, so the test has probably been buggy for a while.

JelleZijlstra:
It's likely related to the mypy feature where it does `os._exit` to exit faster.

JelleZijlstra:
Fix in #7478

PR #7481, opened 2022-03-13
dict.pop: Remove default for second argument

JelleZijlstra:
The first overload takes care of the case where there is only one argument, so there should be no default in the second overload.


hauntsaninja (approved):
Sounds good. This is mainly aesthetic / doesn't cause any issue as was right?

srittau (approved):

srittau:
> Sounds good. This is mainly aesthetic / doesn't cause any issue as was right?

I believe so, although in theory it could make a difference when return types from multiple overloads get merged.

JelleZijlstra:
I got here from https://github.com/jab/bidict/pull/242#discussion_r825315765 because I thought the extra default was confusing mypy's incompatible override check. Seems like that was something else, but this change is still a little more hygienic.

PR #7478, opened 2022-03-13
Fix mypy test

JelleZijlstra:
Because we didn't set mypy's clean_exit parameter, it was
exiting immediately after checking the stdlib.

This exposed two issues:
- types-requests has a version-limited dependency on types-urllib3, which we crashed on
- mypy crashes checking the SQLAlchemy stubs. Not sure if that's a known issue.

PR #7484, opened 2022-03-13
Further improve `mypy_test.py`

AlexWaygood:
- Run mypy on the 3.11 stdlib in CI, as a regression test for https://github.com/python/mypy/issues/12220
- Correct the docstring at the top of the file following the changes made in https://github.com/python/typeshed/pull/7478
- Stop the test from crashing when run locally if there's an extra file/folder in the `stubs` directory. (I had a few `.mypy_cache` folders and extra codemodding scripts in various places, which meant that the test was crashing. The `dist_path / "METADATA.toml"` call in the `is_supported` function was evaluating to a nonexistent filepath.)

### DIFF
--- a/tests/mypy_test.py
+++ b/tests/mypy_test.py
@@ -274,9 +274,9 @@ def test_third_party_distribution(distribution: str, major: int, minor: int, arg
     return code, len(files)
 
 
-def is_probably_stubs_folder(dist_path: Path) -> bool:
+def is_probably_stubs_folder(distribution: str, distribution_path: Path) -> bool:
     """Validate that `dist_path` is a folder containing stubs"""
-    return (dist_path / "METADATA.toml").exists()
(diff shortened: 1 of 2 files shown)


tests/mypy_test.py
```diff
@@ -327,7 +328,12 @@ def main():
             if distribution == "SQLAlchemy":
                 continue  # Crashes
 
-            if not is_supported(distribution, major):
+            distribution_path = Path("stubs", distribution)
+
+            if not is_probably_stubs_folder(distribution_path):
```
JelleZijlstra, line 333: How will we catch it if somebody forgets to put a METADATA.toml in a new stubs folder?

AlexWaygood: Any better after 83b05f3?

JelleZijlstra (approved):

PR #7485, opened 2022-03-13
Make `__hash__` a `ClassVar` for several classes where it is set to `None`

AlexWaygood:
If we write `__hash__: None` in a class statement, mypy infers that `__hash__` is an instance attribute, allowing nonsensical code like this, which fails at runtime:

```python
{1, 2}.__hash__ = None
```

I'm making changes to the Python 2 stubs as well as the py3 stubs in this PR, because I'd like to use flake8-pyi to write a regression test for this.


JelleZijlstra:
Would this still make mypy allow `dict.__hash__ = None`? That fails at runtime too.

AlexWaygood:
> Would this still make mypy allow `dict.__hash__ = None`? That fails at runtime too.

It would, but there's no easy solution to that. We can't use `Final`, since it's fine to override `__hash__` on a subclass, and we can't use the `@property` trick, because that makes it an instance attribute rather than a class attribute. There's not really any way to express "immutable class attribute which is fine to override in a subclass".

This PR at least improves the situation somewhat.

JelleZijlstra:
> it's fine to override `__hash__` on a subclass

Won't any override of `__hash__: ClassVar[None]` already need a `# type: ignore`?

AlexWaygood:
> > it's fine to override `__hash__` on a subclass
> 
> Won't any override of `__hash__: ClassVar[None]` already need a `# type: ignore`?

True. But users will get a [pretty confusing error message](https://mypy-play.net/?mypy=latest&python=3.10&gist=a913ed05d3d0ebb98205fa2c2f2d12e6) if we use `Final`.

JelleZijlstra:
I see, that does seem worse.

PR #7487, opened 2022-03-14
Minor stub fix for jsonschema RefResolver

### LINKED ISSUE #7398: jsonschema RefResolver `referrer` should be marked as `dict[str, Any]`, not `str`
I came across this today:

https://github.com/python/typeshed/blob/b8421eb5d4c2560031233bb15491effaafbbac16/stubs/jsonschema/jsonschema/validators.pyi#L54

The type written for `jsonschema.RefResolver.__init__` is not right for the `referrer` parameter.

(I'll try to remember to put in a PR to fix this, but putting this issue in before I forget.)

sirosen:
`RefResolver.referrer` is a `dict[str, Any]`

resolves #7398

### DIFF
--- a/stubs/jsonschema/jsonschema/validators.pyi
+++ b/stubs/jsonschema/jsonschema/validators.pyi
@@ -44,14 +44,14 @@ class Draft202012Validator(_Validator): ...
 _Handler = Callable[[str], Any]
 
 class RefResolver:
-    referrer: str
+    referrer: dict[str, Any]
     cache_remote: Any
     handlers: dict[str, _Handler]
     store: URIDict
     def __init__(
         self,
         base_uri: str,
-        referrer: str,
+        referrer: dict[str, Any],
         store: SupportsKeysAndGetItem[str, str] | Iterable[tuple[str, str]] = ...,
         cache_remote: bool = ...,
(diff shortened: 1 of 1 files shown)


srittau:
Wouldn't it make more sense to use the `_Schema` type var and potentially fix that to be a dict (if a schema always has to be a dict)?

sirosen:
I'm not really clear on what `_Schema` is or is doing in this stub -- it's just an alias to `Any`? I'm not clear on why it exists.

`referrer` is "the referring document", so not necessarily a schema. Maybe something like `_JsonDict = dict[str, Any]` and use that instead for both?

Since `referrer` _technically_ can be any valid JSON datatype, I could mark it `Any` or `_JsonDict | list[Any]` if that seems better.
The non-dict, non-list cases seem very niche, unlikely to occur (intentionally) in practice, and better to leave out of the annotation.

srittau (approved):

sirosen:
Thanks for the speedy review!
Why don't I open a follow up with proposed changes for `_Schema = dict[str, Any]`? I don't want to lose context on this since I'm unlikely to look at the stub again until/unless I hit a type error.

PR #7488, opened 2022-03-14
sqlalchemy: Session.__enter__ & AsyncSession.__aenter__

### LINKED ISSUE #7440: AsyncSession sqlalchemy no return type annotated
https://github.com/python/typeshed/blob/cda3d9cf722f38f5646e99cacbc169620e268786/stubs/SQLAlchemy/sqlalchemy/ext/asyncio/session.pyi#L53

Because return type of `__aenter__` not anotated, pylance cant give a suggestion on `session` variable.
`session` variable not have `AsyncSession` type hint in vscode. it show `Any` type hint instead.
`session` variable not show any suggestion with this code structure.

```python
async def getParams(product_id: str, db_session: AsyncSession) -> dict:
    async with db_session as session:
        session.execute()
        
```

jpy-git:
closes #7440.

As mentioned in the issue this method (and the non-async equivalent) should return `Self` so that type hints can be used within a context manager.


AlexWaygood (approved):
LGTM - thanks!

PR #7489, opened 2022-03-14
jsonschema: mark schemas as Mapping[str, Any]

sirosen:
follow-up to #7487

This adjustment is meant to make the annotations for schema objects more restrictive in accordance with the behavior and documentation of the jsonschema library.

The library documents methods accepting schemas as taking any mapping type, so preserve that behavior here with `Mapping[str, Any]`.

Additionally, annotate a few locations which consume schema objects but which had other annotations like `Any` or `object`.

The only exceptional case is `validator_for` which takes a mapping or bool, even though the argument is named `schema`. This is not only a runtime behavior, but documented as part of the usage, so it belongs appropriately in the stub.


srittau (approved):
Thanks!

PR #7490, opened 2022-03-14
Use stubtest 0.940

srittau:


Eppydean7:
Unsurcribe

hauntsaninja (commented):
Thanks!

Is it intentional that we're dropping the tomli grep? Also it looks like mypy's released 0.941, so we could update to that (I don't expect CI status to change)

PR #6779, opened 2022-01-02
Change `create_task` type

sobolevn:
I've manually tested that `Awaitable`s are not supported:

```python
import asyncio

class My:
    def __await__(self):
        async def inner():
            print('here')
        return inner

async def main():
  await asyncio.create_task(My())

asyncio.run(main())
```

Output:

```
Traceback (most recent call last):
  File "/Users/sobolev/Desktop/cpython/build/ex.py", line 15, in <module>
    asyncio.run(main())
  File "/Users/sobolev/.pyenv/versions/3.9.9/lib/python3.9/asyncio/runners.py", line 44, in run
    return loop.run_until_complete(main)
  File "/Users/sobolev/.pyenv/versions/3.9.9/lib/python3.9/asyncio/base_events.py", line 642, in run_until_complete
    return future.result()
  File "/Users/sobolev/Desktop/cpython/build/ex.py", line 13, in main
    await asyncio.create_task(My())
  File "/Users/sobolev/.pyenv/versions/3.9.9/lib/python3.9/asyncio/tasks.py", line 361, in create_task
    task = loop.create_task(coro)
  File "/Users/sobolev/.pyenv/versions/3.9.9/lib/python3.9/asyncio/base_events.py", line 433, in create_task
    task = tasks.Task(coro, loop=self, name=name)
TypeError: a coroutine was expected, got <__main__.My object at 0x10c38fe20>
```

But, `Generator` objects are also supported:

```python
import asyncio
def gen():
   i = yield
   print(1, i)  # prints: `1 None`

async def main():
  await asyncio.create_task(gen())

asyncio.run(main())
```

Closes #6770

PR #7491, opened 2022-03-14
Improve various signatures that shouldn't be `async def`, but currently are

AlexWaygood:
As discussed in #7475, the following methods are all "async def" functions at runtime, but should _not_ be "async def" functions in the stub:
- `AsyncIterator.__anext__`
- `AyncGenerator.__anext__`
- `AsyncGenerator.aclose`
- `AsyncGenerator.asend`
- `AsyncGenerator.athrow` 

`typing.AsyncIterator` and `typing.AsyncGenerator` represent abstract interfaces rather than concrete classes, and [PEP 525](https://peps.python.org/pep-0525/#support-for-asynchronous-iteration-protocol), which introduced the asynchronous iteration protocol, explicitly states that it is fine for these methods to return any awaitable. The current annotations, which state that these methods have to return _coroutines_, are therefore too restrictive, and cause false positives.

In addition to these changes:

- `AsyncGenerator.__anext__` is not abstract at runtime, and I don't see any reason why it should be in the stub either.
- `AsyncGenerator.__aiter__` is also not abstract at runtime, and not even defined on `AsyncGenerator` at runtime; it is simply inherited from `AsyncIterator`. I think we should do the same in the stub.
- `AsyncGenerator.aclose` is not abstract at runtime, and I don't see any reason why it should be in the stub.
- `_typeshed.SupportsAnext.__anext__` and `contextlib._SupportsAclose.aclose` were both erroneously made `async def` functions in #7107; this PR also reverts the changes _that_ PR made to _those_ classes.

This PR reverts the changes #7105 made to `typing.pyi`, but not the changes #7105 made to `types.pyi`. I think that makes sense, as the classes in `types` are concrete classes, whereas the classes in `typing` are abstract interfaces.

This PR adds several entries to the allowlist, but we can take them off again if https://github.com/python/mypy/pull/12343 is merged.


AlexWaygood:
The `boostedblob` error is related to this line [here](https://github.com/hauntsaninja/boostedblob/blob/47c1e0efaf3c15b35ace14b3c29694149555ae9f/boostedblob/boost.py#L547), and is because `asyncio.create_task` is annotated as requiring a _coroutine_ following #6779, and this PR would change the definition of `AsyncIterator.__anext__` such that it wouldn't necessarily be guaranteed to return a coroutine anymore.

AlexWaygood:
Cc. @sobolevn, since this partially reverts your PR 🙂

graingert:
Would SupportsAnext need fixing? https://github.com/python/typeshed/blob/master/stdlib/_typeshed/__init__.pyi#L36

AlexWaygood:
> Would SupportsAnext need fixing? https://github.com/python/typeshed/blob/master/stdlib/_typeshed/__init__.pyi#L36

Good spot. `contextlib._SupportsAclose` as well.

graingert:
anext is actually slightly odd as it returns whatever `__anext__` returns - eg an Awaitable or even not an awaitable!

```
Python 3.10.2 (main, Jan 15 2022, 18:02:07) [GCC 9.3.0] on linux
>>> class Foo:
...     def __anext__(self):
...         return "hello"
... 
>>> anext(Foo())
'hello'
>>> 
```

graingert:
so the type annotation for `builtins.anext` should look more like this:

```python
class SupportsAnyAnext(Protocol[_T_co]):
    def __anext__(self) -> _T_co: ...

class SupportsAwaitableAnext(Protcol[_T_co]):
    def __anext__(self) -> Awaitable[_T_co]: ...

@overload
def anext(__i: SupportsAnyAnext[_T]) -> _T: ...
@overload
async def anext(__i: SupportsAwaitableAnext[_T], default: _VT) -> _T | _VT: ...
```

so that anext doesn't incorrectly promote an `AsyncIterable.__next__` `Awaitable` into a `Coroutine`

AlexWaygood:
> anext is actually slightly odd as it returns whatever `__anext__` returns - eg an Awaitable or even not an awaitable!

Sure, but I'd imagine that if `__anext__` returns something that isn't awaitable, it's almost certainly indicative of a programming error. So surely we'd be doing the developer a favour by raising an error if they try to call `anext` on an object that has an `__anext__` method that doesn't return an awaitable?

graingert:
my point was more about preserving the return type of the `.__anext__` call: 

```python
_Awaitable_T_co = TypeVar("_Awaitable_T_co", bound=Awaitable)  # bound isn't strictly true - anext supports any return value

class SupportsAnext(Protocol[_Awaitable_T_co]):
    def __anext__(self) -> _Awaitable_T_co: ...

class SupportsAwaitableAnext(Protcol[_T_co]):
    def __anext__(self) -> Awaitable[_T_co]: ...

@overload
def anext(__i: SupportsAnyAnext[_T]) -> _T: ...
@overload
async def anext(__i: SupportsAwaitableAnext[_T], default: _VT) -> _T | _VT: ...
```

AlexWaygood:
> my point was more about preserving the return type of the `.__anext__` call:

It's frying my brain somewhat, but I think I understand. Thanks for the explanation!

AlexWaygood:
Better now @graingert?

stdlib/builtins.pyi
```diff
@@ -1068,10 +1071,17 @@ class _PathLike(Protocol[_AnyStr_co]):
 
 if sys.version_info >= (3, 10):
     def aiter(__async_iterable: _SupportsAiter[_SupportsAnextT]) -> _SupportsAnextT: ...
+
+    class _SupportsSynchronousAnext(Protocol[_AwaitableT_co]):
+        def __anext__(self) -> _AwaitableT_co: ...
+
+    class _SupportsAwaitableAnext(Protocol[_T_co]):
```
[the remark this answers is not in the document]
AlexWaygood: Any suggestions for a good comment wording?

stdlib/builtins.pyi
```diff
@@ -1068,8 +1071,12 @@ class _PathLike(Protocol[_AnyStr_co]):
 
 if sys.version_info >= (3, 10):
     def aiter(__async_iterable: _SupportsAiter[_SupportsAnextT]) -> _SupportsAnextT: ...
+
+    class _SupportsSynchronousAnext(Protocol[_AwaitableT_co]):
+        def __anext__(self) -> _AwaitableT_co: ...
+
     @overload
-    async def anext(__i: SupportsAnext[_T]) -> _T: ...
+    def anext(__i: _SupportsSynchronousAnext[_AwaitableT]) -> _AwaitableT: ...
```
graingert, line 1079: here's a close python implementation of anext: https://github.com/python/cpython/blob/ea786a882b9ed4261eafabad6011bc7ef3b5bf94/Lib/test/test_asyncgen.py#L52-L80

```suggestion
    # anext is not, in fact, an async function. When default is not provided
    # anext is just a passthrough for obj.__anext__
    def anext(__i: _SupportsSynchronousAnext[_AwaitableT]) -> _AwaitableT: ...
```

maybe it's obvious now and doesn't need a comment :shrug:

graingert: to get an idea of what I'm trying to describe, here's a close python implementation of anext: https://github.com/python/cpython/blob/ea786a882b9ed4261eafabad6011bc7ef3b5bf94/Lib/test/test_asyncgen.py#L52-L80

graingert (approved):

stdlib/builtins.pyi
AlexWaygood: Oh, that's really useful! We should definitely include a link to that in the comment!

AlexWaygood:
Thanks for the review @graingert! (I'll leave this open for a while to give other maintainers a chance to review.)

tests/stubtest_allowlists/py3_common.txt
```diff
@@ -246,6 +246,10 @@ xml.parsers.expat.expat_CAPI
 # ==========
 # Allowlist entries that cannot or should not be fixed
 # ==========
+_collections_abc.AsyncGenerator.asend  # async at runtime, deliberately not in the stub, see #7475. Pos-only differences also.
```
graingert, line 249: you probably want to link to 7491 as that's where most of the commentry is

AlexWaygood:
(Planning to merge this in a day or so, unless any other maintainers want to review, since it's sort of blocking https://github.com/python/mypy/pull/12321)

PR #7492, opened 2022-03-15
Add stubs for whatthepatch

cuspymd:
Hi, I added stubs for [whatthepatch](https://github.com/cscorley/whatthepatch).


stubs/whatthepatch/whatthepatch/exceptions.pyi
```diff
@@ -0,0 +1,16 @@
+from typing import Any
+
+class WhatThePatchException(Exception): ...
+
+class HunkException(WhatThePatchException):
+    hunk: Any
```
Akuli, line 6:
```suggestion
    hunk: int | None
```

stubs/whatthepatch/whatthepatch/patch.pyi
```diff
@@ -0,0 +1,77 @@
...
+    line: int | None
+    hunk: int
+
+file_timestamp_str: str
+diffcmd_header: Any
+unified_header_index: Any
+unified_header_old_line: Any
+unified_header_new_line: Any
+unified_hunk_start: Any
+unified_change: Any
```
Akuli, line 31: These could be annotated as `Pattern[str]`, where `Pattern` is imported from `typing`.

stubs/whatthepatch/whatthepatch/patch.pyi
```diff
@@ -0,0 +1,77 @@
...
+cvs_header_index: Any
+cvs_header_rcs: Any
+cvs_header_timestamp: Any
+cvs_header_timestamp_colon: Any
+old_cvs_diffcmd_header: Any
+
+def parse_patch(text: str) -> Iterator[diffobj]: ...
+def parse_header(text: str) -> header | None: ...
+def parse_scm_header(text: str) -> header | None: ...
+def parse_diff_header(text: str) -> header | None: ...
```
Akuli, line 65: All of these functions seem to accept `str | Iterable[str]` too.

stubs/whatthepatch/whatthepatch/snippets.pyi
```diff
@@ -0,0 +1,7 @@
+from collections.abc import Iterable
+from typing import Any
+
+def remove(path: str) -> None: ...
```
Akuli, line 4: This function actually works with any `StrOrBytesPath` (defined in `_typeshed`). But this doesn't matter much if the rest of the library only works with string paths, not e.g. bytes paths or `pathlib.Path` objects.

cuspymd: Thanks for your detailed code reviews.

Akuli (approved):
Thanks! I fixed the CI failures.

PR #7494, opened 2022-03-16
psycopg2: Accept Composable in place of query string

andersk:
https://www.psycopg.org/docs/sql.html#psycopg2.sql.Composable
“`Composable` objects can be passed directly to `execute()`, `executemany()`, `copy_expert()` in place of the query string.”

Cc @srittau


JelleZijlstra (approved):

hauntsaninja:
Thank you!

PR #7496, opened 2022-03-16
Revert "psycopg2: Accept Composable in place of query string"

JelleZijlstra:
Reverts python/typeshed#7494

pyright started failing on master after this was merged (see #7495). I can't see why, but let's see if reverting fixes it.


JelleZijlstra:
Nope, still fails

PR #7498, opened 2022-03-16
pyright: Don't report incompatible overrides

srittau:
These overrides are inherited from the implementation and are out of
typeshed's control.

### DIFF
--- a/pyrightconfig.json
+++ b/pyrightconfig.json
@@ -31,8 +31,6 @@
     "reportUntypedNamedTuple": "error",
     "reportPrivateUsage": "none",
     "reportConstantRedefinition": "error",
-    "reportIncompatibleMethodOverride": "error",
-    "reportIncompatibleVariableOverride": "error",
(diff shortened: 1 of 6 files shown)


srittau:
See discussion in #7495.

AlexWaygood (approved):
(I assume the CI will pass, if it ever wakes up from its slumber)

JelleZijlstra:
We should be able to do this for mypy too by turning off the `override` error code.

Not totally sure we should do this, though; often incompatible override errors can be ignored, but sometimes they also provide a useful nudge to adjust the annotations on a method that's part of a protocol.

AlexWaygood:
> We should be able to do this for mypy too by turning off the `override` error code.
> 
> Not totally sure we should do this, though; often incompatible override errors can be ignored, but sometimes they also provide a useful nudge to adjust the annotations on a method that's part of a protocol.

Personally I'm happy with keeping the mypy errors, but switching off the slightly stricter pyright errors? (I think that's a coherent position?)

PR #7500, opened 2022-03-17
multiprocessing: args can be any Iterable

JelleZijlstra:
https://github.com/python/cpython/blob/main/Lib/multiprocessing/process.py#L91


AlexWaygood (approved):
The [docs](https://docs.python.org/3/library/multiprocessing.html#the-process-class) state that `Process` has the same API as `threading.Thread`, and the docs for `threading.Thread` only guarantee that non-tuples will be supported in 3.11+. But this follows the runtime implementation, so LGTM!

PR #7501, opened 2022-03-17
Add typing.TypeVarTuple, Unpack, assert_type

JelleZijlstra:
python/cpython#31021 and python/cpython#30843.

### DIFF
--- a/stdlib/typing.pyi
+++ b/stdlib/typing.pyi
@@ -28,7 +28,9 @@ if sys.version_info >= (3, 11):
         "Tuple",
         "Type",
         "TypeVar",
+        "TypeVarTuple",
         "Union",
+        "Unpack",
         "AbstractSet",
         "ByteString",
         "Container",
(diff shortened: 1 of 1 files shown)


stdlib/typing.pyi
```diff
@@ -524,6 +527,12 @@ if sys.version_info >= (3, 8):
 if sys.version_info >= (3, 11):
     Self: _SpecialForm
     Never: _SpecialForm = ...
+    Unpack: _SpecialForm
+
+    class TypeVarTuple:
+        __name__: str
```
JelleZijlstra, line 533: Anticipating python/cpython#31954.

AlexWaygood: Should we wait until that's merged before merging this, or do you reckon we're good to go?

JelleZijlstra: Better to wait I think.

AlexWaygood: (Converting this to a draft for now so it doesn't get accidentally merged in the meantime)

JelleZijlstra: I merged the CPython change

AlexWaygood: And I merged this one :)

PR #7502, opened 2022-03-17
Improve types for invoke tasks.py

kasium:


stubs/invoke/invoke/tasks.pyi
```diff
@@ -1,57 +1,75 @@
...
-        iterable=...,
-        incrementable=...,
+        iterable: Iterable[str] | None = ...,
+        incrementable: Iterable[str] | None = ...,
     ) -> None: ...
     @property
     def name(self): ...
-    def __eq__(self, other): ...
-    def __hash__(self): ...
+    def __eq__(self, other: object) -> bool: ...
```
AlexWaygood, line 45: It seems like this will actually raise `AttributeError` if you try to compare a `Task` with something that doesn't have `name` or `body` attributes. https://github.com/pyinvoke/invoke/blob/45dc9d03639dac5b6d1445831bf270e686ef88b4/invoke/tasks.py#L100

```suggestion
    def __eq__(self, other: Task) -> bool: ...  # type: ignore[override]
```

stubs/invoke/invoke/tasks.pyi
```diff
@@ -1,57 +1,75 @@
...
+    def __hash__(self) -> int: ...
     def __call__(self, *args, **kwargs): ...
     @property
-    def called(self): ...
+    def called(self) -> int: ...
```
AlexWaygood, line 49:
```suggestion
    def called(self) -> bool: ...
```

It returns a bool https://github.com/pyinvoke/invoke/blob/45dc9d03639dac5b6d1445831bf270e686ef88b4/invoke/tasks.py#L133

stubs/invoke/invoke/tasks.pyi
```diff
@@ -1,57 +1,75 @@
 from _typeshed import Self
-from typing import Any
+from typing import Any, Callable, Iterable, TypeVar
```
AlexWaygood, line 2:
```suggestion
from collections.abc import Callable, Iterable
from typing import Any, TypeVar
```

We prefer imports from `collections.abc` where possible

AlexWaygood (changes requested):
Thanks! Mostly looks good, but a few points below

stubs/invoke/invoke/tasks.pyi
```diff
@@ -1,57 +1,76 @@
...
     def argspec(self, body): ...
-    def fill_implicit_positionals(self, positional): ...
-    def arg_opts(self, name, default, taken_names): ...
-    def get_arguments(self): ...
+    def fill_implicit_positionals(self, positional: Iterable[str] | None) -> Iterable[str]: ...
+    def arg_opts(self, name: str, default: Any, taken_names: Iterable[str]) -> dict[str, Any]: ...
+    def get_arguments(self) -> list[Argument]: ...
 
-def task(*args, **kwargs) -> Task: ...
+_TASK = TypeVar("_TASK", bound=Task)
```
srittau, line 56: `_TaskT` would fit our naming conventions better. Also, this should be moved to the top of the file.

srittau (changes requested):

AlexWaygood (approved):
LGTM

srittau (approved):

PR #7509, opened 2022-03-17
Add `typing._SubscriptableSpecialForm` and use it to fix various special forms

KotlinIsland:
DRAFT

> Strictly speaking, these have `__getitem__` methods, only that these raise a `TypeError` when called. I don't have good solution for this
> \- srittau

We could lie and make a fake subclass of special form that has `NoReturn` on `__getitem__`😳


AlexWaygood:
Thanks, but I agree with @JelleZijlstra that we need to be very careful with changes to `typing`. Personally, I'm open to #7508, as it seems to be solving an actual problem, and is a much smaller change. But this change is quite sweeping, and would require adding entries to the stubtest allowlists in order to propagate a lie in typeshed. I'm also not sure how much of a real problem it is that mypy thinks `NoReturn` is subscriptable. (I suppose you could make more of an argument for `TypeAlias` or `Self`, but it still doesn't seem nearly as much of a problem as the `typing.Union` problem fixed in #7508.)

KotlinIsland:
Fair enough 😀.

btw, why on earth is `Self` not subscriptable?

```py
class Foo(Generic[T]):
    def foo(self) -> Self[int]: ...
```

AlexWaygood:
> Fair enough 😀.
> 
> btw, why on earth is `Self` not subscriptable?
> 
> ```python
> class Foo(Generic[T]):
>     def foo(self) -> Self[int]: ...
> ```
> 
>  > Note that we reject using `Self` with type arguments, such as `Self[int]`. This is because it creates ambiguity about the type of the self parameter and introduces unnecessary complexity
> 
> pfff

Ah, the dream of [higher-kinded types](https://github.com/python/typing/issues/548#issuecomment-1068234931)

PR #7510, opened 2022-03-17
pytz: utcoffset only returns None if dt is None

JelleZijlstra:
https://github.com/stub42/pytz/blob/master/src/pytz/tzinfo.py#L396


JelleZijlstra:
OK I need to do something else here to get the override checks to be happy.

AlexWaygood:
Perhaps we need to make the stdlib classes generic first, as per https://github.com/python/mypy/issues/10067.

JelleZijlstra:
I don't think so, I believe it's a mypy bug that it complains about this.

AlexWaygood (approved):

PR #7515, opened 2022-03-19
PEP 604: Remove some more uses of Union/Optional

JelleZijlstra:
Mypy seems fine with these. Aliases that use type[] or variadic tuple[]
still break mypy.

### DIFF
--- a/stdlib/xmlrpc/client.pyi
+++ b/stdlib/xmlrpc/client.pyi
@@ -13,9 +13,11 @@ class _SupportsTimeTuple(Protocol):
     def timetuple(self) -> time.struct_time: ...
 
 _DateTimeComparable = DateTime | datetime | str | _SupportsTimeTuple
-_Marshallable = Union[None, bool, int, float, str, bytes, tuple[Any, ...], list[Any], dict[Any, Any], datetime, DateTime, Binary]
-_XMLDate = Union[int, datetime, tuple[int, ...], time.struct_time]
-_HostType = Union[tuple[str, dict[str, str]], str]
+_Marshallable = (
+    bool | int | float | str | bytes | None | tuple[Any, ...] | list[Any] | dict[Any, Any] | datetime | DateTime | Binary
+)
+_XMLDate = int | datetime | tuple[int, ...] | time.struct_time
+_HostType = Union[str, tuple[str, dict[str, str]]]
 
 def escape(s: str) -> str: ...  # undocumented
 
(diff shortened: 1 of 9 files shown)


AlexWaygood:
Some of these mypy is cool with on >= 3.10, but can't handle on <3.10

JelleZijlstra:
Right. Looks like these are the patterns:
- `type[]` at top level fails
- `tuple[T1, T2]` at top level fails (but `tuple[T1, ...]` is fine)
- `T1 | Callable[..., T2 | T3]` fails, but only <=3.9

Still ended up with a few fixes. I didn't look at `@python2`.

stdlib/xmlrpc/client.pyi
```diff
@@ -13,9 +13,11 @@ class _SupportsTimeTuple(Protocol):
...
 
 _DateTimeComparable = DateTime | datetime | str | _SupportsTimeTuple
-_Marshallable = Union[None, bool, int, float, str, bytes, tuple[Any, ...], list[Any], dict[Any, Any], datetime, DateTime, Binary]
-_XMLDate = Union[int, datetime, tuple[int, ...], time.struct_time]
-_HostType = Union[tuple[str, dict[str, str]], str]
+_Marshallable = (
+    bool | int | float | str | bytes | None | tuple[Any, ...] | list[Any] | dict[Any, Any] | datetime | DateTime | Binary
+)
+_XMLDate = int | datetime | tuple[int, ...] | time.struct_time
+_HostType = Union[str, tuple[str, dict[str, str]]]
```
AlexWaygood, line 20: Why has this one changed?

JelleZijlstra: I guess I moved the order around to see if it made a difference, then reverted back to Union wrong
```suggestion
_HostType = Union[tuple[str, dict[str, str]], str]
```

AlexWaygood:
> Right. Looks like these are the patterns:
> 
> * `type[]` at top level fails
> * `tuple[T1, T2]` at top level fails (but `tuple[T1, ...]` is fine)
> * `T1 | Callable[..., T2 | T3]` fails, but only <=3.9
> 
> Still ended up with a few fixes. I didn't look at `@python2`.

Of these, as far as I'm aware, only (2) is tracked over at mypy (in https://github.com/python/mypy/issues/11098); (1) and (3) need to have bug reports filed.

AlexWaygood (approved):
LGTM

JelleZijlstra:
I created https://github.com/python/mypy/labels/topic-pep-604 so we can track these better. Going to spend some time today organizing the mypy issue tracker.

AlexWaygood:
> I created https://github.com/python/mypy/labels/topic-pep-604 so we can track these better. Going to spend some time today organizing the mypy issue tracker.

https://github.com/python/mypy/issues/11887 could do with that label being added 👍

AlexWaygood:
Mypy bugs for (1) and (3) have been filled:

- https://github.com/python/mypy/issues/12392
- https://github.com/python/mypy/issues/12393

JelleZijlstra:
Thanks!

PR #7517, opened 2022-03-20
Fixed type of the "host" parameter in asyncio getaddrinfo()

agronholm:
The types should match `socket.getaddrinfo()`.

### DIFF
--- a/stdlib/asyncio/events.pyi
+++ b/stdlib/asyncio/events.pyi
@@ -211,7 +211,14 @@ class AbstractEventLoop:
     # Network I/O methods returning Futures.
     @abstractmethod
     async def getaddrinfo(
-        self, host: str | None, port: str | int | None, *, family: int = ..., type: int = ..., proto: int = ..., flags: int = ...
+        self,
+        host: bytes | str | None,
+        port: str | int | None,
+        *,
+        family: int = ...,
+        type: int = ...,
+        proto: int = ...,
+        flags: int = ...,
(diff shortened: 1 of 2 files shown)


stdlib/asyncio/events.pyi
```diff
@@ -211,7 +211,14 @@ class AbstractEventLoop:
     # Network I/O methods returning Futures.
     @abstractmethod
     async def getaddrinfo(
-        self, host: str | None, port: str | int | None, *, family: int = ..., type: int = ..., proto: int = ..., flags: int = ...
+        self,
+        host: bytes | str | None,
```
AlexWaygood, line 215: Any reason not to include `bytearray` in the union as well, like we do in `socket.getaddrinfo()`?

```suggestion
        host: bytearray | bytes | str | None,
```

srittau: `bytearray` is [redundant with `bytes` in argument types](https://docs.python.org/3/library/typing.html#typing.ByteString) and shouldn't be included.

AlexWaygood: > `bytearray` is [redundant with `bytes` in argument types](https://docs.python.org/3/library/typing.html#typing.ByteString) and shouldn't be included.

Okay — no action to be taken here then; we should just change `socket.getaddrinfo`

AlexWaygood (approved):

agronholm:
Thanks!

PR #4484, opened 2020-08-26
fix assigning to tkinter.Tk().report_callback_exception

Akuli:
Fixes #1767 (again)

PR #7520, opened 2022-03-21
Update `property.fget`, `property.fset`, and `property.fdel`

### LINKED ISSUE #7516: Wrong type for `fget` of property
```py
class A:
    @property
    def foo(self):
        return 'a'

A.foo.fget(A())
```
pyright reports error `error: Expected 0 positional arguments (reportGeneralTypeIssues)`

jpy-git:
Closes #7516.

```python
class A:
    def __init__(self, foo: int) -> None:
        self._foo = foo

    @property
    def foo(self) -> int:
        return self._foo
    
    @foo.setter
    def foo(self, foo: int) -> None:
        self._foo = foo
    
    @foo.deleter
    def foo(self) -> None:
        del self._foo

a = A(1)
print(A.foo.fget(a))

A.foo.fset(a, 2)
print(a.foo)

A.foo.fdel(a)
```
These property methods take `property` type as first arg and then the desired instance.

### DIFF
--- a/stdlib/builtins.pyi
+++ b/stdlib/builtins.pyi
@@ -1021,9 +1021,9 @@ class range(Sequence[int]):
     def __reversed__(self) -> Iterator[int]: ...
 
 class property:
-    fget: Callable[[Any], Any] | None
-    fset: Callable[[Any, Any], None] | None
-    fdel: Callable[[Any], None] | None
+    fget: Callable[[property, Any], Any] | None
+    fset: Callable[[property, Any, Any], None] | None
+    fdel: Callable[[property, Any], None] | None
     __isabstractmethod__: bool
     def __init__(
         self,


jpy-git:
N.B. mypy complains about these methods since it seems to treat `<class>.foo` as the `foo` method rather than as a property but that's unrelated to this change and probably needs to be fixed in mypy
```
test.py:18: error: overloaded function has no attribute "fget"
test.py:20: error: overloaded function has no attribute "fset"
test.py:23: error: overloaded function has no attribute "fdel"
```

jpy-git:
On seeing that primer diff, maybe typeshed is right and this is a pyright issue?

JelleZijlstra:
Mypy and pyright disagree here. Here's a minimal example:

```
from typing import Callable

class X:
    f: Callable[[], int]
    g: Callable[["X"], int]

reveal_type(X().f())
reveal_type(X().g())
```

Pyright complains about the second call (`Expected 1 more positional argument`), mypy about the first one (`main.py:8: error: Attribute function "f" with type "Callable[[], int]" does not accept self argument`).

I feel like mypy is right here, because the behavior where `self` gets implicitly bound on instances is specific to functions, not all callables. But the type system sometimes conflates functions with callables, so I can see where pyright is coming from.

In either case, the discrepancy is unfortunate because it means we can't write typeshed stubs that satisfy both type checkers. Perhaps we should raise a discussion at the github.com/python/typing repo, come to a consensus, and update one type checker or the other.

Akuli:
I find mypy's behavior weird, because not all callables are descriptors. In #4484, I used this workaround for it:

```
class _Foo(Protocol):
    def __call__(self) -> int: ...

class X:
    f: _Foo
```

I believe this works consistently in mypy and pyright, but I haven't checked.

jpy-git:
So @Akuli's workaround seems to fix @JelleZijlstra's minimal example. However, if I try and make the same changes to `property` I still get errors.

For the code in the linked issue:
```python
class A:
    @property
    def foo(self):
        return 'a'

A.foo.fget(A())
```

mypy: `"Callable[[A], Any]" has no attribute "fget"`
pyright: `Expected 0 positional arguments`

Seems like mypy can't infer that `A.foo` should have `property` type and pyright has a similar issue as discussed above (but doesn't seem to be helped by the workaround)

Will make an issue in the typing repo for discussion and close this PR for now

AlexWaygood:
Let's see what mypy_primer says about this now that mypy 0.981 is released.

AlexWaygood:
Hmm, that still looks pretty bad. And, looking at this issue again, I think we might have misdiagnosed a little bit here.

Following the release of mypy 0.981, mypy and pyright now agree on the snippet @JelleZijlstra posted in https://github.com/python/typeshed/pull/7520#issuecomment-1074049966: https://mypy-play.net/?mypy=latest&python=3.10&gist=276f0399f36c37f6e291f43b277cec65.

And looking at the stubs again, I think maybe typeshed is correct at the moment. Maybe this is just an issue with pyright's special-casing here.

AlexWaygood:
And indeed, I can no longer reproduce the error in #7516 using VSCode. So, it seems like this was a pyright problem, and that it's now fixed 🥳

PR #7525, opened 2022-03-21
mypy_primer: mention if output is truncated

hauntsaninja:
When changes are really disruptive, like #7430, we truncate output.
Mention this, so that it's known that the change is even more disruptive
than it might appear.


JelleZijlstra (approved):
Thanks! This would be helpful on mypy too.

srittau (approved):

Issue #7526, opened 2022-03-22
random.sample() returning list[_T | int] instead of list[_T]


Harry-Lees:
```python
import random

chosen_colours: list[str] = random.sample(["red", "blue"], counts=[4, 2], k=5)
print(chosen_colours)
```

The example from the documentation raises an error as the type returned by random.sample is `list[str | int]`. Currently the type hint is:

```python
def sample(
    self, population: Sequence[_T] | AbstractSet[_T], k: int, *, counts: Iterable[_T] | None = ...
) -> list[_T]: ...
```

I believe changing it to

```python
def sample(
    self, population: Sequence[_T] | AbstractSet[_T], k: int, *, counts: Iterable[int] | None = ...
) -> list[_T]: ...
```

better reflects the type of the function, as `counts` must be an integer. This fixes the example given.

PR #7527, opened 2022-03-22
random.sample take Iterable[int] instead of Iterable[_T]

Harry-Lees:
The random.sample() function takes in an optional counts parameter, this should be set to `Iterable[int]` instead of `Iterable[_T]` as the function raises a `TypeError` if the counts variable is not an integer. This fixes issue #7526


hauntsaninja (approved):
Thanks for catching!

PR #7528, opened 2022-03-22
random.sample() doesn't take a Set in Python3.11.

Harry-Lees:
From the [docs](https://docs.python.org/3.11/library/random.html#random.sample)

"Changed in version 3.11: The population must be a sequence. Automatic conversion of sets to lists is longer supported."


hauntsaninja (approved):
Thank you once more!

PR #7531, opened 2022-03-22
Mark pre/post tasks in invoke tasks as iterable

kasium:
In fact, they can be tuples, lists, etc. so just mark them as iterable


srittau (changes requested):
I'd suggest to update `__init__()` accordingly.

Issue #5613, opened 2021-06-10
Idea: "Incomplete" marker


srittau:
When creating stubs, we encourage contributors to leave items unannotated instead of using `Any`. This often works, but there are cases where you have to fall back on `Any`, for example:

```python
class Foo:
    def bar(self, some_arg, other_arg=None):
        self.a_field = something()
        if some_cond():
            return whatever()
        else:
            return None
```

In this case the best we can do is:

```python
class Foo:
    a_field: Any
    def bar(self, some_arg, other_arg: Any | None = ...) -> Any | None: ...
```

In this case it's not clear if `Any` is used, because there is no better alternative or because this needs better annotations.

Idea: We could introduce an alias for `Any` in `_typeshed`, called something like `Incomplete`, `Unknown` (my current favorite), `Something`, `Untyped`, or `Unannotated`. This could clearly mark that this argument type still needs work:

```python
class Foo:
    some_arg: Unknown
    def bar(self, some_arg, other_arg: Unknown | None = ...) -> Unknown | None: ...
```
This has a few advantages:

* Communicates that the type is not final and will be improved (and narrowed) in the future.
* Communicates that this particular annotation needs work.
* Gives us a better overview of the completion of stubs and prevents "lies". (In fact, an idea I have is to produce graphs that shows the completion percentage of stubs over time.)
* Gives us the ability to grep for `Any` to find areas where the type system could be improved or where newer type system features could be used.

srittau:
Of course, stubgen should be taught this type if we would do something like this.

JelleZijlstra:
I like the idea of being more explicit what is meant by Any. Pyright uses a similar concept internally (https://github.com/microsoft/pyright/blob/7bb059ecbab5c0c446d4dcf5376fc5ce8bd8cd26/docs/type-inference.md#type-inference), so maybe we can help Pyright by similarly separating Any and Unknown in typeshed.

jakebailey:
Selfishly from the pyright POV, `Unknown` has a specific meaning (a version of `Any` with special tracing and associated errors), so I'd be a bit concerned that the two would be confused and harder to distinguish, especially in the editor where we're showing the type info in tooltips.

[...]

PR #7535, opened 2022-03-22
Add _typeshed.Incomplete

srittau:
Closes: #5613


AlexWaygood:
> I have not idea why mypy "Cannot determine type of "Any"". This error also exist when not using `TypeAlias`. This seems incredibly weird to me.

@sobolevn figured out a fix for this mypy bug in https://github.com/python/mypy/pull/12134, but it hasn't been merged yet

AlexWaygood (approved):

PR #7540, opened 2022-03-23
Use conditional overloads to simplify several stdlib functions

AlexWaygood:
This feature is now supported by pyright and, as of version 0.940, mypy (https://github.com/python/mypy/pull/10712). Not sure whether pytype supports it or not, but wanted to see what the CI thought about it.


rchen152:
Yeah, this should work with pytype! (I played around with the example in https://github.com/python/mypy/pull/10712, and pytype's parsing output looks fine.)

AlexWaygood:
> Yeah, this should work with pytype! (I played around with the example in [python/mypy#10712](https://github.com/python/mypy/pull/10712), and pytype's parsing output looks fine.)

Awesome!

srittau (approved):

Issue #7542, opened 2022-03-24
mock.patch.object type annotation fails with **kwargs


graingert:
https://mypy-play.net/?mypy=latest&python=3.10&gist=6f7d6d1036646b8e8a41ed3a5dab3b98&flags=strict

```python
from unittest import mock
import subprocess


def foo() -> None:
    with mock.patch.object(
        subprocess,
        "Popen",
        **{"return_value.communicate.side_effect": KeyboardInterrupt}
    ):
        pass
```

AlexWaygood:
Reopening as per https://github.com/python/typeshed/pull/7543#pullrequestreview-1123222530 -- the original snippet still causes mypy to emit a false-positive error, even using mypy master and typeshed master.

PR #7543, opened 2022-03-24
Make unittest.patch.* methods static

srittau:
Closes: #7542


srittau:
Weird primer output. In the first failure, it seems that the first overload matches:

```
urllib3 (https://github.com/urllib3/urllib3)
+ test/test_queue_monkeypatch.py:25: error: No overload variant of "object" of "_patcher" matches argument types Module, "str", "Type[BadError]"  [call-overload]
+ test/test_queue_monkeypatch.py:25: note: Possible overload variants:
+ test/test_queue_monkeypatch.py:25: note:     def [_T] object(target, attribute: str, new: _T, spec: Optional[Any] = ..., create: bool = ..., spec_set: Optional[Any] = ..., autospec: Optional[Any] = ..., new_callable: Optional[Any] = ..., **kwargs: Any) -> _patch[_T]
+ test/test_queue_monkeypatch.py:25: note:     def object(target, attribute: str, *, spec: Optional[Any] = ..., create: bool = ..., spec_set: Optional[Any] = ..., autospec: Optional[Any] = ..., new_callable: Optional[Any] = ..., **kwargs: Any) -> _patch[Union[MagicMock, AsyncMock]]
```

`_T` is not constrained. Especially when compared to #7542. Anyone got a clue what's going on here?

srittau:
The primer problems seem due to a mypy bug: python/mypy#7781.

JelleZijlstra:
So what should we do here? I don't think we can afford to merge this until the mypy bug is fixed.

srittau:
I agree. I marked this PR as "deferred" for now.

srittau:
Fixed in python/mypy#13482, so should be part of the next mypy release.

AlexWaygood:
> Fixed in [python/mypy#13482](https://github.com/python/mypy/pull/13482), so should be part of the next mypy release.

Sadly, the mypy fix didn't *quite* make it into 0.981, so we'll have to wait a little longer. But it should be in 0.99x.

hauntsaninja:
If this is definitely fixed on mypy master, we should merge.

AlexWaygood:
> If we want to run primer on this with the mypy fix, let me know and I can run it ad hoc.

That would be great!

AlexWaygood (approved):
I tried the original snippet in the issue out using mypy master, and unfortunately it doesn't seem as though this actually fixes the problem in the issue. With this patch applied, using mypy master, here is mypy's output on @graingert's snippet:

```
test.py:6: error: No overload variant of "object" of "_patcher" matches argument types Module, "str", "Dict[str, Type[KeyboardInterrupt]]"  [call-overload]
test.py:6: note: Possible overload variants:
test.py:6: note:     def [_T] object(target: Any, attribute: str, new: _T, spec: Optional[Any] = ..., create: bool = ..., spec_set: Optional[Any] = ..., autospec: Optional[Any] = ..., new_callable: Optional[Any] = ..., **kwargs: Any) -> _patch[_T]
test.py:6: note:     def object(target: Any, attribute: str, *, spec: Optional[Any] = ..., create: bool = ..., spec_set: Optional[Any] = ..., autospec: Optional[Any] = ..., new_callable: Optional[Any] = ..., **kwargs: Any) -> _patch[Union[MagicMock, AsyncMock]]
```

However, this PR does seem like a good thing to do regardless of whether it fixes the issue.

PR #7544, opened 2022-03-24
Update adapters.pyi

KalleDK:
The cert can be
* A string / bytes which is a path to a certfile
* A tuple with two string / bytes, where the first is the certfile and the second is a keyfile
* None (optional)

The tuple could be anything indexable, but there are strict requirement to actualle have two items, don't know if there is something more generic

See https://github.com/psf/requests/blob/main/requests/adapters.py#L242-L248

### DIFF
--- a/stubs/requests/requests/adapters.pyi
+++ b/stubs/requests/requests/adapters.pyi
@@ -1,4 +1,4 @@
-from typing import Any, Container, Mapping, Text
+from typing import Any, Mapping, Text
 
 from urllib3 import exceptions as urllib3_exceptions, poolmanager, response
 from urllib3.util import retry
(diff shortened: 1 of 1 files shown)


KalleDK:
I can rebase to a single commit, if it's wanted

AlexWaygood:
> I can rebase to a single commit, if it's wanted

No need — we'll use github's "squash and merge" feature to merge it, which will combine all commits into a single commit on master :)

KalleDK:
One thing I noticed was the use of Text | bytes and not AnyStr? Any reason or simply typing from before AnyStr was introduced?

AlexWaygood:
> One thing I noticed was the use of Text | bytes and not AnyStr? Any reason or simply typing from before AnyStr was introduced?

`AnyStr` is a `TypeVar`, which is different to `Text | bytes` (a union expression).

Having said that, `Text` shouldn't really be used in the `requests` stubs at all (it should be `str | bytes` rather than `Text | bytes`). The `requests` stubs used to support Python 2, but no longer do, but we haven't got round to changing `Text` to `str` in the `requests` stubs yet.

KalleDK:
Ah okay - then I wont change anymore :D Just got annoyed when I had to make a subclass of the HTTPAdapter and the send args didnøt make sense :)

PR #7545, opened 2022-03-24
Fix primer comment workflow

srittau:


hauntsaninja (commented):
Thanks for fixing! Unfortunately, with the perms, this workflow is hard to test

PR #7546, opened 2022-03-24
allow `asyncio.start_server()` to accept a Sequence of hosts

Harry-Lees:
From the asyncio [documentation](https://docs.python.org/3/library/asyncio-stream.html#asyncio.start_server), the start_server function passes the host argument directly through to the underlying `create_server()` function. Since 3.5.1, the `create_server()` function accepts a `Sequence[str]` for the host, this should be mirrored in the `start_server()` function as-well.


srittau (approved):
Thanks!

PR #7553, opened 2022-03-25
errno: fix typo

JelleZijlstra:
In case you're wondering, they're aliases for each other, at least on my system. See https://github.com/python/cpython/blob/main/Modules/errnomodule.c (I checked main and 3.6 and it's the same).

Noticed this while looking at #7551

```
In [17]: errno.EDEADLOCK
Out[17]: 35

In [18]: errno.EDEADLK
Out[18]: 35
```

### DIFF
--- a/stdlib/errno.pyi
+++ b/stdlib/errno.pyi
@@ -37,7 +37,7 @@ EPIPE: int
 EDOM: int
 ERANGE: int
 EDEADLK: int
-EDEADLCK: int
+EDEADLOCK: int
 ENAMETOOLONG: int
 ENOLCK: int
 ENOSYS: int


JelleZijlstra:
Oh wait, it's already there further down.

Eric-Arellano (approved):
Thanks!

JelleZijlstra:
I found a few more missing errnos with this script:

```
grep -Eo '^E[A-Z]+:' typeshed/stdlib/errno.pyi | grep -Eo '[A-Z]+' | sort > cpython.txt
grep -Eo 'add_errcode."E[A-Z]+' cpython/Modules/errnomodule.c | grep -o 'E.*' | sort | uniq > typeshed.txt
diff cpython.txt typeshed.txt
```

<details>

```
7a8,9
> EAUTH
> EBADARCH
8a11
> EBADEXEC
10a14
> EBADMACHO
12a17
> EBADRPC
24d28
< EDEADLCK
27a32
> EDEVERR
33a39
> EFTYPE
47a54
> EL
53a61
> ELOCKUNMAPPED
61a70
> ENEEDAUTH
66a76
> ENOATTR
80a91
> ENOPOLICY
85a97
> ENOTACTIVE
102a115,118
> EPROCLIM
> EPROCUNAVAIL
> EPROGMISMATCH
> EPROGUNAVAIL
105a122,123
> EPWROFF
> EQFULL
112a131,132
> ERPCMISMATCH
> ESHLIBVERS
```

</details>

I'll add them in this PR.

Eric-Arellano (approved):

AlexWaygood:
In case anybody else is wondering, these weren't caught by stubtest because of this line in the allowlists:

https://github.com/python/typeshed/blob/9a76b13127ffa8365431dcc105fc111cdd267e7e/tests/stubtest_allowlists/py3_common.txt#L375

srittau (approved):
This shows platform differences well. At least the ones I checked doesn't exist on my Debian system.

JelleZijlstra:
> In case anybody else is wondering, these weren't caught by stubtest because of this line in the allowlists:
> 
> https://github.com/python/typeshed/blob/9a76b13127ffa8365431dcc105fc111cdd267e7e/tests/stubtest_allowlists/py3_common.txt#L375

Thanks! We probably can do better with platform checks in this module (e.g., EDEADLOCK doesn't exist on Mac), but fixing that would be quite tedious. I suspect many that I added exist depending on the Linux kernel version or only on some obscure POSIX variant. For now let's just add everything that may exist.

srittau:
I'll ignore the primer error, which looks to be the same I had in my mypy PR earlier.

PR #7554, opened 2022-03-25
Add Server to asyncio.base_events' __all__

### LINKED ISSUE #7552: Include Server in __all__ in asyncio.base_events
https://github.com/python/typeshed/blob/9a76b13127ffa8365431dcc105fc111cdd267e7e/stdlib/asyncio/base_events.pyi#L18-L19
The standard library had an issue wrt asyncio.Server being documented but not existing.
This issue has been fixed in https://bugs.python.org/issue46955
What needs to happen for this to be fixed here?

vringar:
Closes #7552

### DIFF
--- a/stdlib/asyncio/base_events.pyi
+++ b/stdlib/asyncio/base_events.pyi
@@ -11,13 +11,12 @@ from socket import AddressFamily, SocketKind, _Address, _RetAddress, socket
 from typing import IO, Any, Awaitable, Callable, Coroutine, Generator, Sequence, TypeVar, overload
 from typing_extensions import Literal
 
-if sys.version_info >= (3, 10):
+if sys.version_info >= (3, 7):
     from contextvars import Context
 
+if sys.version_info >= (3, 9):
     __all__ = ("BaseEventLoop", "Server")
 elif sys.version_info >= (3, 7):
-    from contextvars import Context
-
     __all__ = ("BaseEventLoop",)
 else:
     __all__ = ["BaseEventLoop"]
--- a/stdlib/asyncio/base_events.pyi
+++ b/stdlib/asyncio/base_events.pyi
@@ -11,7 +11,11 @@ from socket import AddressFamily, SocketKind, _Address, _RetAddress, socket
 from typing import IO, Any, Awaitable, Callable, Coroutine, Generator, Sequence, TypeVar, overload
 from typing_extensions import Literal
 
-if sys.version_info >= (3, 7):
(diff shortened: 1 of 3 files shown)


vringar:
[Stub Versioning](https://github.com/python/typeshed/blob/9a76b13127ffa8365431dcc105fc111cdd267e7e/CONTRIBUTING.md#stub-versioning) told me to check for Python 3.10 when checking for a feature introduced in 3.10.3

stdlib/asyncio/base_events.pyi
```diff
@@ -11,7 +11,11 @@ from socket import AddressFamily, SocketKind, _Address, _RetAddress, socket
 from typing import IO, Any, Awaitable, Callable, Coroutine, Generator, Sequence, TypeVar, overload
 from typing_extensions import Literal
 
-if sys.version_info >= (3, 7):
+if sys.version_info >= (3, 10):
```
srittau, line 14: I think this was backported to 3.9, so:

```suggestion
if sys.version_info >= (3, 9):
```

JelleZijlstra:
> [Stub Versioning](https://github.com/python/typeshed/blob/9a76b13127ffa8365431dcc105fc111cdd267e7e/CONTRIBUTING.md#stub-versioning) told me to check for Python 3.10 when checking for a feature introduced in 3.10.3 but this is now failing in CI as they are running against 3.10.2. What should I do? Wait until the tests are updated?

We can just ignore the error for now by adding to the stubtest allowlist. We can remove that once GH is on 3.10.3.

stdlib/asyncio/base_events.pyi
```diff
@@ -11,7 +11,11 @@ from socket import AddressFamily, SocketKind, _Address, _RetAddress, socket
...
+    from contextvars import Context
```
srittau, line 15: Optional: We could also move the import into its own block. I found it a bit weird that they were combined before.

vringar:
I split the import from the `__all__` and adjusted for the backport.
So if you want to adjust the Allowlist, I think this is ready

hauntsaninja:
You'll need to add `asyncio.base_events.__all__` and `asyncio.Server` to:
https://github.com/python/typeshed/blob/master/tests/stubtest_allowlists/py39.txt
https://github.com/python/typeshed/blob/master/tests/stubtest_allowlists/py310.txt

hauntsaninja:
(pushed a commit to mypy_primer to get it green)

AlexWaygood (approved):
Thanks!

PR #7153, opened 2022-02-07
Add missing type hint for AbstractSet._hash()

jab:
`collections.abc.Set` provides a `_hash()` method (see https://github.com/python/cpython/blob/f20ca766/Lib/_collections_abc.py#L677) that includes the following in its docstring:
> Note that we don't define `__hash__`: not all sets are hashable. But if you define a hashable set type, its `__hash__` should call this function.

Following this instruction currently causes mypy to give an `attr-defined` error because `AbstractSet` is missing a type hint for this `_hash()` method.

I've been hitting this for a while in my [bidict](https://github.com/jab/bidict) library, whose `frozenbidict.__hash__()` returns `ItemsView(self)._hash()`, and working around it with a `type: ignore` comment, but I believe the fix is as simple as this PR.

Thanks for your consideration.

PR #7555, opened 2022-03-26
Add a 'copy' method to Mapping

DevilXD:
The `copy` method is missing from `Mapping`, which is supposedly to be an immutable representation of a dictionary. Since simple shallow copy operation doesn't modify anything, it belongs in `Mapping` instead of `MutableMapping`. This PR adds this missing method. Using the `copy` method avoids `from copy import copy` imports for creating a quick copy of the original dictionary. The return type is a `MutableMapping` since this method is usually used to create a local mutable copy to be passed somewhere, but if immutability needs to be preserved, one can still assign it to a `Mapping` variable type or use a cast.

I ran into this while trying to deal with `TypedDict` not being compatible with `Dict[str, Any]` (per https://github.com/python/mypy/issues/4976) and converting a JSON-saving function to accept a mapping, which is then modified before saving.


DevilXD:
The only error that appears as the result of this change (excluding the primer) seems to be tied to the same piece of code:

https://github.com/DevilXD/typeshed/blob/1b1933a10f924095b139cdd644633d02adb7a03f/stdlib/typing.pyi#L1210-L1212
https://github.com/DevilXD/typeshed/blob/1b1933a10f924095b139cdd644633d02adb7a03f/stdlib/typing_extensions.pyi#L63-L68

This seems to be because `TypedDict` inherits `Mapping` instead of `MutableMapping` for some reason, despite `TypedDict` instances being mutable. Since I'm not familiar with why that's the case, I'd appreciate someone more knowledgeable to chime in on this topic here.

srittau:
`typing.Mapping`/`collections.abc.Mapping` do not have a `copy()` method at runtime, so we can't add it to typeshed.

DevilXD:
That's... strange then. Isn't `Mapping` supposed to be an immutable representation of a dictionary? As far as I've understood it so far, `Mapping` should provide everything you need to work with a dictionary without modifying it (iterate over, read values, etc.), while `MutableMapping` adds a way to set values and modify the dictionary in some way, just like `Sequence` and `MutableSequence` does it for a list.

This sounds similar to https://github.com/python/typeshed/pull/7153 where `Set` (and thus `MutableSet`) defines a `_hash` method that's also missing at runtime due to `set` not actually inheriting from `MutableSet`. But then again, I think it's better not to throw an error at a valid code instead of the other way around, hence this addition.

JelleZijlstra:
As @srittau said, there is no `Mapping.copy` method at runtime, so it shouldn't be in typeshed either.

You can suggest adding it to CPython first, but it may be difficult because the ABC doesn't provide a way to create an instance of an arbitrary concrete Mapping.

Akuli:
> `Mapping` should provide everything you need to work with a dictionary without modifying it

This is close to how `Mapping` is intended to work, but not quite right. `Mapping` represents anything that has keys and values, not necessarily something copyable. If I create a class that represents something that just can't be reasonably copied, it should still be possible to make it a mapping. You can do `dict(any_mapping)`, but it doesn't make for `.copy()` to do that, because you would expect a copy to have the same type as the original object.

It might help to compare this to `list` and `MutableMapping`: lists have a `.sort()` method while `MutableMapping`s don't. I think this makes sense because Python's sorting implementation was designed for sorting lists efficiently, and if you want to sort something else, you're supposed to convert it to a list anyway (which is what `sorted()` does). So it doesn't make sense to sort non-list `MutableMapping`s in-place.

PR #7556, opened 2022-03-26
Add many missing `__match_args__` attributes

AlexWaygood:
Although stubtest doesn't (yet) complain if `__match_args__` is missing from the stub, it _does_ complain if `__match_args__` is _present but incorrect_ in the stub, so the CI should be fairly reliable here :)


AlexWaygood:
I guess I'll be right back with a flake8-pyi patch, then...

stdlib/_ast.pyi
```diff
@@ -27,126 +29,180 @@ if sys.version_info >= (3, 8):
...
+    if sys.version_info >= (3, 10):
+        __match_args__ = ("target", "annotation", "value", "simple")
     target: expr
     annotation: expr
     value: expr | None
     simple: int
 
 class For(stmt):
+    if sys.version_info >= (3, 10):
+        __match_args__ = ("target", "iter", "body", "orelse", "type_comment")
```
JelleZijlstra, line 121: There is no `type_comment` attribute in the stub, seems like we're missing it.

stdlib/_ast.pyi
```diff
@@ -27,126 +29,180 @@ if sys.version_info >= (3, 8):
...
     names: list[_identifier]
 
 class Nonlocal(stmt):
+    if sys.version_info >= (3, 10):
+        __match_args__ = ("names",)
     names: list[_identifier]
 
 class Expr(stmt):
+    if sys.version_info >= (3, 10):
+        __match_args__ = ("body",)
```
JelleZijlstra, line 205:
```suggestion
        __match_args__ = ("value",)
```

stdlib/_ast.pyi
AlexWaygood: It's inherited from `_ast.AST`, of which `_ast.stmt` is a subclass:

https://github.com/python/typeshed/blob/f40747f81df67098543ddf520194a6859e4ddf02/stdlib/_ast.pyi#L12-L22

stdlib/_ast.pyi
AlexWaygood: Huh, I wonder why stubtest failed on that one.

stdlib/_ast.pyi
JelleZijlstra: Thanks, I missed that

AlexWaygood: So did I the first time I looked :)

stdlib/concurrent/futures/_base.pyi
```diff
@@ -79,6 +79,7 @@ def as_completed(fs: Iterable[Future[_T]], timeout: float | None = ...) -> Itera
 
 # Ideally this would be a namedtuple, but mypy doesn't support generic tuple types. See #1976
 class DoneAndNotDoneFutures(Sequence[set[Future[_T]]]):
+    __match_args__ = ("done", "not_done")
```
AlexWaygood, line 82:
```suggestion
    if sys.version_info >= (3, 10):
        __match_args__ = ("done", "not_done")
```

AlexWaygood:
flake8-pyi PR here: https://github.com/PyCQA/flake8-pyi/pull/192

JelleZijlstra (approved):

AlexWaygood:
Many thanks for the reviews and flake8-pyi bugfix @JelleZijlstra!

PR #7558, opened 2022-03-26
Reduce code duplication in the email module

AlexWaygood:
2 type aliases and one `TypeVar` have duplicate definitions in several `email` submodules. I think it makes sense to move these definitions to a submodule in `_typeshed` instead, to reduce code duplicaiton.


srittau:
Sorry for the late reply, but do we need this in `_typeshed`? Wouldn't it make more sense to have those aliases in one of the `email` modules and import from there?

AlexWaygood:
> Sorry for the late reply

No worries — no rush!

> but do we need this in `_typeshed`? Wouldn't it make more sense to have those aliases in one of the `email` modules and import from there?

Sure — though if we do that, we'll probably have to give them names with a leading underscore, and it feels a bit weird to import names with a leading underscore from other modules.

I also thought that we already did something similar with `_typeshed.xml` and `_typeshed.wsgi` — but perhaps I just misunderstood the purpose of those submodules :)

srittau:
At least the aliases and protocols in `.wsgi` are mainly intended for external use, while the aliases here are more for internal use, if I understand correctly?

AlexWaygood:
> the aliases here are more for internal use, if I understand correctly?

Yes.

srittau:
To me, `_typeshed` is mainly intended for types that are also useful for typeshed users, or at least for types that are used by multiple parts of typeshed, while module-specific types should stay close to where they are used. But I'd be interested in what the general opinion on that is.

AlexWaygood:
What do we think about perhaps adding a fictitious submodule to the `email` package (`_aliases.py`?) that, like `_typeshed`, doesn't exist at runtime?

AlexWaygood:
> What do we think about perhaps adding a fictitious submodule to the `email` package (`_aliases.pyi`? `_typing.pyi`?) that, like `_typeshed`, doesn't exist at runtime?

There appears to be little enthusiasm for this idea. I'll move the aliases to `email.__init__` :)

srittau (approved):

AlexWaygood:
Thanks @srittau!

srittau:
> What do we think about perhaps adding a fictitious submodule to the email package (_aliases.pyi? _typing.pyi?) that, like _typeshed, doesn't exist at runtime?

Sorry for not answering earlier: `_typeshed.wsgi` used to be `wsgiref._types`, so there is precedent for that. But I prefer not to have "fake" modules, although I don't have a good justification for that, it's just a gut feeling.

PR #7562, opened 2022-03-28
locale: getlocale may return None's

henryiii:
According to the docs at https://docs.python.org/3/library/locale.html#locale.getlocale, this function may return None for either of the two items in the return sequence, which is missed in the current form.

This returns a length two tuple, even if the docs state "length 2 Sequence"; it seems more accurate and useful to make this a `Tuple[str | None, str | None]`, just like `getdefaultlocale` directly above this one. Both functions end up calling the same underlying function, so I think they should have the same return type, and the return type of `getdefaultlocale` is more useful.

The minimal fix would be `return Sequence[str | None]`, but I've chosen to match `getdefaultlocale` above.

https://github.com/python/cpython/blob/5944807b09717d43bb017f700e8c451dd07199ed/Lib/locale.py#L534-L603


AlexWaygood (changes requested):
Please adjust the type of `locale.setlocale()` as well, to get rid of the second false-positive error mypy-primer is reporting. It seems as though it will happily accept a tuple of `None`s as the second argument:

```python
>>> import locale
>>> locale.setlocale(locale.LC_CTYPE, (None, None))
'C'
```

henryiii:
`setlocale` with `(None, None)` is valid. The other issue pointed at in Pandas has a `all(x is not None`, which indicates this is correct (though sadly mypy is not...

Yes, what you said.

AlexWaygood:
> The other issue pointed at in Pandas has a `all(x is not None`, which indicates this is correct

Yeah, I think they'll just have to rewrite this snippet to use `TypeGuard`. As you say, it indicates that this is a valid concern.

henryiii:
I actually tested that before the original PR; using setlocale with `(None, None)` was how I quickly convinced myself that it is possible to get `(None, None)` back from the default (no argument) version of getlocale.

AlexWaygood (approved):
Thanks!

PR #7565, opened 2022-03-29
Update pyVmomi stubs

HansAarneLiblik:
Add some missing pyVmomi stubs. Not all, but some I'm missing

### DIFF
--- a/stubs/pyvmomi/pyVmomi/vmodl/__init__.pyi
+++ b/stubs/pyvmomi/pyVmomi/vmodl/__init__.pyi
@@ -1,5 +1,8 @@
 from typing import Any
 
+from .query import *
+from .fault import *
+
 class DynamicProperty:
     def __init__(self, *, name: str = ..., val: Any = ...) -> None: ...
     name: str
--- a/stubs/pyvmomi/pyVmomi/vmodl/__init__.pyi
+++ b/stubs/pyvmomi/pyVmomi/vmodl/__init__.pyi
@@ -1,6 +1,25 @@
 from typing import Any
 
+from fault import *
+from query import *
+
 class DynamicProperty:
     def __init__(self, *, name: str = ..., val: Any = ...) -> None: ...
     name: str
     val: Any
+
+class ManagedObject: ...
+
+class KeyAnyValue:
+    key: str
+    value: Any
+
+class LocalizableMessage:
+    key: str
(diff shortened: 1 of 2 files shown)


stubs/pyvmomi/pyVmomi/vmodl/__init__.pyi
```diff
@@ -1,6 +1,25 @@
 from typing import Any
 
+from fault import *
+from query import *
```
HansAarneLiblik, line 4: Is this the correct way to re-export them? Currently I'm getting
image

HansAarneLiblik: Actually i do not want to re-export them, so this import was unnecessary, but my IDE (IntelliJ 2021.3.3 with Python plugin 213.7172.26) still complains

srittau: How do you import `vmodl`? You could try `from . import query as query` in the stubs.

HansAarneLiblik: In my code: `from pyVmomi import vmodl`

sodul:
@HansAarneLiblik @srittau 

With this changes mypy is now raising:
`error: Exception type must be derived from BaseException`

I suspect that the `class ManagedObjectNotFound:` should be updated to `class ManagedObjectNotFound(Exception):`

Ideally we should have the entire class inheritance:
```
>>> import inspect
>>> inspect.getmro(vmodl.fault.ManagedObjectNotFound)
(<class 'pyVmomi.VmomiSupport.vmodl.fault.ManagedObjectNotFound'>, <class 'pyVmomi.VmomiSupport.vmodl.RuntimeFault'>, <class 'pyVmomi.VmomiSupport.vmodl.MethodFault'>, <class 'pyVmomi.VmomiSupport.vmodl.DynamicData'>, <class 'pyVmomi.VmomiSupport.DataObject'>, <class 'Exception'>, <class 'BaseException'>, <class 'object'>)
```

But fixing the regression would be good. Rolling back with `pip install types-pyvmomi==7.0.5` fixed the new errors. I get the issue with 7.0.6 and 7.0.7.

HansAarneLiblik:
@sodul Yup, `ManagedObjectNotFound` should also derive from `Exception`, my bad!

PR #7566, opened 2022-03-30
logging/handlers.pyi: atTime is datetime.time

stroxler:
It is marked as having type `datetime.datetime` but `atTime`
refers to a time of day; only the fields on `datetime.time`
are used. The [docs](https://docs.python.org/3/library/logging.handlers.html#timedrotatingfilehandler)
clearly state that it is a `time`.

One unfortunate problem is that `datetime.datetime` only inherits
from `datetime.date`, not `datetime.time` even though it exposes
the same API as `datetime.time`.

As a result, even though it is possible to pass a `datetime.datetime`
due to duck-typing, this change will causes code that does so to
not type check.

We could use a union type to allow passing either, or even create
a protocol for a "time-like interface". I'll leave that up to reviewers.

### DIFF
--- a/stdlib/logging/handlers.pyi
+++ b/stdlib/logging/handlers.pyi
@@ -78,7 +78,7 @@ class TimedRotatingFileHandler(BaseRotatingHandler):
     when: str  # undocumented
     backupCount: int  # undocumented
     utc: bool  # undocumented
-    atTime: datetime.datetime | None  # undocumented
+    atTime: datetime.time | None  # undocumented
     interval: int  # undocumented
     suffix: str  # undocumented
     dayOfWeek: int  # undocumented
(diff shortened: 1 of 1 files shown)


AlexWaygood:
> or even create a protocol for a "time-like interface".

I haven't looked at the source code, but from your description, that sounds like the best option to me! We should be as permissive as possible with parameter annotations. `_TimeLike | None` sounds like it would work well here.

stroxler:
Creating a protocol is probably the best answer in principle.

But there are tradeoffs because protocols are a more advanced feature, and generally less efficient to typecheck so I'm curious what other folks think - particularly given that the docs for the actual implementation specifically say `datetime.time`, and there's not a great reason to use a `datetime.datetime` for expressing a time of day (the fact that we can pass a datetime is not so much intent or reasonable semantics as a coincidence).

JelleZijlstra:
Given that the docs explicitly say `time` (I didn't check but trusting @stroxler) and that mypy-primer didn't report any cases where `datetime` was being used, it seems fine to just use `time`. We can switch to a protocol or Union if we get user reports that they want to pass something other than a `time`.

AlexWaygood:
> Given that the docs explicitly say `time` (I didn't check but trusting @stroxler) and that mypy-primer didn't report any cases where `datetime` was being used, it seems fine to just use `time`. We can switch to a protocol or Union if we get user reports that they want to pass something other than a `time`.

Fair enough

PR #7567, opened 2022-03-30
Fix source comment typo

luzpaz:
Found downstream in mypy (https://github.com/python/mypy/pull/12495)


JelleZijlstra:
There were a bunch of other typeshed typos in your PR, could you address those too?

luzpaz:
@JelleZijlstra they don't exist anymore AFAICT

PR #7568, opened 2022-03-31
_imp.create_dynamic returns ModuleType

collinanderson:
Just like `_imp.create_builtin`. It's only really documented in https://peps.python.org/pep-0489/


collinanderson:
Thanks!

PR #7577, opened 2022-04-01
Improve `imp._FileLike.__exit__`

AlexWaygood:
The signature of this method currently doesn't work quite as intended: see https://github.com/PyCQA/flake8-pyi/pull/199#issuecomment-1086087342

Classes will [still be accepted](https://mypy-play.net/?mypy=latest&python=3.10&gist=77efe095d01edeb1a4614166f0c9cf68) as conforming to this protocol if they have more permissive signatures such as `def __exit__(self, *args: object) -> None: ...`


JelleZijlstra (approved):
Contravariance to the rescue

Issue #7574, opened 2022-04-01
`sum` stub too permissive


jpy-git:
Currently the stub for the `sum` built-in allows any iterable as an argument
https://github.com/python/typeshed/blob/ae6ff79c0e14cd054bf07374e7091a098a46cf1f/stdlib/builtins.pyi#L1480-L1489

This means mypy has no issue with:
```python
sum("hello")
```
Which results in this error at runtime:
```
TypeError: unsupported operand type(s) for +: 'int' and 'str'
```

context: I was introducing someone to mypy and used the example above as something that mypy should flag, but turns out it wasn't being flagged so logging here.

Akuli:
It's easy to put a TypeVar bound on the first overload, but `sum()` doesn't work for strings even if you define `start=""`. It intentionally works with everything except strings.

```
>>> sum("asd", start="")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: sum() can't sum strings [use ''.join(seq) instead]
```

Fixing this properly would require https://github.com/python/typing/issues/1043.

AlexWaygood:
> It's easy to put a TypeVar bound on the first overload

Even that's actually a little tricky, because you get into the question of [what is a number](https://stackoverflow.com/questions/69334475/how-to-hint-at-number-types-i-e-subclasses-of-number-not-numbers-themselv/69383462#69383462).

For example, this works fine, even though there's not really anything that these classes all have in common either from a nominal or structural point of view:

```python
>>> import numpy as np
>>> from fractions import Fraction
>>> sum((3, np.float64(), Fraction(7, 22), complex(8, 0), 9.83))
(21.14818181818182+0j)
```

This does too:
```python
>>> from decimal import Decimal
>>> sum((3, Decimal('0.98')))
Decimal('3.98')
```

But this breaks:
```python
>>> from decimal import Decimal
>>> sum((2.34, Decimal('3.87')))
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: unsupported operand type(s) for +: 'float' and 'decimal.Decimal'
```

JelleZijlstra:
We could at least bound the TypeVar to a Protocol defining `__add__`. That wouldn't help with Alex's mixed numeric types, but it would help type checkers catch plausible mistakes like `sum([1, None])`.

PR #7578, opened 2022-04-01
`sum`: `Protocol` defining `__add__`

jpy-git:
Partially addresses #7574.

As suggested by @JelleZijlstra in this [comment](https://github.com/python/typeshed/issues/7574#issuecomment-1086019883) we should define an `__add__` protocol for use in `sum` to prevent things like 
```python
sum([1, None])
```


stdlib/builtins.pyi
```diff
@@ -1477,16 +1477,23 @@ def sorted(
...
+
+class _SupportsSum(Protocol):
+    def __add__(self, __x: Any) -> Any: ...
+
+_SumT = TypeVar("_SumT", bound=_SupportsSum)
+_SumS = TypeVar("_SumS", bound=_SupportsSum)
+
 @overload
-def sum(__iterable: Iterable[_T]) -> _T | Literal[0]: ...
+def sum(__iterable: Iterable[_SumT]) -> Any | Literal[0]: ...
```
AlexWaygood, line 1488: In 90% of cases, I think people will probably be using `sum` to add up variables that are all of the same type, e.g. `sum((1.4, 2.3, 3.7, 4.2))`. We can safely say for this that the result will be of type `float`.

So, although strictly speaking you're entirely correct — the return type isn't guaranteed to always be `_SumT`, far from it! — I nonetheless think I prefer using `_SumT` over `Any` for the return type, so that we can preserve precise type inference for the common case.

But I'm curious about what other people think :)

AlexWaygood: Also, note that the TypeVar is currently unbound in this overload, so either the return type needs to use `_SumT` instead of `Any`, or the parameter type needs to change to use `_SupportsSum` instead of `_SumT` :)

JelleZijlstra: What if we make the protocol generic over the return type of `__add__`, and make it `sum(Iterable[_SupportsSum[_T]]) -> _T`?

jpy-git: Ahh shoot this was actually unintentional, was playing round with the typings and forgot to change it back 😅 
I agree that `_SumT` makes sense, so will change it back for now 👍

jpy-git: > What if we make the protocol generic over the return type of `__add__`, and make it `sum(Iterable[_SupportsSum[_T]]) -> _T`?

Good shout, will give it a go

AlexWaygood: @JelleZijlstra, not sure it's that simple, because there's `__radd__` to consider, which can change the return type. If Y is a subclass of X, and Y defines `__radd__`, then `Y.__radd__ `takes precedence over `X.__add__` when evaluating the operation `x + y`

AlexWaygood: Idk  how much that actually matters though

JelleZijlstra: Yes, getting a completely precise return type in the face of mixed types and `__radd__` is very complicated. The current version is technically incorrect for the case where two arguments are passed, the iterable is empty, and the default returns an unrelated type from its `__add__`.

mypy-primer found a lot of new false positives from my suggestion, so we should probably undo that and just go back to assuming that `__add__` returns the same type.

jpy-git: Yeah I've reverted back to using TypeVars in the latest commit 😄

JelleZijlstra:

What happened to the freqtrade one?

AlexWaygood:
> > According to [mypy_primer](https://github.com/hauntsaninja/mypy_primer), this change has no effect on the checked open source code. 🤖🎉
> 
> What happened to the freqtrade one?
> 
> (code link: https://github.com/freqtrade/freqtrade/blob/0b40c37cc54dbf97ff8ed1b82330e9d158d91d6b/freqtrade/persistence/models.py#L1204)

Yes, that one looked like a true positive. I think  `close_profit_abs` attribute is declared here as `Optional[float]`: https://github.com/freqtrade/freqtrade/blob/0b40c37cc54dbf97ff8ed1b82330e9d158d91d6b/freqtrade/persistence/models.py#L296

jpy-git:
> What happened to the freqtrade one?

Was just wondering the same thing!

All that's changed vs the first commit is altering the return type
```
diff --git a/stdlib/builtins.pyi b/stdlib/builtins.pyi
index 9372b2a1..ce2042bc 100644
--- a/stdlib/builtins.pyi
+++ b/stdlib/builtins.pyi
@@ -1485,7 +1485,7 @@ _SumT = TypeVar("_SumT", bound=_SupportsSum)
 _SumS = TypeVar("_SumS", bound=_SupportsSum)
 
 @overload
-def sum(__iterable: Iterable[_SumT]) -> Any | Literal[0]: ...
+def sum(__iterable: Iterable[_SumT]) -> _SumT | Literal[0]: ...
 
     @overload
```

AlexWaygood:
This is very strange, I can't reproduce it on [MyPy playground](https://mypy-play.net/?mypy=latest&python=3.10&flags=strict&gist=a8bfcfcbac15dba29bb6a6b355ff9b11):

```python
from typing import *

class _SupportsSum(Protocol):
    def __add__(self, __x: Any) -> Any: ...

_SumT = TypeVar("_SumT", bound=_SupportsSum)
_SumS = TypeVar("_SumS", bound=_SupportsSum)

@overload  # type: ignore[no-overload-impl]
def sum(__iterable: Iterable[_SumT]) -> _SumT | Literal[0]: ...
@overload
def sum(__iterable: Iterable[_SumT], start: _SumS) -> _SumT | _SumS: ...

it: Iterable[float | None]

sum(it)  # error: Value of type variable "_SumT" of "sum" cannot be "Optional[float]"
```

AlexWaygood:
I've triggered a new run of mypy_primer

jpy-git:
So testing locally these examples work as expected for me
```python
from typing import Iterator, Union


def foo() -> Iterator[int]: ...

def bar() -> Iterator[Union[int, None]]: ...

sum([1, 2])
sum([1, None])
sum(foo())
sum(bar())
```
```
test.py:9: error: Value of type variable "_SumT" of "sum" cannot be "Optional[int]"
test.py:11: error: Value of type variable "_SumT" of "sum" cannot be "Optional[int]"
```

But if I run on `freqtrade` `mypy --custom-typeshed-dir . ../freqtrade/freqtrade/persistence/models.py | grep "models.py:1203"` returns nothing when we have the `_SumT` return type but with the `Any` return type I get:
`../freqtrade/freqtrade/persistence/models.py:1203: error: Value of type variable "_SumT" of "sum" cannot be "Optional[float]"` 🤔

JelleZijlstra:
I thought type context might help, but this variation that's closer to the freqtrade code doesn't repro it either: https://mypy-play.net/?mypy=latest&python=3.10&flags=strict&gist=3ba27f6c0834c95cfdb5b2d752387275

jpy-git:
Okay here's something odd. I've put a reveal_type around the line in question:
```python
total_profit = sum(
                reveal_type(t.close_profit_abs for t in LocalTrade.get_trades_proxy(is_open=False)))
```

If I run `pyright` then I get this output:
```
/Users/joe/repos/freqtrade/freqtrade/persistence/models.py:1204:29 - information: Type of "t.close_profit_abs for t in LocalTrade.get_trades_proxy(is_open=False)" is "Generator[float | None, None, None]"
  /Users/joe/repos/freqtrade/freqtrade/persistence/models.py:1204:17 - error: Argument of type "Generator[float | None, None, None]" cannot be assigned to parameter "__iterable" of type "Iterable[_SumT@sum]" in function "sum"
```
All works as expected

However, with `mypy`:
```
../freqtrade/freqtrade/persistence/models.py:1204: note: Revealed type is "typing.Generator[Union[builtins.float, None], None, None]"
../freqtrade/freqtrade/persistence/models.py:1204: note: Revealed type is "typing.Generator[Any, None, None]"
```
No error raised and 2 revealed types?? Seems like that second one could be causing the error to get ignored?

So is this a mypy specific issue?

AlexWaygood:
> So is this a mypy specific issue?

I think so, almost certainly :) We should probably merge this as it is, if @JelleZijlstra is in agreement with me. It's just very strange that we can't reproduce it in miniature.

The "double output for `reveal_type`" thing is similar to https://github.com/python/mypy/issues/10739, which might give us a clue.

AlexWaygood:
Something else that's interesting: this PR has also exposed a (completely unrelated) pyright bug, I think. The first commit, where the TypeVar was unbound in the first overload, should have triggered pyright's `reportInvalidTypeVarUse` check. So we really hit the jackpot here!

JelleZijlstra:
Yes, I think we should merge.

I tried some other variations in mypy playground but couldn't get the double reveal type. Normally that happens with TypeVars with constraints, but there aren't any here (or in the issue Alex linked above). Not sure what's happening there.

AlexWaygood (approved):
Thanks!

Akuli:
I think I know what's going on with freqtrade. The change in this PR that silenced `mypy_primer` was:

```diff
diff --git a/stdlib/builtins.pyi b/stdlib/builtins.pyi
index 9372b2a1..ce2042bc 100644
--- a/stdlib/builtins.pyi
+++ b/stdlib/builtins.pyi
@@ -1485,7 +1485,7 @@ _SumT = TypeVar("_SumT", bound=_SupportsSum)
 _SumS = TypeVar("_SumS", bound=_SupportsSum)
 
 @overload
-def sum(__iterable: Iterable[_SumT]) -> Any | Literal[0]: ...
+def sum(__iterable: Iterable[_SumT]) -> _SumT | Literal[0]: ...
 
 if sys.version_info >= (3, 8):
     @overload
```

In freqtrade, the result of `sum(...)` is assigned to `total_profit`, which has type `Any`. Previously mypy tried `_SumT = Optional[int]` when solving the TypeVars, and that caused the error. But now that `_SumT` is also used in the return type, and the result is assigned to `Any`, it doesn't give up there. It instead tries `_SumT = Any`, which succeeds and silences the error. You can see the two tries in the `reveal_type()` output, too.

JelleZijlstra:
Thanks @Akuli, I missed that the variable was of type `Any`. I can reproduce the double reveal_type() now: https://mypy-play.net/?mypy=latest&python=3.10&flags=strict&gist=4b96aaf8e1261db4bf484c0f76138638 (though interestingly only with overloads involved).

PR #7582, opened 2022-04-03
Upgrade `black` version

AlexWaygood:
black 22.3.0 contains [a patch](https://github.com/psf/black/pull/2966) by somebody called "Jelle Zijlstra", which should fix the CI failure in #7558


JelleZijlstra:
Thanks!

Issue #7576, opened 2022-04-01
Missing attributes for MappingView and subclasses in stdlib/collections_abc_.pyi


laundmo:
Due to `stdlib/_collections_abc.pyi` importing most directly from `stdlib/typing.pyi`, MappingView and its subclasses are missing the `self._mapping` attribute. This is an issue if inheriting from them for implementing custom views. This causes pyright and mypy (with `--check-untyped-defs`) to fail.

Minimal Example:
```py
import collections.abc

class MyView(collections.abc.KeysView):
    def __iter__(self):
        for key in self._mapping:
            yield key + 1
```
`mypy --check-untyped-defs test.py`
```
test.py:5: error: "MyView" has no attribute "_mapping"
```
`pyright test.py`
```
test.py
  test.py:5:25 - error: Cannot access member "_mapping" for type "MyView"
    Member "_mapping" is unknown (reportGeneralTypeIssues)
1 error, 0 warnings, 0 informations
```

I'm open to trying to PR this myself, though I have little experience writing stub files.

laundmo:
@AlexWaygood i would propose only adding them to `stdlib/_collections_abc.pyi` and not to typing.pyi, as they seem to be specific to the collections.abc implementation of MappingView, ValuesView etc.

in fact, i would be against adding them to typing.pyi as the attribute is not needed when implementing a completely custom MappingView that doesn't inherit from collections.abc

As for the usecase, in my case i wanted to create a custom mapping class that would convert the keys automatically, for which i need custom views to apply the conversions in case someone uses `.items()`, `.values()` etc. on my custom class.

AlexWaygood:
If you're proposing decoupling the types in `collections.abc` from the types in `typing`, that would be a significant undertaking 🙂 (see #6257 for previous discussion and attempts to do just that). I'd advise not trying to do that.

Akuli:
This is similar to #7153.

AlexWaygood:
> This is similar to #7153.

Yes, though https://github.com/python/typeshed/pull/7153 was regarding a method with an extensive docstring, whereas this attribute is entirely undocumented from what I can see.

laundmo:
> > This is similar to #7153.
> 
> Yes, though #7153 was regarding a method with an extensive docstring, whereas this attribute is entirely undocumented from what I can see.

entirely undocumented but also required if you want any tangible benefit from using collections.abc.MappingView and descendants

Akuli:
[...]

PR #7583, opened 2022-04-04
Add undocumented _mapping to collections.abc MappingView, KeysView, ValuesView, ItemsView

laundmo:
Added self._mapping attribute which was missing from collections.abc.MappingView and subclasses.

fixes: #7576

### DIFF
--- a/stdlib/_collections_abc.pyi
+++ b/stdlib/_collections_abc.pyi
@@ -14,12 +14,12 @@ from typing import (
     Generator as Generator,
     Generic,
     Hashable as Hashable,
-    ItemsView as ItemsView,
+    ItemsView as _ItemsView,
     Iterable as Iterable,
     Iterator as Iterator,
-    KeysView as KeysView,
+    KeysView as _KeysView,
     Mapping as Mapping,
-    MappingView as MappingView,
+    MappingView as _MappingView,
     MutableMapping as MutableMapping,
     MutableSequence as MutableSequence,
     MutableSet as MutableSet,
@@ -27,7 +27,8 @@ from typing import (
     Sequence as Sequence,
     Sized as Sized,
     TypeVar,
-    ValuesView as ValuesView,
+    ValuesView as _ValuesView,
+    Any as _Any,
 )
 from typing_extensions import final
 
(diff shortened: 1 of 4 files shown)


laundmo:
I'm unsure what's causing the allowlist failures, since I'm unsure what the allowlist does. Could anyone explain or help out?

AlexWaygood:
> I'm unsure what's causing the allowlist failures, since I'm unsure what the allowlist does. Could anyone explain or help out?

The stubtest failures are actually a good thing, in this case!

Stubtest is a script we run in CI that tests whether the stubs are consistent with the runtime — e.g. if a function at runtime has three arguments, but it only has two arguments in the stub, stubtest will complain about it. But it's not always possible for us to match the runtime implementation 100%, so we have an "allowlist" of objects that have slight discrepancies with the runtime implementation, and stubtest ignores for us.

In this situation, it seems like several allowlist entries are no longer needed for some reason. I'm not quite sure _why_, tbh, but fewer allowlist entries is always a good thing!

laundmo:
> 
> ```diff
> + homeassistant/components/yeelight/scanner.py:133: error: Incompatible return value type (got "dict_values[str, Any]", expected "ValuesView[Any]")  [return-value]
> + homeassistant/components/fritz/switch.py:201: error: Argument 3 to "device_filter_out_from_trackers" has incompatible type "dict_values[Any, Any]"; expected "ValuesView[Any]"  [arg-type]
> + homeassistant/components/fritz/device_tracker.py:62: error: Argument 3 to "device_filter_out_from_trackers" has incompatible type "dict_values[Any, Any]"; expected "ValuesView[Any]"  [arg-type]
> ```

Interesting, looking at these issues, they seem to typehint a function that takes `dict.values()` as collections.abc.ValuesView - which will now fail because `dict_values` can't contain _mapping.

I wonder whether this is common practice, and whether it should be. To me it seems wrong to use collections.abc here... but there might also be a way to prevent these errors that i dont know about.

AlexWaygood:
The allowlists are found in this folder here: https://github.com/python/typeshed/tree/master/tests/stubtest_allowlists

If you delete the relevant unused entries, the CI will go green :)

laundmo:
> If you delete the relevant unused entries, the CI will go green :)

Could you explain what makes them unused? i dont feel great deleting something without knowing what it does.

AlexWaygood:
> I wonder whether this is common practice, and whether it should be. To me it seems wrong to use collections.abc here...

To the contrary: [as the documentation states](https://docs.python.org/3/library/typing.html#typing.ValuesView), `typing.{Mapping,Keys,Values,Items}View` are all deprecated in 3.9+. Using the equivalent classes from `collections.abc` is the recommended practice for this kind of situation.

laundmo:
> > I wonder whether this is common practice, and whether it should be. To me it seems wrong to use collections.abc here...
> 
> To the contrary: [as the documentation states](https://docs.python.org/3/library/typing.html#typing.ValuesView), `typing.{Mapping,Keys,Values,Items}View` are all deprecated in 3.9+. Using the equivalent classes from `collections.abc` is the recommended practice for this kind of situation.

Would you have any idea how I could work around the issue that the collections.abc are supposed to be used for type hinting dict.values(), but also have the undocumented _mapping? It seems to me that adding _mapping will always cause false positives in that spot, but not adding it will cause false positives when subclassing the views.

AlexWaygood:
> It seems to me that adding _mapping will always cause errors in that spot

Yes, this is partly why I'm still not 100% sure that  merging this PR is a good idea :)

For the subclassing use case, have you considered just doing this? It would mean we wouldn't have to add the undocumented attribute to the stub, and [silences MyPy's complaints](https://mypy-play.net/?mypy=latest&python=3.10&gist=dbeaac3400d2f8f5cf93ad33bb4a87f1):

```python
import collections.abc
from typing import Any

class MyView(collections.abc.KeysView):
   _mapping: dict[str, Any]

   def __iter__(self):
        for key in self._mapping:
            yield key + 1
```

laundmo:
> For the subclassing use case, have you considered just doing this? 
>    _mapping: dict[str, Any]

Yes, that's actually what I'm currently doing, and what SQLAlchemy does too (they also subclass the views)
If it proves to be this much of a hassle, it might be better not to merge. Though, I wonder what might happen if a similar issue arises with more critical pieces of code.

AlexWaygood:
Another way of getting rid of the mypy_primer diff might be to just add the `_mapping` attribute to `typing.Mapping`, instead of attempting to decouple the `typing`/`collections.abc` classes. That's basically what we did with `collections.abc.Set._hash()` in #7153

laundmo:
> Another way of getting rid of the mypy_primer diff might be to just add the `_mapping` attribute to `typing.Mapping`, instead of attempting to decouple the `typing`/`collections.abc` classes. That's basically what we did with `collections.abc.Set._hash()` in #7153

Currently, I inherit `dict_{items,key,values}` from `typing.{Items,Keys,Values}View`, which is why they are erroring out when compared to `collections.abc.{Items,Keys,Values}View`.

I could instead inherit them from the new collections.abc Views which contain the _mapping, achieving the same effect.

This would make dict.values() etc show up as having the _mapping attribute (false positive). I wonder whether there's a way to make _mapping fail for dict_values while still not raising errors when passed to something type hinted as ValuesView.

AlexWaygood (changes requested):
I don't think it's a good idea to start adding attributes to `collections.abc.MappingView`, and deliberately not include them on `typing.MappingView`. For typing purposes, you're meant to be able to use the two interchangeably, and the mypy_primer output confirms that this would have unexpected and unfortunate consequences.

Please update your PR to simply add the `_mapping` attribute to `typing.MappingView`, instead of adding a whole new class to `_collections_abc.pyi`.

AlexWaygood:
Closing due to lack of response from OP.

@laundmo, if you're still interested in making this change, feel free to file a new PR that adds the attribute to `typing.MappingView` (etc.), instead of trying to separate `collections.abc.MappingView` into a separate class. Thanks! 🙂

PR #7589, opened 2022-04-04
Fix several redundant `bytes`/`bytearray`/`memoryview` unions

AlexWaygood:
Unblocks https://github.com/PyCQA/flake8-pyi/pull/201


JelleZijlstra:
I don't feel great about losing precision here. This isn't even specified in PEP 484.

For instance, I just looked at `builtins.pyi` and the first usage of bytes I found was for `ord()`: `def ord(__c: str | bytes) -> int: ...`. I looked at the implementation and it allows bytearray, but not memoryview. 

Next is the constructor of `UnicodeDecodeError`. We have it as `bytes`, but it actually allows any buffer type (so also e.g. `array.array`). 

Then `int.__new__`. We have one overload that takes `bytes` (among others) and another that takes `bytes | bytearray`. Checking the implementation, the first overload allows any buffer type and the second overload actually allows bytes and bytearray, but not memoryview.

I think we're better off reflecting the runtime reality exactly (using `_typeshed.ReadableBuffer` where appropriate), instead of pretending that these types are all the same.

AlexWaygood:
Thanks for calling me out. The `typing` docs seem very bad in this regard. According to [this BPO issue](https://bugs.python.org/issue33864), the intended meaning of the entry for `ByteString` is that "`memoryview` is sometimes a subtype of `ByteString` but not all the time". That wasn't obvious at all from my reading of the section, and I don't really understand how it can be put to practical use when using type hints.

Anyway, you're right that this approach is far too sweeping :)

PR #7591, opened 2022-04-04
Third-party stubs: make many attributes read-only properties

AlexWaygood:


srittau (approved):

stubs/beautifulsoup4/bs4/element.pyi
```diff
@@ -262,7 +262,8 @@ class Tag(PageElement):
     def __copy__(self: Self) -> Self: ...
     @property
     def is_empty_element(self) -> bool: ...
-    isSelfClosing = is_empty_element
+    @property
+    def isSelfClosing(self) -> bool: ...
```
hauntsaninja, line 266: Hmm, this one is actually less accurate to the runtime, maybe mypy / stubtest is missing a trick

AlexWaygood: Mypy interprets aliases of methods inside class bodies as instance variables. I noticed it a while back when working on a stubtest patch,but forgot to file an issue about it.

PR #7593, opened 2022-04-05
google-cloud-ndb depends on types-pytz

srittau:


srittau:
While this doesn't fix #7592, it still fixes a bug.

srittau:
Just updating the stubs to 1.11.* seems to fix stubtest.

PR #7594, opened 2022-04-05
Replace redundant sections with link to the typing documentation

srittau:


CONTRIBUTING.md
```diff
@@ -227,258 +227,11 @@ When the script has finished running, it will print instructions telling you wha
...
-See https://github.com/python/mypy/issues/7214 for more details.
-
-`__enter__` methods and other methods that return instances of the
-current class should be annotated with the `_typeshed.Self` type
-variable ([example](https://github.com/python/typeshed/pull/5698)).
-
-A few guidelines for protocol names below. In cases that don't fall
-into any of those categories, use your best judgement.
-
-* Use plain names for protocols that represent a clear concept
```
JelleZijlstra, line 476: This isn't in the linked doc

CONTRIBUTING.md
```diff
@@ -227,258 +227,11 @@ When the script has finished running, it will print instructions telling you wha
...
-When adding type annotations for context manager classes, annotate
-the return type of `__exit__` as bool only if the context manager
-sometimes suppresses exceptions -- if it sometimes returns `True`
-at runtime. If the context manager never suppresses exceptions,
-have the return type be either `None` or `bool | None`. If you
-are not sure whether exceptions are suppressed or not or if the
-context manager is meant to be subclassed, pick `bool | None`.
-See https://github.com/python/mypy/issues/7214 for more details.
-
-`__enter__` methods and other methods that return instances of the
```
JelleZijlstra, line 469: The typing docs don't mention this

CONTRIBUTING.md
```diff
@@ -227,258 +227,11 @@ When the script has finished running, it will print instructions telling you wha
...
 the latest mypy (`pip install -r requirements-tests.txt`) before running the script.
 
-### Supported type system features
-
-Since PEP 484 was accepted, there have been many other PEPs that added
-new features to the Python type system. In general, new features can
-be used in typeshed as soon as the PEP has been accepted and implemented
-and most type checkers support the new feature.
-
-Accepted features that *cannot* yet be used in typeshed include:
```
JelleZijlstra, line 237: I think this list is still useful, and it makes sense to keep it in typeshed because it's specifically about this repo's policies.

JelleZijlstra (changes requested):
I think this removes too much. I'm on board with shortening this text, but some parts are typeshed-specific and should be kept, and others aren't yet covered in the typing docs.

CONTRIBUTING.md
srittau: Whoops, I kind of forgot about this PR.

Maybe it's best to link to https://github.com/python/typeshed/issues?q=is%3Aissue+is%3Aopen+label%3Afeature-tracker, instead of listing the missing features? The issues are more likely to be kept up-to-date than the list in CONTRIBUTING.

JelleZijlstra: Sure, sounds good

AlexWaygood: We could also consider using a meta-issue tracker to track the issue trackers. The advantage of using an issue to track the issue trackers is that items will automatically be ticked off as completed if issue trackers on the bullet-pointed list are ticked off. We could link to the meta-issue tracker from `CONTRIBUTING.md` and pin it on the issue tracker.

PR #7595, opened 2022-04-05
Use Self return type for the 'replace()' methods in datetime classes

mzagozen:
Since Python 3.6, the `replace()` methods support subclasses - the return
type is the subclass being replaced.

Link to Python issue: https://bugs.python.org/issue31222

### DIFF
--- a/stdlib/datetime.pyi
+++ b/stdlib/datetime.pyi
@@ -65,7 +65,12 @@ class date:
     def isoformat(self) -> str: ...
     def timetuple(self) -> struct_time: ...
     def toordinal(self) -> int: ...
-    def replace(self, year: int = ..., month: int = ..., day: int = ...) -> date: ...
+    if sys.version_info >= (3, 6):
+        def replace(self: Self, year: int = ..., month: int = ..., day: int = ...) -> Self: ...
+    else:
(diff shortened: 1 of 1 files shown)


srittau:
Thanks, you still need to annotate `self` with `Self`, as this is currently simply a type var, not `typing.Self` from PEP 673.

mzagozen:
> Thanks, you still need to annotate `self` with `Self`, as this is currently simply a type var, not `typing.Self` from PEP 673.

@srittau I just noticed my editor was giving me the same error🤦

srittau (approved):

mzagozen:
This `mypy_primer` robot is pretty neat. I guess I should now make the equivalent PR in the pandas repo to fix https://github.com/pandas-dev/pandas/blob/main/pandas/_libs/tslibs/timestamps.pyi#L116?

JelleZijlstra:
> This `mypy_primer` robot is pretty neat. I guess I should now make the equivalent PR in the pandas repo to fix https://github.com/pandas-dev/pandas/blob/main/pandas/_libs/tslibs/timestamps.pyi#L116?

That would be great, but it's not required. We use the tool to get a sense of the impact of changes, and sometimes we decide new errors are acceptable and the library will have to fix them.

srittau:
@Akuli Good catch, #7602.

PR #7596, opened 2022-04-05
Remove a few more instances of `Union`

AlexWaygood:


srittau (approved):

Akuli:
There's not many old-style `Union`s left :)

```
akuli@akuli-desktop:~/typeshed$ git grep 'Union\[' | grep -v @python2 | wc -l
37
```

PR #7603, opened 2022-04-07
SQLAlchemy improvements for generative methods

hasier:
Fix return types for methods marked as `@_generative` in `sqlalchemy.sql.selectable`.

Inspired by https://github.com/python/typeshed/pull/7237 🙂


srittau (approved):
Thanks! I didn't verify against the implementation, but the changes look sensible.

PR #7604, opened 2022-04-07
SQLAlchemy improvements for Operators

hasier:
I've set the return of `ColumnOperator` methods as `ColumnOperators[_T]` as that feels like the intended behaviour. I did some manual tests as well and it looks like it works. I'd appreciate some comments and a nudge towards a better solution if I didn't get it right.

Also, I've changed method alias assignments for their actual signature.


stubs/SQLAlchemy/sqlalchemy/sql/operators.pyi
```diff
@@ -7,13 +7,13 @@ _T = TypeVar("_T")
...
-    def bool_op(self, opstring, precedence: int = ...): ...
-    def operate(self, op, *other, **kwargs) -> None: ...
-    def reverse_operate(self, op, other, **kwargs) -> None: ...
+    def __and__(self, other) -> Any: ...
+    def __or__(self, other) -> Any: ...
+    def __invert__(self) -> Any: ...
+    def op(self, opstring, precedence: int = ..., is_comparison: bool = ..., return_type: Any | None = ...) -> Any: ...
+    def bool_op(self, opstring, precedence: int = ...) -> Any: ...
+    def operate(self, op, *other, **kwargs) -> Any: ...
+    def reverse_operate(self, op, other, **kwargs) -> Any: ...
```
srittau, line 16: Please don't set the return type to `Any`, except if we know that there is no other way to type these return types. Just leave out the return type annotation.

stubs/SQLAlchemy/sqlalchemy/sql/sqltypes.pyi
```diff
@@ -256,7 +256,7 @@ class ARRAY(SchemaEventTarget, Indexable, Concatenable, TypeEngine):
     zero_indexes: bool
 
     class Comparator(Indexable.Comparator, Concatenable.Comparator):
-        def contains(self, *arg, **kw) -> None: ...
+        def contains(self, *arg, **kw) -> ARRAY.Comparator: ...
```
hasier, line 259: I'm thinking this should be `ColumnOperators` instead? wdyt?

PR #7606, opened 2022-04-07
Add `NoReturn` overload to `builtins.sum`

AlexWaygood:
`sum` fails if you try to pass an iterable of strings to it:

```python
>>> sum(('a', 'b', 'c'), start='d')
TypeError: sum() can't sum strings [use ''.join(seq) instead]
```

As pointed out by @Akuli, there's no ideal way to describe this in the stubs without https://github.com/python/typing/issues/1043. However, a `NoReturn` overload is better than nothing.

Helps with #7574

### DIFF
--- a/stdlib/builtins.pyi
+++ b/stdlib/builtins.pyi
@@ -1487,6 +1487,7 @@ _SumS = TypeVar("_SumS", bound=_SupportsSum)
 if sys.version_info >= (3, 8):
     @overload
     def sum(__iterable: Iterable[str], start: _SupportsSum = ...) -> NoReturn: ...
+
 else:
     @overload
     def sum(__iterable: Iterable[str], __start: _SupportsSum = ...) -> NoReturn: ...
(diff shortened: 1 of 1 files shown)


JelleZijlstra:
Not sure this is an improvement. NoReturn makes mypy silently discard code as unreachable, and pyright doesn't give an error if you use it either.

Consider this code:
```python
from typing import NoReturn

def f() -> NoReturn:
    raise Exception

def capybara() -> None:
    f()
    1 + "1"
```
Mypy type checks it without errors. Pyright detects the `1 + "1"`, but says nothing about the `f()` call.

AlexWaygood:
> `NoReturn` makes mypy silently discard code as unreachable

I hoped that was the advantage of this proposal -- that at least mypy would start raising errors for "unreachable code" if people had the `--warn-unreachable` setting switched on.

But I had forgotten that mypy simply starts ignoring type errors if it considers the code that contains the type error to be unreachable.

AlexWaygood:
Too many mypy_primer false negatives, and no clear indication that this would meaningfully improve things.

AlexWaygood:
> NoReturn makes mypy silently discard code as unreachable

Maybe we should reconsider `NoReturn` overloads like this one:

https://github.com/python/typeshed/blob/17b2f31b8cf67fe75fa71b7251535952229c5a02/stdlib/datetime.pyi#L73-L81

PR #7607, opened 2022-04-07
psycopg2: correct return type

qubidt:
Fixes the return type of `psycopg2.cursor.fetchone()` to match the psycopg2 code:

https://github.com/psycopg/psycopg2/blob/1d3a89a0bba621dc1cc9b32db6d241bd2da85ad1/psycopg/cursor_type.c#L647-L651
https://github.com/psycopg/psycopg2/blob/1d3a89a0bba621dc1cc9b32db6d241bd2da85ad1/psycopg/cursor_type.c#L748-L786

It also matches the [psycopg2 documentation](https://www.psycopg.org/docs/cursor.html?highlight=copy_from#cursor.fetchone) as well as the [DB-API](https://peps.python.org/pep-0249/#fetchone)


qubidt:
I just realized I overlooked something!  It's possible for `psycopg2.cursor` to take a `cursor_factor=psycopg2.extras.DictCursor`, which makes `cursor.fetchone()` return a `dict`. (`psycopg2.connection` has a corresponding `cursor_factory` parameter). Unfortunately I don't know how to make the type dependent on the runtime value of parameters. If there's not a better solution, perhaps this should be reverted.

JelleZijlstra:
Good catch! We could potentially handle this by making use of generics or overloads somewhere. Maybe open an issue with a full code sample demonstrating the problem so we can discuss it.

PR #7608, opened 2022-04-08
Add stubs for stdlib-list

kasium:

PR #7609, opened 2022-04-08
pytype_test: Don't mix up stdlib and stub packages starting with stdlib.

rchen152:
This should fix the pytype_test failure in
https://github.com/python/typeshed/pull/7608.


JelleZijlstra (approved):

JelleZijlstra:
Thanks for the quick fix!

PR #7611, opened 2022-04-09
asyncio: use WriteTransport in StreamWriter

### LINKED ISSUE #7610: asyncio.StreamWriter.transport not specific enough
This works at runtime:

```
import asyncio

async def handle_connection(reader: asyncio.StreamReader, writer: asyncio.StreamWriter) -> None:
    print(writer.transport.get_write_buffer_size())
    writer.write(b"x" * (256*1024*1024))
    print(writer.transport.get_write_buffer_size())

async def main() -> None:
    async with await asyncio.start_server(handle_connection, port=1234) as server:
        await server.serve_forever()

asyncio.run(main())
```

but doesn't type check:

```
/tmp/echo2.py:4: error: "BaseTransport" has no attribute "get_write_buffer_size"
/tmp/echo2.py:6: error: "BaseTransport" has no attribute "get_write_buffer_size"
```

According to asyncio docs, a `WriteTransport` has a `get_write_buffer_size()` method, and guessing from the names of things, I think `writer.transport` should always be a `WriteTransport`.

Akuli:
Fixes #7610


JelleZijlstra (approved):
StreamWriter's own code uses several attributes on the transport that only exist on WriteTransport.

PR #7613, opened 2022-04-12
Stubs for Python-Jose

matt-slack:


stubs/python-jose/jose/backends/__init__.pyi
```diff
@@ -0,0 +1,9 @@
+from typing import Any, Callable
+
+from .base import DIRKey as DIRKey
+
+AESKey = Any
+HMACKey = Any
+RSAKey = Any
+ECKey = Any
```
srittau, line 8: Could you add a comment to the code why these are `Any`?

stubs/python-jose/jose/backends/__init__.pyi
```diff
@@ -0,0 +1,9 @@
+from typing import Any, Callable
```
srittau, line 1: Nit: `Callable` should be imported from `collections.abc`.

stubs/python-jose/jose/backends/base.pyi
```diff
@@ -0,0 +1,19 @@
+from typing import Any
+
+from ..utils import base64url_encode as base64url_encode, ensure_binary as ensure_binary
```
srittau, line 3: This is the first instance, but there are more below. This is probably a stubgen artifact, but we should not re-export, except from `__init__.pyi`. In this particular case, we can just completely remove the import.

srittau (changes requested):
Thanks, a few remarks below.

stubs/python-jose/jose/backends/__init__.pyi
matt-slack: This has been updated to the correct import

stubs/python-jose/jose/backends/__init__.pyi
matt-slack: I changed the types to better align with what is happening here. Added some comments to clarify. Hope this helps.

stubs/python-jose/jose/backends/base.pyi
matt-slack: Correct, I just glazed over the imports without thinking about it. Removed

srittau (approved):

PR #7614, opened 2022-04-12
Drop some literal types from argparse (add_argument)

JukkaL:
These were introduced in #7329 and they cause false positives
in code that used to be accepted before. There was a false
positive in https://github.com/pycqa/pylint (encountered in
https://github.com/python/mypy/pull/12321) and I also saw
false positives in an internal codebase.

I suggest not using literal types here, since they can be kind of
awkward to annotate in callers that don't pass a literal string.
Requiring callers to write annotations like
`Literal["?", "*", "+", "...", "A...", "==SUPPRESS=="]` is not
very user friendly.

### DIFF
--- a/stdlib/argparse.pyi
+++ b/stdlib/argparse.pyi
@@ -106,11 +106,8 @@ class _ActionsContainer:
     def add_argument(
         self,
         *name_or_flags: str,
-        action: Literal[
-            "store", "store_const", "store_true", "store_false", "append", "append_const", "count", "help", "version", "extend"
-        ]
-        | type[Action] = ...,
-        nargs: int | Literal["?", "*", "+", "...", "A...", "==SUPPRESS=="] | _SUPPRESS_T = ...,
+        action: str | type[Action] = ...,
+        nargs: int | str | _SUPPRESS_T = ...,
         const: Any = ...,
         default: Any = ...,
         type: Callable[[str], _T] | FileType = ...,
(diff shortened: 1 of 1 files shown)


srittau:
I'm not terribly happy about reverting these changes. Catching typos in "enum"-like strings is a perfect use case for type checking. But the lack of usability for these types of annotations remains a concern that needs to be discussed.

AlexWaygood:
I have similar thoughts to @srittau. Downgrading the precision of the stubs is something I'm never particularly happy about. But I can definitely see the usability concern here, so I guess I'm +0 on this change. 

Pinging @JelleZijlstra and @Akuli, who were involved in reviewing the original PR.

JukkaL:
> But the lack of usability for these types of annotations remains a concern that needs to be discussed.

I think that this is the key issue. The types aren't available anywhere where they can be easily found. Even if we add a type alias in the stub, they won't be easily discoverable and can't be imported at runtime. I agree that it would be nice if we could check these, but I don't see a way to do this without a bad user experience. 

Also, the benefit of type checking these seems a bit limited since the error would be caught immediately after running the program. `argparse` does pretty decent runtime checking, and argument parsing is usually performed always at the beginning of execution. Here's an example traceback:
```
  File "/Users/jukka/.../t.py", line 4, in <module>
    p.add_argument('--py', action='xyz')
  File "/Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/argparse.py", line 1417, in add_argument
    raise ValueError('unknown action "%s"' % (action_class,))
ValueError: unknown action "xyz"
```

AlexWaygood (approved):

srittau:
Let's at least define aliases to `str` (`_ActionType` and `_NArgs`?) for the former literals and keep the literal values as comments.

srittau (approved):
Thanks!

PR #7615, opened 2022-04-12
Add missing definitions to six.moves.http_client

JukkaL:
These were removed as a side-effect of #7327, which added
`__all__` to `http.client`. I checked that all the definitions
are present at runtime via `six.moves.http_client`.


srittau (approved):
I didn't double check, but LGTM.

PR #7618, opened 2022-04-13
Fix various `__all__` bugs and omissions

AlexWaygood:
Also, `typing_extensions.get_type_hints`, `typing_extensions.get_args` and `typing_extensions.get_origin` are "available" on Python 3.6 now that `typing_extensions` [has dropped support for 3.6](https://github.com/python/typing/commit/0ca0e458324ed27f96d81265faf94595f57d84af).


stdlib/_pydecimal.pyi
```diff
@@ -1,3 +1,85 @@
+import sys
+
 # This is a slight lie, the implementations aren't exactly identical
 # However, in all likelihood, the differences are inconsequential
 from decimal import *
+
+if sys.version_info >= (3, 7):
+    __all__ = [
```
AlexWaygood, line 8: Surprisingly, `_pydecimal` appears to have `__all__` even though `decimal` doesn't. `¯\_(ツ)_/¯`

stdlib/typing_extensions.pyi
```diff
@@ -82,15 +132,14 @@ TypedDict: object
 
 OrderedDict = _Alias()
 
-if sys.version_info >= (3, 7):
```
JelleZijlstra, line 85: Technically we shouldn't do this because the change hasn't been released yet, but I don't want to have to duplicate `__all__` here.

PR #7623, opened 2022-04-14
Accept `mtime=None` in `linecache.cache` entries.

till-varoquaux:
The mtime in linecache entries can be set to `None` fir files loaded via a `__loader__` (cf: https://github.com/python/cpython/blob/3.10/Lib/linecache.py#L70)

### DIFF
--- a/stdlib/linecache.pyi
+++ b/stdlib/linecache.pyi
@@ -7,7 +7,7 @@ else:
     __all__ = ["getline", "clearcache", "checkcache"]
 
 _ModuleGlobals = dict[str, Any]
-_ModuleMetadata = tuple[int, float, list[str], str]
+_ModuleMetadata = tuple[int, float | None, list[str], str]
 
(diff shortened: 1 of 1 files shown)


stdlib/linecache.pyi
```diff
@@ -7,7 +7,7 @@ else:
     __all__ = ["getline", "clearcache", "checkcache"]
 
 _ModuleGlobals = dict[str, Any]
-_ModuleMetadata = tuple[int, float, list[str], str]
+_ModuleMetadata = tuple[int, float, list[str] | None, str]
```
AlexWaygood, line 10: Isn't `mtime` the second item in the tuple, rather than the third item?

https://github.com/python/cpython/blob/3.10/Lib/linecache.py#L68

till-varoquaux: Crap. You're right

AlexWaygood (approved):

PR #7626, opened 2022-04-15
ExceptionGroup: make it generic

JelleZijlstra:
See motivation in python/cpython#91549.

I used the following tests:

```python
beg = BaseExceptionGroup("x", [SystemExit()])
assert_type(beg, BaseExceptionGroup[SystemExit])
assert_type(beg.exceptions, tuple[SystemExit | BaseExceptionGroup[SystemExit], ...])
assert_type(
    beg.subgroup(KeyboardInterrupt), BaseExceptionGroup[KeyboardInterrupt] | None
)
assert_type(
    beg.subgroup((KeyboardInterrupt, SystemExit)),
    BaseExceptionGroup[KeyboardInterrupt | SystemExit] | None,
)
assert_type(
    beg.split(KeyboardInterrupt),
    tuple[
        BaseExceptionGroup[KeyboardInterrupt] | None,
        BaseExceptionGroup[SystemExit] | None,
    ],
)
assert_type(
    beg.split((KeyboardInterrupt, SystemExit)),
    tuple[
        BaseExceptionGroup[KeyboardInterrupt | SystemExit] | None,
        BaseExceptionGroup[SystemExit] | None,
    ],
)

eg = ExceptionGroup("x", [ValueError(), KeyError()])
assert_type(eg, ExceptionGroup[ValueError | KeyError])
assert_type(
    eg.exceptions,
    tuple[ValueError | KeyError | ExceptionGroup[ValueError | KeyError], ...],
)
```

These all pass in pyright, but mypy for some reason fails on a lot of the `split` and `subgroup` calls, with errors like `samples/exceptiongroup.py:107: error: Expression is of type "Optional[BaseExceptionGroup[Any]]", not "Optional[BaseExceptionGroup[KeyboardInterrupt]]"`. Seems like a bug.

The overloads I added trade precision in the generic argument against precision in the generic type itself. If you have a subclass `ChildExceptionGroup(ExceptionGroup)` and call `.split(SomeException)` on it, you get an `ExceptionGroup[SomeException]` from the type checker, though at runtime you'd get a `ChildExceptionGroup[SomeException]`. I think that's the right tradeoff because subclassing ExceptionGroup is likely to be rare.


AlexWaygood (approved):
I haven't studied the finer details of exception groups, but this seems reasonable, and mypy/pyright seem happy.

stdlib/builtins.pyi
```diff
@@ -1747,19 +1746,45 @@ if sys.version_info >= (3, 10):
...
-        def subgroup(self: Self, __condition: _SplitCondition) -> Self | None: ...
-        def split(self: Self, __condition: _SplitCondition) -> tuple[Self | None, Self | None]: ...
-        def derive(self: Self, __excs: Sequence[BaseException]) -> Self: ...
+        def exceptions(self) -> tuple[_BaseExceptionT_co | BaseExceptionGroup[_BaseExceptionT_co], ...]: ...
+        @overload
+        def subgroup(
+            self, __condition: type[_BaseExceptionT] | tuple[type[_BaseExceptionT], ...]
+        ) -> BaseExceptionGroup[_BaseExceptionT] | None: ...
+        @overload
+        def subgroup(self: Self, __condition: Callable[[_BaseExceptionT_co], bool]) -> Self | None: ...
```
srittau, line 1765: One minor gripe I have is that `_BaseExceptionT` and `_BaseExceptionT_co` have a very similar name, which means that it's easy to miss the subtleties in cases like above. But I have no immediate idea of better names.

stdlib/builtins.pyi
```diff
@@ -1747,19 +1746,45 @@ if sys.version_info >= (3, 10):
...
+        @overload
+        def split(
+            self: Self, __condition: type[_BaseExceptionT] | tuple[type[_BaseExceptionT], ...]
+        ) -> tuple[BaseExceptionGroup[_BaseExceptionT] | None, Self | None]: ...
+        @overload
+        def split(self: Self, __condition: Callable[[_BaseExceptionT_co], bool]) -> tuple[Self | None, Self | None]: ...
+        def derive(self: Self, __excs: Sequence[_BaseExceptionT_co]) -> Self: ...
```
srittau, line 1772: This is a bit of a tricky one. If I read PEP 654 correctly, the `__excs` provided don't need to have anything in common with the exceptions of `self`. So a "correct" annotation would be:

```python
def derive(self: Self, __excs: Sequence[_BaseExceptionT]) -> Self[_BaseExceptionT]: ...
```

But I don't think that's possible. So I'm fine with this, until someone comes up with a better alternative.

srittau (approved):

stdlib/builtins.pyi
AlexWaygood: Yet another use case for https://github.com/python/typing/issues/548...

stdlib/builtins.pyi
JelleZijlstra: I initially had them as `_BaseExceptionT` and `_BaseExceptionU` but that seemed worse.

youtux:
why should this be the case?
```python
assert_type(beg.exceptions, tuple[SystemExit | BaseExceptionGroup[SystemExit], ...])
```

shouldn't instead be like this?
```python
assert_type(beg.exceptions, tuple[SystemExit, ...])
```

Asking because it causes unexpected typing when using `except*`: https://mypy-play.net/?mypy=latest&python=3.12&gist=059952b22801648257f86bee1fbfd404

youtux:
nevermind, just tried https://mypy-play.net/?mypy=latest&python=3.12&gist=059952b22801648257f86bee1fbfd404 and indeed it can have a nested ExceptionGroup in it. Quite unexpected for me, but this is not the forum to raise the issue.

PR #7627, opened 2022-04-15
Add typing_extensions.assert_type

JelleZijlstra:
Hasn't been released yet but I'd like it to be in the next mypy release.

### DIFF
--- a/stdlib/typing_extensions.pyi
+++ b/stdlib/typing_extensions.pyi
@@ -186,12 +186,13 @@ else:
 
 # New things in 3.11
 if sys.version_info >= (3, 11):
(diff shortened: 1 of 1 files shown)


hauntsaninja (approved):

AlexWaygood:
Should this also be added to `__all__`?

JelleZijlstra:
> Should this also be added to `__all__`?

Yes, thanks!

AlexWaygood:
> Should this also be added to `__all__`?

Seems like it was actually missing from the runtime implementation! I filed https://github.com/python/typing/pull/1136

PR #7634, opened 2022-04-16
Use re-exports instead of TypeAliases in several places

### LINKED ISSUE #7632: Missing update method on the CaseInsensitiveDict
We got mypy errors after the release 6 hours ago:

```
communication/slack/tasks.py:39: error: "CaseInsensitiveDict?[builtins.str]" has no attribute "update"
```

This is when we do (simplified):
```python
session = requests.Session()
session.headers.update({"Authorization": f"Bearer {access_token}"})
```

Also part of the docs: https://docs.python-requests.org/en/master/user/advanced/

I think the `CaseInsensitiveDict` should also contain the `.update()` method. Will open a PR in a second.

AlexWaygood:
Fixes #7632

### DIFF
--- a/stubs/requests/requests/sessions.pyi
+++ b/stubs/requests/requests/sessions.pyi
@@ -1,14 +1,12 @@
 from _typeshed import Self, SupportsItems
-from typing import IO, Any, Callable, Iterable, Mapping, MutableMapping, TypeVar, Union
+from typing import IO, Any, Callable, Iterable, Mapping, MutableMapping, Union
 from typing_extensions import TypeAlias
 
-from urllib3 import _collections
+from urllib3._collections import _collections import RecentlyUsedContainer as RecentlyUsedContainer
 
-from . import adapters, auth as _auth, compat, cookies, exceptions, hooks, models, status_codes, structures, utils
+from . import adapters, auth as _auth, compat, cookies, exceptions, hooks, models, status_codes, utils
 from .models import Response
-
-_KT = TypeVar("_KT")
-_VT = TypeVar("_VT")
+from .structures import CaseInsensitiveDict as CaseInsensitiveDict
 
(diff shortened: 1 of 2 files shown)


stubs/requests/requests/sessions.pyi
```diff
@@ -1,14 +1,12 @@
 from _typeshed import Self, SupportsItems
-from typing import IO, Any, Callable, Iterable, Mapping, MutableMapping, TypeVar, Union
+from typing import IO, Any, Callable, Iterable, Mapping, MutableMapping, Union
 from typing_extensions import TypeAlias
 
-from urllib3 import _collections
+from urllib3._collections import RecentlyUsedContainer
```
hauntsaninja, line 5:
```suggestion
from urllib3._collections import RecentlyUsedContainer as RecentlyUsedContainer
```

hauntsaninja (changes requested):

stubs/requests/requests/sessions.pyi
AlexWaygood: I did that initially, but stubtest called me out. Turns out that it doesn't actually exist at runtime, we just use it for annotations in the stub. Not sure why stubtest didn't flag it before now.

Source: https://github.com/psf/requests/blob/main/requests/sessions.py

Stubtest: https://github.com/python/typeshed/runs/6046468647?check_suite_focus=true

hauntsaninja: Oh interesting. stubtest's checking of type aliases is a little loose, see https://github.com/python/mypy/blob/ce858a6a0c85206ff877faf702293a9b67fae750/mypy/stubtest.py#L960 (I don't quite remember why I did it this way, maybe worth reinvestigating)

hauntsaninja (approved):

Issue #7628, opened 2022-04-15
`gettext.GNUTranslations.__init__`’s argument has an incorrect type


Jayman2000:
`mypy` says that this program is wrong:

```python
from gettext import GNUTranslations

with open("foo.mo", mode='rb') as mo_file:
    GNUTranslations(mo_file)
```

`mypy`’s output:

```
main.py:4: error: Argument 1 to "GNUTranslations" has incompatible type "BufferedReader"; expected "Optional[IO[str]]"
Found 1 error in 1 file (checked 1 source file)
```

I believe that `fp` (`gettext.GNUTranslations.__init__`’s argument) should have the type `Optional[IO[bytes]]`, not `Optional[IO[str]]`.

TeamSpen210:
The problem is that `NullTranslations` (its base class) is hinted as taking an `IO[str] | None`, but `GNUTranslations` needs a binary file. `NullTranslations` doesn't actually do anything with the file, just store it for subclasses to use so the best solution would probably be to make `NullTranslations` take `IO | None` and then constrain in `GNUTranslations`.

srittau:
I'd say we just type `NullTranslations` according to how we type `GNUTranslations` (as long as it doesn't contradict the actual implementation in the former) as the latter is the only "real" translations class we have. We also should replace `IO` with a tight protocol.

PR #7636, opened 2022-04-16
Update gettext translations

srittau:
* Use a protocol instead of `IO` for `fp` argument.
* Annotate a few missing return types.
* Mark some class constants as final.

Fixes: #7628


stdlib/gettext.pyi
```diff
@@ -67,9 +68,14 @@ else:
         "ngettext",
     ]
 
+class _TranslationsReader(Protocol):
+    def read(self) -> bytes: ...
+    # optional:
+    # name: str
```
srittau, line 74: I didn't use `SupportsRead` deliberately:

* The `__length` argument to `read()` is not required.
* To document the optional `name` field. We could add it if something like python/typing#601 gets implemented.

JelleZijlstra (approved):

PR #7639, opened 2022-04-16
Use tuple instead of _typing.Tuple in ast

srittau:


AlexWaygood:
Hah, I guess this foiled flake8-pyi due to the use of an alias when importing `typing`.

AlexWaygood (approved):

PR #7640, opened 2022-04-16
Move code from `decimal.pyi` to `_decimal.pyi`

AlexWaygood:
The stub currently does `from decimal import *` in `_decimal.pyi`, but [the opposite happens at runtime](https://github.com/python/cpython/blob/9300b6d72948b94c0924a75ea14c6298156522d0/Lib/decimal.py#L3-L6).


stdlib/_decimal.pyi
```diff
@@ -1,3 +1,265 @@
...
+from _typeshed import Self
+from types import TracebackType
+from typing import Any, Container, NamedTuple, Sequence, Union, overload
+from typing_extensions import TypeAlias
 
+_Decimal: TypeAlias = Decimal | int
+_DecimalNew: TypeAlias = Union[Decimal, float, str, tuple[int, Sequence[int], int]]
+_ComparableNum: TypeAlias = Decimal | float | numbers.Rational
+
+__doc__: str
```
AlexWaygood, line 12:
```suggestion
```

AlexWaygood: mypy thinks `__doc__` is already defined.

srittau:
Please see the comment in `_pydecimal.pyi`. Both use the same stubs.

AlexWaygood:
> Please see the comment in `_pydecimal.pyi`. Both use the same stubs.

Sure. Should we change the import in `_pydecimal.pyi` to import from `_decimal` instead of `_decimal`, or do you think this is just not worth doing? :)

At the moment I feel like we have two lies, one in `_pydecimal` and one in `decimal`. I don't mind the one in `_pydecimal`, as it avoids code duplication, but I'm not sure the one in `decimal` is really necessary.

stdlib/decimal.pyi
```diff
@@ -1,263 +1,2 @@
...
-    def to_sci_string(self, __x: _Decimal) -> str: ...
-    def to_integral_exact(self, __x: _Decimal) -> Decimal: ...
-    def to_integral_value(self, __x: _Decimal) -> Decimal: ...
-    def to_integral(self, __x: _Decimal) -> Decimal: ...
-
-DefaultContext: Context
-BasicContext: Context
-ExtendedContext: Context
+from _decimal import *
+from _decimal import __doc__ as __doc__, __libmpdec_version__ as __libmpdec_version__, __version__ as __version__
```
AlexWaygood, line 2:
```suggestion
from _decimal import __libmpdec_version__ as __libmpdec_version__, __version__ as __version__
```

JelleZijlstra (approved):

srittau (approved):

PR #7643, opened 2022-04-16
typing updates for 3.11

JelleZijlstra:
- get_overloads
- clear_overloads
- dataclass_transform updates

### DIFF
--- a/stdlib/typing.pyi
+++ b/stdlib/typing.pyi
@@ -84,9 +84,11 @@ if sys.version_info >= (3, 11):
         "assert_never",
         "assert_type",
         "cast",
+        "clear_overloads",
         "final",
         "get_args",
         "get_origin",
+        "get_overloads",
         "get_type_hints",
         "is_typeddict",
         "Never",
@@ -1180,6 +1182,8 @@ if sys.version_info >= (3, 11):
     def reveal_type(__obj: _T) -> _T: ...
     def assert_never(__arg: Never) -> Never: ...
     def assert_type(__val: _T, __typ: Any) -> _T: ...
+    def clear_overloads() -> None: ...
+    def get_overloads(func: Callable[..., object]) -> Sequence[Callable[..., object]]: ...
 
 # Type constructors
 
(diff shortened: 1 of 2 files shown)


stdlib/typing_extensions.pyi
```diff
@@ -200,6 +205,8 @@ else:
     def reveal_type(__obj: _T) -> _T: ...
     def assert_never(__arg: NoReturn) -> NoReturn: ...
     def assert_type(__val: _T, __typ: Any) -> _T: ...
+    def clear_overloads() -> None: ...
+    def get_overloads(func: Callable[..., object]) -> Sequence[Callable[..., object]]: ...
 
 # Experimental (hopefully these will be in 3.11)
```
srittau, line 211: I guess this comment is now outdated?

stdlib/typing.pyi
```diff
@@ -1180,6 +1182,8 @@ if sys.version_info >= (3, 11):
     def reveal_type(__obj: _T) -> _T: ...
     def assert_never(__arg: Never) -> Never: ...
     def assert_type(__val: _T, __typ: Any) -> _T: ...
+    def clear_overloads() -> None: ...
+    def get_overloads(func: Callable[..., object]) -> Sequence[Callable[..., object]]: ...
```
AlexWaygood, line 1186: This is an interesting one. I wonder if this is the best return annotation. You *could* call the overload implementations, but it wouldn't be very useful to do so, since all their bodies are empty.

I don't have anything better, though.

stdlib/typing_extensions.pyi
JelleZijlstra: For most of them, dataclass_transform is still pending. I'll update this.

AlexWaygood (approved):

stdlib/typing.pyi
JelleZijlstra: I guess we could use types.FunctionType. But `inspect.signature` is annotated as taking a Callable, and `inspect.signature` is the main useful thing you could do with the return value of `get_overloads()`.

AlexWaygood: Yeah, that was my basic thought process as well

srittau (approved):

PR #7644, opened 2022-04-16
Add wsgiref.types (Python 3.11+)

srittau:
Use `wsgiref.types` in `_typeshed.wsgi` on Python 3.11+.

Make `_typeshed.wsgi` types match `wsgiref.types` on Python < 3.11.


srittau:
Stubtest doesn't like this for obvious reasons. Should we wait until 3.11b1 or should I add this module to the allowlist?

srittau:
Although I'm unsure why this is an error when running with Python < 3.11 as `VERSIONS` should exclude `wsgiref.types` from being checked.

AlexWaygood:
Looks like you need to change some imports in some other modules:

https://github.com/python/typeshed/blob/master/stdlib/wsgiref/handlers.pyi#L7

Your change to `VERSIONS` means that `wsgiref.types` is no longer available on <3.11, so you'll need to change that import to import things from `_typeshed.wsgiref` instead

srittau:
Good catch, thanks!

JelleZijlstra (approved):

AlexWaygood:
> Good catch, thanks!

Np!

AlexWaygood (approved):

PR #7645, opened 2022-04-16
Add _typeshed.(Opt)ExcInfo

srittau:


stdlib/cgitb.pyi
```diff
@@ -1,9 +1,6 @@
-from _typeshed import StrOrBytesPath
+from _typeshed import OptExcInfo, StrOrBytesPath
 from types import FrameType, TracebackType
 from typing import IO, Any, Callable
-from typing_extensions import TypeAlias
-
-_ExcInfo: TypeAlias = tuple[type[BaseException] | None, BaseException | None, TracebackType | None]
```
AlexWaygood, line 4: Hmm, `tuple[type[BaseException], BaseException, TracebackType] | tuple[None, None, None]` is a different type to `tuple[type[BaseException] | None, BaseException | None, TracebackType | None]`, so this is a semantic change

srittau: Indeed, but usually the latter was a shortcut used for the former. See the typical `__exit__` annotation that should properly be annotated using an overload.

AlexWaygood: Right, the latter is a supertype of the former

AlexWaygood (approved):

stdlib/cgitb.pyi
JelleZijlstra: I believe it's not strictly true that it's always either all-None or all-non-None, at least before 3.11. (Irit did some work in 3.11 that made it always true.) See here: https://github.com/python/cpython/blob/84c279b514141f608cf480905c87d48998e296d1/Python/ceval.c#L4474

stubs/pynput/pynput/_util.pyi
```diff
@@ -29,7 +29,7 @@ class AbstractListener(threading.Thread):
     _thread: threading.Thread  # undocumented
     _condition: threading.Condition  # undocumented
     _ready: bool  # undocumented
-    _queue: Queue[sys._OptExcInfo | None]  # undocumented
+    _queue: Queue[sys._OptExcInfo | None]  # undocumented  # TODO: use _typeshed.OptExcInfo after release of mypy 0.950
```
hauntsaninja, line 32: I believe the release branch of 0.950 has already been cut, so this comment may or may not be accurate

stdlib/cgitb.pyi
srittau: This could be a remnant of Python 2 string exceptions.

AlexWaygood: Maybe it's better to err on the side of safety here and define `_typeshed.OptExcInfo` as `tuple[type[BaseException] | None, BaseException | None, TracebackType | None]` instead of `tuple[type[BaseException], BaseException, TracebackType] | tuple[None, None, None]`

srittau: Primer output looks good and from a type safety perspective, I'd rather go with the more strict/correct annotation. Personally, I've not come across a situation where the type is `None` when the other fields are not, and allowing that could certainly introduce subtle bugs. It would also disagree with the documentation of both `sys.exc_info` and `contextmanager.__exit__`.

AlexWaygood: @JelleZijlstra, do you know of any real-world examples where some (but not all) elements of the tuple are `None`?

If not, I'm fine with merging this, as it's probably just a theoretical issue :)

JelleZijlstra: From https://github.com/python/cpython/issues/89874#issuecomment-1093936442, sounds like it could only happen with interesting use of the C API, so probably not worth worrying about.

AlexWaygood (approved):

PR #7649, opened 2022-04-16
dis: update for py311

hauntsaninja:


stdlib/dis.pyi
```diff
@@ -38,25 +38,60 @@ __all__ = [
...
+else:
+    class Instruction(NamedTuple):
+        opname: str
+        opcode: int
+        arg: int | None
+        argval: Any
+        argrepr: str
+        offset: int
+        starts_line: int | None
+        is_jump_target: bool
```
Akuli, line 69: This would be simpler with `if` statement inside the class:

```suggestion
class Instruction(NamedTuple):
    opname: str
    opcode: int
    arg: int | None
    argval: Any
    argrepr: str
    offset: int
    starts_line: int | None
    is_jump_target: bool
    if sys.version_info >= (3, 11):
        positions: Positions | None = ...
```

hauntsaninja: Unfortunately, at least mypy doesn't like that

Akuli (approved):

PR #7652, opened 2022-04-16
hashlib: update for py311

hauntsaninja:


stdlib/hashlib.pyi
```diff
@@ -143,3 +166,6 @@ class _BlakeHash(_Hash):
 
 blake2b = _BlakeHash
 blake2s = _BlakeHash
+
+if sys.version_info >= (3, 11):
+    def file_digest(__fileobj: BinaryIO, __digest: str | Callable[[], _Hash], *, _bufsize: int = ...) -> _Hash: ...
```
JelleZijlstra, line 171: Can we use a Protocol instead of BinaryIO?

Akuli (approved):

PR #2450, opened 2018-09-12
weakproxy fixes

srittau:
* The callback argument to ref() and proxy() is optional, per documentation
* proxy() returns a ProxyType or CallableProxyType

Closes #1627

PR #7653, opened 2022-04-17
weakref: use positional-only args

hauntsaninja:


hauntsaninja:
Separately, the proxy stuff here looked a little suspicious, since the comment sounds like the second overload should return ProxyType. It looks like this was discussed in https://github.com/python/typeshed/pull/2450. I think I'd be in favour of lying here and saying something like `def (object: t, ...) -> t`

PR #7661, opened 2022-04-18
Make several type aliases private

AlexWaygood:
None of these exist at runtime, except for `sqlite3.dbapi2.OptimizedUnicode`, which exists on <3.10. (All of these were picked up by https://github.com/python/mypy/pull/12608.)


stdlib/sqlite3/dbapi2.pyi
```diff
@@ -410,8 +410,6 @@ class InternalError(DatabaseError): ...
 class NotSupportedError(DatabaseError): ...
 class OperationalError(DatabaseError): ...
 
-OptimizedUnicode = str
```
AlexWaygood, line 413: This one doesn't seem to exist at runtime or be used in the stub

JelleZijlstra: It exists at runtime in 3.6-3.9: https://github.com/python/cpython/blob/3.6/Modules/_sqlite/module.c#L421, https://github.com/python/cpython/blob/636832fb2a4364ca2014aadecf4c35833eff4495/Modules/_sqlite/module.c#L439

Could you add it back with a version guard?

AlexWaygood: Yup, I'd just figured it out from the mypy failures :)

AlexWaygood: I'm not 100% sure whether it's better to do
```python
OptimizedUnicode = str
```
or
```python
OptimizedUnicode: TypeAlias = str
```
or
```python
from builtins import str as OptimizedUnicode
```

Thoughts?

JelleZijlstra (approved):

Issue #1339, opened 2017-05-24
Tests specifically for tricky, white-lie signatures


gnprice:
We've had some useful, long discussions about writing tests for stubs (#754, #917). Most of the discussion has been around having tests for the bulk of the functionality in stubs and the libraries they describe. I think we're all broadly in agreement that
* such comprehensive tests would be valuable; but
* just writing out a test for each function and class stubbed isn't helpful, because it amounts to repeating what we're already saying in the stubs; and
* a higher-tech approach to testing the bulk of the stubs' content won't be trivial to get to work with a good signal-to-noise ratio. (It'll be great if it *can* be done; some folks are working on it, and others are skeptical it can.)

This issue is for a narrower proposal: sometimes when a library's real signature is something complex or hard to express in the type system, our signature ends up pretty tricky or we end up telling sort of a white lie to the type-checkers. For these, even when looking right at the signature in one window and the library's docs (or an experiment in the REPL) in the other, it can be hard to tell whether they match. In these cases, I think a hand-written test would be useful -- probably looking a lot like a mypy unit test, with some code to be type-checked and an indication that certain lines should have errors and the rest shouldn't, and then we could e.g. run mypy and pytype against it.

@JukkaL suggested just this in one bullet point of [a comment on #754](https://github.com/python/typeshed/issues/754#issuecomment-275663086), and gave some examples:
> Add a test that exercises particularly complex signatures. Some recent examples where this could have been helpful are dict.get (a recent change to it broke it on mypy) and requests.api (there was a bad signature).

A new example today is #1331: `auto` doesn't really inherit from `IntFlag`, and we may find a new way of refining how we describe that interaction in the stubs. When we do it'll be useful to have a test for the functionality that that little lie in the stubs is intended to provide.

JelleZijlstra:
I like this idea. This would interact nicely with python/typing#436: the same test runner proposed there could run tests in typeshed.

matthiaskramm:
I'm ±0 on this.

Pro: Recently, we've had a number of pull requests with commit messages along the lines of
"Make `[some example code]` work with typeshed". Clearly, it would have been good to put `[some example code]` somewhere more meaningful.

Contra: It's a slippery slope. Testing leads to tests, which leads to more tests, which leads to complete test coverage, and look! We've just duplicated https://github.com/python/cpython/tree/master/Lib/test.

[...]

PR #6325, opened 2021-11-17
Further improve `pow`

AlexWaygood:
With the current stubs, mypy produces this output with the following code snippet:

```python
from fractions import Fraction

reveal_type(pow(2, 4))                            # Revealed type is "Any"
reveal_type(pow(5, -7))                           # Revealed type is "Any"
reveal_type(pow(Fraction(), 4, None))             # error: No overload variant of "pow" matches argument types "Fraction", "int", "None"
reveal_type(pow(Fraction(3, 7), complex(1, 8)))   # error: Cannot infer type argument 1 of "pow"
reveal_type(pow(complex(4, -8), Fraction(2, 3)))  # error: Cannot infer type argument 1 of "pow"
reveal_type(4 .__pow__(7, 4))                     # Revealed type is "Any"
reveal_type(4 .__pow__(6, None))                  # Revealed type is "Any"
```

With the patch that this PR proposes, the output changes to the following:

```python
from fractions import Fraction

reveal_type(pow(2, 4))                            # Revealed type is "builtins.int"
reveal_type(pow(5, -7))                           # Revealed type is "builtins.float"
reveal_type(pow(Fraction(), 4, None))             # Revealed type is "fractions.Fraction*"
reveal_type(pow(Fraction(3, 7), complex(1, 8)))   # Revealed type is "builtins.complex"
reveal_type(pow(complex(4, -8), Fraction(2, 3)))  # Revealed type is "builtins.complex"
reveal_type(4 .__pow__(7, 4))                     # Revealed type is "builtins.int"
reveal_type(4 .__pow__(6, None))                  # Revealed type is "builtins.int"
```

All other cases that I tested remain unchanged. A more exhaustive test script is attached, as are results of the test script with the current stubs, and with this PR applied:

-  [test_pow.txt](https://github.com/python/typeshed/files/7555705/test_pow.txt) (Uploaded here as a `.txt` file because GitHub won't let me upload a `.py` file, but this is really a Python file. Save it with a `.py` extension and run mypy on it to test `pow`.)
- [pow_results_prior_to_changes.txt](https://github.com/python/typeshed/files/7555710/pow_results_prior_to_changes.txt)
- [pow_results_with_changes.txt](https://github.com/python/typeshed/files/7555712/pow_results_with_changes.txt)
 
The exact command I ran to test this patch was:

```
[...]

PR #7663, opened 2022-04-19
Add regression tests for `builtins.pow` and `object.__reduce__`

AlexWaygood:
As discussed in https://github.com/python/typeshed/issues/1339#issuecomment-974169538 (among other places).

At the moment, this PR has only pyright running on the `test_cases` directory. We can potentially add mypy once there's a mypy release that supports `assert_type`.


JelleZijlstra (commented):
Thanks for setting this up! It's going to be very helpful. I'll send PRs later for my zip_longest cases from the other day.

test_cases/stdlib/test_builtins.py
```diff
@@ -0,0 +1,56 @@
...
+assert_type(Decimal("2.6").__pow__(5, None), Decimal)
+
+# These would ideally be more precise, but `Any` is acceptable
+# They have to be `Any` due to the fact that type-checkers can't distinguish between positive and negative numbers for the second argument to `pow()`
+#
+# int for positive 2nd-arg, float otherwise
+assert_type(pow(4, 65), Any)
+assert_type(pow(3, 57, None), Any)
+# float for positive 2nd-arg, complex otherwise
+assert_type(pow(4.7, 7.4), Any)
```
Akuli, line 52: The comment doesn't seem correct to me. If the *first* arg is positive, the result is a float. Negative first args give complex results.

```
>>> pow(1.2, 3.4)
1.858729691979481
>>> pow(-1.2, 3.4)
(-0.5743790627709717-1.7677569855883686j)
```

test_cases/stdlib/test_builtins.py
```diff
@@ -0,0 +1,56 @@
+from decimal import Decimal
```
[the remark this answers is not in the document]
AlexWaygood: Yeah, I think that might be better too. I'll make the change!

test_cases/README.md
```diff
@@ -0,0 +1,27 @@
...
+files. This is because the purpose of this folder is to test the implications
+of typeshed changes for end users. For example, the stub for `builtins.pow` is
+found in `stdlib/builtins.pyi`, and the regression tests for `pow` are found in
+`test_cases/stdlib/test_builtins.py`.
+
+100% test coverage for typeshed is neither necessary nor desirable, as it would
+lead to unnecessary code duplication. Moreover, typeshed has multiple other
+mechanisms for spotting errors in the stubs. As such, this directory should
+only contain tests for functions and classes which are known to have caused
+problems in the past, where the stubs are difficult to get right.
```
Akuli, line 18: While this is correct, it might need a tl;dr in bold. As discussed, it's difficult to convince some people that they don't need to write tests.

pyrightconfig.json
```diff
@@ -2,7 +2,8 @@
     "typeshedPath": ".",
     "include": [
         "stdlib",
-        "stubs"
+        "stubs",
+        "test_cases"
```
Akuli, line 6: - Do we really want this to be in both pyrightconfigs?
- Does mypy support `assert_type()` yet? If it does, we could instead point `mypy_primer` at typeshed's `test_cases`. That way we would know if a PR to mypy breaks the tests.

JelleZijlstra: Mypy supports assert_type() on master (I added it last week). I am hoping to convince Jukka to put it in the upcoming 0.950 release. We should make mypy run on this directory too as soon as 0.950 is released, or use master if we're feeling brave.

test_cases/stdlib/test_builtins.py
AlexWaygood: Oh, good catch.

hauntsaninja:
Can we keep the comments in builtins? This is a super common point of clarification and there is value in having the stubs be self-documenting.

Another suggestion could be to move this under `tests`. Also note we have the `@tests` directory for each third party stubs, which might be better since the equivalent tests for third party stubs could involve dependencies.

AlexWaygood:
> Can we keep the comments in builtins? This is a super common point of clarification and there is value in having the stubs be self-documenting.

Done.

> Another suggestion could be to move this under `tests`.

To me, it doesn't really feel like this stuff "fits in" with the other content in `tests` particularly well -- most of the content there is scripts for running tests, whereas this is _test data_. But I don't feel that strongly about it; happy to make the change if others agree. 🙂

>  Also note we have the `@tests` directory for each third party stubs, which might be better since the equivalent tests for third party stubs could involve dependencies.

Yeah, the third-party stubs would complicate things. Perhaps we could cross that bridge when we come to it?

test_cases/README.md
AlexWaygood: How's it look after b2e59b5?

test_cases/stdlib/test_builtins.py
AlexWaygood: Fixed in 9d3650f

test_cases/stdlib/test_builtins.py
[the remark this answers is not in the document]
AlexWaygood: 0e2165f

test_cases/stdlib/builtins/test_pow.py
```diff
@@ -0,0 +1,44 @@
...
+assert_type(pow(5.1, 4, None), float)
+assert_type(pow(complex(6), 6.2), complex)
+assert_type(pow(complex(9), 7.3, None), complex)
+assert_type(pow(Fraction(), 4, None), Fraction)
+assert_type(pow(Fraction(3, 7), complex(1, 8)), complex)
+assert_type(pow(complex(4, -8), Fraction(2, 3)), complex)
+assert_type(pow(Decimal("1.0"), Decimal("1.6")), Decimal)
+assert_type(pow(Decimal("1.0"), Decimal("1.0"), Decimal("1.0")), Decimal)
+assert_type(pow(Decimal("4.6"), 7, None), Decimal)
+assert_type((4).__pow__(7, 4), int)
```
JelleZijlstra, line 23: Why are you calling `__pow__` directly here? That doesn't seem idiomatic.

AlexWaygood: Honest answer: no idea. (I've had these test cases sitting around in my back pocket for [a long time](https://github.com/python/typeshed/pull/6325#issue-1055588877).) I'll change it :)

AlexWaygood: Hmm, actually, how do you test the `__pow__` overloads on `int` with three arguments with out accessing `__pow__` directly? If you do `pow(x, y, z)`, you can't be sure you're not testing the overloads for the builtin function `pow()` (rather than the overloads on `int`), and if you do `x ** y`, you can't test the three-argument form.

AlexWaygood: On third thought, it doesn't matter at all which method/function we're testing if the end result is correct. So you're right, these tests are weird; I'll change them.

pyrightconfig.json
AlexWaygood: Update: `assert_type` has been cherry-picked onto the 0.950 branch, so we should be able to point mypy_primer at this directory as soon as 0.950 is released.

JelleZijlstra: Hopefully we can just run mypy regularly, no need to use mypy-primer.

JelleZijlstra (approved):

test_cases/README.md
```diff
@@ -0,0 +1,24 @@
+## Regression tests for typeshed
+
+This directory contains regression tests for the stubs found elsewhere in the
+typeshed repo. Each file contains a number of test cases, all of which should
+pass a type checker without error.
```
srittau, line 5: In the long term we should probably also allow test that fail a type checker to be able to test the negative case.

srittau (commented):
Thanks for doing this!

test_cases/README.md
AlexWaygood: Yup. Once we're running mypy on this we can do that by `# type: ignor`ing (with specific error codes) lines that we want to fail and running mypy with `--warn-unused-ignores`.

PR #7670, opened 2022-04-21
Update PyMySQL constants

Nothing4You:
Still a lot other things to do but this is a start to bring it more up to date.
Based on PyMySQL 1.0.2.

What's the preferred format when e.g. a variable is assigned like [this](https://github.com/PyMySQL/PyMySQL/blob/v1.0.2/pymysql/constants/FIELD_TYPE.py#L30-L31)?

```py3
TINY = 1
CHAR = TINY
```

Should this be typed as
```py3
TINY: int
CHAR: int
```
or
```py3
TINY: int
CHAR = TINY
```
?

Example:
https://github.com/PyMySQL/PyMySQL/blob/v1.0.2/pymysql/constants/FIELD_TYPE.py#L30-L31

`scripts/create_baseline_stubs.py` generates the second variant.


srittau (approved):
Thanks!

Both variants are acceptable, it shouldn't make a difference in the case of primitives. The latter is closer to the implementation, the former is a bit more consistent/readable.

PR #7679, opened 2022-04-23
Buffer support for re

JelleZijlstra:


JelleZijlstra:
Also showed up in python/mypy#12661. Replaces #7158.

Akuli:
Maybe we should update the third-party `regex` stubs too.

hauntsaninja (approved):
Thanks! The homeassistant one looks like a false positive to me, probably a mypy bug, although we should be able to work around it if we wanted by using self types for Match.group

JelleZijlstra:
@Akuli sure, I'll do that in a separate PR.

@hauntsaninja yes that one is weird to me. Not sure where mypy gets bytes from in this code: https://github.com/home-assistant/core/blob/859dcb396a6f7a1712a188fc67060019aea24e9c/homeassistant/components/huawei_lte/device_tracker.py#L170. I don't understand how self types would help either.

(Also, the paroxython one looks like a true positive. I don't know why mypy didn't catch it before.)

hauntsaninja:
I filed https://github.com/python/mypy/issues/12666 for it

PR #5269, opened 2021-04-30
Fix `re.split(...)` return type

tusharsadhwani:
Example:

```python
re.split(r' (?=(x)*.)', 'a b') # ['a', None, 'b']
```

PR #7682, opened 2022-04-25
Loosen `__pow__` overloads a bit

tusharsadhwani:
This is done to alleviate this false positive:

```console
$ cat a.py              

def foo(x: float, y: float) -> float:
    return x ** y

$ mypy --strict a.py
a.py:2: error: Returning Any from function declared to return "float"
```

This change is akin to [this change](https://github.com/python/typeshed/pull/5269#issuecomment-830236965) in the `re` module.

Related: [mypy#7765](https://github.com/python/mypy/issues/7765)


AlexWaygood:
It looks like when mypy sees this signature:

```python
class int:
    def __pow__(self, __x: int, __modulo: None = ...) -> int | float | Any: ...
```

It automatically "simplifies" the signature to:

```python
class int:
    def __pow__(self, __x: int, __modulo: None = ...) -> float | Any: ...
```

This is because of the fact that `int` is an implicit subtype of `float` (according to PEP 484), and therefore `int` in the `int | float` union is redundant.

This then leads to the `rich` hit from the mypy_primer output. The source code's [here](https://github.com/Textualize/rich/blob/65227ba120a3ddcba72af4cd253ed2eabc053c87/rich/filesize.py#L44-L50)); here's the function that causes mypy to emit the error:

```python
def pick_unit_and_suffix(size: int, suffixes: List[str], base: int) -> Tuple[int, str]:
    """Pick a suffix and base for the given size."""
    for i, suffix in enumerate(suffixes):
        unit = base**i
        if size < unit * base:
            break
    return unit, suffix
```

 Will is pretty sure that `unit` is an `int` (I'm guessing he knows the two numbers being `pow`'d together will always be positive). But mypy sees two `int`s being `pow`'d together, and so it's pretty sure that this will create `float | Any`, because that's what the `int.__pow__` signature says. This therefore leads it to produce the somewhat baffling error message that `unit` is of type `float | Any`, when Will is annotating it with `int`.

I'm not sure there's a way of doing this without creating this kind of false-positive error, sadly ☹️

srittau:
As @AlexWaygood points out, I don't think we can improve the types any further at this point. Ideally, there'd be a way to define `int` ranges, or at least types for positive and non-negative integers, which would allow use to narrow the definitions. Thank you for the PR though, which allowed us to judge the impact of such a change!

Akuli:
I wonder if we could return `int | Any`.

AlexWaygood:
> I wonder if we could return `int | Any`.

Yeah, might be worth a shot.

PR #7684, opened 2022-04-26
sqlite3: Add sequence methods to Blob

JelleZijlstra:
python/cpython#91599


hauntsaninja (approved):
Not to be too crass, but I do enjoy the function name `blob_ass_subscript`

Akuli:
I find it weird that `blob[0]` returns `bytes`. With `bytes`, `bytearray` or `memoryview`, it would return a byte as `int`.

PR #7685, opened 2022-04-26
warnings: New arguments to catch_warnings()

JelleZijlstra:
python/cpython#91435


AlexWaygood:
Is it worth adding a comment to remind people to keep the signature of `warnings.catch_warnings` and `warnings.simplefilter` in sync?

JelleZijlstra:
Seems like overkill to me, and they don't match perfectly anyway (action is `| None` here).

AlexWaygood (approved):

PR #7688, opened 2022-04-26
inspect: Add positions attributes that will be new in 3.11

JelleZijlstra:
python/cpython#91531

### DIFF
--- a/stdlib/inspect.pyi
+++ b/stdlib/inspect.pyi
@@ -1,3 +1,4 @@
+import dis
 import enum
 import sys
 import types
@@ -465,6 +466,10 @@ class Traceback(NamedTuple):
     function: str
     code_context: list[str] | None
     index: int | None  # type: ignore[assignment]
+    if sys.version_info >= (3, 11):
+        @property
+        def positions(self) -> dis.Positions | None: ...
+        def __new__(cls: type[Self], filename: str, lineno: int, function: str, code_context: list[str] | None, index: int | None, *, positions=None) -> Self: ...
 
 class FrameInfo(NamedTuple):
     frame: FrameType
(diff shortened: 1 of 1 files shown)


hauntsaninja (approved):
Defining `__new__` looks cleaner than what I did for `dis`

AlexWaygood:
> Defining `__new__` looks cleaner than what I did for `dis`. Edit: it also happens to be what the runtime does

It does however break mypy 🙃

JelleZijlstra:
hm, now I get `stdlib/inspect.pyi:472: error: Cannot overwrite NamedTuple attribute "__new__"  [misc]`. Do we have a better technique or should we just type ignore it?

AlexWaygood:
> hm, now I get `stdlib/inspect.pyi:472: error: Cannot overwrite NamedTuple attribute "__new__" [misc]`. Do we have a better technique or should we just type ignore it?

We _could_ do something similar to what I did with `structseq` classes in other modules. But I don't think there's any need, because (unlike with `structseq`s), there's no attributes on these classes that are `ClassVar`s. And mypy can infer the types fine, even if it doesn't _think_ it can, so I say: just `type: ignore` it away.

https://mypy-play.net/?mypy=latest&python=3.10&gist=2bdfa6f9cdc2bdd07c5d472bf4124ea5

AlexWaygood:
On second thoughts, why don't we just mimic the runtime implementation exactly? It seems to work pretty well for `tokenize.TokenInfo`:

https://github.com/python/typeshed/blob/7cbb579a4473e3fbd87d50f7b69bbf47a8149b00/stdlib/tokenize.pyi#L324-L333

and it would mean we could override `__new__` without a `# type: ignore`. It also means that we can annotate the `positions` attributes as settable instance variables, which they are at runtime in 3.11, even though all other attributes on `Traceback` and `FrameInfo` are read-only.

So in the 3.11 branch we would have:

```python
    class _Traceback(NamedTuple):
        filename: str
        lineno: int
        function: str
        code_context: list[str] | None
        index: int | None  # type: ignore[assignment]

    class Traceback(_Traceback):
        positions: dis.Positions | None
        def __new__(
            cls: type[Self],
            filename: str,
            lineno: int,
            function: str,
            code_context: list[str] | None,
            index: int | None,
            *,
            positions: dis.Positions | None = ...,
        ) -> Self: ...

    class _FrameInfo(NamedTuple):
        frame: FrameType
        filename: str
        lineno: int
        function: str
        code_context: list[str] | None
        index: int | None  # type: ignore[assignment]

    class FrameInfo(_FrameInfo):
        positions: dis.Positions | None
        def __new__(
            cls: type[Self],
            frame: FrameType,
            filename: str,
            lineno: int,
            function: str,
            code_context: list[str] | None,
            index: int | None,
            *,
            positions: dis.Positions | None = ...,
        ) -> Self: ...
```

AlexWaygood (approved):

PR #7692, opened 2022-04-26
multiprocessing: shutdown_timeout param to BaseManager

JelleZijlstra:
python/cpython#32112


stdlib/multiprocessing/managers.pyi
```diff
@@ -76,9 +76,21 @@ class Server:
...
+    else:
+        def __init__(
+            self,
+            address: Any | None = ...,
+            authkey: bytes | None = ...,
+            serializer: str = ...,
+            ctx: BaseContext | None = ...,
+            *,
+            shutdown_timeout: float = ...,
+        ) -> None: ...
```
AlexWaygood, line 92: Shouldn't these be the other way around? 🤔

JelleZijlstra: Indeed, fixed. I think that's a sign I've done enough of these for today.

AlexWaygood: Are you using stubtest to find these, out of interest? :)

JelleZijlstra: No, just going over `git log` in CPython

AlexWaygood (approved):

PR #7694, opened 2022-04-26
`__wrapped__` is read-only on `classmethod`s and `staticmethod`s

AlexWaygood:


JelleZijlstra (approved):
Confirmed:

```c
    {"__wrapped__", T_OBJECT, offsetof(classmethod, cm_callable), READONLY},
```

Issue #7681, opened 2022-04-25
XMLPullParser.feed() accept strings too


Prometheus3375:
From an example in Python [docs](https://docs.python.org/3/library/xml.etree.elementtree.html#pull-api-for-non-blocking-parsing):
```python
>>> parser = ET.XMLPullParser(['start', 'end'])
>>> parser.feed('<mytag>sometext')
>>> list(parser.read_events())
[('start', <Element 'mytag' at 0x7fa66db2be58>)]
>>> parser.feed(' more text</mytag>')
>>> for event, elem in parser.read_events():
...     print(event)
...     print(elem.tag, 'text=', elem.text)
...
end
```
As you can see, plain strings are passed to `feed()` method. Thus, it supports strings. I also checked it myself on Python 3.10.

Typeshed [says](https://github.com/python/typeshed/blob/master/stdlib/xml/etree/ElementTree.pyi#L322) that `feed()` supports only `bytes`. This is a mistake.

srittau:
Indeed, `XMLPullparser.feed()` just delegates to `XMLParser.feed()`, which already accepts `str | bytes`. PR welcome!

PR #7695, opened 2022-04-26
Add str type to XMLPullParser.feed argument

Prometheus3375:
`XMLPullParser.feed()` method supports both bytes and strings


Prometheus3375:
Issue #7681

srittau (approved):
Thanks!

PR #7696, opened 2022-04-26
Additional typing hints for requests

milanboers:
- Fixes issue where positional parameters were accepted for keyword-only arguments.
- Complete typing hints for send, request and HTTP-method functions


milanboers:
Thanks @AlexWaygood . Was struggling with this mypy bug https://github.com/python/mypy/issues/11098 bug seems that you found a workaround 😁

AlexWaygood:
> Thanks @AlexWaygood . Was struggling with this mypy bug [python/mypy#11098](https://github.com/python/mypy/issues/11098) but seems that you found a workaround 😁

No worries! Figured it was easier just to push to your branch in this case, given that the errors were so minor 🙂

Thanks for the PR, it looks extremely well put together for a first-time contribution! I'm taking a look now.

stubs/requests/requests/sessions.pyi
```diff
@@ -45,8 +45,15 @@ class SessionRedirectMixin:
...
-
+_Auth: TypeAlias = Union[tuple[str, str], _auth.AuthBase, Callable[[PreparedRequest], PreparedRequest]]
+_Cert: TypeAlias = Union[str, tuple[str, str]]
+_Data: TypeAlias = str | bytes | Mapping[str, Any] | Iterable[tuple[str, str | None]] | IO[Any]
+_Files: TypeAlias = (
+    MutableMapping[str, IO[Any]]
+    | MutableMapping[str, tuple[str, IO[Any]]]
+    | MutableMapping[str, tuple[str, IO[Any], str]]
+    | MutableMapping[str, tuple[str, IO[Any], str, _TextMapping]]
+)
```
AlexWaygood, line 56: Would you mind possibly providing some explanation as to how you figured out some of these more complex types? It might be quicker than us trawling through the `requests` codebase and repeating the steps you've already done :)

Also, just to check, were you looking at the `requests` `main` branch when you prepared this PR, or the 2.27 tag? If the former, note that the `main` branch is for `requests` 3.0, which hasn't yet been released -- these stubs are for `requests` 2.27. Some of the types look like they might be slightly different between 2.27 and `main`.

milanboers: Most of these types were already there, defined on `Session.request`: https://github.com/python/typeshed/blob/2299c43860c7115fabfd14ad854144e2ef20e132/stubs/requests/requests/sessions.pyi#L83
I only extracted them to a TypeAlias to reuse in get, post, put etc. (these all internally call `send`) and made some minor changes.

I didn't check them thoroughly, but it looks like they're mostly correct, especially if you follow [request's documentation](https://docs.python-requests.org/en/latest/api/).

Request's code seems to accept more than the documentation though, not sure how far we want to go in accepting arguments that would technically work, but are not documented.
For example:
- In `files`, the documentation specifically mentions to use 2/3/4-tuples, but passing lists also works.
- `allow_redirects` is documented as optional and `True` by default. Made the type `bool` now, but I think the code will accept `None`, and make it behave like `False`.

AlexWaygood: Okay great, thanks, that's all really helpful!

stubs/requests/requests/api.pyi
```diff
@@ -1,142 +1,150 @@
 from typing import Any
 
 from .models import Response
-from .sessions import _Data, _Params
+from .sessions import RequestsCookieJar, _Auth, _Cert, _Data, _Files, _HooksInput, _Params, _TextMapping, _Timeout, _Verify
 
 def request(
-    method: str | bytes,
+    method: str,
```
AlexWaygood, line 7: Why change this? Seems like `bytes` works fine, even if it is a little weird:

```python
>>> import requests
>>> requests.request(b'GET', 'https://www.bbc.com/news')
<Response [200]>
```

stubs/requests/requests/sessions.pyi
```diff
@@ -45,8 +45,15 @@ class SessionRedirectMixin:
     def rebuild_proxies(self, prepared_request, proxies): ...
     def should_strip_auth(self, old_url, new_url): ...
 
-_Data: TypeAlias = str | bytes | Mapping[str, Any] | Iterable[tuple[str, str | None]] | IO[Any] | None
-
+_Auth: TypeAlias = Union[tuple[str, str], _auth.AuthBase, Callable[[PreparedRequest], PreparedRequest]]
+_Cert: TypeAlias = Union[str, tuple[str, str]]
+_Data: TypeAlias = str | bytes | Mapping[str, Any] | Iterable[tuple[str, str | None]] | IO[Any]
```
AlexWaygood, line 50:
```suggestion
_Data: TypeAlias = str | bytes | Mapping[str, Any] | Iterable[tuple[str, str | None]] | IO[Any]
_Auth: TypeAlias = Union[tuple[str, str], _auth.AuthBase, Callable[[PreparedRequest], PreparedRequest]]
_Cert: TypeAlias = Union[str, tuple[str, str]]
```

AlexWaygood (commented):
Everything looks good to me, other than my question about `method` changing from `str | bytes` to `str`!

stubs/requests/requests/sessions.pyi
AlexWaygood: The diff for this PR is quite big, so I support keeping the annotations as they are for now and factoring them out into `TypeAlias`es, as you've done in this PR. We can always improve the types in a separate PR later on 👍

JelleZijlstra: For boolean parameters specifically, it's almost always the case that any object that supports `__bool__` (in other words, almost any object anywhere) can be passed and works fine at runtime. But that's not very useful for typing, so we do want to annotate such parameters as `bool` unless there's some specific reason (e.g., documentation) to allow a broader type.

stubs/requests/requests/api.pyi
milanboers: Ah, you're right! Changed it back.

The implementation for `url` has an explicit check+conversion for `bytes`, which is not there for `method`. They also call `.upper()` on it, but apparently that also works for `bytes` :-)

Akuli, line 7:
```suggestion
    method: str | bytes,
```

stubs/requests/requests/sessions.pyi
```diff
@@ -45,8 +45,15 @@ class SessionRedirectMixin:
     def rebuild_proxies(self, prepared_request, proxies): ...
     def should_strip_auth(self, old_url, new_url): ...
 
-_Data: TypeAlias = str | bytes | Mapping[str, Any] | Iterable[tuple[str, str | None]] | IO[Any] | None
-
+_Data: TypeAlias = str | bytes | Mapping[str, Any] | Iterable[tuple[str, str | None]] | IO[Any]
+_Auth: TypeAlias = Union[tuple[str, str], _auth.AuthBase, Callable[[PreparedRequest], PreparedRequest]]
+_Cert: TypeAlias = Union[str, tuple[str, str]]
```
Akuli, line 50: Should this accept bytes? `adapters.pyi` currently does this:

https://github.com/python/typeshed/blob/54c8f45869a67ad2fc73a0d59ce316dd723a2e94/stubs/requests/requests/adapters.pyi#L47

Looking at `sessions.py`, it seems to just pass everything to the adapter, so if adapters work with bytes, then we should allow bytes here too. In `adapters.py`, the certs are validated with `os.path.exists`, which works with bytes and strings.

In general, `adapters.pyi` seems to be a bit out of sync with `api.pyi` and `sessions.pyi`, but I don't think fixing it belongs to this PR.

milanboers: Looks like it would work for `requests`, but in the end it ends up at urllib3's `HTTPSConnectionPool`. Not sure if that accepts & keeps accepting `bytes`.  Their current version doesn't have type stubs, but their [upcoming major version](https://urllib3.readthedocs.io/en/latest/reference/urllib3.connectionpool.html#urllib3.HTTPSConnectionPool) requires `str`.

I can make a follow-up PR later to work away some of the inconsistencies and take a look at `models`.

Akuli (approved):
Thanks!

PR #7700, opened 2022-04-27
pytype now supports Python 3.10

srittau:


.github/workflows/tests.yml
```diff
@@ -44,8 +44,6 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - uses: actions/setup-python@v2
-        with:
-          python-version: 3.9
```
Akuli, line 48: Would it be better to specify 3.10 explicitly? That way nothing will break when Python 3.11 becomes the default in github actions.

Akuli (approved):

PR #7702, opened 2022-04-27
Add various missing generic arguments

srittau:


stubs/Markdown/markdown/inlinepatterns.pyi
```diff
@@ -94,7 +94,7 @@ class LinkInlineProcessor(InlineProcessor):
 class ImageInlineProcessor(LinkInlineProcessor): ...
 
 class ReferenceInlineProcessor(LinkInlineProcessor):
-    NEWLINE_CLEANUP_RE: typing.Pattern
+    NEWLINE_CLEANUP_RE: typing.Pattern[str]
```
Akuli, line 97:
```suggestion
    NEWLINE_CLEANUP_RE: ClassVar[typing.Pattern[str]]
```

stubs/Pillow/PIL/PdfParser.pyi
```diff
@@ -47,10 +47,10 @@ class PdfName:
 class PdfArray(list[Any]):
     def __bytes__(self): ...
 
-class PdfDict(collections.UserDict):
+class PdfDict(collections.UserDict[Any, Any]):
```
Akuli, line 50: Maybe:

```suggestion
class PdfDict(collections.UserDict[bytes, Any]):
```

stubs/humanfriendly/humanfriendly/case.pyi
```diff
@@ -1,9 +1,9 @@
-import collections
+from collections import OrderedDict
 from typing import Any
 
 from humanfriendly.compat import unicode
 
-class CaseInsensitiveDict(collections.OrderedDict):
+class CaseInsensitiveDict(OrderedDict[Any, Any]):
```
Akuli, line 6: Maybe `CaseInsensitiveDict` should be a generic class? It seems to treat non-string keys just like `OrderedDict`.

stubs/Pillow/PIL/PdfParser.pyi
srittau: I was very confused about this class, possibly because I haven't worked with `UserDict` before and sub-classing `dict` works differently. But can't the Key either be a `bytes` or the literal string `'data'`?

Akuli: With the string `"data"`, it calls `UserDict.__setattr__` which sets `self.data` to the value. It has nothing to do with `self[...]`.

srittau: Still confused, but I'll do what you say. :)

srittau: I mixed up `__setitem__` and `__setattr__`, makes sense now.

Akuli (approved):

PR #7703, opened 2022-04-27
Drop Python 2 support in third-party stubs

srittau:
Part of #7367


tests/mypy_test.py
```diff
@@ -336,7 +321,7 @@ def main():
             if not is_probably_stubs_folder(distribution, distribution_path):
                 continue
 
-            if not is_supported(distribution_path, major):
+            if major == 2:
```
Akuli, line 324: It would be cleaner to put the whole loop inside `if major != 2`. This way it won't print `Running mypy ...` when it isn't actually going to run mypy.

tests/mypy_test.py
```diff
@@ -93,20 +92,6 @@ def parse_version(v_str):
...
-def is_supported(distribution_path: Path, major: int) -> bool:
-    data = dict(tomli.loads((distribution_path / "METADATA.toml").read_text()))
-    if major == 2:
-        # Python 2 is not supported by default.
-        return bool(data.get("python2", False))
-    # Python 3 is supported by default.
-    return has_py3_stubs(distribution_path)
-
-
-# Keep this in sync with stubtest_third_party.py
```
Akuli, line 105: We should clean up `stubtest_third_party.py` as well.

Akuli (approved):

PR #7713, opened 2022-04-27
Remove Python 2 support from toml

srittau:


stubs/toml/toml/decoder.pyi
```diff
@@ -1,19 +1,13 @@
...
-if sys.version_info >= (3, 0):
-    from pathlib import PurePath
-
-    FNFError = FileNotFoundError
-    _PathLike: TypeAlias = str | bytes | PurePath
-else:
-    FNFError = IOError
-    _PathLike: TypeAlias = Text
-
+FNFError = FileNotFoundError
```
Akuli, line 10:
```suggestion
FNFError: TypeAlias = FileNotFoundError
```

srittau: This isn't a type alias, just a regular class alias, isn't it?

Akuli: I think type checkers will treat it the same either way.

PR #7715, opened 2022-04-27
Drop Python 2 support from python-dateutil

srittau:


stubs/python-dateutil/dateutil/tz/tz.pyi
```diff
@@ -1,11 +1,11 @@
 import datetime
-from typing import IO, Any, Text, TypeVar
+from typing import IO, Any, TypeVar
 from typing_extensions import TypeAlias
 
 from ..relativedelta import relativedelta
 from ._common import _tzinfo as _tzinfo, enfold as enfold, tzname_in_python2 as tzname_in_python2, tzrangebase as tzrangebase
 
-_FileObj: TypeAlias = str | Text | IO[str] | IO[Text]
+_FileObj: TypeAlias = str | IO[str]
```
Akuli, line 8: `IO[str]` doesn't seem correct. If fileobj is given as a string, it does `open(fileobj, 'rb')`. It also does things like `fileobj.read(4).decode()` below.

srittau: It seems that some streams are expected to be binary, while others are expected to be text streams. This needs a closer look (and protocols). I've used `IO[Any]` for now.

Akuli (approved):

PR #7717, opened 2022-04-27
dateutil.tz.tz: Replace IO with protocols

srittau:


srittau:
In theory, the reader protocol `__enter__` methods could return a different object and we could use two protocols, but I decided to ignore that case as it's unlikely to be used in practice.

stubs/python-dateutil/dateutil/tz/tz.pyi
```diff
@@ -84,12 +93,21 @@ class tzrange(tzrangebase):
...
 class tzstr(tzrange):
     hasdst: bool
-    def __init__(self, s: bytes | _FileObj, posix_offset: bool = ...) -> None: ...
+    def __init__(self, s: str, posix_offset: bool = ...) -> None: ...
     @classmethod
     def instance(cls, name, offset) -> tzoffset: ...
 
+class _ICalReader(Protocol):
+    # optional attribute:
+    # read: str
```
Akuli, line 102:
```suggestion
    # name: str
```

stubs/python-dateutil/dateutil/tz/tz.pyi
```diff
@@ -84,12 +93,21 @@ class tzrange(tzrangebase):
...
+    def __enter__(self: Self) -> Self: ...
+    def __exit__(
+        self, exc_type: type[BaseException] | None, exc_value: BaseException | None, traceback: TracebackType | None
+    ) -> bool | None: ...
```
Akuli, line 106: These aren't needed, because the file object is wrapped in `_nullcontext`. It doesn't close files passed in, it only closes the file if a string was passed in and it opened the file.

Akuli (approved):

PR #7718, opened 2022-04-27
asyncio: expose WriteTransport.get_write_buffer_limits on all Python versions

### LINKED ISSUE #5779: Bad docs, or bad types for asyncio.WriteTransport?
I have to admit I am not fully certain of this one, there's a few layers of abstraction to sift through.

Typeshed declares this:

```python
class _FlowControlMixin(Transport):
    def __init__(self, extra: Optional[Mapping[Any, Any]] = ..., loop: Optional[AbstractEventLoop] = ...) -> None: ...
    def get_write_buffer_limits(self) -> Tuple[int, int]: ...
```

Which does seem true: CPython does indeed define get_write_buffer_limits() as part of that `_FlowControlMixin` class. Where I lose the plot, however, is the fact that https://docs.python.org/3/library/asyncio-protocol.html#write-only-transports seems to imply that *all* `asyncio.WriteTransport` types will offer this method. In practice, I suppose that isn't strictly true, since the CPython implementation declares the class like this:

```python
class WriteTransport(BaseTransport):
    ...
```

And does not in fact guarantee the presence of this method. It turns out that the *actual* type of the transport beneath the StreamWriter that I have is ... `asyncio.selector_events._SelectorSocketTransport`, which inherits from `_SelectorTransport`, which in turn inherits from `transports._FlowControlMixin` and `transports.Transport`, the latter of which inherits from *both* `ReadTransport` and `WriteTransport`.

Phew, OK.

I suppose this means the docs are "wrong" in the sense that `WriteTransport` does not necessarily guarantee the presence of this method, but "in practice" all implementations of `WriteTransport` use the `_FlowControlMixin` that offers it.

Does the WriteTransport class just need a `NotImplementedError` stub for `get_write_buffer_limits()` ...? or is it intentional that this method isn't *required* and the docs are wrong?

Akuli:
https://github.com/python/typeshed/issues/5779#issuecomment-1110984990 shows code that uses this method. It works on 3.6, 3.7 and 3.8, but doesn't type-check without this PR on those versions.

Fixes #5779


JelleZijlstra (approved):
https://docs.python.org/3/library/asyncio-protocol.html#asyncio.WriteTransport.get_write_buffer_limits says it was new in 3.4.2, not sure where we got 3.9 from.

AlexWaygood:
> https://docs.python.org/3/library/asyncio-protocol.html#asyncio.WriteTransport.get_write_buffer_limits says it was new in 3.4.2, not sure where we got 3.9 from.

It was a weird one where somebody noticed on BPO a few weeks ago that these had been documented forever, but not actually implemented in the code. So they were added the other week, and backported to 3.9.

PR #7728, opened 2022-04-27
requests: allow str and bytes for fileobj in files parameter

### LINKED ISSUE #7724: requests: Allow files={"example": ("filename", bytes)}
According to https://github.com/psf/requests/blob/2d5517682b3b38547634d153cea43d48fbc8cdb5/requests/models.py#L170 It should be possible to provide `files={"fieldname": ("filename", [bytes-object])}`, but since the last update of mypy this fails with
```
packet_analyser/daemon/uploader.py:78: error: Argument "files" to "post" has incompatible type "Dict[str, Tuple[str, bytes]]"; expected "Optional[Union[MutableMapping[str, IO[Any]], MutableMapping[str, Tuple[Optional[str], IO[Any]]], MutableMapping[str, Tuple[Optional[str], IO[Any], str]], MutableMapping[str, Tuple[Optional[str], IO[Any], str, MutableMapping[str, str]]]]]"  [arg-type]
```

The issue seems to be in https://github.com/python/typeshed/blob/63fb9af7438613dd2bfebe07911e49022ce78f57/stubs/requests/requests/sessions.pyi#L51-L56

Can this be fixed?

milanboers:
Some people are having issues with this one, see https://github.com/python/typeshed/issues/7724

[Documentation](https://docs.python-requests.org/en/latest/api/) says to pass a "file-like-object". [Implementation](https://github.com/psf/requests/blob/v2.27.1/requests/models.py#L158) allows for `str`, `bytes`, `bytearray` or anything that implements `.read()`. 
Left `bytearray` out for now because urllib3 documents `str` and `bytes` as valid in [RequestField](https://urllib3.readthedocs.io/en/latest/reference/urllib3.fields.html) in the next major version.

Resolves #7724

### DIFF
--- a/stubs/requests/requests/sessions.pyi
+++ b/stubs/requests/requests/sessions.pyi
@@ -49,10 +49,10 @@ _Data: TypeAlias = str | bytes | Mapping[str, Any] | Iterable[tuple[str, str | N
 _Auth: TypeAlias = Union[tuple[str, str], _auth.AuthBase, Callable[[PreparedRequest], PreparedRequest]]
 _Cert: TypeAlias = Union[str, tuple[str, str]]
 _Files: TypeAlias = (
-    MutableMapping[str, IO[Any]]
-    | MutableMapping[str, tuple[str | None, IO[Any]]]
-    | MutableMapping[str, tuple[str | None, IO[Any], str]]
-    | MutableMapping[str, tuple[str | None, IO[Any], str, _TextMapping]]
+    MutableMapping[str, IO[Any] | str | bytes]
+    | MutableMapping[str, tuple[str | None, IO[Any] | str | bytes]]
+    | MutableMapping[str, tuple[str | None, IO[Any] | str | bytes, str]]
+    | MutableMapping[str, tuple[str | None, IO[Any] | str | bytes, str, _TextMapping]]
 )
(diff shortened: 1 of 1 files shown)


AlexWaygood (commented):
The implementation looks like it can be `str`, `bytes`, or anything with a `.read()` method: https://github.com/psf/requests/blob/31a89d9c8463c3394ca00f408f4b86d814421a09/requests/models.py#L160-L161

Rather than using `IO[Any]`, we could make it even more permissive, and just use `_typeshed.SupportsRead`:

https://github.com/python/typeshed/blob/1f2a970b940536ce45f2b3a9ee7e3e4855d3601f/stdlib/_typeshed/__init__.pyi#L180-L181

milanboers:
> 

Good one, done! Kept the `IO` in `_Data` as-is, because it also needs `.tell()` there.

AlexWaygood:
I've confirmed that this annotation will allow objects with a `.read()` returning `bytes`, and objects with a `.read()` returning `str`: https://mypy-play.net/?mypy=latest&python=3.10&gist=c5afaf626edee4068ca05bc806f33f64

AlexWaygood (approved):
Thanks!

PR #7730, opened 2022-04-28
jack: Fix MidiPort properties

### LINKED ISSUE #7729: Stubtest failed on Thu Apr 28 2022

JelleZijlstra:
Fixes part of #7729 

https://github.com/spatialaudio/jackclient-python/blob/26b648a36143b1e3db6e6fc827ca927b0c93cbec/src/jack.py#L1950


stubs/JACK-Client/jack/__init__.pyi
```diff
@@ -180,8 +180,10 @@ class Port:
     def request_monitor(self, onoff: bool) -> None: ...
 
 class MidiPort(Port):
-    is_audio: bool = ...
-    is_midi: bool = ...
+    @property
+    def is_audio(self) -> bool: ...
+    @property
+    def is_midi(self) -> bool: ...
```
AlexWaygood, line 186:
```suggestion
    def is_audio(self) -> Literal[False]: ...
    @property
    def is_midi(self) -> Literal[True]: ...
```

Taking a look at [the docstrings](https://github.com/spatialaudio/jackclient-python/blob/26b648a36143b1e3db6e6fc827ca927b0c93cbec/src/jack.py#L1968-L1969):

JelleZijlstra: I was hesitant because of #7258 but it's probably fine.

AlexWaygood (approved):

PR #7732, opened 2022-04-28
requests: allow non-mutable Mapping for files/hooks parameters

milanboers:
Another one 😒 still an issue with #7724 . Problem is `MutableMapping` is invariant, so the union type will actually never match.
Changed it to `Mapping` (which is covariant in the value type) now. It doesn't seem to be mutated anywhere, I also passed `frozendict` to try. Also tried assigning a `dict[str, BufferedReader]` to it as in the issue, which passes the type check now.

Same problem would occur in `_Hooks`, so changed that one too. Kept `_TextMapping` as-is because it requires `str` (would still fail on a subtype of `str`).

### DIFF
--- a/stubs/requests/requests/sessions.pyi
+++ b/stubs/requests/requests/sessions.pyi
@@ -49,14 +49,14 @@ _Data: TypeAlias = str | bytes | Mapping[str, Any] | Iterable[tuple[str, str | N
 _Auth: TypeAlias = Union[tuple[str, str], _auth.AuthBase, Callable[[PreparedRequest], PreparedRequest]]
 _Cert: TypeAlias = Union[str, tuple[str, str]]
 _Files: TypeAlias = (
-    MutableMapping[str, SupportsRead[str | bytes] | str | bytes]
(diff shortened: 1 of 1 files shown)


AlexWaygood:
Thanks for plugging away on this!

Akuli:
Typeshed doesn't do much tests in the usual sense. There has been a lot of discussion about this, but in short, we don't want to duplicate information that is already specified in the stubs.

For standard library stubs, we have mypy-primer, which runs mypy across many open-source projects on each new PR and posts a diff of mypy outputs as a comment. It would work very well for problems like this. Unfortunately it is currently not being used for third-party stubs, see https://github.com/hauntsaninja/mypy_primer/issues/7.

JelleZijlstra:
Though we *do* have a few `assert_type()`-based tests now, and we could start adding more.

brandonchinn178:
FYI This breaks `session.hooks['response'] = ...` #7776

Akuli:
> Unfortunately it is currently not being used for third-party stubs

Now it is, and regressions like this are much less likely to happen: https://github.com/hauntsaninja/mypy_primer/pull/31

Issue #6307, opened 2021-11-16
mypy --warn-unused-ignores doesn't work


srittau:
I'm unsure about the reason, but when I locally run `mypy --warn-unused-ignores stdlib/builtins.pyi` (or any other file in `stdlib` or `stubs`) where I have manually added an unnecessary `# type: ignore`, I get no warnings. mypy is current in my venv (0.910) and I deleted `.mypy_cache` to make sure. The same happened when I ran `./tests/mypy_test.py --warn-unused-ignores`, which is how I noticed.

Does anyone else have this problem?

srittau:
And I just double checked: If I create a file `foo.pyi` in my home directory:

```python
def foo() -> None: ...  # type: ignore
```

and run `mypy ~/foo.pyi --warn-unused-ignores` from the typeshed directory, I get a warning. As soon as I move the file to the top-level typeshed directory, I don't get a warning.

srittau:
And another data point: When I run `mypy foo.pyi --warn-unused-ignores` in the typeshed dir, then move the file to my home directory, then run `mypy ~/foo.pyi --warn-unused-ignores`, I don't a warning unless I delete `.mypy_cache`.

Akuli:
Happens to me too.

```
(env) akuli@akuli-desktop:~/typeshed$ mypy --warn-unused-ignores stdlib/builtins.pyi 
Success: no issues found in 1 source file
(env) akuli@akuli-desktop:~/typeshed$ rm -rf .mypy_cache/
(env) akuli@akuli-desktop:~/typeshed$ mypy --warn-unused-ignores stdlib/builtins.pyi 
Success: no issues found in 1 source file
(env) akuli@akuli-desktop:~/typeshed$ git diff
diff --git a/stdlib/builtins.pyi b/stdlib/builtins.pyi
index 6bc58d2a..52fa409d 100644
--- a/stdlib/builtins.pyi
+++ b/stdlib/builtins.pyi
@@ -269,10 +269,10 @@ class float:
     def is_integer(self) -> bool: ...
     @classmethod
     def fromhex(cls, __s: str) -> float: ...
-    @property
-    def real(self) -> float: ...
-    @property
-    def imag(self) -> float: ...
+    @property  # type: ignore
+    def real(self) -> float: ...  # type: ignore
+    @property  # type: ignore  # type: ignore
+    def imag(self) -> float: ...  # type: ignore
     def conjugate(self) -> float: ...
[...]

PR #7738, opened 2022-04-28
Cleanup `mypy_test.py`

AlexWaygood:
- Remove the `--warn-unused-ignores` command-line option (it doesn't work, see #6307).
- Add more type hints.

I am planning a second PR after this one, to add new functionality to `mypy_test.py` so that it runs on files in the `tests` and `test_cases` directories.

### DIFF
--- a/tests/mypy_test.py
+++ b/tests/mypy_test.py
@@ -16,7 +16,12 @@ import re
 import sys
 import tempfile
 from pathlib import Path
-from typing import Dict, NamedTuple
+from typing import TYPE_CHECKING, Dict, List, NamedTuple, Set, Tuple
+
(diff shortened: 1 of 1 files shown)


tests/mypy_test.py
```diff
@@ -76,7 +74,7 @@ def parse_versions(fname):
                 continue
             m = _VERSION_LINE_RE.match(line)
             assert m, "invalid VERSIONS line: " + line
-            mod = m.group(1)
+            mod: str = m.group(1)
```
JelleZijlstra, line 77: Interesting, why is this necessary?

JelleZijlstra (approved):
One question, but feel free to merge

PR #7740, opened 2022-04-29
Add stubs for huey and django-huey

henribru:
This adds type stubs for https://github.com/coleifer/huey and https://github.com/gaiacoop/django-huey. These are _mostly_ partial, directly from stubgen. The main changes are in `stubs/huey/huey/api.pyi`, where I've added some generics and made the necessary changes to pass through those generics. The purpose here is to preserve the argument and return types of user-defined functions that are wrapped in Huey's `task` decorator, because otherwise the downstream consumer of Huey loses all type safety for those functions. I think I've covered the surfaces needed to accomplish this for the most common usecases. I feel like this has good value even without typing the rest of Huey's API because that has less of a cascading effect on the type safety.

In addition various functions in `stubs/huey/huey/contrib/djhuey/__init__.pyi` and `stubs/django-huey/django_huey/__init__.pyi` with the same API as methods in `huey.Huey` have been made to match that class.


AlexWaygood (commented):
Consider splitting this into multiple PRs. That might make it easier for you to fix the CI failures, and easier for us to review once you've done so :)

You can add stubs incrementally by using module `__getattr__`, which is a pattern is supported by type checkers. E.g. initially you could do a PR adding just three files, `huey/__init__.pyi`, `huey/api.pyi` and `huey/exceptions.pyi`:

```python
###
# huey.__init__.pyi
###
from huey.api import (
    BlackHoleHuey as BlackHoleHuey,
    FileHuey as FileHuey,
    Huey as Huey,
    MemoryHuey as MemoryHuey,
    PriorityRedisExpireHuey as PriorityRedisExpireHuey,
    PriorityRedisHuey as PriorityRedisHuey,
    RedisExpireHuey as RedisExpireHuey,
    RedisHuey as RedisHuey,
    SqliteHuey as SqliteHuey,
    crontab as crontab,
)
from huey.exceptions import CancelExecution as CancelExecution, RetryTask as RetryTask

###
# huey/api.pyi
###
from typing import Any
def __getattr__(name: str) -> Any: ...

###
# huey/exceptions.pyi
###
from typing import Any
def __getattr__(name: str) -> Any: ...
```

And then you could incrementally add stubs for more and more modules in future PRs.

AlexWaygood:
Although mind you, there aren't that many CI failures.

AlexWaygood:
I think the mypy CI failure is because you need to specify `requires=[types-huey]` in the `METADATA.toml` file for `django-huey`. E.g. this is how we do it for our `requests` stubs, which have our `urllib` stubs as a dependency:

https://github.com/python/typeshed/blob/4e946b74bf18dc932caf41d0257957d85bbabe91/stubs/requests/METADATA.toml#L2

But `types-huey` doesn't exist yet, because the stubs for `huey` are included in this PR. So you'll have to split this into at least 2 PRs: one for `huey`, and then a second one for `django-huey`.

PR #7745, opened 2022-04-30
Make test scripts pass `mypy --strict`

AlexWaygood:
- Add several type hints to untyped functions.
- While we're at it, upgrade several annotations to use modern syntax by using `from __future__ import annotations`. This means that the tests can't be run on Python 3.6, but 3.6 is EOL, and it is already the case that some scripts in this directory can only be run on more recent Python versions. E.g. `check_new_syntax.py` uses `ast.unparse`, which is only available in Python 3.9+.
- Fix a few pieces of code that didn't type check.


hauntsaninja (approved):
Looks good! And in general I think having a relatively high minimum Python version for these tests is fine.

PR #7752, opened 2022-04-30
Use `object` for `value` annotation in `Enum.__new__`

izumrudik:
This pull request reverts part of #7362 that brought back a bug discussed in https://github.com/python/mypy/issues/5788 and initially fixed in #2539

In short, the issue was that the following program always resulted
in an error when running mypy with the `--disallow-any-expr` flag:

    from enum import Enum

    class MyEnum(Enum):
        FOO = 1
        BAR = 2

    blah = MyEnum   # Error here

The root issue was that because the signature of Enum's
`__new__` method was typed as:

    def __new__(self: Type[Self], value: Any) -> Self: ...

This caused mypy to decide that the type of `MyEnum` was
`(Any) -> MyEnum`. This is correct based on the current
type signature, but unfortunately means that it becomes
impossible to ever use enums with the `--disallow-any-expr` flag.


stdlib/enum.pyi
```diff
@@ -155,7 +155,7 @@ class Enum(metaclass=EnumMeta):
     def _missing_(cls, value: object) -> Any: ...
     @staticmethod
     def _generate_next_value_(name: str, start: int, count: int, last_values: list[Any]) -> Any: ...
-    def __new__(cls: type[Self], value: Any) -> Self: ...
+    def __new__(cls: type[Self], value: object) -> Self: ...
```
AlexWaygood, line 158:
```suggestion
    # It's not true that `__new__` will accept any argument type,
    # so ideally we'd use `Any` to indicate that the argument type is inexpressible.
    # However, using `Any` causes too many false-positives for those using mypy's `--disallow-any-expr`
    # (see #7752, #2539, mypy/#5788),
    # and in practice using `object` here has the same effect as using `Any`.
    def __new__(cls: type[Self], value: object) -> Self: ...
```

AlexWaygood (commented):
This arguably makes the stub arguably slightly less "correct" on a purely theoretical level, but I'm happy to merge as long as we add a comment so this whole thing doesn't happen again :)

AlexWaygood (approved):
Thanks!

PR #7758, opened 2022-04-30
Fixed paramiko setblocking argument

### LINKED ISSUE #7697: paramiko channel.setblocking() argument type doesn't match the docs
According the paramiko docs, channel.setblocking() takes an integer 0 or 1:
https://docs.paramiko.org/en/stable/api/channel.html#paramiko.channel.Channel.setblocking

According to typeshed channel.pyi it takes a bool:
https://github.com/python/typeshed/blob/2299c43860c7115fabfd14ad854144e2ef20e132/stubs/paramiko/paramiko/channel.pyi#L72

This is causing my code, written to the docs, to throw mypy errors.  The implementation doesn't care:
```
        if blocking:
            self.settimeout(None)
        else:
            self.settimeout(0.0)
```

I guess either the docs should be changed, or the typeshed .pyi file should be updated...

Thanks for your work!

jbebic:
Added int 0,1 as an allowed argument using Literal[0,1]
Closes #7697


AlexWaygood (approved):
Thanks!

madscientist:
Thanks... someday I would have gotten around to it but procrastination wins again! :)

PR #7760, opened 2022-04-30
Add test cases for `pow` that are meant to fail a type check

AlexWaygood:


test_cases/stdlib/builtins/test_pow.py
```diff
@@ -71,3 +71,11 @@
 assert_type(pow(4.7, 9.2, None), Any)
 # See #7046 -- float for a positive 1st arg, complex otherwise
 assert_type((-95) ** 8.42, Any)
+
+# All of these should fail a type-checker
+# mypy will emit an error if any of them do not fail,
+# since we use --warn-unused-ignores when checking this subdirectory
```
JelleZijlstra, line 77: Can we turn on `reportUnnecessaryTypeIgnoreComment` for pyright too? Not sure if that can be path-specific though.

AlexWaygood: @erictraut, any pointers on the easiest way to configure pyright to get this to happen? :)

For context, this is a new `test_cases` directory we've recently added to typeshed. The idea is that it contains code samples that type checkers should be able to run without emitting any errors. This PR adds some test cases that _should fail_ a type checker -- by using `# type: ignore`s, combined with mypy's `--warn-unused-ignore` flag, we can get mypy to emit an error if any of these cases unexpectedly start passing a type check, rendering the `type: ignore` unnecessary. It would be nice if we could configure pyright so that it also warns if any of `# type: ignore` comments in this new subdirectory become unneeded.

erictraut: Pyright has an equivalent to mypy's `--warn-unused-ignore` feature. It's called `reportUnnecessaryTypeIgnoreComment`. You can configure it in the `pyrightconfig.json` file if you want to enable it.

I can see that you're using mypy-specific error codes in the code added as part of this PR. That can be problematic for other type checkers. See discussion in [this mypy issue](https://github.com/python/mypy/issues/12358). It would be nice if mypy added support for mypy-specific ignore comments rather than promoting the use of mypy error codes with the standard `# type: ignore` comment.

In this particular case, it will probably still work OK because pyright ignores any text after `# type: ignore`, so it will treat `# type: ignore[misc]` as the equivalent of `# type: ignore`.

AlexWaygood: Thanks! Yeah, our approach everywhere else in typeshed is to use mypy error codes after `type: ignore`s is to use mypy error codes (unless it's a pyright-specific `type: ignore`), so (without debating the relative merits or demerits of that policy in this PR discussion) I'm going to stick to that policy here.

I knew about the `reportUnnecessaryTypeIgnoreComment` --  my question was whether it's possible in `pyrightconfig.json` to turn that on only for the `test_cases` subdirectory specifically?

erictraut: You can't enable `reportUnnecessaryTypeIgnoreComment` for only a single directory. You'd need to create another `pyrightconfig.json` file and invoke pyright specifically on that directory. Search for `pyrightconfig.stricture.json` for an example of where we've done that already.

Alternatively, if you want to enable it for specific test source files, you could add the following line to the top of each file:
```python
# pyright: reportUnnecessaryTypeIgnoreComment=true
```

AlexWaygood: Great, thanks!

AlexWaygood:
Well, I officially have no idea what I'm doing with these `pyrightconfig` files.

srittau:
Could we more the changes to mypy_test to a separate PR? I think they are individually useful.

AlexWaygood:
Great, looks like both mypy and pyright will fail if there's an unused `type: ignore` in the `test_cases` directory, with this patch applied.

- Here's what the mypy failure looks like: https://github.com/python/typeshed/runs/6271321242?check_suite_focus=true
- Here's what the pyright failure looks like: https://github.com/python/typeshed/runs/6271329437?check_suite_focus=true

test_cases/stdlib/builtins/test_pow.py
AlexWaygood: > Alternatively, if you want to enable it for specific test source files, you could add the following line to the top of each file:
>
> `# pyright: reportUnnecessaryTypeIgnoreComment=true`

I couldn't figure out how to get the pyrightconfigs to work, so I've gone with this for now. We can maybe work out a way to get `reportUnnecessaryTypeIgnoreComment` activated for this whole directory in a separate PR.

Issue #6347, opened 2021-11-20
functools._lru_cache_wrapper should be a descriptor class, not a callable


erictraut:
I'm not entirely sure this is a bug in the type stub. It depends on the interpretation of ParamSpec when used with methods.

This is related to the discussion [here](https://github.com/python/typing/discussions/946) and this [bug report](https://github.com/microsoft/pyright/issues/2578) filed in the pyright issue tracker.

AlexWaygood:
Could this be a use case for `Concatenate`?

Leave the stub for `_lru_cache_wrapper` as it is:

```python 
from typing import Generic, TypeVar, ParamSpec, Callable

_P = ParamSpec("_P)
_T = TypeVar("_T)

class _lru_cache_wrapper(Generic[_P, _T]):  # type: ignore
    __wrapped__: Callable[_P, _T]  # type: ignore
    def __call__(self, *args: _P.args, **kwargs: _P.kwargs) -> _T: ...  # type: ignore
```

But then change the stub for `lru_cache`/`cache` to the following:

```python
import sys
from typing import overload, Concatenate, Any

if sys.version_info >= (3, 8):
    @overload
    def lru_cache(maxsize: int | None = ..., typed: bool = ...) -> Callable[[Callable[Concatenate[Any, _P], _T]], _lru_cache_wrapper[_P, _T]]: ...  # type: ignore
    @overload
    def lru_cache(maxsize: Callable[Concatenate[Any, _P], _T], typed: bool = ...) -> _lru_cache_wrapper[_P, _T]: ...  # type: ignore

else:
    def lru_cache(maxsize: int | None = ..., typed: bool = ...) -> Callable[[Callable[Concatenate[Any, _P], _T]], _lru_cache_wrapper[_P, _T]]: ...  # type: ignore

if sys.version_info >= (3, 9):
    def cache(__user_function: Callable[Concatenate[Any, _P], _T]) -> _lru_cache_wrapper[_P, _T]: ...  # type: ignore
```

erictraut:
I think `cache` and `lru_cache` can be used with either methods or non-method functions, right? Wouldn't that approach break for non-methods?

AlexWaygood:
> Wouldn't that approach break for non-methods?

Good point! Are you able to flesh out the rewrite-it-as-a-descriptor proposal a little?

erictraut:
I was hoping that someone who is more familiar with how these calls are meant to work could suggest a solution.

[...]

PR #7771, opened 2022-05-02
improve lru_cache by using ParamSpec

### LINKED ISSUE #6987: `functools.lru_cache` wrapped function with TypeVar has wrong return type.
Hi, 

I think there is a problem when using `functools.lru_cache` (and `functools.cache`). When the decorated function uses a `TypeVar` for typing arguments and return type, the wrapped function returend by `functools.lru_cache` has wrong (or unresolved) types.
This is not the case when replacing the `TypeVar` with an explicit type. See the mwe below. Thx.

```python
import functools
from typing import TypeVar

_T = TypeVar("_T")


def my_func(elem: _T) -> list[_T]:
    return [elem]
    
my_func_cached = functools.lru_cache(my_func)

a = my_func("a")
reveal_type(a)  # Revealed type is "builtins.list[builtins.str*]"
a_c = my_func_cached("a")

# ERROR: this should be "builtins.list*[builtins.str]"
reveal_type(a_c)  # Revealed type is "builtins.list*[_T`-1]"


# this works

def my_func_str(elem: str) -> list[str]:
    return [elem]
    
my_func_str_cached = functools.lru_cache(my_func_str)

a_str = my_func_str("a")
reveal_type(a_str)  # Revealed type is "builtins.list[builtins.str]"
a_str_c = my_func_str_cached("a")
reveal_type(a_str_c)  # Revealed type is "builtins.list*[builtins.str]"
```
Running:
Python 3.10.1
mypy 0.931
Linux 5.16.0-arch1-1

asottile:
https://github.com/python/mypy/issues/5107

Fixes #6987.


asottile:
hmmm, this is close but not quite there.  seems my little descriptor trick fixes a lot of situations but doesn't quite work for classmethods.  also seems to trigger an internal error in mypy (though pyright is ok with it) -- mypy's matching of Callable with ParamSpec seems a little incomplete

I had originally written this idea as a WIP mypy plugin

AlexWaygood:
Linking to previous discussion on this topic:

- https://github.com/python/typeshed/issues/6347

stdlib/functools.pyi
```diff
@@ -72,22 +74,48 @@ class _CacheInfo(NamedTuple):
...
 
 if sys.version_info >= (3, 8):
     @overload
-    def lru_cache(maxsize: int | None = ..., typed: bool = ...) -> Callable[[Callable[..., _T]], _lru_cache_wrapper[_T]]: ...
+    def lru_cache(maxsize: int | None = ..., typed: bool = ...) -> _LRUCacheDecorator: ...
+    @overload
+    def lru_cache(maxsize: Callable[Concatenate[_S, _P], _R], typed: bool = ...) -> _lru_cache_wrapper[_S, _P, _R]: ...
     @overload
-    def lru_cache(maxsize: Callable[..., _T], typed: bool = ...) -> _lru_cache_wrapper[_T]: ...
+    def lru_cache(maxsize: Callable[_P, _R], typed: bool = ...) -> _lru_cache_wrapper_0[_P, _R]: ...
```
AlexWaygood, line 109: Even if this didn't cause a mypy crash, don't these overloads overlap? How would a type checker ever know to choose the third overload over the second overload?

asottile: here's some examples which cannot match the first overload but can match the second:

- `def f(**k):`  (no positional argument to concatenate)
- `def f(*, k):` (no positional argument to concatenate)
- `def f():` (no positional argument)

AlexWaygood: Right, because a `ParamSpec` can be bound to an "empty" parameter specification, but `Concatenate[_S, _P]` cannot. Thanks.

AlexWaygood:
> I've reported the mypy crash at [python/mypy#12922](https://github.com/python/mypy/issues/12922).

The crash should be fixed in 0.970 now that https://github.com/python/mypy/pull/12953 has been merged :)

AlexWaygood:
Marking this as deferred until mypy 0.970 has been released.

stdlib/functools.pyi
```diff
@@ -45,22 +47,48 @@ class _CacheInfo(NamedTuple):
...
     def cache_clear(self) -> None: ...
-    def __copy__(self) -> _lru_cache_wrapper[_T]: ...
-    def __deepcopy__(self, __memo: Any) -> _lru_cache_wrapper[_T]: ...
+    def __copy__(self) -> _lru_cache_wrapper[_S, _P, _R]: ...
+    def __deepcopy__(self, __memo: Any) -> _lru_cache_wrapper[_S, _P, _R]: ...
+    def __get__(self, inst: object, owner: type) -> _lru_cache_wrapper_0[_P, _R]: ...
+
+class _LRUCacheDecorator(Protocol):
+    @overload
+    def __call__(self, func: Callable[Concatenate[_S, _P], _R]) -> _lru_cache_wrapper[_S, _P, _R]: ...
```
AlexWaygood, line 72:
```suggestion
    def __call__(self, func: Callable[Concatenate[_S, _P], _R]) -> _lru_cache_wrapper[_S, _P, _R]: ...  # type: ignore[misc]
```

AlexWaygood:
The patch no longer causes mypy to crash on 0.971, but there's now a stubtest failure. Also, the mypy_primer report is a bit of a mixed bag.

kkirsche:
Sorry if this derails at all. I noticed on:
https://github.com/python/typeshed/pull/7771/files#diff-02c51ed45b741b6765c365fb5d6578c586510bbbd57b10a1dba10aeba37ae157R50

That this makes `lru_cache` a `Generic[_P, _T]` (names differ, but for commenting this is easier to represent). Would this PR be the same model to be applied to weakref as noted by the todo comment here:
https://github.com/python/typeshed/blob/de1a79bd007e73bd57821a89e21bc08d4d07e75a/stdlib/weakref.pyi#L127

If this is too far off topic, just let me know and I can either create a new issue or table this question entirely

AlexWaygood:
@kkirsche, yeah, that's quite off-topic for this PR thread, I'd recommend opening an issue if you want to seek clarification on that :)

Greedquest:
Hello, has this been implemented somewhere. `lru_cache` and `cache` would really benefit from `ParamSpec` they seem like the textbook usecase. Right now mypy is seeing the function args as `Any` which means my callsites are not enforced.

kkirsche:
I don't believe so. I know when I tried it originally, I had run into some problems getting everything to work with the various dunder (double underscore) properties. Some of my personal obligations are clearing up so I may be able to take another stab at this. I may have still misunderstood ParamSpec at that time (I wanted it to be more like typescript function parameter utility type). At the very least, giving it another shot should allow me to articulate more clearly what specific issues I ran into in case others may be able to help solve those.

JelleZijlstra:
I believe the difficulty is in dealing with `__get__` to make it work correctly as both a function and a method decorator.

PR #7772, opened 2022-05-03
Add weakref_slot parameter to dataclass decorator

dignissimus:


stdlib/dataclasses.pyi
```diff
@@ -100,6 +100,22 @@ else:
         *, init: bool = ..., repr: bool = ..., eq: bool = ..., order: bool = ..., unsafe_hash: bool = ..., frozen: bool = ...
     ) -> Callable[[type[_T]], type[_T]]: ...
 
+if sys.version_info >= (3, 11):
```
JelleZijlstra, line 103: This needs to go up to line 82, changing the `if sys.version_info >= (3, 10):` to an `elif`. It replaces the overloads from the second set, it's not a new alternative.

AlexWaygood (approved):
Looks good! For future reference, the CPython PR that implemented these changes is here: https://github.com/python/cpython/pull/92160

PR #7773, opened 2022-05-03
`requests`: Allow session headers to be of type None

janrito:
Session headers can be unset by setting them to none, see: https://github.com/psf/requests/blob/40956723f27daf5e0d9759208ca69cef236ab339/requests/sessions.py#L61


Akuli:
Currently many functions accept `headers: _TextMapping`, where `_TextMapping` is defined to be `MutableMapping[str, str]`. Can you fix that too? The correct type is probably `MutableMapping[str, str] | MutableMapping[str, str | None]`, because `MutableMapping` is invariant and so a `dict[str, str]` is not a valid `MutableMapping[str, str | None]`.

janrito:
@Akuli does it make sense for headers to be `None`? 

in a session, `None` is used to remove headers from a specific request. Empty headers should be invalid I think?

janrito:
```python

sess = requests.Session()
sess.headers.update({"Content-Type": "application/json"})

# request with content type
session.get("http://api.com")

# request with no content type
session.get("http://api.com", headers={"Content-Type": None})
```

stubs/requests/requests/sessions.pyi
```diff
@@ -72,7 +72,7 @@ _Verify: TypeAlias = bool | str
 
 class Session(SessionRedirectMixin):
     __attrs__: Any
-    headers: CaseInsensitiveDict[str]
+    headers: CaseInsensitiveDict[str | None]
```
nsoranzo, line 75: As mentioned in https://docs.python-requests.org/en/latest/user/quickstart/#custom-headers , header values can also be bytestring.

```suggestion
    headers: CaseInsensitiveDict[str | bytes | None]
```

janrito: Interesting. I think I added None to the wrong place. We want to allow str or bytes on actual headers, but None only on updates

stubs/requests/requests/sessions.pyi
```diff
@@ -67,12 +67,14 @@ _Params: TypeAlias = Union[
     str | bytes,
 ]
 _TextMapping: TypeAlias = MutableMapping[str, str]
+_HeadersMapping = TypeAlias = MutableMapping[str, str | bytes]
```
JelleZijlstra, line 70: This is inconvenient because MutableMapping is invariant in its value type, so a `dict[str, str]` would not be acceptable.

janrito: What about 
```
MutableMapping[str, str] | MutableMapping[str, bytes]
```
?

janrito: 0e8ff7e

stubs/requests/requests/sessions.pyi
```diff
@@ -67,12 +67,14 @@ _Params: TypeAlias = Union[
     str | bytes,
 ]
 _TextMapping: TypeAlias = MutableMapping[str, str]
+_HeadersMapping: TypeAlias = MutableMapping[str, str] | MutableMapping[str, bytes]
+_HeadersUpdateMapping: TypeAlias = MutableMapping[str, str] | MutableMapping[str, bytes] | MutableMapping[str, None]
```
nsoranzo, line 71: I'm not sure that's correct, header values can be some str and some bytes?

janrito: Mmm I see - would we have to specify all combinations of mutable mappings for str, bytes and None?

janrito: b5ec834

stubs/requests/requests/models.pyi
```diff
@@ -95,7 +95,7 @@ class Response:
     __attrs__: Any
     _content: bytes | None  # undocumented
     status_code: int
-    headers: CaseInsensitiveDict[str]
+    headers: CaseInsensitiveDict[str | bytes]
```
JelleZijlstra, line 98: Are you sure this is correct? These headers are from the HTTP response, and I'd expect requests to return the same type consistently.

This seems to be the cause of a false positive on this line found by mypy-primer: https://github.com/yurijmikhalevich/rclip/blob/98a2454b988986f6ed67ae6e9fe3342ea82e0f92/rclip/utils.py#L92

janrito:
I'm unsure what the next steps are

Akuli:
You removed `| bytes` from the wrong place. It's fine to have it in session headers, but not in response headers.

janrito:
> You removed `| bytes` from the wrong place. It's fine to have it in session headers, but not in response headers.

whelp f9aaf0c, 0d80bdf

AlexWaygood (commented):
`CaseInsensitiveDict[str | bytes]` has the same issues that `MutableMapping[str | bytes]` has (invariance) — `CaseInsensitiveDict[str] | CaseInsensitiveDict[bytes] | CaseInsensitiveDict[str | bytes]` would be better

AlexWaygood:
Well, my suggested change to work around the invariance of `CaseInsensitiveDict` appeared to introduce a lot of new false-positives, so, uh... maybe ignore my remarks there and revert the last commit? 🙃

janrito:
I find it quite challenging to have a local workflow for this. Not even testing other projects, but my own. I'd love some pointers if anyone has some

JelleZijlstra:
> I find it quite challenging to have a local workflow for this. Not even testing other projects, but my own. I'd love some pointers if anyone has some

One quick-and-dirty way is to just edit the stub files used by your type checker (somewhere in site-packages) in place. You can also use `$MYPYPATH`.

Akuli:
To clarify, here's how you would use `MYPYPATH`:

```
$ MYPYPATH=/home/yourname/typeshed/stubs/requests mypy project_using_requests
```

This lets mypy import from `/home/yourname/typeshed/stubs/requests`. There's also `--custom-typeshed-dir`, but it unfortunately affects stdlib stubs only.

Akuli (approved):
Primer is happy and the code makes sense to me (`None` can be used to override whatever was configured in the session). Thanks for contributing! This one was trickier than expected.

rpdelaney:
After this change, my type checker is throwing an error:

```
error: Argument "headers" to "get" has incompatible type "MappingProxyType[str, str]"; expected "Optional[MutableMapping[str, str]]"  [arg-type]
```

Do headers have to be mutable? Can we also accept `MappingProxyType`?

PR #7782, opened 2022-05-05
flask-cors: add regex to decorator

ikonst:
Per https://github.com/corydolphin/flask-cors/blob/79bb3721f72ff540ac2575e810f9bfd6c3380432/flask_cors/decorator.py#L29

### DIFF
--- a/stubs/Flask-Cors/flask_cors/decorator.pyi
+++ b/stubs/Flask-Cors/flask_cors/decorator.pyi
@@ -1,7 +1,7 @@
 from collections.abc import Callable
 from datetime import timedelta
 from logging import Logger
-from typing import Any, Pattern
+from typing import Any, Iterable, Pattern
 from typing_extensions import ParamSpec
 
 _P = ParamSpec("_P")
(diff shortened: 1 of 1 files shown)


stubs/Flask-Cors/flask_cors/decorator.pyi
```diff
@@ -10,10 +10,10 @@ LOG: Logger
 
 def cross_origin(
     *args: Any,
-    origins: str | list[str] | None = ...,
+    origins: str | Pattern[str] | list[str | Pattern[str]] | None = ...,
```
JelleZijlstra, line 13: `list[str | Pattern[str]]` is bad because list is invariant, so a `list[str]` wouldn't be accepted. The nicest solution would be if we can write `Sequence[str | Pattern[str]]`, since `Sequence` is covariant. Does it really require a list?

ikonst: Looking at the code, `Iterable` should suffice

AlexWaygood (approved):
The test suite also has an example of a `re.Pattern` object being passed to the `origins` parameter: https://github.com/corydolphin/flask-cors/blob/79bb3721f72ff540ac2575e810f9bfd6c3380432/tests/decorator/test_origins.py#L65

PR #7785, opened 2022-05-06
bs4: update to 4.11

hauntsaninja:
I mainly need the new warning. I'll get around to fixing all the other things later.


AlexWaygood (commented):
I'm a bit nervous about the number of new allowlist entries here -- should we really be bumping the version if the stubs have got so many errors in them when compared with the 4.11 runtime?

hauntsaninja:
It's not worse than using types-beautifulsoup4 as is / the actual errors aren't too bad / we have plenty of lengthy allowlists. I'll fix all of them soon enough, but this saves me a type ignore at work :-)

stubs/beautifulsoup4/bs4/builder/__init__.pyi
```diff
@@ -64,3 +64,5 @@ class HTMLTreeBuilder(TreeBuilder):
 
 class ParserRejectedMarkup(Exception):
     def __init__(self, message_or_exception) -> None: ...
+
+class XMLParsedAsHTMLWarning(UserWarning): ...
```
AlexWaygood, line 68:
```suggestion
class XMLParsedAsHTMLWarning(UserWarning):
    MESSAGE: ClassVar[str]
```

https://bazaar.launchpad.net/~leonardr/beautifulsoup/bs4/view/head:/bs4/builder/__init__.py#L39

AlexWaygood:
> I'll fix all of them soon enough, but this saves me a type ignore at work :-)

Okay, if you pinky promise :)

PR #7786, opened 2022-05-06
Try to make mypy_primer work for third-party stubs

Akuli:
Adding all stub directories to MYPYPATH is kind of a hack, but if it works, it will be better than nothing.


AlexWaygood:
This is fantastic!

PR #7788, opened 2022-05-06
`csv`: annotate keyword arguments in various functions and methods

AlexWaygood:
`_csv.reader`, `_csv.writer`, `_csv.register_dialect`, `csv.DictReader.__init__`, `csv.DictWriter.__init__`: none of these take arbitrary keyword arguments like the stub currently says. Instead, they all take keyword arguments corresponding to the fieldnames on `_csv.Dialect`.


stdlib/csv.pyi
```diff
@@ -1,5 +1,5 @@
 import sys
-from _csv import (
+from _csv import (  # actually csv.Dialect is a different class to _csv.Dialect at runtime,; but for typing purposes, they're identical
```
AlexWaygood, line 2:
```suggestion
# actually csv.Dialect is a different class to _csv.Dialect at runtime, but for typing purposes, they're identical
from _csv import (
```

AlexWaygood:
(Planning on splitting this up into smaller PRs so that it's easier to review, just experimenting to see the primer diff for now.)

stdlib/csv.pyi
```diff
@@ -98,26 +109,40 @@ class DictReader(Generic[_T], Iterator[_DictReadMapping[_T, str]]):
         restkey: str | None = ...,
         restval: str | None = ...,
         dialect: _DialectLike = ...,
-        *args: Any,
-        **kwds: Any,
+        *,
+        delimiter: str = ...
```
AlexWaygood, line 113:
```suggestion
        delimiter: str = ...,
```

JelleZijlstra (approved):
This is correct, but while reviewing I noticed that our stubs for Dialect are wrong. I'll try to address that separately.

AlexWaygood:
> This is correct, but while reviewing I noticed that our stubs for Dialect are wrong. I'll try to address that separately.

It doesn't seem like our csv stubs have had much love in a while.

PR #7790, opened 2022-05-06
`csv.pyi`: minor cleanup

AlexWaygood:
- Add a comment about `_csv.Dialect` and `csv.Dialect` (they're two different classes at runtime, but in the stub we re-export for simplicity).
- Instantiating a `DictWriter` fails at runtime if the object passed to the `f` parameter doesn't have a `write()` method.
- Instantiating a `DictWriter` fails at runtime if the string passed to the `extrasaction` parameter is not either "raise" or "ignore".

### DIFF
--- a/stdlib/csv.pyi
+++ b/stdlib/csv.pyi
@@ -1,4 +1,6 @@
 import sys
+
+# actually csv.Dialect is a different class to _csv.Dialect at runtime, but for typing purposes, they're identical
 from _csv import (
     QUOTE_ALL as QUOTE_ALL,
     QUOTE_MINIMAL as QUOTE_MINIMAL,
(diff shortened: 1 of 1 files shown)


stdlib/csv.pyi
```diff
@@ -107,14 +110,14 @@ class DictReader(Generic[_T], Iterator[_DictReadMapping[_T, str]]):
 class DictWriter(Generic[_T]):
     fieldnames: Collection[_T]
     restval: Any | None
-    extrasaction: str
+    extrasaction: Literal["raise", "ignore"]
```
srittau, line 113: In theory, this could be spelled in any case, but I'm fine with just allowing lower case for now.

srittau (approved):

PR #7796, opened 2022-05-07
os.popen: add encoding argument for 3.11

JelleZijlstra:
python/cpython#92374


hauntsaninja (commented):
Might need to be changed if https://github.com/python/cpython/pull/92415 is merged

hauntsaninja (commented):
Also am I missing something obvious or does the original PR not do anything?

Akuli:
We shouldn't merge this until the argument actually works in cpython. If they ship a broken argument to the next release, it's better to leave it out from the stubs.

JelleZijlstra:
The argument was removed again: python/cpython#92836.

PR #7798, opened 2022-05-07
Add new arguments to subprocess.Popen

JelleZijlstra:
Most recent one in python/cpython#23930. This makes me wish we
could have had type evaluation functions.


hauntsaninja:
"subprocess.💩en", Jelle tell us how you really feel

stdlib/subprocess.pyi
```diff
@@ -804,7 +804,556 @@ class Popen(Generic[AnyStr]):
...
+            encoding: str | None = ...,
+            errors: str | None = ...,
+            user: str | int | None = ...,
+            group: str | int | None = ...,
+            extra_groups: Iterable[str | int] | None = ...,
+            umask: int = ...,
+            pipesize: int = ...,
+        ) -> Popen[Any]: ...
+    elif sys.version_info >= (3, 9):
+        # user, group, extra_groups, umask were added in 3.9
```
Akuli, line 1180: These shouldn't be available on windows, but we probably shouldn't change the stub to reflect that. It's already difficult enough to read. The same goes for `process_group`.

Akuli (approved):
Reviewed by diffing the overloads against each other.

Issue #6153, opened 2021-10-11
subprocess stdout read1


Akuli:
```python
import subprocess
process = subprocess.Popen("echo hello", shell=True, stdout=subprocess.PIPE)
assert process.stdout is not None
print(process.stdout.read1())
```

mypy error:
```
asd.py:4: error: "IO[bytes]" has no attribute "read1"; maybe "read"?
Found 1 error in 1 file (checked 1 source file)
```
But at runtime there is `read1()` method. It is useful because unlike `read()`, it will return the output as soon as it is available, instead of reading more until it gets n bytes or end of file.

xmo-odoo:
> We should use the concrete runtime types in the subprocess stubs instead of `IO[bytes]`. I just tried and it's an `_io.BufferedReader` for me, but not sure whether that's always the case.

If `buffering=0` is passed in the output streams are unbuffered `FileIO` without read1.

PR #7802, opened 2022-05-07
tempfile.SpooledTemporaryFile: inherit from IOBase on 3.11

JelleZijlstra:
python/cpython#29560


hauntsaninja (approved):

AlexWaygood (commented):
I'd prefer us to use a wordier-but-more-readable alias name, like `_SpooledTemporaryFileBase`, rather than `_STFBase` which feels a bit opaque.

stdlib/tempfile.pyi
```diff
@@ -1,3 +1,4 @@
+import io
 import os
 import sys
 from _typeshed import Self
```
AlexWaygood, line 4:
```suggestion
from _typeshed import Self, WriteableBuffer
```

stdlib/tempfile.pyi
```diff
@@ -318,6 +324,13 @@ class SpooledTemporaryFile(IO[AnyStr]):
     def fileno(self) -> int: ...
     def flush(self) -> None: ...
     def isatty(self) -> bool: ...
+    if sys.version_info >= (3, 11):
+        def readable(self) -> bool: ...
```
AlexWaygood, line 328:
```suggestion
```

stdlib/tempfile.pyi
```diff
@@ -318,6 +324,13 @@ class SpooledTemporaryFile(IO[AnyStr]):
...
+        def read1(self, __size: int = ...) -> AnyStr: ...
+        def readinto(self, b: WriteableBuffer) -> int: ...
+        def readinto1(self, b: WriteableBuffer) -> int: ...
+        def seekable(self) -> bool: ...
```
AlexWaygood, line 332:
```suggestion
```

AlexWaygood (commented):
In typeshed we've always pretended that `SpooledTemporaryFile` has `readable()` and `seekable()` methods on <3.11, even though it doesn't at runtime, because it inherits from `IO` in typeshed, and `IO` has these as abstract methods.

stdlib/tempfile.pyi
JelleZijlstra: Thanks, applied this and updated the comments.

AlexWaygood (commented):
What about the new `detach()` method? https://github.com/python/cpython/pull/29560/files#diff-6553a99f3ae04c9fc9c2349ac27037bfe3b274d5ca54d5800d9c5b5f11e29d21R813

stdlib/tempfile.pyi
```diff
@@ -318,20 +324,26 @@ class SpooledTemporaryFile(IO[AnyStr]):
     def fileno(self) -> int: ...
     def flush(self) -> None: ...
     def isatty(self) -> bool: ...
+    if sys.version_info >= (3, 11):
+        def read1(self, __size: int = ...) -> AnyStr: ...
+        def readinto(self, b: WriteableBuffer) -> int: ...
+        def readinto1(self, b: WriteableBuffer) -> int: ...
```
AlexWaygood, line 330: I'm confused by these methods.

According to the runtime, these call `self._file.read1()`, `self._file.readinto()`, `self._file.readinto1()`: https://github.com/python/cpython/pull/29560/files#diff-6553a99f3ae04c9fc9c2349ac27037bfe3b274d5ca54d5800d9c5b5f11e29d21R766-R774

But `self._file` could be a `BytesIO` or a `TextIOWrapper`: https://github.com/python/cpython/pull/29560/files#diff-6553a99f3ae04c9fc9c2349ac27037bfe3b274d5ca54d5800d9c5b5f11e29d21R653-R656

And according to our current stubs, `TextIOWrapper` doesn't have a `read1`, `readinto` or `readinto1` method: https://mypy-play.net/?mypy=latest&python=3.10&gist=c5b47d63657de3a0a8c9a275baf16705

Akuli: Whether or not these exist seems to depend on the mode:

```
>>> tempfile.TemporaryFile("r").read1
AttributeError: '_io.TextIOWrapper' object has no attribute 'read1'
>>> tempfile.TemporaryFile("rb").read1
<built-in method read1 of _io.BufferedReader object at 0x7fb6c9dd67d0>
```

I can't think of a practical use case for `read1()` or `readinto1()`. With subprocesses, `proc.stdout.read1(100)` reads data from the stdout of a subprocess without waiting for more data to arrive (#6153). But with tempfiles, a plain `.read()` works like that and just returns empty bytes once it's at the end of the file.

stdlib/tempfile.pyi
```diff
@@ -318,20 +324,26 @@ class SpooledTemporaryFile(IO[AnyStr]):
...
+        def detach(self) -> io.RawIOBase: ...
+
     def read(self, n: int = ...) -> AnyStr: ...
-    def readline(self, limit: int = ...) -> AnyStr: ...
-    def readlines(self, hint: int = ...) -> list[AnyStr]: ...
+    def readline(self, limit: int | None = ...) -> AnyStr: ...  # type: ignore[override]
+    def readlines(self, hint: int = ...) -> list[AnyStr]: ...  # type: ignore[override]
```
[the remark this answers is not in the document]
JelleZijlstra: I'll mark them as positional-only to match the `*args` declaration.

stdlib/tempfile.pyi
JelleZijlstra: Adding a comment.

stdlib/tempfile.pyi
```diff
@@ -217,9 +218,14 @@ class _TemporaryFileWrapper(Generic[AnyStr], IO[AnyStr]):
     def write(self, s: AnyStr) -> int: ...
     def writelines(self, lines: Iterable[AnyStr]) -> None: ...
 
-# It does not actually derive from IO[AnyStr], but it does implement the
-# protocol.
-class SpooledTemporaryFile(IO[AnyStr]):
+# It does not actually derive from IO[AnyStr], but it does mostly behave
+# like one.
```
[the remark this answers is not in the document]
JelleZijlstra: Done

AlexWaygood (approved):

JelleZijlstra:
hm, I expected unused stubtest allowlist entries. I'll look into that in a followup PR.

PR #7804, opened 2022-05-07
Run mypy_primer in all pull requests

Akuli:
It now works even for third-party stubs. https://github.com/hauntsaninja/mypy_primer/pull/31


Akuli:
It might be a good idea to close and reopen some of our open third-party stubs pull requests once this is merged, e.g. #7773

hauntsaninja (approved):
Thanks for making this happen!

AlexWaygood (approved):

PR #7806, opened 2022-05-08
Fix stdlib `subprocess.Popen.returncode` typing

mnixry:
According to [the document](https://docs.python.org/3/library/subprocess.html#subprocess.Popen.returncode), the `Popen.returncode` should be `Optional[int]`:

>The child return code, set by [poll()](https://docs.python.org/3/library/subprocess.html#subprocess.Popen.poll) and [wait()](https://docs.python.org/3/library/subprocess.html#subprocess.Popen.wait) (and indirectly by [communicate()](https://docs.python.org/3/library/subprocess.html#subprocess.Popen.communicate)). **A None value indicates that the process hasn’t terminated yet.**

### DIFF
--- a/stdlib/subprocess.pyi
+++ b/stdlib/subprocess.pyi
@@ -798,7 +798,7 @@ class Popen(Generic[AnyStr]):
     stdout: IO[AnyStr] | None
     stderr: IO[AnyStr] | None
     pid: int
-    returncode: int | None
+    returncode: int | Any
     universal_newlines: bool
 
     # Technically it is wrong that Popen provides __new__ instead of __init__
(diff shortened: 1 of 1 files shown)


srittau:
The problem with this change is that there are situations where the caller knows that the exit code can't be `None`, see the mypy primer output above. That said, we should at least change the return type to `int | Any` to cover the `None` case. Also obligatory link to python/typeshed#566.

mnixry:
> The problem with this change is that there are situations where the caller knows that the exit code can't be `None`, see the mypy primer output above. 

Thank you for your explanation. 

So according to my understanding, using `int` instead of `Optional<int>` is a compromised solution to keep compatibility with the existing code? 
I just remember that the type was `int | None` some time ago, this makes the recent change reasonable for me.

srittau:
> So according to my understanding, using `int` instead of `Optional<int>` is a compromised solution to keep compatibility with the existing code?

A mixture of compatibility with existing code, but also not requiring authors of new code to have to add `assert` in places like this, even though a human reader can clearly spot that the value can't be `None`.

hauntsaninja (approved):

Issue #7805, opened 2022-05-08
PIllow: Image.mode should be _Mode, not str


Julian-O:
When you create Image instances in Pillow, you pass a `mode` which tells it what colour system to use (e.g. "RGB", "CMYK") by passing a string. There are only 12 valid string values.

[Image.pyi](https://github.com/python/typeshed/blob/master/stubs/Pillow/PIL/Image.pyi) (line 17) defines a `_Mode` TypeAlias for the Literal values it can be. 

Once you have created the Image instance, you can find what mode it is in by accessing the `mode` attribute. [Image.pyi](https://github.com/python/typeshed/blob/master/stubs/Pillow/PIL/Image.pyi) (line 108) defines the `mode` attribute as a str

As a result, if you try to create one image instance with the same mode as another:

    from PIL import Image
    im1 = Image.new(mode="RGBA", size=(100, 100), color="white")
    im2 = Image.new(mode=im1.mode, size=im1.size, color="black")

you get a type warning that it expected a type Literal, but got a str instead.

----

My recommendation is that Image.pyi defines the `mode` attribute to be of type `_Mode`.

I am still a beginner with some of these concepts, so I suggest taking my recommendation with a grain of salt.

JelleZijlstra:
That sounds right, feel free to submit a PR!

PR #7807, opened 2022-05-08
#7805: Ensure all references to mode are Literals, not str.

Julian-O:
Made edits, waiting for automated tests to run.

I am unclear how to actually *test* the changes - e.g. to make sure the example code in #7805 is now passing without type warnings.


srittau (approved):

srittau:
Thanks! We have a good set of automated tests in typeshed that already test quite a few things. In fact, mypy primer had a complaint above, but in this case I think pytorch should just adapt its type annotations.

We have a few manual tests in `test_cases`, but for a fairly simple change as this, we don't need any specific tests.

Julian-O:

Hmm... What is going on here? [meta_.py](https://github.com/pytorch/vision/blob/main/torchvision/prototype/transforms/functional/_meta.py) makes  call to convert (line 156) based on literals listed in line 136-141, one of which is "LA" which is outside the accepted values for `_Mode`.

[Pillow](https://pillow.readthedocs.io/en/stable/handbook/concepts.html#concept-modes) explains that LA is not one of the "standard modes", but that 

> Pillow also provides limited support for a few additional modes, including:
>
> `LA` (L with alpha)
> [...]

I am not sure what the next step should be,

hauntsaninja:
Maybe we should add it to `_Mode`?

PR #7808, opened 2022-05-08
Fix bugs in `check_new_syntax.py`

AlexWaygood:
`check_new_syntax` should have emitted an error for this commit, but didn't: https://github.com/python/typeshed/pull/7806/commits/b9893473964d8451078a180c5539b4f13a1066e4

(It was almost certainly me who introduced this bug, so, uh, my bad!)

### DIFF
--- a/tests/check_new_syntax.py
+++ b/tests/check_new_syntax.py
@@ -44,10 +44,12 @@ def check_new_syntax(tree: ast.AST, path: Path, stub: str) -> list[str]:
     class OldSyntaxFinder(ast.NodeVisitor):
         def visit_AnnAssign(self, node: ast.AnnAssign) -> None:
             AnnotationUnionFinder().visit(node.annotation)
+            self.generic_visit(node)
 
         def visit_arg(self, node: ast.arg) -> None:
             if node.annotation is not None:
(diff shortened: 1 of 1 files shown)


tests/check_new_syntax.py
```diff
@@ -80,10 +84,12 @@ class TextFinder(ast.NodeVisitor):
         def visit_ImportFrom(self, node: ast.ImportFrom) -> None:
             if node.module == "typing" and any(thing.name == "Text" for thing in node.names):
                 errors.append(f"{path}:{node.lineno}: Use `str` instead of `typing.Text` in a Python-3-only stub.")
+            self.generic_visit(node)
 
         def visit_Attribute(self, node: ast.Attribute) -> None:
             if isinstance(node.value, ast.Name) and node.value.id == "typing" and node.attr == "Text":
```
JelleZijlstra, line 90: We should put this in flake8-pyi instead.

JelleZijlstra (approved):

tests/check_new_syntax.py
AlexWaygood: It's on my to-do list :)

PR #7809, opened 2022-05-08
Run pyright on the `tests` directory in CI

AlexWaygood:


tests/pytype_test.py
```diff
@@ -103,10 +103,6 @@ def _get_module_name(filename: str) -> str:
     return ".".join(module_parts).replace(".pyi", "").replace(".__init__", "")
 
 
-def _is_version(path: str, version: str) -> bool:
-    return any("{}{}{}".format(d, os.path.sep, version) in path for d in TYPESHED_SUBDIRS)
-
-
```
AlexWaygood, line 109: pyright pointed out that this private function isn't used at all.

AlexWaygood:
Ah, I tested locally in my virtual environment. This might be a bit more complicated. Closing for now.

hauntsaninja (commented):
(the benefit of this feels pretty marginal, so I wouldn't try too hard to get it to work)

AlexWaygood:
> (the benefit of this feels pretty marginal, so I wouldn't try too hard to get it to work)

Yeah

PR #7810, opened 2022-05-09
ast: FormattedValue.conversion cannot be None

JelleZijlstra:
```
In [4]: ast.dump(ast.parse('f"{x!r}"'))
Out[4]: "Module(body=[Expr(value=JoinedStr(values=[FormattedValue(value=Name(id='x', ctx=Load()), conversion=114)]))], type_ignores=[])"

In [5]: ast.dump(ast.parse('f"{x}"'))
Out[5]: "Module(body=[Expr(value=JoinedStr(values=[FormattedValue(value=Name(id='x', ctx=Load()), conversion=-1)]))], type_ignores=[])"
```
(On 3.9 but I don't think this has changed since 3.6.)

The stdlib also assumes this: https://github.com/python/cpython/blob/main/Lib/ast.py#L1211 (`chr(None)` doesn't work).

### DIFF
--- a/tests/stubtest_allowlists/py3_common.txt
+++ b/tests/stubtest_allowlists/py3_common.txt
@@ -685,6 +685,9 @@ pkgutil.ImpLoader.is_package
 pkgutil.ImpLoader.load_module
 pkgutil.ImpLoader.source
 
+# None on the class, but never None on instances
+_ast.FormattedValue.conversion
+
 # ==========
 # Exists at runtime, but missing from stubs
 # ==========
--- a/tests/stubtest_allowlists/py3_common.txt
+++ b/tests/stubtest_allowlists/py3_common.txt
@@ -687,6 +687,7 @@ pkgutil.ImpLoader.source
 
 # None on the class, but never None on instances
 ast.FormattedValue.conversion
+_ast.FormattedValue.conversion
 
 # ==========
 # Exists at runtime, but missing from stubs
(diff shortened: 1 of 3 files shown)


tests/stubtest_allowlists/py3_common.txt
```diff
@@ -685,6 +685,9 @@ pkgutil.ImpLoader.is_package
 pkgutil.ImpLoader.load_module
 pkgutil.ImpLoader.source
 
+# None on the class, but never None on instances
+_ast.FormattedValue.conversion
```
JelleZijlstra, line 689:
```suggestion
ast.FormattedValue.conversion
```

tests/stubtest_allowlists/py3_common.txt
```diff
@@ -685,6 +685,9 @@ pkgutil.ImpLoader.is_package
 pkgutil.ImpLoader.load_module
 pkgutil.ImpLoader.source
 
+# None on the class, but never None on instances
+ast.FormattedValue.conversion
```
JelleZijlstra, line 689:
```suggestion
ast.FormattedValue.conversion
_ast.FormattedValue.conversion
```

hauntsaninja:
I think you might only need the allowlist on py39

JelleZijlstra:
I guess so, I wonder what changed in 3.9.

hauntsaninja:
Well 3.9 makes sense, the better question is what happened in 3.10 and the answer is Batuhan made things better: https://github.com/python/cpython/pull/30467

PR #7816, opened 2022-05-10
beautifulsoup4: fix stubtest errors

hauntsaninja:


srittau (approved):
I did not verify this against the implementation, but trust you and stubtest.

PR #7818, opened 2022-05-10
Update testing_engine() annotations

srittau:
Partly addresses #7815.

### DIFF
--- a/stubs/SQLAlchemy/sqlalchemy/testing/engines.pyi
+++ b/stubs/SQLAlchemy/sqlalchemy/testing/engines.pyi
@@ -45,7 +45,7 @@ class ReconnectFixture:
 def reconnecting_engine(url: Any | None = ..., options: Any | None = ...): ...
 @overload
 def testing_engine(  # type: ignore[misc]
-    url: URL | None = ...,
+    url: URL | str | None = ...,
     options: Mapping[str, Any] | None = ...,
     future: Any | None = ...,
     asyncio: Literal[False] = ...,
@@ -53,7 +53,7 @@ def testing_engine(  # type: ignore[misc]
 ) -> Engine: ...
 @overload
 def testing_engine(
-    url: URL | None = ...,
+    url: URL | str | None = ...,
     options: Mapping[str, Any] | None = ...,
     future: Any | None = ...,
     asyncio: Literal[True] = ...,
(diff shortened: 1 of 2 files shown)


stubs/SQLAlchemy/sqlalchemy/testing/engines.pyi
```diff
@@ -37,14 +43,22 @@ class ReconnectFixture:
     def restart(self) -> None: ...
 
 def reconnecting_engine(url: Any | None = ..., options: Any | None = ...): ...
+@overload
+def testing_engine(  # type: ignore[misc]
+    url: URL | None = ...,
```
Akuli, line 48: It does `url = make_url(url)`, so passing in a string instead of `URL` instance will work too.

stubs/SQLAlchemy/sqlalchemy/testing/engines.pyi
```diff
@@ -37,14 +43,22 @@ class ReconnectFixture:
...
+) -> Engine: ...
+@overload
 def testing_engine(
-    url: Any | None = ...,
-    options: Any | None = ...,
+    url: URL | None = ...,
+    options: Mapping[str, Any] | None = ...,
     future: Any | None = ...,
-    asyncio: bool = ...,
+    asyncio: Literal[True] = ...,
```
Akuli, line 59:
```suggestion
    asyncio: Literal[True],
```

stubs/SQLAlchemy/sqlalchemy/testing/engines.pyi
```diff
@@ -37,14 +43,22 @@ class ReconnectFixture:
     def restart(self) -> None: ...
 
 def reconnecting_engine(url: Any | None = ..., options: Any | None = ...): ...
+@overload
+def testing_engine(  # type: ignore[misc]
+    url: URL | None = ...,
+    options: Mapping[str, Any] | None = ...,
+    future: Any | None = ...,
```
Akuli, line 50: `bool | None` maybe? It does `future or (... and future is not False)`.

stubs/SQLAlchemy/sqlalchemy/testing/engines.pyi
Akuli: I think the `# type: ignore` might be unnecessary after this.

srittau: Won't work, since it follows arguments with defaults.

Akuli (approved):

PR #7819, opened 2022-05-10
Remove explicit inheritance from `object`

AlexWaygood:
These stubs no longer support Python 2, and explicit inheritance from `object` is never necessary in Python 3.

Unblocks https://github.com/PyCQA/flake8-pyi/pull/217


Akuli (approved):
Protobuf stubs are auto-generated, but we can edit them again or tweak the auto-generation when they are updated next time.

AlexWaygood:
> Protobuf stubs are auto-generated, but we can edit them again or tweak the auto-generation when they are updated next time.

Only the files ending with `_pb2` are autogenerated :)

Issue #7597, opened 2022-04-06
Upgrade Redis stubs to 4.2.x


Ankhas:
The redis-py package has been updated to 4.2.2 : https://github.com/redis/redis-py/releases/tag/v4.2.2 and currently stubs are in version 4.1.*

kovalevvlad:
A big change here and the reason why I upvoted is the addition of asyncio

AlexWaygood:
PR welcome!

srittau:
#7675 didn't add async support. Reopening.

squaresmile:
Thank you for the updated redis typing. 2 things I noticed with the new types:

```
from redis.asyncio import Redis 

async def func(redis: Redis, redis_key: str): # 1
    return await redis.get(redis_key)         # 2
```
- 1 leads to mypy error `Missing type parameters for generic type` while `TypeError: <class 'redis.asyncio.client.Redis'> is not a generic class`.
- 2 leads to `Incompatible types in "await" (actual type "Optional[Any]", expected type "Awaitable[Any]")`.

PR #7820, opened 2022-05-10
fix(redis): commands from asyncio module must be async

sileht:
redis-py has introduced an asyncio module since 4.2 and almost all
commands should be async if the redis.asyncio.Redis is used instead of
redis.Redis.

This changes adds the types for all core commands.

Related to #7597


sileht:
Thx @AlexWaygood for the CI fix.

PR #5120, opened 2021-03-15
Pysftp Stubs

hatal175:
Adding stubs for pysftp.

Notes:
flake8 warned about:
./stubs/pysftp/pysftp/__init__.pyi:76:5: F811 redefinition of unused 'cd' from line 13
./stubs/pysftp/pysftp/__init__.pyi:102:5: F811 redefinition of unused 'walktree' from line 13

But for the life of me I couldn't understand why its a redefinition - it's a function inside a class. I've put a comment for flake to ignore it but I'll fix if you know how.

Other than that, pretty straight forward as this is a pretty thin wrapper for paramiko.

PR #7822, opened 2022-05-10
pysftp: remove stubs

hauntsaninja:
Helps with #7815

While these stubs were contributed to us relatively recently, a little
over a year ago, the underlying project has had no releases or commits
in the last six years and the last Python they have a classifier for is
Python 3.5


JelleZijlstra:
cc @hatal175 who added these in #5120

hatal175:
No hard feelings :P

hauntsaninja:
The types-pysftp package will continue to exist and help people! :-)

I am slightly confused about the CI failures in #7815 (feels like it should be picking up a new enough paramiko), so it's still a little TBD what the options are

AlexWaygood (approved):
I'm fine with this if @hatal175 is :)

hauntsaninja:
Closing this till we figure out what happened in #7815

PR #7825, opened 2022-05-11
statistics: improve type for median_grouped

### LINKED ISSUE #7824: median_grouped() returns a float
There was a bug in *median_grouped()* that allowed the output type for match in the input type for iterables of length one.  Otherwise, it always returned a float.

    def median_grouped(data: Iterable[_NumberT], interval: _NumberT = ...) -> _NumberT: ...

That bug was fixed.  The input types are anything convertible to a float and the output is always a float.

hauntsaninja:
Fixes #7824


stdlib/statistics.pyi
```diff
@@ -94,7 +94,13 @@ else:
 def median(data: Iterable[_NumberT]) -> _NumberT: ...
 def median_low(data: Iterable[SupportsRichComparisonT]) -> SupportsRichComparisonT: ...
 def median_high(data: Iterable[SupportsRichComparisonT]) -> SupportsRichComparisonT: ...
-def median_grouped(data: Iterable[_NumberT], interval: _NumberT = ...) -> _NumberT: ...
+
+if sys.version_info >= (3, 12):
```
JelleZijlstra, line 98: Possibly 3.11 depending on https://github.com/python/cpython/pull/92533#issuecomment-1123056470

hauntsaninja: Oh glad you asked about it, I wasn't sure so I just tried to say "is so far only on the 3.12 branch" loudly and see if Raymond commented

srittau (approved):

PR #7838, opened 2022-05-15
Add version to redis

Dreamsorcerer:
https://github.com/redis/redis-py/blob/master/redis/__init__.py#L48-L54

### DIFF
--- a/stubs/redis/redis/__init__.pyi
+++ b/stubs/redis/redis/__init__.pyi
@@ -59,3 +59,6 @@ RedisError = exceptions.RedisError
 ResponseError = exceptions.ResponseError
 TimeoutError = exceptions.TimeoutError
 WatchError = exceptions.WatchError
+
+__version__: str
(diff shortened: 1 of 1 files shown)


stubs/redis/redis/__init__.pyi
```diff
@@ -59,3 +59,6 @@ RedisError = exceptions.RedisError
 ResponseError = exceptions.ResponseError
 TimeoutError = exceptions.TimeoutError
 WatchError = exceptions.WatchError
+
+__version__: str
+VERSION: tuple[int | str]
```
AlexWaygood, line 64:
```suggestion
VERSION: tuple[int | str, ...]
```

Otherwise you're saying it will always be a tuple of length 1

Dreamsorcerer: Of course

AlexWaygood (approved):

PR #7839, opened 2022-05-16
markdown: `isBlockLevel()` returns a bool value.

leven-cn:


stubs/Markdown/markdown/util.pyi
```diff
@@ -16,7 +16,7 @@ INSTALLED_EXTENSIONS: Any
 RTL_BIDI_RANGES: Any
 
 def deprecated(message: str, stacklevel: int = ...): ...
-def isBlockLevel(tag): ...
+def isBlockLevel(tag: Any) -> bool: ...
```
srittau, line 19: As this accepts literally anything, we should unintuitively use `object` instead of `Any`.

```suggestion
def isBlockLevel(tag: object) -> bool: ...
```

srittau (changes requested):

stubs/Markdown/markdown/util.pyi
leven-cn: According to PEP 484:

> A special kind of type is Any. Every type is consistent with Any. It can be considered a type that has all values and all methods. Note that Any and builtin type object are completely different.
>
> When the type of a value is object, the type checker will reject almost all operations on it, and assigning it to a variable (or using it as a return value) of a more specialized type is a type error. On the other hand, when a value has type Any, the type checker will allow all operations on it, and a value of type Any can be assigned to a variable (or used as a return value) of a more constrained type.

I think using `Any` better than `object`?

```suggestion
def isBlockLevel(tag: Any) -> bool: ...
```

leven-cn: All right. Thanks for reminder.

PR #7844, opened 2022-05-17
`asyncio`: updates for 3.11

JelleZijlstra:
CPython changes:

- https://github.com/python/cpython/commit/13c10bfb777483c7b02877aab029345a056b809c
- https://github.com/python/cpython/commit/9523c0d84f351a610dc651b234461eb015fa3b82
- https://github.com/python/cpython/commit/9f04ee569cebb8b4c6f04bea95d91a19c5403806
- https://github.com/python/cpython/commit/d03acd7270d66ddb8e987f9743405147ecc15087
- https://github.com/python/cpython/commit/195a46d6ffd4cec6c5fb69c5890f8b1758ac91ca


hauntsaninja (approved):

hauntsaninja (commented):
If we merge https://github.com/python/typeshed/pull/7845 first we'll know whether this got everything ;-)

JelleZijlstra:
I have a couple more things to fix tomorrow.

The mypy-primer hit is a true positive: `asyncio.run` only accepts coroutines, not arbitrary awaitables, and that's been true as far back as 3.7.

stdlib/asyncio/exceptions.pyi
```diff
@@ -1,11 +1,23 @@
...
-    "IncompleteReadError",
-    "LimitOverrunError",
-    "SendfileNotAvailableError",
-)
+import sys
+
+if sys.version_info >= (3, 11):
+    __all__ = (
+        "CancelledError",
+        "InvalidStateError",
```
hauntsaninja, line 6:
```suggestion
    __all__ = (
        "BrokenBarrierError",
        "CancelledError",
        "InvalidStateError",
```

stdlib/asyncio/timeouts.pyi
```diff
@@ -0,0 +1,19 @@
+from _typeshed import Self
+from types import TracebackType
+from typing import final
```
AlexWaygood, line 3:
```suggestion
from typing_extensions import final
```

JelleZijlstra: Thanks!

stdlib/asyncio/base_events.pyi
```diff
@@ -86,7 +98,15 @@ class BaseEventLoop(AbstractEventLoop):
     # Future methods
     def create_future(self) -> Future[Any]: ...
     # Tasks methods
-    if sys.version_info >= (3, 8):
+    if sys.version_info >= (3, 11):
+        def create_task(
+            self,
+            coro: Coroutine[Any, Any, _T] | Generator[Any, None, _T],
```
AlexWaygood, line 104: Should we create an alias somewhere (`_typeshed`?) for `Coroutine[Any, Any, _T] | Generator[Any, None, _T]`? I think we could use it quite a lot in our `asyncio` stubs. (No need to do that in this PR.)

AlexWaygood (approved):
Looks good! I read through the whole thing and compared with the runtime. Just a few small questions/comments.

stdlib/asyncio/base_events.pyi
```diff
@@ -390,6 +507,10 @@ class BaseEventLoop(AbstractEventLoop):
         def sock_sendall(self, sock: socket, data: bytes) -> Future[None]: ...
         def sock_connect(self, sock: socket, address: _Address) -> Future[None]: ...
         def sock_accept(self, sock: socket) -> Future[tuple[socket, _RetAddress]]: ...
+    if sys.version_info >= (3, 11):
+        async def sock_recvfrom(self, sock: socket, bufsize: int) -> bytes: ...
+        async def sock_recvfrom_into(self, sock: socket, buf: WriteableBuffer, nbytes: int = ...) -> int: ...
+        async def sock_sendto(self, sock: socket, data: bytes, address: _Address) -> None: ...
```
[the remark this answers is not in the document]
JelleZijlstra: This is also true for the existing `sock_*` methods. I'll add a comment.

stdlib/asyncio/base_events.pyi
JelleZijlstra: It's probably worth checking whether we still need the Generator variant. That feels like a holdover from the pre-`await` days.

stdlib/asyncio/base_events.pyi
```diff
@@ -86,7 +98,15 @@ class BaseEventLoop(AbstractEventLoop):
     # Future methods
     def create_future(self) -> Future[Any]: ...
     # Tasks methods
-    if sys.version_info >= (3, 8):
+    if sys.version_info >= (3, 11):
+        def create_task(
+            self,
+            coro: Coroutine[Any, Any, _T] | Generator[Any, None, _T],
+            *,
+            name: str | None = ...,
```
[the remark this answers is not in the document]
JelleZijlstra: Not sure why I did that, sounds like it should be `object`: https://docs.python.org/3/library/asyncio-task.html#asyncio.Task.set_name

PR #7845, opened 2022-05-17
stubtest: test against 3.11

hauntsaninja:


hauntsaninja:
Okay, this should be good. I haven't yet moved over any comments from py310.txt. I'll do that before we remove positional-only checks from 3.10 / I have some reorganisation in mind.

PR #7852, opened 2022-05-17
Improve I/O types in socket.pyi

srittau:
* Use a protocol for socket.sendfile() "file" argument.
* Use concrete classes for socket.makefile() return types.


stdlib/socket.pyi
```diff
@@ -538,6 +538,16 @@ AI_V4MAPPED_CFG: AddressInfo
...
 
+class _SendableFile(Protocol):
+    def read(self, __size: int) -> bytes: ...
+
+    # optional fields:
+    #
+    # @property
+    # def mode(self) -> str: ...
+    # def fileno(self) -> int: ...
+    # def seek(self, __offset: int) -> object: ...
```
JelleZijlstra, line 549: If offset is given, `_sendfile_use_send` calls this method unconditionally (at least on main). We should perhaps put it in the protocol.

stdlib/socket.pyi
```diff
@@ -549,26 +559,56 @@ class socket(_socket.socket):
...
+        encoding: str | None = ...,
+        errors: str | None = ...,
+        newline: str | None = ...,
+    ) -> SocketIO: ...
+    @overload
     def makefile(
         self,
-        mode: Literal["r", "w", "rw", "wr", ""] = ...,
+        mode: Literal["rwb", "rbw", "wrb", "wbr", "brw", "bwr"],
         buffering: int | None = ...,
```
JelleZijlstra, line 575: This will pick the wrong overload when the user passes something that looks like an `int` to the typechecker but is `0` at runtime. For `builtins.open`, we handle that with a fallback overload that returns `BinaryIO`; we should do that here too.

srittau: I'll use `IOBase` for the fallback for now as I'm interested what stubtest says. Another case for my pet peeve python/typing#566. :)

srittau: @JelleZijlstra primer passes. I'd prefer to use the concrete `IOBase` here, instead of the `BinaryIO` pseudo-protocol. We can always change it later if someone complains. Maybe even returning a union here is alternative. Type checkers should treat it the same as `IOBase`, but it's closer to the implementation.

JelleZijlstra: Sure. We could try these approaches also with `open()`.

JelleZijlstra (approved):

PR #6445, opened 2021-11-29
Add `CodeType.co_lines()` and `CodeType.co_linetable`

AlexWaygood:
PEP 626, introducing these new features, [states](https://www.python.org/dev/peps/pep-0626/#the-new-co-lines-method-of-code-objects):

> ## The new `co_lines()` method of code objects
> The `co_lines()` method will return an iterator which yields tuples of values, each representing the line number of a range of bytecodes. Each tuple will consist of three values:
>
> - `start` -- The offset (inclusive) of the start of the bytecode range
> - `end` -- The offset (exclusive) of the end of the bytecode range
> - `line` -- The line number, or `None` if the bytecodes in the given range do not have a line number.
> The sequence generated will have the following properties:
>
> - The first range in the sequence with have a start of 0
> - The `(start, end)` ranges will be non-decreasing and consecutive. That is, for any pair of tuples the start of the second will equal to the end of the first.
> - No range will be backwards, that is `end >= start` for all triples.
> - The final range in the sequence with have end equal to the size of the bytecode.
> - `line` will either be a positive integer, or None
>
> ## The `co_linetable` attribute
> The `co_linetable` attribute will hold the line number information. The format is opaque, unspecified and may be changed without notice. The attribute is public only to support creation of new code objects.

PR #7856, opened 2022-05-17
CodeType.co_linetable is of type `bytes`

carljm:
Although the specific format of the bytes is opaque and undocumented, the fact that it will be bytes is stable.

This is necessary in order for type checkers to allow creating one code object from another, since the `lnotab`
positional argument to `__init__` is typed as bytes.

### DIFF
--- a/stdlib/types.pyi
+++ b/stdlib/types.pyi
@@ -308,11 +308,11 @@ class CodeType:
             co_cellvars: tuple[str, ...] = ...,
             co_filename: str = ...,
             co_name: str = ...,
-            co_linetable: object = ...,
+            co_linetable: bytes = ...,
         ) -> CodeType: ...
         def co_lines(self) -> Iterator[tuple[int, int, int | None]]: ...
         @property
-        def co_linetable(self) -> object: ...
+        def co_linetable(self) -> bytes: ...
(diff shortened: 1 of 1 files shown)


JelleZijlstra:
This was added in #6445 but it's not clear why we chose `object` there.

JelleZijlstra (approved):

carljm:
Yes, I checked the originating PR to see if there was rationale but didn't see one.

srittau:
Sounds like a good use case for `NewType`?

carljm:
> Sounds like a good use case for NewType?

I don't see the value of using `NewType` here. If someone wants to provide arbitrary bytes for the linetable of a code object they are creating, that's fine and the results are on them; there isn't value in making them wrap it in some arbitrary bytes-alike type that doesn't exist at runtime.

AlexWaygood:
> Yes, I checked the originating PR to see if there was rationale but didn't see one.

It was my PR. I chose `object` because of this paragraph in PEP 626:

> The co_linetable attribute will hold the line number information. The format is opaque, unspecified and may be changed without notice. The attribute is public only to support creation of new code objects.

But your argument is a good one, so I don't object to changing the type to `bytes` :)

PR #7861, opened 2022-05-18
Import SQLAlchemy types in Flask-SQLAlchemy

akakitani:
This pulls in some missing SQLAlchemy types for Flask-SQLAlchemy. Seems like SQLAlchemy may not have been in typeshed when these stubs were first added


stubs/Flask-SQLAlchemy/flask_sqlalchemy/__init__.pyi
```diff
@@ -1,19 +1,18 @@
...
-
 models_committed: Any
 before_models_committed: Any
 
-class SignallingSession(_SessionBase):
+class SignallingSession(Session):
     app: Any
     def __init__(self, db, autocommit: bool = ..., autoflush: bool = ..., **options) -> None: ...
-    def get_bind(self, mapper: Any | None = ..., clause: Any | None = ...): ...
+    def get_bind(self, mapper: Any | None = ..., clause: Any | None = ...): ...  # type: ignore[override]
```
srittau, line 15: Looking at the implementation, this just passes through all of its kwargs, except for the mapper argument. I see two options: Either just copy the arguments from `Session.get_bind()` or just use this for now:

```suggestion
    def get_bind(self, mapper: Any | None = ..., **kwargs): ...
```

stubs/Flask-SQLAlchemy/flask_sqlalchemy/__init__.pyi
```diff
@@ -40,7 +39,9 @@ class Pagination:
         self, left_edge: int = ..., left_current: int = ..., right_current: int = ..., right_edge: int = ...
     ) -> None: ...
 
-class BaseQuery(_Query):
+_T = TypeVar("_T")
+
+class BaseQuery(Query[_T]):
```
srittau, line 44: We should also make this generic, like the base class:

```
class BaseQuery(Query[_T], Generic[_T]):
```

(Needs `Generic` import from `typing`.)

srittau (commented):
Thanks, a few remarks below.

stubs/Flask-SQLAlchemy/flask_sqlalchemy/__init__.pyi
JelleZijlstra: That's not necessary, inheriting from a generic base that is not fully specialized is enough to make the class generic.

stubs/Flask-SQLAlchemy/flask_sqlalchemy/__init__.pyi
akakitani: Ah yeah I was struggling with this a bit, I actually tried both those options and got some failing tests. I think what's happening is that the typeshed tests are using the latest release on pypi (2.5.1), which still has the old signature of just `mapper` and `clause` https://github.com/pallets-eco/flask-sqlalchemy/blob/2.5.1/flask_sqlalchemy/__init__.py#L187

If we go with the `**kwargs` option, would we want to add something to the stubtest allowlist?

akakitani:
If anyone has pointers on what's triggering this pytype failure I'd appreciate it https://github.com/python/typeshed/runs/6484376671?check_suite_focus=true

```
Ran pytype with 2566 pyis, got 2 errors.
stubs/Flask-SQLAlchemy/flask_sqlalchemy/__init__.pyi (3.10): AttributeError: Class not found: sqlalchemy.util.langhelpers.HasMemoized
stubs/SQLAlchemy/sqlalchemy/orm/context.pyi (3.10): AttributeError: Class not found: sqlalchemy.util.langhelpers.MemoizedSlots
```

I'll keep investigating, but I'm not very familiar with pytype

stubs/Flask-SQLAlchemy/flask_sqlalchemy/__init__.pyi
srittau: `**kwargs` should not need an allowlist, as this is how it's actually implemented. I also think there's some special casing around `**kwargs` that disables stubtest.

akakitani: Gave `**kwargs` a shot here https://github.com/python/typeshed/pull/7861/commits/9de1d33ae63065b3034762c8e48eb9f697a81bbc and ran into mypy failures unfortunately - `# type: ignore[override]` here might make it work?
```
stubs/Flask-SQLAlchemy/flask_sqlalchemy/__init__.pyi:15: error: Signature of "get_bind" incompatible with supertype "Session"  [override]
stubs/Flask-SQLAlchemy/flask_sqlalchemy/__init__.pyi:15: note:      Superclass:
stubs/Flask-SQLAlchemy/flask_sqlalchemy/__init__.pyi:15: note:          def get_bind(self, mapper: Optional[Any] = ..., clause: Optional[Any] = ..., bind: Optional[Any] = ..., _sa_skip_events: Optional[Any] = ..., _sa_skip_for_implicit_returning: bool = ...) -> Any
stubs/Flask-SQLAlchemy/flask_sqlalchemy/__init__.pyi:15: note:      Subclass:
stubs/Flask-SQLAlchemy/flask_sqlalchemy/__init__.pyi:15: note:          def get_bind(self, mapper: Optional[Any] = ..., **kwargs: Any) -> Any
```

Tried copying the superclass signature here too https://github.com/python/typeshed/pull/7861/commits/522abc38cf6735360f1c1374e733af2fc739c5d7. In this case what fails is stubtest, since the `**kwargs` signature isn't on a released version yet I think:
```
error: flask_sqlalchemy.SignallingSession.get_bind is inconsistent, runtime does not have argument "bind"
Stub: at line 15
def (self: flask_sqlalchemy.SignallingSession, mapper: Union[Any, None] =, clause: Union[Any, None] =, bind: Union[Any, None] =, _sa_skip_events: Union[Any, None] =, _sa_skip_for_implicit_returning: builtins.bool =) -> Any
Runtime: at line 187 in file /tmp/tmp2rbfa8cj/lib/python3.9/site-packages/flask_sqlalchemy/__init__.py
def (self, mapper=None, clause=None)

error: flask_sqlalchemy.SignallingSession.get_bind is inconsistent, runtime does not have argument "_sa_skip_events"
Stub: at line 15
def (self: flask_sqlalchemy.SignallingSession, mapper: Union[Any, None] =, clause: Union[Any, None] =, bind: Union[Any, None] =, _sa_skip_events: Union[Any, None] =, _sa_skip_for_implicit_returning: builtins.bool =) -> Any
Runtime: at line 187 in file /tmp/tmp2rbfa8cj/lib/python3.9/site-packages/flask_sqlalchemy/__init__.py
def (self, mapper=None, clause=None)

error: flask_sqlalchemy.SignallingSession.get_bind is inconsistent, runtime does not have argument "_sa_skip_for_implicit_returning"
Stub: at line 15
def (self: flask_sqlalchemy.SignallingSession, mapper: Union[Any, None] =, clause: Union[Any, None] =, bind: Union[Any, None] =, _sa_skip_events: Union[Any, None] =, _sa_skip_for_implicit_returning: builtins.bool =) -> Any
Runtime: at line 187 in file /tmp/tmp2rbfa8cj/lib/python3.9/site-packages/flask_sqlalchemy/__init__.py
def (self, mapper=None, clause=None)


Ran with the following environment:
click==8.1.3
Flask==2.1.2
Flask-SQLAlchemy==2.5.1
greenlet==1.1.2
importlib-metadata==4.11.3
itsdangerous==2.1.2
Jinja2==3.1.2
MarkupSafe==2.1.1
mypy==0.950
mypy-extensions==0.4.3
SQLAlchemy==1.4.36
tomli==2.0.1
types-SQLAlchemy==1.4.34
typing_extensions==4.2.0
Werkzeug==2.1.2
zipp==3.8.0
``` 

Any preference on what approach to go forward with?

srittau (approved):

PR #6114, opened 2021-10-05
Use overrides for IntFlag dunder-r methods

jakebailey:
For #6110.

~~I don't think this should be necessary, but pyright 1.1.175 reports:~~

```
/home/runner/work/typeshed/typeshed/stdlib/enum.pyi
  /home/runner/work/typeshed/typeshed/stdlib/enum.pyi:78:15 - error: Expression of type "(self: _T@__or__, other: int | _T@__or__) -> _T@__or__" cannot be assigned to declared type "(self: int, n: int) -> int"
    Type "(self: _T@__or__, other: int | _T@__or__) -> _T@__or__" cannot be assigned to type "(self: int, n: int) -> int"
      Parameter name mismatch: "n" versus "other" (reportGeneralTypeIssues)
  /home/runner/work/typeshed/typeshed/stdlib/enum.pyi:79:16 - error: Expression of type "(self: _T@__and__, other: int | _T@__and__) -> _T@__and__" cannot be assigned to declared type "(self: int, n: int) -> int"
    Type "(self: _T@__and__, other: int | _T@__and__) -> _T@__and__" cannot be assigned to type "(self: int, n: int) -> int"
      Parameter name mismatch: "n" versus "other" (reportGeneralTypeIssues)
  /home/runner/work/typeshed/typeshed/stdlib/enum.pyi:80:16 - error: Expression of type "(self: _T@__xor__, other: int | _T@__xor__) -> _T@__xor__" cannot be assigned to declared type "(self: int, n: int) -> int"
    Type "(self: _T@__xor__, other: int | _T@__xor__) -> _T@__xor__" cannot be assigned to type "(self: int, n: int) -> int"
      Parameter name mismatch: "n" versus "other" (reportGeneralTypeIssues)
```

~~Here, values of type `_T` inherit from int, but this doesn't seem to make it assignable, yet, overriding it seems to be allowed.~~

@erictraut

PR #7862, opened 2022-05-18
`enum`: More changes for 3.11

AlexWaygood:
In 3.11+, `enum.Flag`s are iterable and sized:

```python
C:\Users\alexw\coding\cpython>python
Running Debug|x64 interpreter...
Python 3.12.0a0 (main, May 17 2022, 16:54:26) [MSC v.1931 64 bit (AMD64)] on win32
>>> from enum import *
>>> class Foo(Flag):
...     X = 1
...     Y = 2
...     Z = X | Y
...
>>> len(Foo.X)
1
>>> len(Foo.Z)
2
>>> list(Foo.X)
[<Foo.X: 1>]
>>> list(Foo.Z)
[<Foo.X: 1>, <Foo.Y: 2>]
```


stdlib/enum.pyi
```diff
@@ -213,15 +223,21 @@ class Flag(Enum):
...
     def __new__(cls: type[Self], value: int) -> Self: ...
     def __or__(self: Self, other: int) -> Self: ...
     def __and__(self: Self, other: int) -> Self: ...
     def __xor__(self: Self, other: int) -> Self: ...
-    def __ror__(self: Self, other: int) -> Self: ...
-    def __rand__(self: Self, other: int) -> Self: ...
-    def __rxor__(self: Self, other: int) -> Self: ...
+    __ror__ = __or__
+    __rand__ = __and__
+    __rxor__ = __xor__
```
AlexWaygood, line 240: This brings the stub closer to the runtime, which uses assignments for these methods. It reverts changes that were made in #6114 by @jakebailey -- it looks like the changes in that PR were made to make pyright happy? Not sure.

jakebailey: Hm, I can see now that it was caused by a parameter name clash with a parent class; I'd go check to see if they still differ... It seems weird that we wouldn't report that, unless that rule has been disabled?

AlexWaygood: Ah, I think we might have disabled the relevant check for typeshed: https://github.com/python/typeshed/pull/7498

In typeshed, we generally prefer to match the implementation at runtime, even if the runtime has... dubious respect for LSP in some cases :)

jakebailey: This isn't an LSP thing, it's an assignability error because named parameters change how a caller is able to use the function (unless they are marked as positional only). A call site with assuming the names of the parent class will crash on a subclass which changes those names.

The saving grace here is that this is a dunder function that isn't often called directly.

I would hope this rule can be reenabled (and lines like this use an ignore which explicitly mention the rule name or something). It's not great to leave this rule off.

AlexWaygood: > I would hope this rule can be reenabled (and lines like this use an ignore which explicitly mention the rule name or something). It's not great to leave this rule off.

Again, I'm not defending the implementation necessarily! (Though I will note that using method assignment like this is pretty idiomatic Python imo, especially for reflected dunders.) But bugs in the implementation should be reported over at the CPython repo -- our job over here at typeshed is to reflect the runtime. I don't think having lots of `type: ignore`s because of this pyright check particularly helps us to do that :)

JelleZijlstra (approved):

PR #7863, opened 2022-05-18
Improve output of `mypy_test` when reporting errors

AlexWaygood:
Currently, all errors are printed to stderr, which can look pretty confusing -- in https://github.com/python/typeshed/runs/6470488996?check_suite_focus=true, mypy emitted an error when running on the test suite, but the error message was printed to a completely different location to the announcement that mypy was about to start running on the test suite.

I've deliberately broken a few bits of typeshed to show what `mypy_test` output will look like with these changes. I'll unbreak those changes as soon as the CI run is complete.

### DIFF
--- a/stdlib/builtins.pyi
+++ b/stdlib/builtins.pyi
@@ -25,18 +25,7 @@ from _typeshed import (
     SupportsTrunc,
     SupportsWrite,
 )
-from collections.abc import (
-    Awaitable,
-    Callable,
-    Iterable,
-    Iterator,
-    MutableSequence,
-    MutableSet,
-    Reversible,
-    Sequence,
-    Set as AbstractSet,
-    Sized,
-)
+from collections.abc import Awaitable, Callable, Iterable, Iterator, MutableSet, Reversible, Set as AbstractSet, Sized
 from io import BufferedRandom, BufferedReader, BufferedWriter, FileIO, TextIOWrapper
 from types import CodeType, TracebackType, _Cell
 
@@ -50,8 +39,10 @@ from typing import (  # noqa: Y027
     Generic,
     Mapping,
     MutableMapping,
+    MutableSequence,
     NoReturn,
     Protocol,
+    Sequence,
     SupportsAbs,
     SupportsBytes,
     SupportsComplex,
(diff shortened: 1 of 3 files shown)


srittau (approved):

srittau:
We could adapt the format I used for stubtest: https://github.com/python/typeshed/runs/6480640648?check_suite_focus=true (open the "Run stubtest" section). I think that's well readable and confirms that the tests (and which tests) were run.

AlexWaygood:
> What do you think of removing the lines like `testing protobuf (37 files)..`? If there's no error in that library, it's just noise that makes it harder to find the error.

Hmm, I quite like those when running the test locally. Testing the `stubs` directory can take a while; I kinda like the reassurance that something is actually happening :)

AlexWaygood:
> We could adapt the format I used for stubtest: https://github.com/python/typeshed/runs/6480640648?check_suite_focus=true (open the "Run stubtest" section). I think that's well readable and confirms that the tests (and which tests) were run.

Actually yes, this is very pretty. And the coloured output is very helpful in identifying errors as opposed to other kinds of output!

AlexWaygood:
> What do you think of removing the lines like `testing protobuf (37 files)..`? If there's no error in that library, it's just noise that makes it harder to find the error.

Relatedly, does anybody have any idea why I get a lot more output whenever I run the test locally than we seem to get in CI? When I run the test locally I get repeated spam about there being "no [mypy] section in config file":

image

Akuli:
A couple things that might be causing it:
- In `mypy/config_parser.py`, mypy calls `config_parser.read(config_file)`. This method silences errors, and returns a list of successfully read files. Try printing what it returns (after installing mypy from source with e.g. `pip install -e .\mypy`).
- I wonder if `temp.flush()` in `tests/mypy_test.py` is enough to get the content showing there correctly. You could try:
    ```diff
    --- a/tests/mypy_test.py
    +++ b/tests/mypy_test.py
    @@ -174,6 +174,8 @@ def run_mypy(
                     temp.write("{} = {}\n".format(k, v))
             temp.flush()
     
    +        print(Path(temp.name).read_text())
    +
             flags = get_mypy_flags(args, major, minor, temp.name, custom_typeshed=custom_typeshed)
             mypy_args = [*flags, *files]
             if args.verbose:
    ```

PR #7864, opened 2022-05-18
stubtest third party: remove extra line when skipping a distribution

srittau:


AlexWaygood (approved):

srittau:
For reference: https://github.com/python/typeshed/runs/6480640648?check_suite_focus=true (open "Run stubtest").

PR #7868, opened 2022-05-18
Add return type for `dateutil.rrule._rrulestr.__call__`.

emosenkis:
Based on inferred types at each of the `return` statements in the implementation.


AlexWaygood (approved):
Union types for return-annotations can be annoying in some situations, but here I think it's okay since they have a common base class

PR #7870, opened 2022-05-18
Introduce _typeshed.Maybe

srittau:
We use the `X | Any` pattern for return types more and more. Introducing a type alias legitimizes the pattern, makes it easier to recognize, easier to grep, and means that we don't need to repeat the documentation.


stdlib/_typeshed/__init__.pyi
```diff
@@ -33,6 +33,17 @@ Self = TypeVar("Self")  # noqa: Y001
 # "Incomplete | None" instead of "Any | None".
 Incomplete: TypeAlias = Any
 
+# Annotation for return types that can return a value or None, but where the
+# caller should not be forced to check for None. For example:
+#
+#   if m := re.match(r"(a)(b)?"):
```
Akuli, line 39:
```suggestion
#   if m := re.match(r"(a)(b)?", string):
```

Akuli (commented):
I like getting rid of the comment copy/pasta.

I'm not sure if `Maybe` is a good name for this, since it means "could be None and you must check for it" in some other languages, but I don't have better suggestions in mind.

srittau:
> I'm not sure if `Maybe` is a good name for this, since it means "could be None and you must check for it" in some other languages, but I don't have better suggestions in mind.

True, but it's succinct and *has* the meaning "could be unset/null/none/undefined" in other languages, so the name at least has some meaning, even if it's not a 100% match to what other languages do.

hauntsaninja:
We've talked about WeakUnion before, so WeakOptional seems a pretty natural name to me

srittau:
I don't think we should use "optional". That name has caused lots of confusion in the past. Also, it's a fairly long name.

Akuli:
`RarelyNone`?

srittau:
We could bikeshed this forever, but I still think `Maybe` is the best option.

Akuli (approved):

AlexWaygood:
> We could bikeshed this forever

Indeed we could, but I do think the choice of name is important here -- I'd prefer the status quo over a confusing alias name. Personally, I prefer either `Probably` or `RarelyNone` to `Maybe` 🙂

srittau:
`Probably` and `RarelyNone` suggest that `None` is somehow the less common case. That's not true in many cases, for example `Match.group()` with optional groups.

AlexWaygood:
`Often`? That has less of an implication that the result is necessarily the _majority_ of cases.

Akuli:
I think the name shouldn't be too clever or short. If you see `Often` in the middle of some code, how are you supposed to know that it has something to do with unions and `None`, let alone guess its meaning?

Maybe we should just call this `TheAnyTrick`. It's very obvious for typeshed maintainers. It isn't obvious at first for people new to typing, but this is a difficult concept, and I don't think we can make it guessable anyway.

AlexWaygood:
> Maybe we should just call this `TheAnyTrick`. It's very obvious for typeshed maintainers. It isn't obvious at first for people new to typing, but this is a difficult concept, and I don't think we can make it guessable anyway.

Or maybe just `AnyOr`? `Any | str` becomes `AnyOr[str]`

AlexWaygood:
> Maybe we should just call this `TheAnyTrick`. It's very obvious for typeshed maintainers. It isn't obvious at first for people new to typing, but this is a difficult concept, and I don't think we can make it guessable anyway.

I'd be fine with `AnyTrick`. I agree it's better to have an opaque name than a name that's too clever or succinct.

srittau:
No consensus.

PR #7871, opened 2022-05-18
Use `dict` for requests.Session().hooks

### LINKED ISSUE #7776: Regression in requests hooks stubs
Since https://github.com/python/typeshed/commit/f330e12f3ef6ca3deae6936ea7b0844d9a72f2ae it's not possible to add new hooks to `requests.Session`.

According to the [documentation](https://docs.python-requests.org/en/latest/user/advanced/#event-hooks), the recommended way to add a hook to every request is by adding them to `requests.Session.hook`:

```python
s = requests.Session()
s.hooks['response'].append(print_url)
```

This snippet now fails with the following error:

```
error: Unsupported target for indexed assignment ("Mapping[str, Union[Callable[[Response], Any], List[Callable[[Response], Any]]]]")
```

## Example project

source: https://gitlab.com/cki-project/cki-lib/-/blob/main/cki_lib/session.py#L47

```python
    response_hooks: typing.List[typing.Callable[[requests.Response], typing.Any]] = []
    session.hooks['response'] = response_hooks
```

CI job: https://gitlab.com/cki-project/cki-lib/-/jobs/2407190024

```
cki_lib/session.py:47: error: Unsupported target for indexed assignment ("Mapping[str, Union[Callable[[Response], Any], List[Callable[[Response], Any]]]]")
```

Akuli:
Fixes #7776

Mutating hooks, as in `session.hooks['response'] = ...`, should work. Reassigning it like `session.hooks = ...` is probably a bad idea, so it will always be a `dict`.

### DIFF
--- a/stubs/requests/requests/sessions.pyi
+++ b/stubs/requests/requests/sessions.pyi
@@ -55,7 +55,7 @@ _Files: TypeAlias = (
     | Mapping[str, tuple[str | None, SupportsRead[str | bytes] | str | bytes, str, _TextMapping]]
 )
 _Hook: TypeAlias = Callable[[Response], Any]
-_Hooks: TypeAlias = Mapping[str, _Hook | list[_Hook]]
+_Hooks: TypeAlias = dict[str, _Hook | list[_Hook]]
 _HooksInput: TypeAlias = Mapping[str, Iterable[_Hook] | _Hook]
 
 _ParamsMappingKeyType: TypeAlias = str | bytes | int | float


Akuli:
We still need to discuss whether it's better to use a union or just require hooks to be lists. The usual reason for avoiding union return types applies here: if we merge this PR as is, #7776 still wouldn't work, because it would end up calling `.append()` on a `_Hook | list[_Hook]`. https://github.com/python/typeshed/issues/7776#issuecomment-1130566193

AlexWaygood (approved):
Makes sense

JelleZijlstra (approved):

PR #7873, opened 2022-05-18
Add overloads for `asyncio.sleep()`.

### LINKED ISSUE #7866: `asyncio.sleep()` should have an overload
This is from https://github.com/microsoft/pyright/issues/3475.

With the latest version of pyright (1.1.247), this code results in an error:

```py
from typing import Awaitable, TypeVar
import asyncio
T = TypeVar("T")

async def wrap(coro: Awaitable[T]) -> T:
    return await coro

async def wrap_sleep() -> None:
    await wrap(asyncio.sleep(0))
```

The result when I run `pyright` on it:

```
Found 1 source file
/..../test/test.py
  /..../test/test.py:9:16 - error: Argument type is partially unknown
    Argument corresponds to parameter "coro" in function "wrap"
    Argument type is "Coroutine[Any, Any, Unknown]" (reportUnknownArgumentType)
1 error, 0 warnings, 0 informations 
```

The fix, as suggested by @JelleZijlstra and @erictraut, is to add an overload like this:

```py
 async def sleep(delay: float) -> None: ...
```

wch:
Closes #7866. This adds an overload to `asyncio.sleep()`, so that when it is called _without_ `return=None`, the type checker knows that the return type is `None` instead of `unknown`.

Also related to https://github.com/microsoft/pyright/issues/3475.


stdlib/asyncio/tasks.pyi
```diff
@@ -270,6 +270,9 @@ def run_coroutine_threadsafe(coro: _FutureT[_T], loop: AbstractEventLoop) -> con
 
 if sys.version_info >= (3, 10):
     def shield(arg: _FutureT[_T]) -> Future[_T]: ...
+    @overload
+    async def sleep(delay: float) -> None: ...
+    @overload
     async def sleep(delay: float, result: _T = ...) -> _T: ...
```
JelleZijlstra, line 276:
```suggestion
    async def sleep(delay: float, result: _T) -> _T: ...
```
So the overloads don't overlap.

stdlib/asyncio/tasks.pyi
```diff
@@ -281,6 +284,9 @@ if sys.version_info >= (3, 10):
 
 else:
     def shield(arg: _FutureT[_T], *, loop: AbstractEventLoop | None = ...) -> Future[_T]: ...
+    @overload
+    async def sleep(delay: float, *, loop: AbstractEventLoop | None = ...) -> None: ...
+    @overload
     async def sleep(delay: float, result: _T = ..., *, loop: AbstractEventLoop | None = ...) -> _T: ...
```
JelleZijlstra, line 290:
```suggestion
    async def sleep(delay: float, result: _T, *, loop: AbstractEventLoop | None = ...) -> _T: ...
```

wch:
Looking for a bit of advice. This part fails the CI checks on Python <= 3.9. (Python >= 3.10 uses a different code path which seems to be fine.)

```py
    @overload
    async def sleep(delay: float, *, loop: AbstractEventLoop | None = ...) -> None: ...
    @overload
    async def sleep(delay: float, result: _T = ..., *, loop: AbstractEventLoop | None = ...) -> _T: ...
```

```
stdlib/asyncio/tasks.pyi:288: error: Overloaded function signatures 1 and 2 overlap with incompatible return types  [misc]
```

Is it possible to make this work?

AlexWaygood:
> Is it possible to make this work?

Committing @JelleZijlstra's suggestions he made in review just now should do it :)

wch:
@AlexWaygood Thanks, he must have posted that as I was typing. :)

stdlib/asyncio/tasks.pyi
wch: This one doesn't seem to cause errors in CI. Should I still commit this change?

AlexWaygood: I just committed it for you :) I'm not sure why it's not causing errors in CI. I feel like it should do.

JelleZijlstra (approved):

PR #7877, opened 2022-05-19
ast: fix stubtest issues

JelleZijlstra:


AlexWaygood (approved):

hauntsaninja:
Probably worth fixing CPython for the ImportFrom thing, I can do that...

PR #7872, opened 2022-05-18
Add colour to the output of `mypy_test`

AlexWaygood:
And gracefully exit on `KeyboardInterrupt`.

PR #7879, opened 2022-05-19
Add colour to the output of `mypy_test`, take 2

AlexWaygood:
A second attempt at #7872


AlexWaygood:
Samples of what this looks like in CI (and locally):

- https://github.com/python/typeshed/runs/6505698999?check_suite_focus=true
- https://github.com/python/typeshed/runs/6505740565?check_suite_focus=true

tests/mypy_test.py
```diff
@@ -183,9 +186,24 @@ def run_mypy(
...
+                captured_stdout = stdout_redirect.getvalue()
+                captured_stderr = stderr_redirect.getvalue()
+                if returned_stderr:
+                    print_error(returned_stderr)
+                if captured_stderr:
+                    print_error(captured_stderr)
+                if returned_stdout:
+                    print_error(returned_stdout)
+                if captured_stdout:
+                    print_error(captured_stdout, end="")
```
srittau, line 204: Should we really color the whole error output red? This could be a bit hard to read.

tests/stubtest_third_party.py
```diff
@@ -100,7 +88,7 @@ def run_stubtest(dist: Path, *, verbose: bool = False) -> bool:
         try:
             subprocess.run(stubtest_cmd, env={"MYPYPATH": str(dist), "MYPY_FORCE_COLOR": "1"}, check=True, capture_output=True)
         except subprocess.CalledProcessError as e:
-            print(colored(" fail", "red"))
+            print_error(" fail")
```
srittau, line 91:
```suggestion
            print_error("fail")
```

srittau (commented):
LGTM, but one suggestion and one nit

tests/stubtest_third_party.py
AlexWaygood: Thanks!

tests/mypy_test.py
AlexWaygood: Hmm. We could just colour the `failure` message at the beginning red. But I don't find https://github.com/python/typeshed/runs/6505698999?check_suite_focus=true too hard to read, and I like how easy it is for me to see where the error messages start and where they end.

srittau (approved):

PR #7883, opened 2022-05-19
Only run CI on pushes to main/master and pull requests

srittau:
Currently, when a maintainer opens a pull request from a branch in the
typeshed repository (instead of a fork), CI is run twice: once for the
PR and once for the push. This not only wastes CI minutes (and it seems
we're running into some limits lately), but also makes CI output twice
as large and more difficult to process.

The disadvantage is that a PR needs to be opened to run CI.
Alternatively, CI can be triggered manually.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -3,6 +3,9 @@ name: Test
 on:
   workflow_dispatch:
   push:
+    branches:
+      - main
+      - master
   pull_request:
 
 permissions:


AlexWaygood:
> Alternatively, CI can be triggered manually.

Oh, how do you do that? I wasn't aware (and I've found it useful in the past to run CI on a branch).

We can also consider only running `stubtest_stdlib.py` on PRs that touch the stdlib, `tests.yml`, and/or `tests/stdlib.py` itself.

AlexWaygood:
> Oh, how do you do that?

Instructions here, for future reference: https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow.

It's pretty easy, so I'm fine with this change.

AlexWaygood (approved):

.github/workflows/tests.yml
```diff
@@ -3,6 +3,9 @@ name: Test
 on:
   workflow_dispatch:
   push:
+    branches:
+      - main
```
Akuli, line 7: We currently don't have a `main` branch, so this is redundant. If we some day decide to rename `master` to `main`, we'll have to grep for `master` anyway.

Akuli (commented):
In the past, I have used a similar configuration in other projects, and it works well. Just one nit.

PR #7884, opened 2022-05-19
Update emoji (3rd party) library

RensOliemans:
I'm new to typeshed, hopefully this is OK.


RensOliemans:
Wow, checking the stubs with the actual third party code is awesome! I have a question on how I can solve this error.

The error given is:

    error: emoji.emoji_lis is inconsistent, runtime argument "language" has a default value of type builtins.type, which is incompatible with stub argument type builtins.str

Which makes sense, since they changed the `language` argument to default to `_DeprecatedParameter` which is a custom class of theirs. See [here](https://github.com/carpedm20/emoji/blob/master/emoji/core.py#L294) for their code. How can I refer to their custom class as a type?

AlexWaygood:
> I'm new to typeshed, hopefully this is OK.

Welcome to the project! PRs improving our stubs are very welcome here :)

> Concretely, how can I refer to their custom class as a type?

You can do it like this:

```python
class _DeprecatedParameter: ...

def demojize(
    string: str,
    use_aliases: bool | type[_DeprecatedParameter] = ...,
    delimiters: tuple[str, str] = ...,
    language: str = ...,
    version: float | int | None = ...,
    handle_version: str | Callable[[str, dict[str, str]], str] | None = ...,
) -> str: ...
```

Since the `use_aliases` parameter here is, well, deprecated, we could also consider just removing it from the stub (there are ways of getting stubtest to shut up about small discrepancies compared to the runtime). But it's usually best to stick as closely to the runtime as possible, and I think that's what I'd prefer here.

RensOliemans:
> Welcome to the project! PRs improving our stubs are very welcome here :)

Thanks!
 
> Since the `use_aliases` parameter here is, well, deprecated, we could also consider just removing it from the stub (there are ways of getting stubtest to shut up about small discrepancies compared to the runtime). But it's usually best to stick as closely to the runtime as possible, and I think that's what I'd prefer here.

Alright makes sense, thanks for the review and tip!

stubs/emoji/emoji/core.pyi
```diff
@@ -1,17 +1,38 @@
...
+class _DeprecatedParameter: ...
+
 def emojize(
     string: str,
-    use_aliases: bool = ...,
+    use_aliases: bool | type[_DeprecatedParameter] = ...,
     delimiters: tuple[str, str] = ...,
     variant: Literal["text_type", "emoji_type", None] = ...,
     language: str = ...,
+    version: float | int | None = ...,
```
AlexWaygood, line 15: PEP 484 specifies that type checkers should treat `int` as an implicit subtype of `float`, so `int` is redundant in a union with `float`

```suggestion
    version: float | None = ...,
```

stubs/emoji/emoji/core.pyi
```diff
@@ -1,17 +1,38 @@
...
+    handle_version: str | Callable[[str, dict[str, str]], str] | None = ...,
+) -> str: ...
+def demojize(
+    string: str,
+    use_aliases: bool | type[_DeprecatedParameter] = ...,
+    delimiters: tuple[str, str] = ...,
+    language: str = ...,
+    version: float | int | None = ...,
```
AlexWaygood, line 23: Same as above

```suggestion
    version: float | None = ...,
```

stubs/emoji/emoji/core.pyi
```diff
@@ -1,17 +1,38 @@
...
+    handle_version: str | Callable[[str, dict[str, str]], str] | None = ...,
+) -> str: ...
+def replace_emoji(
+    string: str,
+    replace: str | Callable[[str, dict[str, str]], str] = ...,
+    language: str | type[_DeprecatedParameter] = ...,
+    version: float | int | None = ...,
```
AlexWaygood, line 30: Same as above

```suggestion
    version: float | None = ...,
```

stubs/emoji/emoji/core.pyi
srittau, line 15: `int` is redundant with `float` in argument types:

```suggestion
    version: float | None = ...,
```

@AlexWaygood Maybe that's a check we could add to flake8-pyi?

stubs/emoji/emoji/core.pyi
```diff
@@ -1,17 +1,38 @@
...
-def emoji_lis(string: str, language: str = ...) -> list[dict[str, int | str]]: ...
-def distinct_emoji_lis(string: str) -> list[str]: ...
-def emoji_count(string: str) -> int: ...
+def get_emoji_regexp(language: str | None = ...) -> Pattern[str]: ...
+def emoji_lis(string: str, language: str | type[_DeprecatedParameter] = ...) -> list[dict[str, int | str]]: ...
+def emoji_list(string: str) -> list[dict[str, int | str]]: ...
+def distinct_emoji_lis(string: str, language: str | type[_DeprecatedParameter] = ...) -> list[str]: ...
+def distinct_emoji_list(string: str) -> list[str]: ...
+def emoji_count(string: str, unique: bool = ...) -> int: ...
+def version(string: str) -> float | int: ...
```
AlexWaygood, line 38: Same as above

```suggestion
def version(string: str) -> float: ...
```

stubs/emoji/emoji/core.pyi
srittau, line 23:
```suggestion
    version: float | None = ...,
```

stubs/emoji/emoji/core.pyi
srittau, line 30:
```suggestion
    version: float | None = ...,
```

stubs/emoji/emoji/core.pyi
```diff
@@ -1,17 +1,38 @@
...
+    language: str | type[_DeprecatedParameter] = ...,
+    version: float | int | None = ...,
 ) -> str: ...
-def demojize(string: str, use_aliases: bool = ..., delimiters: tuple[str, str] = ..., language: str = ...) -> str: ...
-def get_emoji_regexp(language: str = ...) -> Pattern[str]: ...
-def emoji_lis(string: str, language: str = ...) -> list[dict[str, int | str]]: ...
-def distinct_emoji_lis(string: str) -> list[str]: ...
-def emoji_count(string: str) -> int: ...
+def get_emoji_regexp(language: str | None = ...) -> Pattern[str]: ...
+def emoji_lis(string: str, language: str | type[_DeprecatedParameter] = ...) -> list[dict[str, int | str]]: ...
```
AlexWaygood, line 33: You could consider returning a list of `TypedDict`s here: the type associated with the `location` key will always be `int`; the type associated with the `emoji` key will always be `str`. https://docs.python.org/3/library/typing.html#typing.TypedDict.

Same point applies for the `emoji_list` function below.

srittau (changes requested):
Thanks, just a few nits below.

AlexWaygood (commented):
Thanks, this is great! Just a few small points to fix:

stubs/emoji/emoji/core.pyi
AlexWaygood: Whooops, looks like we did simultaneous reviews! And, yeah, we could, certainly :)

RensOliemans:
Thanks for the useful review tips! I have added `TypedDict` for the `emoji_lis` and `emoji_list` return types.

AlexWaygood (approved):
Stubtest's happy, primer's happy, I'm happy. Thanks again!

PR #7886, opened 2022-05-20
traceback: python3.11 additions

JelleZijlstra:


stdlib/traceback.pyi
```diff
@@ -108,6 +114,33 @@ class TracebackException:
     offset: int
     msg: str
     if sys.version_info >= (3, 10):
```
hauntsaninja, line 116:
```suggestion
    if sys.version_info >= (3, 11):
```

JelleZijlstra: Thanks, just pushed

hauntsaninja (approved):

PR #7893, opened 2022-05-20
`Pillow`: use union type for `Image.paste`

### LINKED ISSUE #7778: Pillow Image.Image.paste() stub is incorrect
I'm getting this error from mypy:
`Argument 1 to "paste" of "Image" has incompatible type "int"; expected "Image"`
The method paste on the Image class does accept an image for the first argument but it also accepts other things mentioned [here in the docs](https://pillow.readthedocs.io/en/stable/reference/Image.html#PIL.Image.Image.paste):

> Instead of an image, the source can be a integer or tuple containing pixel values. The method then fills the region with the given color. When creating RGB images, you can also use color strings as supported by the ImageColor module.

jkklapp:
closes #7778


stubs/Pillow/PIL/Image.pyi
```diff
@@ -163,7 +163,9 @@ class Image:
     def getprojection(self) -> tuple[list[int], list[int]]: ...
     def histogram(self, mask: Image | None = ..., extrema: tuple[int, int] | tuple[float, float] | None = ...) -> list[int]: ...
     def entropy(self, mask: Image | None = ..., extrema: tuple[int, int] | tuple[float, float] | None = ...) -> float: ...
-    def paste(self, im: Image, box: tuple[float, float] | _Box | None = ..., mask: Image | None = ...) -> None: ...
+    def paste(
+        self, im: Image | tuple[int, int] | int, box: tuple[float, float] | _Box | None = ..., mask: Image | None = ...
+    ) -> None: ...
```
srittau, line 168: One thing I noticed when I just tested this is that the `box` argument must be given when `im` is an `int` (and I assume a tuple). We could express this using overloads, but I'm also fine accepting this as is for now.

AlexWaygood:
This one is a bit more complex than I realised when I added the "good-first-issue" label on the issue.

I'm not sure that using a tuple of length 2 here is correct. https://pillow.readthedocs.io/en/stable/handbook/concepts.html seems to imply that a "pixel value" can be a tuple of length 3 or 4 (but not of length 2). For other "pixel values" in this stub, we use the `_Color` type alias that we define on line 25. But that's currently defined as `_Color = float | tuple[float, ...]`, and I'm not sure that's correct (it looks like it should probably be `_Color = Union[int, tuple[int, int, int], tuple[int, int, int, int]]` instead, at least for this case).

stubs/Pillow/PIL/Image.pyi
```diff
@@ -163,7 +163,9 @@ class Image:
     def getprojection(self) -> tuple[list[int], list[int]]: ...
     def histogram(self, mask: Image | None = ..., extrema: tuple[int, int] | tuple[float, float] | None = ...) -> list[int]: ...
     def entropy(self, mask: Image | None = ..., extrema: tuple[int, int] | tuple[float, float] | None = ...) -> float: ...
-    def paste(self, im: Image, box: tuple[float, float] | _Box | None = ..., mask: Image | None = ...) -> None: ...
+    def paste(
+        self, im: Image | tuple[int, int] | int, box: tuple[float, float] | _Box | None = ..., mask: Image | None = ...
```
AlexWaygood, line 167: For now, this is probably the safest change to make, @jkklapp

```suggestion
        self, im: Image | _Color, box: tuple[float, float] | _Box | None = ..., mask: Image | None = ...
```

AlexWaygood (approved):
Thanks!

PR #7894, opened 2022-05-20
Ignore mypy errors in Python 2 builtins and typing

JukkaL:
Similar errors are already ignored in Python 3 stubs. The
errors break various Python 2 tests of mypy.

See https://github.com/python/typeshed/issues/7367#issuecomment-1131472225 for context.


AlexWaygood (approved):
I assume the reason why we didn't catch this in our CI is because of this line in `mypy_test.py`:

https://github.com/python/typeshed/blob/adf48f1f9653ad347887ea5ca4190d5ddf041d50/tests/mypy_test.py#L337

PR #7895, opened 2022-05-20
CI: Don't run stubtest_stdlib on PRs that don't touch the stdlib

AlexWaygood:
Also, refine a little more the paths that `mypy_primer` runs on

### DIFF
--- a/.github/workflows/mypy_primer.yml
+++ b/.github/workflows/mypy_primer.yml
@@ -6,8 +6,7 @@ on:
     paths:
     - 'stdlib/**'
     - 'stubs/**'
-    - '.github/workflows/mypy_primer.yml'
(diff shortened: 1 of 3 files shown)


srittau (approved):

AlexWaygood:
Not sure I understand why stubtest_stdlib isn't running on this PR itself; I feel like it should do, since I'm altering files in the `.github/workflows` directory

srittau:
> Not sure I understand why stubtest_stdlib isn't running on this PR itself; I feel like it should do, since I'm altering files in the `.github/workflows` directory

Possibly because the workflow is not in master yet, but that's only a speculation. I'd suggest to merge and open an experimental PR to check whether it works correctly and revert if it doesn't.

PR #7896, opened 2022-05-20
[DO-NOT-MERGE] Testing #7895

AlexWaygood:
A PR to test the changes made in #7895

PR #7897, opened 2022-05-20
CI: Fix stubtest workflow; make paths for other workflows more specific

AlexWaygood:
Reverts python/typeshed#7895. The new workflow doesn't seem to be being triggered: see #7896


.github/workflows/stubtest_stdlib.yml
```diff
@@ -1,43 +0,0 @@
-name: Run stubtest on the stdlib
-
-on:
-  workflow_dispatch:
-  push:
-    branches:
-      - main
-      - master
-    pull_request:
```
JelleZijlstra, line 9: Is this overindented? In tests.yml it's at the same level as `push:`.

srittau: Good spot! I believe this is the reason.

AlexWaygood: Oh wow. 🤦 Thanks!

AlexWaygood: Yup, that was it!

JelleZijlstra (approved):

.github/workflows/stubtest_stdlib.yml
```diff
@@ -1,16 +1,16 @@
...
-      paths:
-        - 'stdlib/**'
-        - '.github/workflows/**'
-        - 'tests/**'
+  pull_request:
+    paths:
+      - 'stdlib/**'
+      - '.github/workflows/stubtest_stdlib.yml'
+      - 'tests/stubtest_stdlib.py'
```
JelleZijlstra, line 13: I'd rather just keep this as `tests/**`, in case we update the test file to use colors.py or something.

AlexWaygood: Fair point, I wasn't 100% sure about that one.

AlexWaygood:
```suggestion
      - 'tests/**'
```

JelleZijlstra (approved):

PR #7899, opened 2022-05-20
stubtest: put all the incorrect __get__ sigs together

JelleZijlstra:


AlexWaygood (approved):

JelleZijlstra:
Explicitly linking to python/cpython#93021

PR #7900, opened 2022-05-20
types: 3.11 (and earlier) fixes

JelleZijlstra:
- New attributes in 3.11
- Arguments to CodeType.__init__ are always positional (have fun)
- co_consts can use object instead of Any


Akuli:
> Arguments to `CodeType.__init__` are always positional (have fun)

:DDDDDD

PR #7905, opened 2022-05-21
Add a JSON alias

srittau:
Cf. python/typing#182

I have deliberately not added this to any existing stubs, since that might be disruptive and needs experimentation independent from the alias itself.

### DIFF
--- a/stdlib/_typeshed/__init__.pyi
+++ b/stdlib/_typeshed/__init__.pyi
@@ -248,3 +248,15 @@ class structseq(Generic[_T_co]):
     # but only has any meaning if you supply it a dict where the keys are strings.
     # https://github.com/python/typeshed/pull/6560#discussion_r767149830
     def __new__(cls: type[Self], sequence: Iterable[_T_co], dict: dict[str, Any] = ...) -> Self: ...
+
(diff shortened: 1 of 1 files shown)


Akuli:
I don't see how this would be useful:
- Because of invariance, a `dict[str, str]` is not a valid `JSON`. This makes `JSON` quite useless for argument types.
- We don't have weak unions or `AnyOf`, so `JSON` is not good for return types: you would need to `assert` the structure of the JSON to get anything done.

davidfstr (commented):
It's worth asking the question of what a `JSON` type would be used for.

* If it was documentation _only_ then defining it as either Any or a NewType would be sufficient.
* If it is NOT intended to be used as the return type for the standard library's `json.loads()` (and instead for application code only), than getting a `JSON` value would necessarily involve a `cast(...)` (if defined in the way proposed here, or if it was defined as a NewType).

My 2¢ would be to recommend attempting to define a `JSON` type that would be good enough to use as the return type of `json.loads()`.

JelleZijlstra:
We should now be able to use a properly recursive type.

However, I share the reservations about adding this if we don't know how we'll use it. I think invariance makes it so that we can't really use this alias anywhere in typeshed.

PR #7906, opened 2022-05-21
Simplify and correct many numeric unions

AlexWaygood:
Unblocks https://github.com/PyCQA/flake8-pyi/pull/222


stdlib/filecmp.pyi
```diff
@@ -13,12 +13,12 @@ __all__ = ["clear_cache", "cmp", "dircmp", "cmpfiles", "DEFAULT_IGNORES"]
...
 BUFSIZE: Literal[8192]
 
-def cmp(f1: StrOrBytesPath, f2: StrOrBytesPath, shallow: int | bool = ...) -> bool: ...
+def cmp(f1: StrOrBytesPath, f2: StrOrBytesPath, shallow: bool | Literal[0, 1] = ...) -> bool: ...
 def cmpfiles(
     a: AnyStr | PathLike[AnyStr],
     b: AnyStr | PathLike[AnyStr],
     common: Iterable[AnyStr | PathLike[AnyStr]],
-    shallow: int | bool = ...,
+    shallow: bool | Literal[0, 1] = ...,
```
srittau, line 21: These are cases where we can consider removing the `Literal` at some point. (Otherwise, we'd basically have to replace all instances in argument types with `bool | Literal[0, 1]`.) But no need to do this in this PR.

srittau (commented):
I'm a bit concerned with replacing `int | bool` with `int`. While it's of course technically true, it's a bit of a historical accident that `bool` is a subclass of `int`. (As `bool` was not originally in Python.) I'd rather see it stay in cases where `foo(123)` and `foo(True)` are different use cases. On the other hand, in the long run we should replace `bool | int` with just `bool` in cases, where the argument is used like a boolean (and you started working towards this in this PR by replacing `int` with `Literal[0, 1]`).

For example, in urllib3 I think it would make more sense to annotate the `retries` argument with `Retry | int | Literal[False] | None`, even if `Literal[False]` is technically redundant.

I think this differs from `float | int` -> `int` in that `int` is morally (and per PEP 484) a subtype (but not in implementation) while `bool` is morally not a subtype (while it is in implementation), if that makes sense.

stdlib/filecmp.pyi
AlexWaygood: Yes. I held off doing that because I didn't want to introduce any breaking semantic changes in this PR.

AlexWaygood:
Okay, I've deleted the proposed `int | bool` check from my flake8-pyi PR, and I'll remove the cases where I'm simplifying `int | bool` to just `int` in this PR.

stubs/waitress/waitress/adjustments.pyi
```diff
@@ -8,7 +8,7 @@ from .proxy_headers import PROXY_HEADERS as PROXY_HEADERS
 truthy: frozenset[Any]
 KNOWN_PROXY_HEADERS: frozenset[Any]
 
-def asbool(s: bool | str | int | None) -> bool: ...
+def asbool(s: str | bool | None) -> bool: ...
```
AlexWaygood, line 11: This one is just incorrect: it doesn't take `int`. https://github.com/Pylons/waitress/blob/4b6b5832f3a30d82dc8cb359cf04d2a8edf6712b/src/waitress/adjustments.py#L30-L41

JelleZijlstra: It technically does, since it calls `str()` on its argument and then checks whether the result is `"1"` (among others).

AlexWaygood: I clearly need a coffee; I misread the source code.

AlexWaygood: Changes have been reverted :)

srittau (approved):

AlexWaygood:
@JelleZijlstra, just want to check you're okay with this as well? I know you had some Thoughts when I tried to do something similar with `bytes`/`bytearray` :)

JelleZijlstra (commented):
I'm fine with the change. I'm not really a fan of the int/float special case but it's here to stay.

AlexWaygood:
> I'm fine with the change. I'm not really a fan of the int/float special case but it's here to stay.

Neither am I, but yeah, there's no way we're going to change that at this point :/ and I think it's good to be consistent across typeshed about this principle.

stdlib/filecmp.pyi
hauntsaninja: `bool | int` -> `bool | Literal[0, 1]` is a breaking semantic change. E.g., the following will now no longer type check:
```
def foo(x: int):
    cmp(..., shallow=x)
```
mypy_primer doesn't say anything, so probably nbd, but worth mentioning

stdlib/random.pyi
```diff
@@ -68,7 +68,7 @@ class Random(_random.Random):
     def __init__(self, x: Any = ...) -> None: ...
     # Using other `seed` types is deprecated since 3.9 and removed in 3.11
     if sys.version_info >= (3, 9):
-        def seed(self, a: int | float | str | bytes | bytearray | None = ..., version: int = ...) -> None: ...  # type: ignore[override]
+        def seed(self, a: float | str | bytes | bytearray | None = ..., version: int = ...) -> None: ...  # type: ignore[override]
```
hauntsaninja, line 71: This one feels a little worse to me, any objections to noqa?

AlexWaygood: No, feel free to revert and add a `noqa`!

stdlib/filecmp.pyi
AlexWaygood: > `bool | int` -> `bool | Literal[0, 1]` is a breaking semantic change. 

Correct; I should have chosen my words better. I do however feel that it's useful for a user to get an error if, for whatever reason, they try to pass in `3` for this parameter — that probably indicates a logical error in their code somewhere. So my opinion is that this is probably a worthwhile breaking change, and also less breaking than it would have been if we simply changed it to `bool`.

But anyway, as you say, no primer output, so probably nbd :)

PR #7916, opened 2022-05-21
random: explicitly mention int for seed

hauntsaninja:
https://github.com/python/typeshed/pull/7906#discussion_r878750404
I felt this better documents how seed is used in practice


AlexWaygood (approved):
I'd prefer if we added a comment about why we're `noqa`ing , like we do in `turtle.pyi`, but feel free to merge

AlexWaygood (approved):
Awesome, thanks!

PR #7918, opened 2022-05-21
codecs: add missing function, reflect runtime import *

JelleZijlstra:


JelleZijlstra:
> 
> ```diff
> werkzeug (https://github.com/pallets/werkzeug)
> + src/werkzeug/debug/repr.py:134: error: Unused "type: ignore" comment
> - tests/test_wsgi.py:355: error: Function is missing a return type annotation  [no-untyped-def]
> ```

This found a bug in the stub, a false positive error here: https://github.com/pallets/werkzeug/blob/bb21bf90b0b121e3ed45b9950b823e4b43a81fd8/src/werkzeug/debug/repr.py#L134. I'll fix it.

hauntsaninja (approved):
Looks good to me. codecs is actually something I wouldn't mind having test cases for, IIRC we've regressed some examples before.

PR #7919, opened 2022-05-22
concurrent.futures: new params in 3.11

JelleZijlstra:
https://github.com/python/cpython/commit/fdc0e09c3316098b038996c428e88931f0a4fcdb

### DIFF
--- a/stdlib/concurrent/futures/process.pyi
+++ b/stdlib/concurrent/futures/process.pyi
@@ -53,7 +53,12 @@ class _ResultItem:
     work_id: int
     exception: Exception
     result: Any
-    def __init__(self, work_id: int, exception: Exception | None = ..., result: Any | None = ...) -> None: ...
+    if sys.version_info >= (3, 11):
+        def __init__(
+            self, work_id: int, exception: Exception | None = ..., result: Any | None = ..., exit_pid: int | None = ...
+        ) -> None: ...
+    else:
+        def __init__(self, work_id: int, exception: Exception | None = ..., result: Any | None = ...) -> None: ...
 
 class _CallItem:
     work_id: int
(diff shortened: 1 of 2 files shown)


stdlib/concurrent/futures/process.pyi
```diff
@@ -53,7 +53,12 @@ class _ResultItem:
     work_id: int
     exception: Exception
     result: Any
-    def __init__(self, work_id: int, exception: Exception | None = ..., result: Any | None = ...) -> None: ...
+    if sys.version_info >= (3, 11):
+        def __init__(
```
AlexWaygood, line 57:
```suggestion
        exit_pid: int | None
        def __init__(
```

stdlib/concurrent/futures/process.pyi
```diff
@@ -153,7 +178,17 @@ class ProcessPoolExecutor(Executor):
     _executor_manager_thread_wakeup: _ThreadWakeup
     _result_queue: SimpleQueue[Any]
     _work_ids: Queue[Any]
-    if sys.version_info >= (3, 7):
+    if sys.version_info >= (3, 11):
+        def __init__(
```
AlexWaygood, line 182: It's a private attribute, but we seem to be typing other private attributes for this class 🤷‍♂️

```suggestion
        _max_tasks_per_child: int | None
        def __init__(
```

AlexWaygood (commented):
Looks good, just two things I noticed:

stdlib/concurrent/futures/process.pyi
Akuli: I don't think we should add this just because other private attributes are there too. Stubtest isn't good at checking instance attributes, so we won't notice if it's renamed or removed for whatever reason.

AlexWaygood: Yeah fair, feel free to ignore. I have no strong opinion on this.

JelleZijlstra: Yeah, we should only add private attributes when someone actually needs them. There's quite a lot of private stuff in this file; not sure we really need it all.

hauntsaninja (approved):

AlexWaygood (approved):

PR #7920, opened 2022-05-22
inspect: cannot fix this one

JelleZijlstra:
https://github.com/python/cpython/blob/748c83d96da735f8210167d676469860d4e11b80/Lib/inspect.py#L2631


Akuli (approved):
It exists only on the instances (i.e. enum members), not on the class itself:
```
>>> 'description' in dir(inspect._ParameterKind)
False
>>> 'description' in vars(inspect._ParameterKind)
False
>>> inspect._ParameterKind.POSITIONAL_OR_KEYWORD.description
'positional or keyword'
```

PR #7923, opened 2022-05-22
mmap: add MADV_FREE on darwin

hauntsaninja:


AlexWaygood (approved):
I don't have a mac to test on, but I'm happy if stubtest is happy

JelleZijlstra (approved):
It indeed exists on my Mac, the value is 5

PR #7924, opened 2022-05-22
termios: update for py311

hauntsaninja:


stdlib/termios.pyi
```diff
@@ -244,5 +244,8 @@ if sys.platform != "win32":
     def tcdrain(__fd: FileDescriptorLike) -> None: ...
     def tcflush(__fd: FileDescriptorLike, __queue: int) -> None: ...
     def tcflow(__fd: FileDescriptorLike, __action: int) -> None: ...
+    if sys.version_info >= (3, 11):
+        def tcgetwinsize(__fd: FileDescriptorLike) -> tuple[int, int]: ...
+        def tcsetwinsize(__fd: FileDescriptorLike, __winsize: tuple[int, int]) -> None: ...
```
JelleZijlstra, line 249: It actually works with any two-item PySequence (which isn't the same as Sequence), but we have no way to express that. I'm OK with leaving it as only tuple.

JelleZijlstra (approved):
Added in https://github.com/python/cpython/commit/ae224bb566301d3602e9b090e37c1dcf5a48c914

PR #7927, opened 2022-05-23
bdb: update for py311

hauntsaninja:
https://github.com/python/cpython/commit/ad442a674ca443feec43a88a2d3671784712e550


hauntsaninja:
It's undocumented, not sure whether intentionally or not

srittau (approved):

PR #7932, opened 2022-05-23
requests: allow immutable headers

Akuli:
Fixes a problem reported by @rpdelaney in  https://github.com/python/typeshed/pull/7773#issuecomment-1134888216

I spent a while looking at requests source code, and I'm quite convinced that the headers can't be mutated. Each request is prepared before auth etc headers are applied, and the headers are copied when preparing.

### DIFF
--- a/stubs/requests/requests/sessions.pyi
+++ b/stubs/requests/requests/sessions.pyi
@@ -66,16 +66,8 @@ _Params: TypeAlias = Union[
     str | bytes,
 ]
 _TextMapping: TypeAlias = MutableMapping[str, str]
-_HeadersMapping: TypeAlias = MutableMapping[str, str] | MutableMapping[str, bytes] | MutableMapping[str, str | bytes]
-_HeadersUpdateMapping: TypeAlias = (
-    MutableMapping[str, str]
-    | MutableMapping[str, bytes]
-    | MutableMapping[str, None]
-    | MutableMapping[str, str | bytes]
-    | MutableMapping[str, str | None]
-    | MutableMapping[str, bytes | None]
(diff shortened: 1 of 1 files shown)


JelleZijlstra:
Thanks! Is it worth adding some test cases here? We've gone through quite a few iterations with these requests mappings.

Akuli:
1. Actually it is open-source.v

AlexWaygood:
> Most of the iterations were before `mypy_primer` worked for third-party stubs. Ideally we would just add @rpdelaney's project to mypy_primer, if it happens to be open-source.

+1 to adding more projects to mypy-primer if they use code that isn't otherwise covered. But I feel like "adding more projects to mypy-primer" and "adding specific regression tests" aren't mutually exclusive.

The huge value of mypy-primer is that it can detect problems that we don't even know are problems yet. But the disadvantage is we'd instantly lose coverage if a certain project decided to move from using `requests` to `HTTPie` (or whatever). We have more control over the regression tests in the `test_cases` directory.

AlexWaygood:
> `downforeveryone` is Apache 2.0 licensed. I don't know how `mypy_primer` works, but can't you check out a package at a particular commit? Or just copy-paste in the code you need to check the regression.

I think we probably all agree that adding `downforeverone` to mypy-primer would be both Possible And Good — @Akuli's filed https://github.com/hauntsaninja/mypy_primer/issues/33 to track that :)

What we're debating is whether to also add regression tests to typeshed itself via our fairly new `test_cases` directory. We currently have a smattering of test cases for the stdlib stubs, but no test cases as yet for the third-party stubs typeshed maintains.

Issue #7928, opened 2022-05-23
TypeVars in some generic functions can go unsolved, resulting in an `Any` in the evaluated type


erictraut:
Pyright users have recently reported a couple of issues where they received an `Unknown` (i.e. an implicit `Any`) type when evaluating a call expression. These problems were both related to a case where a TypeVar was used to annotate a parameter with a default argument, but the caller didn't provide an argument for that parameter and relied on the default. This left the TypeVar unsolved.

Here's [an example](https://github.com/microsoft/pyright/issues/3494) from the pyright issue tracker.

@JelleZijlstra suggested that pyright add a check for this condition. I've enhanced the `reportInvalidTypeVarUse` check to detect and report this case. See [this issue](https://github.com/microsoft/pyright/issues/3495) for details.

The new check identifies 10 cases in the stdlib typeshed stubs and 27 cases in third-party stubs. Jelle asked that I create this issue with the full list, which I have pasted below.

```
<redacted>/typeshed/stdlib/argparse.pyi
  <redacted>/typeshed/stdlib/argparse.pyi:215:32 - error: Type variable "_ArgumentParserT" may go unsolved if caller supplies no argument for parameter "parser_class"
    Provide an overload that specifies the return type when the argument is not supplied (reportInvalidTypeVarUse)
<redacted>/typeshed/stdlib/fileinput.pyi
  <redacted>/typeshed/stdlib/fileinput.pyi:32:54 - error: Type variable "AnyStr" may go unsolved if caller supplies no argument for parameter "openhook"
    Provide an overload that specifies the return type when the argument is not supplied (reportInvalidTypeVarUse)
<redacted>/typeshed/stdlib/gettext.pyi
  <redacted>/typeshed/stdlib/gettext.pyi:157:22 - error: Type variable "_T" may go unsolved if caller supplies no argument for parameter "class_"
    Provide an overload that specifies the return type when the argument is not supplied (reportInvalidTypeVarUse)
<redacted>/typeshed/stdlib/optparse.pyi
  <redacted>/typeshed/stdlib/optparse.pyi:243:41 - error: Type variable "AnyStr" may go unsolved if caller supplies no argument for parameter "args"
    Provide an overload that specifies the return type when the argument is not supplied (reportInvalidTypeVarUse)
<redacted>/typeshed/stdlib/tempfile.pyi
[...]

PR #7934, opened 2022-05-24
fileinput: Fix TypeVar usage

### LINKED ISSUE #7922: Wrong stub for `fileinput.input()`
@JelleZijlstra:
> the stubs appear to be wrong here:
https://github.com/python/typeshed/blob/6e18441a2c1e5836eca4e6763a9590d2b9270378/stdlib/fileinput.pyi#L32
Only the `openhook` argument constrains the TypeVar, but it has a default, and if the argument is not given there's nothing for the type checker to use to decide on the value of the TypeVar. The typeshed stub should be changed to add a separate overload where `openhook` is not given.

https://github.com/microsoft/pyright/issues/3494#issuecomment-1133985204

JelleZijlstra:
Fixes #7922, part of #7928.

Also discovered an apparent CPython bug in the process: python/cpython#93157

### DIFF
--- a/stdlib/fileinput.pyi
+++ b/stdlib/fileinput.pyi
@@ -2,8 +2,8 @@ import sys
 from _typeshed import Self, StrOrBytesPath
 from collections.abc import Callable, Iterable, Iterator
 from types import TracebackType
-from typing import IO, Any, AnyStr, Generic, Literal, overload
-from typing_extensions import Protocol
+from typing import IO, Any, AnyStr, Generic, overload, TypeVar
+from typing_extensions import Literal, Protocol
 
 __all__ = [
     "input",
@@ -20,8 +20,10 @@ __all__ = [
     "hook_encoded",
 ]
 
-class _HasReadlineAndFileno(Protocol[AnyStr]):
-    def readline(self) -> AnyStr: ...
+_AnyStr_co = TypeVar("_AnyStr_co", str, bytes)
+
+class _HasReadlineAndFileno(Protocol[_AnyStr_co]):
+    def readline(self) -> _AnyStr_co: ...
     def fileno(self) -> int: ...
 
 if sys.version_info >= (3, 9):
--- a/stdlib/fileinput.pyi
+++ b/stdlib/fileinput.pyi
@@ -3,7 +3,7 @@ from _typeshed import Self, StrOrBytesPath
 from collections.abc import Callable, Iterable, Iterator
 from types import TracebackType
 from typing import IO, Any, AnyStr, Generic, Protocol, TypeVar, overload
-from typing_extensions import Literal
+from typing_extensions import Literal, TypeAlias
 
 if sys.version_info >= (3, 9):
     from types import GenericAlias
@@ -24,9 +24,9 @@ __all__ = [
 ]
 
 if sys.version_info >= (3, 11):
-    _TextMode = Literal["r"]
+    _TextMode: TypeAlias = Literal["r"]
 else:
-    _TextMode = Literal["r", "rU", "U"]
+    _TextMode: TypeAlias = Literal["r", "rU", "U"]
 
 _AnyStr_co = TypeVar("_AnyStr_co", str, bytes, covariant=True)
 
--- a/stdlib/fileinput.pyi
+++ b/stdlib/fileinput.pyi
@@ -20,7 +20,7 @@ __all__ = [
     "hook_encoded",
 ]
 
-_AnyStr_co = TypeVar("_AnyStr_co", str, bytes)
+_AnyStr_co = TypeVar("_AnyStr_co", str, bytes, covariant=True)
 
 class _HasReadlineAndFileno(Protocol[_AnyStr_co]):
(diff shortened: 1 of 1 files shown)


JelleZijlstra:
Good catch! U support was removed in 3.11 (commit 19ba2122ac7313ac29207360cfa864a275b9489e) and I didn't check the earlier branches carefully enough. I'll add it.

stdlib/fileinput.pyi
```diff
@@ -2,7 +2,11 @@ import sys
 from _typeshed import Self, StrOrBytesPath
 from collections.abc import Callable, Iterable, Iterator
 from types import TracebackType
-from typing import IO, Any, AnyStr, Generic
+from typing import IO, Any, AnyStr, Generic, Protocol, TypeVar, overload
+from typing_extensions import Literal
```
AlexWaygood, line 6:
```suggestion
from typing_extensions import Literal, TypeAlias
```

stdlib/fileinput.pyi
```diff
@@ -19,40 +23,133 @@ __all__ = [
     "hook_encoded",
 ]
 
-if sys.version_info >= (3, 9):
-    from types import GenericAlias
+if sys.version_info >= (3, 11):
+    _TextMode = Literal["r"]
```
AlexWaygood, line 27:
```suggestion
    _TextMode: TypeAlias = Literal["r"]
```

stdlib/fileinput.pyi
```diff
@@ -19,40 +23,133 @@ __all__ = [
...
+else:
+    _TextMode = Literal["r", "rU", "U"]
```
AlexWaygood, line 29:
```suggestion
    _TextMode: TypeAlias = Literal["r", "rU", "U"]
```

stdlib/fileinput.pyi
```diff
@@ -19,40 +23,133 @@ __all__ = [
...
 ]
 
-if sys.version_info >= (3, 9):
-    from types import GenericAlias
+if sys.version_info >= (3, 11):
+    _TextMode: TypeAlias = Literal["r"]
+else:
+    _TextMode: TypeAlias = Literal["r", "rU", "U"]
+
+_AnyStr_co = TypeVar("_AnyStr_co", str, bytes, covariant=True)
```
AlexWaygood, line 31: We define the same `TypeVar` (with the same name) in `builtins.pyi` and `os/__init__.pyi` — is it worth adding it to `_typeshed`, do you think?

stdlib/fileinput.pyi
JelleZijlstra: Thanks, I keep forgetting! You do know you can also edit the file directly in https://github.com/JelleZijlstra/typeshed/edit/fileinput/stdlib/fileinput.pyi?pr=%2Fpython%2Ftypeshed%2Fpull%2F7934, right?

AlexWaygood: I do know, yes :) but thanks!

stdlib/fileinput.pyi
JelleZijlstra: It's simple enough that I don't mind redefining it, but I don't feel strongly.

AlexWaygood: We can leave it for now

AlexWaygood (approved):

stdlib/fileinput.pyi
```diff
@@ -65,36 +162,131 @@ def isstdin() -> bool: ...
...
             inplace: bool = ...,
             backup: str = ...,
             bufsize: int = ...,
-            mode: str = ...,
-            openhook: Callable[[StrOrBytesPath, str], IO[AnyStr]] = ...,
+            mode: _TextMode = ...,
+            openhook: Callable[[StrOrBytesPath, str], _HasReadlineAndFileno[str]] | None = ...,
+        ) -> None: ...
+        # Because mode isn't keyword-only here yet, we need two overloads each for
+        # the bytes case and the fallback case.
```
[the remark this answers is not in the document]
JelleZijlstra: Sebastian opened a conversation a while ago about allowing arguments without defaults to follow arguments with defaults, which would make this easier, but it didn't go anywhere.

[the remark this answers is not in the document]
srittau: There was some pushback, because that's mostly useful for typing and considered bad API design otherwise, but honestly more from typing-critical developers, if I remember correctly.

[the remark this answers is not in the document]
AlexWaygood: > An alternative could be to add a special marker like `typing.NO_DEFAULT`, which we'd use as the default.

Oh, I like that quite a lot, actually. It would mean we wouldn't have to change Python's grammar at all. It does also feel like the kind of thing that would disproportionately benefit typeshed, though.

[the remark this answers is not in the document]
Akuli: It should be possible to write a script that automatically applies `NO_DEFAULT` in typeshed wherever possible. I would be interested to see the results,

PR #7935, opened 2022-05-24
gettext: fix unconstrained TypeVar

JelleZijlstra:
Part of #7928

### DIFF
--- a/stdlib/gettext.pyi
+++ b/stdlib/gettext.pyi
@@ -1,7 +1,8 @@
+import io
 import sys
 from _typeshed import StrPath
 from collections.abc import Container, Iterable, Sequence
-from typing import Any, Protocol, TypeVar, overload
+from typing import Any, Callable, Protocol, TypeVar, overload
 from typing_extensions import Final, Literal
 
 if sys.version_info >= (3, 11):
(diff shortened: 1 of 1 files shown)


stdlib/gettext.pyi
```diff
@@ -113,33 +114,91 @@ def find(domain: str, localedir: StrPath | None = ..., languages: Iterable[str]
 _T = TypeVar("_T")
```
AlexWaygood, line 114: The docs for `gettext.translations` say:

> Return a *Translations instance based on the domain, localedir, and languages, which are first passed to [find()](https://docs.python.org/3/library/gettext.html#gettext.find) to get a list of the associated .mo file paths. Instances with identical .mo file names are cached. The actual class instantiated is class_ if provided, otherwise [GNUTranslations](https://docs.python.org/3/library/gettext.html#gettext.GNUTranslations). The class’s constructor must take a single [file object](https://docs.python.org/3/glossary.html#term-file-object) argument. If provided, codeset will change the charset used to encode translated strings in the [lgettext()](https://docs.python.org/3/library/gettext.html#gettext.NullTranslations.lgettext) and [lngettext()](https://docs.python.org/3/library/gettext.html#gettext.NullTranslations.lngettext) methods.

For me, that implies that the expectation is that the value passed to the `class_` parameter should be a subclass of `NullTranslations` (`gettext.translations` won't return a `*Translations` instance in all cases otherwise).

```suggestion
_T = TypeVar("_T", bound=NullTranslations)
```

stdlib/gettext.pyi
```diff
@@ -113,33 +114,91 @@ def find(domain: str, localedir: StrPath | None = ..., languages: Iterable[str]
...
+        localedir: StrPath | None = ...,
+        languages: Iterable[str] | None = ...,
+        class_: None = ...,
+        *,
+        fallback: Literal[True],
+    ) -> GNUTranslations | NullTranslations: ...
+    @overload
+    def translation(
+        domain: str, localedir: StrPath | None, languages: Iterable[str] | None, class_: None, fallback: Literal[True]
+    ) -> GNUTranslations | NullTranslations: ...
```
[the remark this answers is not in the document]
JelleZijlstra: Good point, made it all return NullTranslations

stdlib/gettext.pyi
JelleZijlstra: Thanks, done. The implementation also calls the `add_fallback` method on the returned object.

AlexWaygood (approved):

AlexWaygood:
It might also be good to rename the `_T` TypeVar to `_NullTranslationsT` now that it's bound to `_NullTrabslations`

JelleZijlstra:
Good point, you're right.

AlexWaygood (approved):

PR #7938, opened 2022-05-24
`multiprocessing.managers`: fix `TypeVar` usage

AlexWaygood:
#7928

`dict()` and `list()` just return empty dictionaries and lists (respectively) if no arguments are supplied:

```python
>>> from multiprocessing.managers import SyncManager
>>> with SyncManager() as s:
...     print(s.dict())
...
{}
```


stdlib/multiprocessing/managers.pyi
```diff
@@ -124,8 +124,14 @@ class SyncManager(BaseManager):
     def Semaphore(self, value: Any = ...) -> threading.Semaphore: ...
     def Array(self, typecode: Any, sequence: Sequence[_T]) -> Sequence[_T]: ...
     def Value(self, typecode: Any, value: _T) -> ValueProxy[_T]: ...
-    def dict(self, sequence: Mapping[_KT, _VT] = ...) -> _dict[_KT, _VT]: ...
-    def list(self, sequence: Sequence[_T] = ...) -> _list[_T]: ...
+    @overload
+    def dict(self, sequence: Mapping[_KT, _VT]) -> _dict[_KT, _VT]: ...
+    @overload
+    def dict(self) -> _dict[Any, Any]: ...
```
JelleZijlstra, line 130: This is another instance of the issue @erictraut found in https://github.com/microsoft/pyright/issues/3501#issuecomment-1135455934, though here it's on a function, not a constructor.

AlexWaygood: Yeah, empty collections are tough :(

JelleZijlstra (commented):
Don't have a strong opinion on what to do here, but observations:

- As Eric pointed out, with changes like this, we potentially make it so users silently get `Any` types, instead of getting warned by pyright's strict mode about "unknown" types.
- The parameter types here are wrong. multiprocessing is a maze but as far as I can tell these should accept the same arguments as the `list` and `dict` constructors. For example, `dict()` allows kwargs and `list()` allows any iterable.
- The return types are wrong: these methods return special proxy objects, not actual dicts or lists.

Some examples from playing with them:
```
>>> m.list({1, 2, 3})
<ListProxy object, typeid 'list' at 0xffff933192b0>
>>> l = _
>>> type(l)
<class 'multiprocessing.managers.ListProxy'>
>>> len(l)
3
>>> str(l)
'[1, 2, 3]'
>>> l[0]
1
>>> m.dict(x=3)
<DictProxy object, typeid 'dict' at 0xffff933cd040>
>>> d=_
>>> d["x"]
3
>>> d["x"] = 5
>>> d["x"]
5
```

AlexWaygood:
> 
> ```diff
> optuna (https://github.com/optuna/optuna)
> + tests/samplers_tests/tpe_tests/test_sampler.py:1044: error: Incompatible types in assignment (expression has type "DictProxy[Any, Any]", variable has type "Dict[int, List[int]]")
> + tests/samplers_tests/tpe_tests/test_sampler.py:1045: error: Incompatible types in assignment (expression has type "DictProxy[Any, Any]", variable has type "Dict[int, int]")
> ```

The code in question was doing this:

```python
sequence_dict: Dict[int, List[int]] = manager.dict()
hash_dict: Dict[int, int] = manager.dict()
```

https://github.com/optuna/optuna/blob/ce2b95111cd4ce094fd81c044128fa8851c786b3/tests/samplers_tests/tpe_tests/test_sampler.py#L1044-L1045

This was the best way of annotating that code prior to this PR, given that in typeshed we didn't acknowledge the existence of these classes previously. But it will have to be updated if this PR is merged.

stdlib/multiprocessing/managers.pyi
```diff
@@ -66,6 +65,62 @@ class ValueProxy(BaseProxy, Generic[_T]):
...
+    def pop(self, __key: _KT) -> _VT: ...
+    @overload
+    def pop(self, __key: _KT, __default: _VT | _T) -> _VT | _T: ...
+    def keys(self) -> list[_KT]: ...  # type: ignore[override]
+    def values(self) -> list[tuple[_KT, _VT]]: ...  # type: ignore[override]
+    def items(self) -> list[_VT]: ...  # type: ignore[override]
+    if sys.version_info < (3, 7):
+        def has_key(self, k: _KT) -> bool: ...
+
+class BaseListProxy(BaseProxy, MutableSequence[_T]):
```
JelleZijlstra, line 90: Missing `__contains__` and `reverse`

stdlib/multiprocessing/managers.pyi
```diff
@@ -66,6 +65,62 @@ class ValueProxy(BaseProxy, Generic[_T]):
     if sys.version_info >= (3, 9):
         def __class_getitem__(cls, item: Any) -> GenericAlias: ...
 
+class DictProxy(BaseProxy, MutableMapping[_KT, _VT]):
```
JelleZijlstra, line 68: Also `__contains__`, `clear`, `popitem`, `setdefault`, `update`.

(Or are those inherited? In that case, why do we have `__len__`?)

stdlib/multiprocessing/managers.pyi
```diff
@@ -66,6 +65,62 @@ class ValueProxy(BaseProxy, Generic[_T]):
...
+    def copy(self) -> dict[_KT, _VT]: ...
+    @overload
+    def get(self, __key: _KT) -> _VT | None: ...
+    @overload
+    def get(self, __key: _KT, __default: _VT | _T) -> _VT | _T: ...
+    @overload
+    def pop(self, __key: _KT) -> _VT: ...
+    @overload
+    def pop(self, __key: _KT, __default: _VT | _T) -> _VT | _T: ...
+    def keys(self) -> list[_KT]: ...  # type: ignore[override]
```
JelleZijlstra, line 84: This is because of https://github.com/python/cpython/blob/518595652759462b13904df767f69b8cc2c61143/Lib/multiprocessing/managers.py#L51 (took me a while to figure out)

stdlib/multiprocessing/managers.pyi
```diff
@@ -124,8 +179,21 @@ class SyncManager(BaseManager):
...
+    @overload
+    def dict(self, **kwargs: _VT) -> DictProxy[str, _VT]: ...
+    @overload
+    def dict(self, __map: SupportsKeysAndGetItem[_KT, _VT], **kwargs: _VT) -> DictProxy[_KT, _VT]: ...
+    @overload
+    def dict(self, __iterable: Iterable[tuple[_KT, _VT]], **kwargs: _VT) -> DictProxy[_KT, _VT]: ...
+    @overload
+    def dict(self, __iterable: Iterable[list[str]]) -> DictProxy[str, str]: ...
+    @overload
+    def list(self, sequence: Sequence[_T]) -> ListProxy[_T]: ...
```
JelleZijlstra, line 194:
```suggestion
    def list(self, __sequence: Sequence[_T]) -> ListProxy[_T]: ...
```
It's pos-only.

stdlib/multiprocessing/managers.pyi
AlexWaygood: Both inherited from `MutableSequence`

stdlib/multiprocessing/managers.pyi
AlexWaygood: All inherited from `MutableMapping`. `MutableMapping.__len__` is an abstract method, so must be overridden here in order to make this class concrete.

PR #7939, opened 2022-05-24
`tempfile`: Fix `TypeVar` usage

AlexWaygood:
#7928

I find the name `_DirT` really confusing, considering it's a type alias rather than a `TypeVar`, but decided against fixing that in this PR in order to keep the diff small.


AlexWaygood:
> 
> ```diff
> manticore (https://github.com/trailofbits/manticore)
> + manticore/core/workspace.py:220: error: Argument 1 to "exists" has incompatible type "Optional[str]"; expected "Union[Union[str, bytes, PathLike[str], PathLike[bytes]], int]"
> + manticore/core/workspace.py:221: error: Argument 1 to "isdir" has incompatible type "Optional[str]"; expected "Union[Union[str, bytes, PathLike[str], PathLike[bytes]], int]"
> + manticore/core/workspace.py:223: error: Argument 1 to "mkdir" has incompatible type "Optional[str]"; expected "Union[str, bytes, PathLike[str], PathLike[bytes]]"
> ```

I have no idea why this change makes mypy think that the result of `os.path.abspath(tempfile.mkdtemp(prefix='foo', dir='bar'))` can be `None`.

https://github.com/trailofbits/manticore/blob/9e11bc9ae9cc961d38c2a4932372d89fec2c6b26/manticore/core/workspace.py#L212-L223

hauntsaninja:
Looks like it's one of those cases where it's sensitive to the ordering of the contraints in AnyStr, if we had `AnyStr = TypeVar("AnyStr", bytes, str)` we'd get a different result. cc @JelleZijlstra

JelleZijlstra (commented):
This disallows some valid calls, such as `tempfile.mkstemp(None, b"x")`. I think we can fix this and work around the mypy bug with three overloads:

1. One where all the args are None (you already have it)
2. One where all the args are `str | None`, returning `tuple[int, str]`
3. One where all the args are `bytes | None`, returning `tuple[int, bytes]`

`mkdtemp` would need similar treatment.

`mktemp` indeed doesn't accept bytes, so it was correct to remove the `AnyStr` for it.

hauntsaninja (commented):
nit: I'd remove the "according to mypy" since these overloads genuinely do overlap (just to disambiguate from other comments where we workaround incorrect mypy behaviour).

AlexWaygood:
> nit: I'd remove the "according to mypy" since these overloads genuinely do overlap (just to disambiguate from other comments where we workaround incorrect mypy behaviour).

I think the only reason pyright never complains about this is we have the check switched off in pyrightconfig.json.

stdlib/tempfile.pyi
```diff
@@ -380,20 +380,22 @@ class TemporaryDirectory(Generic[AnyStr]):
     if sys.version_info >= (3, 9):
         def __class_getitem__(cls, item: Any) -> GenericAlias: ...
 
+# The overloads are overlapping according to mypy, but they should still work fine.
```
AlexWaygood, line 383:
```suggestion
# The overloads overlap, but they should still work fine.
```

stdlib/tempfile.pyi
```diff
@@ -380,20 +380,22 @@ class TemporaryDirectory(Generic[AnyStr]):
...
-    suffix: AnyStr | None = ..., prefix: AnyStr | None = ..., dir: _DirT[AnyStr] | None = ..., text: bool = ...
-) -> tuple[int, AnyStr]: ...
-@overload
-def mkdtemp() -> str: ...
-@overload
-def mkdtemp(suffix: AnyStr | None = ..., prefix: AnyStr | None = ..., dir: _DirT[AnyStr] | None = ...) -> AnyStr: ...
+    suffix: bytes | None = ..., prefix: bytes | None = ..., dir: _DirT[bytes] | None = ..., text: bool = ...
+) -> tuple[int, bytes]: ...
+
+# The overloads are overlapping according to mypy, but they should still work fine.
```
AlexWaygood, line 393:
```suggestion
# The overloads overlap, but they should still work fine.
```

AlexWaygood:
> This disallows some valid calls, such as `tempfile.mkstemp(None, b"x")`.

Good spot! I'm fine with the new approach :)

JelleZijlstra (approved):

PR #7942, opened 2022-05-25
redis: Fix unconstrained TypeVars in a few redis commands

JelleZijlstra:
Part of #7928


AlexWaygood (approved):
That's 4 more functions where something like `typing.NO_DEFAULT` would be useful

PR #7944, opened 2022-05-25
builtins: Fix unconstrained overloads in set() and frozenset()

JelleZijlstra:
See https://github.com/microsoft/pyright/issues/3501#issuecomment-1135979479

Related to #7928


stdlib/builtins.pyi
```diff
@@ -909,7 +909,7 @@ class list(MutableSequence[_T], Generic[_T]):
 class dict(MutableMapping[_KT, _VT], Generic[_KT, _VT]):
     # __init__ should be kept roughly in line with `collections.UserDict.__init__`, which has similar semantics
```
hauntsaninja, line 910: Should we make a similar change to `collections.UserDict.__init__`

hauntsaninja (commented):
Looks good!

stdlib/builtins.pyi
JelleZijlstra: Good point, made similar changes to constructors in that file.

hauntsaninja (approved):

stdlib/builtins.pyi
```diff
@@ -962,6 +962,9 @@ class dict(MutableMapping[_KT, _VT], Generic[_KT, _VT]):
         def __ior__(self: Self, __value: Iterable[tuple[_KT, _VT]]) -> Self: ...
 
 class set(MutableSet[_T], Generic[_T]):
+    @overload
+    def __init__(self) -> None: ...
+    @overload
     def __init__(self, __iterable: Iterable[_T] = ...) -> None: ...
```
AlexWaygood, line 968: We don't need the default here anymore if we're using overloads for `__init__`.

```suggestion
    def __init__(self, __iterable: Iterable[_T]) -> None: ...
```

stdlib/builtins.pyi
```diff
@@ -998,6 +1001,9 @@ class set(MutableSet[_T], Generic[_T]):
         def __class_getitem__(cls, __item: Any) -> GenericAlias: ...
 
 class frozenset(AbstractSet[_T_co], Generic[_T_co]):
+    @overload
+    def __init__(self) -> None: ...
+    @overload
     def __init__(self, __iterable: Iterable[_T_co] = ...) -> None: ...
```
AlexWaygood, line 1007:
```suggestion
    def __init__(self, __iterable: Iterable[_T_co]) -> None: ...
```

PR #7945, opened 2022-05-25
mock: Fix unconstrained TypeVar

JelleZijlstra:
The first overload covers the case where `new` is not given.

Part of #7928


AlexWaygood (approved):
A refreshingly easy fix

PR #7948, opened 2022-05-25
Update mypy to 0.960

srittau:


AlexWaygood (approved):

AlexWaygood:
Ah, we should change the `stubtest_stdlib` workflow so that it runs when we change `requirements-tests.txt`. I don't think there are any significant stubtest changes this time, but there obviously have been in the past.

srittau:
Yeah, I just noticed that too. Wondered why the tests ran so quickly.

PR #6583, opened 2021-12-14
Add `SupportsRichComparison` type to `_typeshed`

AlexWaygood:
Fixes #6575

Cc: @JukkaL

PR #7950, opened 2022-05-25
Flesh out more of jsonschema stubs

sirosen:
Apply more detailed annotations to the format module and most of the exceptions module.

---

Some background:

`jsonschema` is starting to add internal annotations and in the last year added `mypy` to its own linting/testing. Hopefully at some point we can publish the package with annotations for general use, but we're not there today.

So there's going to be an ongoing project of adding internal annotations, and I'm going to try to port that stuff back into typeshed where appropriate.
There are also some updates here which aren't part of the `jsonschema` annotations, and I would like to at some point port them back in the other direction -- into `jsonschema`.

If anyone has ideas about how to get annotations gradually added and eventually published as part of a widely used package which is already in typeshed, I'm open to learning a better way of doing this.


sirosen:
I'm not sure what the mypy_primer error means.

I tried running `mypy_primer` myself (for the first time) with slightly different options from the github workflow, and here's what I got:
```
$ mypy_primer --custom-typeshed-repo ./typeshed --output concise
tornado (https://github.com/tornadoweb/tornado)
+ tornado/web.py:1687: error: Function does not return a value

aiohttp (https://github.com/aio-libs/aiohttp)
+ aiohttp/web_app.py:381: error: Function does not return a value  [func-returns-value]
```
Neither of which are, as far as I know, related to this change.

So possibly this schemathesis line is a false negative?

stubs/jsonschema/jsonschema/exceptions.pyi
```diff
@@ -1,40 +1,50 @@
...
+        schema: Any = ...,
+        schema_path: Sequence[str] = ...,
+        parent: _Error | None = ...,
     ) -> None: ...
     @classmethod
-    def create_from(cls, other): ...
+    def create_from(cls: type[Self], other: _Error) -> Self: ...
     @property
-    def absolute_path(self): ...
+    def absolute_path(self) -> Sequence[str]: ...
```
AlexWaygood, line 42: We generally prefer to be as precise as possible with return types in typeshed, so that would imply using `deque[str]` here instead of `Sequence[str]`. If you think it's an implementation detail which precise type is returned here, however, then you may want to ignore me here :)

The same point applies to `absolute_schema_path` below.

```suggestion
    def absolute_path(self) -> deque[str]: ...
```

stubs/jsonschema/jsonschema/exceptions.pyi
```diff
@@ -1,40 +1,50 @@
...
     @property
-    def absolute_schema_path(self): ...
+    def absolute_schema_path(self) -> Sequence[str]: ...
     @property
-    def json_path(self): ...
+    def json_path(self) -> str: ...
+    def _contents(self) -> dict[str, Any]: ...
```
AlexWaygood, line 47: It would be a bit involved, but you _could_ consider returning a `TypedDict` here for more precise typing (each key will have a specific type associated with it for the corresponding value, if I understand the source code correctly.

stubs/jsonschema/jsonschema/exceptions.pyi
```diff
@@ -1,40 +1,50 @@
+from _typeshed import Self
+from collections import deque
+from collections.abc import Callable, Container, Iterable, Sequence
 from typing import Any
 
-WEAK_MATCHES: Any
-STRONG_MATCHES: Any
+from jsonschema import _utils, protocols
+
+_RelevanceFuncType = Callable[[ValidationError], Any]
```
AlexWaygood, line 8: Functions [passed to the `key` argument of `builtins.max`](https://github.com/python-jsonschema/jsonschema/blob/65f591e5152e37867cdd5db1baf6ae175229713e/jsonschema/exceptions.py#L359) must return objects that satisfy the `_typeshed.SupportsRichComparison` interface:

https://github.com/python/typeshed/blob/2d2b34c1ee25f9c63554250d2b5f897a395ae300/stdlib/builtins.pyi#L1295

```suggestion
_RelevanceFuncType = Callable[[ValidationError], SupportsRichComparison]
```

(You'll have to import it from `_typeshed` at the top of the file.)

AlexWaygood:
> I'm not sure what the mypy_primer error means.

`¯\_(ツ)_/¯`

stubs/jsonschema/jsonschema/_format.pyi
```diff
@@ -16,28 +20,28 @@ draft7_format_checker: FormatChecker
 draft201909_format_checker: FormatChecker
 draft202012_format_checker: FormatChecker
 
-def is_email(instance) -> bool: ...
-def is_ipv4(instance) -> bool: ...
-def is_ipv6(instance) -> bool: ...
+def is_email(instance: Any) -> bool: ...
```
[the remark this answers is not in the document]
sirosen: I'll switch this.

Aside/tangent: using `object` in similar functions (e.g. `ensure_is_foo(x: object) -> Foo`) can be less nice than `Any` because of how it responds to type narrowing. After you write enough "assert-is-X" utilities, it becomes a habit to use `Any` in basically any case in which one could use `object`. This is all off-topic, but I wonder if changing the way that `object` narrows would make it more comfortable/normal for typing users like myself to use it. When I first ran into this, I found it _very_ surprising that `isinstance(x, str)` does not narrow `x: object` to `str`.

stubs/jsonschema/jsonschema/exceptions.pyi
sirosen: > If you think it's an implementation detail which precise type is returned here, however, then you may want to ignore me here :)

Yeah, I considered annotating it as `deque` -- that was in my first draft -- but then decided exactly this. Callers should not be relying on that value being a `deque` IMO.
In a way, your comment helps validate my decision. It means I was thinking the right kinds of things!

stubs/jsonschema/jsonschema/_format.pyi
```diff
@@ -16,28 +20,28 @@ draft7_format_checker: FormatChecker
...
+def is_ipv4(instance: Any) -> bool: ...
```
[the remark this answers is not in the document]
AlexWaygood: > If you're okay with this "white lie" in the types, I'd like to keep this as the annotation and then go fix the library later.

I'm okay with that 😄 I guessed that that was probably the case!

stubs/jsonschema/jsonschema/_format.pyi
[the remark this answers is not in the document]
JelleZijlstra: > When I first ran into this, I found it _very_ surprising that `isinstance(x, str)` does not narrow `x: object` to `str`.

That's not true:

https://mypy-play.net/?mypy=latest&python=3.10&gist=5104b5bedec0b6dd05bd8b6cf23fc1ef

stubs/jsonschema/jsonschema/exceptions.pyi
sirosen: Ah, that's why `Ctrl+F: "Comparable"` didn't get me anything on the typing docs!

I wanted this but decided against adding
```python
class _Comparable(Protocol):
    def __lt__(self, other: Any) -> bool: ...
```
to the stubs.

I'll make the change, with great enthusiasm.

(Is `SupportsRichComparison` on a path towards inclusion in `typing-extensions` and `typing`?)

stubs/jsonschema/jsonschema/exceptions.pyi
sirosen: I'm going to punt on this for now, and just add a TODO comment to make the type more precise with `TypedDict`, if that's okay. Right now, I want to get this changeset "over the finish line" and then come back when I have some more time to work on `jsonschema` and on the typing.

stubs/jsonschema/jsonschema/exceptions.pyi
AlexWaygood: > Is `SupportsRichComparison` on a path towards inclusion in `typing-extensions` and `typing`?

I'm rooting for it 🤞 but I think it's still too new to add it to `typing_extensions` just yet — we only added it a few months ago in #6583, so I don't think it can really be considered stable yet

stubs/jsonschema/jsonschema/exceptions.pyi
AlexWaygood: Sure

stubs/jsonschema/jsonschema/_format.pyi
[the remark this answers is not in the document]
sirosen: Oh. Hmm... I know I've run into unexpected issues in which `object` doesn't narrow` but `Any` does. But clearly it wasn't such a simple case. If I see this again, I'll make sure to pay closer attention.

Sorry for the inaccuracy and distraction.

AlexWaygood (approved):
LGTM, but I'll wait a bit to see if any other maintainers have any thoughts before merging

JelleZijlstra:
> LGTM, but I'll wait a bit to see if any other maintainers have any thoughts before merging

I haven't done a full review but nothing stands out to me on skimming the stubs. If you're OK with the change, feel free to merge.

AlexWaygood:
> > LGTM, but I'll wait a bit to see if any other maintainers have any thoughts before merging
> 
> I haven't done a full review but nothing stands out to me on skimming the stubs. If you're OK with the change, feel free to merge.

Merged ;)

hauntsaninja:
> I tried running mypy_primer myself (for the first time) with slightly different options from the github workflow

The output you get from the command is the effect of diffing mypy master vs mypy 0.960 (while using a custom typeshed repo). I'm not sure what the schemathesis diff is about, but we do some output munging / mypy is sensitive to various things that cause it to change file processing order, and sometimes we get weird diffs.

stubs/jsonschema/jsonschema/exceptions.pyi
```diff
@@ -1,40 +1,52 @@
...
+        schema: Any = ...,
+        schema_path: Sequence[str] = ...,
+        parent: _Error | None = ...,
     ) -> None: ...
     @classmethod
-    def create_from(cls, other): ...
+    def create_from(cls: type[Self], other: _Error) -> Self: ...
     @property
-    def absolute_path(self): ...
+    def absolute_path(self) -> Sequence[str]: ...
```
DMRobertson, line 42: This change is incorrect (at least with jsonschema 4.4.0) because path entries may be indexes into a JSON array. For example:

```python
import jsonschema

SCHEMA = {
    "$schema": "https://json-schema.org/draft/2020-12/schema",
    "type": "array",
    "items": {"type": "string"},
}


document = ["a", 12345, "c", "d"]

try:
    jsonschema.validate(document, SCHEMA)
except jsonschema.ValidationError as e:
    print(repr(e))
    for path_entry in e.absolute_path:
        print(repr(path_entry), type(path_entry))
```

```
$ pip show jsonschema | \grep Version
Version: 4.4.0

$ python temp.py
<ValidationError: "12345 is not of type 'string'">
1 <class 'int'>
```

I think `Sequence[str | int]` is the correct annotation? I'll make a PR with that

sirosen: Yes! Apologies for the oversight. These can be ints in precisely the case you mention. I'll double-check for other cases of this. (I'm not a typeshed maintainer, but I'll also go 👍 your PR.)

DMRobertson: Not to worry, glad it was straightforward to sort out!

PR #7957, opened 2022-05-26
posixpath: Rearrange overloads

JelleZijlstra:
I think this may remove the false positives from #7939's mypy-primer output.


JelleZijlstra:
Seems like it doesn't help.

Issue #7956, opened 2022-05-26
pillow image modes are missing


misimisi171718:
some of the image modes are missing from type hints like `pa` and others documented [here](https://pillow.readthedocs.io/en/stable/handbook/concepts.html#concept-modes)

```python
PIL.Image.new("pa",(5,5),(255,255))
```
throes the error
```
Argument 1 to "new" has incompatible type "Literal['PA']"; expected "Literal['1', 'CMYK', 'F', 'HSV', 'I', 'L', 'LAB', 'P', 'RGB', 'RGBA', 'RGBX', 'YCbCr']"
```

AlexWaygood:
Thanks — PR welcome!

PR #7960, opened 2022-05-26
Pillow: Add missing image modes to `_Mode` type alias in `Image.pyi` stub (issue #7956)

Crozzers:
This PR adds all of the image modes documented [here](https://pillow.readthedocs.io/en/stable/handbook/concepts.html#concept-modes) to the `_Mode` type alias in the `PIL.Image` type stub

### DIFF
--- a/stubs/Pillow/PIL/Image.pyi
+++ b/stubs/Pillow/PIL/Image.pyi
@@ -14,7 +14,7 @@ from ._imaging import (
 from .ImageFilter import Filter
 from .ImagePalette import ImagePalette
 
-_Mode: TypeAlias = Literal["1", "CMYK", "F", "HSV", "I", "L", "LAB", "P", "RGB", "RGBA", "RGBX", "YCbCr"]
+_Mode: TypeAlias = Literal["1", "BGR;15", "BGR;16", "BGR;24", "BGR;32", "CMYK", "F", "HSV", "I", "I;16", "I;16B", "I;16L", "I;16N", "L", "LA", "LAB", "La", "P", "PA", "RGB", "RGBA", "RGBX", "RGBa", "YCbCr"]
 _Resample: TypeAlias = Literal[0, 1, 2, 3, 4, 5]
 _Size: TypeAlias = tuple[int, int]
 _Box: TypeAlias = tuple[int, int, int, int]
--- a/stubs/Pillow/PIL/Image.pyi
+++ b/stubs/Pillow/PIL/Image.pyi
@@ -14,7 +14,32 @@ from ._imaging import (
 from .ImageFilter import Filter
 from .ImagePalette import ImagePalette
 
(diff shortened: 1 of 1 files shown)


AlexWaygood (approved):
Thanks!

Julian-O:
Oh dear. I had set aside some time today to research this issue, but when I went to look at the code, I found I was too late!

Just in case anyone visits this again, so my effort wasn't wasted: 

The valid values of mode in Pillow are lightly documented and somewhat inconsistent. This table shows the current state of affairs:

| Mode  | [Documented](https://pillow.readthedocs.io/en/stable/handbook/concepts.html) | Accepted by `Image.new()` | Recognised by [`ImageMode.getmode()`](https://github.com/python-pillow/Pillow/blob/main/src/PIL/ImageMode.py)` | Accepted in `Image.pyi`before this PR | Accepted after
|  --  |  --  |  --  |  --  |  --  |  --  |
| 1 | Standard|True|Core|True|True
| L |Standard|True|Core|True|True
| P |Standard|True|Core|True|True
| RGB |Standard|True|Core|True|True
| RGBA |Standard|True|Core|True|True
| CMYK |Standard|True|Core|True|True
| YCbCr  |Standard|True|Core|True|True
| LAB |Standard|True|Core|True|True
| HSV |Standard|True|Core|True|True
| I |Standard|True|Core|True|True
| F |Standard|True|Core|True|True
| LA |Limited Support|True|Experimental|False|True
| PA |Limited Support|True|Experimental|False|True
| RGBX |Limited Support|True|Core|True|True
| RGBa |Limited Support|True|Experimental|False|True
| La |Limited Support|True|Experimental|False|True
| I;16 |Limited Support|True|Mapping|False|True
| I;16L |Limited Support|True|Mapping|False|True
| I;16B |Limited Support|True|Mapping|False|True
| I;16N |Limited Support|True|Mapping|False|True
| BGR;15 |Limited Support|False|Experimental|False|True
| BGR;16 |Limited Support|False|Experimental|False|True
| BGR;24 |Limited Support|False|Experimental|False|True
| BGR;32 |Limited Support|False|Experimental|False|True
| I;16S |Undocumented|False|Mapping|False|False
| I;16LS |Undocumented|False|Mapping|False|False
| I;16BS |Undocumented|False|Mapping|False|False
| I;16NS |Undocumented|False|Mapping|False|False
| I;32 |Undocumented|False|Mapping|False|False
| I;32B |Undocumented|False|Mapping|False|False
| I;32L |Undocumented|False|Mapping|False|False
| I;32S |Undocumented|False|Mapping|False|False
| I;32BS |Undocumented|False|Mapping|False|False
| I;32LS |Undocumented|False|Mapping|False|False

AlexWaygood:
> This table shows the current state of affairs

Wow, that's excellent, thank you! It looks like we may now be erring on the side of being slightly too permissive, in that case. However, this aligns with typeshed's general policy of preferring false-negative errors over false-positives. So I think the new state of affairs is fine.

Thanks again for your research here, I'm sure it will be helpful for others who come across this issue/PR :)

srittau:
Considering the state of affairs, I wonder whether it may be better to just change the mode types to `str`.

AlexWaygood:
> Considering the state of affairs, I wonder whether it may be better to just change the mode types to `str`.

I wouldn't object to that!

PR #7962, opened 2022-05-27
markdown: remove `__getattr__()`.

leven-cn:


hauntsaninja (changes requested):
We shouldn't have getattr at all. It's just used by markdown.util to issue deprecation warnings. Including it in the stub will result in type checker false negatives.

hauntsaninja (approved):

PR #7963, opened 2022-05-27
contextlib: Remove explicit base class from ExitStack

### LINKED ISSUE #7961: Wrong return type from ExitStack.enter_context when given a child of ExitStack
**Bug Report**

The return type from `enter_context()` is incorrect when the argument is a child of `ExitStack`.

**To Reproduce**

```python
# ctx.py

from contextlib import ExitStack

class Thing(ExitStack):
    def foo(self) -> None:
        print("foo")

stack = ExitStack()
thing = stack.enter_context(Thing())
reveal_type(thing)
thing.foo()
```

**Expected Behavior**

The type of `thing` should be `Thing`, not `ExitStack`.

**Actual Behavior**

```bash
$ mypy ctx.py 
ctx.py:11: note: Revealed type is "contextlib.ExitStack"
ctx.py:12: error: "ExitStack" has no attribute "foo"
```

**Your Environment**

```bash
$ mypy --version 
mypy 0.960 (compiled: yes)
```

JelleZijlstra:
Fixes #7961


AlexWaygood (approved):
Thank you! This looks great to me.

Issue #7497, opened 2022-03-16
Replace # type: ignores without error codes


srittau:
pyright recently added `# pyright: ignore[...]`. We should replace all plain `# type: ignore`s with either `# type: ignore[...]` or `# pyright: ignore[...]` if possible. There might be a few ignores we can't replace.

```
$ rgrep "# type: ignore" stubs/ stdlib/ | grep -v "# type: ignore\[" | grep -v @python2 | wc -l
101
```

srittau:
Cf #7495

AlexWaygood:
If we do this for all the `type: ignore`s that are irrelevant to mypy, we can also consider trying out the new `ignore-without-code` mypy option, which will complain if we have a `type: ignore` comment without an error code.

JelleZijlstra:
Some ignores may also be necessary for pytype (and maybe pyre, which we don't run in CI).

rchen152:
Pytype discards `# type: ignore`s without doing anything with them, so removing them shouldn't affect it :)

Akuli:
Let's say someone makes a new type checker that doesn't attempt to parse `# pyright: ignore[...]` comments, but understands the standard `# type: ignore` comments. Should it display an error if it notices that there's something wrong with the stubs? Or just silence all errors coming from stub files?

I generally don't like introducing type checker specific things into stubs. While `# type: ignore[error-code]` is mypy-specific, other type checkers tend to treat it the same as `# type: ignore`. But how is a type checker supposed to know what to do with `# pyright: ignore[...]`?

I find it unlikely that we need to say "ignore this error in pyright but not in any other type checker". An error like that would be basically a bug in pyright, and those get fixed very quickly.

srittau:
The ignores serve our internal tests. Type checkers shouldn't warn about perceived problems in typeshed or external stubs. (I also think it's also problematic for a type checker to treat "malformed" ignores as a regular ignore.) Everything else would mean we'd potentially have to annotate every line in typeshed with a ignore, because who knows what a type checkers decides to warn about.

Using plain `# type: ignore` in stubs is dangerous as it hides potential problems. It's also a documentation issue as it's often not immediately clear why a certain line is ignored.

srittau:
One further advantage of using error codes is that it gives us the chance to collect statistics about the types of `# type: ignore`s we use. This could influence the options we use in tests and point to areas of improvement in type checkers.

AlexWaygood:
[...]

PR #7964, opened 2022-05-27
psycopg2: stub improvements

JelleZijlstra:
Fixes an entry from #7928 along with a number of other improvements.

I went off the C code:
https://github.com/psycopg/psycopg2/blob/master/psycopg/connection_type.c


stubs/psycopg2/psycopg2/_psycopg.pyi
```diff
@@ -371,64 +371,92 @@ class connection:
...
     def info(self) -> ConnectionInfo: ...
     isolation_level: Any
     notices: Any
     notifies: Any
     @property
-    def pgconn_ptr(self) -> Any: ...
+    def pgconn_ptr(self) -> int | None: ...
     @property
     def protocol_version(self) -> int: ...
+    # For these two the getter returns bool | None, but the setter
```
AlexWaygood, line 394: Could also consider using read-write properties that return different types from their getter to the ones they receive in their setter, like we do here:

https://github.com/python/typeshed/blob/62a8a6922cb052704198ebd28d26887d3b6d17f4/stdlib/builtins.pyi#L82

JelleZijlstra: Oh good point, I thought mypy didn't support that.

stubs/psycopg2/psycopg2/_psycopg.pyi
```diff
@@ -371,64 +371,99 @@ class connection:
...
     @property
     def info(self) -> ConnectionInfo: ...
-    isolation_level: Any
-    notices: Any
-    notifies: Any
     @property
-    def pgconn_ptr(self) -> Any: ...
+    def isolation_level(self) -> int | None: ...
+    @isolation_level.setter
+    def isolation_level(self, __value: str | bytes | int | None) -> None: ...  # pyright: ignore
```
AlexWaygood, line 390: I'm fine with using `# pyright: ignore` here (rather than a bare `# type: ignore`, which would be the other possibility), but I think some other maintainers had some concerns when we last discussed it in #7497. @srittau suggested in https://github.com/python/typeshed/issues/7497#issuecomment-1070817560 that we should maybe just disable this specific pyright check in our `pyrightconfig.json`, since it's the only type checker in our CI that complains about it, and it basically just surfaces issues in the implementation rather than the stubs.

JelleZijlstra: Good point, #7985

AlexWaygood (commented):
I don't particularly trust my ability to cross-check this against the C code, but this all looks fine to me, and I trust you to have read the C code thoroughly!

PR #7967, opened 2022-05-27
Pillow: change `_Mode` type alias to `str` in `Image.pyi` stub

Crozzers:
As mentioned by @Julian-O in [their comment](https://github.com/python/typeshed/pull/7960#issuecomment-1139252724) on the previous PR, support for various image types is inconsistent.

This PR changes the `_Mode` type alias to simply be `str`


srittau (approved):
Thanks!

PR #7969, opened 2022-05-27
openpyxl: Updates openpyxl.utils.cell stubs

XF-FW:
Updates this stub file, whose types were not filled in.


stubs/openpyxl/openpyxl/utils/cell.pyi
```diff
@@ -9,17 +9,17 @@ ABSOLUTE_RE: Any
...
-def range_boundaries(range_string): ...
+def get_column_letter(idx: int) -> str: ...
+def column_index_from_string(str_col: str) -> int: ...
+def range_boundaries(range_string: str) -> tuple[int, int, int, int]: ...
 def rows_from_range(range_string) -> Generator[Any, None, None]: ...
 def cols_from_range(range_string) -> Generator[Any, None, None]: ...
-def coordinate_to_tuple(coordinate): ...
-def range_to_tuple(range_string): ...
-def quote_sheetname(sheetname): ...
+def coordinate_to_tuple(coordinate: str) -> tuple[int, str]: ...
```
AlexWaygood, line 23: The second element in the returned tuple is a value in the `_COL_STRING_CACHE` dictionary, which has `str`s for keys and `int`s for values.

https://foss.heptapod.net/openpyxl/openpyxl/-/blob/branch/3.0/openpyxl/utils/cell.py#L202

```suggestion
def coordinate_to_tuple(coordinate: str) -> tuple[int, int]: ...
```

AlexWaygood (commented):
Thanks! Just one thing I noticed, otherwise this looks good.

stubs/openpyxl/openpyxl/utils/cell.pyi
XF-FW: Ah nice catch. Thank you!

I've applied your suggestion.

AlexWaygood (approved):
Looks great!

PR #7970, opened 2022-05-27
Introduce `_typeshed.GenericPath` and `_typeshed.AnyStr_co`

AlexWaygood:


stdlib/_typeshed/__init__.pyi
```diff
@@ -115,6 +118,7 @@ class SupportsItemAccess(SupportsGetItem[_KT_contra, _VT], Protocol[_KT_contra,
 # These aliases are simple strings in Python 2.
```
srittau, line 118: We should probably remove this comment, as it will be outdated soon and only applies to some of the aliases now.

stdlib/_typeshed/__init__.pyi
```diff
@@ -115,6 +118,7 @@ class SupportsItemAccess(SupportsGetItem[_KT_contra, _VT], Protocol[_KT_contra,
...
 StrPath: TypeAlias = str | PathLike[str]  # stable
 BytesPath: TypeAlias = bytes | PathLike[bytes]  # stable
+GenericPath: TypeAlias = AnyStr | PathLike[AnyStr]
```
srittau, line 121: Good name!

stdlib/_typeshed/__init__.pyi
AlexWaygood:
```suggestion
```

srittau (approved):

PR #7971, opened 2022-05-27
Add a `getattr` overload for `BaseException.__notes__`

AlexWaygood:
This change means that type checkers can easily infer that the type of `getattr(base_exception_obj, "__notes__", [])` will be `list[str]`, without requiring an explicit annotation.

As suggested by @Akuli in https://github.com/python/typeshed/pull/7860#discussion_r876692941

### DIFF
--- a/stdlib/builtins.pyi
+++ b/stdlib/builtins.pyi
@@ -1192,6 +1192,8 @@ class filter(Iterator[_T], Generic[_T]):
 
 def format(__value: object, __format_spec: str = ...) -> str: ...  # TODO unicode
 @overload
+def getattr(__o: BaseException, __name: Literal["__notes__"], __default: list[str] = ...) -> list[str]: ...
+@overload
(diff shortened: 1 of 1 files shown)


AlexWaygood:
> 
> ```diff
> + homeassistant/components/synology_dsm/__init__.py:114: error: "List[str]" has no attribute "macs"  [attr-defined]
> + homeassistant/components/canary/camera.py:134: error: Argument 2 to "async_get_image" has incompatible type "List[str]"; expected "str"  [arg-type]
> ```

These both look like mypy bugs of some kind to me, but I think the benefit of this PR is pretty marginal anyway, so I don't think it's really worth introducing any false-positive errors for the sake of this PR.

JelleZijlstra:
Likely the new overload gets picked in unexpected places when the arguments are Any. This may have worse effects in pyright because pyright's overload picking algorithm is more likely to pick the first overload.

PR #6560, opened 2021-12-10
Add a `structseq` class to `_typeshed`

AlexWaygood:
Fixes #6546.

PR #7972, opened 2022-05-28
Add stubs for dj-database-url

XF-FW:
Adds stubs for https://github.com/jazzband/dj-database-url/

Thanks to @sbdchd for providing a great starting point here: https://github.com/chdsbd/kodiak/pull/559 .

This will also close: https://github.com/jazzband/dj-database-url/issues/135


stubs/dj-database-url/dj_database_url.pyi
```diff
@@ -0,0 +1,19 @@
+from collections.abc import Mapping
+from typing_extensions import TypedDict
+
+DEFAULT_ENV: str
+SCHEMES: Mapping[str, str]
```
AlexWaygood, line 5: Why use `Mapping` here rather than `dict`?

stubs/dj-database-url/dj_database_url.pyi
```diff
@@ -0,0 +1,19 @@
...
+
+class _DBConfig(TypedDict):
+    NAME: str
+    USER: str
+    PASSWORD: str
+    HOST: str
+    PORT: str
+    CONN_MAX_AGE: int
+    ENGINE: str
```
AlexWaygood, line 14: Not all keys will _necessarily_ always be in the dictionary returned from `parse()` and `config`. In fact, if I'm reading the source code correctly, the only key that's _guaranteed_ to be present is `"ENGINE"`. Ideally we'd use [PEP 655](https://peps.python.org/pep-0655/) to express that here, but not all type checkers support PEP 655 yet, so the best we can do in typeshed is something like this. The "required" key is given in `_DBConfigBase`, and the "optional" keys are given in the subclass, `_DBConfig`, which is marked with the `total=False` keyword (see [the description of this trick in the `typing` docs](https://docs.python.org/3/library/typing.html#typing.TypedDict.__optional_keys__)).

You also missed the `"OPTIONS"` key (https://github.com/jazzband/dj-database-url/blob/836740dd814878cfa9957a0a68b545060ba89c10/dj_database_url.py#L157-L158) :)

```suggestion
class _DBConfigBase(TypedDict);
    NAME: str

class _DBConfig(_DBConfigBase, total=False):
    USER: str
    PASSWORD: str
    HOST: str
    PORT: str
    CONN_MAX_AGE: int
    OPTIONS: dict[str, Any]
    ENGINE: str
```

AlexWaygood (commented):
Thanks! Just two comments

stubs/dj-database-url/dj_database_url.pyi
XF-FW: Ah, it was the default from create_baseline_stubs. Fixed.

AlexWaygood: Nit for future reference btw: please don't force-push on typeshed PRs. All typeshed PRs are squash-merged at the end of the day, so [it's fine to make a mess](https://github.com/python/typeshed/pull/6560). Also, force-pushing interacts badly with GitHub's UI -- it makes it harder for reviewers to see what changed between commits :)

stubs/dj-database-url/dj_database_url.pyi
XF-FW: Yeah, I think you're right on all points. I've used your suggested code to fix the issues pointed out.

Thanks for the review. I really appreciate it.
Also, thanks for letting me know about PEP 655 and the `total=False` trick.

stubs/dj-database-url/dj_database_url.pyi
XF-FW: Ah, sorry about that. Habits. It won't happen again. 

Thanks for letting me know.

stubs/dj-database-url/dj_database_url.pyi
AlexWaygood: > Thanks for the review. I really appreciate it.

Pleasure! :D

AlexWaygood (approved):
Thanks!

palfrey:
The `TypedDict` for this only has the fields for dj-database-url, not [the rest of them](https://docs.djangoproject.com/en/4.0/ref/settings/#databases). Even with the `total=False` setting, I still get `error: Extra key "TEST" for TypedDict "_DBConfig"` out of mypy when trying to set the `TEST` setting later on.

PR #7975, opened 2022-05-28
`builtins.sum`: Add overload for `bool` special case

### LINKED ISSUE #7974: problem with the type definition of sum
Everything is explained in https://github.com/microsoft/pylance-release/issues/2865. I copy it to avoid redirection.

*Environment data*

Language Server version: 2022.5.3
OS and version: macOS 10.14.6 (Mojave)
Python version (& distribution if applicable, e.g. Anaconda): 3.10.4
Code Snippet

```python
def test(N:int) -> bool:
  from random import choice
  return sum(a for a in [choice([True,False]) for _ in range(N)])==2
```

*Expected behavior*

This code should be perfectly fine.

*Actual behavior*

Problem detected on the last line
Pylance computes that it will always return False. Because it implies that sum on bool is always 0 or a bool.

```
Condition will always evaluate to False since the types "bool | Literal[0]" and "Literal[2]"
have no overlapPylance[reportUnnecessaryComparison]
(https://github.com/microsoft/pylance-release/blob/main/DIAGNOSTIC_SEVERITY_RULES.md#diagnostic-severity-rules)
```

Two problems here :
1/ bool should be considered as int, because they are ! (True+True==2)
2/ Even with ==1 at the end of the line, pylance seems to think that (True==1) is False.

________

Answer of Eric Traut when I posted this bug report on pylance :

I'm glad to see that you're using the reportUnnecessaryComparison diagnostic check! It's a strict-mode check, so most pylance users are not aware of it.

This looks like a bug in the type definition of sum, which is defined in the typeshed stubs [here](https://github.com/python/typeshed/blob/1eef7365b9f68f6af023da731369096f55437654/stdlib/builtins.pyi#L1552):

```python
def sum(__iterable: Iterable[_SumT]) -> _SumT | Literal[0]: ...
```
That leads to the following:

```python
x = sum([True, True, True])
reveal_type(x) # bool | Literal[0]
```

However, the correct (runtime) type of this expression is int.

So the bug will need to be fixed in the typeshed stubs. It will probably require a special-case overload for sum to properly model its behavior. Please file a bug (or a PR) in the typeshed project.

Here are a few suggested workarounds:

Use ints rather than bools in your code:
```python
def test(N: int) -> bool:
    return sum([choice([1, 0]) for _ in range(N)]) == 2
```
Add a # pyright: ignore[reportUnnecessaryComparison] or # type: ignore comment to the end of the line.
Disable that diagnostic check for the file by adding a # pyright: reportUnnecessaryComparison=false comment at the top of the file.

AlexWaygood:
Closes #7974

### DIFF
--- a/stdlib/builtins.pyi
+++ b/stdlib/builtins.pyi
@@ -1548,6 +1548,14 @@ def sorted(__iterable: Iterable[_T], *, key: Callable[[_T], SupportsRichComparis
 _SumT = TypeVar("_SumT", bound=SupportsAdd)
 _SumS = TypeVar("_SumS", bound=SupportsAdd)
 
+if sys.version_info >= (3, 8):
+    @overload
+    def sum(__iterable: Iterable[bool], start: int = ...) -> int: ...
+
+else:
+    @overload
+    def sum(__iterable: Iterable[bool], __start: int = ...) -> int: ...
+
 @overload
 def sum(__iterable: Iterable[_SumT]) -> _SumT | Literal[0]: ...
 
--- a/stdlib/builtins.pyi
+++ b/stdlib/builtins.pyi
@@ -1548,6 +1548,10 @@ def sorted(__iterable: Iterable[_T], *, key: Callable[[_T], SupportsRichComparis
 _SumT = TypeVar("_SumT", bound=SupportsAdd)
 _SumS = TypeVar("_SumS", bound=SupportsAdd)
 
+# In general, the return type of `x + x` is *not* guaranteed to be the same type as x.
+# However, we can't express that in the stub for `sum()`
+# without creating many false-positive errors (see #7578).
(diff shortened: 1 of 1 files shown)


stdlib/builtins.pyi
```diff
@@ -1548,6 +1548,14 @@ def sorted(__iterable: Iterable[_T], *, key: Callable[[_T], SupportsRichComparis
 _SumT = TypeVar("_SumT", bound=SupportsAdd)
 _SumS = TypeVar("_SumS", bound=SupportsAdd)
 
+if sys.version_info >= (3, 8):
```
JelleZijlstra, line 1551: I think this merits a comment

srittau (approved):

JelleZijlstra (approved):

PR #7976, opened 2022-05-28
pyright: enable 'reportOverlappingOverload' check

AlexWaygood:
I don't think there's a good reason to have this check enabled for mypy but not for pyright.


srittau:
I don't like all the extra type ignores that are now required. In fact, I wonder whether we shouldn't disable the corresponding mypy check. Overlapping overloads are often required in stubs, but only rarely find an actual problem.

AlexWaygood:
> Overlapping overloads are often required in stubs, but only rarely find an actual problem.

This is true. I'm open to doing the opposite here, and disabling the mypy check.

AlexWaygood:
You're right: this didn't expose any new bugs, so I'm closing this for now. I'll consider filing another PR to fiddle with the mypy config.

PR #7977, opened 2022-05-28
`mypy_test.py`: Skip `Flask-SQLAlchemy`

AlexWaygood:
Ever since we upgraded to mypy 0.960, I've hit https://github.com/python/mypy/issues/11899 every second time I run `mypy_test.py` locally. Mypy always crashes on `Flask-SQLAlchemy`, so I propose we skip `mypy_test` on those stubs for now, unless and until https://github.com/python/mypy/pull/12324 is merged.

### DIFF
--- a/tests/mypy_test.py
+++ b/tests/mypy_test.py
@@ -369,8 +369,9 @@ def test_third_party_stubs(code: int, major: int, minor: int, args: argparse.Nam
     files_checked = 0
 
     for distribution in sorted(os.listdir("stubs")):
-        if distribution == "SQLAlchemy":
-            continue  # Crashes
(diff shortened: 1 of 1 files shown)


hauntsaninja:
I couldn't repro, could you confirm whether https://github.com/python/mypy/pull/12324 fixes?

AlexWaygood:
> I couldn't repro, could you confirm whether [python/mypy#12324](https://github.com/python/mypy/pull/12324) fixes?

I'll try to confirm tomorrow tomorrow if python/mypy#12324 fixes the issue.

It happened very consistently for me locally. Deleting the cache sorted it, but it would then just crash again after running `mypy_test` twice. And always with the exact same traceback as in python/mypy#11899

hauntsaninja:
Thanks for checking!

PR #7978, opened 2022-05-28
stripe: update a few signatures

sbdchd:
Fill out a few more types on existing methods and add a new method to `WebhookSignature`.

code for reference: https://github.com/stripe/stripe-python/blob/master/stripe


stubs/stripe/stripe/api_resources/abstract/createable_api_resource.pyi
```diff
@@ -6,10 +7,10 @@ from stripe.api_resources.abstract.api_resource import APIResource as APIResourc
...
     def create(
-        cls,
+        cls: type[Self],
         api_key: Any | None = ...,
         idempotency_key: Any | None = ...,
         stripe_version: Any | None = ...,
         stripe_account: Any | None = ...,
         **params,
-    ): ...
+    ) -> Self: ...
```
AlexWaygood, line 16: Any tips on how to review this change? The code for this method is a bit of a maze, and playing around in the REPL seems sort-of nontrivial as well, for this library.

stubs/stripe/stripe/api_resources/invoice.pyi
```diff
@@ -11,12 +11,12 @@ from stripe.api_resources.abstract import (
...
-    def finalize_invoice(self, idempotency_key: Any | None = ..., **params): ...
-    def mark_uncollectible(self, idempotency_key: Any | None = ..., **params): ...
-    def pay(self, idempotency_key: Any | None = ..., **params): ...
-    def send_invoice(self, idempotency_key: Any | None = ..., **params): ...
-    def void_invoice(self, idempotency_key: Any | None = ..., **params): ...
+    def finalize_invoice(self, idempotency_key: str | None = ..., **params) -> Invoice: ...
+    def mark_uncollectible(self, idempotency_key: str | None = ..., **params) -> Invoice: ...
+    def pay(self, idempotency_key: str | None = ..., **params) -> Invoice: ...
+    def send_invoice(self, idempotency_key: str | None = ..., **params) -> Invoice: ...
+    def void_invoice(self, idempotency_key: str | None = ..., **params) -> Invoice: ...
```
AlexWaygood, line 18: These all look like they return `self` at runtime: https://github.com/stripe/stripe-python/blob/97ee54f5e54b4af1d474de3059b26a6b92dca53d/stripe/api_resources/invoice.py#L28-L56

```suggestion
    def finalize_invoice(self: Self, idempotency_key: str | None = ..., **params) -> Self: ...
    def mark_uncollectible(self: Self, idempotency_key: str | None = ..., **params) -> Self: ...
    def pay(self: Self, idempotency_key: str | None = ..., **params) -> Self: ...
    def send_invoice(self: Self, idempotency_key: str | None = ..., **params) -> Self: ...
    def void_invoice(self: Self, idempotency_key: str | None = ..., **params) -> Self: ...
```

(You'll need to import `Self` from `_typeshed`.)

AlexWaygood, line 18: These all look like they return `self` at runtime: https://github.com/stripe/stripe-python/blob/97ee54f5e54b4af1d474de3059b26a6b92dca53d/stripe/api_resources/invoice.py#L28-L56

```suggestion
    def finalize_invoice(self: Self, idempotency_key: str | None = ..., **params) -> Self: ...
    def mark_uncollectible(self: Self, idempotency_key: str | None = ..., **params) -> Self: ...
    def pay(self: Self, idempotency_key: str | None = ..., **params) -> Self: ...
    def send_invoice(self: Self, idempotency_key: str | None = ..., **params) -> Self: ...
    def void_invoice(self: Self, idempotency_key: str | None = ..., **params) -> Self: ...
```

stubs/stripe/stripe/api_resources/payment_method.pyi
```diff
@@ -9,5 +7,5 @@ from stripe.api_resources.abstract import (
 
 class PaymentMethod(CreateableAPIResource, ListableAPIResource, UpdateableAPIResource):
     OBJECT_NAME: str
-    def attach(self, idempotency_key: Any | None = ..., **params): ...
-    def detach(self, idempotency_key: Any | None = ..., **params): ...
+    def attach(self, idempotency_key: str | None = ..., **params) -> PaymentMethod: ...
+    def detach(self, idempotency_key: str | None = ..., **params) -> PaymentMethod: ...
```
AlexWaygood, line 11: These return `self` at runtime:

```suggestion
    def attach(self: Self, idempotency_key: str | None = ..., **params) -> Self: ...
    def detach(self: Self, idempotency_key: str | None = ..., **params) -> Self: ...
```

https://github.com/stripe/stripe-python/blob/97ee54f5e54b4af1d474de3059b26a6b92dca53d/stripe/api_resources/payment_method.py#L20-L30

AlexWaygood, line 11: These return `self` at runtime:

```suggestion
    def attach(self: Self, idempotency_key: str | None = ..., **params) -> Self: ...
    def detach(self: Self, idempotency_key: str | None = ..., **params) -> Self: ...
```

https://github.com/stripe/stripe-python/blob/97ee54f5e54b4af1d474de3059b26a6b92dca53d/stripe/api_resources/payment_method.py#L20-L30

stubs/stripe/stripe/webhook.pyi
```diff
@@ -1,13 +1,15 @@
...
-    def construct_event(payload, sig_header, secret, tolerance=..., api_key: Any | None = ...): ...
+    def construct_event(
+        payload: bytes | str, sig_header: str, secret: str, tolerance: int = ..., api_key: str | None = ...
+    ) -> Event: ...
 
 class WebhookSignature:
     EXPECTED_SCHEME: str
     @classmethod
-    def verify_header(cls, payload, header, secret, tolerance: Any | None = ...): ...
+    def verify_header(cls, payload: bytes | str, header: str, secret: str, tolerance: int | None = ...) -> bool: ...
```
AlexWaygood, line 13: It actually only ever returns `True` (you'll need to import `Literal` from `typing_extensions`)

```suggestion
    def verify_header(cls, payload: bytes | str, header: str, secret: str, tolerance: int | None = ...) -> Literal[True]: ...
```

AlexWaygood, line 13: It actually only ever returns `True` (you'll need to import `Literal` from `typing_extensions`)

```suggestion
    def verify_header(cls, payload: bytes | str, header: str, secret: str, tolerance: int | None = ...) -> Literal[True]: ...
```

AlexWaygood (changes requested):
Thanks! This isn't a full review, but here's some things I spotted.

Do you have any tips on how to review this PR? For some of these modules, figuring out the types from the source code isn't particularly easy, nor is playing around in the REPL. The docs don't seem to be hugely helpful either :/

stubs/stripe/stripe/api_resources/abstract/createable_api_resource.pyi
AlexWaygood, line 16: This I'm having trouble figuring out from looking at the source code, can you copy a REPL demonstration for me to see?

stubs/stripe/stripe/api_resources/abstract/listable_api_resource.pyi
```diff
@@ -1,10 +1,14 @@
...
 
 class ListableAPIResource(APIResource):
     @classmethod
-    def auto_paging_iter(cls, *args, **params): ...
+    def auto_paging_iter(cls, *args, **params) -> Iterator[Any]: ...
     @classmethod
-    def list(cls, api_key: Any | None = ..., stripe_version: Any | None = ..., stripe_account: Any | None = ..., **params): ...
+    def list(
+        cls, api_key: Any | None = ..., stripe_version: Any | None = ..., stripe_account: Any | None = ..., **params
+    ) -> ListObject: ...
```
AlexWaygood, line 14: Same here, the source code is a bit of a maze, could I possibly see a REPL demonstration?

stubs/stripe/stripe/api_resources/abstract/updateable_api_resource.pyi
```diff
@@ -1,8 +1,8 @@
-from typing import Any
+from _typeshed import Self
 
 from stripe.api_resources.abstract.api_resource import APIResource as APIResource
 
 class UpdateableAPIResource(APIResource):
     @classmethod
-    def modify(cls, sid, **params): ...
-    def save(self, idempotency_key: Any | None = ...): ...
+    def modify(cls: type[Self], sid: str, **params) -> Self: ...
```
AlexWaygood, line 7: Here too

stubs/stripe/stripe/api_resources/application_fee_refund.pyi
```diff
@@ -1,12 +1,13 @@
...
 
 from stripe.api_resources import ApplicationFee as ApplicationFee
 from stripe.api_resources.abstract import UpdateableAPIResource as UpdateableAPIResource
 
 class ApplicationFeeRefund(UpdateableAPIResource):
     OBJECT_NAME: str
     @classmethod
-    def modify(cls, fee, sid, **params): ...
-    def instance_url(self): ...
+    def modify(cls: type[Self], fee, sid: str, **params) -> Self: ...  # type: ignore[override]
```
AlexWaygood, line 10: And here

stubs/stripe/stripe/api_resources/invoice.pyi
```diff
@@ -11,12 +12,12 @@ from stripe.api_resources.abstract import (
...
+    def finalize_invoice(self: Self, idempotency_key: str | None = ..., **params) -> Self: ...
+    def mark_uncollectible(self: Self, idempotency_key: str | None = ..., **params) -> Self: ...
+    def pay(self: Self, idempotency_key: str | None = ..., **params) -> Self: ...
+    def send_invoice(self: Self, idempotency_key: str | None = ..., **params) -> Self: ...
+    def void_invoice(self: Self, idempotency_key: str | None = ..., **params) -> Self: ...
     @classmethod
     def upcoming(
         cls, api_key: Any | None = ..., stripe_version: Any | None = ..., stripe_account: Any | None = ..., **params
-    ): ...
+    ) -> Invoice: ...
```
AlexWaygood, line 23: And here (also, are you sure this is `Invoice` and not `Self`? Seems a bit suspect given some of the other annotations 🤔)

stubs/stripe/stripe/api_resources/list_object.pyi
```diff
@@ -22,7 +22,7 @@ class ListObject(StripeObject):
     def __iter__(self): ...
     def __len__(self): ...
```
AlexWaygood, line 23: May as well annotate this while we're here

```suggestion
    def __len__(self) -> int: ...
```

stubs/stripe/stripe/api_resources/list_object.pyi
```diff
@@ -22,7 +22,7 @@ class ListObject(StripeObject):
...
     def __reversed__(self): ...
-    def auto_paging_iter(self) -> Generator[Any, None, None]: ...
+    def auto_paging_iter(self) -> Iterator[Any]: ...
     @classmethod
     def empty_list(cls, api_key: Any | None = ..., stripe_version: Any | None = ..., stripe_account: Any | None = ...): ...
     @property
```
AlexWaygood, line 28: I can't add a suggestion but the `is_empty` property seems like it returns a `bool`, may as well update it while we're here

stubs/stripe/stripe/api_resources/invoice.pyi
sbdchd: yeah I think it should be Self, went with Invoice since I was being lazy and I think most people probably aren't subclassing these

stubs/stripe/stripe/api_resources/abstract/createable_api_resource.pyi
sbdchd:
```python
>>> type(stripe.Customer.create())
<class 'stripe.api_resources.customer.Customer'>
```

essentially all the resources inherit from `CreateableAPIResource` and that `.create` method is used to create the resource instance

stubs/stripe/stripe/api_resources/application_fee_refund.pyi
sbdchd: this class inherits from the updatable api resource so it has the same behavior as Customer

```python
>>> customer = stripe.Customer.create(); type(customer.modify(customer.id, name="foo"))
<class 'stripe.api_resources.customer.Customer'>
```

stubs/stripe/stripe/api_resources/invoice.pyi
sbdchd:
```python
>>> type(stripe.Invoice.upcoming(subscription="sub_1JrD7ZGBs2RJTknIQSCKG7Xk"))
<class 'stripe.api_resources.invoice.Invoice'>
```

sbdchd: actually, looking at the code I think this will always be an Invoice since it hits the `/upcoming` invoices endpoint and then creates a Stripe object based off that (which will always be an invoice)

https://github.com/stripe/stripe-python/blob/97ee54f5e54b4af1d474de3059b26a6b92dca53d/stripe/api_resources/invoice.py#L66-L77

AlexWaygood (approved):
I haven't been able to confirm every detail of these changes, but it all looks reasonable to me, and mypy_primer has no complaints.

PR #7979, opened 2022-05-28
Fix dj_database_url required key

XF-FW:
Hey. It's me again.

I've happily installed `types-dj-database-url`, but it seems I've made a small mistake with the PR from earlier. For reference: #7972 .

Required key is ENGINE and not NAME. This fixes that issue.

A question, I have, you might be able to answer me: I have a project that uses dj-database-url. If I wanted to test these stubs against my project, how would I go about doing so? 

Thanks.

### DIFF
--- a/stubs/dj-database-url/dj_database_url.pyi
+++ b/stubs/dj-database-url/dj_database_url.pyi
@@ -5,16 +5,16 @@ DEFAULT_ENV: str
 SCHEMES: dict[str, str]
 
 class _DBConfigBase(TypedDict):
-    NAME: str
+    ENGINE: str
 
 class _DBConfig(_DBConfigBase, total=False):
+    NAME: str
     USER: str
     PASSWORD: str
     HOST: str
     PORT: str
     CONN_MAX_AGE: int
     OPTIONS: dict[str, Any]
-    ENGINE: str
 
 def parse(url: str, engine: str | None = ..., conn_max_age: int = ..., ssl_require: bool = ...) -> _DBConfig: ...
 def config(
--- a/stubs/dj-database-url/dj_database_url.pyi
+++ b/stubs/dj-database-url/dj_database_url.pyi
@@ -4,10 +4,8 @@ from typing_extensions import TypedDict
 DEFAULT_ENV: str
 SCHEMES: dict[str, str]
 
-class _DBConfigBase(TypedDict):
+class _DBConfig(TypedDict, total=False):
     ENGINE: str
-
-class _DBConfig(_DBConfigBase, total=False):
     NAME: str
     USER: str
     PASSWORD: str


AlexWaygood:
Interesting. From my reading of the source code, NAME is guaranteed to be present in the dictionary in both code paths:

- https://github.com/jazzband/dj-database-url/blob/931a4e4198e255c4a1b2211a48c09b94de8018a7/dj_database_url.py#L85
- https://github.com/jazzband/dj-database-url/blob/931a4e4198e255c4a1b2211a48c09b94de8018a7/dj_database_url.py#L130

But ENGINE is only guaranteed to be present in the dictionary in one of the two code paths:

- https://github.com/jazzband/dj-database-url/blob/931a4e4198e255c4a1b2211a48c09b94de8018a7/dj_database_url.py#L85
- https://github.com/jazzband/dj-database-url/blob/931a4e4198e255c4a1b2211a48c09b94de8018a7/dj_database_url.py#L163

Where am I misreading the code? :)

XF-FW:
You're not. That's indeed the case, I had misinterpented the problem, but there's still one:

Assigning, for example, `DATABASES["default"]` to a call to `dj_database_url.config`, will set its type to `_DBConfig`, *but* Django itself, does not require either `NAME` or `ENGINE`:

https://github.com/django/django/blob/0dd29209091280ccf34e07c9468746c396b7778e/django/db/utils.py#L149

My problem is that I have something like this on the settings of my Django project:
```
import sys

import dj_database_url

# Database
# https://docs.djangoproject.com/en/4.0/ref/settings/#databases
DATABASES = {}
DATABASES["default"] = dj_database_url.config(
    conn_max_age=600, default="postgres://postgres:postgres@localhost/my-app"
)
DATABASES["default"]["OPTIONS"] = {"application_name": "my-app"}

# We don't need a live database to check if there are pending migrations
if_checking_if_has_pending_migrations = {
    "makemigrations",
    "--check",
    "--dry-run",
}.issubset(set(sys.argv))
if if_checking_if_has_pending_migrations:  # pragma: no cover
    DATABASES = {"default": {"ENGINE": "django.db.backends.dummy"}}
```

Since mypy got its type from `dj_database_url`, I'm getting this on the last line: `error: Missing key "NAME" for TypedDict "_DBConfig"`. 

That's the problem. I think the solution, would be to have everything as optional, as that's what django expects.

Alternatively, I can just set DATABASES to empty dict after the `if`, as that's the default value, but I like the explicitness and I think the above solution is the correct one.

XF-FW:
I've made the change to make everything optional.

Also, for reference, `django-stubs` defines `DATABASES` as `Dict[str, Dict[str, Any]]`:

https://github.com/typeddjango/django-stubs/blob/43b082850c8ca4e00194165f02be798a0943fb21/django-stubs/conf/global_settings.pyi#L80

I'm beginning, to think, that's the best solution, to make sure it behaves nicely. Especially, since there are more keys missing from what `DATABASES[alias]` accepts. It's missing, for example, `ATOMIC_REQUESTS` and `AUTOCOMMIT`.

What do you think?

AlexWaygood (approved):
Got it. Thanks for the fix!

> If I wanted to test these stubs against my project, how would I go about doing so?

If you want to test changes to these stubs prior to filing a PR, you can make the changes in your local clone of typeshed, and then point mypy towards these stubs by using the `MYPYPATH` environment variable and setting it to the `stubs` directory in your local clone of typeshed. If your project's open-source, you could also make a request to have it added to the [mypy_primer](https://github.com/hauntsaninja/mypy_primer) project :)

PR #7980, opened 2022-05-29
jsonschema: {relative,absolute}_path can hold ints

DMRobertson:
In https://github.com/matrix-org/synapse/issues/12901 I noticed an problem with the changes to the jsonschema stubs in #7950.

This change remedies that by teaching mypy that paths, relative paths and absolute paths on Error objects can hold strings, ints, or a mixture of the two. For evidence that this is the case, see e.g. https://github.com/python-jsonschema/jsonschema/blob/2e2832463e8de7977217d377ca0553c10f6fb48c/jsonschema/tests/test_exceptions.py#L246-L247 or the test case I provided in https://github.com/python/typeshed/pull/7950#discussion_r884279863.

### DIFF
--- a/stubs/jsonschema/jsonschema/exceptions.pyi
+++ b/stubs/jsonschema/jsonschema/exceptions.pyi
@@ -1,7 +1,7 @@
 from _typeshed import Self, SupportsRichComparison
 from collections import deque
 from collections.abc import Callable, Container, Iterable, Sequence
-from typing import Any, Union
+from typing import Any
 
 from jsonschema import _utils, protocols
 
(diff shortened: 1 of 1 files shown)


AlexWaygood:
@sirosen, does this change make sense to you? :)

sirosen:
Yes, this change is definitely an improvement -- the OP's comment/example on the earlier PR is accurate, and I'm sorry I didn't catch this earlier.

I'm looking now to see if there are other `[str]` containers which should be `[str | int]`. But I'm not sure we want to block this while I do that?

AlexWaygood:
> Yes, this change is definitely an improvement -- the OP's comment/example on the earlier PR is accurate, and I'm sorry I didn't catch this earlier.
> 
> I'm looking now to see if there are other `[str]` containers which should be `[str | int]`. But I'm not sure we want to block this while I do that?

I'll merge for now; further fixes can be made in further PRs if necessary.

AlexWaygood (approved):
Thanks for the fix, and thanks @sirosen for confirming!

PR #7987, opened 2022-05-30
Add `__setstate__` for BaseException

pranavrajpal:
BaseException has a `__setstate__` method at runtime that sets `__dict__` to the passed in parameter. See [here](https://github.com/python/cpython/blob/61e008a4ab0b444342f9186efb57207d6fb05092/Objects/exceptions.c#L159) for the implementation.

This came up in python/mypy#12901 (see the first Pandera error in https://github.com/python/mypy/pull/12901#issuecomment-1140538339).


srittau (approved):

srittau:
Thanks!

Issue #7947, opened 2022-05-25
Issue with PyYaml add_constructor


PaulBurridge-Nasstar:
Hi,

Receiving the following error message from mypy:

```
error: Call to untyped function "add_constructor" of "BaseConstructor" in typed context
```

for the following code

```
loader = yaml.SafeLoader
loader.add_constructor(
    "!ReplaceWithConfig", partial(replace_with_config_constructor, config=config)
)
```

Versions are:

```
mypy==0.950
PyYAML==6.0
types-PyYAML==6.0.7
```

Thanks in advance for any assistance.

nakatomotoi:
@PaulBurridge-Nasstar  @Akuli
Hi, may I work on this issue?
This is my first issue to work on in typeshed.

AlexWaygood:
> @PaulBurridge-Nasstar @Akuli Hi, may I work on this issue? This is my first issue to work on in typeshed.

@nakatomotoi feel free! PRs are very welcome here at typeshed, no need to ask before filing a PR :)

PR #7989, opened 2022-05-30
Add types to add_constructor

nakatomotoi:
Solves https://github.com/python/typeshed/issues/7947

I add types to `add_constructor()` referring to [this code](https://github.com/python/typeshed/blob/master/stubs/PyYAML/yaml/__init__.pyi#L286-L289).


Akuli:
There's also `an` `add_constructor()` function in `__init__.pyi`. Presumably the two should be annotated in the same way.

AlexWaygood:
> So I think the type for the `constructor` parameter should probably be something like `Callable[[loader.Loader, nodes.Node], Any]`.

It looks like you actually did something very similar to this in your first commit, but then bailed due to the stubtest check in our CI failing. Try it again, but don't use `from loader import *`. That has a special meaning in a stub file — in a stub file, all things imported using a `from foo import *` import are considered "publicly exported". If you use `from loader import Loader` instead, it will be considered a private-to-the-stub import, and stubtest will pass.

stubs/PyYAML/yaml/constructor.pyi
```diff
@@ -1,9 +1,13 @@
+from collections.abc import Callable
 from typing import Any, Pattern
 from typing_extensions import TypeAlias
 
 from yaml.error import MarkedYAMLError
 from yaml.nodes import ScalarNode
 
+from loader import Loader
+from nodes import Node
```
AlexWaygood, line 9:
```suggestion
from yaml.loader import Loader
from yaml.nodes import Node
```

AlexWaygood:
Okay, I think we're running into difficulties due to the fact that parameter types are contravariant (ew). Not sure what the best solution is here -- it might be best to bail and just use `Callable[[Any, Any], Any]`.

@Akuli, any ideas?

AlexWaygood:
@nakatomotoi, here's a hack that allows us to get covariant behaviour in parameter types, and might resolve the mypy_primer false-positive errors. Could you try it?

```python
from typing import TypeVar

from yaml.loader import Loader
from yaml.nodes import Node

_L = TypeVar("_L", bound=Loader)
_N = TypeVar("_N", bound=Node)

class BaseConstructor:
    # Use TypeVars to get covariant behaviour in the parameters for the Callable
    def add_constructor(cls, tag: str, constructor: Callable[[_L, _N], Any]) -> None: ...
```

stubs/PyYAML/yaml/constructor.pyi
```diff
@@ -1,8 +1,13 @@
-from typing import Any, Pattern
+from collections.abc import Callable
+from typing import Any, Pattern, TypeVar
 from typing_extensions import TypeAlias
 
 from yaml.error import MarkedYAMLError
-from yaml.nodes import ScalarNode
+from yaml.loader import Loader
```
AlexWaygood, line 6:
```suggestion
from yaml.loader import Loader, BaseLoader, FullLoader, SafeLoader, UnsafeLoader
```

stubs/PyYAML/yaml/constructor.pyi
```diff
@@ -1,8 +1,13 @@
...
+from yaml.nodes import Node, ScalarNode
+
+_L = TypeVar("_L", bound=Loader)
```
AlexWaygood, line 9:
```suggestion
_L = TypeVar("_L", bound=Loader | BaseLoader | FullLoader | SafeLoader | UnsafeLoader)
```

stubs/PyYAML/yaml/constructor.pyi
```diff
@@ -27,7 +32,7 @@ class BaseConstructor:
     def construct_mapping(self, node, deep=...): ...
     def construct_pairs(self, node, deep=...): ...
     @classmethod
-    def add_constructor(cls, tag, constructor): ...
+    def add_constructor(cls, tag: str, constructor: Callable[[_L, _N], Any]) -> None: ...
```
AlexWaygood, line 35:
```suggestion
    # Use typevars so we can have covariant behaviour in the parameter types
    def add_constructor(cls, tag: str, constructor: Callable[[_L, _N], Any]) -> None: ...
```

AlexWaygood:
> There's also an `add_constructor()` function in `__init__.pyi`. Presumably the two should be annotated in the same way.

I agree with @Akuli here, but this can wait for a separate PR.

PR #7990, opened 2022-05-30
Fix exception types for jsonschema._format

sirosen:
The annotated type for the `raises` argument on format checkers was

    Exception | tuple[Exception, ...]

when it should read

    type[Exception] | tuple[type[Exception], ...]

---

This is a mistake I introduced in #7950 . So far, there have been two or three issues found with those changes, depending on how you count. I will try to make smaller changes to `jsonschema` stubs in the future to avoid creating this situation again.


stubs/jsonschema/jsonschema/_format.pyi
```diff
@@ -2,14 +2,15 @@ from collections.abc import Callable, Iterable
 from typing import Any, TypeVar
 
 _F = TypeVar("_F", bound=Callable[..., Any])
+_RaisesType = type[Exception] | tuple[type[Exception], ...]
```
AlexWaygood, line 5: You've hit a known mypy bug in its handling of PEP-604 type aliases. The workaround is to use the old syntax (`Union`) here instead. The rest of our CI won't complain about it in this instance.

sirosen: Ah, easy enough -- I'll have that done in just a second.

Also, thanks to you and everyone on typeshed for being _super_ responsive! It makes it very easy to contribute. ❤️

stubs/jsonschema/jsonschema/_format.pyi
```diff
@@ -1,15 +1,16 @@
 from collections.abc import Callable, Iterable
-from typing import Any, TypeVar
+from typing import Any, TypeVar, Union
 
 _F = TypeVar("_F", bound=Callable[..., Any])
+_RaisesType = Union[type[Exception], tuple[type[Exception], ...]]
```
srittau, line 5:
```suggestion
from typing import Any, TypeAlias, TypeVar, Union

_F = TypeVar("_F", bound=Callable[..., Any])
_RaisesType: TypeAlias = Union[type[Exception], tuple[type[Exception], ...]]
```

srittau (changes requested):

stubs/jsonschema/jsonschema/_format.pyi
sirosen: Thanks! I was just now trying to figure out how I'm supposed to write this. I'm still not accustomed to using `TypeAlias` yet.

AlexWaygood: Need to import `TypeAlias` from `typing_extensions`; `typing.TypeAlias` is only a thing on 3.10+

srittau:
Sorry, my bad, `TypeAlias` needs to be imported from `typing_extensions`.

srittau (approved):

stubs/jsonschema/jsonschema/_format.pyi
AlexWaygood: > thanks to you and everyone on typeshed for being super responsive! It makes it very easy to contribute. ❤️

I'm very glad to hear that! Also, don't beat yourself up too much about the regressions — it's nearly inevitable if you're adding annotations to stubs that currently don't have many type hints at all. One way we could work to reduce these in the future, however, would be to add more open-source projects using `jsonschema` to mypy_primer.

sirosen: I'll give that a shot. [`check-jsonschema`](https://github.com/python-jsonschema/check-jsonschema) is where I'm doing my own work.
And there was a PR from the maintainers of https://github.com/matrix-org/synapse , so perhaps I can integrate that as well. I'll have to see how they have mypy setup, of course.

PR #7999, opened 2022-05-31
stubs/requests: expand types for files

### LINKED ISSUE #7998: requests: allow passing a list of tuples for `files`
Relates to https://github.com/python/typeshed/issues/7724

The `files` argument in the `requests` library is incompletely typed. The current types include the `Mapping` type (over which there has been [considerable discourse already](https://github.com/python/typeshed/issues/7724)), but the official requests documentation states:

> You can send multiple files in one request. [...]
> To do that, just set files to a list of tuples of (form_field_name, file_info):

The [official example code given](https://requests.readthedocs.io/en/latest/user/advanced/#post-multiple-multipart-encoded-files) is:

```python
url = 'https://httpbin.org/post'
multiple_files = [
    ('images', ('foo.png', open('foo.png', 'rb'), 'image/png')),
    ('images', ('bar.png', open('bar.png', 'rb'), 'image/png'))]
r = requests.post(url, files=multiple_files)
r.text
```

This snippet currently fails typechecking with the latest stubs. A minimal reproduction [is available here](https://github.com/tommilligan/python-typeshed-repro-7724), which also confirms the code works as expected, but fails with.

```log
minimal.py:15: error: Argument "files" to "post" has incompatible type "List[Tuple[str, Tuple[str, BufferedReader, str]]]"; expected "Optional[Union[Mapping[str, Union[SupportsRead[Union[str, bytes]], str, bytes]], Mapping[str, Tuple[Optional[str], Union[SupportsRead[Union[str, bytes]], str, bytes]]], Mapping[str, Tuple[Optional[str], Union[SupportsRead[Union[str, bytes]], str, bytes], str]], Mapping[str, Tuple[Optional[str], Union[SupportsRead[Union[str, bytes]], str, bytes], str, MutableMapping[str, str]]]]]"
```

tommilligan:
Relates to https://github.com/python/typeshed/issues/7724
Closes #7998

This PR allows passing a sequence of tuple for the `files` argument, where applicable in the `requests` library.

Existing types have been aliased for readability, and to document that all type of file specification are supported in either the `Mapping` or `Sequence[tuple]` form.


stubs/requests/requests/sessions.pyi
```diff
@@ -48,11 +48,23 @@ class SessionRedirectMixin:
...
-    | Mapping[str, tuple[str | None, SupportsRead[str | bytes] | str | bytes, str]]
-    | Mapping[str, tuple[str | None, SupportsRead[str | bytes] | str | bytes, str, _TextMapping]]
+    Mapping[str, _FileContent]
+    | Mapping[str, _FileSpecTuple2]
+    | Mapping[str, _FileSpecTuple3]
+    | Mapping[str, _FileSpecTuple4]
+    | Sequence[tuple[str, _FileContent]]
+    | Sequence[tuple[str, _FileSpecTuple2]]
+    | Sequence[tuple[str, _FileSpecTuple3]]
+    | Sequence[tuple[str, _FileSpecTuple4]]
```
AlexWaygood, line 67: I don't think we need to arduously spell these out as separate kinds of mappings. From the source code, it looks like each `key, value` pair in `files` is checked individually to see what the structure is: https://github.com/psf/requests/blob/a5e7169e759817ca309f8271c3c4fa79393a22f5/requests/models.py#L177-L186.

Local testing appears to confirm that "mixing and matching" different structures in the same request seems to work fine, e.g.

```python
import requests

url = 'https://httpbin.org/post'
multiple_files = [
    ('images', ('foo.png', open('foo.png', 'rb'))),
    ('images', ('bar.png', open('bar.png', 'rb'), 'image/png')),
    ('images', ('baz.png', open('baz.png', 'rb'), 'image/png', {}))
]
r = requests.post(url, files=multiple_files)
print(r.status_code)
print(r.text)
```

As such, we can take advantage of covariance, and do this instead:

```python
_OneOrSeveralFiles: TypeAlias = _FileContent | _FileSpecTuple2 | _FileSpecTuple3 | _FileSpecTuple4

_Files : TypeAlias = (
    Mapping[str, _OneOrSeveralFiles]
    | Sequence[tuple[str, _OneOrSeveralFiles]]
)
```

AlexWaygood (commented):
Thanks! I dug around in the source code and tried out your repro, all looks good. Just one comment about how we can simplify the code a little, and also make the stub a little more permissive.

stubs/requests/requests/sessions.pyi
```diff
@@ -48,12 +48,16 @@ class SessionRedirectMixin:
...
-    | Mapping[str, tuple[str | None, SupportsRead[str | bytes] | str | bytes, str, _TextMapping]]
-)
+# Files can be passed as either a mapping, or a list of tuples.
+_FileContent: TypeAlias = SupportsRead[str | bytes] | str | bytes
+_FileName: TypeAlias = str | None
+_FileContentType: TypeAlias = str
+_FileCustomHeaders: TypeAlias = _TextMapping
+_FileSpecTuple2 = tuple[_FileName, _FileContent]
+_FileSpecTuple3 = tuple[_FileName, _FileContent, _FileContentType]
+_FileSpecTuple4 = tuple[_FileName, _FileContent, _FileContentType, _FileCustomHeaders]
```
AlexWaygood, line 58: Our CI isn't smart enough to catch it, but these should also be `TypeAlias`es

```suggestion
_FileSpecTuple2: TypeAlias = tuple[_FileName, _FileContent]
_FileSpecTuple3: TypeAlias = tuple[_FileName, _FileContent, _FileContentType]
_FileSpecTuple4: TypeAlias = tuple[_FileName, _FileContent, _FileContentType, _FileCustomHeaders]
```

stubs/requests/requests/sessions.pyi
```diff
@@ -48,12 +48,16 @@ class SessionRedirectMixin:
...
-    Mapping[str, SupportsRead[str | bytes] | str | bytes]
-    | Mapping[str, tuple[str | None, SupportsRead[str | bytes] | str | bytes]]
-    | Mapping[str, tuple[str | None, SupportsRead[str | bytes] | str | bytes, str]]
-    | Mapping[str, tuple[str | None, SupportsRead[str | bytes] | str | bytes, str, _TextMapping]]
-)
+# Files can be passed as either a mapping, or a list of tuples.
+_FileContent: TypeAlias = SupportsRead[str | bytes] | str | bytes
+_FileName: TypeAlias = str | None
+_FileContentType: TypeAlias = str
+_FileCustomHeaders: TypeAlias = _TextMapping
```
Akuli, line 55: `_TextMapping` is `MutableMapping[str, str]`. I think we should just delete it, because it's confusing, but in a separate PR. But for these headers:

- Does it really have to be mutable?
- Does it accept bytes keys or values?

Akuli (commented):
I'm mostly concerned about using `_TextMapping`. The rest are small nits.

stubs/requests/requests/sessions.pyi
tommilligan: Thanks; I saw there was already some discussion around this here: https://github.com/python/typeshed/issues/7724#issuecomment-1115767799

The outcome so far appears to be:
- It's _probably_ not required to be mutable for `headers`, however
- It's the same type as is used elsewhere for `proxies`, which _is_ mutated, and
- The upstream maintainers aren't committed to ensuring immutability for those arguments, and ask for a `dict` in the official documentation

So I definitely agree that should be a separate PR :grin: because it sounds like a thorny issue all round.

Akuli: Bytes seem to work for me. I tried `requests.post('http://localhost:12345/', files={"x": ("foo", "bar", "text/plain", {b'x': b'y'})})`, it does create a request and probably shouldn't be an error.

tommilligan: Just to be clear - is this something you'd like to be addressed in this PR? I would prefer to open that can of worms in a follow up if possible, as this PR is addressing an orthogonal issue.

Happy to open that PR after this one is closed.

Akuli: We should allow bytes and immutable mappings (in this PR), because it works at runtime. Note that even the immutable `Mapping` is invariant in the key type, so you will need `Mapping[str | Foo] | Mapping[bytes | Foo] | Mapping[str | bytes, Foo]`.

tommilligan: So after some investigation, I disagree that the headers type should be changed to include `bytes`. This is also why I'd like to avoid rolling the headers type change into this PR.

With the example request you posted above, I see an invalid body prepared for the request:

```python
from requests import Request

prepared = Request(
    "POST",
    "http://localhost:12345/",
    files={"x": ("foo", "bar", "text/plain", {b"x": b"y"})},
).prepare()
print(prepared.body.decode())
```
Output:
```
--ae71473b6c428c3d94bfbd5415f7bf1e
Content-Disposition: form-data; name="x"; filename="foo"
Content-Type: text/plain
b'x': b'y'

bar
--ae71473b6c428c3d94bfbd5415f7bf1e--
```

Note the 'header' of `b'x': b'y'`, which definitely doesn't look like the header a user would expect to be sending.

This is caused by formatting inside the `urllib3` library, which results in the python representation of a bytestring:
- https://github.com/psf/requests/blob/0f358eabe6f8e4c472a5030d9cd513965868bd6c/requests/models.py#L197
- https://github.com/urllib3/urllib3/blob/6de3330eb54f73a57c7860f75123bde8b043dbd2/src/urllib3/fields.py#L244

Akuli (approved):
Thanks!

tommilligan:
@isFakeAccount I can confirm these changes were released as of at least `types-requests 2.27.30`

```bash
pip install types-requests==2.27.30 -U 
```

isFakeAccount:
Thanks for the quick response @tommilligan

PR #8000, opened 2022-05-31
`builtins.sum`: Items in the iterable must support addition with `int` if no `start` value is given

### LINKED ISSUE #7574: `sum` stub too permissive

AlexWaygood:
Closes #7574 (cc. @jpy-git). I think this is just about the best we can do without introducing many false-positive errors.

### DIFF
--- a/stdlib/_typeshed/__init__.pyi
+++ b/stdlib/_typeshed/__init__.pyi
@@ -72,6 +72,9 @@ SupportsRichComparisonT = TypeVar("SupportsRichComparisonT", bound=SupportsRichC
 class SupportsAdd(Protocol[_T_contra, _T_co]):
     def __add__(self, __x: _T_contra) -> _T_co: ...
 
+class SupportsRAdd(Protocol[_T_contra, _T_co]):
+    def __radd__(self, __x: _T_contra) -> _T_co: ...
+
 class SupportsDivMod(Protocol[_T_contra, _T_co]):
     def __divmod__(self, __other: _T_contra) -> _T_co: ...
 
--- a/stdlib/_typeshed/__init__.pyi
+++ b/stdlib/_typeshed/__init__.pyi
@@ -69,11 +69,8 @@ SupportsRichComparisonT = TypeVar("SupportsRichComparisonT", bound=SupportsRichC
 
 # Dunder protocols
 
-class SupportsAdd(Protocol):
-    def __add__(self, __x: Any) -> Any: ...
-
-class SupportsAddWithInt(Protocol):
-    def __add__(self, __x: int) -> Any: ...
+class SupportsAdd(Protocol[_T_contra, _T_co]):
+    def __add__(self, __x: _T_contra) -> _T_co: ...
 
 class SupportsDivMod(Protocol[_T_contra, _T_co]):
     def __divmod__(self, __other: _T_contra) -> _T_co: ...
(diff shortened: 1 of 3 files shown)


stdlib/_typeshed/__init__.pyi
```diff
@@ -69,9 +69,12 @@ SupportsRichComparisonT = TypeVar("SupportsRichComparisonT", bound=SupportsRichC
 
 # Dunder protocols
 
-class SupportsAdd(Protocol):
+class SupportsAddition(Protocol):
```
hauntsaninja, line 72: (I think I prefer the old name since it mirrors the dunder method name)

AlexWaygood: I disliked the idea of `SupportsAddWithInt` ://

But fair enough, I'll revert the naming change :)

AlexWaygood:
Any thoughts from anyone on whether this is a good idea? :)

stdlib/builtins.pyi
```diff
@@ -1561,15 +1562,15 @@ else:
     def sum(__iterable: Iterable[bool], __start: int = ...) -> int: ...  # type: ignore[misc]
 
 @overload
-def sum(__iterable: Iterable[_SumT]) -> _SumT | Literal[0]: ...
+def sum(__iterable: Iterable[_AddableWithIntT]) -> _AddableWithIntT | Literal[0]: ...
```
srittau, line 1565: This doesn't seem right:

```python
In [2]: class Foo:
   ...:     def __add__(self, other):  # accepts int as "other"
   ...:         return Foo()
   ...: 

In [3]: sum([Foo(), Foo()])
<ipython-input-3-55dd5d052a18> in <module>
----> 1 sum([Foo(), Foo()])

TypeError: unsupported operand type(s) for +: 'int' and 'Foo'
```

AlexWaygood: That's a weird `TypeError` there though. Good spot. Maybe `__radd__` is involved somehow.

AlexWaygood: Indeed, `__add__` and `__radd__` are both required if no default is specified:

```python
>>> class Foo:
...     def __add__(self, other):
...         return Foo()
...
>>> sum([Foo(), Foo()])
TypeError: unsupported operand type(s) for +: 'int' and 'Foo'
>>> class Bar:
...     def __radd__(self, other):
...         return Bar()
...
>>> sum([Bar(), Bar()])
TypeError: unsupported operand type(s) for +: 'Bar' and 'Bar'
>>> class Baz:
...     def __add__(self, other):
...         return Baz()
...     def __radd__(self, other):
...         return Baz()
...
>>> sum([Baz(), Baz()])
<__main__.Baz object at 0x000001E71D90BD30>
```

srittau (approved):

JelleZijlstra:
This just found a bug where we were doing `sum()` on a type that defined `__add__` but not `__radd__`. Thanks!

PR #8002, opened 2022-06-01
markdown: remove `INSTALLED_EXTENSIONS`.

leven-cn:


AlexWaygood:
Did this just not exist at runtime?

PR #8010, opened 2022-06-01
Add type stubs for jcconv3 package

richardjharris:
Points of confusion:
 - input type (str | bytes) matches output type, but unsure how to
   express this
 - mode parameter of convert must be a valid jcconv.CONSTANT but
   for now I'm using integer literals


AlexWaygood (commented):
We still support Python 3.6, so you'll have to use `typing_extensions.Literal` instead of `typing.Literal`

stubs/jcconv3/jcconv3/jcconv3.pyi
```diff
@@ -0,0 +1,42 @@
+from typing import Literal
```
AlexWaygood, line 1:
```suggestion
```

richardjharris: d'oh, thanks

AlexWaygood:
It looks like this library probably meets our criteria for having stubs added to typeshed. At the same time, [the library doesn't appear to have been updated in over two years](https://github.com/LIQRGV/jcconv), and I'm not hugely enthusiastic about adding stubs for a library that's unmaintained.

richardjharris:
Thanks for the input. I just found out I can put my own typings in a `typings` directory for personal use. That's enough for my needs so happy to close this.

AlexWaygood:
> Thanks for the input. I just found out I can put my own typings in a `typings` directory for personal use. That's enough for my needs so happy to close this.

Thanks for understanding! Typeshed is a pretty huge repository, and adding any new stubs always adds a maintenance burden for us, so we do have to have some limits on what we accept :)

PR #8011, opened 2022-06-01
Add type stubs for romkan module

richardjharris:


AlexWaygood (commented):
We still support Python 3.6, so you'll need to use `typing.Pattern` and `typing.Match`

AlexWaygood (commented):
You're now importing `Match` and `Pattern` from `typing`, but the CI is still failing because you're not actually using them in the stub — you're still using `re.Match` and `re.Pattern` in the stub, which don't exist on 3.6.

AlexWaygood:
I'm not sure about adding these stubs. The library doesn't seem to have been updated since 2013, and [the package metadata says that the highest Python version it supports is Python 3.3](https://github.com/soimort/python-romkan/blob/c82dc16f1154794fab765cceb2c036d20251717d/romkan.json#L24-L31). Our [policy](https://github.com/python/typeshed/blob/master/CONTRIBUTING.md#third-party-library-stubs) states:

> We accept stubs for third-party packages into typeshed as long as:
> - the package is publicly available on the [Python Package Index](https://pypi.org/);
> - the package supports any Python version supported by typeshed; and
> - the package does not ship with its own stubs or type annotations.

Criteria (2) does not seem to be met here, since typeshed only supports Python 3.6+.

richardjharris:
Closing as per discussion at https://github.com/python/typeshed/pull/8010

PR #8013, opened 2022-06-02
Add stubs for "gdb"

pmderodat:
Hello,

Here are type stubs for the "gdb" package, the Python API to extend GDB (https://sourceware.org/gdb/onlinedocs/gdb/Python-API.html). This API is implemented using CPython’ C API, so I’ve tried my best using the documentation and live GDB session to check the types.

Thanks for having a look!


pmderodat:
Here is the last failure that the CI reports:
```
Testing gdb...

Failed to install
ERROR: Could not find a version that satisfies the requirement gdb==12.1.* (from versions: none)
ERROR: No matching distribution found for gdb==12.1.*
```

This one looks expected: the `gdb` module is available only when running Python scripts under GDB, there is nothing you can install with pip (no wheel, no actual package files, …).

What should I do to keep the PR checks happy? Thanks!

stubs/gdb/METADATA.toml
```diff
@@ -0,0 +1,5 @@
+version = "12.1.*"
+extra_description = """\
+  Type hints for GDB's \
+  [Python API](https://sourceware.org/gdb/onlinedocs/gdb/Python-API.html).\
+"""
```
srittau, line 5: You can add the following line to skip running stubtest:

```python
stubtest = false
```

pmderodat: I’ll try that, thanks!

pmderodat:
Thank you for your answer.

My use case is that I want to use mypy to type-check a Python codebase (https://github.com/adacore/gnat-gdb-scripts) that uses this API. My understanding so far that type stubs were a way for mypy to do its work without needing the actual code base installed (because there’s no such codebase… or at least not a Python one).

This addition would allow me to `pip install types-gdb` and then just run mypy on my codebase… and hopefully would allow anyone writing GDB scripts to do the same (for instance, why not [libstdc++’s helpers](https://gcc.gnu.org/git/?p=gcc.git;a=tree;f=libstdc%2B%2B-v3/python/libstdcxx;h=68c2f13a46c1f29a2fc4c546932784aa9b993c88;hb=HEAD)).

hauntsaninja:
Currently types-X refers to PyPI distribution X, which this would change. It looks like there is no project at https://pypi.org/project/gdb/ , so maybe this is a bit theoretical, but worth mentioning.

pmderodat:
Interesting… I would argue that publishing a “concurrent” gdb package on PyPI would look like a bad thing for the Python ecosystem independently, but that’s just my opinion. Do you think I should instead manually create and upload a package for these types stubs, for instance `gnu-gdb-python-types-stubs`?

srittau:
FWIW, while this is of course a special case, I'm fine with uploading this types package, even if no corresponding implementation package exists. We should document this clearly using the `extra_description` field in `METADATA.toml`, though.

pmderodat:
Understood, I’ve just added a note, please let me know if we need more.

srittau:
> Understood, I’ve just added a note, please let me know if we need more.

Sounds good to me, although I'd like to wait for the feedback of the other maintainers.

Akuli:
I think it's fine to have `types-gdb` on pypi without having `gdb` on pypi, as long as it's documented clearly. I think we should mention this in CONTRIBUTING.md though. It's not uncommon for typeshed maintainers to just `pip install` a package based on the name of the folder in `stubs/`, and I don't want maintainers to be tricked into installing a malicious package: anyone could upload a malicious pypi package named `pdb` and make a pull request to our `pdb` stubs.

AlexWaygood:
I don't have experience with `gdb`, but I'm also fine with adding these stubs even if there's not a pip-installable package. I'm a little concerned that these subs could easily become outdated without stubtest, but we already have a few other `types-*` packages that aren't stubtested, so it's not *that* big a deal.

srittau (approved):
I did not compare this to the implementation, but LGTM.

JelleZijlstra:
I agree with including this but I'd like to compare it against the implementation, give me a few days.

stubs/gdb/gdb/__init__.pyi
```diff
@@ -0,0 +1,695 @@
...
+class _PrettyPrinter(Protocol):
+    # TODO: The "children" and "display_hint" methods are optional for
+    # pretty-printers. Unfortunately, there is no such thing as an optional
+    # method in Mypy at the moment.
+    #
+    # def children(self) -> Iterator[tuple[str, _ValueOrNative]]: ...
+    # def display_hint(self) -> str | None: ...
+    def to_string(self) -> str | LazyString: ...
+
+_PrettyPrinterLookupFunction: TypeAlias = Callable[[Value], Optional[_PrettyPrinter]]
```
AlexWaygood, line 187:
```suggestion
_PrettyPrinterLookupFunction: TypeAlias = Callable[[Value], _PrettyPrinter | None]
```

stubs/gdb/gdb/__init__.pyi
```diff
@@ -0,0 +1,695 @@
...
+    def __call__(self, frame: Frame) -> UnwindInfo | None: ...
+
+# Xmethods: the API is defined in the "xmethod" module
+
+# Inferiors
+
+def inferiors() -> tuple[Inferior, ...]: ...
+def selected_inferior() -> Inferior: ...
+
+BufferType: TypeAlias = Union[memoryview, bytes]
```
AlexWaygood, line 245:
```suggestion
BufferType: TypeAlias = memoryview | bytes
```

AlexWaygood (changes requested):
Nit: please use PEP 604 syntax where possible

stubs/gdb/gdb/__init__.pyi
JelleZijlstra: Probably should be `_typeshed.ReadableBuffer` (I haven't checked the code though)

AlexWaygood: > Probably should be `_typeshed.ReadableBuffer` (I haven't checked the code though)

I also haven't checked the code, but yeah, sounds plausible :)

pmderodat: I’ve double-checked passing a `mmap.mmap` instance: it looks like it works, thanks!

stubs/gdb/gdb/__init__.pyi
```diff
@@ -0,0 +1,695 @@
...
+TYPE_CODE_NAMESPACE: int
+TYPE_CODE_DECFLOAT: int
+TYPE_CODE_INTERNAL_FUNCTION: int
+
+# Pretty Printing
+
+class _PrettyPrinter(Protocol):
+    # TODO: The "children" and "display_hint" methods are optional for
+    # pretty-printers. Unfortunately, there is no such thing as an optional
+    # method in Mypy at the moment.
```
JelleZijlstra, line 181:
```suggestion
    # method in the type system at the moment.
```
We don't just support mypy

stubs/gdb/gdb/__init__.pyi
```diff
@@ -0,0 +1,695 @@
...
+    def is_valid(self) -> bool: ...
+    def switch(self) -> None: ...
+    def is_stopped(self) -> bool: ...
+    def is_running(self) -> bool: ...
+    def is_exited(self) -> bool: ...
+    def handle(self) -> bytes: ...
+
+# Recordings
+
+def start_recording(method: str = ..., format: str = ...) -> Record: ...
```
JelleZijlstra, line 282:
```suggestion
def start_recording(__method: str = ..., __format: str = ...) -> Record: ...
```
These are positional-only (see py-record.c)

JelleZijlstra (commented):
Thanks! I looked at the top-level functions in the main module and noticed only a few small problems, mostly around positional-only arguments.

For reference the source code is here: https://github.com/bminor/binutils-gdb/blob/master/gdb/python/python.c

`gdb.format_address` is missing.

pmderodat:
Thanks a lot for your feedback.

I knew that most functions in GDB take positional-only arguments, and did not know how to describe them in stubs: I wasn’t aware of the `__` naming convention. I’ll do another pass on all the stubs to fix this.

Regarding `gdb.format_address`, it seems that it is a recent addition that isn’t yet part of an official release. It is fine to include it anyway? (I’ve set the last GDB release version in the metadata file to track what GDB version the stubs try to describe) If we go this way, I’ll also add stubs for `gdb.current_language` and `gdb.Frame.language`, which were also added recently.

JelleZijlstra:
> Thanks a lot for your feedback.
> 
> I knew that most functions in GDB take positional-only arguments, and did not know how to describe them in stubs: I wasn’t aware of the `__` naming convention. I’ll do another pass on all the stubs to fix this.

Thanks! For technical reasons we can't use PEP 570 pos-only arguments, so we're stuck with the double underscore hack.

> 
> Regarding `gdb.format_address`, it seems that it is a recent addition that isn’t yet part of an official release. It is fine to include it anyway? (I’ve set the last GDB release version in the metadata file to track what GDB version the stubs try to describe) If we go this way, I’ll also add stubs for `gdb.current_language` and `gdb.Frame.language`, which were also added recently.

Oh, good catch! We shouldn't add it yet if it hasn't been released.

pmderodat:
> Thanks! For technical reasons we can't use PEP 570 pos-only arguments, so we're stuck with the double underscore hack.

I wasn’t aware of that PEP, interesting read, thank you! I’m eager to be able to use it in stubs. :-)

> Oh, good catch! We shouldn't add it yet if it hasn't been released.

Understood.

I’ve tried my best to fix the positional-only arguments in the stubs double-checking with the C definitions, I hope it’s now fine.

pmderodat:
Thanks a lot for the review and the merge!

stubs/gdb/gdb/events.pyi
```diff
@@ -0,0 +1,131 @@
...
+    def connect(self, __object: Callable[[ExitedEvent], object]) -> None: ...
+    def disconnect(self, __object: Callable[[ExitedEvent], object]) -> None: ...
+
+exited: ExitedEventRegistry
+
+class StopEvent(ThreadEvent):
+    stop_signal: str
+
+class BreakpointEvent(StopEvent):
+    breakpoints = Sequence[gdb.Breakpoint]
```
AlexWaygood, line 30: This line was just flagged by a new check I'm working on for [flake8-pyi](https://github.com/PyCQA/flake8-pyi), the linter we use at typeshed. Did you mean for this to be a type alias, or did you mean to have `breakpoints: Sequence[gdb.Breakpoint]`?

stubs/gdb/gdb/printing.pyi
```diff
@@ -0,0 +1,31 @@
...
+    def __call__(self, val: gdb.Value) -> gdb._PrettyPrinter | None: ...
+
+class SubPrettyPrinter:
+
+    name: str
+    enabled: bool
+
+    def __init__(self, name: str) -> None: ...
+
+GenPrinterFunction = Callable[[gdb.Value], PrettyPrinter]
```
AlexWaygood, line 22: Same here -- did you mean for this to be a type alias, or did you mean to do `GenPrinterFunction: Callable[[gdb.Value], PrettyPrinter]`

JelleZijlstra: This name doesn't appear in the gdb source code, I think it should be a private type alias.

stubs/gdb/gdb/events.pyi
JelleZijlstra: It should be an attribute according to https://sourceware.org/gdb/onlinedocs/gdb/Events-In-Python.html (search for "BreakpointEvent.breakpoints"). For some reason I can't find it in the source code though.

pmderodat: Thank you for spotting this! I confirm it should be an attribute. Re-reading this some time after writing it, I now also see a typo on line 31. I’ve just opened a PR to fix this: #8144

stubs/gdb/gdb/printing.pyi
pmderodat: I definitely meant it as a private type alias, but forgot about it when writing the related part in `__init__.py:187`. I’ll just remove it, thanks! Fix also in #8144

PR #8018, opened 2022-06-04
Turn a selection of __contains__-methods into TypeGuards

harahu:
Just an idea I wanted to try out. I think this could be beneficial.

Was considering also doing this to Container, but am imagining the consequences to be bigger.


AlexWaygood (commented):
It doesn't look like mypy, at least, is able to use this type hint to perform type-narrowing with `if y in x` guards: https://mypy-play.net/?mypy=latest&python=3.10&flags=show-traceback&gist=a9137d5292f17ea41c39f74c6362b4f8. So I'm not sure that this is worth it :/

Even if this did work, I'm not sure I'd be in favour of having this merged. I feel like it would be quite disruptive to force everybody subclassing a builtin container to use `TypeGuard` as the return type for `__contains__`. I think I'd prefer to see type checkers special-case `if x in y` guards instead.

AlexWaygood:
> I think I'd prefer to see type checkers special-case `if x in y` guards instead.

Pyright in fact [has already implemented special-casing of this kind](https://github.com/microsoft/pyright/blob/main/docs/type-concepts.md#type-guards):

```python
# PYRIGHT:
from random import choice

z = [1, 2]
y = choice([object(), 2])
reveal_type(y)  # Type of "y" is "object | int"
if y in z:
    reveal_type(y)  # Type of "y" is "int"
```

So, I think we'll decline this one. But thanks for the PR -- it was a clever idea 🙂

PR #8022, opened 2022-06-04
`asyncio.sslproto`: More updates for py311

AlexWaygood:
https://github.com/python/cpython/commit/13c10bfb777483c7b02877aab029345a056b809c


stdlib/asyncio/sslproto.pyi
```diff
@@ -70,8 +89,14 @@ class _SSLProtocolTransport(transports._FlowControlMixin, transports.Transport):
     def write(self, data: bytes) -> None: ...
     def can_write_eof(self) -> Literal[False]: ...
     def abort(self) -> None: ...
+    if sys.version_info >= (3, 11):
```
JelleZijlstra, line 92:
```suggestion
    if sys.version_info >= (3, 11):
        def get_write_buffer_limits(self) -> tuple[int, int]:
```

stdlib/asyncio/sslproto.pyi
```diff
@@ -70,8 +89,14 @@ class _SSLProtocolTransport(transports._FlowControlMixin, transports.Transport):
...
+        def get_read_buffer_limits(self) -> tuple[int, int]: ...
+        def set_read_buffer_limits(self, high: int | None = ..., low: int | None = ...) -> None: ...
+        def get_read_buffer_size(self) -> int: ...
 
 class SSLProtocol(protocols.Protocol):
```
JelleZijlstra, line 97: This now inherits from BufferedProtocol

stdlib/asyncio/sslproto.pyi
AlexWaygood: I was confused about why this hadn't been flagged by stubtest -- looks like in the stub it's identical to the signature of `get_write_buffer_limits()` in the superclass, `asynico.transports._FlowControlMixin`. Happy to add the override here, though, since I agree it's a bit weird and counterintuitive to have the other three methods here but not this one.

JelleZijlstra (approved):

PR #8023, opened 2022-06-06
`tkinter`: remove TypeAlias of `_MenuIndex` for `Menu`.

leven-cn:


AlexWaygood (commented):
I'll let @Akuli decide on this, since he's the tkinter expert, but I'm -1 on this. It makes the code less repetitive, but only slightly, and, in my opinion, it also makes the code more verbose and less readable.

PR #8031, opened 2022-06-07
Fix various `__all__`-related errors and omissions

AlexWaygood:
These were all discovered using the stubtest patch I have over at https://github.com/python/mypy/pull/12214


AlexWaygood:
Ah, it looks like pytype doesn't have `ParamSpecArgs` or `ParamSpecKwargs` in its copy of `typing.pyi`.

hauntsaninja (approved):
Thanks for fixing these! I will get around to reviewing the stubtest PR (although you may want to get some version of it to run on third party stubs)

AlexWaygood:
> although you may want to get some version of it to run on third party stubs

Yah, I'll have a go at doing that tomorrow with the same method I employed here :)

(Namely, what I did here was to edit the stubtest workflow on my fork of typeshed so that it pointed towards the branch on my mypy fork with the stubtest patch. Then I manually ran the CI from my local fork.)

Issue #8027, opened 2022-06-07
How should sqlite3.Row be hinted


altendky:
`sqlite3.Row` was made generic in [python/typeshed#7641](https://github.com/python/typeshed/pull/7641).  With that change I'm not clear how to hint an `sqlite3.Row`. It seems that it was made generic against a single type, but many tables will not use the same type for all columns. Also, i have setup to use the dict-like rows which would require typed-dict-like hinting I guess.  I didn't see any relevant discussion in the PR.

altendky:
Is there a case I'm overlooking where being generic like it is now is useful?  Outside of the probably rare corner case of all columns in a table having the same type.  At which point there still isn't a hint as to how many columns there are.

JelleZijlstra:
Yes, not sure why I made that change; I agree it's likely not useful. I'd be OK with making the class non-Generic again.

altendky:
Is there a path forward, maybe after some other expected changes/features in mypy even, towards a more complete solution?  If so then the thrash of removing it may not be worthwhile.  It has been present in a few recent mypy releases already.  If not, then maybe removal would be good.

JelleZijlstra:
There's no currently planned feature that would help here. We'd need something like a TypedDict that is based on an arbitrary class instead of a dict, but that's not a simple feature to add.

PR #8036, opened 2022-06-08
Remove Generic from sqlite3.Row

altendky:
Introduced in https://github.com/python/typeshed/pull/7641.  Removal discussed at https://github.com/python/typeshed/issues/8027.


altendky:
I am not entirely certain this is the right path.  Probably mostly because I haven't tried to get into the overall context etc of mypy development.  But given the comments at https://github.com/python/typeshed/issues/8027 it seems unlikely to be useful for `Row` to be generic now or in the foreseeable future.  So, here's an offering to be considered.

stdlib/sqlite3/dbapi2.pyi
```diff
@@ -420,15 +419,15 @@ class PrepareProtocol:
 
 class ProgrammingError(DatabaseError): ...
 
-class Row(Generic[_T_co]):
-    def __init__(self, __cursor: Cursor, __data: tuple[_T_co, ...]) -> None: ...
+class Row:
+    def __init__(self, __cursor: Cursor, __data: tuple[Any]) -> None: ...
```
JelleZijlstra, line 423:
```suggestion
    def __init__(self, __cursor: Cursor, __data: tuple[Any, ...]) -> None: ...
```
A tuple containing any number of objects, not a tuple containing a single object.

altendky: Oops...  thanks.

JelleZijlstra (approved):

Issue #8012, opened 2022-06-02
Remove stubs for `characteristic`?


AlexWaygood:
The `characteristic` library hasn't had any changes since October 2020. It also has a big warning on [its GitHub homepage](https://github.com/hynek/characteristic) saying:

> No Maintenance Intended
> Warning
> Characteristic is unmaintained. Please have a look at its successor [attrs](https://attrs.readthedocs.io/).

It doesn't seem like we've had any issues or PRs relating to these stubs for a long time. Should we remove them?

(Side note: I'm not sure what our policy is for removing stubs for unmaintained libraries that don't have type hints -- we don't have anything in `CONTRIBUTING.md`.)

JelleZijlstra:
I don't think we should necessarily remove stubs for unmaintained packages, since people may still use them and providing types could provide a valuable service. Even if the package only declares support for Python 3.3-3.4, it may still work fine on newer versions.

In this case though, I think it's fine to remove, since there's been little interest in the stubs since they were contributed in #771, and the package has been unmaintained for so long.

srittau:
>Third-party packages are generally removed from typeshed when one of the
following criteria is met:
>
> [...]
> * the package does not support any of the Python versions supported by
  typeshed.

That said, I have no objections with removing it. The types package remains, after all, we just can't update it easily. Cc @hynek as author.

We could add another field, similar to `obsolete_since` to `METADATA.toml` to add a comment to a final upload.

AlexWaygood:
> I don't think we should necessarily remove stubs for unmaintained packages

I wasn't suggesting going on a blitz to remove all unmaintained packages from typeshed — I was just looking at the source code for this one for another reason, and was struck by the warning on the GitHub homepage :)

> In this case though, I think it's fine to remove, since there's been little interest in the stubs since they were contributed in https://github.com/python/typeshed/pull/771, and the package has been unmaintained for so long.

> The types package remains, after all, we just can't update it easily.

+1 to both of these thoughts.

hynek:
For me the question is mostly about how useful those stubs are at all? Since it's attrs-like, you rather need a plugin for it to be useful with mypy?

The last release is 8 years ago, I had no idea that stubs actually exist, and it has only 30k downloads per month (which is suspect are only mirrors). So I have no hard feelings if y'all dropped them.

PR #8038, opened 2022-06-08
Mark characteristic as no longer updated

srittau:
Part of #8012, depends on #8037


AlexWaygood (approved):
LGTM (once #8037 is merged and the CI is green)

PR #8041, opened 2022-06-08
`dateparser`: Remove redundant import of `_typeshed.Self`

AlexWaygood:
It doesn't exist at runtime:

```python
>>> import dateparser.date
>>> dateparser.date.Self
AttributeError: module 'dateparser.date' has no attribute 'Self'
```

...and it's not used in the stub either.

I'm not sure why stubtest didn't pick it up. My stubtest patch over at https://github.com/python/mypy/pull/12214 _does_ pick it up (and I'm not sure why that is, either... but it's a true positive!)

### DIFF
--- a/stubs/dateparser/dateparser/date.pyi
+++ b/stubs/dateparser/dateparser/date.pyi
@@ -1,5 +1,4 @@
 import collections
-from _typeshed import Self as Self
 from collections.abc import Callable, Iterable, Iterator
 from datetime import datetime
 from typing import ClassVar, Pattern, overload


hauntsaninja (approved):
Thanks, curious why none of our CI picks it up

hauntsaninja:
@AlexWaygood I got around to looking into this. Turns out I fixed it in https://github.com/python/mypy/pull/12851

AlexWaygood:
> @AlexWaygood I got around to looking into this. Turns out I fixed it in [python/mypy#12851](https://github.com/python/mypy/pull/12851)

Heh, that makes sense!

PR #8046, opened 2022-06-09
rm -rf stubs/filelock

hauntsaninja:
See #8043

Please confirm that a new types-filelock has been released before
merging


AlexWaygood:
Depends on

- https://github.com/python/mypy/pull/12960

AlexWaygood:
> There isn't actually any dependency. mypy_primer just needs to be updated to install filelock when setting up mypy (but it also doesn't really matter if we don't update mypy_primer, since it'll happily diff things out).
> 
> The mypy PR is just a nit, since self check would fail if someone installed filelock 3.2

Ahh sorry I misunderstood, and didn't want to inadvertently break mypy! Thanks, great, let's merge then.

hauntsaninja:
Updated primer in https://github.com/hauntsaninja/mypy_primer/commit/d91de1f533963a41d362531147c2ff407c286433

AlexWaygood (approved):

AlexWaygood:
> Well we do need to wait for a new types-filelock :-)

"let's merge in a little bit"

srittau:
https://pypi.org/project/types-filelock/3.2.7/

PR #8049, opened 2022-06-09
Suggest updating mypy on stub removals

hauntsaninja:


hauntsaninja:
The list is legacy, to avoid bad backward compatibility issues from modular typeshed, e.g. we don't want new additions to end up there. I guess it's still possible to have this update only removals from typeshed, but it's a pretty rare thing. Feel free to close, merge, or update mypy's vendoring script :-)

srittau (approved):

PR #8053, opened 2022-06-09
Add stub for emoji.is_emoji

vdonato:
Guessing from the documentation, the `is_emoji` function was added to the [emoji](https://pypi.org/project/emoji/)
package around version 1.6, but the type stubs weren't updated to reflect this.


AlexWaygood (approved):
Thanks!

PR #8056, opened 2022-06-11
`xml.etree.ElementTree.Element`: remove inheritance from `MutableSequence`

### LINKED ISSUE #8030: `xml.etree.ElementTree.Element` is not a subclass of `MutableSequence`
According to [Python docs](https://docs.python.org/3/library/collections.abc.html) any `MutableSequence` must define next methods:
```
# Must be defined by user
__getitem__
__setitem__
__delitem__
__len__
insert
# Defined in Sequence
__contains__
__iter__
__reversed__
index
count
# Defined in MutableSequence
append
reverse
extend
pop
remove
__iadd__
```
Methods defined in `Sequence` and `MutableSequence` are available in user's class only if that class inherits from them (or user defines these methods in the class).

`typeshed` [claims](https://github.com/python/typeshed/blob/master/stdlib/xml/etree/ElementTree.pyi#L135) that `xml.etree.ElementTree.Element` is a subclass of `MutableSequence`, i. e., `Element` defines all methods above. Unfortunately this is wrong. `Element` does not inherit from `MutableSequence` and it [defines](https://docs.python.org/3.10/library/xml.etree.elementtree.html#element-objects) only next methods from `MutableSequence`:
```
__len__
__getitem__
__setitem__
__delitem__
append
extend
insert
remove
```

In my opinion inheriting from `MutableSequence` must be removed to avoid errors like method `pop` being a part of `Element`.

AlexWaygood:
Fixes #8030. I'm not convinced that this is a good idea, but I'd like to see what the CI says.


AlexWaygood:
I was expecting mypy_primer to report a larger fallout, so I guess I'm neutral on this. More thoughts here: https://github.com/python/typeshed/issues/8030#issuecomment-1152949592

JelleZijlstra (approved):

stdlib/xml/etree/ElementTree.pyi
```diff
@@ -115,8 +116,10 @@ class Element(MutableSequence[Element]):
     @overload
     def __getitem__(self, __i: SupportsIndex) -> Element: ...
     @overload
-    def __getitem__(self, __s: slice) -> MutableSequence[Element]: ...
+    def __getitem__(self, __s: slice) -> list[Element]: ...
     def __len__(self) -> int: ...
+    # Doesn't actually exist at runtime, but instance of the class are indeed iterable due to __getitem__.
+    def __iter__(self) -> Iterator[Element]: ...
```
srittau, line 122: This is really something that type checkers should special case. :/

srittau (approved):

PR #8057, opened 2022-06-11
`inspect`, `asyncio`: Use more `TypeGuard`s

### LINKED ISSUE #8009: Should `asyncio.iscoroutinefunction` return some kind of `TypeGuard`?
I have the following use-case of a class that takes a regular function or a coroutine as an attribute, and then provides access to it through two properties. Basically the code looks like this:

```python
"""example.py"""

from asyncio import iscoroutinefunction
from typing import TypeVar, Generic, Callable, Awaitable, Any

T = TypeVar('T')

class Response(Generic[T]):
    ...

class Request(Generic[T]):
    ...

class Foo:
    def __init__(self, callback: Callable[[Request[T]], Response[T] | Awaitable[Response[T]]]):
        self._callback = callback

    @property
    def func(self) -> Callable[[Request[T]], Response[T]]:
        if iscoroutinefunction(self._callback):
            raise AttributeError()

        return self._callback

    @property
    def coro(self) -> Callable[[Request[T]], Awaitable[Response[T]]:
        if not iscoroutinefunction(self._callback):
            raise AttributeError()

        return self._callback

async def bar(req: Request[T]) -> Response[T]:
    ...

foo = Foo(bar)

```

Now, I expected mypy to give no errors due to type narrowing from `iscoroutinefunction, but instead I get
```
example.py:21: error: Incompatible return value type (got "Callable[[Request[T]], Union[Response[T], Awaitable[Response[T]]]]", expected "Callable[[Request[T]], Response[T]]")
example.py:28: error: Incompatible return value type (got "Callable[[Request[T]], Union[Response[T], Awaitable[Response[T]]]]", expected "Callable[[Request[T]], Awaitable[Response[T]]]")
```

After finding this, I started searching through the issues in both mypy and typeshed but couldn't find anything related to `iscoroutinefunction`. I then looked into typeshed and saw that the return type of `iscoroutinefunction` is just `bool`, and not a `TypeGuard`, despite the similar `iscoroutine` having one. This makes me think that `iscoroutinefunction` is annotated like that on purpose.

So my question is, should the return type of `asyncio.iscoroutinefunction` be a `TypeGuard` of some kind?

I'm new to the more advanced features of Python typing, and my naive implementation would be something like
```python
from typing import TypeVar, TypeGuard

T = TypeVar('T')

def iscoroutinefunction(func: T) -> TypeGuard[T]: ...
```

If you feel like this should be changed, I'd be happy to submit a PR once I know what the type should be.

AlexWaygood:
Also, use a `TypeAlias` for the return type of `inspect.isroutine()`, instead of repeating the whole function definition across both of the branches.

Fixes #8009.


AlexWaygood:
Interesting. @erictraut, any idea why the test cases I've added in this PR are failing pyright? Mypy seems to have no trouble with them.

AlexWaygood:
**Mypy primer analysis**

---

Most of these are just noise: mypy is just emitting slightly different error messages for things that already failed to type check.

```diff
+ discord/ext/tasks/__init__.py:622: error: Incompatible return value type (got "Callable[..., CoroutineType[Any, Any, Any]]", expected "ET")
```

This one looks like a true positive to me. Source code here: https://github.com/Rapptz/discord.py/blob/334ef1d7facce9dbfba2a6924bf57fc59bc827b5/discord/ext/tasks/__init__.py#L598-L622

> ```diff
> aiohttp (https://github.com/aio-libs/aiohttp)
> + aiohttp/helpers.py:126: error: Incompatible redefinition (redefinition with type "Callable[[Any], bool]", original type overloaded function)  [misc]
> 
> pyinstrument (https://github.com/joerick/pyinstrument)
> + pyinstrument/vendor/decorator.py:66: error: Incompatible redefinition (redefinition with type "Callable[[Any], Any]", original type overloaded function)
> ```

These are both a little unfortunate, but can't really be helped. I think they will simply have to add `type: ignore`s to their code if this PR is merged.

- `aiohttp` source code: https://github.com/aio-libs/aiohttp/blob/3a41da05bcb1edde9f6f878d9092e108e70dff2a/aiohttp/helpers.py#L122-L129
- https://github.com/joerick/pyinstrument/blob/b833b3f4a6b039354ed8bb747371d5f0d539010c/pyinstrument/vendor/decorator.py#L62-L67

erictraut:
@AlexWaygood, pyright doesn't currently support `TypeGuard` on overloaded functions. Evaluating overloads is very expensive in the type narrowing code path because all argument types need to be evaluated first, so it creates a bunch of dependencies between expressions. For that reason, I didn't initially add support for this case. I can think of ways to limit the performance impact, but it will take some work. I've filed [this feature enhancement request](https://github.com/microsoft/pyright/issues/3572) to track this.

AlexWaygood:
> I've filed [this feature enhancement request](https://github.com/microsoft/pyright/issues/3572) to track this.

Thanks @erictraut! 

> I can think of ways to limit the performance impact, but it will take some work. 

I'm in no particular rush with this PR :)

> Evaluating overloads is very expensive in the type narrowing code path

If the performance impact is too large for pyright to tolerate, I suppose I could simply exclude the test cases from this PR -- these are the only bits that fail pyright. It seems that since pyright doesn't yet support combining `overload`s with `TypeGuard`s, it just infers the return type of these overloaded functions as `bool` -- and that's what they are at the moment, anyway, so there wouldn't be a regression in pyright's understanding of these functions.

I think the number of false-positive errors surfaced by mypy-primer makes it unfeasible to consider a version of this PR without the overloads, however.

(Yet another option is obviously to close this PR and not bother too much about it.)

erictraut:
This is included in pyright 1.1.255, which I just published.

JelleZijlstra (approved):

PR #8059, opened 2022-06-12
add undocumented attributes for threading.Timer

hoefling:
Signed-off-by: Oleg Hoefling <oleg.hoefling@gmail.com>

### DIFF
--- a/stdlib/threading.pyi
+++ b/stdlib/threading.pyi
@@ -215,6 +215,12 @@ if sys.version_info >= (3, 8):
     ExceptHookArgs = _ExceptHookArgs
 
 class Timer(Thread):
+    args: Iterable[Any] | None  # undocumented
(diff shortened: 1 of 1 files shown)


stdlib/threading.pyi
```diff
@@ -215,6 +215,12 @@ if sys.version_info >= (3, 8):
     ExceptHookArgs = _ExceptHookArgs
 
 class Timer(Thread):
+    args: Iterable[Any] | None  # undocumented
```
Akuli, line 218: `args` and `kwargs` can't be `None`. If you pass in `None` to `__init__`, they get replaced by an empty list or dict.

hoefling: @Akuli oops, my bad - corrected.

Akuli (approved):

PR #8061, opened 2022-06-12
Improve `csv.DictReader`

### LINKED ISSUE #7953: csv.DictReader marked as producing Dict[str, str] but can have a None key
Here is an example to illustrate:
```python
import csv
from io import StringIO

csvtext = "foo,bar\n1,2\n3,4,5"

csvreader = csv.DictReader(StringIO(csvtext), delimiter=',')

#def checkrow(row: dict[str | None, str]) -> None:
def checkrow(row: dict[str, str]) -> None:
	if None in row:
		print(f"None found as key with value {row[None]!r}, someone broke a promise!")

for row in csvreader:
	print(row)
	checkrow(row)
```

Type checkers complain: `error: Invalid index type "None" for "Dict[str, str]"; expected type "str"`

Signature visible here: https://github.com/python/typeshed/blob/master/stdlib/csv.pyi#L107

I'm not certain what could/should be done, but by construction `if None in foo: something(row[None])" is assuredly fine.
And if it were unreachable code, then it would never execute. But it does execute:
```
None found as key with value ['5'], someone broke a promise!
```

Trying the commented signature doesn't make typechecking happy either:
```
error: Argument 1 to "checkrow" has incompatible type "Dict[str, str]"; expected "Dict[Optional[str], str]"
```

I suppose that the type in the typeshed isn't quite a match with the library currently.

AlexWaygood:
Fixes #7953 (fixes the stuff typeshed can fix, anyway).

### DIFF
--- a/stdlib/csv.pyi
+++ b/stdlib/csv.pyi
@@ -79,7 +79,7 @@ class unix_dialect(Dialect):
     lineterminator: str
     quoting: _QuotingType
 
-class DictReader(Generic[_T], Iterator[_DictReadMapping[_T | None, str | None]]):
+class DictReader(Generic[_T], Iterator[_DictReadMapping[_T | Any, str | Any]]):
     fieldnames: Sequence[_T] | None
     restkey: str | None
     restval: str | None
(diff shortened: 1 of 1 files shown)


Akuli (approved):
Reflecting this more accurately would be really quite difficult:

```
>>> next(csv.DictReader(io.StringIO('x,y'), fieldnames=[1,2,3]))
{1: 'x', 2: 'y', 3: None}
>>> next(csv.DictReader(io.StringIO('x,y,a,b'), fieldnames=[1,2,3]))
{1: 'x', 2: 'y', 3: 'a', None: ['b']}
```

AlexWaygood:
> Reflecting this more accurately would be really quite difficult:
> 
> ```
> >>> next(csv.DictReader(io.StringIO('x,y'), fieldnames=[1,2,3]))
> {1: 'x', 2: 'y', 3: None}
> >>> next(csv.DictReader(io.StringIO('x,y,a,b'), fieldnames=[1,2,3]))
> {1: 'x', 2: 'y', 3: 'a', None: ['b']}
> ```

Yeah, I had a go locally, and quickly gave up. More trouble than it's worth imo.

PR #8063, opened 2022-06-12
Use `_typeshed.IdentityFunction` more consistently

AlexWaygood:


stubs/mypy-extensions/mypy_extensions.pyi
```diff
@@ -1,5 +1,5 @@
 import abc
-from _typeshed import Self
+from _typeshed import IdentityFunction, Self
 from collections.abc import Callable, ItemsView, KeysView, Mapping, ValuesView
```
AlexWaygood, line 3:
```suggestion
from collections.abc import ItemsView, KeysView, Mapping, ValuesView
```

JelleZijlstra (approved):

PR #8068, opened 2022-06-13
Add pre-commit config for flake8

### LINKED ISSUE #7991: Make it easier to run flake8 for contributors
For an external contributor who drops by, it's not immediately obvious what the linting/testing expectations are. Testing is harder and more complicated, but I think linting could be smoothed out. For now, just looking at the flake8 usage: two easier options than the current flow.

1. Add `flake8` to pre-commit config
2. Setup a `tox.ini` which allows for `tox -e lint` or similar to run linting

There are, of course, other options too, but these two are my preferred ones. The goal is to make installing flake8, flake8-pyi, and any other flake8 plugins a managed process via some tool.

Would either of these additions be welcome? If `flake8` goes into pre-commit, then pre-commit.ci takes care of running it and perhaps it can be removed from GitHub actions.

sirosen:
This makes it easier to lint your own work with `pre-commit`, but doesn't change the CI config by adding this to the pre-commit.ci skip list.

There's no way to automatically reflect the additional dependencies from the requirements file, so the flake8 plugin list needs to be a verbatim copy.

resolves #7991

---

Two asides:
- Sorry this took me so long to put in. I meant to do it right away and got distracted.
- If there's any discomfort about copying the requirements data, let's just nix this and close #7991 as a wontfix (at least for now). I think it's worth keeping the copy, but I also understand not wanting duplicate configs.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -10,7 +10,7 @@ repos:
       - id: isort
         name: isort (python)
   - repo: https://github.com/pycqa/flake8
-    rev: 4.0.1
+    rev: 4.0.1  # must match requirements-tests.txt
     hooks:
       - id: flake8
         additional_dependencies:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -14,6 +14,7 @@ repos:
     hooks:
       - id: flake8
         additional_dependencies:
+          # must match requirements-tests.txt
           - 'flake8-bugbear==21.11.29'
           - 'flake8-pyi==22.5.1'
 
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -9,11 +9,18 @@ repos:
     hooks:
       - id: isort
         name: isort (python)
+  - repo: https://github.com/pycqa/flake8
+    rev: 4.0.1
+    hooks:
+      - id: flake8
+        additional_dependencies:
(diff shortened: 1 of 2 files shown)


.pre-commit-config.yaml
```diff
@@ -9,11 +9,18 @@ repos:
     hooks:
       - id: isort
         name: isort (python)
+  - repo: https://github.com/pycqa/flake8
+    rev: 4.0.1
+    hooks:
+      - id: flake8
+        additional_dependencies:
+          - 'flake8-bugbear==21.11.29'
+          - 'flake8-pyi==22.5.1'
```
srittau, line 18: Could you add a comment here and in the requirements file that they should be kept in sync?

sirosen: I added these in the same style as the existing comments in `requirements-tests.txt`. We can tweak the text if it's too pithy.

.pre-commit-config.yaml
```diff
@@ -9,11 +9,19 @@ repos:
     hooks:
       - id: isort
         name: isort (python)
+  - repo: https://github.com/pycqa/flake8
+    rev: 4.0.1
```
AlexWaygood, line 13: Let's add a comment to this line as well

```suggestion
    rev: 4.0.1  # must match requirements-tests.txt
```

AlexWaygood (approved):
Thanks @sirosen! 🎉

Issue #7524, opened 2022-03-21
`subprocess.{run, call, check_call, check_output}` do not have arguments new in 3.9+


septatrix:
The subprocess methods do not have `group`, `extra_groups`, `user`, and `umask` in their signature.

AlexWaygood:
#8071 adds the new-in-3.9 parameters to `subprocess.run`, `suprocess.call`, `subprocess.call` and `subprocess.check_output`. However, there are new parameters in 3.10 and 3.11 that should be added to these functions in future PRs:

- 3.10 adds the `pipesize` parameter.
- 3.11 adds the `process_group` parameter.

PR #8070, opened 2022-06-13
`subprocess.(check_)call`: add `text` argument new in 3.7

### LINKED ISSUE #7777: `subprocess.call()` is missing `text` argument (in Python >=3.7)
The type annotation for `subprocess.call()` is missing the `text` argument (introduced in Python 3.7).

The type annotation for `call()` is accompanied by a comment that says `# Same args as Popen.__init__` but is not Python-version-specific and does not include the `text` argument:
https://github.com/python/typeshed/blob/b1ccd03d51f5efe5f600e8ede1810a253d35e1cd/stdlib/subprocess.pyi#L446-L467

The annotation for `Popen.__init__`, nay `Popen.__new__`, on the other hand, is in a Py >=3.7 branch to include `text`:
https://github.com/python/typeshed/blob/b1ccd03d51f5efe5f600e8ede1810a253d35e1cd/stdlib/subprocess.pyi#L807-L833

I guess `call()` needs a similar 3.7+ branch?

AlexWaygood:
Fixes #7777


hauntsaninja (commented):
Huh, I guess this was missed because passing `text` to check_call is a little unusual given there's check_output.

Lgtm, except that it's missing the version gate to 3.7

hauntsaninja (commented):
Could also consider adding the new 3.11 Popen args everywhere

AlexWaygood:
> Could also consider adding the new 3.11 Popen args everywhere

There's still some from 3.9 that haven't yet percolated through the stub: #7524. But I'd prefer to leave those (and the new 3.11) arguments) to future PRs.

hauntsaninja:
Oh missed that comment. I'm fine with whatever!

AlexWaygood:
> Oh missed that comment. I'm fine with whatever!

Well I've added the version guard now, so let's go with an accurate stub I guess 😆

PR #8071, opened 2022-06-14
Add new-in-3.9 parameters to `subprocess.{run, call, check_call, check_output}`

AlexWaygood:
Wow is this module a nightmare to type.

Refs #7524.


JelleZijlstra:
Would you mind adding comments like "3.9 adds parameter foo"? That should make these a bit easier to understand.

PR #8074, opened 2022-06-14
Fix miscellaneous invalid `TypeVar` usages

AlexWaygood:
Refs #7928


stubs/flake8-plugin-utils/flake8_plugin_utils/plugin.pyi
```diff
@@ -18,7 +18,10 @@ class Error:
 
 class Visitor(ast.NodeVisitor, Generic[TConfig]):
     errors: list[Error]
-    def __init__(self, config: TConfig | None = ...) -> None: ...
+    @overload
+    def __init__(self, config: TConfig) -> None: ...
+    @overload
+    def __init__(self, config: None = ...) -> None: ...
```
JelleZijlstra, line 24: These overloads should be swapped because otherwise `None` will be bound to  `TConfig`.

JelleZijlstra (commented):
Thanks, just noticed one thing.

stubs/flake8-plugin-utils/flake8_plugin_utils/plugin.pyi
AlexWaygood: Ahh, good spot!

AlexWaygood, line 24:
```suggestion
    @overload
    def __init__(self, config: None = ...) -> None: ...
    @overload
    def __init__(self, config: TConfig) -> None: ...
```

JelleZijlstra (approved):

PR #8087, opened 2022-06-16
Fix `TypeVar`s in `beautifulsoup` and `SQLAlchemy`

AlexWaygood:
Refs #7928


stubs/beautifulsoup4/bs4/element.pyi
```diff
@@ -351,4 +351,7 @@ class SoupStrainer:
 
 class ResultSet(list[_PageElementT], Generic[_PageElementT]):
     source: SoupStrainer
-    def __init__(self, source: SoupStrainer, result: Iterable[_PageElementT] = ...) -> None: ...
+    @overload
+    def __init__(self, source: SoupStrainer) -> None: ...
```
srittau, line 355: Sorry, I haven't following this so far, but shouldn't this annotate `self` to bind `_PageElementT`?

AlexWaygood: This is basically the same approach that we've had for a while for some generic collections in `builtins` and `collections`, e.g. this has been the first overload for `builtins.list` for two years now:

https://github.com/python/typeshed/blob/43f923b6c2fa7eb7b1f35890b46ea169beea25f3/stdlib/builtins.pyi#L958-L960

If the constructor-method overload picked is one where there are no `TypeVar`s present at all, despite the class being generic, the type checker will require the user to provide an explicit type annotation, the same as happens if you try to initialise an empty list 🙂

https://mypy-play.net/?mypy=latest&python=3.10&gist=dcbf1006ff7606d02d5670171bc6cae1

srittau (approved):

Issue #6905, opened 2022-01-12
Inconsistencies between `__init__` and `__new__`


erictraut:
I just implemented a new check in pyright called `reportInconsistentConstructor` that detects inconsistencies between the `__init__` and `__new__` methods. It's disabled by default but can be enabled in the configuration. It's in pyright 1.1.207, which I just published.

The check is pretty conservative, but it does identify three places in typeshed that look like legitimate problems.

1. The class `WeakMethod` in `weakref.pyi` defines a `__new__` that is incompatible with the `__init__` declared in its parent class `ReferenceType`.
2. The class `_TokenType` in `pygments/token.pyi` defines an `__init__` that is incompatible with the `__new__` in `tuple[str]`.
3. The class `CascadeOptions` in `SQLAlchemy/sqlalchemy/orm/util.pyi` defines a `__new__` that is incompatible with the `__init__` in its parent class `frozenset[Any]`.

If/when these are addressed, we might want to update pyright to 1.1.207 and enable the new diagnostic check to prevent future inconsistencies from being introduced.

srittau:
Sounds like a good plan.

AlexWaygood:
Relatedly (but slightly tangentially), can anybody explain to me why `classmethod.__new__` and `classmethod.__init__` have different signatures in `builtins`? I've never really understood why, and there's no comment clarifying it:

https://github.com/python/typeshed/blob/5d206358907c419802cd9afcfcc83cce4f01bebb/stdlib/builtins.pyi#L117-L118

(Same issue for `staticmethod`.)

PR #8088, opened 2022-06-16
`protobuf`: Delete some `__new__` and `__init__` methods

AlexWaygood:
Refs #6905.

Pyright points out that two classes in the stubs for `protobuf` have `__init__` and `__new__` methods that contradict. However, neither of these classes can be directly instantiated anyway:

```python
>>> from google.protobuf.descriptor import Descriptor
>>> Descriptor()
  File "C:\Users\alexw\coding\protobuf\lib\site-packages\google\protobuf\descriptor.py", line 313, in __new__
    _message.Message._CheckCalledFromGeneratedFile()
TypeError: Descriptors cannot not be created directly.
If this call came from a _pb2.py file, your generated code is out of date and must be regenerated with protoc >= 3.19.0.
If you cannot immediately regenerate your protos, some other possible workarounds are:
 1. Downgrade the protobuf package to 3.20.x or lower.
 2. Set PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION=python (but this will use pure-Python parsing and will be much slower).

More information: https://developers.google.com/protocol-buffers/docs/news/2022-05-06#python-updates
```

As such, I propose that we simply delete the `__init__` and `__new__` methods for these two classes.

Cc. @nipunn1313

### DIFF
--- a/stubs/protobuf/@tests/stubtest_allowlist.txt
+++ b/stubs/protobuf/@tests/stubtest_allowlist.txt
@@ -23,3 +23,12 @@ google.protobuf.text_format.MessageToBytes
 # Stubbed as static method, but actually exists as a property that's
 # a function. Typeshed's typing is more useful
 google.protobuf.service.Service.GetDescriptor
+
+# These are deliberately omitted in the stub
+# due to the fact that these classes have conflicting signatures for __init__/__new__.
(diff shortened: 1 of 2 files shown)


srittau (approved):

nipunn1313:
sg

Issue #6301, opened 2021-11-15
PEP 655 (Required TypedDict items) tracker


srittau:
This issue is supposed to track when we can start to use `Required`/`NotRequired` ([PEP 655](https://www.python.org/dev/peps/pep-0655/)) in typeshed. Support is needed in released versions of:

* [x] PEP accepted
* [X] mypy (python/mypy#10370)
* [ ] pytype
* [x] pyright (microsoft/pyright#1554)
* [x] pyre
* [x] PyCharm ([PY-53611](https://youtrack.jetbrains.com/issue/PY-53611/Support-PEP-655-typingRequiredtypingNotRequired-for-TypedDicts))

JelleZijlstra:
I feel like we should also wait for the PEP to be accepted. typing-extensions can include experimental features that may be rejected, but I feel like typeshed should not.

srittau:
This is now fully supported in mypy.

JelleZijlstra:
I think it supports only one of Required and NotRequired? That's what the initial PR said but I haven't verified with the release.

srittau:
Quick tests confirms that it works with both:

```python
from typing_extensions import TypedDict, Required, NotRequired

class NonTotal(TypedDict, total=False):
    foo: int
    bar: Required[str]
    
nt1: NonTotal = {"foo": 123, "bar": ""}
nt2: NonTotal = {"foo": "", "bar": 123}  # incompatible types
nt3: NonTotal = {"foo": 123}  # missing key "bar"
nt4: NonTotal = {"bar": ""}

class Total(TypedDict, total=True):
    foo: int
    bar: NotRequired[str]
    
t1: Total = {"foo": 123, "bar": ""}
t2: Total = {"foo": "", "bar": 123}  # incompatible types
t3: Total = {"bar": ""}  # missing key "foo"
t4: Total = {"foo": 123}  
```

gregoiredx:
Hello @srittau, just to let you know you can check the "PEP accepted" checkbox, see https://groups.google.com/g/dev-python/c/odKM_OpDES0 and https://peps.python.org/pep-0655/

srittau:
I've used `NotRequired` in #8089 and our tests passed. So it seems that pytype at least accepts `NotRequired`, even if it doesn't fully support it, which would be enough for me. Cc @rchen152 for feedback.

Also cc @pradeep90 and @shannonzhu: Does pyre support `Required`/`NotRequired` for `TypedDict`?

[...]

PR #8089, opened 2022-06-16
Add types for zxcvbn

srittau:


stubs/zxcvbn/zxcvbn/__init__.pyi
```diff
@@ -0,0 +1,21 @@
...
+from .time_estimates import _CrackTimesDisplay, _CrackTimeSeconds
+
+class _Result(TypedDict):
+    password: str
+    guesses: Decimal
+    guesses_log10: float
+    sequence: list[_Match]
+    calc_time: NotRequired[datetime.timedelta]
+    crack_times_seconds: NotRequired[_CrackTimeSeconds]
+    crack_times_display: NotRequired[_CrackTimesDisplay]
```
AlexWaygood, line 17: I thought we still couldn't use PEP 655 in typeshed (#6301)?

stubs/zxcvbn/zxcvbn/time_estimates.pyi
```diff
@@ -0,0 +1,24 @@
...
+    offline_slow_hashing_1e4_per_second: Decimal
+    offline_fast_hashing_1e10_per_second: Decimal
+
+class _CrackTimesDisplay(TypedDict):
+    online_throttling_100_per_hour: str
+    online_no_throttling_10_per_second: str
+    offline_slow_hashing_1e4_per_second: str
+    offline_fast_hashing_1e10_per_second: str
+
+def estimate_attack_times(guesses: Decimal) -> _TimeEstimate: ...
```
AlexWaygood, line 21: Are you sure this has to be a `Decimal`? It looks like the `Decimal` constructor is called on `guesses` inside this function, which would be a bit odd if it has to already be a `Decimal`: https://github.com/dwolfhub/zxcvbn-python/blob/f596bef499c02bb6f0c8aef33b260364add2dc2b/zxcvbn/time_estimates.py#L5-L8

stubs/zxcvbn/zxcvbn/matching.pyi
```diff
@@ -0,0 +1,88 @@
...
+class _DM(TypedDict):
+    month: int
+    day: int
+
+class _DMY(TypedDict):
+    year: int
+    month: int
+    day: int
+
+def map_ints_to_dmy(ints: tuple[int, int, int]) -> _DMY: ...
```
AlexWaygood, line 86: This can return `None`: https://github.com/dwolfhub/zxcvbn-python/blob/f596bef499c02bb6f0c8aef33b260364add2dc2b/zxcvbn/matching.py#L591

stubs/zxcvbn/zxcvbn/matching.pyi
```diff
@@ -0,0 +1,88 @@
...
+
+from .adjacency_graphs import _Graph
+
+class _Match(TypedDict):
+    pattern: Literal["dictionary", "spatial", "repeat", "sequence", "regex", "date"]
+    token: str
+    i: int
+    j: int
+    guesses: int
+    guesses_log10: float
```
AlexWaygood, line 14: `date_match()` is annotated as returning `list[_Match]` but the dictionaries in that list don't appear to have `guesses` or `guesses_log10` keys in them, so maybe these should be `NotRequired`?

Those dictionaries also have `year`, `month` and `day` keys, which don't currently appear in this `TypedDict`. https://github.com/dwolfhub/zxcvbn-python/blob/f596bef499c02bb6f0c8aef33b260364add2dc2b/zxcvbn/matching.py#L520-L529

And they sometimes have a `separator` key: https://github.com/dwolfhub/zxcvbn-python/blob/f596bef499c02bb6f0c8aef33b260364add2dc2b/zxcvbn/matching.py#L552

AlexWaygood (commented):
Mostly looks good, but I'm not sure about the huge `_Match` `TypedDict` in `matching.pyi`. It seems like we could probably have a smaller base `TypedDict`, and return subtypes of that `TypedDict` from specific functions where `_Match` always has certain other keys.

srittau:
> Mostly looks good, but I'm not sure about the huge `_Match` `TypedDict` in `matching.pyi`. It seems like we could probably have a smaller base `TypedDict`, and return subtypes of that `TypedDict` from specific functions where `_Match` always has certain other keys.

Basically everything except the `zxcvbn()` function in `__init__.py` is internal API anyway, so I wouldn't want to put the work in for a finer-grained API.

stubs/zxcvbn/zxcvbn/time_estimates.pyi
srittau: Whoops, I was under the (mistaken) assumption that `float` is a subtype of `Decimal`.

srittau: I'm skipping the other possible values except `float` here, since I don't think they are relevant.

stubs/zxcvbn/zxcvbn/matching.pyi
srittau: I'll have a look later.

AlexWaygood:
What do you want to do here?

1. Wait until pyre/pytype have PEP 655 support?
2. Rewrite the PR to use `TypedDict` inheritance for now?
3. YOLO it and go with PEP 655 even though pytype/pyre don't fully support it yet?

srittau:
I honestly think, we could just merge it. pytype should just assume `Any` if I understand pytype's type handling correctly. We can wait for pyre support, but on the other hand this is a new package, so it not working for a few weeks is not a regression.

stubs/zxcvbn/zxcvbn/__init__.pyi
```diff
@@ -0,0 +1,21 @@
...
+from .feedback import _Feedback
+from .matching import _Match
+from .time_estimates import _CrackTimesDisplay, _CrackTimeSeconds
+
+class _Result(TypedDict):
+    password: str
+    guesses: Decimal
+    guesses_log10: float
+    sequence: list[_Match]
+    calc_time: NotRequired[datetime.timedelta]
```
JelleZijlstra, line 15: This is set unconditionally, why is it `NotRequired`?

stubs/zxcvbn/zxcvbn/__init__.pyi
```diff
@@ -0,0 +1,21 @@
+import datetime
+from collections.abc import Iterable
+from decimal import Decimal
+from typing_extensions import Literal, NotRequired, TypedDict
+
+from .feedback import _Feedback
+from .matching import _Match
+from .time_estimates import _CrackTimesDisplay, _CrackTimeSeconds
+
+class _Result(TypedDict):
```
JelleZijlstra, line 10: Why don't you inherit from `_TimeEstimate`, since it copies over all the data from that dict?

stubs/zxcvbn/zxcvbn/matching.pyi
```diff
@@ -0,0 +1,94 @@
...
+    month: int
+    day: int
+
+class _DMY(TypedDict):
+    year: int
+    month: int
+    day: int
+
+def map_ints_to_dmy(ints: tuple[int, int, int]) -> _DMY | None: ...
+def map_ints_to_dm(ints: tuple[int, int]) -> _DM: ...
```
JelleZijlstra, line 93:
```suggestion
def map_ints_to_dm(ints: tuple[int, int]) -> _DM | None: ...
```

`zxcvbn.matching.map_ints_to_dm((0, 0))` returns None.

stubs/zxcvbn/zxcvbn/scoring.pyi
```diff
@@ -0,0 +1,47 @@
...
+    sequence: list[_Match]
+
+def nCk(n: int, k: int) -> float: ...
+def most_guessable_match_sequence(password: str, matches: Iterable[_Match], _exclude_additive: bool = ...) -> _GuessesResult: ...
+def estimate_guesses(match: _Match, password: str) -> Decimal: ...
+def bruteforce_guesses(match: _Match) -> int: ...
+def dictionary_guesses(match: _Match) -> int: ...
+def repeat_guesses(match: _Match) -> Decimal: ...
+def sequence_guesses(match: _Match) -> int: ...
+def regex_guesses(match: _Match) -> int: ...
```
JelleZijlstra, line 31:
```suggestion
def regex_guesses(match: _Match) -> int | None: ...
```

JelleZijlstra (commented):
A few small comments

PR #8090, opened 2022-06-16
Fix `box` parameter type of `Pillow.PIL.Image.paste`

eggplants:
When I ran `img.paste(cropped, box=(10.5, 10.5))` I got:

`TypeError: 'float' object cannot be interpreted as an integer`

So I think the `box` parameter's type should be not `tuple[float, float]` but `tuple[int, int]`. (same as `_Box`)


JelleZijlstra (approved):
Confirmed in the code here: https://github.com/python-pillow/Pillow/blob/c58d2817bc891c26e6b8098b8909c0eb2e7ce61b/src/_imaging.c#L1344

It wants `i` objects, which are ints (https://docs.python.org/3/c-api/arg.html#numbers)

Issue #8094, opened 2022-06-18
mypy complains about subclassing `imaplib.IMAP4.error`


JohnVillalovos:
Reproducer:

```
$ cat reproducer.py
#!/usr/bin/python3 -ttu

import imaplib

class MyException(imaplib.IMAP4.error): ...
```

Error:
```
$ mypy reproducer.py
reproducer.py:5: error: Variable "imaplib.IMAP4.error" is not valid as a type
reproducer.py:5: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#variables-vs-type-aliases
reproducer.py:5: error: Invalid base class "imaplib.IMAP4.error"
Found 2 errors in 1 file (checked 1 source file)
```

I believe this is because typeshed has `error` as `error: type[Exception]` when it should be `class error(Exception): ...`

PR #2273, opened 2018-06-22
Fix type for imaplib.IMAP4.error and related

nakamuray:
fix #2268

PR #8095, opened 2022-06-18
fix: type-hints for imaplib exceptions

JohnVillalovos:
`mypy` would complain when trying to sub-class the `imaplib`
exceptions.

Change the exceptions from a `type` to a `class`

Closes: #8094


JohnVillalovos:
Back in 2018 it was suggested to do this in PR #2273 that added these, but a test didn't work then. Now the CI tests are working.

hauntsaninja (approved):

hauntsaninja:
Thanks!

PR #8096, opened 2022-06-18
stubtest: use separate table in METADATA.toml

hauntsaninja:
I want to start enabling `--ignore-missing-stub` on third party packages, like we do for stdlib, since this helps ensure type completeness. I thought it might make sense to separate out the metadata little bit.

### DIFF
--- a/tests/check_consistent.py
+++ b/tests/check_consistent.py
@@ -25,8 +25,11 @@ metadata_keys = {
     "extra_description",
     "obsolete_since",
     "no_longer_updated",
-    "stubtest",
-    "stubtest_apt_dependencies",
+    "tool",
+}
+stubtest_keys = {
+    "skip",
+    "apt_dependencies",
 }
 allowed_files = {"README.md"}
 
@@ -183,6 +186,10 @@ def check_metadata() -> None:
                 assert dep_version.count(".") <= 2, f"Bad version '{dep_version}' in dependency {dep}"
                 for part in dep_version.split("."):
                     assert part.isnumeric(), f"Bad version '{part}' in dependency {dep}"
+
+        assert set(data.get("tool", [])).issubset({"stubtest"}), f"Unrecognised tool for {distribution}"
+        for key in data.get("tool", {}).get("stubtest", {}):
+            assert key in stubtest_keys, f"Unrecognised stubtest key {key} for {distribution}"
 
 
 if __name__ == "__main__":
--- a/tests/check_consistent.py
+++ b/tests/check_consistent.py
@@ -20,7 +20,7 @@ import tomli
 
 consistent_files = [{"stdlib/@python2/builtins.pyi", "stdlib/@python2/__builtin__.pyi"}]
 metadata_keys = {"version", "requires", "extra_description", "obsolete_since", "no_longer_updated", "tool"}
-stubtest_keys = {"skip", "apt_dependencies"}
+tool_keys = {"stubtest": {"skip", "apt_dependencies"}}
 allowed_files = {"README.md"}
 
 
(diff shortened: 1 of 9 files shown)


hauntsaninja:
I think the locale failures might be a 3.10.5 thing, since https://github.com/python/typeshed/runs/6946462440?check_suite_focus=true from a couple minutes ago was green with 3.10.4

AlexWaygood:
> I think the locale failures might be a 3.10.5 thing, since https://github.com/python/typeshed/runs/6946462440?check_suite_focus=true from a couple minutes ago was green with 3.10.4

Yeah looks like it. Now failing on master too. Weird that it's only failing on Darwin https://github.com/python/typeshed/runs/6946737526?check_suite_focus=true

AlexWaygood:
> I want to start enabling `--ignore-missing-stub` on third party packages, like we do for stdlib, since this helps ensure type completeness.

Idk about this. We've (correctly) always said in the past that "some stubs" are better than "no stubs". Isn't this going to increase the burden for third-party contributors to contribute stubs to typeshed?

hauntsaninja:
Yes, some stubs are often better than no stubs. The idea is that we make this per project configurable (and the default can be as is today).

This PR would mean that such per project stubtest configuration is more cleanly separated from the more important stuff in METADATA.toml

If that setup still sounds concerning, let's open an issue and discuss more? (since ignore-missing-stub is a little tangential to this PR)

AlexWaygood:
> Yes, some stubs are often better than no stubs. The idea is that we make this per project configurable (and the default can be as is today).

Okay, sg!

This PR also looks good. We could also consider merging the `requirements-stubtest.txt` file that we have for some stubs into `METADATA.toml`. It's a bit weird that we currently have `apt-dependencies` in `METADATA.toml`, but we keep regular dependencies in a separate file, e.g.: https://github.com/python/typeshed/blob/master/stubs/google-cloud-ndb/%40tests/requirements-stubtest.txt

tests/check_consistent.py
```diff
@@ -19,15 +19,8 @@
...
-    "version",
-    "requires",
-    "extra_description",
-    "obsolete_since",
-    "no_longer_updated",
-    "stubtest",
-    "stubtest_apt_dependencies",
-}
+metadata_keys = {"version", "requires", "extra_description", "obsolete_since", "no_longer_updated", "tool"}
+stubtest_keys = {"skip", "apt_dependencies"}
```
srittau, line 23: I think we should convert this to a dict of sets:

```suggestion
tool_keys = {"stubtest": {"skip", "apt_dependencies"}}
```

and then ...

tests/check_consistent.py
```diff
@@ -184,6 +177,10 @@ def check_metadata() -> None:
                 for part in dep_version.split("."):
                     assert part.isnumeric(), f"Bad version '{part}' in dependency {dep}"
 
+        assert set(data.get("tool", [])).issubset({"stubtest"}), f"Unrecognised tool for {distribution}"
+        for key in data.get("tool", {}).get("stubtest", {}):
+            assert key in stubtest_keys, f"Unrecognised stubtest key {key} for {distribution}"
```
srittau, line 182: ... we don't need to hardcode the tool here:

```suggestion
        assert set(data.get("tool", [])).issubset(tool_keys.keys()), f"Unrecognised tool for {distribution}"
        for tool, tk in tool_keys.items():
            for key in data.get("tool", {}).get(tool, {}):
                assert key in tk, f"Unrecognised {tool} key {key} for {distribution}"
```

teddyoweh (approved):

hauntsaninja:
I took a look at requirements-stubtest.txt. It seems like we have two use cases: 1) installing extras, 2) pinning things for cases where upstream pins aren't correct.
Agreed it would be nice to have consistency with apt_dependencies and put it in `METADATA.toml`. For case 1 it could also be nice to have "extra" installation be explicit, e.g. for html5lib we could do `package_extras = ["all"]` and then not have a requirements file.

That said, I think that's out of scope for this PR. I'd also want to look into the history of why we chose a separate file before doing any refactoring :-)

Issue #8098, opened 2022-06-18
Stubtest is failing on master 


AlexWaygood:
Specifically, on 3.10.5 on macos: https://github.com/python/typeshed/runs/6946773305?check_suite_focus=true

<details>

```
error: locale.bind_textdomain_codeset is not present at runtime
Stub: at line 147
def (builtins.str, Union[builtins.str, None]) -> Union[builtins.str, None]
Runtime:
MISSING

error: locale.bindtextdomain is not present at runtime
Stub: at line 146
def (builtins.str, Union[Union[builtins.str, os.PathLike[builtins.str]], None]) -> builtins.str
Runtime:
MISSING

error: locale.dcgettext is not present at runtime
Stub: at line 144
def (Union[builtins.str, None], builtins.str, builtins.int) -> builtins.str
Runtime:
MISSING

error: locale.dgettext is not present at runtime
Stub: at line 143
def (Union[builtins.str, None], builtins.str) -> builtins.str
Runtime:
MISSING

error: locale.gettext is not present at runtime
Stub: at line 142
def (builtins.str) -> builtins.str
Runtime:
MISSING

error: locale.textdomain is not present at runtime
Stub: at line 145
def (Union[builtins.str, None]) -> builtins.str
Runtime:
MISSING

error: socket.MsgFlag.MSG_NOSIGNAL is not present in stub
Stub:
MISSING
Runtime:
MsgFlag.MSG_NOSIGNAL
```

</details>

hauntsaninja:
#8096 just got a 3.10.4 run and passed. Maybe someone reverted something?

hauntsaninja:
Okay, built 3.10.5 locally and the locale ones do indeed seem like false positives. So seems like a problem with the Github build.
The socket one seems like a true positive (but it exists for me on 3.10.4 as well).
If we get failures, note that there's a nice stubtest allowlist trick we can use with an optional group to avoid unused allowlist entries:
https://github.com/python/mypy/blob/eb1b1e007e7fd1e976e6dd0f49a71b662a30a2d6/mypy/stubtest.py#L1459

PR #8099, opened 2022-06-19
socket: consistency between constants and enums

### LINKED ISSUE #5696: AF_BLUETOOTH mssing on win32, but available at runtime
```
C:\Users\IEUser>py
Python 3.9.5 (tags/v3.9.5:0a7dcbd, May  3 2021, 17:27:52) [MSC v.1928 64 bit (AMD64)] on win32
>>> import socket
>>> socket.AF_BLUETOOTH
<AddressFamily.AF_BLUETOOTH: 32>
>>>
```

also AF_UNIX is available in typeshed, but missing on win32 at runtime

hauntsaninja:
This matches what happens at runtime, from `Enum._convert_`
Helps with #8098. Fixes #5696.

### DIFF
--- a/stdlib/socket.pyi
+++ b/stdlib/socket.pyi
@@ -425,7 +425,7 @@ class AddressFamily(IntEnum):
         if sys.version_info >= (3, 8):
             AF_QIPCRTR: AddressFamily
     AF_LINK: AddressFamily  # availability: BSD, macOS
-    if sys.platform != "win32" and sys.platform != "darwin":
+    if sys.platform != "darwin":
         AF_BLUETOOTH: AddressFamily
 
 AF_UNIX: AddressFamily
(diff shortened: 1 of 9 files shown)


stdlib/socket.pyi
```diff
@@ -390,39 +390,43 @@ class AddressFamily(IntEnum):
...
     AF_SECURITY: int
     AF_SNA: int
     AF_SYSTEM: int
-    AF_TIPC: int
     AF_UNSPEC: int
-    AF_VSOCK: int
     AF_WANPIPE: int
     AF_X25: int
+    if sys.platform == "linux":
+        AF_CAN: AddressFamily
```
AlexWaygood, line 417: Shouldn't these all be annotated with `int`, like the other `AddressFamily` members?

hauntsaninja: Yup it should, thanks for catching!

AlexWaygood (approved):
Haven't double-checked them all, but all looks good, and I trust you/stubtest!

PR #8103, opened 2022-06-19
socket.AddressFamily: Fix AF_LINK

AlexWaygood:
Slipped through the net in #8099


hauntsaninja (approved):
Ahh, thanks for fixing my shoddy work!

PR #8104, opened 2022-06-19
Tweak pre-3.8 iscoroutine stub

florimondmanca:
Refs https://github.com/python/typeshed/pull/6105#discussion_r901080130

<Description WIP>


JelleZijlstra:
I don't think we should do this. `AwaitableGenerator` is a mypy-specific name that doesn't actually exist, and we shouldn't have it in `typing` in the first place.

hauntsaninja:
There is a real usability issue here though. I wouldn't mind lying here and having `Annotated[TypeGuard[Coroutine[Any, Any, Any]], "can actually be a generator-style coroutine"]` for 3.7 as well.

florimondmanca:
Hi, I updated the PR with the `Annotated` suggestion. Please let me know if that seems acceptable. In the end mypy treats `iscoroutine` as returning the same `TypeGuard[Coroutine[Any, Any, Any]]`, so the issue I mentioned would be resolved, but we do keep a note that it might be an `@asyncio.coroutine`.

stdlib/asyncio/coroutines.pyi
```diff
@@ -24,4 +23,6 @@ if sys.version_info >= (3, 8):
     def iscoroutine(obj: object) -> TypeGuard[Coroutine[Any, Any, Any]]: ...
 
 else:
-    def iscoroutine(obj: object) -> TypeGuard[types.GeneratorType[Any, Any, Any] | Coroutine[Any, Any, Any]]: ...
+    def iscoroutine(
+        obj: object,
+    ) -> Annotated[TypeGuard[Coroutine[Any, Any, Any]], "can actually be a generator-style coroutine"]: ...
```
JelleZijlstra, line 28: I'd prefer to have a comment instead of using Annotated here, in the off-chance this interacts poorly with someone else's use of Annotated.

AlexWaygood (approved):
Sounds like we're all happy with this now. Thanks for the PR!

Issue #8101, opened 2022-06-19
Add new or missing socket constants on linux


hauntsaninja:
We currently silence many errors from socket constants in stubtest, because the availability of some constants cannot be expressed in the type system (and probably also because historically we weren't always the best at specifying all platform differences exactly). However, this means we don't notice new additions.

On macOS, with python3.10:
```
λ python3.10 -m mypy.stubtest --custom-typeshed-dir ~/dev/typeshed socket --concise | grep 'is not present in stub'
socket.CAPI is not present in stub
socket.IP_RECVTOS is not present in stub
socket.PF_SYSTEM is not present in stub
socket.SYSPROTO_CONTROL is not present in stub
socket.TCP_KEEPALIVE is not present in stub
socket.TCP_NOTSENT_LOWAT is not present in stub
[...]
```

I can add PRs for these / check this for Linux. Filing this issue mainly as a reminder for myself to do so. Would be good if someone wants to take another look at Windows.

AlexWaygood:
I'm in favour of adding these, and of getting some more coverage from stubtest. I can take a look at the windows ones.

Missing `socket` constants have also been requested a few times by users:

- #7077
- #5696

AlexWaygood:
Stubtest is enabled for all the socket constants on Windows and macOS these days, but we still have two very broad allowlist entries on linux:

https://github.com/python/typeshed/blob/18cd196109938b690a97dc1dadf1c92b391b9639/tests/stubtest_allowlists/linux.txt#L1
https://github.com/python/typeshed/blob/18cd196109938b690a97dc1dadf1c92b391b9639/tests/stubtest_allowlists/linux.txt#L6

tungol:
I removed the two broad linux allowlist entries as part of https://github.com/python/typeshed/pull/13044 . This issue can probably be closed now.

AlexWaygood:
Fantastic, thank you!

PR #8106, opened 2022-06-19
Overhaul `socket` stubs on Windows and MacOS

AlexWaygood:
Helps with #8101.

This PR reduces the `(_)socket` allowlists for win32 and darwin to only a handful of missing constants/functions.

### DIFF
--- a/stdlib/_socket.pyi
+++ b/stdlib/_socket.pyi
@@ -166,15 +166,18 @@ IP_MULTICAST_LOOP: int
 IP_MULTICAST_TTL: int
 IP_OPTIONS: int
 IP_RECVDSTADDR: int
-if sys.version_info >= (3, 10) or (sys.platform != "win32" and sys.platform != "darwin"):
+if sys.version_info >= (3, 10):
+    IP_RECVTOS: int
+elif sys.platform != "win32" and sys.platform != "darwin":
     IP_RECVTOS: int
 IP_TOS: int
 IP_TTL: int
 MSG_CTRUNC: int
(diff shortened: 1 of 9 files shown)


hauntsaninja (commented):
Thanks, this is a pretty heroic PR. I'd want to double check there aren't any Linux regressions before merging.

Can confirm stubtest also looks good on my Mac laptop (in case there were OS version or hardware specific things we're missing). For what it's worth, I have `socket.MSG_NOSIGNAL` on 3.8 and above — curious what you were seeing.

AlexWaygood:
> For what it's worth, I have `socket.MSG_NOSIGNAL` on 3.8 and above — curious what you were seeing.

I was using GH Actions on my fork to run stubtest, and it _seemed_ like it varied from one stubtest run to the next as to whether it existed at runtime.

AlexWaygood:
> I'd want to double check there aren't any Linux regressions before merging.

Here's a failing Ubuntu stubtest run from yesterday on my fork, before I added the allowlist back for linux, if that's helpful: https://github.com/AlexWaygood/typeshed/runs/6955971392?check_suite_focus=true

hauntsaninja:
Confirmed that we didn't accidentally regress availability of some constants on Linux on Python 3.9

hauntsaninja (approved):

AlexWaygood:
> Confirmed that we didn't accidentally regress availability of any constants on Linux on Python 3.9 (same set of "is not present in stub" errors)

Awesome, thanks for checking!

PR #8109, opened 2022-06-20
dbapi2 set_progress_handler should return integer

### LINKED ISSUE #8105: dbapi2 set_progress_handler callback return type
I have a question regarding the type in dbapi's sqlite `set_progress_handler` callback return type.

It seems to be defined here:
https://github.com/python/typeshed/blob/a750a42c65b77963ff097b6cbb6d36cef5912eb7/stdlib/sqlite3/dbapi2.pyi#L344

and to set the return type to an optional bool.

But from what I see in the [docs](https://docs.python.org/3/library/sqlite3.html#sqlite3.Connection.set_progress_handler) and from the behaviour when trying to use it the callback should return an `int` and if that is a non-zero int the execution will be terminated.

Now, with boolean this will also work, but execution would be terminated if `True` is returned, and it would keep on going if `False` is returned.

I think it would make more sense to have the type definitions aligned with the docs and have it return an Optional int.

LefterisJP:
Fix #8105

dbapi2 set_progress_handler callback should be an int according to the
stdlib documentation: https://docs.python.org/3/library/sqlite3.html#sqlite3.Connection.set_progress_handler


srittau (approved):
Thanks!

PR #8114, opened 2022-06-21
asyncio.iscoroutine: fix return type (`Generator` -> `AwaitableGenerator`)

KotlinIsland:
As noted by @florimondmanca [here](https://github.com/python/typeshed/pull/6105#discussion_r901080130) the correct return type should be the sus type_check_only type `AwaitableGenerator`.


hauntsaninja:
Duplicate of #8104

PR #8115, opened 2022-06-21
`typing.AwaitableGenerator` add `type_check_only`

KotlinIsland:
`typing.AwaitableGenerator` is a type-time only member, it should be decorated with `type_check_only` (despite the fact that mypy doesn't support it python/mypy/issues/9531 :trollface:)

`type_check_only` was moved to the top of `typing` so that it can be used within the file.

### DIFF
--- a/stdlib/typing.pyi
+++ b/stdlib/typing.pyi
@@ -1,3 +1,5 @@
+from typing import type_check_only
+
 import collections  # Needed by aliases like DefaultDict, see mypy issue 2986
 import sys
 from _typeshed import IdentityFunction, ReadableBuffer, Self as TypeshedSelf, SupportsKeysAndGetItem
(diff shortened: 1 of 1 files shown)


stdlib/typing.pyi
```diff
@@ -391,6 +394,7 @@ class Coroutine(Awaitable[_V_co], Generic[_T_co, _T_contra, _V_co]):
 
 # NOTE: This type does not exist in typing.py or PEP 484 but mypy needs it to exist.
 # The parameters correspond to Generator, but the 4th is the original type.
+@type_check_only  #
```
JelleZijlstra, line 397:
```suggestion
@type_check_only
```
Not sure what the empty comment is for

JelleZijlstra (approved):

PR #8130, opened 2022-06-21
Replace str with bytes for imaplib  append and ParseFlags

tbelaire:
Last Argument of APPEND
https://github.com/python/cpython/blob/main/Lib/imaplib.py#L1497
https://github.com/python/cpython/blob/main/Lib/imaplib.py#L413
You can see it's parsed with the bytes regex MapCLRF `re.compile(br'\r\n|\r|\n')`

https://github.com/python/cpython/blob/main/Lib/imaplib.py#L1497
You can see it's parsed with the bytes regex Flags
`re.compile(br'.*FLAGS \((?P<flags>[^\)]*)\)')`


stdlib/imaplib.pyi
```diff
@@ -171,5 +171,5 @@ class _Authenticator:
 
 def Internaldate2tuple(resp: bytes) -> time.struct_time: ...
 def Int2AP(num: int) -> str: ...
-def ParseFlags(resp: str) -> tuple[str, ...]: ...
+def ParseFlags(resp: str) -> tuple[bytes, ...]: ...
```
Akuli, line 174: `ParseFlags("hi")` gives an error to me. The argument must be `bytes` too.

```suggestion
def ParseFlags(resp: bytes) -> tuple[bytes, ...]: ...
```

Akuli (approved):
Thanks!

stdlib/imaplib.pyi
tbelaire: Fixed.

PR #8134, opened 2022-06-22
mypy_test.py: Fix argument-parsing for `--python-version`

AlexWaygood:
Currently, the argument-parsing for the `--python-version` option is a little broken. If you try to pass more than one value to `--python-version` when running the test locally, it doesn't work.

This PR means that either of the following two invocations will work:

```
python tests/mypy_test.py -p3.6 -p3.7
```

```
python tests/mypy_test.py --python-version 3.6 3.7
```

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -17,7 +17,7 @@ in the `CONTRIBUTING.md` document. In particular, we recommend running with Pyth
 
 ## mypy\_test.py
 
-This test requires Python 3.7+. Run using:
+This test requires Python 3.8+. Run using:
 ```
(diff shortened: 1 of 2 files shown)


AlexWaygood:
Several of the other command-line options are also a little bit broken, but the fixes for some of those are a little more involved, and I'd rather tackle one problem at a time.

AlexWaygood:
> One thing I don't like about this argument (independent from your change) is that unknown Python versions are silently ignored. E.g. `--python-version 3.8 3.14 whatever` with only run for Python 3.8 and ignore the other arguments.

I'll fix that as part of my next PR.

Issue #8107, opened 2022-06-20
xml.sax.ContentHandler missing


stfujnkk:
I'm using a project that relies on typeshed, but there's a problem with the type hint.

[The issues link](https://github.com/microsoft/pylance-release/issues/2955)

raprocks:
Working on this

PR #8140, opened 2022-06-22
fix(`stdlib.xml`): re-export ContentHandler

raprocks:
fixes python/typeshed#8107

### DIFF
--- a/stdlib/xml/sax/__init__.pyi
+++ b/stdlib/xml/sax/__init__.pyi
@@ -2,7 +2,7 @@ import sys
 from _typeshed import SupportsRead, _T_co
 from collections.abc import Iterable
 from typing import Any, NoReturn, Protocol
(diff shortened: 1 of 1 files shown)


JelleZijlstra (approved):
The documentation only mentions it as `xml.sax.handler.ContentHandler` (https://docs.python.org/3.10/library/xml.sax.handler.html#xml.sax.handler.ContentHandler), but it's imported in `xml.sax.__init__` even though it doesn't get used there directly, so I think we can assume the re-export is intentional.

Should we also re-export ErrorHandler?

raprocks:
i think we should, should i commit with the changes?

AlexWaygood:
> i think we should, should i commit with the changes?

Go for it!

raprocks:
Done!

AlexWaygood (approved):

PR #8142, opened 2022-06-23
Improve `errno` stubs on Windows and MacOS

AlexWaygood:


hauntsaninja (commented):
Thanks for doing this, looks reasonable. Nit: can we coalesce the similar conditions and unflatten the nested ones?

AlexWaygood:
> Thanks for doing this, looks reasonable. Nit: can we coalesce the similar conditions and unflatten the nested ones?

We can, yes! Mainly did it this way to make the diff more readable, but I agree that it doesn't make the end result particularly readable. I'll make the change :)

stdlib/errno.pyi
```diff
@@ -119,40 +86,135 @@ EHOSTUNREACH: int
...
+    EPROCLIM: int
+    EPROCUNAVAIL: int
+    EPROGMISMATCH: int
+    EPROGUNAVAIL: int
+    EPWROFF: int
+    ERPCMISMATCH: int
+    ESHLIBVERS: int
+
+    if sys.platform != "darwin" or sys.version_info >= (3, 11):
+        EQFULL: int  # undocumented
```
AlexWaygood, line 117: Not possible to unflatten this nested check without triggering flake8-pyi's Y022

hauntsaninja (approved):

hauntsaninja:
Thank you!

PR #8146, opened 2022-06-24
re-exporting `HTTPError` in `urllib.request`

ewerybody:
to mitigate typechecker complains about unknown module members. [cpython/#94204](https://github.com/python/cpython/issues/94204)


hauntsaninja (commented):
(My thoughts on this here: https://github.com/python/cpython/issues/94204#issuecomment-1165543543 . Curious how others feel)

AlexWaygood (approved):
I also approve, as I said in the CPython issue. Concur with @hauntsaninja's thoughts on this.

srittau (approved):
Not too happy, but used in production.

PR #8147, opened 2022-06-24
add typing to ResponseCallbackProtocol objects

Syffe:
in order to be more compliant with typing overall


stubs/redis/redis/asyncio/client.pyi
```diff
@@ -13,10 +13,10 @@ from redis.typing import ChannelT, EncodableT, KeyT
 PubSubHandler: TypeAlias = Callable[[dict[str, str]], Awaitable[None]]
 
 class ResponseCallbackProtocol(Protocol):
-    def __call__(self, response: Any, **kwargs): ...
+    def __call__(self, response: Any, **kwargs) -> None: ...
```
JelleZijlstra, line 16: This should likely return `object` instead in order to be more permissive. Presumably the runtime just ignores whatever is returned.

AlexWaygood (changes requested):
As @JelleZijlstra says, `None` would be the wrong annotation for the return type here.

`ResponseCallbackProtocol` is used as a building block for the `ResponseCallbackT` type alias here: https://github.com/redis/redis-py/blob/e6cd4fdf3b159d7f118f154e28e884069da89d7e/redis/asyncio/client.py#L80:

```python
ResponseCallbackT = Union[ResponseCallbackProtocol, AsyncResponseCallbackProtocol]
```

The `ResponseCallbackT` alias is then used in `redis-py`'s codebase to annotate the `response_callbacks` attribute of the `Redis` class: https://github.com/redis/redis-py/blob/e6cd4fdf3b159d7f118f154e28e884069da89d7e/redis/asyncio/client.py#L98:

```python
class Redis(
    AbstractRedis, AsyncRedisModuleCommands, AsyncCoreCommands, AsyncSentinelCommands
):
    # <--- snip --->
    response_callbacks: MutableMapping[Union[str, bytes], ResponseCallbackT]
```

The `response_callbacks` dictionary is then used here in `redis-py`'s codebase to determine the returned value when the `Redis.parse_response` method is called: https://github.com/redis/redis-py/blob/e6cd4fdf3b159d7f118f154e28e884069da89d7e/redis/asyncio/client.py#L510-L514.

It seems like many non-`None` values could be returned from an object annotated as being `ResponseCallbackProtocol` or `AsyncResponseCallbackProtocol`, without `redis-py`'s codebase breaking. As such, saying "these functions have to return `None`" would be too restrictive.

Please change the return annotations of `ResponseCallbackProtocol` and `AsyncResponseCallbackProtocol` to be `object`, as requested, and then we can merge this 🙂

AlexWaygood:
Closing due to lack of response from OP

PR #8158, opened 2022-06-25
stubsabot: add project links

hauntsaninja:
See https://github.com/python/typeshed/pull/8157 for an example

### DIFF
--- a/scripts/stubsabot.py
+++ b/scripts/stubsabot.py
@@ -66,6 +66,7 @@ class PypiInfo:
     # https://warehouse.pypa.io/api-reference/json.html#get--pypi--project_name--json
     # Corresponds to a single entry from `releases` for the given version
     release_to_download: dict[str, Any]
+    info: dict[str, Any]
 
 
 async def fetch_pypi_info(distribution: str, session: aiohttp.ClientSession) -> PypiInfo:
@@ -82,6 +83,7 @@ async def fetch_pypi_info(distribution: str, session: aiohttp.ClientSession) -> 
             version=packaging.version.Version(version),
             upload_date=date,
             release_to_download=release_to_download,
+            info=j["info"],
         )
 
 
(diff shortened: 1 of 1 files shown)


JelleZijlstra (approved):
Nice, thanks

scripts/stubsabot.py
```diff
@@ -268,7 +282,9 @@ async def suggest_typeshed_update(update: Update, session: aiohttp.ClientSession
             return
         subprocess.check_call(["git", "push", "origin", branch_name, "--force-with-lease"])
 
-    body = """\
+    body = "\n".join(f"{k}: {v}" for k, v in update.links.items())
+    body += """
+
 If stubtest fails for this PR:
 - Leave this PR open (as a reminder, and to prevent stubsabot from opening another PR)
 - Fix stubtest failures in another PR, then close this PR
```
Akuli, line 290: I have never understood these instructions. Why not just fix the PR by pushing commits directly to it, and merge when stubtest passes again?

AlexWaygood: Yes, that seems to be the more common rhythm we're settling into.

hauntsaninja: It's a good question. It's because of what the current code will do if we run it automatically on a regular basis (rather than me running it manually in batches). stubsabot.py just makes a commit off of master and pushes it. master changes regularly, so these pushes would fail. Rather than failing, silently doing nothing, or some more complicated logic, it currently force pushes to the remote: https://github.com/python/typeshed/blob/a9fb8c5119989bc6cf754b6bd966d790d79bca39/scripts/stubsabot.py#L283
So this means it could undo changes we made.

But thinking about it a little bit more, silently doing nothing might not be that bad an option?

The other reason for these instructions is in theory non-maintainers could contribute + might benefit from instructions more so than maintainers.

PR #8159, opened 2022-06-25
PyYAML: check missing defs

hauntsaninja:


stubs/PyYAML/yaml/_yaml.pyi
```diff
@@ -33,6 +33,7 @@ class CParser:
     def get_single_node(self) -> Node | None: ...
     def raw_parse(self) -> int: ...
     def raw_scan(self) -> int: ...
+    __pyx_vtable__: Any
```
AlexWaygood, line 36: The whole `_yaml` namespace is a legacy thing kept for backwards compatibility, according to this comment in the source code: https://github.com/yaml/pyyaml/blob/8cdff2c80573b8be8e8ad28929264a913a63aa33/lib/_yaml/__init__.py#L1

I don't mind keeping the `_yaml` namespace in the stub, but I'm not sure about adding things like this that seem to be implementation details. Maybe it would be better to allowlist this one?

hauntsaninja:
Okay, changed to allowlist

AlexWaygood (approved):

PR #8166, opened 2022-06-26
Markdown: Fix BlockProcessor method return types

andersk:


hauntsaninja (approved):
Thanks, lg. It's not clear to me when or what the return value of `run` is used for

andersk:
It’s used in [`markdown.blockparser.BlockParser.parseBlocks`](https://github.com/Python-Markdown/markdown/blob/3.3.7/markdown/blockparser.py#L123). Also see the [documentation](https://python-markdown.github.io/extensions/api/#blockprocessors):

> `run` may return `False` (not `None`) to signal failure, meaning that it did not process the blocks after all.

The [built-in `BlockProcessors`](https://github.com/Python-Markdown/markdown/blob/3.3.7/markdown/blockprocessors.py) return all of `None`, `True`, and `False`.

PR #8168, opened 2022-06-26
DateTimeRange: complete stubs

hauntsaninja:


stubs/DateTimeRange/datetimerange/__init__.pyi
```diff
@@ -17,6 +20,14 @@ class DateTimeRange:
...
         end_time_format: str = ...,
     ) -> None: ...
+    @classmethod
+    def from_range_text(
+        cls: type[Self],
+        range_text: str,
+        separator: str = ...,
+        start_time_format: str | None = ...,
+        end_time_format: str | None = ...,
+    ) -> Self: ...
```
AlexWaygood, line 30: The return type is actually hardcoded here as being `DateTimeRange`, so using the `Self` TypeVar isn't actually correct here: https://github.com/thombashi/DateTimeRange/blob/17faff9dcda09a18490041c925b58811d4aac9a7/datetimerange/__init__.py#L870

hauntsaninja: Thanks, good catch!

AlexWaygood (approved):

PR #8171, opened 2022-06-26
`typing.TextIO`: add `write_through` and `reconfigure`.

ppentchev:
Hi,

Thanks a lot for taking care of typeshed, mypy, and the related projects!

What do you think about the attached trivial patch that makes it possible to use e.g. `sys.stdin.reconfigure(encoding="UTF-8")` in cases when it is needed?

Thanks again, and keep up the great work!

G'luck,
Peter

### DIFF
--- a/stdlib/typing.pyi
+++ b/stdlib/typing.pyi
@@ -706,6 +706,18 @@ class TextIO(IO[str]):
     def newlines(self) -> Any: ...  # None, str or tuple
     @abstractmethod
     def __enter__(self) -> TextIO: ...
+    if sys.version_info >= (3, 7):
+        @property
+        def write_through(self) -> bool: ...
+        def reconfigure(
+            self,
+            *,
+            encoding: str | None = ...,
+            errors: str | None = ...,
+            newline: str | None = ...,
+            line_buffering: bool | None = ...,
+            write_through: bool | None = ...,
+        ) -> None: ...
 
 class ByteString(Sequence[int], metaclass=ABCMeta): ...
 


ppentchev:
Hm. Okay. So the test suite seems to be correct: Python 3.7's typing module does indeed not declare TextIO.reconfigure() and TextIO.write_through. However, they are present in the actual sys.stdin/stdout/stderr objects. So... what do we do here? And okay, fine, "do nothing and add type: ignore on the sys.stdin.reconfgure() line" is an acceptable answer.

Akuli:
This was already "fixed" in #4035, but that fix doesn't work because `sys.stdin` is typed as `TextIO`, not as `TextIOWrapper` which is the class that provides `reconfigure()`. Changing `sys.stdin` to be a `TextIOWrapper` isn't great either, because then you can't do `sys.stdin = io.StringIO()`.

A couple things we could do instead:
- Use `sys.__stdin__.recoonfigure()`. The `__stdin__` attribute isn't meant to be reassigned to a `StringIO` or whatever other file-like object, so it can be typed as a `TextIOWrapper`, and it's already typed that way. If `sys.stdin` hasn't been reassigned, `__stdin__` and `stdin` refer to the same object.
- Change the type of `stdin` to be `TextIOWrapper | Any`. This means your code must work if `stdin` is a `TextIOWrapper`, but it could technically be something else too, and assigning something else to it is not an error.

srittau:
typeshed's `TextIO` type follows the implementation of `TextIO` in Python's standard library. The latter doesn't have `write_through()` or `reconfigure()`. Also, considering the pseudo-protocol nature of `TextIO`, it's difficult to add items to it, even in the standard library.

Akuli:
To be clear, here's how to do this with `isinstance`:

```python3
if isinstance(sys.stdout, io.TextIOWrapper) and sys.version_info >= (3, 7):
    sys.stdout.reconfigure(encoding="UTF-8")
```

ppentchev:
Thanks, that helps a lot! And yes, I get it, the I/O types are, hm, complicated. So yeah, thanks, and sorry for the noise!

PR #8175, opened 2022-06-26
Third-party stubs: audit `Callable[<parameters>, None]` annotations

AlexWaygood:
All of these callbacks have return values that are ignored at runtime.

Relevant to PyCQA/flake8-pyi/issues/237 (I was already looking at this before the issue was filed ;)


stubs/pynput/pynput/_util.pyi
```diff
@@ -67,6 +68,6 @@ class Events(Generic[_T, _AbstractListener_T]):
     def __iter__(self: Self) -> Self: ...
     def __next__(self) -> _T: ...
     def get(self, timeout: float | None = ...) -> _T | None: ...
-    def _event_mapper(self, event: Callable[_P, None]) -> Callable[_P, None]: ...
+    def _event_mapper(self, event: Callable[_P, _R]) -> Callable[_P, _R]: ...
```
srittau, line 71: The returned callback actually returns `None`.

stubs/waitress/waitress/utilities.pyi
```diff
@@ -43,7 +43,7 @@ class Error:
     body: str = ...
     def __init__(self, body: str) -> None: ...
     def to_response(self) -> tuple[str, Sequence[tuple[str, str]], str]: ...
-    def wsgi_response(self, environ: Any, start_response: Callable[[str, Sequence[tuple[str, str]]], None]) -> str: ...
+    def wsgi_response(self, environ: Any, start_response: Callable[[str, Sequence[tuple[str, str]]], object]) -> str: ...
```
srittau, line 46: Should probably use `_typeshed.wsgi.StartResponse`. For example, the second argument is a list in the implementation, which is actually required by the WSGI protocol. On the other hand, the return type is ignored, but I still think it's safer to use the predefined type.

srittau (changes requested):

stubs/pynput/pynput/_util.pyi
AlexWaygood: I was led astray by the presence of `functools.wraps`. But that just means that the returned callback might pretend, in some situations, that it doesn't return `None`. It will always return `None` in reality -- you're right.

srittau (approved):

PR #8173, opened 2022-06-26
Bump fpdf2 to 2.5.*

hauntsaninja:
Release: https://pypi.org/project/fpdf2/2.5.5/
Homepage: https://pyfpdf.github.io/fpdf2/

Mostly fixes #7792.

PR #8176, opened 2022-06-26
fpdf2: improve types

hauntsaninja:
Note that I did these looking at latest master of fpdf2. There are several differences between 2.4 and master: https://github.com/python/typeshed/runs/7059844991


stubs/fpdf2/fpdf/fpdf.pyi
```diff
@@ -57,11 +56,10 @@ class ToCPlaceholder(NamedTuple):
 
 class SubsetMap:
     def __init__(self, identities: list[int]) -> None: ...
-    def pick(self, unicode: int): ...
-    def dict(self): ...
+    def pick(self, unicode: int) -> int | None: ...
```
AlexWaygood, line 59: I think this always returns `int`. If `unicode` isn't in `self._map` when the method is called, it's inserted into `self._map` before `self._map.get(unicode)` is called. https://github.com/PyFPDF/fpdf2/blob/48f837963b7ae1956a0e5b23815063d23c1964f5/fpdf/fpdf.py#L246

stubs/fpdf2/fpdf/fpdf.pyi
```diff
@@ -139,43 +132,46 @@ class FPDF:
...
+    auto_page_break: bool
+    b_margin: float
+    page_break_trigger: float
     def set_auto_page_break(self, auto: bool, margin: float = ...) -> None: ...
-    zoom_mode: Any
-    layout_mode: Any
-    def set_display_mode(self, zoom, layout: str = ...) -> None: ...
-    def set_compression(self, compress) -> None: ...
-    title: Any
+    zoom_mode: str | float
```
AlexWaygood, line 152: Could be more precise here with Literal types: https://github.com/PyFPDF/fpdf2/blob/48f837963b7ae1956a0e5b23815063d23c1964f5/fpdf/fpdf.py#L587

But that might also lead to false positives.

stubs/fpdf2/fpdf/fpdf.pyi
```diff
@@ -139,43 +132,46 @@ class FPDF:
...
+    def set_display_mode(self, zoom: str | float, layout: str = ...) -> None: ...
```
AlexWaygood, line 153: Same here, `layout` could possibly be typed more precisely with `Literal`

stubs/fpdf2/fpdf/fpdf.pyi
```diff
@@ -139,43 +132,46 @@ class FPDF:
...
+    keywords: str
     def set_keywords(self, keywords: str) -> None: ...
-    creator: Any
+    creator: str
     def set_creator(self, creator: str) -> None: ...
-    producer: Any
+    producer: str
     def set_producer(self, producer: str) -> None: ...
-    creation_date: Any
+    creation_date: datetime.datetime | None
```
AlexWaygood, line 170: Can also be `True`: https://github.com/PyFPDF/fpdf2/blob/48f837963b7ae1956a0e5b23815063d23c1964f5/fpdf/fpdf.py#L438

stubs/fpdf2/fpdf/fpdf.pyi
```diff
@@ -185,37 +181,37 @@ class FPDF:
...
-    def circle(self, x, y, r, style: Any | None = ...) -> None: ...
+    def rect(self, x: float, y: float, w: float, h: float, style: Incomplete | None = ...) -> None: ...
+    def ellipse(self, x: float, y: float, w: float, h: float, style: Incomplete | None = ...) -> None: ...
+    def circle(self, x: float, y: float, r, style: Incomplete | None = ...) -> None: ...
     def add_font(self, family: str, style: _FontStyle = ..., fname: str | None = ..., uni: bool = ...) -> None: ...
     def set_font(self, family: str | None = ..., style: _FontStyles = ..., size: int = ...) -> None: ...
     def set_font_size(self, size: int) -> None: ...
-    def set_stretching(self, stretching) -> None: ...
-    def add_link(self): ...
+    def set_stretching(self, stretching: float) -> None: ...
```
AlexWaygood, line 204: `font_stretching` on line 92 of this stub needs to be changed to be `float` rather than `int`: https://github.com/PyFPDF/fpdf2/blob/48f837963b7ae1956a0e5b23815063d23c1964f5/fpdf/fpdf.py#L1922

stubs/fpdf2/fpdf/fpdf.pyi
```diff
@@ -185,37 +181,37 @@ class FPDF:
...
+    def add_link(self) -> int: ...
     def set_link(self, link, y: int = ..., x: int = ..., page: int = ..., zoom: str = ...) -> None: ...
-    def link(self, x, y, w, h, link, alt_text: Any | None = ...) -> None: ...
-    def text_annotation(self, x, y, text) -> None: ...
-    def add_action(self, action, x, y, w, h) -> None: ...
-    def text(self, x, y, txt: str = ...) -> None: ...
-    def rotate(self, angle, x: Any | None = ..., y: Any | None = ...) -> None: ...
-    def rotation(self, angle, x: Any | None = ..., y: Any | None = ...) -> _GeneratorContextManager[None]: ...
+    def link(self, x: float, y: float, w: float, h: float, link, alt_text: str | None = ...) -> None: ...
```
AlexWaygood, line 207: https://github.com/PyFPDF/fpdf2/blob/48f837963b7ae1956a0e5b23815063d23c1964f5/fpdf/fpdf.py#L1970

```suggestion
    def link(self, x: float, y: float, w: float, h: float, link: str | int, alt_text: str | None = ...) -> None: ...
```

AlexWaygood (changes requested):
Spotted a few things.

Also, could you change lines 71-73 to be `ClassVar`s?

hauntsaninja:
Thanks, made all those changes + a couple others!

stubs/fpdf2/fpdf/fpdf.pyi
AlexWaygood:
```suggestion
    zoom_mode: Literal["fullpage", "fullwidth", "real", "default"] | float
    page_layout: PageLayout | None
```

stubs/fpdf2/fpdf/fpdf.pyi
```diff
@@ -185,37 +184,38 @@ class FPDF:
...
+        transition: Incomplete | None = ...,
     ) -> None: ...
     def header(self) -> None: ...
     def footer(self) -> None: ...
     def page_no(self) -> int: ...
-    def set_draw_color(self, r, g: int = ..., b: int = ...) -> None: ...
-    def set_fill_color(self, r, g: int = ..., b: int = ...) -> None: ...
-    def set_text_color(self, r, g: int = ..., b: int = ...) -> None: ...
-    def get_string_width(self, s, normalized: bool = ..., markdown: bool = ...): ...
+    def set_draw_color(self, r: int, g: int = ..., b: int = ...) -> None: ...
```
AlexWaygood, line 192: The `draw_color` attribute is currently typed as being `str`, but that doesn't look right: https://github.com/PyFPDF/fpdf2/blob/48f837963b7ae1956a0e5b23815063d23c1964f5/fpdf/fpdf.py#L923

Same with `fill_color` and `text_color`

hauntsaninja: It's correct on 2.4.6. We can change with #8173. There are a couple other things like that

AlexWaygood (approved):
Thanks!

hauntsaninja:
(PageLayout will also need all the new enums in whatever 2.5.* variant they got introduced in)

AlexWaygood:
> (PageLayout will also need all the new enums in whatever 2.5.* variant they got introduced in)

Yeah, I spotted that after your previous comment :)

PR #8177, opened 2022-06-26
`JACK-Client`: audit callback annotations

AlexWaygood:
The returned values from these functions are all ignored at runtime.

Relevant to https://github.com/PyCQA/flake8-pyi/issues/237


hauntsaninja (approved):

srittau (approved):
Didn't verify, but I doubt someone would rely on a `None` return value.

PR #8188, opened 2022-06-27
Improve `xmlrpc.client` callbacks

AlexWaygood:
Relevant to PyCQA/flake8-pyi/issues/237


stdlib/xmlrpc/client.pyi
```diff
@@ -97,32 +97,29 @@ class ExpatParser:  # undocumented
     def feed(self, data: str | bytes) -> None: ...
     def close(self) -> None: ...
 
-class Marshaller:
-
-    dispatch: dict[
-        type[Any], Callable[[Marshaller, Any, Callable[[str], Any]], None]
-    ]  # TODO: Replace 'Any' with some kind of binding
+_WriteType: TypeAlias = Callable[[str], object]
```
srittau, line 100: Optional: Maybe `_WriteCallback`? "Type" is a bit unspecific.

stdlib/xmlrpc/client.pyi
```diff
@@ -97,32 +97,29 @@ class ExpatParser:  # undocumented
...
 
+class Marshaller:
+    # TODO: Replace 'Any' with some kind of binding
+    dispatch: dict[type[Any], Callable[[Marshaller, Any, _WriteType], None]]
```
srittau, line 104: Just use `_typeshed.Incomplete` instead of `Any`, in case this is potentially fixable.

srittau (changes requested):

stdlib/xmlrpc/client.pyi
AlexWaygood: I had a look at the source code writing this PR, and I get what the comment is saying, but I don't think it's really fixable tbh.

I'm wondering if we should introduce `_typeshed.Inexpressible` for cases like this, where we know what the type _should_ be (so it's not incomplete _per se_), it's just that we don't currently have any way of expressing it in the current type system. It might also be useful for places where we really need recursive type aliases, but can't use them yet.

srittau: Well, `Inexpressible` is basically `Any`, at least if we use `Incomplete` consistently.

srittau (approved):

PR #8191, opened 2022-06-27
Improve None as predicate for Python 2 ifilter

ddaskari:
Properly support `ifilter(None, ...)` on Python 2. Used Python 3 [`filter`](https://github.com/python/typeshed/blob/3e274583236ab58858025f9e4b2bc7bf060ef52d/stdlib/builtins.pyi#L1275) as a basis. 

I stumbled over this as `six.moves.filter` returns `itertools.ifilter`.


AlexWaygood:
Thanks! Unfortunately, I know very little about Python 2, so I won't be able to review this (though one of the other maintainers might).

Please be aware, however, that we're planning on deleting all of our remaining Python 2 stubs as soon as mypy drops support for Python 2, which could be very soon (#7367, https://github.com/python/mypy/issues/12237)

stdlib/@python2/itertools.pyi
```diff
@@ -22,8 +22,14 @@ class chain(Iterator[_T], Generic[_T]):
...
 def compress(data: Iterable[_T], selectors: Iterable[Any]) -> Iterator[_T]: ...
 def dropwhile(predicate: Callable[[_T], Any], iterable: Iterable[_T]) -> Iterator[_T]: ...
-def ifilter(predicate: Callable[[_T], Any] | None, iterable: Iterable[_T]) -> Iterator[_T]: ...
-def ifilterfalse(predicate: Callable[[_T], Any] | None, iterable: Iterable[_T]) -> Iterator[_T]: ...
+@overload
+def ifilter(predicate: None, iterable: Iterable[_T | None]) -> Iterator[_T]: ...
+@overload
+def ifilter(predicate: Callable[[_T], Any], iterable: Iterable[_T]) -> Iterator[_T]: ...
+@overload
+def ifilterfalse(predicate: None, iterable: Iterable[_T | None]) -> Iterator[_T]: ...
```
srittau, line 30: I don't think this is correct. Using Python 3:

```python
>>> from itertools import filterfalse
>>> list(filterfalse(None, [0,1,2,3,None]))
[0, None]
```

I don't think an overload for `ifilterfalse` makes sense here.

srittau (changes requested):

ddaskari:
@AlexWaygood thanks for the heads-up, I was expecting something like this 😬

@srittau I removed the `ifilterfalse` overload.

srittau (approved):

PR #8197, opened 2022-06-28
`markdown`: remove deleted methods of `Registry`.

leven-cn:
This change will be included in version 3.4


AlexWaygood:
Thanks! In general, however, we prefer to wait until a release with these changes has come out before making changes like this in the stub. Otherwise the stub won't reflect what the package looks like at runtime when a user does `pip install markdown`.

AlexWaygood (approved):
Markdown 3.4 is out! Thanks 😊

PR #8202, opened 2022-06-28
Improve `multiprocessing` stubs

AlexWaygood:
- Use bound methods in `multiprocessing/__init__.pyi` instead of duplicating the signatures from `multiprocessing.context.pyi`. Improve the signatures in `multiprocessing.context.pyi` where the ones in `multiprocessing.__init__.pyi` are more precise. This matches the runtime.
- Use re-exports instead of assignments for several functions in `multiprocessing/dummy/__init__.pyi`. This reduces ambiguity in the stub (they're not type aliases), and matches the runtime.
- Use an assignment for `multiprocessing.dummy.current_process`, instead of duplicating the signature from `threading.pyi`. Again, this matches the runtime.
- `Callable[<parameters>, Any]` -> `Callable[<parameters>, object]` for several callbacks where the returned value is ignored.

Relevant to https://github.com/PyCQA/flake8-pyi/issues/237.


stdlib/multiprocessing/context.pyi
```diff
@@ -42,12 +44,10 @@ class BaseContext:
     @staticmethod
     def active_children() -> list[BaseProcess]: ...
     def cpu_count(self) -> int: ...
-    # TODO: change return to SyncManager once a stub exists in multiprocessing.managers
-    def Manager(self) -> Any: ...
-    # TODO: change return to Pipe once a stub exists in multiprocessing.connection
-    def Pipe(self, duplex: bool = ...) -> Any: ...
+    def Manager(self) -> SyncManager: ...
+    def Pipe(self, duplex: bool = ...) -> tuple[Connection, Connection]: ...
```
hauntsaninja, line 48: This seems to return a Pipe: https://github.com/python/cpython/blob/68fb03249f3b17146db42b00a75718b823a2280c/Lib/multiprocessing/context.py#L63

stdlib/multiprocessing/context.pyi
```diff
@@ -120,7 +120,7 @@ class BaseContext:
         @overload
         def get_context(self, method: str) -> BaseContext: ...
 
-    def get_start_method(self, allow_none: bool = ...) -> str: ...
+    def get_start_method(self, allow_none: bool = ...) -> str | None: ...
```
hauntsaninja, line 123: I don't think this is right, where do you see None?

AlexWaygood: It is documented as potentially returning `None`: https://docs.python.org/3/library/multiprocessing.html#multiprocessing.get_start_method

stdlib/multiprocessing/context.pyi
hauntsaninja: Ugh, thanks multiprocessing. Should've known to look another level :-)

Hmm, on Windows it returns a tuple of PipeConnection https://github.com/python/cpython/blob/68fb03249f3b17146db42b00a75718b823a2280c/Lib/multiprocessing/connection.py#L577
Maybe we annotate this (and multiprocessing.connection.Pipe) as returning `tuple[_ConnectionBase, _ConnectionBase]`

stdlib/multiprocessing/context.pyi
hauntsaninja: Seems like we'd ideally have an overload based on allow_none?
```suggestion
    @overload
    def get_start_method(self, allow_none: Literal[False] = ...) -> str: ...
    @overload
    def get_start_method(self, allow_none: bool) -> str | None: ...
```

hauntsaninja (approved):

PR #8203, opened 2022-06-28
`subprocess.Popen`: `__new__` -> `__init__`

### LINKED ISSUE #3363: when calling super method arguments are inherited from `object` instead of parent class
* Are you reporting a bug, or opening a feature request?
Bug

* Please insert below the code you are checking with mypy,
  or a mock-up repro if the source is private. We would appreciate
  if you try to simplify your case to a minimal repro.

```python
import subprocess

class MyPopen(subprocess.Popen):

    def __init__(self) -> None:
        args = ['echo', 'test']
        super().__init__(
            args=args,
        )
```

* What is the actual behavior/output?
```
$ python -m mypy mypy_subprocess.py
mypy_subprocess.py:9: error: Unexpected keyword argument "args" for "__init__" of "object"
```

* What are the versions of mypy and Python you are using?
```
$ python -m mypy --version
mypy 0.740+dev.742d33a41271206d6eb7dd2e4b04659de4683597

$ python --version
Python 3.7.4
```

AlexWaygood:
Fixes #3363

### DIFF
--- a/stdlib/subprocess.pyi
+++ b/stdlib/subprocess.pyi
@@ -2155,14 +2155,11 @@ class Popen(Generic[AnyStr]):
     returncode: int | Any
     universal_newlines: bool
 
-    # Technically it is wrong that Popen provides __new__ instead of __init__
-    # but this shouldn't come up hopefully?
-
     if sys.version_info >= (3, 11):
         # process_group is added in 3.11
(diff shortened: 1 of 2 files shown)


srittau (approved):
Nice!

PR #8205, opened 2022-06-29
asyncio task factory can accept Coroutines too

### LINKED ISSUE #8194: asyncio task factory and Task type compatibility
The typeshed for asyncio's [`get_task_factory` and `set_task_factory`](https://github.com/python/mypy/blob/master/mypy/typeshed/stdlib/asyncio/events.pyi#L196-L199) declare the factory as a callable accepting `Generator[Any, None, _T]` for the `coro` argument. Meanwhile, [`Task` accepts `Generator[_TaskYieldType, None, _T] | Awaitable[_T]`](https://github.com/python/mypy/blob/master/mypy/typeshed/stdlib/asyncio/tasks.pyi#L297) and [`create_task` accepts `Coroutine[Any, Any, _T] | Generator[Any, None, _T]`](https://github.com/python/mypy/blob/master/mypy/typeshed/stdlib/asyncio/events.pyi#L189-L190). This creates incompatibilities with code working with factory type in `get/set_task_factory`. For example,

```
import asyncio

@asyncio.coroutine
def old_style() -> None:
    pass

async def coro() -> None:
    pass

loop = asyncio.get_running_loop()
factory = loop.get_task_factory()
if factory:
    asyncio.create_task(coro())
    factory(loop, coro()) #  error: Argument 2 has incompatible type "Coroutine[Any, Any, None]"; expected "Generator[Any, None, <nothing>]"  [arg-type]
    asyncio.create_task(old_style())
    factory(loop, old_style())
```

From my understanding of the [task factory docs](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.set_task_factory) that the `coro` factory argument be a coroutine, I'd expect for the code above to pass all type checks. Should the typeshed for `get/set_task_factory` be updated to match `Task` and `create_task`, allowing either a `Generator` or `Awaitable`?

ccotter:
Fixes #8194


AlexWaygood (commented):
Thanks, this looks good to me! Just one nit: `Callable[[AbstractEventLoop, Coroutine[Any, Any, _T] | Generator[Any, None, _T]], Future[_T]]` is a fairly complex type to have repeated across two files. Could you create a type alias in `asyncio/events.pyi`, and then import that alias in `asyncio/base_events.pyi`?

E.g. you'd do something like this in `events.pyi`:

```python
# In the global scope
_TaskFactory: TypeAlias = Callable[[AbstractEventLoop, Coroutine[Any, Any, _T] | Generator[Any, None, _T]], Future[_T]]

# Inside the class definition
def set_task_factory(self, factory: _TaskFactory | None) -> None: ...
def get_task_factory(self) -> _TaskFactory | None: ...
```

And something like this in `base_events.pyi`:

```python
# In the global scope
from asyncio.events import _TaskFactory

# Inside the class definition
def set_task_factory(self, factory: _TaskFactory | None) -> None: ...
def get_task_factory(self) -> _TaskFactory | None: ...
```

AlexWaygood:
I'm looking at the pyright errors

stdlib/asyncio/events.pyi
```diff
@@ -80,6 +80,7 @@ _Context: TypeAlias = dict[str, Any]
 _ExceptionHandler: TypeAlias = Callable[[AbstractEventLoop, _Context], Any]
 _ProtocolFactory: TypeAlias = Callable[[], BaseProtocol]
 _SSLContext: TypeAlias = bool | None | ssl.SSLContext
+_TaskFactory: TypeAlias = Callable[[AbstractEventLoop, Coroutine[Any, Any, _T] | Generator[Any, None, _T]], Future[_T]]
```
AlexWaygood, line 83: Pyright now thinks that we want this to be a generic type alias. We can fix this by using a callback protocol instead:

```suggestion

class _TaskFactory(Protocol):
    def __call__(self, __loop: AbstractEventLoop, __factory: Coroutine[Any, Any, _T] | Generator[Any, None, _T]) -> Future[_T]: ...
```

AlexWaygood (approved):
Thanks!

ccotter:
Thanks!

AlexWaygood:
> Thanks!

Thanks for the PR!

PR #8207, opened 2022-06-29
Bump prettytable to 3.3.*

hauntsaninja:
Release: https://pypi.org/project/prettytable/3.3.0/
Homepage: https://github.com/jazzband/prettytable


hauntsaninja:
ignore_missing_stub showing its usefulness here :-)

stubs/prettytable/prettytable/colortable.pyi
```diff
@@ -0,0 +1,41 @@
...
+        vertical_color: str = ...,
+        horizontal_char: str = ...,
+        horizontal_color: str = ...,
+        junction_char: str = ...,
+        junction_color: str = ...,
+    ) -> None: ...
+    def format_code(s: str) -> str: ...  # type: ignore
+
+class Themes:
+    DEFAULT: Theme
```
AlexWaygood, line 31:
```suggestion
    DEFAULT: ClassVar[Theme]
```

stubs/prettytable/prettytable/colortable.pyi
```diff
@@ -0,0 +1,41 @@
...
+    OCEAN: Theme
```
AlexWaygood, line 32:
```suggestion
    OCEAN: ClassVar[Theme]
```

stubs/prettytable/prettytable/colortable.pyi
```diff
@@ -0,0 +1,41 @@
+from _typeshed import Incomplete
+from collections.abc import Callable
+
+from .prettytable import PrettyTable as PrettyTable
```
AlexWaygood, line 4: Is this re-export necessary here?

```suggestion
from .prettytable import PrettyTable
```

AlexWaygood (commented):
(Haven't looked at prettytable.pyi yet)

stubs/prettytable/prettytable/prettytable.pyi
```diff
@@ -141,6 +145,62 @@ class PrettyTable:
     def oldsortslice(self): ...
     @oldsortslice.setter
     def oldsortslice(self, val) -> None: ...
+    @property
+    def bottom_junction_char(self) -> Incomplete: ...
```
AlexWaygood, line 149: I think we can just leave this unannotated rather than doing an explicit `-> Incomplete` here. Same goes for most of the other properties in this class that you've added.

```suggestion
    def bottom_junction_char(self): ...
```

stubs/prettytable/prettytable/prettytable.pyi
```diff
@@ -11,6 +12,9 @@ PLAIN_COLUMNS: int
 MARKDOWN: int
 ORGMODE: int
 RANDOM: int
+SINGLE_BORDER: int
+DOUBLE_BORDER: int
+BASE_ALIGN_VALUE: str
 
 class PrettyTable:
```
AlexWaygood, line 19: This class appears to be missing the `header_align` property, and I'm not sure I understand why stubtest isn't picking up on it being missing: https://github.com/jazzband/prettytable/blob/4345954111637455549bb7dab1e6f3714c32b0a6/src/prettytable/prettytable.py#L663-L664

stubs/prettytable/prettytable/prettytable.pyi
```diff
@@ -141,6 +145,62 @@ class PrettyTable:
...
+    @top_right_junction_char.setter
+    def top_right_junction_char(self, val) -> None: ...
+    @property
+    def xhtml(self) -> bool: ...
+    @xhtml.setter
+    def xhtml(self, val: bool) -> None: ...
+    @property
+    def rows(self) -> list[list[Any]]: ...
+    def add_autoindex(self, fieldname: str = ...): ...
+    def get_latex_string(self, **kwargs: Incomplete) -> str: ...
```
AlexWaygood, line 203: I'd just leave this unannotated

```suggestion
    def get_latex_string(self, **kwargs) -> str: ...
```

stubs/prettytable/prettytable/prettytable.pyi
```diff
@@ -141,6 +145,62 @@ class PrettyTable:
...
+    @property
+    def top_right_junction_char(self) -> Incomplete: ...
+    @top_right_junction_char.setter
+    def top_right_junction_char(self, val) -> None: ...
+    @property
+    def xhtml(self) -> bool: ...
+    @xhtml.setter
+    def xhtml(self, val: bool) -> None: ...
+    @property
+    def rows(self) -> list[list[Any]]: ...
```
AlexWaygood, line 201: Are you sure? It looks like a list of three-element tuples to me: https://github.com/jazzband/prettytable/blob/4345954111637455549bb7dab1e6f3714c32b0a6/src/prettytable/prettytable.py#L2489

stubs/prettytable/prettytable/prettytable.pyi
hauntsaninja: It's not released https://github.com/jazzband/prettytable/blob/3.3.0/src/prettytable/prettytable.py

stubs/prettytable/prettytable/prettytable.pyi
hauntsaninja: This is what I was looking at: https://github.com/jazzband/prettytable/blob/8ce5bd8956e12726f313db9644b03292400a49a7/src/prettytable/prettytable.py#L1414

stubs/prettytable/prettytable/prettytable.pyi
AlexWaygood: Ah, that explains it!

AlexWaygood (approved):

PR #8209, opened 2022-06-29
stdlib: audit more callback annotations

AlexWaygood:
Relevant to PyCQA/flake8-pyi/issues/237


stdlib/shutil.pyi
```diff
@@ -82,17 +82,15 @@ else:
         ignore_dangling_symlinks: bool = ...,
     ) -> _PathReturn: ...
 
+_OnErrorCallback: TypeAlias = Callable[[Callable[[StrOrBytesPath, OptExcInfo], object], StrOrBytesPath, OptExcInfo], object]
```
AlexWaygood, line 85:
```suggestion
_OnErrorCallback: TypeAlias = Callable[[Callable[..., object], StrOrBytesPath, OptExcInfo], object]
```

stdlib/shutil.pyi
```diff
@@ -82,17 +82,15 @@ else:
         ignore_dangling_symlinks: bool = ...,
     ) -> _PathReturn: ...
 
+_OnErrorCallback: TypeAlias = Callable[[Callable[..., object], StrOrBytesPath, OptExcInfo], object]
```
AlexWaygood, line 85:
```suggestion
_OnErrorCallback: TypeAlias = Callable[[Callable[..., object], Any, Any], object]
```

stdlib/shutil.pyi
```diff
@@ -1,6 +1,6 @@
 import os
 import sys
-from _typeshed import BytesPath, StrOrBytesPath, StrPath, SupportsRead, SupportsWrite
+from _typeshed import BytesPath, OptExcInfo, StrOrBytesPath, StrPath, SupportsRead, SupportsWrite
```
AlexWaygood, line 3:
```suggestion
from _typeshed import BytesPath, StrOrBytesPath, StrPath, SupportsRead, SupportsWrite
```
```suggestion
from _typeshed import BytesPath, OptExcInfo, StrOrBytesPath, StrPath, SupportsRead, SupportsWrite
```

stdlib/concurrent/futures/process.pyi
```diff
@@ -10,6 +10,9 @@ from weakref import ref
 
 from ._base import Executor, Future
 
+_R = TypeVar("_R")
+_S = TypeVar("_S")
```
JelleZijlstra, line 14: Any reason not to use a single TypeVar named `_T`?

stdlib/sched.pyi
```diff
@@ -1,23 +1,26 @@
 import sys
 from collections.abc import Callable
 from typing import Any, NamedTuple
+from typing_extensions import TypeAlias
 
 __all__ = ["scheduler"]
 
+_ActionCallback: TypeAlias = Callable[..., Any]
```
JelleZijlstra, line 8:
```suggestion
_ActionCallback: TypeAlias = Callable[..., object]
```
Looks like the return value is ignored

stdlib/shutil.pyi
```diff
@@ -82,17 +82,15 @@ else:
         ignore_dangling_symlinks: bool = ...,
     ) -> _PathReturn: ...
 
+_OnErrorCallback: TypeAlias = Callable[[Callable[..., object], Any, Any], object]
```
JelleZijlstra, line 85:
```suggestion
_OnErrorCallback: TypeAlias = Callable[[Callable[..., Any], Any, Any], object]
```
Argument positions are contravariant so it's better to be permissive.

stdlib/wsgiref/validate.pyi
```diff
@@ -25,9 +26,11 @@ class ErrorWrapper:
     def writelines(self, seq: Iterable[str]) -> None: ...
     def close(self) -> NoReturn: ...
 
+_WriterCallback: TypeAlias = Callable[[bytes], Any]
```
JelleZijlstra, line 29:
```suggestion
_WriterCallback: TypeAlias = Callable[[bytes], object]
```
Return value is ignored

stdlib/sched.pyi
AlexWaygood: It's used as an attribute annotation as well, though, and I wanted to avoid unexpected false positives on code like

```python
e: Event
x = e.action()
x.foo()
```

I'm not sure if it makes sense to do that, but I didn't want to cause a regression. Same for `_WriterCallback` in `wsgiref.validate`.

stdlib/concurrent/futures/process.pyi
AlexWaygood: Done

PR #8210, opened 2022-06-30
fix: `_Color` type alias of `PIL.Image`

Antyos:
The `_Color` type for Pillow's `Image` is defined as an integer, not a float.

```python
from PIL import Image
>>> Image.new("RGB", (100, 100), 0.0)
TypeError: color must be int or tuple
>>> Image.new("RGB", (100, 100), (0.0, 0.0, 0.0))  
TypeError: 'float' object cannot be interpreted as an integer
```

Colors will also throw an error if they are not 1-, 3-, or 4- tuples, and strings can be interpolated as colors.

```python
from PIL import Image
>>> Image.new("RGB", (100, 100), (0, 0))
TypeError: color must be int, or tuple of one, three or four elements
>>> Image.new("RGB", (100, 100), "black")  # ok
>>> Image.new("RGB", (100, 100), "#000000")  # ok
```

Also, I think #8090 may have missed references to `_Box` on lines 155 and 178, though I could be wrong. Would that be better as its own PR?


AlexWaygood:
You've hit a known mypy bug with PEP 604 type aliases: https://github.com/python/mypy/issues/11098

The solution is to use an old-style `Union` here.

eggplants:
> I think https://github.com/python/typeshed/pull/8090 may have missed references to _Box on lines 155 and 178.

I only noticed the mistake and corrected it while using `paste`, but did not actually check or verify Pillow's source for the others.

Lines 155 and 178 are suspicious, but you have to make sure their type actually is wrong.

AlexWaygood:
> Also, I think #8090 may have missed references to `_Box` on lines 155 and 178, though I could be wrong. Would that be better as its own PR?

Yeah, let's fix one problem at a time.

stubs/Pillow/PIL/Image.pyi
```diff
@@ -22,7 +22,7 @@ _Box: TypeAlias = tuple[int, int, int, int]
 _ConversionMatrix: TypeAlias = Union[
     tuple[float, float, float, float], tuple[float, float, float, float, float, float, float, float, float, float, float, float],
 ]
-_Color: TypeAlias = float | tuple[float, ...]
+_Color: TypeAlias = Union[int, tuple[int], tuple[int, int, int], tuple[int, int, int, int], str]
```
AlexWaygood, line 25: This isn't correct. For some image modes, a `float` value will do. Here's an example using the "F" mode for both `.new()` and `.putpixel()`:

```python
>>> from PIL import Image
>>> x = Image.new('F', (100, 100), 34.98)
>>> x.putpixel((98, 97), 76.32)
>>> x.show()
```

We could be fancy with overloads and generics, but for now I suggest doing this:

```suggestion
# `str` values are only accepted if mode="RGB" for an `Image` object
# `float` values are only accepted for certain modes such as "F"
# See https://pillow.readthedocs.io/en/stable/reference/Image.html#PIL.Image.new
_Color: TypeAlias = Union[int, tuple[int], tuple[int, int, int], tuple[int, int, int, int], str, float]
```

AlexWaygood (changes requested):
Thanks, but this isn't quite correct at the moment: see my comment below.

stubs/Pillow/PIL/Image.pyi
Antyos: Good catch. Overloaded generics sounds complicated and error-prone-- let's not do that.

In my brief testing, you can do a 1-tuple for a floating-point color, (but not any longer than that), so we should add that as well.

```python
>>> from PIL import Image
>>> Image.new('F', (100, 100), (34.98,))
```

I'm still not too familiar with code reviews; can I make a change to your suggestion before committing, or do I need to make two commits?

Akuli: It's easiest to make two commits.

AlexWaygood: Yeah, either works -- either commit my suggestion and then do a follow-up commit, or disregard my suggestion and make your own commit. It'll all be squash-merged in the end anyway, so don't worry about making a mess with the commit history :)

Antyos (commented):
add tuple[float] to _Color

stubs/Pillow/PIL/Image.pyi
```diff
@@ -22,7 +22,10 @@ _Box: TypeAlias = tuple[int, int, int, int]
 _ConversionMatrix: TypeAlias = Union[
     tuple[float, float, float, float], tuple[float, float, float, float, float, float, float, float, float, float, float, float],
 ]
-_Color: TypeAlias = float | tuple[float, ...]
+# `str` values are only accepted if mode="RGB" for an `Image` object
+# `float` values are only accepted for certain modes such as "F"
+# See https://pillow.readthedocs.io/en/stable/reference/Image.html#PIL.Image.new
+_Color: TypeAlias = Union[int, tuple[int], tuple[int, int, int], tuple[int, int, int, int], str, float]
```
AlexWaygood, line 28:
```suggestion
_Color: TypeAlias = Union[int, tuple[int], tuple[int, int, int], tuple[int, int, int, int], str, float, tuple[float]]
```

AlexWaygood (approved):
Thanks!

PR #8211, opened 2022-06-30
[stubsabot] Bump dataclasses to 0.8

hauntsaninja:
Release: https://pypi.org/project/dataclasses/0.8/
Homepage: https://github.com/ericvsmith/dataclasses

### DIFF
--- a/stubs/dataclasses/METADATA.toml
+++ b/stubs/dataclasses/METADATA.toml
@@ -1,1 +1,1 @@
-version = "0.6"
+version = "0.8"


hauntsaninja:
Hmm, dataclasses 0.7 onwards requires Python 3.6. See https://github.com/ericvsmith/dataclasses/commit/44724c459fcca9ddeab77b8d79d0790bb1c3163e. (Of course, prior to that, on Python 3.7 the install would work, but the resulting module would not be importable)

Given Python 3.6 EOL, I think we should mark these as "no longer updated" or whatever, then delete them.

AlexWaygood:
> Given Python 3.6 EOL, I think we should mark these as "no longer updated" or whatever, then delete them.

Yeah, no point in investing time & energy to update 3.6-only stubs if we could be removing support for 3.6 entirely in just a few days.

srittau:
Obsoleted by #8216.

PR #8216, opened 2022-06-30
Remove the dataclasses third-party stub

srittau:
dataclasses is part of the Python stdlib and the third-party package is
not supported past Python 3.6.


srittau:
I manually triggered the stub uploader: https://pypi.org/project/types-dataclasses/#description

PR #8218, opened 2022-06-30
support lists of same type futures in asyncio.gather()

perrinjerome:


perrinjerome:
I submitted by mistake, this is not ready at all

PR #8226, opened 2022-07-01
[stubsabot] Bump Pillow to 9.2.*

hauntsaninja:
Release: https://pypi.org/project/Pillow/9.2.0/
Homepage: https://python-pillow.org
Changelog: https://github.com/python-pillow/Pillow/blob/main/CHANGES.rst


stubs/Pillow/PIL/Image.pyi
```diff
@@ -158,7 +157,7 @@ class Image:
     def getextrema(self): ...
     def getexif(self) -> Exif: ...
     def getim(self): ...
-    def getpalette(self) -> list[int] | None: ...
+    def getpalette(self, rawmode: str = ...) -> list[int] | None: ...
```
srittau, line 160: Accepts `None`, even if a `str` is the default:

```suggestion
    def getpalette(self, rawmode: str | None = ...) -> list[int] | None: ...
```

srittau (changes requested):

srittau (approved):

PR #8227, opened 2022-07-01
create_baseline_stubs: check all defs by default

hauntsaninja:
Since stubgen will usually create definition complete stubs, we should have
stubtest check missing definitions so we can know to reflect any future
additions.


srittau:
What about making `ignore_missing_stub` the default? This would need a change to quite a few `METADATA.toml` files, but by now incomplete stubs are more of a legacy feature, I think.

AlexWaygood (approved):
I worried at first that this would lead to contributors unnecessarily including implementation details in new stubs they contributed. But stubgen already does that anyway, so this wouldn't change anything in that regard.

I think making `ignore_missing_stub` `False` by default is something we should work towards, but it doesn't necessarily have to be done right now.

PR #3464, opened 2019-11-14
Add warnings.catch_warning type hints with Literal.

mslapek:
Fixes #3463

PR #8229, opened 2022-07-02
Improve `warnings.catch_warnings`

AlexWaygood:
- `__new__` -> `__init__`. This kind of `__init__` overloading wasn't possible when the current stubs were written (see discussion in #3464).
- Remove the fictitious classes `_catch_warnings_without_records` and `_catch_warnings_with_records`, which don't exist at rutnime. Make the class generic instead.

### DIFF
--- a/stdlib/warnings.pyi
+++ b/stdlib/warnings.pyi
@@ -2,7 +2,7 @@ import sys
 from _warnings import warn as warn, warn_explicit as warn_explicit
 from collections.abc import Sequence
 from types import ModuleType, TracebackType
-from typing import Any, TextIO, overload
+from typing import Any, Generic, TextIO, TypeVar, overload
 from typing_extensions import Literal, TypeAlias
 
 __all__ = [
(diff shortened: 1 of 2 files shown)


AlexWaygood:
> 
> ```diff
> pytest (https://github.com/pytest-dev/pytest)
> + src/_pytest/recwarn.py:160: error: Missing type parameters for generic type "catch_warnings"  [type-arg]
> ```

This is a custom subclass of `warnings.catch_warnings`: https://github.com/pytest-dev/pytest/blob/b08c1dcf0e421b408653d185574baec7855e2439/src/_pytest/recwarn.py#L160

A comment in the source code indicates that, as they are well aware, this subclass already violates LSP by returning `self` in their `__enter__` method: https://github.com/pytest-dev/pytest/blob/b08c1dcf0e421b408653d185574baec7855e2439/src/_pytest/recwarn.py#L209-L211

srittau (approved):
Much cleaner.

PR #8232, opened 2022-07-03
(🐞) `__future__._Feature.getMandatoryRelease` should be Optional

KotlinIsland:
And it shouldn't be a `sys._version_info`, it should be a 5-tuple.

> This is a 5-tuple, of the same form as sys.version_info, or, if the feature was dropped, or it's release date is undetermined, is None.

This is also needed for https://github.com/python/cpython/pull/93628


stdlib/__future__.pyi
```diff
@@ -1,9 +1,12 @@
 import sys
+from typing_extensions import TypeAlias
+
+_version_info: TypeAlias = tuple[int, int, int, str, int]
 
 class _Feature:
```
graingert, line 6: Might be cool to have this as a

```
_T_minver = TypeVar("_T_minver", bound=tuple[int, int, int, str, int])
_T_maxver = TypeVar("_T_maxver", bound=tuple[int, int, int, str, int] | None)

class _Feature(Generic[_T_minver, _T_maxver]):
    ...
```

So that the runtime constants can be exposed at type checking time

AlexWaygood: I don't think that's necessary here tbh, and making classes generic in the stub when they're not at runtime can have some unfortunate consequences.

KotlinIsland: mypy doesn't narrow the tuple here.

https://mypy-play.net/?mypy=latest&python=3.10&gist=40fa9e86e067c31b64565ef816390e0a

AlexWaygood: Although it does narrow the type here: https://mypy-play.net/?mypy=latest&python=3.10&gist=d7701f4a0f78a3c0e2a2e514ca12d269

AlexWaygood (approved):

stdlib/__future__.pyi
KotlinIsland: @AlexWaygood it doesn't narrow to the the literal values

graingert: https://mypy-play.net/?mypy=latest&python=3.10&gist=83e0c037ef083c9bf4ec50c2b2f9d551&flags=strict

Works for me

KotlinIsland: When the variable is marked as `Final` it does narrow.

KotlinIsland: https://mypy-play.net/?mypy=latest&python=3.10&gist=7736b03b5d7a04b0ed930680e2234cfd

KotlinIsland: But it wouldn't hurt to make it generic, I'll do it now

graingert: https://mypy-play.net/?mypy=latest&python=3.10&gist=83e0c037ef083c9bf4ec50c2b2f9d551&flags=strict

Works for me if you specify the type in the assignment, note the stub won't have the concrete value, so you'd have to put the `tuple[Literal[2], ...` anyway

graingert: > and making classes generic in the stub when they're not at runtime can have some unfortunate consequences.

`_Feature` is private so it's not exposed right?

KotlinIsland: or not...

AlexWaygood: > `_Feature` is private so it's not exposed right?

It _is_ documented, though. And I still think in general we shouldn't make things generic in the stub when they're not at runtime unless there's a compelling use case  for it, which I don't really see here.

KotlinIsland:
Now have to make sure mypy works with it...

KotlinIsland:
Mypy should be fine, it doesn't look at the values.

stdlib/__future__.pyi
KotlinIsland: https://github.com/python/typing/issues/956

graingert: Could you do something like:

```python
class _Feature:
    ...

class _StubOnlyFeature(_Feature, Generic[...]):
    ...
```

AlexWaygood: > Could you do something like:
> 
> ```python
> class _Feature:
>     ...
> 
> class _StubOnlyFeature(_Feature, Generic[...]):
>     ...
> ```

I'm even less a fan of that ;)

I prefer making classes generic in the stub to introducing lies about what class certain objects are at runtime.

PR #8235, opened 2022-07-04
Type stubs for tqdm library

Gilthans:
While tqdm is very popular, it is not only not typed, but strips the types of sequences it receives due to the way it works.
This should resolve the issue for most users.

An issue to add type hints to tqdm has been around [since 2016](https://github.com/tqdm/tqdm/issues/260), but hasn't seen any progress.


AlexWaygood:
It's fine to use `Literal` in typeshed, you just have to import it from `typing_extensions` instead of `typing`

AlexWaygood (commented):
Thanks for the PR! We'll take a full look when the tests are passing (feel free to ask if you need a hand with any of them).

Please don't make unrelated changes to `CONTRIBUTING.md` in this PR, though. If you think that needs to be changed, please do that in its own PR. It'll make for a very confusing git history otherwise.

stubs/tqdm/tqdm/dask.pyi
```diff
@@ -0,0 +1,10 @@
+from _typeshed import Incomplete
+
+from dask.callbacks import Callback  # type: ignore
```
AlexWaygood, line 3: It's not currently possible to have third-party `py.typed` libraries as dependencies for typeshed stubs. Your best bet is to do something like this:

```python
from typing import Any
Callback: TypeAlias = Any # actually dark.callbacks.Callback
```

Gilthans: Thanks, I was wondering how to get around that!

AlexWaygood:
If the implementation breaks LSP by having an incompatible override, the best thing to do is to mimic the implementation exactly and add a `# type: ignore[override]` to silence mypy on the line in question

Gilthans:
Thanks for the feedback and tips, I've incorporated them!
There were failing tests in stubtest wherever there was an external dependency, so I added them to `stubtest_allowlist.txt`, I hope that's alright.

AlexWaygood:
> Thanks for the feedback and tips, I've incorporated them!
> There were failing tests in stubtest wherever there was an external dependency, so I added them to `stubtest_allowlist.txt`, I hope that's alright.

It looks like stubtest just needs to `pip install` a few more things before running the script. I think a better solution than allowlisting these modules is to add a `stubtest_requirements.txt` file to tell stubtest to install a few extra things. See here for an example: https://github.com/python/typeshed/blob/master/stubs/google-cloud-ndb/%40tests/requirements-stubtest.txt

AlexWaygood:
(You don't need to pin the requirements precisely -- the `google-cloud-ndb` stubs are a bit of an outlier in that regard!)

Gilthans:
I had to put `tqdm.contrib.discord` back to the allowlist, since the dependency has conflicts with tensorflow. Since it's a failry small, read-only [repo](https://github.com/b1naryth1ef/disco) I'm OK with not putting more effort into it, but if you have a suggestion I'll be happy to try something else

AlexWaygood:
> I had to put `tqdm.contrib.discord` back to the allowlist, since the dependency has conflicts with tensorflow. Since it's a failry small, read-only [repo](https://github.com/b1naryth1ef/disco) I'm OK with not putting more effort into it, but if you have a suggestion I'll be happy to try something else

Nope, that's totally cool -- thanks!

stubs/tqdm/tqdm/std.pyi
```diff
@@ -0,0 +1,272 @@
...
+        ascii: bool | str | None = ...,
+        disable: bool = ...,
+        unit: str = ...,
+        unit_scale: bool | float = ...,
+        dynamic_ncols: bool = ...,
+        smoothing: float = ...,
+        bar_format: str | None = ...,
+        initial: float = ...,
+        position: int | None = ...,
+        postfix: Mapping[object, object] | str | None = ...,
```
AlexWaygood, line 98: Another `Mapping[object, object]`

stubs/tqdm/tqdm/std.pyi
```diff
@@ -0,0 +1,272 @@
...
+        lock_args: tuple[bool | None, float | None] | tuple[bool | None] | None = ...,
+        nrows: int | None = ...,
+        colour: str | None = ...,
+        delay: float | None = ...,
+        gui: bool = ...,
+    ) -> None: ...
+    def __new__(cls, *_, **__): ...
+    @classmethod
+    def write(
+        cls, s: str, file: str | io.TextIOWrapper | io.StringIO | None = ..., end: str = ..., nolock: bool = ...
```
AlexWaygood, line 110: Again, the only required thing of the `file` parameter here is that the object passed in has a `write()` method:

```suggestion
        cls, s: str, file: SupportsWrite[str] | None = ..., end: str = ..., nolock: bool = ...
```

https://github.com/tqdm/tqdm/blob/4f208e72552c4d916aa4fe6a955349ee8b2ed353/tqdm/std.py#L613-L620

stubs/tqdm/tqdm/std.pyi
```diff
@@ -0,0 +1,272 @@
...
+    ) -> None: ...
+    @classmethod
+    def external_write_mode(
+        cls, file: str | io.TextIOWrapper | io.StringIO | None = ..., nolock: bool = ...
+    ) -> AbstractContextManager[None]: ...
```
AlexWaygood, line 115: `file` here should probably also be `SupportsWrite[str]`. Also, we prefer precise return types in typeshed, so it would be better to return `contextlib._GeneratorContextManager[None]` rather than `AbstractContextManager[None]`.

stubs/tqdm/tqdm/std.pyi
```diff
@@ -0,0 +1,272 @@
...
+    last_print_t: Incomplete
+    start_t: Incomplete
+
+    def __bool__(self): ...
+    def __nonzero__(self): ...
+    def __len__(self): ...
+    def __reversed__(self): ...
+    def __contains__(self, item) -> bool: ...
+    def __enter__(self): ...
+    def __exit__(self, exc_type, exc_value, traceback) -> None: ...
```
AlexWaygood, line 188:
```suggestion
    def __bool__(self) -> bool: ...
    def __nonzero__(self) -> bool: ...
    def __len__(self) -> int: ...
    def __reversed__(self) -> Iterator[_T]: ...
    def __contains__(self, item: object) -> bool: ...
    def __enter__(self: Self) -> Self: ...
    def __exit__(self, exc_type: object, exc_value: object, traceback: object) -> None: ...
```

stubs/tqdm/tqdm/std.pyi
```diff
@@ -0,0 +1,272 @@
...
+    def close(self) -> None: ...
+    def clear(self, nolock: bool = ...) -> None: ...
+    def refresh(
+        self, nolock: bool = ..., lock_args: tuple[bool | None, float | None] | tuple[bool | None] | None = ...
+    ) -> None: ...
+    def unpause(self) -> None: ...
+    def reset(self, total: float | None = ...) -> None: ...
+    def set_description(self, desc: str | None = ..., refresh: bool | None = ...) -> None: ...
+    def set_description_str(self, desc: str | None = ..., refresh: bool | None = ...) -> None: ...
+    def set_postfix(self, ordered_dict: Mapping[object, object] | None = ..., refresh: bool | None = ..., **kwargs) -> None: ...
```
AlexWaygood, line 202: Another `Mapping[object, object]`

stubs/tqdm/tqdm/std.pyi
```diff
@@ -0,0 +1,272 @@
...
+    def set_postfix_str(self, s: str = ..., refresh: bool = ...) -> None: ...
+    def moveto(self, n) -> None: ...
+    @property
+    def format_dict(self): ...
```
AlexWaygood, line 206:
```suggestion
    def format_dict(self) -> MutableMapping[str, Any]: ...
```

https://github.com/tqdm/tqdm/blob/4f208e72552c4d916aa4fe6a955349ee8b2ed353/tqdm/std.py#L1460-L1476

stubs/tqdm/tqdm/std.pyi
```diff
@@ -0,0 +1,272 @@
...
+    ascii: bool | str | None = ...,
+    disable: bool = ...,
+    unit: str = ...,
+    unit_scale: bool | float = ...,
+    dynamic_ncols: bool = ...,
+    smoothing: float = ...,
+    bar_format: str | None = ...,
+    initial: float = ...,
+    position: int | None = ...,
+    postfix: Mapping[object, object] | str | None = ...,
```
AlexWaygood, line 236: Another `Mapping[object, object]`

stubs/tqdm/tqdm/std.pyi
```diff
@@ -0,0 +1,272 @@
...
```
AlexWaygood, line 265: Another `Mapping[object, object]`

stubs/tqdm/tqdm/utils.pyi
```diff
@@ -0,0 +1,44 @@
+from _typeshed import Incomplete
+
+CUR_OS: Incomplete
+IS_WIN: Incomplete
+IS_NIX: Incomplete
+RE_ANSI: Incomplete
```
AlexWaygood, line 6:
```suggestion
from _typeshed import Incomplete
from typing import Pattern

CUR_OS: str
IS_WIN: bool
IS_NIX: bool
RE_ANSI: Pattern[str]
```

stubs/tqdm/tqdm/utils.pyi
```diff
@@ -0,0 +1,44 @@
...
+
+class FormatReplace:
+    replace: Incomplete
+    format_called: int
+    def __init__(self, replace: str = ...) -> None: ...
+    def __format__(self, _): ...
```
AlexWaygood, line 12:
```suggestion
    replace: str
    format_called: int
    def __init__(self, replace: str = ...) -> None: ...
    def __format__(self, _) -> str: ...
```

stubs/tqdm/tqdm/utils.pyi
```diff
@@ -0,0 +1,44 @@
...
+
+class Comparable:
+    def __lt__(self, other): ...
+    def __le__(self, other): ...
+    def __eq__(self, other): ...
+    def __ne__(self, other): ...
+    def __gt__(self, other): ...
+    def __ge__(self, other): ...
```
AlexWaygood, line 20:
```suggestion
    def __lt__(self, other) -> bool: ...
    def __le__(self, other) -> bool: ...
    def __eq__(self, other) -> bool: ...
    def __ne__(self, other) -> bool: ...
    def __gt__(self, other) -> bool: ...
    def __ge__(self, other) -> bool: ...
```

stubs/tqdm/tqdm/utils.pyi
```diff
@@ -0,0 +1,44 @@
...
+
+class ObjectWrapper:
+    def __getattr__(self, name): ...
+    def __setattr__(self, name, value): ...
```
AlexWaygood, line 24:
```suggestion
    def __getattr__(self, name: str): ...
    def __setattr__(self, name: str, value): ...
```

stubs/tqdm/tqdm/utils.pyi
```diff
@@ -0,0 +1,44 @@
...
+    def wrapper_getattr(self, name): ...
+    def wrapper_setattr(self, name, value): ...
+    def __init__(self, wrapped) -> None: ...
+
+class SimpleTextIOWrapper(ObjectWrapper):
+    def __init__(self, wrapped, encoding) -> None: ...
+    def write(self, s): ...
+    def __eq__(self, other): ...
```
AlexWaygood, line 32:
```suggestion
    def write(self, s: str): ...
    def __eq__(self, other: object) -> bool: ...
```

stubs/tqdm/tqdm/utils.pyi
```diff
@@ -0,0 +1,44 @@
...
+
+class DisableOnWriteError(ObjectWrapper):
+    @staticmethod
+    def disable_on_exception(tqdm_instance, func): ...
```
AlexWaygood, line 36:
```suggestion

_P = ParamSpec("_P")
_R = TypeVar("_R")

class DisableOnWriteError(ObjectWrapper):
    @staticmethod
    def disable_on_exception(tqdm_instance, func: Callable[_P, _R]) -> Callable[_P, _R]: ...
```

(You'll need to import `TypeVar` from `typing`, `ParamSpec` from `typing_extensions` and `Callable` from `collections.abc`)

stubs/tqdm/tqdm/utils.pyi
```diff
@@ -0,0 +1,44 @@
...
+    def __init__(self, wrapped, tqdm_instance) -> None: ...
+    def __eq__(self, other): ...
```
AlexWaygood, line 38:
```suggestion
    def __eq__(self, other: object) -> bool: ...
```

stubs/tqdm/tqdm/utils.pyi
```diff
@@ -0,0 +1,44 @@
...
+
+class CallbackIOWrapper(ObjectWrapper):
+    def __init__(self, callback, stream, method: str = ...): ...
```
AlexWaygood, line 41:
```suggestion
    def __init__(self, callback: Callable[[int], object], stream, method: str = ...) -> None: ...
```

stubs/tqdm/tqdm/utils.pyi
```diff
@@ -0,0 +1,44 @@
...
+
+def disp_len(data): ...
+def disp_trim(data, length): ...
```
AlexWaygood, line 44:
```suggestion
def disp_len(data: str) -> int: ...
def disp_trim(data: str, length: int) -> str: ...
```

AlexWaygood (changes requested):
This is a pretty big library to contribute, so thanks for all the work done here!

Here's a first pass at reviewing -- sorry about the number of comments! Most are pretty minor.

I'll do another pass after these have been addressed. I haven't looked at the `mypy_primer` report yet.

stubs/tqdm/tqdm/notebook.pyi
```diff
@@ -0,0 +1,93 @@
...
+        ascii: bool | str | None = ...,
+        disable: bool = ...,
+        unit: str = ...,
+        unit_scale: bool | float = ...,
+        dynamic_ncols: bool = ...,
+        smoothing: float = ...,
+        bar_format: str | None = ...,
+        initial: float = ...,
+        position: int | None = ...,
+        postfix: Mapping[object, object] | str | None = ...,
```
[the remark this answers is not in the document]
AlexWaygood: This is marked as resolved but doesn't seem to be fixed

stubs/tqdm/tqdm/notebook.pyi
```diff
@@ -0,0 +1,93 @@
...
```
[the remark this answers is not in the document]
AlexWaygood: Also unresolved

stubs/tqdm/tqdm/dask.pyi
```diff
@@ -0,0 +1,12 @@
+from _typeshed import Incomplete
+from typing import Any
+from typing_extensions import TypeAlias
+
+_Callback: TypeAlias = Any  # Actually dask.callbacks.Callback
+
+class TqdmCallback(_Callback):
+    tqdm_class: Incomplete
+    def __init__(
+        self, start: Incomplete | None = ..., pretask: Incomplete | None = ..., tqdm_class=..., **tqdm_kwargs
```
AlexWaygood, line 10:
```suggestion
    tqdm_class: type[Incomplete]
    def __init__(
        self, start: Incomplete | None = ..., pretask: Incomplete | None = ..., tqdm_class: type[Incomplete] = ..., **tqdm_kwargs
```

stubs/tqdm/tqdm/gui.pyi
```diff
@@ -0,0 +1,28 @@
...
+    hspan: Incomplete
+    wasion: Incomplete
+    ax: Incomplete
+    def __init__(self, *args, **kwargs) -> None: ...
+    disable: bool
+    def close(self) -> None: ...
+    def clear(self, *_, **__) -> None: ...
+    def display(self, *_, **__) -> None: ...
+
+def tgrange(*args, **kwargs): ...
```
AlexWaygood, line 25:
```suggestion
def tgrange(*args, **kwargs) -> tqdm_gui[Incomplete]: ...
```

stubs/tqdm/tqdm/keras.pyi
```diff
@@ -0,0 +1,29 @@
...
+    on_batch_end: Incomplete
+    def __init__(
+        self,
+        epochs: Incomplete | None = ...,
+        data_size: Incomplete | None = ...,
+        batch_size: Incomplete | None = ...,
+        verbose: int = ...,
+        tqdm_class=...,
+        **tqdm_kwargs,
+    ): ...
```
AlexWaygood, line 25:
```suggestion
    ) -> None: ...
```

stubs/tqdm/tqdm/keras.pyi
```diff
@@ -0,0 +1,29 @@
...
+    def on_train_begin(self, *_, **__) -> None: ...
+    def on_epoch_begin(self, epoch, *_, **__): ...
```
AlexWaygood, line 27:
```suggestion
    def on_epoch_begin(self, epoch, *_, **__) -> None: ...
```

stubs/tqdm/tqdm/notebook.pyi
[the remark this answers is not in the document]
AlexWaygood:
```suggestion
        postfix: Mapping[str, object] | str | None = ...,
```

stubs/tqdm/tqdm/notebook.pyi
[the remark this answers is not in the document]
AlexWaygood:
```suggestion
        postfix: Mapping[str, object] | str | None = ...,
```

stubs/tqdm/tqdm/rich.pyi
```diff
@@ -0,0 +1,36 @@
...
+_T = TypeVar("_T")
+
+class tqdm_rich(Generic[_T], std_tqdm[_T]):
+    def __init__(self, *args, **kwargs) -> None: ...
+    def close(self) -> None: ...
+    def clear(self, *_, **__) -> None: ...
+    def display(self, *_, **__) -> None: ...
+    def reset(self, total: Incomplete | None = ...) -> None: ...
+
+def trrange(*args, **kwargs): ...
```
AlexWaygood, line 33:
```suggestion
def trrange(*args, **kwargs) -> tqdm_rich[Incomplete]: ...
```

stubs/tqdm/tqdm/std.pyi
```diff
@@ -0,0 +1,270 @@
...
+        postfix: Mapping[str, object] | str | None = ...,
+        unit_divisor: float = ...,
+        write_bytes: bool | None = ...,
+        lock_args: tuple[bool | None, float | None] | tuple[bool | None] | None = ...,
+        nrows: int | None = ...,
+        colour: str | None = ...,
+        delay: float | None = ...,
+        gui: bool = ...,
+    ) -> None: ...
+    def __new__(cls, *_, **__): ...
```
AlexWaygood, line 107:
```suggestion
    def __new__(cls: type[Self], *_, **__) -> Self: ...
```

stubs/tqdm/tqdm/std.pyi
```diff
@@ -0,0 +1,270 @@
...
+        initial: float = ...,
+        position: int | None = ...,
+        postfix: Mapping[str, object] | str | None = ...,
+        unit_divisor: float = ...,
+        write_bytes: bool | None = ...,
+        lock_args: tuple[bool | None, float | None] | tuple[bool | None] | None = ...,
+        nrows: int | None = ...,
+        colour: str | None = ...,
+        delay: float | None = ...,
+    ): ...
```
AlexWaygood, line 146:
```suggestion
    ) -> None: ...
```

stubs/tqdm/tqdm/std.pyi
```diff
@@ -0,0 +1,270 @@
...
+    def set_postfix(self, ordered_dict: Mapping[str, object] | None = ..., refresh: bool | None = ..., **kwargs) -> None: ...
+    def set_postfix_str(self, s: str = ..., refresh: bool = ...) -> None: ...
+    def moveto(self, n) -> None: ...
+    @property
+    def format_dict(self) -> MutableMapping[str, Any]: ...
+    def display(self, msg: str | None = ..., pos: int | None = ...) -> None: ...
+    @classmethod
+    def wrapattr(
+        cls, stream, method: Literal["read", "write"], total: float | None = ..., bytes: bool | None = ..., **tqdm_kwargs
+    ) -> AbstractContextManager[Incomplete]: ...
```
AlexWaygood, line 209:
```suggestion
    ) -> contextlib._GeneratorContextManager[Incomplete]: ...
```

stubs/tqdm/tqdm/tk.pyi
```diff
@@ -0,0 +1,23 @@
...
+    def close(self) -> None: ...
+    def clear(self, *_, **__) -> None: ...
+    def display(self, *_, **__) -> None: ...
+    def set_description(self, desc: str | None = ..., refresh: bool | None = ...) -> None: ...
+    desc: Incomplete
+    def set_description_str(self, desc: str | None = ..., refresh: bool | None = ...) -> None: ...
+    def cancel(self) -> None: ...
+    def reset(self, total: Incomplete | None = ...) -> None: ...
+
+def ttkrange(*args, **kwargs): ...
```
AlexWaygood, line 20:
```suggestion
def ttkrange(*args, **kwargs) -> tqdm_tk[Incomplete]: ...
```

AlexWaygood (changes requested):
Second pass. We're nearly there.

stubs/tqdm/tqdm/tk.pyi
```diff
@@ -0,0 +1,23 @@
+from _typeshed import Incomplete
+from typing import Generic, TypeVar
+
+from .std import tqdm as std_tqdm
+
+_T = TypeVar("_T")
+
+class tqdm_tk(Generic[_T], std_tqdm[_T]):
+    def __init__(self, *args, **kwargs): ...
```
AlexWaygood, line 9:
```suggestion
    def __init__(self, *args, **kwargs) -> None: ...
```

stubs/tqdm/tqdm/utils.pyi
```diff
@@ -0,0 +1,49 @@
...
+    def __lt__(self, other) -> bool: ...
+    def __le__(self, other) -> bool: ...
+    def __eq__(self, other) -> bool: ...
+    def __ne__(self, other) -> bool: ...
+    def __gt__(self, other) -> bool: ...
+    def __ge__(self, other) -> bool: ...
+
+class ObjectWrapper:
+    def __getattr__(self, name: str): ...
+    def __setattr__(self, name: str, value): ...
```
AlexWaygood, line 26:
```suggestion
    def __setattr__(self, name: str, value) -> None: ...
```

stubs/tqdm/tqdm/contrib/utils_worker.pyi
```diff
@@ -0,0 +1,15 @@
...
+from concurrent.futures import Future, ThreadPoolExecutor
+from typing import TypeVar
+from typing_extensions import ParamSpec
+
+_P = ParamSpec("_P")
+_R = TypeVar("_R")
+
+class MonoWorker:
+    pool: ThreadPoolExecutor
+    futures: deque[Incomplete]
```
AlexWaygood, line 13:
```suggestion
    futures: deque[Future[Incomplete]]
```

AlexWaygood:
(I might need to iterate a little bit to tackle these primer errors)

AlexWaygood (approved):
Awesome, thanks for your patience! The remaining mypy_primer hits all look like true positives to me (except the `pyppeteer` one, which I can't make head or tail of... but you can't catch them all `¯\_(ツ)_/¯`)

JelleZijlstra:
Weird, https://github.com/pyppeteer/pyppeteer/blob/f98f9030d3d40dc629bf08f93672f05caf8dfb0e/pyppeteer/chromium_downloader.py#L94 doesn't have a `tqdm` line. The next line does, but there's no `file` argument. Is mypy-primer looking at a different version of the library? The file hasn't changed since January.

AlexWaygood:
> Weird, https://github.com/pyppeteer/pyppeteer/blob/f98f9030d3d40dc629bf08f93672f05caf8dfb0e/pyppeteer/chromium_downloader.py#L94 doesn't have a `tqdm` line. The next line does, but there's no `file` argument. Is mypy-primer looking at a different version of the library? The file hasn't changed since January.

Yeah, I have no idea. Weirdly, the default branch for `pyppeteer` seems to be `dev` rather than `master` — but the error doesn't make sense on `master` either.

JelleZijlstra:
There's a hardcoded revision: https://github.com/hauntsaninja/mypy_primer/blob/d91de1f533963a41d362531147c2ff407c286433/mypy_primer.py#L1431

This is the version it's checking: https://github.com/pyppeteer/pyppeteer/blob/2d27bfdf9b6d0df32e5ebe869b057409042417a8/pyppeteer/chromium_downloader.py#L94
It's passing `os.devnull` which is indeed a `str`, not a `SupportsWrite[str]`. I don't know whether that code is just broken.

AlexWaygood:
> There's a hardcoded revision: https://github.com/hauntsaninja/mypy_primer/blob/d91de1f533963a41d362531147c2ff407c286433/mypy_primer.py#L1431
> 
> This is the version it's checking: https://github.com/pyppeteer/pyppeteer/blob/2d27bfdf9b6d0df32e5ebe869b057409042417a8/pyppeteer/chromium_downloader.py#L94 It's passing `os.devnull` which is indeed a `str`, not a `SupportsWrite[str]`. I don't know whether that code is just broken.

Hmm, seems like primer's been hardcoded to that revision for 2 years now. I wonder if `tqdm` has changed in the meantime so that `str` is no longer good to be passed to the `file` parameter. It doesn't look to me like `file` is meant to be a `str` (but I haven't tried it, just read the source code)

alkatar21:
@AlexWaygood  I do not know if I should create a new issue or if it will be seen here. Anyway, under `tqdm.contrib.logging` are 2 context managers that are currently annotated as iterator and therefore generate the following messages:
```py
    with logging_redirect_tqdm():
        pass
```
```
"Iterator[None]" has no attribute "__enter__"; maybe "__iter__"?  [attr-defined] mypy(error)
"Iterator[None]" has no attribute "__exit__"; maybe "__next__"?  [attr-defined] mypy(error)
Object of type "Iterator[None]" cannot be used with "with" because it does not implement __enter__ Pylance[reportGeneralTypeIssues]
Object of type "Iterator[None]" cannot be used with "with" because it does not implement __exit__ Pylance[reportGeneralTypeIssues]
```

AlexWaygood:
@alkatar21 thanks! I've filed #8251.

Issue #5768, opened 2021-07-13
Allow non-types dependencies


srittau:
This was previously discussed in #5618 and came up in #5765. I think the consensus is that we should allow this. This issue is mainly to track this.

intgr:
Copying relevant discussion from #5618 for context:

[srittau commented on Jun 18](https://github.com/python/typeshed/issues/5618#issuecomment-863559855)

> Not speaking about stubs packages in general, but I think it's reasonable if typeshed packages depend on dependencies of the corresponding upstream package. I.e. if package `foo` depends on `cryptography`, it's fine that our `foo-types` depends on `cryptography` as well, since it doesn't introduce "new" dependencies. Apart from that we should vet dependencies carefully.

[JukkaL commented on Jun 18](https://github.com/python/typeshed/issues/5618#issuecomment-863937821)

> I agree with srittau.
> 
> I think that it would be sufficient to maintain an allowlist of vetted dependencies. I'd expect that the number of such dependencies will be fairly small, so it shouldn't be much effort to maintain.

hmc-cs-mdrissi:
I'm running into this issue for [tensorflow stubs](https://github.com/python/typeshed/pull/7319). Most tensorflow type annotations depend on a couple basic numpy types. So at moment my options look like don't include numpy types and have most annotations be too strict or have Any almost everywhere. 

What's needed to work on this task? I can see one issue is pyright check produces errors from dependency not being installed.

Akuli:
If I understand correctly, the next steps are:
- Add a check to `tests/check_consistent.py` that fails if `stubs/foo` depends on anything else than:
    - A `types-bar` package where `bar` is so that `stubs/bar` exists in typeshed
    - Or a dependency of `foo` on pypi that has `py.typed`
        - We can use pypi's API for this, although checking the presence of `py.typed` with just the API is tricky. Can we just assume that `py.typed` exists, or do we need to potentially download the sdist to check? How about packages that don't provide sdist?
- Make a pull request that changes one stubs-package to depend on upstream types
- Fix pull request until CI is green
- Before merging, run the third-party stubs check in `daily.yml` passes (this can be ran manually)

A potential problem is if packages `foo` and `bar` have dependencies that cannot be installed simultaneously. Then `daily.yml` will not work at all. I think we should fix this later, in a separate issue and a separate PR, because users are already complaining about not being able to use upstream type hints, and this issue won't come up until we have many non-stubs dependencies in typeshed.

[...]

Issue #5618, opened 2021-06-11
Remove cryptography (depends on #5952)


srittau:
`cryptography` added `py.typed` in [release 3.4.4](https://github.com/pyca/cryptography/blob/main/CHANGELOG.rst#344---2021-02-09), released 2021-02-09.

hauntsaninja:
One fun and different thing about types-cryptography is that other typeshed packages depend on it: paramiko, PyJWT, openssl-python.

So we will need to ensure stub packages can depend on non-typeshed-stub-packages.

hauntsaninja:
Just in case, does anyone feel strongly against allowing stub packages from depending on real PEP 561 packages? cc @JukkaL

srittau:
What do you mean by "real" PEP 561 packages?

Akuli:
Perhaps packages that contain `.py` files and a `py.typed` file.

hauntsaninja:
Yup, I started with "real packages", clarified to "PEP 561 packages", but missed editing out the "real".

srittau:
Not speaking about stubs packages in general, but I think it's reasonable if typeshed packages depend on dependencies of the corresponding upstream package. I.e. if package `foo` depends on `cryptography`, it's fine that our `foo-types` depends on `cryptography` as well, since it doesn't introduce "new" dependencies. Apart from that we should vet dependencies carefully.

JukkaL:
I agree with @srittau. 

I think that it would be sufficient to maintain an allowlist of vetted dependencies. I'd expect that the number of such dependencies will be fairly small, so it shouldn't be much effort to maintain.

srittau:
Blocked by #5952.

Akuli:
What is the current status of this? Can we now make other packages depend on `cryptography`?

PR #8239, opened 2022-07-05
Fix cryptography WithSerialization aliases

hasier:
Align with changes in https://github.com/pyca/cryptography/commit/6a8c0b55b99f95e1fcc8bb9c2b03fd9158d2081e, still present in [latest commit in main](https://github.com/pyca/cryptography/tree/86d9e39db80726e2dabd55ddde1699cc7b689374/src/cryptography/hazmat/primitives/asymmetric).

### DIFF
--- a/stubs/cryptography/cryptography/hazmat/primitives/asymmetric/dsa.pyi
+++ b/stubs/cryptography/cryptography/hazmat/primitives/asymmetric/dsa.pyi
@@ -34,14 +34,14 @@ class DSAPrivateKey(metaclass=ABCMeta):
     def public_key(self) -> DSAPublicKey: ...
     @abstractmethod
     def sign(self, data: bytes, algorithm: HashAlgorithm | Prehashed) -> bytes: ...
-
-class DSAPrivateKeyWithSerialization(DSAPrivateKey):
     @abstractmethod
     def private_bytes(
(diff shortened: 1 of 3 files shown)


AlexWaygood:
Hi, @hasier -- thanks for the PR!

I think stubtest is probably failing because the `version` field in the `METADATA.toml` file for these stubs is pinned to 3.3: https://github.com/python/typeshed/blob/592707bb97cb8e36220b5dfcb53a7c1a8a91d998/stubs/cryptography/METADATA.toml#L1. This is because `cryptography` has shipped with inline type annotations for over a year now, so the stubs typeshed ships are now obsolete -- you should use the type hints provided by the library instead.

Usually in a situation like this, we would have deleted the stubs from typeshed by now. However, in this case things are a little more complicated due to the fact that other stubs packages in typeshed depend on these stubs (#5618).

If you'd like to help out with typeshed, a really great contribution to make would be to put some work into tweaking our testing infrastructure so that we can include non-stub packages as dependencies for typeshed stubs. Once we're able to do that, we'll be able to remove these out-of-date `cryptography` stubs (#5768).

For now, I think it's probably not worth putting the effort into improving these long-obsolete stubs, sadly :(

hasier:
Ah I see. I had not realised they already had inline annotations because we are not in the most recent version of `cryptography`, so some of them were missing 🤦  I'll close this PR then, thanks for explaining @AlexWaygood!

PR #6732, opened 2021-12-29
Create dependabot.yml

sobolevn:
Docs: https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/configuration-options-for-dependency-updates

It is pretty much the same file CPython has: https://github.com/python/cpython/blob/main/.github/dependabot.yml

But, I think "weekly" check is better than "monthly": not too frequent, not too late.
We also don't use labels in this repo.

PR #8240, opened 2022-07-05
Add dependabot

wrt54g:
I added dependabot for Actions.


srittau:
I don't have good experiences with dependabot for actions as it will always pin an exact actions version, instead of just a major version, which makes no sense, at least for GitHub-created actions.

AlexWaygood:
I agree with @srittau and @hauntsaninja on this, and it was already discussed in #6732. Thanks, but dependabot just doesn't seem like it would be particularly helpful for typeshed.

PR #8243, opened 2022-07-06
Improve `tkinter.simpledialog`

TeamSpen210:
This adds the remaining methods to `simpledialog.Dialog`, which are intended to be overridden by subclass code. I also replaced `**kwargs` in the functions with the specific keywords that `_QueryDialog` accepts.


stdlib/tkinter/simpledialog.pyi
```diff
@@ -1,10 +1,13 @@
...
-from typing import Any
+from tkinter import Event, Frame, Misc, Toplevel
 
 class Dialog(Toplevel):
     def __init__(self, parent: Misc | None, title: str | None = ...) -> None: ...
-    def body(self, master) -> None: ...
+    def body(self, master: Frame) -> Misc | None: ...
     def buttonbox(self) -> None: ...
+    def ok(self, event: Event[Misc] | None = ...) -> None: ...
+    def cancel(self, event: Event[Misc] | None = ...) -> None: ...
```
Akuli, line 8: The events come from `self.bind(...)`, so their widget is always the type of `Self` (so a `Dialog` or some subclass of `Dialog`):

```suggestion
    def ok(self: Self, event: Event[Self] | None = ...) -> None: ...
    def cancel(self: Self, event: Event[Self] | None = ...) -> None: ...
```

You will need to import `Self` from `_typeshed`.

stdlib/tkinter/simpledialog.pyi
```diff
@@ -22,6 +25,30 @@ class SimpleDialog:
...
+    minvalue: float | None = ...,
+    maxvalue: float | None = ...,
+    parent: Misc | None = ...,
+) -> float | None: ...
+def askinteger(
+    title: str | None,
+    prompt: str,
+    *,
+    initialvalue: int | None = ...,
+    minvalue: float | None = ...,  # Any object that can compare to int.
```
Akuli, line 42: It might be better to require `int`, because the value is also used in error messages, and displaying a float in an error message would be confusing to users.

stdlib/tkinter/simpledialog.pyi
```diff
@@ -22,6 +25,30 @@ class SimpleDialog:
...
+    maxvalue: float | None = ...,
+    parent: Misc | None = ...,
+) -> int | None: ...
+def askstring(
+    title: str | None,
+    prompt: str,
+    *,
+    initialvalue: str | None = ...,
+    minvalue: str | None = ...,
+    maxvalue: str | None = ...,
```
Akuli, line 52: - Why would you want a string to be "between" two other strings? That doesn't make sense to me. I think passing in `minvalue` or `maxvalue` to `askstring()` is more likely a bug than what you wanted to achieve.
- There's a `"show"` keyword argument: `show="*"` would be useful for asking a password, for example.

Akuli (commented):
Thanks! Sorry about taking so long to review this PR, my life has been quite busy for the past few days :)

stdlib/tkinter/simpledialog.pyi
TeamSpen210: I included min/maxvalue for completeness since they do work (implemented in the common dialog class), but it is indeed fairly useless.

stdlib/tkinter/simpledialog.pyi
TeamSpen210: I did consider constraining this further, but since the `buttonbox` method (adding the buttons) is intended to be overridden, in that case it seems reasonable that the user might want to bind these to other widgets, so it would no longer apply. The default implementations don't use the widget at all, and in either case you can still access `self` directly. I should have put a comment to explain that though.

Akuli (approved):
Thanks!

PR #8244, opened 2022-07-06
🐛 Fix type annotation for `traceback.StackSummary.from_list()` to include `FrameSummary` items

tiangolo:
🐛 Fix type annotation for `traceback.StackSummary.from_list()` to include `FrameSummary` items.

The current type annotation only expects tuples, but the docs say it supports `FrameSummary` items and mentions tuples as "old-style".

> Construct a [StackSummary](https://docs.python.org/3/library/traceback.html#traceback.StackSummary) object from a supplied list of [FrameSummary](https://docs.python.org/3/library/traceback.html#traceback.FrameSummary) objects or old-style list of tuples. Each tuple should be a 4-tuple with filename, lineno, name, line as the elements.

Rerf: https://docs.python.org/3/library/traceback.html#traceback.StackSummary.from_list

And the implementation checks each item to see if it's a `FrameSummary` or a tuple: https://github.com/python/cpython/blob/50b9a7762f06335277d9962edc8d39498601a4e4/Lib/traceback.py#L439-L456


AlexWaygood:
E.g. https://mypy-play.net/?mypy=latest&python=3.10&gist=f03667b3b56f05dd0bdb5122062052d5

tiangolo:
Ah! Good point, thank you @AlexWaygood! Fixed it :heavy_check_mark:

AlexWaygood (approved):
Awesome, thanks for the fix!

PR #8246, opened 2022-07-06
`tqdm`: Improve a few `__init__` methods

AlexWaygood:
Fixes a few issues I noticed after merging #8235. These `__init__` overrides in the subclasses currently have the effect of making type checkers unable to perform type inference for instances of these subclasses, unless we tediously copy/paste the whole overload stack like I propose in this PR. Some of the existing `__init__` overrides are also missing a few optional parameters that are present for the subclasses, but not the superclass.


AlexWaygood:
Planning on merging this in a few days, unless there are any objections, since I was the reviewer for the original PR adding `tqdm` stubs.

PR #8247, opened 2022-07-06
Delete some works of fiction

AlexWaygood:
These things all used to exist, but none of them exist on 3.6+


hauntsaninja (approved):
Look at all this _OldStyle code left _dangling

PR #8248, opened 2022-07-06
Fix various `TypeAlias` issues

AlexWaygood:
Refs https://github.com/PyCQA/flake8-pyi/pull/240


hauntsaninja (commented):
The reasoning behind removing the conditional type alias annotations isn't clear to me

AlexWaygood:
> The reasoning behind removing the conditional type alias annotations isn't clear to me

Yeah, tbh I'm slightly unsure for those ones about whether it's better to consider them as _type_ aliases or _variable_ aliases. Happy to revert those changes if you think it's better the way it is; don't really have a strong opinion there.

hauntsaninja (approved):
Thanks! These are simple and uncontroversial. I don't have a strong opinion on the removed ones, so if there is some benefit feel free to open another PR.

AlexWaygood:
> Thanks! These are simple and uncontroversial. I don't have a strong opinion on the removed ones, so if there is some benefit feel free to open another PR.

I thought there were some benefits, but I was playing around on mypy playground and couldn't find any reason to make the change in terms of actually improving the stub. It reduces the hits from https://github.com/PyCQA/flake8-pyi/pull/240, but I think that that approach probably isn't viable anyway, so I don't think that's a good reason to make those changes `¯\_(ツ)_/¯`

Issue #8214, opened 2022-06-30
Wrong return type for signal.default_int_handler


kumaraditya303:
The current method signature specifies `None` as the return type whereas it should be `Never`.

https://github.com/python/typeshed/blob/200891fed1f8c7f290cec3bdc90b35da95cb8260/stdlib/signal.pyi#L67-L68

Runtime behavior:

```py
import signal
import sys
signal.default_int_handler(1, sys._getframe(0))
```
Output:
```console
@kumaraditya303 ➜ /workspaces/cpython (specialize-noargs ✗) $ ./python main.py 
Traceback (most recent call last):
  File "/workspaces/cpython/main.py", line 3, in <module>
    signal.default_int_handler(1, sys._getframe(0))
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
KeyboardInterrupt


```

PR #8249, opened 2022-07-06
fix: Return `Never` rather than `None` from `signal.default_int_handler`

kkirsche:
Fix: #8214 

Fixes the type of `signal.default_int_handler` to return `Never` rather than `None` per https://github.com/python/cpython/blob/774ef28814d0d9d57ec813cb31b0a7af6c476127/Modules/signalmodule.c#L258

It appears to always throw the Keyboard exception.

The only question I had was whether `Never` was correct or `NoReturn` was correct

### DIFF
--- a/stdlib/signal.pyi
+++ b/stdlib/signal.pyi
@@ -4,7 +4,7 @@ from collections.abc import Callable, Iterable
 from enum import IntEnum
 from types import FrameType
 from typing import Any, Union
-from typing_extensions import Final, TypeAlias, final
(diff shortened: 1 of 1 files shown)


JelleZijlstra:
Thanks! Either `Never` or `NoReturn` is fine; see https://typing.readthedocs.io/en/latest/source/unreachable.html#never-and-noreturn.

kkirsche:
Thanks for the link to that explanation, that's really helpful :)

Issue #6189, opened 2021-10-22
Python 3.6 EOL (not before July 2022)


hauntsaninja:
We're coming close to Python 3.6 EOL. Last year we were pretty prompt about dropping support for Python 3.5, dropping support within a month or two of EOL. I think I'd like us to be a little slower this time round (hence my posting of this in advance of anyone making a PR).

My impression is that Python 3.6 is still widely used. pypistats a couple packages says on the order of 20%. It's also the default on Ubuntu 18.04. Waiting six months would be in line with our removal policy for third party stubs, so that seems like a number we could default to, but curious if people have opinions.

srittau:
I have updated the title with the "support until" date and added the "removal" label (which is not 100% appropriate, but close enough).

KotlinIsland:
@hauntsaninja Could you please link to the PyPI Stats page you mentioned in the OP, thanks.

hauntsaninja:
Oh I wasn't aware of https://pypistats.org/packages/__all__ , that's nice.

I have no idea which specific packages I looked at, but if you made me guess:
https://pypistats.org/packages/typing-extensions 
https://pypistats.org/packages/mypy
https://pypistats.org/packages/requests
https://pypistats.org/packages/setuptools
https://pypistats.org/packages/numpy

mypy has the least 3.6 usage of these at ~10%, others go up to ~20%

AlexWaygood:
I'm pinning this issue, since our tentative removal date is coming up in around a month now.

srittau:
As July is approaching fast: Are there any blockers for removing Python 3.6 support? I get the feeling that Python 3.6 support is becoming slightly annoying to maintain, so it will be a plus if we can remove it.

PR #8250, opened 2022-07-06
Drop support for Python 3.6, part 1

srittau:
* Document that typeshed support 3.7+.
* Don't run tests on Python 3.6.
* Remove Python 3.6 allow lists.
* Merge common allowlist items.

Part of #6189


srittau:
Part two will remove obsolete branches.

.github/workflows/daily.yml
```diff
@@ -49,7 +49,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: actions/setup-python@v4
         with:
-          python-version: 3.9
+          python-version: "3.9"
```
AlexWaygood, line 52: Any reason not to bump this to Python 3.10, while we're here?

srittau: I did at first, but didn't want to muddle the PR. I'm not sure there aren't any side effects.

AlexWaygood: Okay fair, let's leave it for now then

hauntsaninja (approved):
This looks good to me!

PR #8251, opened 2022-07-07
Improve `tqdm.contrib.logging` context managers

AlexWaygood:
Fixes https://github.com/python/typeshed/pull/8235#issuecomment-1177191204

### DIFF
--- a/stubs/tqdm/tqdm/contrib/logging.pyi
+++ b/stubs/tqdm/tqdm/contrib/logging.pyi
@@ -1,10 +1,19 @@
 import logging
-from collections.abc import Iterator, Sequence
-from typing import Any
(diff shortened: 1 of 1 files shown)


stubs/tqdm/tqdm/contrib/logging.pyi
```diff
@@ -1,10 +1,19 @@
 import logging
-from collections.abc import Iterator, Sequence
-from typing import Any
+from _typeshed import Incomplete
+from collections.abc import Callable, Iterator, Sequence
```
AlexWaygood, line 3:
```suggestion
from collections.abc import Callable, Sequence
```

srittau (approved):

PR #8252, opened 2022-07-07
Improve `SQLAlchemy` type aliases

AlexWaygood:
Refs https://github.com/PyCQA/flake8-pyi/pull/240


srittau (approved):
While all the type aliases imported from `mypy` exist at runtime, I don't think they need to be re-exported from `sqlalchemy.ext.mypy`.

AlexWaygood:
> While all the type aliases imported from `mypy` exist at runtime, I don't think they need to be re-exported from `sqlalchemy.ext.mypy`.

My thoughts too.

Issue #8261, opened 2022-07-09
No type-hint for `imaplib.Commands`


JohnVillalovos:
While attempting to add type-hinting and type-checking to https://github.com/mjs/imapclient discovered that `imaplib.Commands` does not have a type-hint.

This causes an error with `mypy` like this:
```
imapclient/imapclient.py:49: error: Module has no attribute "Commands"
```

PR #8262, opened 2022-07-09
stdlib(imaplib): add type-hint for `imaplib.Commands`

JohnVillalovos:
Add type-hints for `imaplib.Commands` which is defined here in
imaplib.py:
https://github.com/python/cpython/blob/6442a9dd212fa18343db21849cf05c0181662c1f/Lib/imaplib.py#L58-L102

Closes: #8261


hauntsaninja (approved):
Thank you!

PR #8263, opened 2022-07-09
Extend annotations for `psycopg2.sql`

lonvia:
Adds return types for all functions in `psycopg2.sql`.

Types as per https://github.com/psycopg/psycopg2/blob/3c58e96e1000ef60060fb8139687028cb274838d/lib/sql.py.


lonvia:
The errors in the zulib library are all due to automatic type deduction returning a type that is too narrow. Explicit type annotation for local variables would be needed. I suspect that zulib won't be the only ones running into that problem.

What is the established protocol in such a case?

stubs/psycopg2/psycopg2/sql.pyi
```diff
@@ -1,48 +1,49 @@
...
+    def string(self) -> str: ...
+    def as_string(self, context) -> str: ...
+    def format(self, *args, **kwargs) -> Composed: ...
+    def join(self, seq) -> Composed: ...
 
 class Identifier(Composable):
     def __init__(self, *strings) -> None: ...
     @property
-    def strings(self): ...
+    def strings(self) -> tuple[str]: ...
```
srittau, line 32: `tuple[str]` is a tuple of length one, therefore:

```suggestion
    def strings(self) -> tuple[str, ...]: ...
```

srittau (approved):
Thanks, I noticed one thing below. Regarding primer output: As the problem exists due to previously "untyped" (`Any` typed) stubs, it's fine if new "problems" are being pointed out. I wish mypy was a bit smarter in cases like this, though.

srittau (changes requested):
Sorry, I meant to request changes with my previous review.

stubs/psycopg2/psycopg2/sql.pyi
lonvia: Fixed.

srittau (approved):

PR #8265, opened 2022-07-09
sysconfig: Add `get_default_scheme` and `get_preferred_scheme`

q0w:
Should `get_preferred_scheme` take key arg as  `Literal["prefix", "home", "user"]`?


stdlib/sysconfig.pyi
```diff
@@ -14,12 +16,21 @@ __all__ = [
     "parse_config_h",
 ]
 
+if sys.version_info >= (3, 10):
+    __all__ += ["get_default_scheme", "get_preferred_scheme"]
+
```
AlexWaygood, line 21: The runtime doesn't include these in `__all__`, so we shouldn't include them in `__all__` in the stub either

```suggestion
```

stdlib/sysconfig.pyi
```diff
@@ -14,12 +16,21 @@ __all__ = [
...
 @overload
 def get_config_vars() -> dict[str, Any]: ...
 @overload
 def get_config_vars(arg: str, *args: str) -> list[Any]: ...
 def get_scheme_names() -> tuple[str, ...]: ...
+
+if sys.version_info >= (3, 10):
+    def get_default_scheme() -> str: ...
+    _Key: TypeAlias = Literal["prefix", "home", "user"]
+    def get_preferred_scheme(key: _Key) -> str: ...
```
AlexWaygood, line 32: Not sure I really see the value of using a type alias here, given that it's only used once

```suggestion
    def get_preferred_scheme(key: Literal["prefix", "home", "user"]) -> str: ...
```

AlexWaygood (changes requested):
Thanks! A few nits:

AlexWaygood (approved):
Thanks!

(In general, in the future, please avoid force-pushing with typeshed PRs — it interacts badly with GitHub's UI, making it hard to see what changed between commits. Everything is squashed into one commit before merging anyway, so there's no need to worry about a messy commit history.)

PR #8266, opened 2022-07-10
`int.denominator`, `int.imag`: More precise return types

AlexWaygood:


JelleZijlstra:
Would this fall under #7258? I can't imagine much real-world code being affected though.

AlexWaygood:
> Would this fall under #7258? I can't imagine much real-world code being affected though.

I'd argue "no", since the problems raised in that issue are mainly to do with class-level and module-level constants (https://github.com/python/typeshed/issues/7258#issuecomment-1047667279). So the risk seems small imo... but yeah, the benefits are also probably small as well.

I don't have a particularly strong opinion on this — I was just looking at `int` because of all the renewed talk about typing & numbers recently, and noticed that `fractions.Fraction` has a much more precise annotation here: https://github.com/python/typeshed/blob/29c17ffb47dfc8fa688e1357c73eb522e86dd955/stdlib/fractions.pyi#L151

srittau:
Please note that the moratorium is for literal constants, not for all literal values. Especially return types should be fairly uncontroversial.

srittau (approved):

PR #8271, opened 2022-07-11
Fix `asyncio.gather` regression

### LINKED ISSUE #8270: Calling `await asyncio.gather(*a)` returns empty tuple in Pyright
In the following code:

```
import asyncio
from typing import Awaitable, List


async def f(a: List[Awaitable[int]]):
    b = await asyncio.gather(*a)
    reveal_type(b)  # Type of "b" is "tuple[()]"
```

Pyright type checking now assumes that b is empty, leading to potential errors down the line.

Most likely related to this commit: https://github.com/python/typeshed/commit/0740d1c48e48ddd426ac8cb56572ebd12a3652b1

More context from the discussion in the Pyright issue tracker: https://github.com/microsoft/pyright/issues/3685

AlexWaygood:
Adding the empty-tuple overload caused major problems for pyright, and that overload only deals with an unlikely edge case anyway. Get rid of it, and replace the fallback overload with a more general overload.

Fixes #8270.


hauntsaninja:
Worth a regression test?

AlexWaygood:
The reason why this PR is not a simple reversion of https://github.com/python/typeshed/commit/0740d1c48e48ddd426ac8cb56572ebd12a3652b1 is that the added tests do not pass with a simple revert PR.

hauntsaninja (commented):
Thanks! The test looks good, and helps make changes here much more sane.

That said, it might be nice to get rid of the type ignores. What goes wrong with adding the empty overload at the bottom?
```
    ...
    @overload
     def gather(
         __coro_or_future1: _FutureLike[Any],
         __coro_or_future2: _FutureLike[Any],
         __coro_or_future3: _FutureLike[Any],
         __coro_or_future4: _FutureLike[Any],
         __coro_or_future5: _FutureLike[Any],
         __coro_or_future6: _FutureLike[Any],
         *coros_or_futures: _FutureLike[Any],
         return_exceptions: bool = ...,
     ) -> Future[list[Any]]: ...
     @overload
     def gather(*, return_exceptions: bool = ...) -> Future[tuple[()]]: ...
```

AlexWaygood:
> What goes wrong with adding the empty overload at the bottom?

That's what I tried at first: https://github.com/python/typeshed/pull/8271/files/a345b43e68b56359e4decead1a261e283f8081a4.

But the test cases failed: https://github.com/python/typeshed/runs/7299574031?check_suite_focus=true

It seems like pyright picks the empty overload for `asyncio.gather(*[...])` _even if_ that overload is the last overload!

> minor, but maybe the clearest thing to do would be to move the big overload to the top, since that matches what type checkers are doing when they type check `gather(*[...])`

Meh, not sure I agree that that would improve readability :-)

hauntsaninja (approved):
Hmm, not sure I understand what pyright is doing then, so you may want someone else to review this. Test case looks good though!

AlexWaygood:
> Hmm, not sure I understand what pyright is doing then, so you may want someone else to review this. Test case looks good though!

I _think_ I understand what pyright's doing, and the test cases make me confident that this fixes the regression, so I'm inclined to merge. If somebody figures a way to get all the test cases to pass without the `type ignore`s, they can always file another PR in the future. Does that sound okay?

hauntsaninja:
Wait, why is pyright doing what it's doing? @erictraut also mentioned moving the empty overload to the bottom as something that's expected to work in https://github.com/microsoft/pyright/issues/3685#issuecomment-1180777945

Yeah, fine to merge, I was just clarifying what my "approve" of the PR promised :-)

PR #8272, opened 2022-07-11
Drop support for Python 2

### LINKED ISSUE #7367: Dropping Python 2 support

AlexWaygood:
Closes #7367


JelleZijlstra (approved):

srittau (approved):

ManuelZ:
If I'm using a MyPy version that still supports Python 2, will it be able to download stubs from Typeshed without issues?

hauntsaninja:
mypy vendors a copy of typeshed. All currently released versions of mypy both a) support Python 2, b) vendor a copy of typeshed that supports Python 2. The next release of mypy will do neither.

mypy does let you use your own copy of typeshed using the `--custom-typeshed-dir` flag. Very few people use this, but in case this is what you were asking about, the answer is you won't be able to type check Python 2 code, but type checking Python 3 code should still work.

ManuelZ:
Thanks.

PR #8274, opened 2022-07-11
socketserver: Set StreamRequestHandler.connection to SSLSocket

### LINKED ISSUE #8264: StreamRequestHandler.connection should be SSLSocket?
I get this error using werkzeug.WSGIRequestHandler:
```shell
error: "socket" has no attribute "getpeercert" 
[attr-defined]
                peer_cert = self.connection.getpeercert(binary_form=True)
```

q0w:
Fixes #8264

### DIFF
--- a/stdlib/socketserver.pyi
+++ b/stdlib/socketserver.pyi
@@ -3,7 +3,6 @@ import types
 from _typeshed import Self
 from collections.abc import Callable
 from socket import socket as _socket
-from ssl import SSLSocket
 from typing import Any, BinaryIO, ClassVar, Union
 from typing_extensions import TypeAlias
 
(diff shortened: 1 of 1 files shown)


stdlib/socketserver.pyi
```diff
@@ -153,7 +154,7 @@ class StreamRequestHandler(BaseRequestHandler):
     wbufsize: ClassVar[int]  # undocumented
     timeout: ClassVar[float | None]  # undocumented
     disable_nagle_algorithm: ClassVar[bool]  # undocumented
-    connection: _socket  # undocumented
+    connection: SSLSocket  # undocumented
```
JelleZijlstra, line 157: In the code, I see `self.connection = self.request`, and we type `BaseRequestHandler.request` as Any. What guarantees that the socket is always an SSLSocket?

q0w: maybe `type[_socket]` then?

JelleZijlstra: I'm pretty sure it's an instance, not a class. Perhaps we should just use Any here too.

srittau: I really should look into preparing an `AnyOf` PEP, despite the resistance. There are so many instances where this would be useful.

AlexWaygood: Very happy to give it a proof-read (or help co-write) if you ever do :)

PR #8277, opened 2022-07-12
Import `Match` and `Pattern` from `re`, not `typing`

AlexWaygood:
This is possible now that we have dropped support for Python 3.6


srittau (approved):
🎉

PR #8278, opened 2022-07-12
Revert "add overload to tuple.__new__ to better express an empty tuple"

### LINKED ISSUE #8275: Recent change to `tuple` constructor breaks subclasses of tuples
A recent change to the `tuple.__new__` method in `builtins.pyi` introduced an overload:
```python
    @overload
    def __new__(cls) -> tuple[()]: ...
```

This overload is problematic for classes that derive from specialized tuples.

```python
class MyTuple(tuple[int, str, str]):
    ...

x = MyTuple()
reveal_type(x) # tuple[()] !!!!
```

Because of this overload, the wrong type is evaluated.

I think this overload should return `Self`, not `tuple[()]`. Admittedly, that doesn't produce the ideal type evaluation for `tuple()`. It produces `tuple[Any, ...]` rather than `tuple[()]`.

JelleZijlstra:
Reverts python/typeshed#7454

Fixes #8275


AlexWaygood (approved):
I find the variadic-tuple-subclass case much more clearly problematic than the specialised-tuple-subclass case. Looks good.

PR #8280, opened 2022-07-12
Use error codes for type ignores

srittau:
Disable reportSelfClsParameterName for pytype as this is out of typeshed's
control

Closes: #7497


stdlib/subprocess.pyi
```diff
@@ -92,7 +92,13 @@ class CompletedProcess(Generic[_T]):
     stdout: _T
     stderr: _T
     # type ignore on __init__ because the TypeVar can technically be unsolved, but see comment above
```
AlexWaygood, line 94:
```suggestion
    # pyright ignore on __init__ because the TypeVar can technically be unsolved, but see comment above
```

AlexWaygood (approved):
Nice

PR #8282, opened 2022-07-12
Re-enable SQLAlchemy mypy tests

srittau:


tests/mypy_test.py
```diff
@@ -414,10 +414,6 @@ def test_third_party_stubs(code: int, args: TestConfig) -> TestResults:
     files_checked = 0
 
     for distribution in sorted(os.listdir("stubs")):
-        if distribution in {"SQLAlchemy", "Flask-SQLAlchemy"}:
```
AlexWaygood, line 417: These still crash locally for me if I run `mypy_test.py` twice in quick succession. It's an issue with the mypy cache; it should be fixed in mypy 0.970 when that comes out.

srittau: Ah yes, I didn't try that. Same here.

srittau:
Actually, I'll just mark this as "defered" so we don't forget about it.

AlexWaygood (approved):
The crashes are gone on 0.971 🎉

PR #8283, opened 2022-07-12
os: Params for posix_fallocate and posix_fadvise are pos-only

JelleZijlstra:
https://github.com/python/cpython/blob/90a6e56e5663233fe986331ef3f10422fe57dcbc/Modules/posixmodule.c#L10845


srittau (approved):
LGTM, once the tests pass.

PR #8284, opened 2022-07-12
os: pos-only for get_terminal_size

JelleZijlstra:
https://github.com/python/cpython/blob/90a6e56e5663233fe986331ef3f10422fe57dcbc/Modules/posixmodule.c#L13200


srittau (approved):

srittau (approved):
LGMT, once the tests pass.

PR #8285, opened 2022-07-12
`fractions.pyi`: Correct dunder parameter names

AlexWaygood:
This wasn't possible prior to #8280, which disabled pyright's `reportSelfClsParameterName` check


test_cases/stdlib/builtins/test_tuple.py
```diff
@@ -1,5 +1,5 @@
-from typing_extensions import assert_type
 from typing import Tuple
+from typing_extensions import assert_type
```
JelleZijlstra, line 2: Weird, I wonder how this passed on my PR.

JelleZijlstra (approved):

test_cases/stdlib/builtins/test_tuple.py
AlexWaygood: It surprised me too :)

Issue #8286, opened 2022-07-13
enum auto must not be derived from IntFlag


ghpqans:
When using auto enumeration within an `class Example(enum.Enum): X = auto()`, the `auto()` is actually constructing an instance of `enum.auto`, which just has a default constructor without any extra arguments. However, when deriving `class auto(IntFlag)` in `enum.pyi` the stub actually has a constructor argument of type int, the value for the `IntFlag`. 
This contradicts the definition in the original `enum.py` impolementation.
As a matter of fact this leads to confusing type hints, e.g. `enum class auto(IntFlag)`.

srittau:
I agree that this is confusing an error prone. In the last few years we tend to align stubs as closely to the implementation as possible. In this case, I believe the best approach is for type checkers would be to special case `auto`.

That said, as a stop-gap we could try not to derive `auto` from `IntFlag` and copy `IntFlag`s field over, instead or alternatively to just override `auto.__init__()`.

srittau:
Also cf. #1331 for PR where the current implementation was added.

Ovsyanka83:
+ 1
When working with fastapi, for example, we tend to use a lot of string enums, and we would love to have native support for them and any other types that could possibly use auto.

PR #8287, opened 2022-07-13
Type enum.auto according to implementation

srittau:
Exploratory PR.

Cf. #8286


srittau:
3.11-dev seems to have switch Python versions. I'll make a separate PR addressing those issues.

srittau:
This looks promising.

stdlib/enum.pyi
```diff
@@ -193,14 +193,12 @@ class IntEnum(int, _IntEnumBase):
...
-    _value_: Any
-    @_magic_enum_attr
-    def value(self) -> Any: ...
-    def __new__(cls: type[Self]) -> Self: ...
+class auto:
+    value: Never
+    # This is a lie, auto() returns an auto instance. But auto instances
+    # are solely intended to be used inside the body of Enum sub-classes
+    # and get replaced with Enum instances at runtime.
+    def __new__(cls: type[Self]) -> Enum: ...  # type: ignore[misc]
```
AlexWaygood, line 201: There's no need to annotate `cls` in a `__new__` method if there isn't a type-dependency between the type of `cls` and the return type.

```suggestion
    def __new__(cls) -> Enum: ...
```

srittau:
I thought that mypy supports `__new__()`, but that doesn't seem to be the case:

```python
class X:
    def __new__(cls) -> int:  #  type: ignore[misc]
        return 43

reveal_type(X())  # Revealed type is "foo.X"
print(type(X()))  # <class 'int'>
```

AlexWaygood:
> I thought that mypy supports `__new__()`, but that doesn't seem to be the case:
> 
> ```python
> class X:
>     def __new__(cls) -> int:  #  type: ignore[misc]
>         return 43
> 
> reveal_type(X())  # Revealed type is "foo.X"
> print(type(X()))  # <class 'int'>
> ```

I think it only supports `__new__` if the `__new__` method returns a type that's a subtype of `cls` :/

We're never going to be able to fully express what happens at runtime here; it's too magical. It needs to be special-cased by the type checker.

What if, instead of subclassing `IntFlag`, we subclassed `Any`? That should get rid of the false-positive errors, and also hopefully fix the Pycharm issue.

srittau:
Subclassing `Any` has the disadvantage of losing all type checking. I'll try another approach first.

stdlib/enum.pyi
```diff
@@ -193,14 +193,11 @@ class IntEnum(int, _IntEnumBase):
...
-_auto_null: Any
-
-# subclassing IntFlag so it picks up all implemented base functions, best modeling behavior of enum.auto()
+# This is a lie, auto() does not derive from IntFlag. But auto instances
+# are solely intended to be used inside the body of Enum sub-classes
+# and get replaced with Enum instances at runtime.
 class auto(IntFlag):
-    _value_: Any
-    @_magic_enum_attr
-    def value(self) -> Any: ...
```
AlexWaygood, line 202: Why delete these?

AlexWaygood: `IntFlag` has them as `int`, though, which isn't correct for `auto`. The `_value_` and `value` attributes of enum members created using `auto()` depends on the return type of `_generate_next_value_`, which can be overriden in custom enum classes.

stdlib/enum.pyi
```diff
@@ -193,14 +193,11 @@ class IntEnum(int, _IntEnumBase):
 
 def unique(enumeration: _EnumerationT) -> _EnumerationT: ...
 
-_auto_null: Any
```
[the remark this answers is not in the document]
AlexWaygood: It seems like it's been there (both at runtime and in the stub) for a very long time, and I'd rather not run the risk of breaking somebody.

AlexWaygood (commented):
I'm happy to change the constructor if it helps out PyCharm, but I'm still not sure I see a strong rationale for the other changes. `enum.auto()` obviously needs to be heavily special-cased by type checkers, but I'm worried that these changes might lead to regressions in the medium term until that special-casing is added.

Maybe adding some test cases would help here? Then we could be confident in knowing whether this causes any regressions.

Also, are we sure that this fixes the PyCharm issue? Have we got any way of testing that?

Issue #2383, opened 2018-08-14
Adding lists of subtypes together


alanhdu:
Using `mypy==0.620` and Python 3.6.6, the following code:

```python
from typing import List, Optional

x: List[Optional[int]] = [1, 2, 3]
[None] + x
```

fails with:

```
a.py:4: error: Unsupported operand types for + ("List[None]" and "List[Optional[int]]")
```

This is presumably because lists are invariant w.r.t their generic parameter, and the declration requires a `List` of the same type

https://github.com/python/typeshed/blob/4e40b035c8337cddcf4f94fcd83a2f0d454488b6/stdlib/3/builtins.pyi#L619.

I don't know what the right solution here is -- can we just replace the `x: List[_T]` with `x: List[_T_co]` and make it covariant? Or doe we need another typevar `_U = TypeVar("U", bound=_T)` or something?

wy193777:
#2404 has been reverted. Do we have a solution for this issue?

tamuhey:
+1 from https://github.com/microsoft/pyright/issues/1777

Issue #8292, opened 2022-07-13
list.__add__ isn't permissive enough


Akuli:
Works at runtime, doesn't type check:

```python3
class Foo:
    def asd(self) -> int:
        return 1

class Bar:
    def asd(self) -> int:
        return 2

combined = [Foo()] + [Bar()]
for item in combined:
    print(item.asd())
```

I would expect `[Foo()] + [Bar()]` to produce `list[Foo | Bar]`.

Converting to tuples works as expected:

```python3
combined = tuple([Foo()]) + tuple([Bar()])
```

But invariance problems don't apply here, because we're constructing a new list that doesn't refer to the old lists in any way.

There has previously been discussion about how mypy doesn't understand `[Foo(), Bar()]`, but this is slightly different: I'm adding lists, not mixing types within different elements of a single list literal.

AlexWaygood:
Seems reasonable to change it to:

```python
_T = TypeVar("_T")
_S = TypeVar("_S")

class list(MutableSequence[_T]):
    def __add__(self, __x: list[_S]) -> list[_T | _S]:...
```

 We already do a similar thing with `dict.__or__`: https://github.com/python/typeshed/blob/82745be9fce70442f90b781d6cdd756b718ce86f/stdlib/builtins.pyi#L1051

JelleZijlstra:
This is trickier than it seems, see #3183 for the last attempt.

Akuli:
I'll try again. The last attempt was done mostly before `mypy_primer` was a thing, so this time we should at least know if something breaks.

PR #8293, opened 2022-07-13
Yet another attempt to fix list.__add__

Akuli:
#8292


stdlib/builtins.pyi
```diff
@@ -986,8 +986,12 @@ class list(MutableSequence[_T], Generic[_T]):
     @overload
     def __setitem__(self, __s: slice, __o: Iterable[_T]) -> None: ...
     def __delitem__(self, __i: SupportsIndex | slice) -> None: ...
+    # Overloading looks unnecessary, but is needed to work around complex mypy problems
+    @overload
     def __add__(self, __x: list[_T]) -> list[_T]: ...
-    def __iadd__(self: Self, __x: Iterable[_T]) -> Self: ...
+    @overload
+    def __add__(self, __x: list[_S]) -> list[_S | _T]: ...
+    def __iadd__(self: Self, __x: Iterable[_T]) -> Self: ...  # type: ignore
```
AlexWaygood, line 994:
```suggestion
    def __iadd__(self: Self, __x: Iterable[_T]) -> Self: ...  # type: ignore[misc]
```

AlexWaygood:
I like the look of that primer diff. It looks like there's a few regressions in there, but it also looks like it fixes more things than it breaks.

Akuli:
My thoughts on the regressions:

```
+ pandas/core/indexes/frozen.py:65: error: Incompatible types in assignment (expression has type "Callable[[FrozenList, Any], FrozenList]", base class "list" defined the type as overloaded function)  [assignment]
+ pandas/core/reshape/merge.py:1616: error: Cannot determine type of "__add__"  [has-type]
+ pandas/core/reshape/reshape.py:313: error: Cannot determine type of "__add__"  [has-type]
```

First error: A `list[Any]` subclass `FrozenList` does `__add__ = __iadd__ = union`, where `union` is defined as taking an implicitly-`Any`-typed argument and returning `FrozenList`. With this PR, the return type is no longer `list[Any]` in all cases, and maybe `list[Any]` and `list[Any | Foo]` are incompatible. But the pandas code is already quite `Any`-typed, so I think a `# type: ignore` is fine there.

Second error: `join_levels = join_levels + [restore_levels]` confuses mypy, because it causes the type of `join_levels` to depend on itself in a complicated way. I have seen this error message in the past when multiple types (often in multiple files) depend on each other in a cyclic way. (I wish the error message actually said what's wrong instead of making me memorize its meaning...) But I don't think this is a problem, because you can usually `.append()`, or just add an explicit annotation if you don't want to append, which is useful if you have another reference to the same list that you don't want to change.

Third error comes from this code:

```python3
        new_levels: FrozenList | list[Index]

        if isinstance(value_columns, MultiIndex):
            new_levels = value_columns.levels + (self.removed_level_full,)  # error: Cannot determine type of "__add__"  [has-type]
            new_names = value_columns.names + (self.removed_name,)

            new_codes = [lab.take(propagator) for lab in value_columns.codes]
        else:
            new_levels = [
                value_columns,
                self.removed_level_full,
            ]
            new_names = [value_columns.name, self.removed_name]
            new_codes = [propagator]
```

There's already an explicit type annotation on `new_levels`, so how should this be fixed in pandas?

test_cases/stdlib/builtins/test_list.py
```diff
@@ -0,0 +1,16 @@
...
+        return 1
+
+
+class Bar:
+    def asd(self) -> int:
+        return 2
+
+
+combined = [Foo()] + [Bar()]
+assert_type(combined, list[Foo | Bar])
```
AlexWaygood, line 16: We have to use old syntax in the test suite, since it's a `.py` file and we test on 3.7+

```suggestion
assert_type(combined, List[Union[Foo, Bar]])
```

Akuli:
Even if we don't know how the pandas error could be fixed, apart from `# type: ignore`, this fixes so many other problems that I think it's worth merging anyway.

AlexWaygood:
Looks like this PR might also fix #2383

Akuli:
It does. I edited the description.

Akuli:
I ran the mypy test suite locally. This break's mypy's `testMapStr` test, but only because the error message changes.

Akuli:
I think the mypy test failure points out a real problem. Consider this code (so buggy that doesn't show up in mypy_primer much):

```python3
asd: list[str] = []
print(asd + 1)
```

Error message before this PR: 

```
asd.py:2: error: Unsupported operand types for + ("List[str]" and "int")
```

Error message after this PR:

```
asd.py:2: error: No overload variant of "__add__" of "list" matches argument type "int"
asd.py:2: note: Possible overload variants:
asd.py:2: note:     def __add__(self, List[str]) -> List[str]
asd.py:2: note:     def [_S] __add__(self, List[_S]) -> List[Union[_S, str]]
```

This isn't exactly very user friendly, as `List[str]` is hidden deep inside the error message and it doesn't mention `+`.

Should figure out what mypy changes would be needed to get a simpler error message before we merge this PR?

hauntsaninja (approved):
Thanks for making this work! I'm also in favor of merging.

I don't think the error usability is bad enough to prevent merging. IIRC we've had other use of overloads for dunder methods recently that have caused similar changes to error messages.

PR #8258, opened 2022-07-08
Complete python-jose typing

hasier:
Complete the work started in https://github.com/python/typeshed/pull/8238 and make minor fixes to changes included there.

PR #8294, opened 2022-07-14
Improve typing of python-jose backends

hasier:
Somehow forgot to work on the `backends` for `python-jose` at https://github.com/python/typeshed/pull/8258

Adding some improvements where possible, also added a few comments where we depend on third party typing.


stubs/python-jose/jose/backends/base.pyi
```diff
@@ -1,17 +1,23 @@
...
-    def wrap_key(self, key_data) -> None: ...
-    def unwrap_key(self, wrapped_key) -> None: ...
+    def sign(self, msg: bytes) -> bytes: ...
+    def verify(self, msg: bytes, sig: bytes) -> bool: ...
+    def public_key(self: Self) -> Self: ...
+    def to_pem(self) -> bytes: ...
+    def to_dict(self) -> dict[str, Any]: ...
+    def encrypt(self, plain_text: str | bytes, aad: bytes | None = ...) -> tuple[bytes, bytes, bytes | None]: ...
+    def decrypt(
+        self, cipher_text: str | bytes, iv: str | bytes | None = ..., aad: bytes | None = ..., tag: bytes = ...
```
hasier, line 16: How do we usually go about cases where the function has a `None` default, as it happens with `tag` here, but then it checks it must not be `None`? https://github.com/mpdavis/python-jose/blob/master/jose/backends/cryptography_backend.py#L496-L497
Tests are currently failing due to this issue https://github.com/python/typeshed/runs/7339240425?check_suite_focus=true

AlexWaygood: I think stubtest is raising a fair point here. It looks to me like `ValueError` is only raised if `tag` is `None` _and_ `mode.name == "GCM"`. Presumably if `mode.name` is equal to something else, passing in `None` for the `tag` parameter is fine. https://github.com/mpdavis/python-jose/blob/96474ecfb6ad3ce16f41b0814ab5126d58725e2a/jose/backends/cryptography_backend.py#L495

stubs/python-jose/jose/backends/cryptography_backend.pyi
```diff
@@ -1,11 +1,11 @@
 from typing import Any
 
-from .base import Key as Key
+from .base import Key
 
-def decode_dss_signature(signature: bytes) -> tuple[int, int]: ...
-def encode_dss_signature(r: int, s: int) -> bytes: ...
```
AlexWaygood, line 6: Have these deliberately been deleted?

hasier: Yes, they don't seem to be exposed https://github.com/mpdavis/python-jose/blob/96474ecfb6ad3ce16f41b0814ab5126d58725e2a/jose/backends/cryptography_backend.py

stubs/python-jose/jose/backends/base.pyi
```diff
@@ -1,17 +1,23 @@
+from _typeshed import Self
 from typing import Any
 
 class Key:
+    # Enable when we can use stubs from installed dependencies,
+    # as `key` can be of type cryptography.x509.base.Certificate:
+    # from cryptography.x509 import Certificate
```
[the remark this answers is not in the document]
hasier: I am not sure most of the values will be available, and adding the types as a dependency seems a bit of an overkill, also knowing they are outdated.

In any case, things like `cryptography_backend` don't exist in the current stubs, I think it should be typed `cryptography.hazmat.backends.openssl.backend.Backend` but that does not seem to exist in the current stubs. So I was thinking it might be better to leave all as they are and just type them whenever it's possible with actual values 😕 

If it is useful, maybe I could go over a few of them and try to type them if they don't depend on `cryptography`, as I did for other backends.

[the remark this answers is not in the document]
AlexWaygood: > If it is useful, maybe I could go over a few of them and try to type them if they don't depend on cryptography, as I did for other backends.

I've never used these stubs or `python-jose` itself, so I can't comment on whether it's worth your time to do that :) Feel free to skip it if you don't think it would be useful 👍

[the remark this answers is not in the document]
hasier: It's the first time for me as well, I'm adding stubs as I need them and appreciate a lot the time y'all take to review them 🙏  If that's ok with you, I think I'll leave it like this for now, will further improve in a separate PR if I find some more headaches 🙂

stubs/python-jose/jose/backends/base.pyi
Akuli: While decrypting with AES-GCM makes sense without a tag, it's usually either a mistake or a part of some quick-and-dirty script, as there's no guarantee that you decrypted the data you intended to decrypt.

But maybe we should disallow passing a tag altogether when mode isn't GCM. Googling "aes tag", every result mentions GCM which makes me believe that tags are purely an AES-GCM thing.

Akuli: And now that I look at the code, it doesn't use `tag` at all when mode isn't GCM. So I think there should be two overloads:
- Mode GCM, tag specified as not None (this actually needs to be split into two overloads: tag passed positionally, tag passed by keyword)
- Any other mode, tag None or unspecified

But this too can be improved in a separate PR, if it needs to be done at all :)

Akuli: I don't think we can do much typing based on IV values: it's just a `bytes` object (possibly `str`?), of length unknown at type checking time. Either way, this is beyond the scope of this PR and we should probably just merge this as is :)

hauntsaninja (approved):
Spot checked a couple things, looks good! Thanks for all your work here!

Issue #7130, opened 2022-02-04
typing_extensions.pyi imports `Protocol` from typing.pyi rather than defining its own


erictraut:
Until recently, `typing_extensions.pyi` declared its own `Protocol` symbol as follows:
```python
Protocol: _SpecialForm = ...
```

It now directly imports `Protocol` from `typing.pyi` and re-exports this symbol. This means pyright can no longer differentiate between the two symbols. There's an important difference because `typing.Protocol` will generate a runtime exception if used in versions of Python prior to 3.7, but `typing_extensions.Protocol` will not.

@JelleZijlstra, is there a reason this change was made? Or put another way, is there a reason not to change it back?

Here's a link to bug that describes the regression in pyright: https://github.com/microsoft/pyright/issues/2987

JelleZijlstra:
I'm ok with changing this back; there wasn't really a compelling reason to merge the two.

AlexWaygood:
Apologies for causing the regression @erictraut (it was my PR IIRC).

Seems like it should be added back to `typing_extensions` along with a comment explaining why it's duplicated there from `typing` rather than being re-exported, even though it's available in `typing` unconditionally (from typeshed's POV — at runtime, it obviously can only be imported from `typing` in 3.8+).

PR #8295, opened 2022-07-14
Add `typing_extensions.NamedTuple`

AlexWaygood:
Fixes https://github.com/python/typing_extensions/issues/56


stdlib/typing_extensions.pyi
```diff
@@ -1,6 +1,8 @@
 import abc
+import collections
 import sys
 from _typeshed import IdentityFunction, Self as TypeshedSelf  # see #6932 for why the Self alias cannot have a leading underscore
+from collections.abc import Iterable,
```
AlexWaygood, line 5:
```suggestion
from collections.abc import Iterable
```

stdlib/typing_extensions.pyi
```diff
@@ -237,3 +241,24 @@ else:
...
+            _field_types: dict[str, type]
+        _field_defaults: dict[str, Any]
+        _fields: tuple[str, ...]
+        _source: str
+        @overload
+        def __init__(self, typename: str, fields: Iterable[tuple[str, Any]] = ...) -> None: ...
+        @overload
+        def __init__(self, typename: str, fields: None = ..., **kwargs: Any) -> None: ...
+        @classmethod
+        def _make(cls: type[_T], iterable: Iterable[Any]) -> _T: ...
```
AlexWaygood, line 258:
```suggestion
        def _make(cls: type[TypeshedSelf], iterable: Iterable[Any]) -> TypeshedSelf: ...
```

stdlib/typing_extensions.pyi
```diff
@@ -194,8 +197,9 @@ else:
 
 # New things in 3.11
 if sys.version_info >= (3, 11):
-    from typing import (
+    from typing import (  # Ability to create generic NamedTuples is new in 3.11
         LiteralString as LiteralString,
+        NamedTuple as NamedTuple,
```
AlexWaygood, line 202:
```suggestion
    from typing import (
        LiteralString as LiteralString,
        # Ability to create generic NamedTuples is new in 3.11
        NamedTuple as NamedTuple,
```

AlexWaygood:
> Would anything break if we unconditionally did `from typing import NamedTuple as NamedTuple`? We already do that for `overload` too.

My thinking was that type checkers might want to distinguish `typing_extensions.NamedTuple` from `typing.NamedTuple`, so that they can emit an error if a user tries to create a generic `typing.NamedTuple` on <3.11, since that will fail at runtime. I.e., it would be a similar argument to the argument made in #7130 for declaring `typing_extensions.Protocol` as a distinct symbol to `typing.Protocol`.

JelleZijlstra (approved):

PR #8298, opened 2022-07-14
threading: __exit__ only returns None

### LINKED ISSUE #8297: Lock.__exit__() signature says it can return bool, but it doesn't
## Environment data
- Python 3.10.5 (tags/v3.10.5:f377153, Jun  6 2022, 16:14:13)
- Windows-10-10.0.22000-SP0

## Issue 
The type information for the `Lock` class [(here)](https://github.com/python/typeshed/blob/10f3238998737fa7c73a1005b61b5c717cc9c429/stdlib/threading.pyi#L107) indicates to the type checkers that the `__exit__()` method may trap exceptions swallowed within a `with` code block, for the returned type can be `bool`.

However looking at the `Lock.__exit__` code shows that it actually never returns `bool`. 
(file `...\Python310\Lib\threading.py` on my Windows installation)

```python
class _RLock:
    def __exit__(self, t, v, tb):
        self.release()
```

As a consequence, the type checkers report false positive.
A detailled code example is [here](https://github.com/microsoft/pylance-release/issues/3043).

JelleZijlstra:
Fixes #8297 

`threading.Lock.__exit__` is implemented in C and can only return None: https://github.com/python/cpython/blob/6cbb57f62d345d7a5d6aeb1b3b5d37a845344d5e/Modules/_threadmodule.c#L200
`threading.Condition.__exit__` returns whatever its lock's `__exit__` returns, but our type annotations indicate that the lock is always a `Lock` or an `_RLock`, and neither returns anything other than None.


srittau (approved):
LGTM, but I didn't look at dummy threading.

hauntsaninja (approved):

hauntsaninja:
dummy_threading is removed and not very useful, so I'm happy not improving it. But if someone feels like we could open another PR for it.

PR #8303, opened 2022-07-15
Add stubsabot Github Action

hauntsaninja:


.github/workflows/stubsabot.yml
```diff
@@ -0,0 +1,48 @@
+name: Run stubsabot weekly
```
srittau, line 1: Would it make more sense to run it daily? This way less "chunks" of work would be generated at a time. And since stubsabot ignores updates that already have a PR (if I understand correctly), this shouldn't make a difference in this regard.

srittau (approved):

hauntsaninja:
(Give me a couple minutes to run a few more tests)

.github/workflows/stubsabot.yml
hauntsaninja: Yup, Akuli is right. I just added what I hope is adequate protection against undesirable force push situations.

I've kept it weekly for now, since I'm not sure how much time I'll have in the next few days to fix it if it explodes. I'll change to daily once I'm sure that things are working well.

hauntsaninja:
Let's see how this goes!

PR #8304, opened 2022-07-15
Run `pycln` as a pre-commit hook in CI

AlexWaygood:
`pycln` is a pre-commit hook to auto-remove unused import statements. The tool is used by several high-profile open-source projects such as Pyodide, Poetry and Rich. It is fully compatible with `.pyi` files as well as `.py` files.


.pre-commit-config.yaml
```diff
@@ -1,4 +1,9 @@
 repos:
+  - repo: https://github.com/hadialqattan/pycln
+    rev: v2.0.2 # must match requirements-tests.txt
+    hooks:
+      - id: pycln
+        args: [--all]
```
AlexWaygood, line 6:
```suggestion
        args: [--all, .]
```

AlexWaygood:
Here's `pycln` successfully removing some unused imports that pyright didn't pick up on: https://github.com/python/typeshed/pull/8304/commits/018c3efa8fa0534455e994ba9951d6252d0f188b

.pre-commit-config.yaml
```diff
@@ -1,4 +1,9 @@
 repos:
+  - repo: https://github.com/hadialqattan/pycln
+    rev: v2.0.2 # must match requirements-tests.txt
+    hooks:
+      - id: pycln
+        args: [--all, .]
```
srittau, line 6: Maybe limit this to `stubs`, `stdlib`, and `scripts`, to be safe.

AlexWaygood: `tests`, as well?

srittau (approved):
LGTM, but I would like to hear feedback of other maintainers.

.pre-commit-config.yaml
srittau: Sure.

hauntsaninja (commented):
I'm fine with this. Not familiar with pycln, but it looks like it uses libcst, which is promising (unlike isort, which was plagued with bugs for a long time).

Akuli:
Have you tried running pycln on all of our existing stubs? Does it delete anything? If it does, a file like that would be quite annoying to contribute to.

AlexWaygood:
> Have you tried running pycln on all of our existing stubs? Does it delete anything? If it does, a file like that would be quite annoying to contribute to.

Yes, I ran it on the whole of typeshed before filing this PR.

It made changes to `aiohttp`, where it picked up a genuine bug (an import wasn't being re-exported when it should have been). I fixed that in the first commit in this PR.

It also identified unused imports in `stdlib/tokenize.pyi` and `stdlib/unittest/loader.pyi`. `pycln` auto-fixed these in the third commit in this PR: 8d0d6d3

PR #8311, opened 2022-07-15
Make VERSIONS more accurate

hauntsaninja:
Some of the versions starting at 3.6 looked suspicious, so I
checked those against git history for the CPython repo.
Note that I currently do not have access to any EOL Python.


JelleZijlstra:
This was intentional, we didn't bother getting precise versions before 3.6 when we added the file. But I guess it's nice to get it accurate.

stdlib/VERSIONS
```diff
@@ -21,39 +21,39 @@ __future__: 2.7-
 __main__: 2.7-
 _ast: 2.7-
 _bisect: 2.7-
-_bootlocale: 3.6-3.9
+_bootlocale: 3.4-3.9
 _codecs: 2.7-
 _collections_abc: 3.3-
-_compat_pickle: 3.6-
-_compression: 3.6-
+_compat_pickle: 2.7-
```
JelleZijlstra, line 27: Not in my 2.7

stdlib/VERSIONS
```diff
@@ -21,39 +21,39 @@ __future__: 2.7-
...
+_compression: 3.5-
 _csv: 2.7-
 _curses: 2.7-
-_decimal: 3.6-
-_dummy_thread: 3.6-3.8
+_decimal: 2.7-
```
JelleZijlstra, line 31: Same

stdlib/VERSIONS
```diff
@@ -21,39 +21,39 @@ __future__: 2.7-
...
+_dummy_thread: 3.0-3.8
 _dummy_threading: 2.7-3.8
 _heapq: 2.7-
-_imp: 3.6-
+_imp: 2.7-
```
JelleZijlstra, line 35: Same

stdlib/VERSIONS
```diff
@@ -21,39 +21,39 @@ __future__: 2.7-
...
 _json: 2.7-
 _markupbase: 2.7-
 _msi: 2.7-
-_operator: 3.6-
+_operator: 3.4-
 _osx_support: 2.7-
-_posixsubprocess: 3.6-
+_posixsubprocess: 2.7-
```
JelleZijlstra, line 41: And again.

stdlib/VERSIONS
```diff
@@ -289,7 +289,7 @@ wsgiref.types: 3.11-
 xdrlib: 2.7-
 xml: 2.7-
 xmlrpc: 3.0-
-xxlimited: 3.6-
+xxlimited: 2.7-
```
JelleZijlstra, line 292: Not in my 2.7

stdlib/VERSIONS
```diff
@@ -21,39 +21,39 @@ __future__: 2.7-
...
+_threading_local: 2.7-
 _tkinter: 2.7-
-_tracemalloc: 3.6-
+_tracemalloc: 3.4-
 _typeshed: 2.7-  # not present at runtime, only for type checking
 _warnings: 2.7-
 _weakref: 2.7-
 _weakrefset: 2.7-
-_winapi: 3.6-
+_winapi: 2.7-
```
JelleZijlstra, line 56: Can't check this one

stdlib/VERSIONS
```diff
@@ -21,39 +21,39 @@ __future__: 2.7-
...
+_pydecimal: 3.5-
 _random: 2.7-
-_sitebuiltins: 3.6-
+_sitebuiltins: 3.4-
 _socket: 3.0-  # present in 2.7 at runtime, but not in typeshed
-_stat: 3.6-
+_stat: 3.4-
 _thread: 2.7-
-_threading_local: 3.6-
+_threading_local: 2.7-
```
JelleZijlstra, line 49: This one does exist in 2.7

hauntsaninja (commented):
I was using `git tag --contains asdf` to match these to releases. I think there's something strange going on with the git history that I don't fully understand — it looks like 3.1 is in the history of 2.7.4

hauntsaninja:
Thanks for checking! I used more git and now I'm a little more confident in these

JelleZijlstra (approved):
Didn't confirm all these, but this seems fine

PR #8314, opened 2022-07-16
Use packaging to parse versions

hauntsaninja:
packaging is the standard tool for dealing with versions and
requirements. This allows us to easily support all PEP 508 specs.
packaging also has good error messages, so I don't create our own here.
This lets us get rid of our handrolled half parser and sanity checks.
I kept the whitespace consistency check, since that's just a formatting
lint and isn't testing the correctness of parsing (which we now
definitely do correctly).

### DIFF
--- a/requirements-tests.txt
+++ b/requirements-tests.txt
@@ -10,3 +10,4 @@ flake8-pyi==22.5.1
 isort==5.10.1
 tomli==1.2.2
 packaging==21.3
+termcolor
--- a/requirements-tests.txt
+++ b/requirements-tests.txt
@@ -9,3 +9,4 @@ flake8-pyi==22.5.1
 # must match .pre-commit-config.yaml
 isort==5.10.1
 tomli==1.2.2
+packaging==21.3
--- a/tests/check_consistent.py
+++ b/tests/check_consistent.py
@@ -134,7 +134,7 @@ def check_metadata() -> None:
 
 
 if __name__ == "__main__":
-    assert sys.version_info >= (3, 9)
+    assert sys.version_info >= (3, 9), "Python 3.9+ is required to run this test"
     check_stdlib()
     check_versions()
     check_stubs()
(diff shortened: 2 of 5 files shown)


hauntsaninja:
I'll make corresponding changes to stub_uploader once https://github.com/typeshed-internal/stub_uploader/pull/49 is merged

requirements-tests.txt
```diff
@@ -9,3 +9,5 @@ flake8-pyi==22.5.1
 # must match .pre-commit-config.yaml
 isort==5.10.1
 tomli==1.2.2
+packaging==21.3
+termcolor
```
AlexWaygood, line 13: I'm fine with this, but can you delete this section in CONTRIBUTING, since it's no longer an "optional dependency" if we do this: https://github.com/python/typeshed/blob/master/CONTRIBUTING.md#optional-dependencies

tests/check_consistent.py
```diff
@@ -155,6 +134,7 @@ def check_metadata() -> None:
 
 
 if __name__ == "__main__":
+    assert sys.version_info >= (3, 9)
```
AlexWaygood, line 137:
```suggestion
    assert sys.version_info >= (3, 9), "Python 3.9+ is required to run this test"
```

The other tests which require a version >3.7 have a note in https://github.com/python/typeshed/blob/master/tests/README.md to that effect; could you also add a note there?

AlexWaygood (commented):
Thanks, this looks great! Just a few nits:

tests/check_consistent.py
srittau: Or maybe we can just document that we require 3.9+ for all tests? (Even if some tests still work with older Python versions.)

AlexWaygood: That sounds even better actually, it'll mean we don't have to update the README in the tests directory so often :)

AlexWaygood (approved):
Nice!

Issue #8302, opened 2022-07-15
pycurl: CurlMulti.setopt not typed


OneMoreSecond:
Problem:
In following file, CurlMulti.setopt is not typed like Curl.setopt
https://github.com/python/typeshed/blob/197dac2c0c216230d39461ec415e6daabe5e8ea1/stubs/pycurl/pycurl.pyi

Reference:
http://pycurl.io/docs/latest/curlobject.html
http://pycurl.io/docs/latest/curlmultiobject.html

PR #8315, opened 2022-07-17
typing CurlMulti.setopt

OneMoreSecond:
Fixes https://github.com/python/typeshed/issues/8302

The new line is copied from the line of Curl.setopt


AlexWaygood (approved):
A lot of the `Any`s in this file look like they could probably be a bit more precise, but I think this is fine for now. Thanks for the PR!

PR #8319, opened 2022-07-17
Add missing modules

hauntsaninja:
mypy 0.970 will include https://github.com/python/mypy/pull/13030 which will start complaining about these for stubs we check strictly (aka without `--ignore-missing-stub`).


AlexWaygood:
Maybe it would be easier just to allowlist these modules when 0.970 comes out?

hauntsaninja:
Yeah, I'll try something else.

PR #8325, opened 2022-07-18
added explicit typing to all async pipeline commands

bentheiii:


stubs/redis/redis/asyncio/client.pyi
```diff
@@ -169,7 +170,7 @@ PSWorkerThreadExcHandlerT: TypeAlias = PubsubWorkerExceptionHandler | AsyncPubsu
 CommandT: TypeAlias = tuple[tuple[str | bytes, ...], Mapping[str, Any]]
 CommandStackT: TypeAlias = list[CommandT]
 
-class Pipeline(Redis[_StrType], Generic[_StrType]):
+class Pipeline(Generic[_StrType]):
```
srittau, line 173: Not a full review yet, but since `Pipeline` derives from `Redis` at runtime, we should keep that relationship. This unfortunately means that we will need lots of `# type: ignore[override]`. See also `redis.client.Pipeline`.

srittau (changes requested):
Thanks! While I didn't verify every argument type, things are looking good. There's only one thing I noticed: Could you replace instances of `Any | None` with `Incomplete | None`? `_typeshed.Incomplete` is a fairly new alias for `Any` (so it's not used much yet in typeshed) that's supposed to mark types that can be improved as opposed to `Any`, which is used for types where we can't do better within the capabilities of the type system.

stubs/redis/@tests/stubtest_allowlist.txt
```diff
@@ -19,6 +19,12 @@ redis.typing.CommandsProtocol.__init__
 # unclear problems
 redis.Sentinel.master_for
 redis.Sentinel.slave_for
+redis.asyncio.client.Pipeline.command_info
+redis.asyncio.client.Pipeline.debug_segfault
+redis.asyncio.client.Pipeline.memory_doctor
+redis.asyncio.client.Pipeline.memory_help
+redis.asyncio.client.Pipeline.script_debug
+redis.asyncio.client.Pipeline.shutdown
```
AlexWaygood, line 27: These allowlist entries are all because the method is an `async def` method at runtime but not an `async def` method in the stub.

1. Are you sure it's correct to have them as not being `async def` in the stub?
2. If you are sure, can you put these allowlist entries in a separate section, and add a comment explaining why we're ignoring the stubtest errors? I don't think these entries belong in the "unclear problems" section, since we know why stubtest is complaining here.

AlexWaygood (changes requested):

stubs/redis/@tests/stubtest_allowlist.txt
bentheiii: Hi @AlexWaygood 

I'm sure in the sense that it doesn't really matter, in principle this function should be syncronous in the pipeline, just like all other redis commands. In practice, because the functions are defined as `async` in the superclass unlike most commands (I can't really tell why), then any attempt to use these commands in a pipeline would result in a bug.

I will place these entreis in their own section

srittau (approved):
LGTM, if we need to tweak the stubs later, we can always do so.

PR #8328, opened 2022-07-19
wave: allow passing anonymous tuple

### LINKED ISSUE #8322: Wave stdlib parameter regression in mypy 0.950
**Bug Report**

When working with the `wave` module of the standard library, valid code from mypy <0.950 works but errors with 
```error: Argument 1 to "setparams" of "Wave_write" has incompatible type "waveparams"; expected "_wave_params"``` 
on 0.950+
**To Reproduce**

Using Example code 
```python
# bug_report.py
import wave
import collections

waveparams = collections.namedtuple(
    "waveparams",
    "nchannels, sampwidth, framerate, nframes, comptype, compname",
)

with wave.open("test.wav", "wb") as wav_file:
    params = waveparams(
        nchannels=1,
        sampwidth=2,
        framerate=16_000,
        nframes=0,
        comptype="NONE",
        compname="NONE",
    )
    wav_file.setparams(params)
```
running `mypy` an older version (say 0.931) it says no issues. but using a fully up to date version throws and error of
`error: Argument 1 to "setparams" of "Wave_write" has incompatible type "waveparams"; expected "_wave_params"`

I am not sure why they would be 'incompatible' here. My team was making a namedtuple to make up for the fact that wav_file.setparameters just took a raw untyped tuple as input. 

**Expected Behavior**

The named tuple here does match the required fields ofr setparams of `Wave_write` object from the stdlib so it should pass.

**Actual Behavior**

On newer versions of mypy this errors

**Your Environment**

- Python version used: 3.8
- Operating system and version: Linux

hauntsaninja:
Fixes #8322


f4hy:
Thanks.

PR #8333, opened 2022-07-19
Remove types-SQLAlchemy requirement from Flask-SQLAlchemy

kkirsche:
This removes the types-SQLAlchemy requirement from Flask-SQLAlchemy because the user may instead be using sqlalchemy-provided stubs rather than third-party stubs.

If there is a way to instead make this conditional, this should be closed and that approach should be used instead to isolate the version appropriately.

### DIFF
--- a/stubs/Flask-SQLAlchemy/METADATA.toml
+++ b/stubs/Flask-SQLAlchemy/METADATA.toml
@@ -1,2 +1,2 @@
 version = "2.5.*"
-requires = ["types-SQLAlchemy"]
+requires = []


kkirsche:
Sounds like maybe this should be an issue instead :/ from the pre-commit results

AlexWaygood:
There's two things that don't make this possible yet, unfortunately:
- We currently don't have the infrastructure set up to allow non-`types` dependencies in typeshed: #5768
- `SQLAlchemy` now has inline type annotations for much of its code, and has a `py.typed` file on its `master` branch. But the most recent release (`1.4.39`) _doesn't_ have a `py.typed` file. Without a `py.typed` file, the type annotations are only for internal use; a type-checker won't be able to see them and use them if `SQLAlchemy` is installed as a package.

kkirsche:
Ah, that's too bad. I was really hoping it'd be possible to use the sqlalchemy `sqlalchemy2-stubs` with this type set as this would drastically simplify working with Flask-SQLAlchemy for users on 1.4 who use their official mypy stubs

kkirsche:
Thank you for the information though and sorry for the incorrect PR

srittau:
If there are useful types from `sqlalchemy2-stubs` that aren't in `types-sqlalchemy` yet, I believe it would be useful to have them integrated.

AlexWaygood:
No worries, thanks for the PR, even if it wasn't successful! :)

PR #8337, opened 2022-07-19
Fix return type of email.header.decode_header()

srittau:


srittau:
Nice intuitive API:

```python
>>> from email.header import decode_header
>>> decode_header("abc")
[('abc', None)]
>>> decode_header("=?iso-8859-1?q?p=F6stal?=")
[(b'p\xf6stal', 'iso-8859-1')]
>>> decode_header("before =?iso-8859-1?q?p=F6stal?= after")
[(b'before ', None), (b'p\xf6stal', 'iso-8859-1'), (b' after', None)]
```

AlexWaygood (approved):
That API is going to haunt my dreams tonight

PR #8321, opened 2022-07-17
Add missing third party modules

hauntsaninja:
Better version of #8319. mypy 0.970 will include https://github.com/python/mypy/pull/13030 which will start complaining about these for stubs we check strictly (aka without --ignore-missing-stub). I removed some modules that I think hold negative value for us.

PR #8340, opened 2022-07-20
Allowlist useless modules

hauntsaninja:
These are the ones I chose not to add from #8321

Fixes most of #8339


stubs/stdlib-list/@tests/stubtest_allowlist.txt
```diff
@@ -0,0 +1 @@
+stdlib_list.fetch
```
hauntsaninja, line 1:
```suggestion
# looks like developer tooling that's not part of the public API
stdlib_list.fetch
```

PR #8346, opened 2022-07-20
Add some missing Windows constants

AlexWaygood:


srittau (approved):
I can't verify the correctness of the constants, but looks plausible to me.

PR #8352, opened 2022-07-21
Improve `redis.backoff` types

sobolevn:
Ref: https://github.com/redis/redis-py/blob/master/redis/backoff.py


stubs/redis/redis/backoff.pyi
```diff
@@ -1,30 +1,31 @@
...
+    def __init__(self, cap: int, base: int) -> None: ...
+    def compute(self, failures: int) -> int: ...
 
 class EqualJitterBackoff(AbstractBackoff):
-    def __init__(self, cap, base) -> None: ...
-    def compute(self, failures): ...
+    def __init__(self, cap: int, base: int) -> None: ...
+    # Technically it returns `float`,
+    # but `# type: ignore[override]` seems like an overkill.
+    def compute(self, failures: int) -> int: ...
```
srittau, line 27: Wouldn't it make more sense to have `AbstractBackoff.compute()` and this `compute()` return `float` and the other sub-classes return `int`?

sobolevn: Or probably everything can be a `float` 🤔 
I will experiment a bit more with this.

sobolevn:
Ok, the result of `compute()` is used in two places:
1. https://github.com/redis/redis-py/blob/b2bcbc8bf1ae5622567acbc20e30e9c2e977e5c0/redis/retry.py#L52-L54
2. https://github.com/redis/redis-py/blob/bea72995fd39b01e2f0a1682b16b6c7690933f36/redis/asyncio/retry.py#L65-L67

Both just send `backoff` to `sleep()`. Which works with any floats.

srittau (approved):

PR #8355, opened 2022-07-21
xml.dom.minidom: add more missing type annotations to methods on Element

tetsuok:
This PR adds more missing type annotation to the methods on `Element`: `setAttribute`, `getAttributeNode`, `removeAttribute`, and `getElementsByTagName` in the same way as #8350. This is to avoid getting an error when we call an unannotated function in an annotated function. According to [Document Object Model (DOM) Level 1 Specification](https://www.w3.org/TR/REC-DOM-Level-1/level-one-core.html), these methods take strings:

```
interface Element : Node {
  ...
  void                      setAttribute(in DOMString name, 
                                         in DOMString value)
  Attr                      getAttributeNode(in DOMString name);
  void                      removeAttribute(in DOMString name)
                                            raises(DOMException);
  NodeList                  getElementsByTagName(in DOMString name);
  ...
}
```


srittau (approved):

srittau:
Thanks!

Issue #8356, opened 2022-07-21
all BeautifulSoup type hints are wrong


AIGeneratedUsername:
## Summary

It is a total chaos. Where I have `BeautifulSoup`, I get `Iterable[PageElement]`. Where I have `Tag`, I get `PageElement`, etc. See the example below.

I have errors from Mypy in every line:
`Item "NavigableString" of "Union[Tag, NavigableString]" has no attribute "find_all"`
`"Iterable[PageElement]" has no attribute "__next__"  [attr-defined]`
`"enumerate" has incompatible type "BeautifulSoup"; expected "Iterable[<nothing>]"`
`Value of type "Union[Tag, NavigableString, None]" is not indexable`
etc.

The only workaround is either `#type: ignore` dozens of lines, or uninstall "types-beautifulsoup4".

## Reproducible code sample
```python
html_doc = """
<html><head><title>The Dormouse's story</title></head>
<body>
<p class="title"><b>The Dormouse's story</b></p>

<p class="story">Once upon a time there were three little sisters; and their names were
<a href="http://example.com/elsie" class="sister" id="link1">Elsie</a>,
<a href="http://example.com/lacie" class="sister" id="link2">Lacie</a> and
<a href="http://example.com/tillie" class="sister" id="link3">Tillie</a>;
and they lived at the bottom of a well.</p>

<p class="story">...</p>
"""
from bs4 import BeautifulSoup
soup = BeautifulSoup(html_doc, 'lxml')
reveal_type(soup)
for el in soup:
    reveal_type(el)
```

## Output

```bash
_local.py:16:13: note: Revealed type is "bs4.BeautifulSoup"
_local.py:17:11: error: "Iterable[PageElement]" has no attribute "__next__"  [attr-defined]
    for el in soup:
              ^
_local.py:18:17: note: Revealed type is "Any"
Found 1 error in 1 file (checked 1 source file)
```

## Expected output
No errors.

## Versions

beautifulsoup4 = "4.11.1"
types-beautifulsoup4 = "4.11.1"

JelleZijlstra:
Seems like the annotation for `Tag.__iter__` is incorrect: it should return an Iterator, not an Iterable.

We make that mistake in a few other places:
```
% git grep 'def __iter__' | grep Iterable
stdlib/_typeshed/wsgi.pyi:        def __iter__(self) -> Iterable[bytes]: ...
stdlib/wsgiref/types.pyi:    def __iter__(self) -> Iterable[bytes]: ...
[...]

PR #8358, opened 2022-07-21
wsgiref.types: use Iterator, not Iterable

hauntsaninja:
https://github.com/python/typeshed/issues/8356


srittau (approved):
Can be merged when the CPython PR is merged.

PR #8361, opened 2022-07-22
Always return `Iterator`, not `Iterable` for `__iter__`

sobolevn:
After https://github.com/python/typeshed/pull/8358 I went looking for other cases of `__iter__` returing `Iterable`. I found two:
1. https://github.com/python/cpython/blob/934b25dcc492dcbca4da9d63d0d71dc940fc0375/Lib/wsgiref/validate.py#L216-L221 which clearly is an `Iterator`
2. https://github.com/google/python-gflags/blob/4f06c3d0d6cbe9b1fb90ee9fb1c082b3bf9285f6/gflags/flagvalues.py#L666-L667 which is also an `Iterator`

According to the docs:

> `object.__iter__(self)`
> This method is called when an [iterator](https://docs.python.org/3/glossary.html#term-iterator) is required for a container. This method should return a new iterator object that can iterate over all the objects in the container. For mappings, it should iterate over the keys of the container.

Maybe this can be added as a new `flake8-pyi` rule.


AlexWaygood:
> Maybe this can be added as a new `flake8-pyi` rule.

@hauntsaninja's already filed https://github.com/PyCQA/flake8-pyi/issues/247 :)

AlexWaygood (approved):
Nice, thanks!

PR #8362, opened 2022-07-22
Improve `redis.utils` types

sobolevn:
Ref: https://github.com/redis/redis-py/blob/b2bcbc8bf1ae5622567acbc20e30e9c2e977e5c0/redis/utils.py
Unused function issue: https://github.com/redis/redis-py/issues/2299


stubs/redis/redis/utils.pyi
```diff
@@ -14,11 +15,8 @@ def from_url(url: str, *, db: int = ..., decode_responses: Literal[True], **kwar
...
-@overload
-def str_if_bytes(value: bytes) -> str: ...  # type: ignore[misc]
-@overload
-def str_if_bytes(value: _T) -> _T: ...
+def str_if_bytes(value: str | bytes) -> str: ...
 def safe_str(value: object) -> str: ...
-def dict_merge(*dicts): ...
-def list_keys_to_dict(key_list, callback): ...
-def merge_result(command, res): ...
+def dict_merge(*dicts: Mapping[str, object]) -> dict[str, object]:
```
AlexWaygood, line 20:
```suggestion
def dict_merge(*dicts: Mapping[str, object]) -> dict[str, object]: ...
```

stubs/redis/redis/utils.pyi
```diff
@@ -14,11 +15,8 @@ def from_url(url: str, *, db: int = ..., decode_responses: Literal[True], **kwar
...
-@overload
-def str_if_bytes(value: bytes) -> str: ...  # type: ignore[misc]
-@overload
-def str_if_bytes(value: _T) -> _T: ...
+def str_if_bytes(value: str | bytes) -> str: ...
 def safe_str(value: object) -> str: ...
-def dict_merge(*dicts): ...
-def list_keys_to_dict(key_list, callback): ...
-def merge_result(command, res): ...
+def dict_merge(*dicts: Mapping[str, object]) -> dict[str, object]: ...
```
AlexWaygood, line 20: We can probably be a bit more sophisticated here — if all the input `Mapping`s have `int` as their value type, the returned `dict` will also have `int` as its value type:

```suggestion
def dict_merge(*dicts: Mapping[str, _T]) -> dict[str, _T]: ...
```

AlexWaygood: But then `_T` just gets solved as `object`, which is no worse than what you have now

AlexWaygood (approved):

sobolevn:
Thank you! 👍

PR #8365, opened 2022-07-22
Change `Stub versioning` wording in `CONTRIBUTING.md`

sobolevn:
I don't think that we ever should use one-tuples (like `(3,)`) in our current setup.
I think that this is a legacy from times when python2 stubs were a thing.
This might confuse people.


srittau (approved):

srittau:
Good catch!

Issue #8370, opened 2022-07-22
pyopenssl Context.set_ocsp_client_callback has incorrect argument types


ShaneHarvey:
PyMongo's mypy just started failing with: 
```
pymongo/pyopenssl_context.py:195: error: Argument "data" to
"set_ocsp_client_callback" of "Context" has incompatible type "_CallbackData";
expected "Optional[bytes]"  [arg-type]
    ...csp_client_callback(callback=_ocsp_callback, data=self._callback_data)
```
This was caused by the changes in https://github.com/python/typeshed/pull/8367

The bug is that set_ocsp_client_callback's data argument can be Any type, not just bytes. If we want to get fancy we can use generics to match the type signature of callback with data:
```
_T = TypeVar("_T")
...
    def set_ocsp_server_callback(
        self, callback: Callable[[Connection, _T | None], bytes], data: _T | None = ...
    ) -> None: ...
    def set_ocsp_client_callback(
        self, callback: Callable[[Connection, bytes, _T | None], bool], data: _T | None = ...
    ) -> None: ...
```

AlexWaygood:
Cc. @sobolevn

PR #8371, opened 2022-07-22
Fix pyopenssl hints for set_ocsp_server_callback/set_ocsp_client_callback

ShaneHarvey:
Fixes https://github.com/python/typeshed/issues/8370:

```
pymongo/pyopenssl_context.py:195: error: Argument "data" to
"set_ocsp_client_callback" of "Context" has incompatible type "_CallbackData";
expected "Optional[bytes]"  [arg-type]
    ...csp_client_callback(callback=_ocsp_callback, data=self._callback_data)
```


sobolevn (approved):
Thank you! I only saw `bytes` use-cases.

PR #8374, opened 2022-07-23
Add missing methods to `OpenSSL.SSL.Connection`

sobolevn:
For now, I've added all methods from 21.0
If there are new ones in 22.x, they are to be added later.

Source: https://github.com/pyca/pyopenssl/blob/21.0.0/src/OpenSSL/SSL.py#L1527-L2501

I am marking this as a draft for now, because I have to finish some details.


stubs/pyOpenSSL/OpenSSL/SSL.pyi
```diff
@@ -155,7 +195,10 @@ class Context:
...
     def set_alpn_select_callback(self, callback: Callable[[Connection, list[bytes]], bytes]) -> None: ...
-    def set_ocsp_server_callback(self, callback: Callable[[Connection, _T | None], bytes], data: _T | None = ...) -> None: ...
+    def get_alpn_proto_negotiated(self) -> bytes: ...
+    def set_ocsp_server_callback(
+        self, callback: Callable[[Connection, bytes | None], bytes], data: bytes | None = ...
+    ) -> None: ...
     def set_ocsp_client_callback(
-        self, callback: Callable[[Connection, bytes, _T | None], bool], data: _T | None = ...
+        self, callback: Callable[[Connection, bytes, bytes | None], bool], data: bytes | None = ...
     ) -> None: ...
```
AlexWaygood, line 204: Looks like you're accidentally reverting #8371 here

sobolevn: I will rebase later :)

sobolevn:
Ok, I've tried to make sure that our stubtest catches missing attributes, but it does not. For example, I removed `__getattr__` from `Context`. It passed our CI, but mypy_primer found several missing methods:

```
+ src/urllib3/contrib/pyopenssl.py:452: error: "Context" has no attribute "get_verify_mode"  [attr-defined]
+ src/urllib3/contrib/pyopenssl.py:459: error: "Context" has no attribute "set_default_verify_paths"  [attr-defined]
+ src/urllib3/contrib/pyopenssl.py:494: error: "Context" has no attribute "set_passwd_cb"  [attr-defined]
```

I will work on `Context` in the next PRs.

stubs/pyOpenSSL/OpenSSL/SSL.pyi
```diff
@@ -115,25 +117,63 @@ def SSLeay_version(type: int) -> str: ...
...
 class Connection:
-    def __getattr__(self, name: str) -> Any: ...  # incomplete
+    def __getattr__(self, name: str) -> Any: ...  # takes attributes from `self._socket`
     def __init__(self, context: Context, socket: socket.socket | None = ...) -> None: ...
+    def get_context(self) -> Context: ...
+    def set_context(self, context: Context) -> None: ...
+    def get_servername(self) -> bytes | None: ...
+    def set_tlsext_host_name(self, name: bytes) -> None: ...
+    def pending(self) -> int: ...
+    def send(self, buf: ReadableBuffer, flags: int = ...) -> int: ...
```
AlexWaygood, line 127: This also accepts `str` (though it's deprecated): https://github.com/pyca/pyopenssl/blob/51717970261d6283b681f5fa8347808409f0991d/src/OpenSSL/SSL.py#L1708

```suggestion
    def send(self, buf: ReadableBuffer | str, flags: int = ...) -> int: ...
```

stubs/pyOpenSSL/OpenSSL/SSL.pyi
```diff
@@ -115,25 +117,63 @@ def SSLeay_version(type: int) -> str: ...
...
+    write = send
+    def sendall(self, buf: ReadableBuffer, flags: int = ...) -> int: ...
```
AlexWaygood, line 129:
```suggestion
    def sendall(self, buf: ReadableBuffer | str, flags: int = ...) -> int: ...
```

stubs/pyOpenSSL/OpenSSL/SSL.pyi
```diff
@@ -115,25 +117,63 @@ def SSLeay_version(type: int) -> str: ...
...
+    def recv(self, bufsiz: int, flags: int | None = ...) -> bytes: ...
+    read = recv
+    def recv_into(self, buffer: bytearray, nbytes: int | None = ..., flags: int | None = ...) -> int: ...
```
AlexWaygood, line 132: Does `buffer` here have to be a bytearray, or can any writeable, mutable buffer do?

stubs/pyOpenSSL/OpenSSL/SSL.pyi
```diff
@@ -115,25 +117,63 @@ def SSLeay_version(type: int) -> str: ...
...
+    def get_verified_chain(self) -> list[X509] | None: ...
     def bio_read(self, bufsiz: int) -> bytes: ...
     def bio_write(self, buf: bytes) -> int: ...
-    def recv(self, bufsiz: int, flags: int | None = ...) -> bytes: ...
-    def sendall(self, buf: bytes, flags: int = ...) -> int: ...
+    def bio_shutdown(self) -> None: ...
+    def renegotiate(self) -> bool: ...
+    def renegotiate_pending(self) -> bool: ...
+    def total_renegotiations(self) -> int: ...
+    def makefile(self, *args: Any, **kwargs: Any) -> NoReturn: ...
```
AlexWaygood, line 148: Maybe it would be better to leave this one out, since it just unconditionally raises an exception?

```suggestion
```

stubs/pyOpenSSL/OpenSSL/SSL.pyi
```diff
@@ -115,25 +117,63 @@ def SSLeay_version(type: int) -> str: ...
...
     def get_cipher_name(self) -> str | None: ...
+    def get_cipher_bits(self) -> int | None: ...
+    def get_cipher_version(self) -> str | None: ...
     def get_protocol_version_name(self) -> str: ...
+    def get_protocol_version(self) -> int: ...
     def get_shutdown(self) -> int: ...
+    def set_shutdown(self, state: int) -> None: ...
+    def get_state_string(self) -> bytes: ...
+    def server_random(self) -> bytes: ...
+    def client_random(self) -> bytes: ...
```
AlexWaygood, line 162: https://github.com/pyca/pyopenssl/blob/38f9b4e524ac6479d57021bba2270df84d85b672/src/OpenSSL/SSL.py#L2277

```suggestion
    def client_random(self) -> bytes | None: ...
```

stubs/pyOpenSSL/OpenSSL/SSL.pyi
```diff
@@ -115,25 +117,63 @@ def SSLeay_version(type: int) -> str: ...
...
+    def master_key(self) -> bytes: ...
```
AlexWaygood, line 163: https://github.com/pyca/pyopenssl/blob/38f9b4e524ac6479d57021bba2270df84d85b672/src/OpenSSL/SSL.py#L2293

```suggestion
    def master_key(self) -> bytes | None: ...
```

stubs/pyOpenSSL/OpenSSL/SSL.pyi
```diff
@@ -115,25 +117,63 @@ def SSLeay_version(type: int) -> str: ...
...
+    def export_keying_material(self, label: Any, olen: Any, context: Any = ...) -> Any: ...  # TODO: type, see RFC-5705
```
AlexWaygood, line 164: Why annotate these if we're adding the TODO comment? Why not just leave it unannotated, which serves as its own implicit TODO?

AlexWaygood (changes requested):

stubs/pyOpenSSL/OpenSSL/SSL.pyi
sobolevn: I don't know 🙂 
pyright was failing without them:
- https://github.com/python/typeshed/runs/7480728281?check_suite_focus=true
- https://github.com/python/typeshed/runs/7480713768?check_suite_focus=true

```
  /home/runner/work/typeshed/typeshed/stubs/pyOpenSSL/OpenSSL/SSL.pyi:166:38 - error: Type of parameter "label" is unknown (reportUnknownParameterType)
  /home/runner/work/typeshed/typeshed/stubs/pyOpenSSL/OpenSSL/SSL.pyi:166:45 - error: Type of parameter "olen" is unknown (reportUnknownParameterType)
  /home/runner/work/typeshed/typeshed/stubs/pyOpenSSL/OpenSSL/SSL.pyi:166:9 - error: Return type is unknown (reportUnknownParameterType)
```

stubs/pyOpenSSL/OpenSSL/SSL.pyi
sobolevn: Good catch!

stubs/pyOpenSSL/OpenSSL/SSL.pyi
sobolevn: Makes sense!

stubs/pyOpenSSL/OpenSSL/SSL.pyi
sobolevn: It has to support:
- `len()`
- ```python
        # This strange line is all to avoid a memory copy. The buffer protocol
        # should allow us to assign a CFFI buffer to the LHS of this line, but
        # on CPython 3.3+ that segfaults. As a workaround, we can temporarily
        # wrap it in a memoryview.
        buffer[:result] = memoryview(_ffi.buffer(buf, result))
   ```

For example, `list` will also work:

```python
>>> l = []
>>> l[:4] = memoryview(b'abcd')
>>> l
[97, 98, 99, 100]
```

Akuli:
```suggestion
    def recv_into(self, buffer: Sequence[int], nbytes: int | None = ..., flags: int | None = ...) -> int: ...
```

stubs/pyOpenSSL/OpenSSL/SSL.pyi
Akuli: You can exclude the file in `pyrightconfig.stricter.json`.

stubs/pyOpenSSL/OpenSSL/SSL.pyi
sobolevn: `Sequence` does not support `[] =`. `MutableSequence` does:

```python
from typing import Sequence, MutableSequence

s: Sequence[int]
m: MutableSequence[int]

s[:1] = memoryview(b'1')  # Unsupported target for indexed assignment ("Sequence[int]")
m[:1] = memoryview(b'1')
```

Akuli: I see, the buffer is what is being assigned into. Technically `MutableSequence` isn't enough, because a mutable sequence could be indexable but not sliceable:

```
>>> d = collections.deque('asd')
>>> d[1] = 'hi'
>>> d[1:] = 'lol'
TypeError: sequence index must be integer, not 'slice'
>>> 
```

I think `bytearray` and `MutableSequence` are both fine in practice. This method is mostly for compatibility with non-SSL sockets anyway, and doesn't perform any better than `.recv()` like it can for non-SSL sockets.

stubs/pyOpenSSL/OpenSSL/SSL.pyi
AlexWaygood: Or we can use `_typeshed.Incomplete`, which is an alias for `Any` but serves as an easily greppable TODO marker :)

stubs/pyOpenSSL/OpenSSL/SSL.pyi
AlexWaygood: `deque` is also a bit of a special case here — most classes that claim to be `MutableSequence`s should really support slicing

AlexWaygood (approved):
LGTM, thanks!

sobolevn:
Thanks everyone!

PR #8375, opened 2022-07-23
Improve `redis.crc` types

sobolevn:
Ref: https://github.com/redis/redis-py/blob/master/redis/crc.py


AlexWaygood (approved):
I [really don't like](https://github.com/PyCQA/flake8-pyi/pull/235) type aliases with names ending in `T`, but this is what the source code does in this instance ://

PR #8376, opened 2022-07-23
Allow Path for create_unix_server()

Dreamsorcerer:


AlexWaygood (dismissed):
This looks good:

- The docs do indeed promise that `Path` objects are accepted in 3.7+: https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.create_unix_server
- The one concrete class in the stdlib that implements `create_unix_server` does a call to `os.fspath()`, indicating that it supports `PathLike` objects as well as `str`s: https://github.com/python/cpython/blob/3c94d3395edbcd299cbdacd09660ed88e654eeef/Lib/asyncio/unix_events.py#L254

It seems like the docstring of `asyncio.events.AbstractEventLoop.create_unix_server` is quite out of date, but that's irrelevant to this PR: https://github.com/python/cpython/blob/3c94d3395edbcd299cbdacd09660ed88e654eeef/Lib/asyncio/events.py#L407

Could you use `_typeshed.StrPath` instead of `str | Path`, however?

AlexWaygood (approved):

PR #8380, opened 2022-07-24
Add missing `None` return to `RawConfigParser.get`

### LINKED ISSUE #7476: `RawConfigParser.get` should return `str | None`, not `str`
The `get` method on `RawConfigParser` method should return `str | None`, along with all with the other `get*` variants. The same applies to `SectionProxy`.

kkirsche:
Fix #7476

This adds support to RawConfigParser.get to return None as seen on these lines in cPython:
https://github.com/python/cpython/blob/main/Lib/configparser.py#L773-L774


Akuli:
Presumably `ConfigParser` shouldn't inherit this behaviour. Usually you just want to parse a config file with `ConfigParser`, and in that case getting `None` seems to be impossible:

```
>>> p = configparser.ConfigParser()
>>> p.read_string("[foo]\nempty =")
>>> p.get("foo", "empty")
''
>>> p.get("foo", "asdasd")
   ...
configparser.NoOptionError: No option 'asdasd' in section: 'foo'
```

A couple ways to fix this:
- Add another `get` method to `ConfigParser` that doesn't have `| None`.
- Return `str | Any`, meaning that the user has to be prepared to get a string, but they can handle other types too if they want.

You will also have to adjust `SectionProxy`. Maybe make it generic, so that it can be a `MutableMapping[str, str]` for `ConfigParser` but `MutableMapping[str, str | None]` for `RawConfigParser`?

kkirsche:
I've added the overloads for `ConfigParser`. Regarding `SectionProxy`, if I make it a generic that can accept either, `SectionProxy` no longer is compatible with `Mapping[str, str]

Regarding making section proxy generic, I can do that but it seems to really ripple throughout as we have to adjust the `__getitem__` implementations, and related TypeAliases like `_Parser` to handle how it can change.

Akuli:
Indeed, the inheritance is a bit unfortunate: `RawConfigParser` should behave like `MutableMapping[str, SectionProxy[str | None]]` but its subclass `ConfigParser` should behave like `MutableMapping[str, SectionProxy[str]]`. I don't think there's any way to express this. Maybe we should just return `str | Any` from `SectionProxy`'s `.get()`, which makes it usable for both `ConfigParser` and `RawConfigParser`, although you won't get errors if you use `RawConfigParser` and you forget to handle `None`.

stdlib/configparser.pyi
```diff
@@ -143,9 +143,11 @@ class RawConfigParser(_Parser):
     ) -> _T: ...
     # This is incompatible with MutableMapping so we ignore the type
     @overload  # type: ignore[override]
-    def get(self, section: str, option: str, *, raw: bool = ..., vars: _Section | None = ...) -> str: ...
+    def get(self, section: str, option: str, *, raw: bool = ..., vars: _Section | None = ...) -> str | None: ...
     @overload
-    def get(self, section: str, option: str, *, raw: bool = ..., vars: _Section | None = ..., fallback: _T) -> str | _T: ...
+    def get(
+        self, section: str, option: str, *, raw: bool = ..., vars: _Section | None = ..., fallback: _T
```
Akuli, line 149: Looking at zproject/config.py in zulip (from last mypy_primer output), mypy seems to think that `_T` is `str`, and complains when you do `fallback=None`. I will try to fix that.

```suggestion
        self, section: str, option: str, *, raw: bool = ..., vars: _Section | None = ..., fallback: _T | None
```

stdlib/configparser.pyi
```diff
@@ -182,7 +189,7 @@ class SectionProxy(MutableMapping[str, str]):
         vars: _Section | None = ...,
         _impl: Any | None = ...,
         **kwargs: Any,
-    ) -> str: ...
+    ) -> str | Any: ...
```
Akuli, line 192:
```suggestion
    ) -> str | Any: ...  # can be None in RawConfigParser's sections
```

Akuli:
This still seems a bit too disruptive to me: all mypy_primer results above (you need to click "1 similar comment") are usages of `RawConfigParser` objects that probably don't contain any `None`s, because to get a `None` value in `RawConfigParser`, you would have to manually insert it instead of just reading a file.

We have a couple options to choose from:
- Just tell everyone to stop using `RawConfigParser`. This isn't great, because it's not necessarily the fault of whoever gets mypy errors. For example, flake8 provides a `RawConfigParser` object for plugin authors.
- Use `| Any` in `RawConfigParser` too

stdlib/configparser.pyi
Akuli: I reverted this. Returning a possibly `None` value is good if you pass `fallback=None`.

Akuli (approved):

Akuli:
Thanks!

PR #8381, opened 2022-07-24
Add return type to MySQLdb.Connection.__enter__

kkirsche:
This pull request adds the return type, as seen in the link below, to the `mysqlclient.MySQLdb.connections:Connection.__enter__`
https://github.com/PyMySQL/mysqlclient/blob/main/MySQLdb/connections.py#L230-L231


AlexWaygood (approved):
Thanks!

PR #8382, opened 2022-07-24
Update types of exceptions in mysqlclient’s connection class

kkirsche:
This updates the types of the classes used in mysqlclient’s Connection class. On this page we can see that by default each value is set to the exception itself.

https://github.com/PyMySQL/mysqlclient/blob/main/MySQLdb/connections.py#L321-L330

These were imported from `._exceptions` as seen here:
https://github.com/PyMySQL/mysqlclient/blob/main/MySQLdb/connections.py#L10-L21

If we look at `_exceptions.py`, we can see that these are all `Exception`'s, so we can type these as `BaseException` to align with the expectation that these will be raisable objects.
https://github.com/PyMySQL/mysqlclient/blob/main/MySQLdb/_exceptions.py


kkirsche:
Closed because it’s unclear why this doesn’t work with the test suite

AlexWaygood:
> Closed because it’s unclear why this doesn’t work with the test suite

The defaults are set to the exception types themselves, whereas your annotations would imply that they're being set to instances of the exception types :)

AlexWaygood (approved):
The other alternative would be to mimic the runtime exactly and do:

```python
class Connection(_mysql.connection):
    Warning = Warning
    Error = Error
    # etc...
```

But I think that might risk type checkers thinking that these are type aliases, which they aren't. And this seems fine!

PR #8383, opened 2022-07-24
Add return types to PyYaml’s Serializer `open`, `close`, `serialize`

kkirsche:
As seen here:
https://github.com/yaml/pyyaml/blob/master/lib/yaml/serializer.py#L27-L41

Neither `open` nor `close` methods return, so `None` added as the return type.

Finally, serialize accepts a Node and does not return as seen here:
https://github.com/yaml/pyyaml/blob/master/lib/yaml/serializer.py#L46-L58


AlexWaygood (approved):
Thanks!

PR #5545, opened 2021-05-28
Rework socket

srittau:
* Extract _socket.pyi from socket.pyi.
* Extract _socket.socket from socket.socket.
* Fix socket.family annotation.
* Annotate SocketIO properly.
* SocketType is an alias of _socket.socket.
* Sort items in socket.pyi in the same order as in socket.py.
* Remove socket.EINTR.
* Add socket.errorTab (Windows only) and _socket.dup().
* Use _typeshed.WriteableBuffer instead of custom alias.
* Use built-in generics.
* Use socket.socket rather than socket.SocketType in stdlib and third-party stubs.

PR #8385, opened 2022-07-24
`socket`: add missing`socket.accept()`.

leven-cn:


leven-cn:
Removed usage of socket.SocketType: <https://github.com/python/typeshed/pull/5545>.

PR #8386, opened 2022-07-25
Type `http.client.HTTPConnection.sock`

sobolevn:
It can be either:
- `None`: https://github.com/python/cpython/blob/310f94871a923f6cf3dc9259e732ce2376578b26/Lib/http/client.py#L856
- or `socket.socket`: https://github.com/python/cpython/blob/310f94871a923f6cf3dc9259e732ce2376578b26/Lib/http/client.py#L955-L956

### DIFF
--- a/stdlib/http/client.pyi
+++ b/stdlib/http/client.pyi
@@ -157,7 +157,7 @@ class HTTPConnection:
     timeout: float | None
     host: str
     port: int
-    sock: socket | None
+    sock: socket | Any  # can be `None` if `.connect()` was not called
     def __init__(
         self,
         host: str,
--- a/stdlib/http/client.pyi
+++ b/stdlib/http/client.pyi
@@ -157,7 +157,7 @@ class HTTPConnection:
     timeout: float | None
     host: str
     port: int
-    sock: Any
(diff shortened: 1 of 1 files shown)


sobolevn:
Oh, I see. Looks like the idea behind `Any` here, because different subtypes can use different `sock` types. What's is the policy in this case?

Do we fallback to `Any` or do we require `# type: ignore[override]` for such cases?

AlexWaygood:
This looks quite disruptive, and in practice, users can probably be confident that `sock` will be a `socket` object after `.connect()` has been called.

Maybe do `sock: socket | Any` instead, with a comment explaining that it could be `None` if `connect()` has not yet been called?

Akuli:
Please add something like `sock: SSLSocket | Any` into `HTTPSConnection`. HTTPS connections never use unencrypted `socket.socket`s.

sobolevn:
Sure! Source: https://github.com/python/cpython/blob/310f94871a923f6cf3dc9259e732ce2376578b26/Lib/http/client.py#L1456-L1457

srittau (approved):

PR #8388, opened 2022-07-25
SQLAlchemy: Fix annotations for Query.{update,delete}; Fix #8387

### LINKED ISSUE #8387: Incorrect hints for SQLAlchemy parameter `synchronize_session`
The parameter `synchronize_session` to the functions `sqlalchemy.orm.Query.{update,delete}` is typed wrongly as a `str`, when it should be a `Union[str, bool]` or `Literal[False, 'evaluate', 'fetch']`. 

The documentation for [update](https://docs.sqlalchemy.org/en/14/orm/query.html#sqlalchemy.orm.Query.update) and [delete](https://docs.sqlalchemy.org/en/14/orm/query.html#sqlalchemy.orm.Query.delete) demonstrates that both strings and bools should be accepted.

This issue was also reported to SQLAlchemy and fixed upstream: https://github.com/sqlalchemy/sqlalchemy/issues/8280 

The discussion in that issue also mentioned https://github.com/sqlalchemy/sqlalchemy2-stubs, and I'm also aware of the Dropbox stubs. Are those independent from the stubs in this repository?

akselaase:
The two functions `Query.update` and `Query.delete` admits a parameter `synchronize_session`, which is currently annotated as a `str`. However, the correct type annotation is `Literal[False, "evaluate", "fetch"]`.

This has been fixed in the upstream repository sqlalchemy/sqlalchemy#8280.


srittau (approved):
Thanks!

PR #8389, opened 2022-07-25
refactor: add type hints to flask_sqlalchemy.model

kkirsche:
The metaclasses are typed to match the following `type.__init__` overload:
https://github.com/python/typeshed/blob/master/stdlib/builtins.pyi#L167

As query's are generic in SQLAlchemy, this begins to make `Model` generic for use in the `flask_sqlalchemy.__init__` file once this is completed.

This also updates `__table_cls__` from `NameMetaMixin`.  This returns a table from:
1. https://github.com/pallets-eco/flask-sqlalchemy/blob/main/src/flask_sqlalchemy/model.py#L78
2. https://github.com/pallets-eco/flask-sqlalchemy/blob/main/src/flask_sqlalchemy/model.py#L86
3. https://github.com/pallets-eco/flask-sqlalchemy/blob/main/src/flask_sqlalchemy/model.py#L94

and then has a `None` return if this branch hits:
https://github.com/pallets-eco/flask-sqlalchemy/blob/main/src/flask_sqlalchemy/model.py#L90-L92

and the code runs to completion.


kkirsche:
I wasn't sure if this should be `Generic[_Model]` or if I should be using `Query[Self]` instead to indicate that it's the current model

stubs/Flask-SQLAlchemy/flask_sqlalchemy/model.pyi
```diff
@@ -1,7 +1,11 @@
 from re import Pattern
-from typing import Any
+from typing import Any, Generic, TypeVar
 
+from sqlalchemy import Table
 from sqlalchemy.ext.declarative import DeclarativeMeta
+from sqlalchemy.orm import Query
+
+_Model = TypeVar("_Model")
```
AlexWaygood, line 8: Style nit: I'd prefer we called this `TypeVar` `_ModelT`; the `T` suffix indicates unambiguously that it's a `TypeVar` rather than a `TypeAlias`.

```suggestion
_ModelT = TypeVar("_ModelT")
```

stubs/Flask-SQLAlchemy/flask_sqlalchemy/model.pyi
```diff
@@ -10,14 +14,14 @@ camelcase_re: Pattern[str]
 def camel_to_snake_case(name: str) -> str: ...
 
 class NameMetaMixin(type):
-    def __init__(cls, name, bases, d) -> None: ...
-    def __table_cls__(cls, *args, **kwargs): ...
+    def __init__(cls, name: str, bases: tuple[type, ...], d: dict[str, Any]) -> None: ...
+    def __table_cls__(cls, *args: Any, **kwargs: Any) -> Table | None: ...
```
AlexWaygood, line 18: This doesn't look right to me; I don't think this method can _really_ be called with _any_ arbitrary `*args, **kwargs` combination. The "true" signature looks like it might be quite complicated; I'd rather we just left it unannotated for now, as an explicit marker that there's still stuff to do here in future PRs:

```suggestion
    def __table_cls__(cls, *args, **kwargs) -> Table | None: ...
```

stubs/Flask-SQLAlchemy/flask_sqlalchemy/model.pyi
```diff
@@ -10,14 +14,14 @@ camelcase_re: Pattern[str]
...
+    def __init__(cls, name: str, bases: tuple[type, ...], d: dict[str, Any]) -> None: ...
 
 class DefaultMeta(NameMetaMixin, BindMetaMixin, DeclarativeMeta): ...
 
-class Model:
-    query_class: Any | None
-    query: Any | None
+class Model(Generic[_Model]):
+    query_class: type[Query[_Model]] | None
+    query: Query[_Model] | None
```
AlexWaygood, line 27: @srittau, does making `Model` generic look correct to you?

stubs/Flask-SQLAlchemy/flask_sqlalchemy/model.pyi
kkirsche: Sorry about that. Makes sense.

stubs/Flask-SQLAlchemy/flask_sqlalchemy/model.pyi
kkirsche: That's fair. This was before I had seen in `requests` that the group will sometimes use `*` to name out upstream `*args` and `**kwargs` usage. 

Table Class is logically a good candidate for `ParamSpec` as it's described here:
https://docs.sqlalchemy.org/en/14/orm/declarative_config.html#table-cls

But I'm not sure if the current ParamSpec description allows for capturing a class's init as a parameter spec or not

stubs/Flask-SQLAlchemy/flask_sqlalchemy/model.pyi
kkirsche: Class implementation can be seen here, if it helps:
https://github.com/pallets-eco/flask-sqlalchemy/blob/main/src/flask_sqlalchemy/model.py#L115

stubs/Flask-SQLAlchemy/flask_sqlalchemy/model.pyi
AlexWaygood: No need to be sorry! :)

stubs/Flask-SQLAlchemy/flask_sqlalchemy/model.pyi
AlexWaygood: We can leave that to another PR :)

stubs/Flask-SQLAlchemy/flask_sqlalchemy/model.pyi
srittau: I have no experience with flask-sqlalchemy, but considering that `sqlalchemy.orm.query` is generic: probably.

AlexWaygood: Let's YOLO it then. If it causes a problem for somebody, we'll know soon enough.

AlexWaygood (approved):

PR #8390, opened 2022-07-25
refactor: add type hints to flask_sqlalchemy.__init__.Pagination

kkirsche:
This adds the following changes to add type hints to Flask-SQLAlchemy:

1. Moves the TypeVar for easier access
2. Move `BaseQuery` above `Pagination`
3. Add return value of `Pagination[_T]` to `BaseQuery.paginate`
4. Make `Pagination` generic
5. Add type hints for `Query` to be `BaseQuery[_T] | None`
    * Accepts value: https://github.com/pallets-eco/flask-sqlalchemy/blob/main/src/flask_sqlalchemy/__init__.py#L353
    * Existing checks for `None` are in place:
        * https://github.com/pallets-eco/flask-sqlalchemy/blob/main/src/flask_sqlalchemy/__init__.py#L375
        * https://github.com/pallets-eco/flask-sqlalchemy/blob/main/src/flask_sqlalchemy/__init__.py#L394
        * Otherwise it assumes it is a BaseQuery object
6. Provides type hints for page, per_page, and total
    * `page` is `int`
        * https://github.com/pallets-eco/flask-sqlalchemy/blob/main/src/flask_sqlalchemy/__init__.py#L384
        * https://github.com/pallets-eco/flask-sqlalchemy/blob/main/src/flask_sqlalchemy/__init__.py#L408
    * `total` is `int | None`
        * https://github.com/pallets-eco/flask-sqlalchemy/blob/main/src/flask_sqlalchemy/__init__.py#L366 None Check otherwise used as `int` in math call
        * `per_page` is `int`
            * https://github.com/pallets-eco/flask-sqlalchemy/blob/main/src/flask_sqlalchemy/__init__.py#L366 for `int` comparisons and conversions to `float`
7. Updates method types as seen on https://github.com/pallets-eco/flask-sqlalchemy/blob/main/src/flask_sqlalchemy/__init__.py


stubs/Flask-SQLAlchemy/flask_sqlalchemy/__init__.pyi
```diff
@@ -16,37 +18,37 @@ class SignallingSession(Session):
...
+    per_page: int
+    total: int | None
     items: Any
-    def __init__(self, query, page, per_page, total, items) -> None: ...
+    def __init__(self, query: BaseQuery[_T] | None, page: int, per_page: int, total: int | None, items) -> None: ...
     @property
-    def pages(self): ...
-    def prev(self, error_out: bool = ...): ...
+    def pages(self) -> int: ...
+    def prev(self: Self, error_out: bool = ...) -> Self: ...
```
AlexWaygood, line 39: `BaseQuery.paginate()` is hardcoded to return a `Pagination` object, so `Pagination.prev()` won't necessarily return whatever the type of `self` is.

To illustrate, if we have a subclass of `Pagination`, `Foo`:

```python
class Foo(Pagination): ...
```

Then the type of `self` in the `Foo.prev()` will be `Foo`, but the returned type will _not_ be `Foo` (unless `Pagination.prev()` is overridden in the subclass); it will still be `Pagination`.

```suggestion
    def prev(self, error_out: bool = ...) -> Pagination: ...
```

stubs/Flask-SQLAlchemy/flask_sqlalchemy/__init__.pyi
```diff
@@ -16,37 +18,37 @@ class SignallingSession(Session):
...
     @property
-    def prev_num(self): ...
+    def prev_num(self) -> int | None: ...
     @property
-    def has_prev(self): ...
-    def next(self, error_out: bool = ...): ...
+    def has_prev(self) -> bool: ...
+    def next(self: Self, error_out: bool = ...) -> Self: ...
```
AlexWaygood, line 44:
```suggestion
    def next(self, error_out: bool = ...) -> Pagination: ...
```

stubs/Flask-SQLAlchemy/flask_sqlalchemy/__init__.pyi
kkirsche: Ah, good point. I hadn't considered the sub-class use case like that. Thanks for explaining that :)

AlexWaygood: No problem :)

AlexWaygood (approved):

PR #8392, opened 2022-07-25
refactor: Add type hints to mysqlclient.times

kkirsche:
This merge request adds missing types to `mysqlclient.times` as seen here:
https://github.com/PyMySQL/mysqlclient/blob/dac24e7b05b83eb1511e25e3cd8f8c20b7fbe112/MySQLdb/times.py

The `ticks` values were chosen based on the type signature of `time.localtime`


stubs/mysqlclient/MySQLdb/times.pyi
```diff
@@ -9,18 +10,18 @@ Timestamp = datetime
...
-def DateTime2literal(d, c): ...
-def DateTimeDelta2literal(d, c): ...
+def format_TIMEDELTA(v: timedelta) -> str: ...
+def format_TIMESTAMP(d: datetime) -> str: ...
+def DateTime_or_None(s: str) -> datetime | None: ...
+def TimeDelta_or_None(s: str) -> timedelta | None: ...
+def Time_or_None(s: str) -> time | None: ...
+def Date_or_None(s: str) -> date | None: ...
+def DateTime2literal(d: datetime, c: Any) -> Any: ...
+def DateTimeDelta2literal(d: datetime, c: Any) -> Any: ...
```
srittau, line 27: As `c` is ignored, it can be literally anything. Confusingly, this is expressed using `object`, not `Any`. The latter is used as a marker for "can't be expressed with the type system". Also, both of there return `str.

```suggestion
def DateTime2literal(d: datetime, c: object) -> str: ...
def DateTimeDelta2literal(d: datetime, c: object) -> str: ...
```

srittau (changes requested):
Thanks! Remarks below.

kkirsche:
Thank you for the feedback, I appreciate the clarification about the use of `object` here. Updated in [e2968df](https://github.com/python/typeshed/pull/8392/commits/e2968df1fe08cf62d6603617d3d4338b2b8a099c) per your feedback

srittau (approved):

PR #8396, opened 2022-07-25
refactor: add missing imports and vars to requests.compat

kkirsche:
This adds missing items from https://github.com/psf/requests/blob/main/requests/compat.py


AlexWaygood (changes requested):
I'd prefer if we used re-exports here (from `urllib.parse import urlparse as urlparse`, etc), like the runtime does. I also don't think we should be including private attributes like `_ver` — we generally avoid including private attributes unless there's a specific reason to include them.

kkirsche:
Fixed via 4afd35c

AlexWaygood (approved):

PR #8400, opened 2022-07-25
Fix #8369: BeautifulSoup iterator returns Tag

### LINKED ISSUE #8369: beautifulsoup: iterating over `bs4.BeautifulSoup` returns `Tag`, but Mypy thinks that `bs4.element.PageElement` type is returned
### The code is correct, but fails with Mypy:
```python
html_doc = """
<html><head><title>The Dormouse's story</title></head>
<body>
<p class="title"><b>The Dormouse's story</b></p>

<p class="story">Once upon a time there were three little sisters; and their names were
<a href="http://example.com/elsie" class="sister" id="link1">Elsie</a>,
<a href="http://example.com/lacie" class="sister" id="link2">Lacie</a> and
<a href="http://example.com/tillie" class="sister" id="link3">Tillie</a>;
and they lived at the bottom of a well.</p>

<p class="story">...</p>
"""
from bs4 import BeautifulSoup, SoupStrainer

soup = BeautifulSoup(html_doc, 'lxml', parse_only=SoupStrainer("a"))
reveal_type(soup)
for el in soup:
    print(type(el))
    reveal_type(el)
    print(el.get("href"))
```

### Mypy output
```bash
_local.py:17:13: note: Revealed type is "bs4.BeautifulSoup"
_local.py:19:17: note: Revealed type is "bs4.element.PageElement"
_local.py:20:11: error: "PageElement" has no attribute "get"  [attr-defined]
        print(el.get("href"))
              ^
```
### Python output if you run the code
```bash
<class 'bs4.element.Tag'>
http://example.com/elsie
<class 'bs4.element.Tag'>
http://example.com/lacie
<class 'bs4.element.Tag'>
http://example.com/tillie
```
### Expected Mypy output

```
soup = BeautifulSoup(...
for item in soup:
  reveal_type(item)  #  <-- type must be `Tag`
```

### Versions
types-beautifulsoup4 = "4.11.4"
beautifulsoup4 = "4.11.1"

kkirsche:
Fix #8369 

The `BeautifulSoup` class inherits from `Tag` here:
https://bazaar.launchpad.net/~leonardr/beautifulsoup/bs4/view/head:/bs4/__init__.py#L74

The `__iter__` method is of `Tag`, which is the one used by `BeautifulSoup` is seen here:
https://bazaar.launchpad.net/~leonardr/beautifulsoup/bs4/view/head:/bs4/element.py#L1521

This iterates over `self.contents`, which is a `list[PageElement]` so this makes both unions with `Tag` to ensure coverage still handles the situations where a non-`Tag` `PageElement` is found


AlexWaygood:
You probably don't need to do quite so many merge commits — they mean that those of us who are subscribed to all activity on this repo get quite a few notifications :)

kkirsche:
Oops, sorry about that! I'll keep that in mind moving forward.

AlexWaygood:
> Oops, sorry about that! I'll keep that in mind moving forward.

No worries! :)

hauntsaninja (changes requested):
I'm a little confused by this, since `Tag`s are `PageElement`s.

I think the kind of error in #8369 is often unavoidable. It might be safe to change `BeautifulSoup.__iter__` (I just don't know), but `Tag.__iter__` can most certainly return e.g. a `NavigableString`

```
html_doc = "some document"
from bs4 import BeautifulSoup
soup = BeautifulSoup(html_doc)
tag = soup.contents[0].contents[0].contents[0]
print(type(tag))
list(tag)[0].get("href")
```

kkirsche:
OK, so it sounds like rather than a merge request, the correct approach would be for the code in the issue to do type-narrowing via isinstance checks or a structural check that the item supports the `.get` method. Is that understanding what you've said correctly?

Akuli:
Ideally passing `parse_only=SoupStrainer("a")` to `BeautifulSoup(...)` would return a soup of `Tag`s. I think it is possible to express that with generics and an overloaded `__init__`. But making `BeautifulSoup` generic might not be a good idea if people are currently using it in type hints.

AIGeneratedUsername:
Defining
```python
def __iter__(self) -> Iterator[Tag | NavigableString]: ...
```
will produce much more meaningful output by a type checker, then the current `-> Iterator[PageElement`].

It will be clear that you may get any of these two types, and the `Tag` has `.get()` method.

### Current (bad) output 
`error: "PageElement" has no attribute "get"  [attr-defined]`

This output makes you think that you have no chances to get something with `.get()` method. Very confusing.

### New (good) output

Something like:

`"NavigableString" of two "Tag | NavigableString" has no attribute "get" [union-attr]`

This new warning makes things clear: you need to check for `NavigableString` and then do not call `.get()` for it.

kkirsche:
Happy to make changes, but I may be in over my head with getting this right. If there is a consensus for the correct update I'm happy to do it, otherwise I'll probably close this as I'm not confident enough with my understanding of BeautifulSoup to handle the `overload` calls at this time

PR #8401, opened 2022-07-25
Add missing `__version__` to `chardet.__init__`

kkirsche:
`__version__` is part of the `__all__` of `chardet` here:
https://github.com/chardet/chardet/blob/main/chardet/__init__.py#L27

`__version__` is a string, as seen on:
https://github.com/chardet/chardet/blob/main/chardet/version.py#L8

### DIFF
--- a/stubs/chardet/chardet/__init__.pyi
+++ b/stubs/chardet/chardet/__init__.pyi
@@ -1,6 +1,7 @@
 import sys
 
 from .universaldetector import UniversalDetector as UniversalDetector, _FinalResultType, _IntermediateResultType
+from .version import __version__ as __version__, VERSION as VERSION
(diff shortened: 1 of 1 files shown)


AlexWaygood (changes requested):
We should re-export both `__version__` and `VERSION` from the `version` submodule, like the runtime does. We should delete these symbols from `__init__.pyi` and replace them with `from .version import __version__ as __version__, VERSION as VERSION` (the weird `from x import foo as foo` syntax indicates to type checkers that these imported symbols are public, unlike most imports in stub files)

kkirsche:
Fixed via ce37d19

AlexWaygood (approved):

PR #8402, opened 2022-07-25
Add missing urllib3.__init__ type annotations/hints

kkirsche:
This adds the missing `__author__` field and types `add_stderr_logger` per https://github.com/urllib3/urllib3/blob/1.26.11/src/urllib3/__init__.py


stubs/urllib3/urllib3/__init__.pyi
```diff
@@ -23,5 +24,5 @@ Retry = retry.Retry
 class NullHandler(logging.Handler):
     def emit(self, record): ...
 
-def add_stderr_logger(level=...): ...
+def add_stderr_logger(level: int = ...) -> logging.StreamHandler[Any]: ...
```
AlexWaygood, line 27:
```suggestion
def add_stderr_logger(level: int = ...) -> logging.StreamHandler[TextIO]: ...
```

The source code indicates that the `StreamHandler` is only ever created with 0 arguments: https://github.com/urllib3/urllib3/blob/aa3def7d242525e6e854991247c4b68583d15135/src/urllib3/__init__.py#L56

That means we follow the code path of the first overload in typeshed's stub, meaning that a `StreamHandler[TextIO]` is returned:

https://github.com/python/typeshed/blob/d3c4cb105acac79e50cdaf72171dd455575c536c/stdlib/logging/__init__.pyi#L777

(You'll have to import `TextIO` from `io`)

AlexWaygood (changes requested):

stubs/urllib3/urllib3/__init__.pyi
kkirsche: Imported `TextIO` from typing (`io` doesn't re-export it) and fixed via 4660dd7 if I understood the feedback correctly.

Issue #6799, opened 2022-01-03
`multiprocessing` is missing several files


sobolevn:
Directory in CPython: https://github.com/python/cpython/tree/main/Lib/multiprocessing
Directory in typeshed: https://github.com/python/typeshed/tree/master/stdlib/multiprocessing

- [x] https://github.com/python/cpython/blob/main/Lib/multiprocessing/forkserver.py
- [x] https://github.com/python/cpython/blob/main/Lib/multiprocessing/heap.py
- [x] https://github.com/python/cpython/blob/main/Lib/multiprocessing/popen_fork.py
- [x] https://github.com/python/cpython/blob/main/Lib/multiprocessing/popen_forkserver.py
- [ ] https://github.com/python/cpython/blob/main/Lib/multiprocessing/popen_spawn_posix.py
- [ ] https://github.com/python/cpython/blob/main/Lib/multiprocessing/popen_spawn_win32.py
- [x] https://github.com/python/cpython/blob/main/Lib/multiprocessing/reduction.py
- [x] https://github.com/python/cpython/blob/main/Lib/multiprocessing/resource_sharer.py
- [x] https://github.com/python/cpython/blob/main/Lib/multiprocessing/resource_tracker.py
- [x] https://github.com/python/cpython/blob/main/Lib/multiprocessing/util.py

hauntsaninja:
Yup. All of these are undocumented, so relatively low pri (as per checklist in https://github.com/python/typeshed/issues/4545 , issue also contains a script to get list of missing modules in stdlib).

kkirsche:
Once the open merge requests are completed, I'll move onto the next series of files. I'm trying to avoid creating too many merge requests at any one time.

I seem to have missed where the script was in the linked issue to locate files/modules without stubs, but I'd be happy to address other modules if I can identify how to locate those.

kkirsche:
Moving onto last two hopefully today, depending on my meeting schedule. If all goes to plan this should be able to be closed by the end of the week

kkirsche:
This will be fixed when the remaining PR's are completed.

Note though that there are some additional opportunities in multiprocessing to expand stubs further, if interested

PR #8404, opened 2022-07-26
fix: Add missing multiprocessing.util types

kkirsche:
X-Ref: #6799

This adds best-effort types to multiprocessing.util to begin to address issue #6799

https://github.com/python/cpython/blob/main/Lib/multiprocessing/util.py


stdlib/multiprocessing/util.pyi
```diff
@@ -0,0 +1,122 @@
...
+]
+
+NOTSET: int = ...
+SUBDEBUG: int = ...
+DEBUG: int = ...
+INFO: int = ...
+SUBWARNING: int = ...
+
+LOGGER_NAME: str = ...
+DEFAULT_LOGGING_FORMAT: str = ...
```
AlexWaygood, line 36:
```suggestion
NOTSET: int
SUBDEBUG: int
DEBUG: int
INFO: int
SUBWARNING: int

LOGGER_NAME: str
DEFAULT_LOGGING_FORMAT: str
```

stdlib/multiprocessing/util.pyi
```diff
@@ -0,0 +1,122 @@
...
+
+_logger: Logger | None = ...
+_log_to_stderr: bool = ...
+
```
AlexWaygood, line 40:
```suggestion
```

stdlib/multiprocessing/util.pyi
```diff
@@ -0,0 +1,122 @@
...
+def sub_debug(msg: object, *args: object) -> None: ...
+def debug(msg: object, *args: object) -> None: ...
+def info(msg: object, *args: object) -> None: ...
+def sub_warning(msg: object, *args: object) -> None: ...
+def get_logger() -> Logger: ...
+def log_to_stderr(level: int | None = ...) -> Logger: ...
+def _platform_supports_abstract_sockets() -> bool: ...
```
AlexWaygood, line 47:
```suggestion
```

stdlib/multiprocessing/util.pyi
```diff
@@ -0,0 +1,122 @@
...
+def is_abstract_socket_namespace(address: str | bytes | None) -> bool: ...
+
+abstract_sockets_supported: bool = ...
```
AlexWaygood, line 50:
```suggestion
abstract_sockets_supported: bool
```

stdlib/multiprocessing/util.pyi
```diff
@@ -0,0 +1,122 @@
...
+
+def _remove_temp_dir(rmtree: Callable[[StrOrBytesPath], None], tempdir: StrOrBytesPath) -> None: ...
```
AlexWaygood, line 52:
```suggestion
```

stdlib/multiprocessing/util.pyi
```diff
@@ -0,0 +1,122 @@
...
+def get_temp_dir() -> str: ...
+
+_afterfork_registry: WeakValueDictionary[tuple[int, int, Callable[[Any], Any]], object]
+_afterfork_counter: count[int]
+
+def _run_after_forkers() -> None: ...
```
AlexWaygood, line 58:
```suggestion
```

stdlib/multiprocessing/util.pyi
```diff
@@ -0,0 +1,122 @@
...
+def register_after_fork(obj: object, func: Callable[[Any], Any]) -> None: ...
```
AlexWaygood, line 59: I'd prefer to use `Incomplete` for the `obj` parameter and for the parameter of the `func` callable, since it isn't clear to me what those types are meant to be (unless you have a suggestion of how this function is meant to be used? :)

However, I'd prefer to use `object` for the return-type of the `func` callable, since it seems the returned value is ignored at runtime in the one place the functions registered in the `_afterfork_registry` are used: https://github.com/python/cpython/blob/f40bc7fa49f8d137f9b38f7a14569e9a9bdbed07/Lib/multiprocessing/util.py#L167

```suggestion
def register_after_fork(obj: Incomplete, func: Callable[[Incomplete], object]) -> None: ...
```

stdlib/multiprocessing/util.pyi
```diff
@@ -0,0 +1,122 @@
...
+
+_finalizer_registry: dict[tuple[int | None, int], Finalize] = ...
+_finalizer_counter: count[int] = ...
+
```
AlexWaygood, line 63:
```suggestion
```

stdlib/multiprocessing/util.pyi
```diff
@@ -0,0 +1,122 @@
...
+class Finalize:
+    _args: Any
+    _kwargs: dict[Any, Any]
+    _callback: Callable[[Any], Any]
+    _key: tuple[int | None, int]
+    _pid: int
+
```
AlexWaygood, line 70:
```suggestion
```

stdlib/multiprocessing/util.pyi
```diff
@@ -0,0 +1,122 @@
...
+    def __init__(
+        self,
+        obj: object | None,
```
AlexWaygood, line 73: Are you sure a user is _meant_ to be passing in literally any object? Or would `_typeshed.Incomplete` be better here? :)

stdlib/multiprocessing/util.pyi
```diff
@@ -0,0 +1,122 @@
...
+        callback: Callable[[Any], Any],
```
AlexWaygood, line 74: Your annotation currently means "A callable that can only accept a single argument of `Any` type, and returns a value of `Any` type. In this situation, a better annotation is "Any callable with any number of parameters of `Any` type, that returns a value of `Any` type". To express that, we use a `...` in place of a parameter list.

```suggestion
        callback: Callable[..., Any],
```

stdlib/multiprocessing/util.pyi
```diff
@@ -0,0 +1,122 @@
...
+        args: Any = ...,
+        kwargs: dict[Any, Any] | None = ...,
```
AlexWaygood, line 76: These are passed as `*args, **kwargs` to `self._callback` in the `__call__` method: https://github.com/python/cpython/blob/f40bc7fa49f8d137f9b38f7a14569e9a9bdbed07/Lib/multiprocessing/util.py#L224

That means that `args` has to be a `Sequence`, and `kwargs` must have `str` keys. I think `kwargs` can probably be any `Mapping`, however; it probably doesn't _have_ to be a `dict`.

```suggestion
        args: Sequence[Any] = ...,
        kwargs: Mapping[str, Any] | None = ...,
```

stdlib/multiprocessing/util.pyi
```diff
@@ -0,0 +1,122 @@
...
+        exitpriority: int | None = ...,
+    ) -> None: ...
+    def __call__(
+        self,
+        wr: Any = ...,
```
AlexWaygood, line 81: The parameter is unused, so literally any object can be passed int

```suggestion
        wr: object = ...,
```

stdlib/multiprocessing/util.pyi
```diff
@@ -0,0 +1,122 @@
...
+        # Need to bind these locally because the globals can have
+        # been cleared at shutdown
```
AlexWaygood, line 83: We don't need these comments in the stub, I don't think

```suggestion
```

stdlib/multiprocessing/util.pyi
```diff
@@ -0,0 +1,122 @@
...
+        _finalizer_registry: dict[Any, Any] = ...,
```
AlexWaygood, line 84: Be as permissive as possible in parameter annotations. It probably doesn't _have_ to be a `dict`.

```suggestion
        _finalizer_registry: MutableMapping[Incomplete, Incomplete] = ...,
```

stdlib/multiprocessing/util.pyi
```diff
@@ -0,0 +1,122 @@
...
+        sub_debug=...,
```
AlexWaygood, line 85:
```suggestion
        sub_debug: Callable[..., object] = ...,
```

stdlib/multiprocessing/util.pyi
```diff
@@ -0,0 +1,122 @@
...
+        getpid: Callable[[], int] = ...,
+    ) -> Any: ...
+    def cancel(self) -> None: ...
+    def still_active(self) -> bool: ...
+
+def _run_finalizers(minpriority: Finalize | None = ...) -> None: ...
```
AlexWaygood, line 91:
```suggestion
```

stdlib/multiprocessing/util.pyi
```diff
@@ -0,0 +1,122 @@
...
+_exiting: bool = ...
+
+def _exit_function(
+    info: Callable[[object, object], None] = ...,
+    debug: Callable[[object, object], None] = ...,
+    _run_finalizers: Callable[[Finalize | None], None] = ...,
+    active_children: Callable[[], list[process.BaseProcess]] = ...,
+    current_process: Callable[[], process.BaseProcess] = ...,
+) -> None: ...
+
```
AlexWaygood, line 103:
```suggestion
```

stdlib/multiprocessing/util.pyi
```diff
@@ -0,0 +1,122 @@
...
+class ForkAwareThreadLock:
+    _lock: threading.Lock
```
AlexWaygood, line 105:
```suggestion
```

stdlib/multiprocessing/util.pyi
```diff
@@ -0,0 +1,122 @@
...
+    acquire: Callable[[bool, float], bool]
+    release: Callable[[], None]
+    def __init__(self) -> None: ...
+    def __enter__(self) -> bool: ...
+    def __exit__(self, *args: object) -> None: ...
+
+class ForkAwareLocal(threading.local):
+    def __init__(self) -> None: ...
+    def __reduce__(self: Self) -> tuple[type[Self], tuple[()]]: ...
```
AlexWaygood, line 114: We tend not to include `__reduce__` methods in typeshed, because they sorta feel like implementation details

```suggestion
```

stdlib/multiprocessing/util.pyi
```diff
@@ -0,0 +1,122 @@
...
+
+MAXFD: int = ...
```
AlexWaygood, line 116:
```suggestion
MAXFD: int
```

stdlib/multiprocessing/util.pyi
```diff
@@ -0,0 +1,122 @@
...
+
+def close_all_fds_except(fds: Iterable[int]) -> None: ...
+def _close_stdin() -> None: ...
+def _flush_std_streams() -> None: ...
```
AlexWaygood, line 120:
```suggestion
```

stdlib/multiprocessing/util.pyi
```diff
@@ -0,0 +1,122 @@
...
+def spawnv_passfds(path: bytes, args: Sequence[str], passfds: Sequence[int]) -> int: ...
+def _cleanup_tests() -> None: ...
```
AlexWaygood, line 122:
```suggestion
```

AlexWaygood (changes requested):

stdlib/multiprocessing/util.pyi
kkirsche: I wasn't sure of the intent, but the reason I chose object was twofold — first, the variable name implied the type of `object` and second, as the value is used with `id` which was the signature `id(__obj: object, /) -> int` my logic was that it made sense to also use `object` for the upstream, as that's the only behavioral requirement.


Happy to change it to incomplete though if you prefer for `obj`

AlexWaygood: `obj` is a fairly common variable name, and doesn't _always_ imply an argument type of `object`; I think I'd prefer to play it safe and use `Incomplete` here :)

stdlib/multiprocessing/util.pyi
kkirsche: Putting in as `Incomplete | None` as the signature for `obj: object | None` was inferred based on the requirements of  `weakref.ref` from how it was used and checks performed on it explicitly include `is not None`

kkirsche: This may make sense to overload, as `obj` and `exitpriority` go hand in hand in the implementation here:

https://github.com/python/cpython/blob/f40bc7fa49f8d137f9b38f7a14569e9a9bdbed07/Lib/multiprocessing/util.py#L192-L195

So potentially having one where `obj` None and `exitpriority` is required and one where it's `Incomplete`

stdlib/multiprocessing/util.pyi
kkirsche: Ah, ok. The various uses of ellipsis sometimes trip me up. Thank you for clarifying that.

stdlib/multiprocessing/util.pyi
AlexWaygood: > This may make sense to overload

I think it probably isn't worth the complexity, given that this entire module is undocumented :)

stdlib/multiprocessing/util.pyi
AlexWaygood: It annoys me the way `...` is overloaded to mean so many different, unrelated things in the type system, but what can you do 🤷‍♂️

stdlib/multiprocessing/util.pyi
kkirsche: perfectly fair :) updating and will push that in a moment

stdlib/multiprocessing/util.pyi
kkirsche: That's reasonable, I'll remove the todo comment next to that

kkirsche:
To the best of my knowledge, all feedback has been addressed. Appreciate the review and feedback. This will help ensure that the remaining pull requests for other undocumented sections include less of the same fixes as was required in this instance. Thanks 👍

stdlib/multiprocessing/util.pyi
AlexWaygood: @kkirsche this hasn't been addressed yet :)

kkirsche: Oops, sorry about that. Addressing this now

kkirsche: This is fixed via [1f76c39](https://github.com/python/typeshed/pull/8404/commits/1f76c393255f41c6ead06b8c5a56a783017719fd)

AlexWaygood: Thanks!

stdlib/multiprocessing/util.pyi
```diff
@@ -0,0 +1,81 @@
...
+    def __enter__(self) -> bool: ...
+    def __exit__(self, *args: object) -> None: ...
+
+class ForkAwareLocal(threading.local):
+    def __init__(self) -> None: ...
+
+MAXFD: int
+
+def close_all_fds_except(fds: Iterable[int]) -> None: ...
+def spawnv_passfds(path: bytes, args: Sequence[str], passfds: Sequence[int]) -> int: ...
```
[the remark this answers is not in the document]
kkirsche: Updated 👍

AlexWaygood (approved):

stdlib/multiprocessing/util.pyi
```diff
@@ -0,0 +1,86 @@
+import threading
+from _typeshed import Incomplete, ReadableBuffer, SupportsTrunc
+from collections.abc import Callable, Iterable, Mapping, MutableMapping, Sequence
+from logging import Logger
+from typing import Any, SupportsIndex, SupportsInt
```
AlexWaygood, line 5:
```suggestion
from typing import Any, SupportsInt
from typing_extensions import SupportsIndex
```

PR #8405, opened 2022-07-26
fix: Add missing multiprocessing/resource-tracker type hints

kkirsche:

This adds best-effort types to `multiprocessing.resource_tracker` to begin to address issue https://github.com/python/typeshed/issues/6799

https://github.com/python/cpython/blob/main/Lib/multiprocessing/resource_tracker.py


kkirsche:
It sounds like this is Python 3.8+ only from this error:

```
error: multiprocessing.resource_tracker failed to import, ModuleNotFoundError: No module named 'multiprocessing.resource_tracker'
Stub: at line 1
MypyFile:1(
  stdlib/multiprocessing/resource_tracker.pyi)
Runtime:
MISSING
```

Before making changes though, would someone be able to confirm the correct way to address this? I want to avoid unnecessary emails and whatnot as I try to work it out myself

JelleZijlstra:
You may need to add it to the VERSIONS file if it's new in 3.8.

kkirsche:
Seems to be the case:
https://github.com/python/cpython/blob/3.8/Lib/multiprocessing/resource_tracker.py
vs. 404 for
https://github.com/python/cpython/blob/3.7/Lib/multiprocessing/resource_tracker.py

Fixed in d5294ead

stdlib/multiprocessing/resource_tracker.pyi
```diff
@@ -0,0 +1,34 @@
...
+from threading import Lock
+from typing import Any
+
+__all__ = ["ensure_running", "register", "unregister"]
+
+_HAVE_SIGMASK: bool = ...
+_IGNORED_SIGNALS: tuple[Signals, Signals] = ...
+
+_CLEANUP_FUNCS: dict[str, Callable[[Any], Any]] = ...
+
```
AlexWaygood, line 13: Same as https://github.com/python/typeshed/pull/8406#discussion_r931170447

```suggestion
```

stdlib/multiprocessing/resource_tracker.pyi
```diff
@@ -0,0 +1,34 @@
...
+class ResourceTracker:
+    _lock: Lock
+    _fd: int | None
+    _pid: int | None
+
```
AlexWaygood, line 18:
```suggestion
```

stdlib/multiprocessing/resource_tracker.pyi
```diff
@@ -0,0 +1,34 @@
...
+    def __init__(self) -> None: ...
+    def _stop(self) -> None: ...
```
AlexWaygood, line 20:
```suggestion
```

stdlib/multiprocessing/resource_tracker.pyi
```diff
@@ -0,0 +1,34 @@
...
+    def getfd(self) -> int | None: ...
+    def ensure_running(self) -> None: ...
+    def _check_alive(self) -> bool: ...
```
AlexWaygood, line 23:
```suggestion
```

stdlib/multiprocessing/resource_tracker.pyi
```diff
@@ -0,0 +1,34 @@
...
+    def register(self, name: Sized, rtype: object) -> None: ...
+    def unregister(self, name: Sized, rtype: object) -> None: ...
+    def _send(self, cmd: object, name: Sized, rtype: object) -> None: ...
```
AlexWaygood, line 26:
```suggestion
```

stdlib/multiprocessing/resource_tracker.pyi
```diff
@@ -0,0 +1,34 @@
...
+
+_resource_tracker: ResourceTracker = ...
+ensure_running: Callable[[], None] = ...
+register: Callable[[Sized, object], None] = ...
+unregister: Callable[[Sized, object], None] = ...
+getfd: Callable[[], int | None] = ...
```
AlexWaygood, line 32: You had the right idea in #8406 for how to tackle bound methods :)

```suggestion
ensure_running = _resource_tracker.ensure_running
register = _resource_tracker.register
unregister = _resource_tracker.unregister
getfd = _resource_tracker.getfd
```

AlexWaygood (changes requested):

stdlib/multiprocessing/resource_tracker.pyi
```diff
@@ -0,0 +1,19 @@
...
+from collections.abc import Sized
+
+__all__ = ["ensure_running", "register", "unregister"]
+
+class ResourceTracker:
+    def __init__(self) -> None: ...
+    def getfd(self) -> int | None: ...
+    def ensure_running(self) -> None: ...
+    def register(self, name: Sized, rtype: object) -> None: ...
+    def unregister(self, name: Sized, rtype: object) -> None: ...
```
AlexWaygood, line 11: It's true that these methods can be passed literally any object without raising an exception, so your instinct here to annotate these parameters with `object` was good. However, I feel like it's unlikely that you're actually _meant_ to be passing literally any object to these methods. I think it's more likely that you're meant to be passing some kind of "resource" to these methods (the "shared memory segments, semaphores, etc." referenced in the block comment at the top of the module). I don't know enough about `multiprocessing` to know what the _true_ type is, so I'd rather we went for `_typeshed.Incomplete` for now, which serves as an explicit marker of "we're not quite sure what this type should be".

```suggestion
    def register(self, name: Sized, rtype: Incomplete) -> None: ...
    def unregister(self, name: Sized, rtype: Incomplete) -> None: ...
```

kkirsche: That makes sense, updated :) thanks for explaining that 👍

AlexWaygood (approved):

PR #8406, opened 2022-07-26
fix: add missing multiprocessing.forkserver type hints

kkirsche:

This adds best-effort types to `multiprocessing.forkserver` to begin to address issue https://github.com/python/typeshed/issues/6799

https://github.com/python/cpython/blob/main/Lib/multiprocessing/forkserver.py

This also adds a missing argument to spawn._main as seen here:
https://github.com/python/cpython/blob/main/Lib/multiprocessing/spawn.py#L124


stdlib/multiprocessing/forkserver.pyi
```diff
@@ -0,0 +1,44 @@
...
+    _forkserver_pid: int | None
+    _inherited_fds: Sequence[int] | None
+    _lock: Lock
+    _preload_modules: list[str]
+
+    def __init__(self) -> None: ...
+    def _stop(self) -> None: ...
+    def _stop_unlocked(self) -> None: ...
+    def set_forkserver_preload(self, modules_names: list[str]) -> None: ...
+    def get_inherited_fds(self) -> Sequence[int] | None: ...
```
AlexWaygood, line 25: For stubs, be as permissive as possible for argument annotations, and as precise as possible for return annotations. The docstring says this returns a "list of fds", and "list" is more precise than "Sequence".

```suggestion
    def get_inherited_fds(self) -> list[int] | None: ...
```

stdlib/multiprocessing/forkserver.pyi
```diff
@@ -0,0 +1,44 @@
...
+    def ensure_running(self) -> None: ...
+
+def main(
+    listener_fd: int | None,
+    alive_r: FileDescriptorLike,
+    preload: Sequence[str],
+    main_path: str | None = ...,
+    sys_path: object | None = ...,
+) -> None: ...
+def _serve_one(child_r: int, fds: Iterable[int], unused_fds: Iterable[int], handlers: Mapping[_SIGNUM, _HANDLER]) -> Any: ...
```
AlexWaygood, line 36:
```suggestion
```

stdlib/multiprocessing/forkserver.pyi
```diff
@@ -0,0 +1,44 @@
...
+def read_signed(fd: int) -> Any: ...
+def write_signed(fd: int, n: Any) -> None: ...
```
AlexWaygood, line 38: `SIGNED_STRUCT` has `'q'` as the `format_value` passed to its `__init__` (https://github.com/python/cpython/blob/2833f3798dc3647e850b303a4d0fa00609a0ae9b/Lib/multiprocessing/forkserver.py#L26), and that means that only `int`s can be passed to its `.pack` method (docs here: https://docs.python.org/3/library/struct.html#format-characters)

```suggestion
def write_signed(fd: int, n: int) -> None: ...
```

AlexWaygood (changes requested):

stdlib/multiprocessing/forkserver.pyi
kkirsche: Ah, sorry about that. Thank you for explaining how you got to that conclusion!

AlexWaygood: > Ah, sorry about that. Thank you for explaining how you got to that conclusion!

No need to be sorry! Typing is hard :)

If the reason for something one of us is saying is every unclear, please don't be afraid to ask for clarification btw!

AlexWaygood:
Ah, the stubtest failures on 3.11 aren't anything you need to address in this PR; it's because there's a new 3.11 beta version that's been released.

AlexWaygood (approved):

PR #8407, opened 2022-07-26
fix: Add missing (private) types and methods for ensurepip

kkirsche:
This adds missing types/variables as seen in:
https://github.com/python/cpython/blob/main/Lib/ensurepip/__init__.py

and adds the previously unincluded:
https://github.com/python/cpython/blob/main/Lib/ensurepip/_uninstall.py


JelleZijlstra:
That's a lot of private functions. What is the motivation for adding these?

kkirsche:
Completeness — After reviewing the CONTRIBUTING.md documentation I didn't see any notes about private types or methods being omitted as a desired trait of typeshed types, so this began to address that gap.

JelleZijlstra:
https://github.com/python/typeshed/blob/master/CONTRIBUTING.md says "Other objects may be included if they are being used in practice or if they are not prefixed with an underscore." The aim here is to add private names if people need them, it's not to get every private function included.

kkirsche:
Ah, sorry about that — I must have misunderstood it. Sorry about that.

PR #8408, opened 2022-07-27
Fix return value for optparse.parse_args

pgiraud:
Apparently accidentaly changed in 158293257e77fe3ca6900490e0699fc7f739d02c

There's no reason why the return value would be different for each overload.


srittau:
This seems incorrect to me. To what is `AnyStr` constrained in that overload? Can that overload ever return `bytes`?

hauntsaninja:
If args is None, it uses sys.argv, which is list[str]

PR #8411, opened 2022-07-27
Stubtest: allowlist all of `distutils` on 3.11

### LINKED ISSUE #8410: Many stubtest_stdlib failures for `distutils` on 3.11b5
See e.g. https://github.com/python/typeshed/runs/7543946914?check_suite_focus=true

I have no idea what might be causing this, but it means CI is currently red for all new PRs we receive.

AlexWaygood:
Fixes #8410


AlexWaygood:
I hoped we could just add a `distutils\..*` entry to the `3.11` allowlist. But that doesn't work, because stubtest complains about all sorts of unused allowlist entries from the `py3_common` allowlist. So those need to be removed from the `py3_common` allowlist and pasted into the {`3.10`, `3.9`, `3.8`, `3.7`} allowlists instead.

hauntsaninja (approved):
Thanks for fixing!

In case it's useful in the future, there's the `x` -> `(x)?` trick if you want to avoid unused allowlist errors. A hidden feature of stubtest is that it will not complain about an unused allowlist entry if it's a regex that could match the empty string. In this case, we could have used it to mark all the distutils things in py3_common.txt.

AlexWaygood:
> In case it's useful in the future, there's the `x` -> `(x)?` trick if you want to avoid unused allowlist errors. A hidden feature of stubtest is that it will not complain about an unused allowlist entry if it's a regex that could match the empty string. In this case, we could have used it to mark all the distutils things in py3_common.txt.

Yah, I considered this, but it felt vaguely immoral for this situation somehow `¯\_(ツ)_/¯`

PR #8412, opened 2022-07-27
Add missing `multiprocessing.heap` type stubs

kkirsche:

This adds best-effort types to multiprocessing.heap to continue to address issue https://github.com/python/typeshed/issues/6799

https://github.com/python/cpython/blob/main/Lib/multiprocessing/heap.py

Some `Incomplete` types in other locations of the `multiprocessing` module make this difficult to complete without first working through the remaining untyped areas, and then going back through with those improvements to other incomplete areas of the multiprocessing module.


kkirsche:
The commit history includes the private values for easier reviewing

AlexWaygood:
A few thoughts on the `Arena` class:

- We usually don't include `pickle` methods as they're basically implementation details, so I would remove `__setstate__` and `__getstate__`.
- It's _basically_ the same interface on Windows as on Posix, so I don't think the whole class needs to be behind the `sys.platform` check in the stub, even though that's how the runtime does it.

I'd change it to this:

```python
class Arena:
    buffer: mmap
    size: int
    if sys.platform == "win32":
        name: str
        def __init__(self, size: int) -> None: ...
    else:
        fd: int
        def __init__(self, size: int, fd: int = ...) -> None: ...
```

stdlib/multiprocessing/heap.pyi
```diff
@@ -0,0 +1,32 @@
...
+    size: int
+    buffer: mmap
+    if sys.platform == "win32":
+        name: str
+        def __init__(self, size: int) -> None: ...
+    else:
+        fd: int
+        def __init__(self, size: int, fd: int = ...) -> None: ...
+
+if not sys.platform == "win32":
```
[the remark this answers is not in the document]
kkirsche: Interesting, why this choice? This seems to go against most recommendations for python comparisons I've seen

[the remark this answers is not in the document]
AlexWaygood: It does? I haven't seen those recommendations, and I'd be curious to know why they recommend against using the `!=` operator :)

Anyway, this is the standard typeshed style for this kind of guard :)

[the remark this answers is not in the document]
kkirsche: No worries, sounds good. I'll look if I can find the related documentation I'm thinking of about the application of de morgan's laws in relationship to the `not` and `!=`/`==` operators.

stdlib/multiprocessing/heap.pyi
```diff
@@ -0,0 +1,32 @@
...
+    if sys.platform == "win32":
+        name: str
+        def __init__(self, size: int) -> None: ...
+    else:
+        fd: int
+        def __init__(self, size: int, fd: int = ...) -> None: ...
+
+if sys.platform != "win32":
+    def reduce_arena(a: Arena) -> tuple[Callable[[int, Incomplete], Arena], tuple[int, Incomplete]]: ...
+    def rebuild_arena(size: int, dupfd: Incomplete) -> Arena: ...
```
AlexWaygood, line 23: For the `dupfd` argument of `rebuild_arena`: we know it has to have a `detach()` method, and we know the `detach()` method needs to return an `int` (as the returned value of the `detach()` method is passed as the second argument to the `Arena` constructor). So you can do something like this:

```suggestion
    class _SupportsDetach(Protocol):
        def detach(self) -> int: ...

    def reduce_arena(a: Arena) -> tuple[Callable[[int, _SupportsDetach], Arena], tuple[int, Incomplete]]: ...
    def rebuild_arena(size: int, dupfd: _SupportsDetach) -> Arena: ...
```

stdlib/multiprocessing/heap.pyi
```diff
@@ -0,0 +1,32 @@
+import sys
+from _typeshed import Incomplete
+from collections.abc import Callable
+from mmap import mmap
+from typing_extensions import TypeAlias
+
+_Block: TypeAlias = tuple[Arena, int, int]
```
AlexWaygood, line 7: Nit: can we move this type alias declaration below the `Arena` class definition? Type checkers support forward references in stub files, but it's making me do a double take every time I see this :)

kkirsche: Sure, no worries. I think of it more in terms of stub organization (TypeAliases first, then implementation using them), but not a big deal at all, especially for short stubs like this

AlexWaygood: > Sure, no worries. I think of it more in terms of stub organization (TypeAliases first, then implementation using them), but not a big deal at all, especially for short stubs like this

In general, yes, but it's nice to avoid forward references where possible :)

stdlib/multiprocessing/heap.pyi
```diff
@@ -0,0 +1,35 @@
...
+        def __init__(self, size: int, fd: int = ...) -> None: ...
+
+_Block: TypeAlias = tuple[Arena, int, int]
+
+if sys.platform != "win32":
+    class _SupportsDetach(Protocol):
+        def detach(self) -> int: ...
+
+    def reduce_arena(a: Arena) -> tuple[Callable[[int, Incomplete], Arena], tuple[int, Incomplete]]: ...
+    def rebuild_arena(size: int, dupfd: Incomplete) -> Arena: ...
```
AlexWaygood, line 26: There's no point defining the protocol if you don't actually use it in the type hints :)

```suggestion
    def reduce_arena(a: Arena) -> tuple[Callable[[int, _SupportsDetach], Arena], tuple[int, Incomplete]]: ...
    def rebuild_arena(size: int, dupfd: _SupportsDetach) -> Arena: ...
```

kkirsche: D'oh! Fixing it, sorry about that

AlexWaygood (approved):

stdlib/multiprocessing/heap.pyi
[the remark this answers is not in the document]
Akuli: De Morgan's laws are meaningful when you have `not` in front of `and` or `or`. For example,
```python3
if not (foo == 0 or bar == 0):
    print("nothing is zero")
```
is (by De Morgan's law) same as
```python3
if foo != 0 and bar != 0:
    print("nothing is zero")
```
The first way to write this can be cleaner with two or more things to check, here `foo` and `bar`. You only have one thing, `sys.platform`.

PR #8413, opened 2022-07-27
feat: add support for `multiprocessing.resource_sharer` type hints

kkirsche:

This adds best-effort types to multiprocessing.sharer to continue to address issue https://github.com/python/typeshed/issues/6799

https://github.com/python/cpython/blob/main/Lib/multiprocessing/resource_sharer.py


kkirsche:
Commit history includes all types including private if it helps for review

AlexWaygood:
> Commit history includes all types including private if it helps for review

It doesn't, please delete them :)

stdlib/multiprocessing/resource_sharer.pyi
```diff
@@ -0,0 +1,51 @@
...
+    def __init__(self) -> None: ...
+    def register(self, send: Callable[[Connection, int], None], close: Callable[[], None]) -> tuple[_Address, int]: ...
+    @staticmethod
+    def get_connection(ident: tuple[_Address, int]) -> Connection: ...
+    def stop(self, timeout: float | None = ...) -> None: ...
+    def _afterfork(self) -> None: ...
+    def _start(self) -> None: ...
+    def _serve(self) -> None: ...
+
+_resource_sharer: _ResourceSharer
```
AlexWaygood, line 50: This is the 1% of instances where you do need the `= ...` in a stub file: where flake8 complains at you that the name  is "undefined" (flake8 isn't fully compatible with stub files at the end of the day)

```suggestion
_resource_sharer: _ResourceSharer = ...
```

AlexWaygood (changes requested):
Please remove all the private things

kkirsche:
Sorry about that, the pre-commit bot prevented the push. Cleaning that up now

stdlib/multiprocessing/resource_sharer.pyi
```diff
@@ -0,0 +1,36 @@
...
+
+# https://docs.python.org/3/library/multiprocessing.html#address-formats
+_Address: TypeAlias = Union[str, tuple[str, int]]
+
+if sys.platform == "win32":
+    __all__ += ["DupSocket"]
+
+    class DupSocket:
+        def __init__(self, sock: socket) -> None: ...
+        def detach(self) -> bytes: ...
```
AlexWaygood, line 19: It returns whatever a call to `socket.fromshare()` returns: https://github.com/python/cpython/blob/bceb197947bbaebb11e01195bdce4f240fdf9332/Lib/multiprocessing/resource_sharer.py#L40

And the typeshed stub says that `socket.fromshare` returns a `socket`: https://github.com/python/typeshed/blob/177bbf7864b544880e7db8f898d14aa5c3561f07/stdlib/socket.pyi#L732

```suggestion
        def detach(self) -> socket: ...
```

stdlib/multiprocessing/resource_sharer.pyi
```diff
@@ -0,0 +1,36 @@
...
+
+else:
+    __all__ += ["DupFd"]
+
+    class DupFd:
+        def __init__(self, fd: int) -> None: ...
+        def detach(self) -> None: ...
```
AlexWaygood, line 26: It returns the first element whatever a call to `multiprocessing.reduction.recv_handle()` returns: https://github.com/python/cpython/blob/bceb197947bbaebb11e01195bdce4f240fdf9332/Lib/multiprocessing/reduction.py#L189

The current typeshed annotation for `multiprocessing.reduction.recv_handle()` (which was written by me :) is incorrect; I'll file a PR to fix it momentarily. I think it returns `list[int]`.

```suggestion
        def detach(self) -> int: ...
```

stdlib/multiprocessing/resource_sharer.pyi
```diff
@@ -0,0 +1,36 @@
...
+
+class _ResourceSharer:
+    def __init__(self) -> None: ...
+    def register(self, send: Callable[[Connection, int], None], close: Callable[[], None]) -> tuple[_Address, int]: ...
+    @staticmethod
+    def get_connection(ident: tuple[_Address, int]) -> Connection: ...
+    def stop(self, timeout: float | None = ...) -> None: ...
+
+_resource_sharer: _ResourceSharer = ...
+stop = _resource_sharer.stop
```
AlexWaygood, line 36: This all looks like a bit too much to maintain, considering it's a private class in an undocumented module. The only public think here is the `stop()` callable:

```suggestion
def stop(timeout: float | None = ...) -> None: ...
```

AlexWaygood (changes requested):

stdlib/multiprocessing/resource_sharer.pyi
kkirsche: I understand the point, though personally I think it's valuable to provide these types of things for developers who choose to start doing development on internals in more depth for, as without the private methods, for me at least, it really hinders learning the cpython stdlib codebase

AlexWaygood (approved):

stdlib/multiprocessing/resource_sharer.pyi
AlexWaygood: I agree that typeshed is a fantastic tool for learning about the stdlib codebase, but that's ultimately not its primary purpose.

The issue with including implementation details in the stub is that there are no backwards compatibility guarantees, so they're liable to change frequently, possibly even in micro releases of CPython. That just imposes too much of a maintenance burden on us as maintainers.

There's also a question of the usefulness of these implementation details for typechecking. End users of this module shouldn't be using implementation details of this module in their applications or libraries (because they're liable to change at any time), so it's good, in a way, if a typechecker emits an error when a user tries to import one of these implementation details.

PR #8416, opened 2022-07-28
Fix commented code due to previous bug (todo) in setuptools.command.test

kkirsche:
With https://github.com/python/mypy/pull/10884 merged and released, this should be safe to re-enable per the todo comment.

Source:
https://github.com/pypa/setuptools/blob/main/setuptools/command/test.py#L117


stubs/setuptools/setuptools/command/test.pyi
```diff
@@ -29,8 +29,8 @@ class test(Command):
     test_runner: Any
     def initialize_options(self) -> None: ...
     def finalize_options(self) -> None: ...
-    # TODO: uncomment once https://github.com/python/mypy/pull/10884 is released
-    # def test_args(self): ...
+    @NonDataProperty
+    def test_args(self) -> list[Any]: ...
```
hauntsaninja, line 33: I think this is `list[str]`?

hauntsaninja (commented):
Thanks for cleaning this up

hauntsaninja (approved):

PR #8417, opened 2022-07-28
Add third-party lief stubs

chinggg:
Add the initial stubs for `lief`, generated by `python3 scripts/create_baseline_stubs.py lief`

Fix https://github.com/lief-project/LIEF/issues/650

These errors should be fixed before PR get merged.
```log
Running black: black stubs/lief
error: cannot format stubs/lief/lief/MachO.pyi: Cannot parse: 40:6:     32BIT_BRANCH: ClassVar[ARM_RELOCATION] = ...
error: cannot format stubs/lief/lief/OAT.pyi: Cannot parse: 172:29:     def magic(self) -> Magic number: ...
error: cannot format stubs/lief/lief/__init__.pyi: Cannot parse: 215:15: class ItemsView[DictStringVersion]:
error: cannot format stubs/lief/lief/DEX.pyi: Cannot parse: 394:10: class lief.DEX.MapList.it_items_t:
error: cannot format stubs/lief/lief/ELF/__init__.pyi: Cannot parse: 455:35:     def eof_offset(self) -> Return the last offset used by the ELF binary according to both: ...
error: cannot format stubs/lief/lief/PE.pyi: Cannot parse: 1339:17:     machine: The target machine architecture
```

- [ ] `32BIT_BRANCH` starts with number, so it is not a valid identifier
- [ ] `Magic number` contains space, not valid
- [ ] `class ItemsView[DictStringVersion]` use `[]` when declaring custom class, not valid
- [ ] `class lief.DEX.MapList.it_items_t` contains dots (`.`), not valid
- [ ] `eof_offset` has wrong return type, which is more like comments
- [ ] `machine` in `PE.pyi` has wrong type annotation, which contains space


stubs/lief/lief/MachO.pyi
```diff
@@ -0,0 +1,1552 @@
...
+    def __ne__(self, arg0: int) -> bool: ...
+    def __setstate__(self, state: int) -> None: ...
+    @property
+    def name(self) -> str: ...
+    @property
+    def value(self) -> int: ...
+
+class ARM_RELOCATION:
+    __members__: ClassVar[dict] = ...  # read-only
+    32BIT_BRANCH: ClassVar[ARM_RELOCATION] = ...
```
srittau, line 40: This is not a valid identifier in Python. Does this exist at runtime?

stubs/lief/lief/OAT.pyi
```diff
@@ -0,0 +1,294 @@
...
+    @property
+    def jni_dlsym_lookup_offset(self) -> int: ...
+    @property
+    def key_value_size(self) -> int: ...
+    @property
+    def key_values(self) -> Header.it_key_values_t: ...
+    @property
+    def keys(self) -> List[HEADER_KEYS]: ...
+    @property
+    def magic(self) -> Magic number: ...
```
srittau, line 172: `Magic number` is also not valid.

stubs/lief/lief/DEX.pyi
```diff
@@ -0,0 +1,416 @@
...
+    @property
+    def dim(self) -> int: ...
+    @property
+    def type(self) -> Type.TYPES: ...
+    @property
+    def underlying_array_type(self) -> Type: ...
+    @property
+    def value(self) -> object: ...
+
+class lief.DEX.MapList.it_items_t:
```
srittau, line 394: Same problem here. `lief.DEX.MapList.it_items_t` is not a valid Python identifier/class name. Those can't have dots in them.

stubs/lief/lief/ELF/__init__.pyi
```diff
@@ -0,0 +1,3115 @@
...
+    @property
+    def dtor_functions(self) -> List[lief.Function]: ...
+    @property
+    def dynamic_entries(self) -> Binary.it_dynamic_entries: ...
+    @property
+    def dynamic_relocations(self) -> Binary.it_filter_relocation: ...
+    @property
+    def dynamic_symbols(self) -> Binary.it_symbols: ...
+    @property
+    def eof_offset(self) -> Return the last offset used by the ELF binary according to both: ...
```
srittau, line 455: I would suggest to either find out the correct type here or just leave it unannotated:

```suggestion
    def eof_offset(self): ...
```

srittau (commented):
Thank you for your contribution to typeshed. This is not a full review yet, just a few notes on the failures you mentioned.

One thing that might help: Just send a PR with just the result of `scripts/create_baseline_stubs.py` and maybe a bit of import cleanup. We can then build on that.

chinggg:
@srittau Thanks for your help! These errors have shown some common patterns so I am trying to fix them in a reasonable way.
> One thing that might help: Just send a PR with just the result of scripts/create_baseline_stubs.py and maybe a bit of import cleanup. We can then build on that.

The first commit https://github.com/python/typeshed/pull/8417/commits/f14343ea24968047c74f0e0e67774fa8d6727f5e is generated by scripts/create_baseline_stubs.py and it has many errors so I try to make a commit for each file to fix them. So you mean I just need the first commit and you can build on that **without error automatically**?

srittau:
Ah I see. The original commit indeed has lots of problems. Normally that script generates acceptable stubs, but not in this case, unfortunately. So please ignore my comment above.

But maybe the easiest solution in case of errors would be to just remove the problematic type annotation for now, if it is not an easy fix. This makes it easier for you to finish the stubs and for us to review the first version.

chinggg:
Now most of my change is to convert C++ enum to Python enum since the auto script cannot express it using Python Enum class. I think this conversion should be done automatically in the future, since we cannot just rely on manual efforts to keep up with the latest change.

AlexWaygood:
Unfortunately there are many CI checks failing here, and I don't think any of us currently have the bandwidth to help this PR pass all of our CI :(

I'm going to close this PR for now due to lack of activity, but I'm happy to reopen it if you're still interested in working on it :)

PR #8418, opened 2022-07-28
Remove unused protocol

hauntsaninja:


stdlib/concurrent/futures/_base.pyi
```diff
@@ -36,13 +35,6 @@ _T = TypeVar("_T")
 _T_co = TypeVar("_T_co", covariant=True)
```
JelleZijlstra, line 35:
```suggestion
```

JelleZijlstra (approved):

JelleZijlstra:
Maybe this could be a flake8-pyi check.

PR #8420, opened 2022-07-28
Add missing multiprocessing.popen_fork type stubs

kkirsche:

This adds best-effort types to multiprocessing.popen_fork to continue to address issue https://github.com/python/typeshed/issues/6799

https://github.com/python/cpython/blob/main/Lib/multiprocessing/popen_fork.py

Popen attributes in some cases may not exist, such as `sentinel` which is set [here](https://github.com/python/cpython/blob/30610d28374f5a9698d456cebf3ae496ac01af51/Lib/multiprocessing/popen_fork.py#L79) but not initialized during the `__init__`. As such, it may desirable for those attributes to be removed from the stub.


tests/stubtest_allowlists/win32.txt
```diff
@@ -8,6 +8,8 @@ mmap.PAGESIZE
 # alias for a class defined elsewhere,
 # mypy infers the variable has type `(*args) -> DupHandle` but stubtest infers the runtime type as <class DupHandle>
 multiprocessing.reduction.AbstractReducer.DupHandle
+# Multiprocessing.popen_fork exists on Windows but fails to import
+multiprocessing.popen_fork
```
AlexWaygood, line 12: Can you move this allowlist entry down to line 60 or so, so it's next to all the other modules that either don't exist or fail to import on Windows?

kkirsche: Sure, I had put it here so it was next to the other multiprocessing entry in the file

AlexWaygood: Yeah, but this section of the allowlist is for things that are _potentially_ fixable, either by changing the typeshed stub or by improving mypy/stubtest somehow. This entry doesn't belong in this section, as it's not something that's fixable; it's just a thing that we need stubtest to ignore on Windows :)

stdlib/multiprocessing/popen_fork.pyi
```diff
@@ -0,0 +1,22 @@
...
+
+from .process import BaseProcess
+from .util import Finalize
+
+if sys.platform != "win32":
+    __all__ = ["Popen"]
+
+    class Popen:
+        finalizer: Finalize | None
+        method: str
```
AlexWaygood, line 11:
```suggestion
        method: ClassVar[str]
```

stdlib/multiprocessing/popen_fork.pyi
```diff
@@ -0,0 +1,22 @@
...
+        pid: int | None
```
AlexWaygood, line 12: Why `| None`? It looks like it's set as the return value of a call to `os.fork()` in the `_launch()` method:  https://github.com/python/cpython/blob/bceb197947bbaebb11e01195bdce4f240fdf9332/Lib/multiprocessing/popen_fork.py#L66

Note that the `_launch()` method is called in the `__init__` method: https://github.com/python/cpython/blob/bceb197947bbaebb11e01195bdce4f240fdf9332/Lib/multiprocessing/popen_fork.py#L19

And the typeshed stub says that `os.fork()` returns an `int`: https://github.com/python/typeshed/blob/177bbf7864b544880e7db8f898d14aa5c3561f07/stdlib/os/__init__.pyi#L821

```suggestion
        pid: int
```

stdlib/multiprocessing/popen_fork.pyi
```diff
@@ -0,0 +1,22 @@
...
+        returncode: int | None
+        sentinel: int | None
```
AlexWaygood, line 14: Same here, why the `| None`? It's set to `parent_r` in `_launch()`:  https://github.com/python/cpython/blob/bceb197947bbaebb11e01195bdce4f240fdf9332/Lib/multiprocessing/popen_fork.py#L79

And `parent_r` is the first element of the two-element tuple returned by `os.pipe()`: https://github.com/python/cpython/blob/bceb197947bbaebb11e01195bdce4f240fdf9332/Lib/multiprocessing/popen_fork.py#L64

 The typeshed stub says `os.pipe()` returns `tuple[int, int]`: https://github.com/python/typeshed/blob/177bbf7864b544880e7db8f898d14aa5c3561f07/stdlib/os/__init__.pyi#L603

```suggestion
        sentinel: int
```

AlexWaygood (commented):
Mostly looks good

stdlib/multiprocessing/popen_fork.pyi
kkirsche: This is where I wasn't sure how to properly model what's occurring. There is no guarantee that the attribute will exist, as it's not initialized before that. So assuming the module is used properly, this should be an int, but other circumstances this would raise an attribute error. As a result, I modelled this with a `| None` to capture conceptually that it may not actually exist

stdlib/multiprocessing/popen_fork.pyi
kkirsche: This was similar to the reason explained above, but for `sentinel` it only is set when the pid already exists, so this was modelled with an `| None` to capture the risk of not having the attribute in all circumstances. I don't think it's the right modelling of it, but I wasn't sure how to capture that

stdlib/multiprocessing/popen_fork.pyi
AlexWaygood: Sometimes it will exist, sometimes it won't, but whatever the case, it will never be `None`, so `| None` is incorrect. Just do `pid: int`, and add a comment saying that it doesn't always exist. In these situations, we prioritise avoiding false positives over avoiding false negatives.

kkirsche: Ah, makes sense. I must have accidentally removed the implementation before that. Added the comment RE: `sentinel`

stdlib/multiprocessing/popen_fork.pyi
```diff
@@ -0,0 +1,23 @@
+import sys
+
+from typing_extensions import ClassVar
```
AlexWaygood, line 3:
```suggestion
from typing import ClassVar
```

AlexWaygood (approved):

Issue #4545, opened 2020-09-16
Missing stubs in Python 3.8 standard library


hauntsaninja:
If a module is undocumented, feel free to strike it off (https://github.com/python/typeshed/issues/1019#issuecomment-693265073).

Filtered list:
- [ ] ctypes.macholib
- [ ] ctypes.macholib.dyld
- [ ] ctypes.macholib.dylib
- [ ] ctypes.macholib.framework
- [ ] curses.has_key
- [ ] distutils.command.install_egg_info
- [ ] distutils.command.upload
- [ ] distutils.config
- [ ] distutils.msvc9compiler
- [ ] distutils.versionpredicate
- [ ] email.base64mime
- [ ] email.quoprimime
- [ ] json.scanner
- [ ] json.tool
- [ ] multiprocessing.forkserver
- [ ] multiprocessing.heap
- [ ] multiprocessing.popen_fork
- [ ] multiprocessing.popen_forkserver
- [ ] multiprocessing.popen_spawn_posix
- [ ] multiprocessing.popen_spawn_win32
- [ ] multiprocessing.reduction
- [ ] multiprocessing.resource_sharer
- [ ] multiprocessing.resource_tracker
- [ ] multiprocessing.util
- [ ] sqlite3.dump
- [ ] tkinter.colorchooser
- [ ] tkinter.dnd
- [ ] tkinter.scrolledtext
- [ ] tkinter.simpledialog
- [ ] tkinter.tix
- [ ] xml.dom.domreg
- [ ] xml.dom.expatbuilder
- [ ] xml.dom.minicompat
- [ ] xml.dom.xmlbuilder
- [ ] xml.sax.expatreader

Full list:
<details>

```
{'_abc',
 '_asyncio',
 '_blake2',
 '_bz2',
 '_codecs_cn',
 '_codecs_hk',
 '_codecs_iso2022',
 '_codecs_jp',
 '_codecs_kr',
 '_codecs_tw',
 '_collections',
 '_collections_abc',
 '_contextvars',
 '_crypt',
 '_ctypes',
 '_curses_panel',
 '_datetime',
 '_dbm',
 '_elementtree',
 '_functools',
 '_gdbm',
 '_hashlib',
 '_io',
 '_locale',
 '_lsprof',
 '_lzma',
 '_md5',
 '_multibytecodec',
 '_multiprocessing',
 '_opcode',
 '_pickle',
 '_posixshmem',
 '_pyio',
 '_queue',
 '_scproxy',
[...]

PR #8421, opened 2022-07-28
fix: add missing curses.has_key stub

kkirsche:
X-Ref: https://github.com/python/typeshed/issues/4545

This adds the `curses.has_key.pyi` stub to complete the curses module (in terms of missing stub files) as outlined in #4545.

Source:
https://github.com/python/cpython/blob/main/Lib/curses/has_key.py

EDIT:
include source file


AlexWaygood:
It imports `_curses`, and `_curses` doesn't exist on Windows, so you'll have to do the same thing as in #8420. You should add it to this section of the allowlist: https://github.com/python/typeshed/blob/804628dc7da81e47e44a53f7d9989fa770cac7c9/tests/stubtest_allowlists/win32.txt#L59-L63

kkirsche:
Fixed via 2664310

AlexWaygood (approved):

PR #8422, opened 2022-07-28
`multiprocessing.reduction`: small fixes

AlexWaygood:
I had to look at the CPython source for this module while reviewing #8413, so: here's two small things I noticed while studying it.


stdlib/multiprocessing/reduction.pyi
```diff
@@ -43,9 +44,9 @@ else:
     else:
         ACKNOWLEDGE: Literal[False]
 
-    def recvfds(sock, size): ...
+    def recvfds(sock: socket, size: int) -> list[int]: ...
     def send_handle(conn, handle, destination_pid) -> None: ...
-    def recv_handle(conn) -> None: ...
+    def recv_handle(conn) -> int: ...
```
srittau, line 49:
```python
    from _typeshed import HasFileno
    def recv_handle(conn: HasFileno) -> int: ...
```

AlexWaygood: Nice.

srittau (approved):

PR #8425, opened 2022-07-28
fix: add missing multiprocessing.popen_forkserver.pyi stubs

kkirsche:

This adds best-effort types to multiprocessing.popen_fork to continue to address issue https://github.com/python/typeshed/issues/6799

https://github.com/python/cpython/blob/main/Lib/multiprocessing/popen_forkserver.py

Note that the forkserver version inherits from `popen_fork`, so `_launch` is called explicitly as was the case there. This only includes new or overriding attributes on the `Popen` class, ignoring attributes like `pid` which are identical with the parent. `sentinel` is defined here even though it matches `popen_fork.Popen` because it doesn't have the same condition as the comment in the other stub.

### DIFF
--- a/tests/stubtest_allowlists/win32.txt
+++ b/tests/stubtest_allowlists/win32.txt
@@ -56,8 +56,9 @@ syslog
 termios
 xxlimited
 
-# Multiprocessing.popen_fork exists on Windows but fails to import
+# Multiprocessing.popen_fork and popen_forkserver exists on Windows but fails to import
 multiprocessing.popen_fork
+multiprocessing.popen_forkserver
 
 # Modules that rely on _curses
 curses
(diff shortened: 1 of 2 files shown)


tests/stubtest_allowlists/win32.txt
```diff
@@ -56,8 +56,9 @@ syslog
 termios
 xxlimited
 
-# Multiprocessing.popen_fork exists on Windows but fails to import
+# Multiprocessing.popen_fork and popen_forkserver exists on Windows but fails to import
```
AlexWaygood, line 59: Grammar

```suggestion
# multiprocessing.popen_fork and popen_forkserver exist on Windows but fail to import
```

AlexWaygood (approved):
LGTM

kkirsche:
Thanks :) two more to go and the module will be complete 👍

PR #4776, opened 2020-11-18
Add type annotations for `peewee`

dargueta:
- [x] AsIs
- [x] AutoField
- [x] BareField
- [x] BigAutoField
- [x] BigBitField
- [x] BigIntegerField
- [x] BinaryUUIDField
- [ ] BitField
- [x] BlobField
- [x] BooleanField
- [x] Case
- [x] Cast
- [x] CharField
- [x] Check
- [ ] chunked
- [x] Column
- [x] CompositeKey
- [ ] Context
- [x] Database
- [x] DatabaseError
- [x] DatabaseProxy
- [x] DataError
- [x] DateField
- [x] DateTimeField
- [x] DecimalField
- [x] DeferredForeignKey
- [x] DeferredThroughModel
- [x] DJANGO_MAP
- [x] DoesNotExist
- [x] DoubleField
- [ ] DQ
- [x] EXCLUDED
- [x] Field
- [x] FixedCharField
- [x] FloatField
- [x] fn
- [x] ForeignKeyField
- [x] IdentityField
- [x] ImproperlyConfigured
- [ ] Index
- [x] IntegerField
- [x] IntegrityError
- [x] InterfaceError
- [x] InternalError
- [x] IPField
- [x] JOIN
- [ ] ManyToManyField
- [x] Model
- [ ] ModelIndex
- [ ] MySQLDatabase
- [x] NotSupportedError
- [x] OP
- [x] OperationalError
- [ ] PostgresqlDatabase
- [ ] PrimaryKeyField (Deprecated)
- [x] ProgrammingError
- [x] Proxy
- [x] QualifiedNames
- [x] SchemaManager
- [x] SmallIntegerField
- [ ] Select
- [x] SQL
- [ ] SqliteDatabase
- [x] Table
- [x] TextField
- [x] TimeField
- [x] TimestampField
- [ ] Tuple
- [x] UUIDField
- [x] Value
- [x] ValuesList
- [x] Window

PR #8426, opened 2022-07-28
refactor: update emoji package to 2.0.0

kkirsche:
Fix: #8359

This updates the stubs for the emoji package to 2.0.0

Parts of the `EMOJI_DICT` could be `TypeDict`'d but I don't think there is support for the attribute may exist or may not exist to appropriately do that typing, so I've marked the inner-dict as Incomplete.


kkirsche:
The pyright failure was:

```
Run jakebailey/pyright-action@v1
Error: Error: read ECONNRESET
```

Fixed the missing at runtime error I had missed though

stubs/emoji/emoji/core.pyi
```diff
@@ -26,22 +22,13 @@ def emojize(
...
 ) -> str: ...
-def replace_emoji(
-    string: str,
-    replace: str | Callable[[str, dict[str, str]], str] = ...,
-    language: str | type[_DeprecatedParameter] = ...,
-    version: float | None = ...,
-) -> str: ...
-def get_emoji_regexp(language: str | None = ...) -> Pattern[str]: ...
-def emoji_lis(string: str, language: str | type[_DeprecatedParameter] = ...) -> list[_EmojiLisReturn]: ...
+def replace_emoji(string: str, replace: str | Callable[[str, dict[str, str]], str] = ..., version: float | None = ...) -> str: ...
```
AlexWaygood, line 30: I think the existing annotation here was incorrect: it can't be `None`, because it's compared with `-1` here: https://github.com/carpedm20/emoji/blob/c398e0c5ff6ceff08c7860078b242765f093f285/emoji/core.py#L228

```suggestion
def replace_emoji(string: str, replace: str | Callable[[str, dict[str, str]], str] = ..., version: float = ...) -> str: ...
```

stubs/emoji/emoji/unicode_codes/__init__.pyi
```diff
@@ -1,26 +1,8 @@
-EMOJI_ALIAS_UNICODE_ENGLISH: dict[str, str]
-EMOJI_UNICODE_ENGLISH: dict[str, str]
-UNICODE_EMOJI_ENGLISH: dict[str, str]
-UNICODE_EMOJI_ALIAS_ENGLISH: dict[str, str]
+from _typeshed import Incomplete
 
-EMOJI_UNICODE_GERMAN: dict[str, str]
-UNICODE_EMOJI_GERMAN: dict[str, str]
+from .data_dict import EMOJI_DATA as EMOJI_DATA, LANGUAGES as LANGUAGES, STATUS as STATUS
```
AlexWaygood, line 3: We should just copy the runtime here, I think

```suggestion
from .data_dict import *
```

stubs/emoji/emoji/unicode_codes/__init__.pyi
```diff
@@ -1,26 +1,8 @@
...
-EMOJI_UNICODE_PORTUGUESE: dict[str, str]
-UNICODE_EMOJI_PORTUGUESE: dict[str, str]
-
-EMOJI_UNICODE: dict[str, dict[str, str]]
-UNICODE_EMOJI: dict[str, dict[str, str]]
-
-EMOJI_DATA: dict[str, dict[str, str]]
-
-STATUS: dict[str, int]
+def get_emoji_unicode_dict(lang: str) -> Incomplete: ...
```
AlexWaygood, line 7:
```suggestion
def get_emoji_unicode_dict(lang: str) -> dict[str, str]: ...
```

AlexWaygood (commented):
Mostly looks good

stubs/emoji/emoji/core.pyi
kkirsche: Ah, good point. The `| None` is only on the downstream functions which this is acting as a wrapper for.

AlexWaygood (approved):
Thanks!

AlexWaygood:
If you fancy tackling #8360, that would also be a very welcome PR :)

kkirsche:
I'll take a look at what that requires. I think that'll be much more straightforward for me than trying to handle #4776, as I don't have enough experience with peewee to feel confident I'll get it right without starting them over from scratch.

PR #8427, opened 2022-07-28
fix: add missing multiprocessing.popen_spawn_posix stubs

kkirsche:

This adds the missing stubs for multiprocessing.popen_spawn_posix from:
https://github.com/python/cpython/blob/main/Lib/multiprocessing/popen_spawn_posix.py

### DIFF
--- a/stdlib/multiprocessing/popen_spawn_posix.pyi
+++ b/stdlib/multiprocessing/popen_spawn_posix.pyi
@@ -15,9 +15,8 @@ if sys.platform != "win32":
         def detach(self) -> int: ...
 
     class Popen(popen_fork.Popen):
-        DupFd: type[_DupFd]
+        DupFd: ClassVar[type[_DupFd]]
         finalizer: Finalize
-        method: ClassVar[str]
         pid: int  # may not exist if _launch raises in second try / except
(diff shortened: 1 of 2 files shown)


stdlib/multiprocessing/popen_spawn_posix.pyi
```diff
@@ -0,0 +1,25 @@
...
+    __all__ = ["Popen"]
+
+    class _DupFd:
+        fd: int
+
+        def __init__(self, fd: int) -> None: ...
+        def detach(self) -> int: ...
+
+    class Popen(popen_fork.Popen):
+        DupFd: type[_DupFd]
```
AlexWaygood, line 18:
```suggestion
        DupFd: ClassVar[type[_DupFd]]
```

stdlib/multiprocessing/popen_spawn_posix.pyi
```diff
@@ -0,0 +1,25 @@
...
+        finalizer: Finalize
+        method: ClassVar[str]
```
AlexWaygood, line 20: This is the same type as in the superclass; no need to override it here

```suggestion
```

AlexWaygood (approved):

PR #8428, opened 2022-07-28
Fix: add missing multiprocessing.popen_spawn_win32

kkirsche:
Fixes: #6799
Requires: https://github.com/python/typeshed/pull/8427 to be complete fix

Source:
https://github.com/python/cpython/blob/main/Lib/multiprocessing/popen_spawn_win32.py

This also follows the _winapi stubs to update the duplicate method based on how it behaves. Additional work can be done on that if desired


stdlib/multiprocessing/popen_spawn_win32.pyi
```diff
@@ -0,0 +1,30 @@
...
+        finalizer: Finalize
+        method: ClassVar[str]
+        pid: int
+        returncode: int | None
+        sentinel: int
+
+        def __init__(self, process_obj: BaseProcess) -> None: ...
+        def duplicate_for_child(self, handle: int) -> int: ...
+        def wait(self, timeout: float | None = ...) -> int | None: ...
+        def poll(self) -> int | None: ...
```
AlexWaygood, line 25: I don't think `wait` (and, by extension, `poll`, which returns whatever `wait` returns) can return `None`: if `self.returncode` is `None` when `wait` is called, there's this whole branch of logic which looks like it's dedicated to guaranteeing `self.returncode` is set to an `int` before `self.returncode` is returned: https://github.com/python/cpython/blob/2fbee85931296bbeddae6358583e400ce5321f89/Lib/multiprocessing/popen_spawn_win32.py#L102-L113

```suggestion
        def wait(self, timeout: float | None = ...) -> int: ...
        def poll(self) -> int: ...
```

AlexWaygood: Oh but on the other hand there's this `if self.wati(timeout=1.0) is None` check here: https://github.com/python/cpython/blob/2fbee85931296bbeddae6358583e400ce5321f89/Lib/multiprocessing/popen_spawn_win32.py#L125

AlexWaygood: Yeah ignore me, this suggestion is incorrect.

stdlib/multiprocessing/reduction.pyi
```diff
@@ -27,9 +27,11 @@ def dump(obj, file, protocol: Any | None = ...) -> None: ...
 
 if sys.platform == "win32":
     if sys.version_info >= (3, 8):
-        def duplicate(handle, target_process: Any | None = ..., inheritable: bool = ..., *, source_process: Any | None = ...): ...
+        def duplicate(
+            handle: int, target_process: Any | None = ..., inheritable: bool = ..., *, source_process: Any | None = ...
```
AlexWaygood, line 31:
```suggestion
            handle: int, target_process: Incomplete | None = ..., inheritable: bool = ..., *, source_process: Incomplete | None = ...
```

stdlib/multiprocessing/reduction.pyi
```diff
@@ -27,9 +27,11 @@ def dump(obj, file, protocol: Any | None = ...) -> None: ...
...
+        ) -> int: ...
     else:
-        def duplicate(handle, target_process: Any | None = ..., inheritable: bool = ...): ...
+        def duplicate(handle: int, target_process: Any | None = ..., inheritable: bool = ...) -> int: ...
```
AlexWaygood, line 34:
```suggestion
        def duplicate(handle: int, target_process: Incomplete | None = ..., inheritable: bool = ...) -> int: ...
```

stdlib/multiprocessing/popen_spawn_win32.pyi
kkirsche: All good :) I had the same line of thought when I was putting it together

AlexWaygood (commented):
The stubs for `multiprocessing.reduction` were written (mostly autogenerated) before `_typeshed.Incomplete` was a thing. We should change the `Any` usages to `Incomplete`, while we're here.

stdlib/multiprocessing/popen_spawn_win32.pyi
AlexWaygood: Glad we could go on this journey together 😄

stdlib/multiprocessing/reduction.pyi
kkirsche: This would instead make more sense to be: `int | None`

* `duplicate` is defined here: https://github.com/python/cpython/blob/main/Lib/multiprocessing/reduction.py#L71
* `target_process` is the third argument to `_winapi.DuplicateHandle` here: https://github.com/python/cpython/blob/main/Lib/multiprocessing/reduction.py#L79
* `_winapi.DuplicateHandle` is here, and only accepts `ints` https://github.com/python/typeshed/blob/master/stdlib/_winapi.pyi#L165

stdlib/multiprocessing/reduction.pyi
kkirsche: See comment on other statement

stdlib/multiprocessing/reduction.pyi
AlexWaygood: Makes sense.

AlexWaygood (approved):

Issue #8366, opened 2022-07-22
[Pillow] Module has no attribute "Resampling"


ow0x:
Hi,

Since Pillow version >9.1.0, the Pillow library has switched the `PIL.Image` filters to `Resampling` enum class and deprecated the old method of using typing.Literal for defining filters. Due to this reason, mypy shows the error stated in this issue title if anyone is using types-Pillow package and is using `PIL.Image.Resampling` approach in our project. The reason for me to switch to Resampling enum method is to avoid DeprecationWarning and safeguard my code ahead of time.
```py
[WARNING] py.warnings: DeprecationWarning: LANCZOS is deprecated and will be removed in Pillow 10 (2023-07-01). Use Resampling.LANCZOS instead.
```

Pillow docs for reference: https://pillow.readthedocs.io/en/stable/reference/Image.html?highlight=lanczos#resampling-filters
Pillow 9.1.0 release notes: https://pillow.readthedocs.io/en/stable/releasenotes/9.1.0.html#constants
the same is documented in https://pillow.readthedocs.io/en/stable/deprecations.html#constants

It would be grateful if this can be accounted for in the pillow typeshed package if possible. I would like to submit a PR to fix but unsure how should I go about this so I decided to open an issue.

hauntsaninja:
Thanks, PR welcome!

You can add the Resampling enum (and other new enums) over here https://github.com/python/typeshed/blob/master/stubs/Pillow/PIL/Image.pyi

PR #8419, opened 2022-07-28
[types-pillow] update deprecated `PIL.Image` constants

ow0x:
Hi,

As per my opened issue #8366 , this PR updates deprecated `PIL.Image` constants (since Pillow >9.1.0) to the documented ones as per Pillow library documentation.
Pillow docs for reference: https://pillow.readthedocs.io/en/stable/deprecations.html#constants
Reference to Pillow v9.1.0 release notes where the deprecation notice was intially documented: https://pillow.readthedocs.io/en/stable/releasenotes/9.1.0.html#constants

I would like to wholeheartedly thank @hauntsaninja for providing the guidance on how and where to make necessary changes and the valuable opportunity to contribute to this open source project.

I have read the contribution guidelines of this project and tried to follow them to the best of my abilities. Please guide me on right path in case I have made the changes inappropriately as this is my first PR on this project and I am a bit unsure if I have done it right! 😅 

This PR resolves #8366

PR #8429, opened 2022-07-28
Add new Resampling enum to Pillow Image

dynobo:
Should solve #8366.

(My first little PR here :see_no_evil: )


dynobo:
Oh,  never mind. #8419 was faster and is more complete...

hauntsaninja:
Just a little too slow, but thank you for contributing! :-)

PR #8439, opened 2022-07-29
refactor: Add missing types related to requests

kkirsche:
This begins the processing of completing the requests package's stubs.

I attempted to use stubtest locally to do the whole thing, but it doesn't seem to load the custom typeshed directory using this command:

```
stubtest --custom-typeshed-dir /Users/kkirsche/git/python/typeshed requests
```

I'm assuming this may have to do with having types-requests installed?

1. Adds missing stubs for `__version__.py`: https://github.com/psf/requests/blob/main/requests/__version__.py
    * Removes the use of `Any` for these values
2. Add missing type alias names as seen here: https://github.com/psf/requests/blob/main/requests/compat.py#L74-L79
3. Adds missing `atomic_open` and `resolve_proxies` functions
    * `atomic_open`: https://github.com/psf/requests/blob/main/requests/utils.py#L295-L305
    * `resolve_proxies`: https://github.com/psf/requests/blob/main/requests/utils.py#L857-L881
 4. Add missing `get` method to `LookupDict`:
    * `get`: https://github.com/psf/requests/blob/main/requests/structures.py#L98-L99
5. Adds missing `help.py` stubs
   * https://github.com/psf/requests/blob/main/requests/help.py


stubs/requests/requests/help.pyi
```diff
@@ -0,0 +1,35 @@
+from typing import TypedDict
```
AlexWaygood, line 1:
```suggestion
from typing_extensions import TypedDict
```

kkirsche: Thanks — was just moving back to this after adding the test case for weakref :)

Akuli:
> stubtest --custom-typeshed-dir /Users/kkirsche/git/python/typeshed requests

As you noticed, `--custom-typeshed-dir` doesn't work for third-party packages. You need to use the `MYPYPATH` environment variable instead:

```
$ MYPYPATH=stubs/requests/ stubtest requests
```

This tells mypy that folders in `stubs/requests/` should be importable.

Maybe we should document this in CONTRIBUTING.md. We generally don't require contributors to run anything locally, but if you want to run stubtest locally, you will likely come across this.

stubs/requests/requests/_internal_utils.pyi
```diff
@@ -0,0 +1,14 @@
...
+
+_VALID_HEADER_NAME_RE_BYTE: Pattern[bytes]
+_VALID_HEADER_NAME_RE_STR: Pattern[str]
+_VALID_HEADER_VALUE_RE_BYTE: Pattern[bytes]
+_VALID_HEADER_VALUE_RE_STR: Pattern[str]
+
+HEADER_VALIDATORS = {
+    bytes: (_VALID_HEADER_NAME_RE_BYTE, _VALID_HEADER_VALUE_RE_BYTE),
+    str: (_VALID_HEADER_NAME_RE_STR, _VALID_HEADER_VALUE_RE_STR),
+}
```
AlexWaygood, line 11:
```suggestion
from typing import Any

HEADER_VALIDATORS: dict[type, tuple[Pattern[Any], Pattern[Any]]]
```

kkirsche: Happy to make this change, but why use `Any` instead of a typed dict with more specific types?

AlexWaygood: I already applied the change to fix the flake8/pytype CI failure. A `TypedDict` won't work here, as a `TypedDict` has to have `str` keys, and this dictionary doesn't.

kkirsche: Ah, because they're `type`. Makes sense, thanks

kkirsche:
I've reduced the strictness temporarily to unblock this from being merged and have noted in the `METADATA.toml` the specifics that are missing for requests to have complete coverage, specifically in the adapters module.

The SOCKSProxyManager uses a function (https://github.com/psf/requests/blob/main/requests/adapters.py#L57-L62) as a fallback on import error. As that's a `NoReturn` due to the raise, I assumed that I didn't need to type for the fallback, though can if necessary.

AlexWaygood:
> The SOCKSProxyManager uses a function (https://github.com/psf/requests/blob/main/requests/adapters.py#L57-L62) as a fallback on import error. As that's a `NoReturn` due to the raise, I assumed that I didn't need to type for the fallback, though can if necessary.

You'll probably have to add `PySocks` to a `requirements-stubtest.txt` file for `requests` as well, otherwise stubtest will only see the fallback.

kkirsche:
I've also added a note that this will be a candidate for use of #8467, if agreed to.

stubs/requests/requests/__init__.pyi
```diff
@@ -27,13 +38,7 @@ from .models import PreparedRequest as PreparedRequest, Request as Request, Resp
...
-__title__: Any
-__build__: Any
-__license__: Any
-__copyright__: Any
-__version__: Any
-
-class NullHandler(logging.Handler):
-    def emit(self, record): ...
+charset_normalizer_version: str | None
+chardet_version: str | None
```
AlexWaygood, line 42: I don't think we need to include these in the stub. They're just implementation details of some initialisation logic that `requests` does when you import it; they're not really meant for public consumption. Does stubtest complain if you omit them from the stub and `ignore_missing_stub` is set to `False`?

stubs/requests/requests/_internal_utils.pyi
```diff
@@ -0,0 +1,7 @@
+from re import Pattern
+from typing import Any
+
+HEADER_VALIDATORS: dict[type, tuple[Pattern[Any], Pattern[Any]]]
+
+def to_native_string(string: str | bytes, encoding: str = ...) -> str: ...
+def unicode_is_ascii(u_string: str) -> bool: ...
```
AlexWaygood, line 7: I'm not really sure we should be including any stubs for a module that's named `_internal_utils`. Does stubtest complain if we omit it and `ignore_missing_stub` is set to `False`?

stubs/requests/requests/adapters.pyi
```diff
@@ -2,9 +2,11 @@ from collections.abc import Mapping
 from typing import Any
 
 from urllib3 import exceptions as urllib3_exceptions, poolmanager, response
+from urllib3.contrib import socks
 from urllib3.util import retry
 
 from . import cookies, exceptions, models, structures, utils
+from .compat import basestring as basestring
 
 Response = models.Response
```
AlexWaygood, line 11: For lines 11-34: All of these are re-exports at runtime. I'm not quite sure why we're using assignments in the stub -- re-exports are generally less buggy, so I think we should use re-exports here as well, like  the runtime does.

stubs/requests/requests/auth.pyi
```diff
@@ -1,6 +1,7 @@
 from typing import Any
 
 from . import cookies, models, utils
+from .compat import basestring as basestring
```
AlexWaygood, line 4: Does this need to be re-exported from this module? It would be a bit weird to do `from requests.auth import basestring`, surely? Shouldn't end users be importing `basestring` from `.compat`?

stubs/requests/requests/compat.pyi
```diff
@@ -16,3 +17,10 @@ from urllib.request import getproxies as getproxies, parse_http_list as parse_ht
 is_py2: bool
 is_py3: bool
 has_simplejson: bool
+
+builtin_str: TypeAlias = str  # noqa: Y042
+# str = str omitted
+# bytes = bytes omitted
```
AlexWaygood, line 23: `builtin_str` doesn't need to be annotated with `TypeAlias`; we can use a simple assignment here, and then we don't need the `# noqa`. Also, why are you omitting the `str` and `bytes` re-exports?

```suggestion
builtin_str = str
str = str
bytes = bytes
```

stubs/requests/requests/compat.pyi
```diff
@@ -16,3 +17,10 @@ from urllib.request import getproxies as getproxies, parse_http_list as parse_ht
...
+basestring: tuple[type, type]
+numeric_types: tuple[type, type]
+integer_types: tuple[type]
```
AlexWaygood, line 26: I feel like the exact lengths of these is an implementation detail that might change in the future

```suggestion
basestring: tuple[type, ...]
numeric_types: tuple[type, ...]
integer_types: tuple[type, ...]
```

stubs/requests/requests/help.pyi
```diff
@@ -0,0 +1,44 @@
+from types import ModuleType
+from typing_extensions import TypedDict
+
+from .__version__ import __version__ as requests_version
+
+chardet: ModuleType | None
+charset_normalizer: ModuleType | None
+cryptography: ModuleType | None
+pyopenssl: ModuleType | None
+OpenSSL: ModuleType | None
```
AlexWaygood, line 10: I don't think we really need any of these; none of them are meant to be imported from this module

```suggestion
```

stubs/requests/requests/help.pyi
```diff
@@ -0,0 +1,44 @@
...
+    using_charset_normalizer: bool
+    pyOpenSSL: _PyOpenSSLDict
+    urllib3: _VersionDict
+    chardet: _OptionalVersionDict
+    charset_normalizer: _OptionalVersionDict
+    cryptography: _VersionDict
+    idna: _VersionDict
+    requests: _VersionDict
+
+def _implementation() -> dict[str, str]: ...
```
AlexWaygood, line 42:
```suggestion
```

stubs/requests/requests/models.pyi
```diff
@@ -7,6 +7,7 @@ from typing import Any
 from urllib3 import exceptions as urllib3_exceptions, fields, filepost, util
 
 from . import auth, cookies, exceptions, hooks, status_codes, utils
+from .compat import basestring as basestring
```
AlexWaygood, line 10: I don't think we need this re-export, do we?
```suggestion
```

stubs/requests/requests/packages.pyi
```diff
@@ -1,3 +1,6 @@
 # requests also imports urllib3, idna, and chardet below
 # requests.packages. The stubs don't reflect that and it's recommended to
 # import these packages directly if needed.
+mod: str
+package: str
+target: str
```
AlexWaygood, line 6: These are all loop variables that leak into the global scope. If stubtest complains about these with `ignore_missing_stub` set to `False`, the proper solution is to add entries to the allowlist with a comment explaining why they're being allowlisted. We definitely shouldn't be adding these to the stub.

stubs/requests/requests/sessions.pyi
```diff
@@ -40,10 +41,23 @@ def merge_setting(request_setting, session_setting, dict_class=...): ...
...
-    def resolve_redirects(self, resp, req, stream=..., timeout=..., verify=..., cert=..., proxies=...): ...
+    def resolve_redirects(
+        self,
+        resp,
+        req,
+        stream: bool = ...,
+        timeout=...,
+        verify=...,
+        cert=...,
+        proxies=...,
```
AlexWaygood, line 52:
```suggestion
        timeout: Incomplete | None = ...,
        verify: Incomplete | None = ...,
        cert: Incomplete | None = ...,
        proxies: Incomplete | None = ...,
```

stubs/requests/requests/structures.pyi
```diff
@@ -17,5 +18,5 @@ class LookupDict(dict[str, _VT]):
     name: Any
     def __init__(self, name: Any = ...) -> None: ...
     def __getitem__(self, key: str) -> _VT | None: ...  # type: ignore[override]
-    def __getattr__(self, attr: str) -> _VT: ...
     def __setattr__(self, __attr: str, __value: _VT) -> None: ...
+    def get(self, key: str, default: _D | None = ...) -> _VT | _D | None: ...  # type: ignore[override]
```
AlexWaygood, line 22: Shouldn't this just have the same signature as `Mapping.get()` in `typing.pyi`? The only difference seems to be that the `key` parameter is positional-or-keyword, instead of being positional-only, as it is on `Mapping.get()`.

stubs/requests/requests/utils.pyi
```diff
@@ -1,10 +1,14 @@
...
 from contextlib import AbstractContextManager
+from io import BufferedWriter
+from os import PathLike
 from typing import Any, AnyStr
 from typing_extensions import TypeAlias
 
 from . import compat, cookies, exceptions, structures
-from .models import PreparedRequest
+from ._internal_utils import HEADER_VALIDATORS as HEADER_VALIDATORS
+from .compat import basestring as basestring, integer_types as integer_types
```
AlexWaygood, line 10: Why do these need to be re-exported here? Should end users be importing them from this module?

```suggestion
```

stubs/requests/requests/utils.pyi
```diff
@@ -22,6 +26,7 @@ def super_len(o): ...
 def get_netrc_auth(url: _Uri, raise_errors: bool = ...) -> tuple[str, str] | None: ...
 def guess_filename(obj): ...
 def extract_zipped_paths(path): ...
+def atomic_open(filename: PathLike[AnyStr]) -> AbstractContextManager[BufferedWriter]: ...
```
AlexWaygood, line 29: - The `AnyStr` `TypeVar` is meaningless here, since it's only used once in the function signature.
- This function works with `str` or `bytes`, not just objects that have an `__fspath__` method
- You can be more specific in the return type -- you can use `contextlib._GeneratorContextManager` instead of `AbstractContextManager`

```suggestion
def atomic_open(filename: StrOrBytesPath) -> _GeneratorContextManager[BufferedWriter]: ...
```

Where `StrOrBytesPath` is imported from `_typeshed`

stubs/requests/requests/utils.pyi
```diff
@@ -47,6 +52,7 @@ def set_environ(env_name: str, value: None) -> AbstractContextManager[None]: ...
 def should_bypass_proxies(url: _Uri, no_proxy: Iterable[str] | None) -> bool: ...
 def get_environ_proxies(url: _Uri, no_proxy: Iterable[str] | None = ...) -> dict[Any, Any]: ...
 def select_proxy(url: _Uri, proxies: Mapping[Any, Any] | None): ...
+def resolve_proxies(request: Request | PreparedRequest, proxies: Mapping[str, str], trust_env: bool = ...): ...
```
AlexWaygood, line 55: Looks like `proxies` can also be `None`: https://github.com/psf/requests/blob/177dd90f18a8f4dc79a7d2049f0a3f4fcc5932a0/requests/utils.py#L868

```suggestion
def resolve_proxies(request: Request | PreparedRequest, proxies: Mapping[str, str] | None, trust_env: bool = ...): ...
```

stubs/requests/requests/_internal_utils.pyi
kkirsche: It was locally, I can't speak to in CI/CD.

I'm torn on this though. If you want to work towards having upstream maintainers take over types, avoiding internal utility methods as have been done so far hurts that as they're critical to how the libraries work, in many cases.

stubs/requests/requests/auth.pyi
kkirsche: This was a stubtest related error locally which is why it was added and re-exported. I don't disagree with you but I don't want to avoid doing what the runtime actually has

stubs/requests/requests/compat.pyi
kkirsche: Because these were explicitly in the upstream

stubs/requests/requests/utils.pyi
kkirsche: Because it's what stubtest said was required to achieve correctness in this module, not what the team would normally recommend doing though.

stubs/requests/requests/structures.pyi
kkirsche: I don't know about mapping, but whatever `dict` uses as it calls `self.__dict__.get(key, default)` and subclasses `dict` — I'll take a look at this shortly

stubs/requests/requests/sessions.pyi
kkirsche: I'd prefer we just type these rather than resorting to incomplete unnecessarily

stubs/requests/requests/compat.pyi
kkirsche: Oops, I misread your comment. 

I was getting errors in my editor that I couldn't do that. So I omitted them because of the error. If it's actually valid though, happy to do that. I think it's because I was trying to use `TypeAlias`

When I hadn't used `TypeAlias` it had been triggering an error locally, but I'd need to pull that up to find it

stubs/requests/requests/sessions.pyi
AlexWaygood: Of course we should aim to have these functions completely typed eventually. But we don't need to do that in _this_ PR; and this is easy to verify, and better than the status quo

AlexWaygood (changes requested):
Stubtest is an extremely useful tool, but it's an imperfect tool. We shouldn't go adding loads of implementation details and unnecessary re-exports to the stub just to satisfy stubtest; adding entries to a stubtest allowlist is often a better option in these situations.

stubs/requests/requests/_internal_utils.pyi
AlexWaygood: If upstream maintainers want to integrate typeshed's stubs into their codebases, that's great news! We're happy to help maintainers do that, and have done in the past. That's _not_ the primary purpose of the stubs packages typeshed provides, however.

kkirsche: But what's the end goal of that? If there is no end game, doesn't that simply mean this is going to infinitely grow with no way to maintain it at scale, assuming appropriate interest. As such, I would argue that it's critical to take maintainers as a core customer not as secondary, because if they don't adopt stubs due to errors and whatnot, you end up in a lose-lose where they leave the work to you and you aren't prioritizing the work for them.

Just feels like this is being made harder on yourselves than you need to be because of these types of decisions, which in my opinion hurt the availability and use of types in upstream libraries

stubs/requests/requests/__init__.pyi
kkirsche: Locally or remotely?

This was added because of local errors when running `MYPYPATH=stubs/requests stubtest requests`

kkirsche:
Before I address these, most seem to boil down to what does stubtest do in this situation.

As such, how do you want me to justify these inclusions? With the error messages, with CI failures, or some other means?

kkirsche:
> Stubtest is an extremely useful tool, but it's an imperfect tool. We shouldn't go adding loads of implementation details and unnecessary re-exports to the stub just to satisfy stubtest; adding entries to a stubtest allowlist is often a better option in these situations.

I personally disagree with this. Accurate types as a user are much more important to me than leaving a whole bunch of stuff out because of differing opinions. They're often critical to expanding on existing libraries via wrappers and whatnot, where implementation details matter.

The removal of things like implementation details also complicates automated testing of stubs by making it impossible to trace the stubs against the implementation from the inner-most types to the outer-most types, validating as you go. 

If you aren't interested , I'd rather just close this and spend my time on other stuff if we don't see eye to eye on the purpose of type hints and what should be included.

stubs/requests/requests/__init__.pyi
AlexWaygood: If stubtest emits errors for these variables, then I think that is a false-positive error that should be solved by adding an allowlist entry when we change the stubtest setting to `ignore_missing_stub = False`

AlexWaygood:
I have stated typeshed's policy on these issues. It would be useful and good to add several of the missing functions you've added in this PR, but as it is, it cannot be merged. Whether you want to address my feedback or close this PR is, of course, up to you.

Akuli:
I'll try to motivate typeshed's philosophy a bit. It is very much based on "practicality beats purity". In general, we add non-public functions and other implementation details if someone genuinely needs them. This works well for several practical reasons:
- When someone starts using a non-public function in type-safe code, they need to either `# type: ignore` it or discuss with us about it. This is a good thing because it often turns out that there's a public function that does the same thing.
- Non-public things are still added as needed in the less common case where they are genuinely needed (as you say, by a wrapper library or something).
- It is less stubs to maintain for us.

So we don't aim for some kind of "purity" of leaving out everything non-public or including everything non-public: neither of those would work well for many users.

In the past, we haven't heard of library maintainers wanting stubs for non-public things. Is this a thing that exists, but library maintainers just don't tell us about it? If a library maintainer wanted stubs for non-public parts of their library, we would likely allow contributors to add them, as we would do with any other justified request to add non-public stubs.

kkirsche:
> I'll try to motivate typeshed's philosophy a bit. It is very much based on "practicality beats purity". In general, we add non-public functions and other implementation details if someone genuinely needs them. This works well for several practical reasons:
> 
> * When someone starts using a non-public function in type-safe code, they need to either `# type: ignore` it or discuss with us about it. This is a good thing because it often turns out that there's a public function that does the same thing.
> * Non-public things are still added as needed in the less common case where they are genuinely needed (as you say, by a wrapper library or something).
> * It is less stubs to maintain for us.
> 
> So we don't aim for some kind of "purity" of leaving out everything non-public or including everything non-public: neither of those would work well for many users.
> 
> In the past, we haven't heard of library maintainers wanting stubs for non-public things. Is this a thing that exists, but library maintainers just don't tell us about it? If a library maintainer wanted stubs for non-public parts of their library, we would likely allow contributors to add them, as we would do with any other justified request to add non-public stubs.

That's fair, and more understandable than simply explaining that it's policy.

As a penetration tester and security application developer, I strongly rely on things like implementation details when automating fuzzing and testing of well known libraries. Not only for testing purposes but also for development of abstractions on top of them.

To verify type hints end-to-end for accuracy up front, I've written internal tooling, which I would need to rewrite to share but am open to doing so, which traces the code's source for the type hints of those libraries and then "lift" those type hints up based on calling parameters. What this does is it validates the actual behavior of the code against the expected type hints.

What this does it allows me to identify where types don't match their downstream callers. As a penetration tester, this gives me what I need to be able to fuzz these libraries for vulnerabilities based on unexpected type errors they may have. An easy example of this is when someone says that an IP address for something like curl must be a dotted quad notation (e.g. 192.168.1.1) when it accepts ints, hex, and many other forms of IP addresses. 

By tracing type hints, specifically the lower level implementation details, you see where those mismatches occur. Once you have a high-confidence list of type mismatches (e.g. upstream is more strict than downstream) you can spot check for instance checks or fully automate the process of using the functions in the same way you could from an OpenAPI/Swagger API specification to fuzz for errors.

As such, from my background and primary use case, I find implementation details to be a critical component of libraries, which is why I so strongly feel that they should always exist, as they enable automation outside of the type-only space that may not usually be shared in public forums like this.

To each their own at the end of the day, I understand the rationale, but disagree on the impact it has on downstream consumers, as someone who uses type hints not only for security application development (e.g. web apps, compliance scanners, vulnerability detection tools, etc.) but also for non-standard use cases such as security analysis purposes where completeness is valuable to understanding how users or maintainers expect the library to be used and detecting where that doesn't align with what's actually supported, allowing for easier detection and reporting of upstream security vulnerabilities. 

At the end of the day though, I'm a random person on the internet at this stage, and haven't done enough to earn any trust in my reasoning, so I respect that at this time it doesn't make sense necessarily to change current policies. Hopefully one day sufficient work will have been completed to justify re-evaluation of this.

Akuli:
I'm starting to think this is a tooling problem.

Library internals can be "typed" very differently from the public API: while public APIs should be typed to accept anything that works, internals of libraries only need to support what the library does with them. One example would be:

```python3
def sign(x):
    """Return 1 for positive numbers, -1 for negative numbers."""
    return x / abs(x)
```

If this is a public function, we would probably add some `TypeVar`/`overload`/`Protocol` tricks to allow passing in complex numbers and all other kinds of objects that support absolute value and division. But as the utility function of a library, it could be typed as `def sign(x: float) -> float`, or even `def sign(x: int) -> float` depending on how the library uses it.

This means that while typing public APIs should be done by hand, it should be possible to auto-generate type hints for the internals of a library just based on how they are used by the public parts. A tool that does this would be helpful for you, as it would automatically create what you tried to create by hand, in a way that is faster and doesn't add more maintenance work for us. A tool like that would also be useful for typeshed, because we could use it to discover bugs in our public API typings, like this:
1. Apply our public API types to the source code (there are already tools that do this).
2. Run the tool to fill in as much of the remaining types as possible.
3. Run a type checker over the tool. The output will be noisy, but we could review it manually or just look at diffs similarly to `mypy_primer`.

> traces the code's source for the type hints of those libraries and then "lift" those type hints up based on calling parameters.

I don't understand what you're saying here. Do you mean "up" as in towards the public API, or towards the internals of a library? Does your tool do what I'm describing, or does it do something else?

AlexWaygood:
> This means that while typing public APIs should be done by hand, it should be possible to auto-generate type hints for the internals of a library just based on how they are used by the public parts.

You might be able to do something like this by running `MonkeyType` on a package's test suite: https://github.com/Instagram/MonkeyType

kkirsche:
Yep. That's what I mean by lifting types.

So in your example, it'd look and see that `abs` supports `_SupportsAbs[_T]` from the standard library. It would "lift" that type to the `sign` function so that sign, behaviorally, supports `_SupportsAbs[_T]` (per https://github.com/python/typeshed/blob/197dac2c0c216230d39461ec415e6daabe5e8ea1/stdlib/_operator.pyi#L51). This then works "outward" in passes typing more of the library as things are solved.

It ends up with a patched version of the stdlib stubs, removing all the conditionals related to system platforms so that it can do cross-platform evaluation more easily

Akuli:
Your tool and what I had in mind arrive at the same result, auto-generated types for functions, but they do it differently:
- My imaginary tool looks at how typed functions are calling untyped functions, and based on that, it adds type annotations for untyped functions. It works "inward", assuming that every internal function is eventually called by the public API (and presumably just leaves things untyped if that isn't the case).
- Your tool looks at what each function does and adds parameter types accordingly. It works "outwards", starting with the innermost internals of a library that only use known standard-library functions, and eventually adds types to the public functions.

It seems that your approach generates something more like what typeshed does for public functions (accepting as much as possible), whereas mine would create results closer to what I recommend for internal functions (narrow and specific, support only what the library needs).

What does your tool do when a function argument is used in many places? Does it generate this?

```
def sign(x: SupportsAbs[_T]) -> _T:
    return x / abs(x)
```

Running a type checker over this would emit errors, because just `x: SupportsAbs[_T]` doesn't mean that you can divide with it. In fact, my `sign()` function is very difficult to properly express with the current type system, unless you only support a few argument types that the library actually passes to it.

And yes, I am aware of MonkeyType, but I think it could still be useful to do a similar thing based on typeshed's stubs, not the tool's test suite.

kkirsche:
> What does your tool do when a function argument is used in many places? Does it generate this?
> 
> ```
> def sign(x: SupportsAbs[_T]) -> _T:
>     return x / abs(x)
> ```

No, I left out the specifics for how this works:

So in that example, the evaluation would need to process from the inside out, so:

1. `abs` uses `x` and requires `SupportsAbs[_T]`
2. `x` is used with `operator.truediv` (https://github.com/python/typeshed/blob/197dac2c0c216230d39461ec415e6daabe5e8ea1/stdlib/_operator.pyi#L68) which is typed broadly as `Any` / `Any`. 
3. Neither function is sufficiently typed to actually solve at this stage. So we need to trace where the `/` operator is coming from. Technically, it's the `truediv` operator, https://github.com/python/cpython/blob/main/Modules/_operator.c#L128 which corresponds to https://docs.python.org/3/c-api/number.html#c.PyNumber_TrueDivide
5. Because truedivide is a protocol to `__truediv__`, it chooses to use a protocol for this, since technically it can take any type var that supports both truedivide and absolute value. Also, since protocols aren't type checked, it doesn't currently worry about correctly typing `__x` in the `__truediv__`

The resulting type reads as:

```
class SupportsAbsAndTrueDiv(SupportsAbs[_T]):
   def __truediv__(self, __x: object) -> float: ...

def sign(x: SupportsAbsAndTrueDiv[object]) -> float:
    return x / abs(x)
```

It's not perfect as the type system isn't expressive enough to easily capture it in a more simple way. It avoids writing the actual protocols into the files until after evaluations to see which ones should be separated from the usage (used in multiple files)  or if it should be kept near the implementation

PR #8442, opened 2022-07-30
Migrate file args in traceback functions from IO to SupportsWrite

### LINKED ISSUE #8423: Should traceback print method file args accept SupportsWrite?
In the stubs for `traceback`, the `file` arg of functions like `print_exc`, `print_last`, `print_stack`, and `print_exception` all seem wider than strictly necessary, as (AFAICT) they all end up calling `print`, whose `file` arg is typed as `file: SupportsWrite[str] | None`.

EX:
https://github.com/python/typeshed/blob/master/stdlib/traceback.pyi#L76-L78

In py3.10:
https://github.com/python/cpython/blob/3.10/Lib/traceback.py#L104-L121

Should these file parameters also be typed as `SupportsWrite[str] | None`?  I can, for example, do the following:

```py
>>> import traceback
>>> class MyFile:
...   def write(self, s):
...     print(s)
...
>>> try:
...   raise ValueError()
... except Exception:
...   traceback.print_exc(file=MyFile())
...


  File "<stdin>", line 2, in <module>


ValueError
```

Or is there a reason why constraining the `file` parameter it to the wider `IO[str]` type is preferred?  I admit I'm not familiar with the ins and outs of typing builtin libraries.

alanisaac:
Migrates the `file` args in `traceback.py` to use `SupportsWrite[str]` in place of `IO[str]`.

fixes #8423


AlexWaygood (approved):
Nice, thank you!

PR #8446, opened 2022-07-30
`ctypes`: `pointer` is a function, not a class

### LINKED ISSUE #7747: mypy error: Name "ctypes._Pointer" is not defined
`class TessBaseAPI(ctypes._Pointer): pass`
[Source]( https://stackoverflow.com/a/21754478)

### LINKED ISSUE #8351: Typing of ctypes' pointer types is wrong
https://docs.python.org/3/library/ctypes.html#pointers explicitly says:

> Calling the pointer type without an argument creates a NULL pointer.

and even has a code example for this:

```python
null_ptr = POINTER(c_int)()
```
But the `pointer` constructor https://github.com/python/typeshed/blob/10bc4b058462c4343a2cca9adfa7b25979c45562/stdlib/ctypes/__init__.pyi#L167 requires  an argument. This should be optional if I understand the code there correctly.

AlexWaygood:
Fixes #8351. Fixes #7747. Closes #7753.


AlexWaygood:
As expected, the "moral" thing to do here is quite disruptive. @srittau, what do you think? Worth it anyway?

AlexWaygood:
> As expected, the "moral" thing to do here is quite disruptive.

Having said that, it looks like the `psycopg` stubs are autogenerated: https://github.com/psycopg/psycopg/blob/cace92a9854130da02e55f416b2bcbefdc7242e1/psycopg/psycopg/pq/_pq_ctypes.py#L727.

So the fix for that library would be pretty trivial, I think.

srittau (approved):
Let's do this. It might hurt now, but will probably prevent more hurt in the future.

AlexWaygood:
@dvarrazzo, just to give you a heads-up about this — the typeshed stubs for `ctypes` have been a bit of a hack for a while. This PR brings them closer to the runtime reality, but it means you might have to change your stubs for `psycopg` (after a future release of mypy) to use `ctypes._Pointer` instead of `ctypes.pointer`.

dvarrazzo:
@AlexWaygood no problem at all! When this change will be released I'll be happy to update our stubs and set a dependency on the newer mypy version :slightly_smiling_face:

nathanpage-credo:
Uhh this breaks things where a user may want to define pointer annotations:
```python
import ctypes as ct

class Foo(ct.Structure):
    ... # other stuff defined here 
    test: "ct.pointer[ct.c_int32]"  # these now error
    test2: "ct.pointer[ct.c_bool]"  # these now error
```

JelleZijlstra:
> Could `ctypes._Pointer` be a non-internal type (e.g., `ctypes.Pointer`)?

Maybe, but that's not a decision we can make in typeshed. If you want it to be, you should open an issue over at CPython.

nathanrpage97:
Ok, I don’t think it is worthy of a change to cpython implementation. The big thing is linters not liking a reference to a private api. I think for now I will re-export ctypes._Pointer -> PointerType to reduce linter silencing needed.

Akuli:
Cpython might accept making it a public type, a bit similarly to how `_curses.window` is now exposed as `curses.window`.

AlexWaygood:
> Cpython might accept making it a public type, a bit similarly to how `_curses.window` is now exposed as `curses.window`.

Especially given that `ctypes._Pointer` is actually documented -- it seems like it's not really _meant_ to be a "private" class.

PR #8448, opened 2022-07-30
chore: add types to urrlib3.util.url

sk-:
Added types to urrlib3.util.url.

The untyped functions are currently deprecated, hence I did not want to spend any efforts typing them.
Code was manually checked and inspired by the typings present in https://github.com/urllib3/urllib3/blob/main/src/urllib3/util/url.py


AlexWaygood (approved):
Thanks!

PR #8449, opened 2022-07-31
Remove `__unicode__` method from `sqlalchemy.exc`

sobolevn:
It does not have this method for a long time now: https://github.com/sqlalchemy/sqlalchemy/blob/14bfbadfdf9260a1c40f63b31641b27fe9de12a0/lib/sqlalchemy/exc.py#L43-L117


AlexWaygood:
> It does not have this method for a long time now: https://github.com/sqlalchemy/sqlalchemy/blob/14bfbadfdf9260a1c40f63b31641b27fe9de12a0/lib/sqlalchemy/exc.py#L43-L117

It's still there on the latest release, though, and we generally don't remove methods from typeshed until there's been a release with them removed 🙂 https://github.com/sqlalchemy/sqlalchemy/blob/f35a025bd1a9b649e5dedc86c8b5f3b12ffb8c9b/lib/sqlalchemy/exc.py#L103

srittau:
SQLAlchemy is a bit tricky, since `main` is not the current release branch, but work on SQLAlchemy 2.0. The `rel_1_4` branch is what is currently released: https://github.com/sqlalchemy/sqlalchemy/tree/rel_1_4

AlexWaygood:
We'd rather not make this change until there's a SQLAlchemy release which has the change in it. And when there is a SQLAlchemy release with the change in it, we'll have to make the change as part of the version bump, rather than as a standalone PR. So, I'm going to close this for now :)

PR #5346, opened 2021-05-05
Add aliases for 'function classes' to multiprocessing

srittau:
Part of #4313

PR #4314, opened 2020-07-06
multiprocessing: revert changes to queue classes

hauntsaninja:
Partial revert of #4289. Fixes #4313

Issue #4313, opened 2020-07-06
Recent multiprocessing.Queue change breaks existing annotations and reduces typing precision


JukkaL:
#4289 changed `multiprocessing.Queue` from a class to a function. This matches the implementation, but this has some possibly unanticipated side effects, as discussed in https://github.com/python/mypy/pull/9100. Here's a summary of the points raised in the linked PR:

* `multiprocessing.Queue` is documented as a class: https://docs.python.org/3/library/multiprocessing.html#multiprocessing.Queue

* `multiprocessing.queues.Queue` can be used in annotations, but it seems to be undocumented. For a user to annotate code using queues, they need to either read the stubs or the stdlib implementation, which is unfortunate.

* The return type of `multiprocessing.Queue` is `Queue[Any]`, so the change loses some typing precision, even if a user can figure out the correct type. Since `multiprocessing.queues.Queue` is not very discoverable (as it's undocumented), for many new users this may effectively remove the ability to type check queue items.

* The change breaks existing code that uses `multiprocessing.Queue` as a type, without an obvious fix.

The above considerations probably explain why `multiprocessing.Queue` used to be defined as a class, even thought it's not quite true.

I'm not sure what's the best way forward. Here are some ideas:

* Revert the change to `Queue` and other similar types (at least for now).
* Teach type checkers to suggest using `multiprocessing.queues.Queue` (etc.) in type annotations. However, if these are undocumented, maybe we shouldn't recommend these?
* Change the return type of `multiprocessing.Queue` to `Queue[T]` to preserve type checking precision (this may not work on all type checkers).

JelleZijlstra:
Making `multiprocessing.Queue` an alias for `multiprocessing.queues.Queue` feels like it might be a good pragmatic solution. (This would be a partial revert of #4289.) This approach could cause a couple of false negatives:

- Type checkers wouldn't catch code that uses the `ctx=` argument to the `multiprocessing.queues.Queue` constructor.
- Type checkers wouldn't catch code that does `isinstance(..., multiprocessing.Queue)` or otherwise treats the function as a type object at runtime.

But these problems feel less severe than the problems that @JukkaL points out with the current stubs.

cc @vishalkuo and @hauntsaninja who were involved in the previous PR.

srittau:
[...]

PR #4289, opened 2020-06-28
[multiprocessing] fixes more incorrect types around context

vishalkuo:
Another iteration on https://github.com/python/typeshed/issues/4266 which should knock out a couple issues (see `diff` output). Comments on the fixes:

```
multiprocessing.JoinableQueue is not a type
multiprocessing.Queue is not a type
multiprocessing.SimpleQueue is not a type
```

These are all actually functions in the `context` class that return queues from the `queues` module [here](https://github.com/python/cpython/blob/2e0a920e9eb540654c0bb2298143b00637dc5961/Lib/multiprocessing/context.py#L100-L113). 

```
multiprocessing.Event is inconsistent, runtime does not have argument "lock"
multiprocessing.set_executable is inconsistent, stub argument "exe" differs from runtime argument "executable"
```

These are also functions in context that differ from runtime: see [event](https://github.com/python/cpython/blob/2e0a920e9eb540654c0bb2298143b00637dc5961/Lib/multiprocessing/context.py#L90-L93) and [set_executable](https://github.com/python/cpython/blob/2e0a920e9eb540654c0bb2298143b00637dc5961/Lib/multiprocessing/context.py#L172-L178)

# Test Plan:
```
(mypy) (⎈ |development:home)[~/projects/typeshed] (mp) $ git co HEAD~1
(mypy) (⎈ |development:home)[~/projects/typeshed] ((HEAD detached at 5d553c95)) $ python3.8 -m mypy.stubtest --custom-typeshed-dir . multiprocessing --concise >  ~/Desktop/pre-change.txt
(mypy) (⎈ |development:home)[~/projects/typeshed] ((HEAD detached at 5d553c95)) $ git co mp
Previous HEAD position was 5d553c95 apply black and isort (#4287)
Switched to branch 'mp'
(mypy) (⎈ |development:home)[~/projects/typeshed] (mp) $ python3.8 -m mypy.stubtest --custom-typeshed-dir . multiprocessing --concise >  ~/Desktop/post-change.txt
(mypy) (⎈ |development:home)[~/projects/typeshed] (mp) $ diff ~/Desktop/pre-change.txt ~/Desktop/post-change.txt 
1,3d0
< multiprocessing.Event is inconsistent, runtime does not have argument "lock"
< multiprocessing.JoinableQueue is not a type
< multiprocessing.Queue is not a type
6d2
< multiprocessing.SimpleQueue is not a type
8,9d3
[...]

PR #8450, opened 2022-07-31
Remove, move or `# noqa` more `TypeAlias` declarations

AlexWaygood:
Unblocks https://github.com/PyCQA/flake8-pyi/pull/255


stdlib/multiprocessing/__init__.pyi
```diff
@@ -62,27 +60,6 @@ __all__ = [
 if sys.version_info >= (3, 8):
     __all__ += ["parent_process"]
 
-# The following type aliases can be used to annotate the return values of
-# the corresponding functions. They are not defined at runtime.
```
Akuli, line 66: These were a bit controversial when @srittau added them, and he probably doesn't want them removed: #5346

AlexWaygood: Ah, thanks for the reference.

AlexWaygood: The other options are:

- Special-case `__init__.pyi` files in the flake8-pyi check (what I did originally, but @JelleZijlstra doesn't like that).
- Just `# noqa` all the ones in `multiprocessing`.
- Abandon the flake8-pyi check entirely.
- Something else?

Akuli: I think this is somewhat of a "true positive". These type aliases are just weird: named as if they were private, intended to be public, not documented, not discoverable. But again, @srittau will probably disagree. I think we should wait for his input.

AlexWaygood: Yeah of course, not going to merge without it :)

stdlib/multiprocessing/__init__.pyi
```diff
@@ -10,12 +10,10 @@ from multiprocessing.context import (
 from multiprocessing.process import active_children as active_children, current_process as current_process
 
 # These are technically functions that return instances of these Queue classes.
-# Using them as annotations is deprecated. Either use imports from
-# multiprocessing.queues or the aliases defined below. See #4266 for discussion.
+# Using them as annotations is deprecated. Use imports from multiprocessing.queues instead.
+# See #4266 for discussion.
```
srittau, line 14: I think this comment is just outdated? As far as I can see, all these are classes in the implementation, at least in Python 3.7+. Was this maybe different in the past? In that case, I don't think we need the type aliases below.

Akuli: They became functions at some point, but it was reverted.

Akuli: Somehow I missed that you asked about the implementation, not about the stub. These are functions (actually methods) in the implementation:

```
>>> from multiprocessing import Queue
>>> Queue
<bound method BaseContext.Queue of <multiprocessing.context.DefaultContext object at 0x7f66f26d73d0>>
```

There is also an undocumented class that is different from the public and documented `Queue`:

```
>>> from multiprocessing.queues import Queue
>>> Queue
<class 'multiprocessing.queues.Queue'>
```

There's similarly e.g. a method `multiprocessing.Event` and a class `multiprocessing.synchronize.Event`.

AlexWaygood: This is what happens at runtime:

```pycon
>>> import multiprocessing
>>> multiprocessing.Queue
<bound method BaseContext.Queue of <multiprocessing.context.DefaultContext object at 0x000002F32AF7E5C0>>
>>> multiprocessing.Queue[int]
TypeError: 'method' object is not subscriptable
```

So the stubs are a bit of a lie at the moment. As @Akuli says, they were made into functions in #4289, but then that was reverted in #4314 after Jukka complained in #4313.

The thing is, either way, I don't think having these aliases helps the situation much. To work around these issues, a user can probably just put `multiprocessing.Queue` in quotes in their annotations, or use `from __future__ import annotations`.

srittau: We should try to make them into functions again in the stubs and see the fallout, now that we have mypy_primer. Annotating something with `muliprocessing.Queue`, which is a function, is not something we should support.

AlexWaygood: > We should try to make them into functions again in the stubs and see the fallout, now that we have mypy_primer. Annotating something with `muliprocessing.Queue`, which is a function, is not something we should support.

Sure. But not in this PR :)

srittau: Should be a separate PR, but since the type aliases are connected to this, we should keep those that are alternatives to a function. We don't need to keep type aliases that are only an alternative for a class anyway.

Akuli: And going back to the controversy we had before: A typical user likely doesn't care that `multiprocessing.Queue` isn't a function. It's a technicality that gets in the way mostly when doing type annotations and when doing `isinstance(foo, Queue)`. My argument was that because `isinstance` checking is quite rare, including this technicality in the stubs is not a "practicality beats purity" solution. But I agree that we should see what mypy_primer says :)

For this PR, should we delete the aliases? We already know that no projects in mypy_primer are using them. I don't expect anyone else to be using them either, because annotating as if they were classes more or less works now.

Akuli: I marked my previous comment as off-topic, as it doesn't belong at all to this PR. I somehow didn't see other people's comments that were posted before I wrote mine.

AlexWaygood: > since the type aliases are connected to this, we should keep those that are alternatives to a function.

I don't really see how the aliases help, even the ones which are alternatives to a function.

If it's a choice between doing this:

```python
from multiprocessing.synchronize import Lock
foo: Lock
```

or this:

```python
from typing import TYPE_CHECKING
if TYPE_CHECKING:
    from multiprocessing import _LockType
foo: "_LockType"
```

Then, I find the first way much cleaner and more ergonomic. And, as @Akuli says in his minimized comment, we know none of the projects in mypy_primer are using these aliases.

srittau: Sorry for being unclear. Yes, if there is a valid type somewhere, like `multiprocessing.synchronize.Lock`, we don't need the alias. It's possible that this is true for all aliases, I didn't check.

AlexWaygood: `_BarrierType`, `_BoundedSemaphoreType`, `_ConditionType`, `_EventType`, `LockType`, `_RLockType` and `_SemaphoreType` are all aliases to valid types in `multiprocessing.synchronize`.

`_QueueType`, `_SimpleQueueType` and `_JoinableQueueType` are all aliases to valid types in `multiprocessing.queues`.

So yeah, it's true for all the aliases :)

AlexWaygood:
So, are we happy for this to go in? :)

stdlib/multiprocessing/__init__.pyi
AlexWaygood, line 14:
```suggestion
# The stub here doesn't reflect reality exactly --
# while e.g. `multiprocessing.queues.Queue` is a class,
# `multiprocessing.Queue` is actually a function at runtime.
# Avoid using `multiprocessing.Queue` as a type annotation; 
# use imports from multiprocessing.queues instead.
# See #4266 and #8450 for discussion.
```

srittau (approved):

PR #8451, opened 2022-07-31
`SimplePath` always use `__truediv__`, not `__div__`

sobolevn:
This was a bug in `SimplePath` protocol definition, not a new thing: https://github.com/python/importlib_metadata/issues/334

It was always assumed to use `__truediv__`, because `__div__` is not a thing in python3.


srittau:
Unfortunately, in Python 3.10, `SimplePath` uses `__div__`. That said, since this is obviously wrong and can lead to problematic runtime behavior, if `__truediv__` is not used in the implementation (and having `__div__` is unnecessary), I think adding this to the stubtest ignore file is correct.

AlexWaygood:
> Unfortunately, in Python 3.10, `SimplePath` uses `__div__`.

The fix should probably be backported to 3.10 in the stdlib; it's a bit weird that it hasn't been.

tests/stubtest_allowlists/py3_common.txt
```diff
@@ -102,6 +102,7 @@ importlib.abc.Loader.exec_module  # See Lib/importlib/_abc.py. Might be defined
 importlib.abc.MetaPathFinder.find_spec  # Not defined on the actual class, but expected to exist.
 importlib.abc.PathEntryFinder.find_spec  # Not defined on the actual class, but expected to exist.
 importlib.machinery.ExtensionFileLoader.get_filename  # Wrapped with _check_name decorator which changes runtime signature
+importlib.metadata._meta.SimplePath  # There was a bug in `SimplePath` definition in cpython, see #8451
```
AlexWaygood, line 105: It needs to go in the py310.txt allowlist, not py3_common.txt, since it'll only be "used" when stubtest is running on Python 3.10

tests/stubtest_allowlists/py310.txt
```diff
@@ -17,6 +17,7 @@ fractions.Fraction.__new__  # overload is too complicated for stubtest to resolv
 gettext.install
 gettext.translation
 hmac.new  # Stub is a white lie; see comments in the stub
+importlib.metadata._meta.SimplePath  # There was a bug in `SimplePath` definition in cpython, see #8451
```
AlexWaygood:
```suggestion
importlib.metadata._meta.SimplePath.__div__  # There was a bug in `SimplePath` definition in cpython, see #8451
importlib.metadata._meta.SimplePath.__truediv__  # There was a bug in `SimplePath` definition in cpython, see #8451
```

stdlib/importlib/metadata/_meta.pyi
```diff
@@ -17,7 +16,6 @@ class SimplePath(Protocol):
     def joinpath(self) -> SimplePath: ...
     def parent(self) -> SimplePath: ...
     def read_text(self) -> str: ...
-    if sys.version_info >= (3, 11):
-        def __truediv__(self) -> SimplePath: ...
-    else:
-        def __div__(self) -> SimplePath: ...
+    # There was a bug in `SimplePath` definition in cpython, see #8451
+    # Previously `__div__` was used, but it should have been `__truediv__`.
```
AlexWaygood, line 20:
```suggestion
    #  Strictly speaking `__div__` was defined in 3.10, not __truediv__,
    # but it should have always been `__truediv__`.
```

Akuli: Shouldn't `in 3.10` be `before 3.10`?

AlexWaygood: No — the class is new in 3.10, and has the `__div__` method in 3.10. In 3.11, the method has been changed to `__truediv__`, but it hasn't been backported to 3.10.

AlexWaygood (approved):

PR #8454, opened 2022-08-01
redis.backoff: fix  `cap` and `base` type (float)

yoziru:
The backoff classes in `redis` call `random.uniform` under the hood, which accepts both `int` and `float` input

Getting errors in the latest `mypy` 0.971 using `types-redis` 4.3.12:
```
error: Argument "base" to "FullJitterBackoff" has incompatible type "float"; expected "int"
```

Dependency versions:
```
redis == "4.3.4"
mypy == "0.971"
types-redis = "4.3.12"
```


srittau (approved):
Thanks!

PR #8457, opened 2022-08-01
Add `urllib3.contrib.socks`; improve `urllib3.connectionpool`

kkirsche:
This is required to address some re-exports in https://github.com/python/typeshed/pull/8439

The source code is available here:
https://github.com/urllib3/urllib3/blob/main/src/urllib3/contrib/socks.py


stubs/urllib3/urllib3/contrib/socks.pyi
```diff
@@ -0,0 +1,40 @@
+from typing_extensions import TypedDict
+from typing import Any, ClassVar, Mapping
+
+from ..connection import HTTPConnection, HTTPSConnection
+from ..connectionpool import HTTPConnectionPool, HTTPSConnectionPool
+from ..poolmanager import PoolManager
+
+class _SockOptions(TypedDict):
```
srittau, line 9: Why not `_TYPE_SOCKS_OPTIONS` like in the implementation?

stubs/urllib3/urllib3/contrib/socks.pyi
```diff
@@ -0,0 +1,41 @@
...
+    socks_version: int
+    proxy_host: str | None
+    proxy_port: str | None
+    username: str | None
+    password: str | None
+    rdns: bool
+
+class SOCKSConnection(HTTPConnection):
+    def __init__(self, _socks_options: _SockOptions, *args: Any, **kwargs: Any) -> None: ...
```
srittau, line 18: As we might reasonably be able to type `args` and `kwargs`:

```suggestion
    def __init__(self, _socks_options: _SockOptions, *args, **kwargs) -> None: ...
```

stubs/urllib3/urllib3/contrib/socks.pyi
```diff
@@ -0,0 +1,41 @@
...
+    proxy_url: str
+
+    def __init__(
+        self,
+        proxy_url: str,
+        username: str | None = ...,
+        password: str | None = ...,
+        num_pools: int = ...,
+        headers: Mapping[str, str] | None = ...,
+        **connection_pool_kw: Any,
```
srittau, line 40: Same here:

```suggestion
        **connection_pool_kw,
```

srittau (dismissed):

AlexWaygood:
As well as @srittau's points, note that this module only imports successfully if the optional dependency `PySocks` is installed (according to https://github.com/urllib3/urllib3/blob/16d2976041c5b8b0fe5f790e8edc9c50f318ee7b/src/urllib3/contrib/socks.py#L50-L52). You'll need the optional dependency to a `requirements-stubtest.txt` file in order for stubtest to be able to import it in CI (see https://github.com/python/typeshed/blob/master/stubs/tqdm/%40tests/requirements-stubtest.txt for an example).

stubs/urllib3/urllib3/contrib/socks.pyi
kkirsche: Updated. How would I type hint these given that they are `args` / `kwargs` of the `HTTPConnection` as seen in this?

https://github.com/urllib3/urllib3/blob/1.26.x/src/urllib3/connection.py#L115

With type hints in 2.x branch (missing the strict parameter):
https://github.com/urllib3/urllib3/blob/main/src/urllib3/connection.py#L136-L148

stubs/urllib3/@tests/requirements-stubtest.txt
```diff
@@ -0,0 +1 @@
+PySocks>=1.5.6,<2.0,!=1.5.7
```
AlexWaygood, line 1: Does this really need to be so specific? (Honest question -- I haven't checked, and I don't know much about `PySocks`!)

kkirsche: https://github.com/urllib3/urllib3/blob/1.26.x/setup.py#L99

AlexWaygood: Thanks for the reference! It'll be annoying if we have to continually update this to stop our CI from failing, but I guess if that's what the runtime requires, we should probably have the same requirement here.

kkirsche: Does stubtest support the concept of `extras`? The easy way to solve this would be to load a list of extras to install with the package when doing tests

AlexWaygood: > Does stubtest support the concept of `extras`? The easy way to solve this would be to load a list of extras to install with the package when doing tests

Yeah, that's basically the purpose of the `stubtest-requirements.txt` files. All required dependencies are automatically installed by stubtest prior to it stubtesting; this file is for any additional dependencies that we want to specify (which are generally the optional dependencies a package has, if it has any)

AlexWaygood: ...I think I misunderstood what you meant — sorry. A better answer for you is that, no, the script we have in CI to install necessary stubtest requirements and then run stubtest doesn't currently know anything about `extras`. But we could probably improve that script — it's here: https://github.com/python/typeshed/blob/master/tests/stubtest_third_party.py

stubs/urllib3/urllib3/contrib/socks.pyi
```diff
@@ -0,0 +1,41 @@
...
+    socks_version: int
+    proxy_host: str | None
+    proxy_port: str | None
+    username: str | None
+    password: str | None
+    rdns: bool
+
+class SOCKSConnection(HTTPConnection):
+    def __init__(self, _socks_options: _TYPE_SOCKS_OPTIONS, *args, **kwargs) -> None: ...
+    def _new_conn(self) -> Any: ...  # "socks.socksocket":
```
AlexWaygood, line 19:
```suggestion
```

stubs/urllib3/urllib3/contrib/socks.pyi
```diff
@@ -0,0 +1,41 @@
...
+
+class SOCKSHTTPSConnection(SOCKSConnection, HTTPSConnection): ...
+class SOCKSHTTPConnectionPool(HTTPConnectionPool): ...
+class SOCKSHTTPSConnectionPool(HTTPSConnectionPool): ...
```
AlexWaygood, line 23:
```suggestion
class SOCKSHTTPConnectionPool(HTTPConnectionPool):
    ConnectionCls: ClassVar[type[SocksConnection]]
class SOCKSHTTPSConnectionPool(HTTPSConnectionPool):
    ConnectionCls: ClassVar[type[SocksConnection]]
```

You may have to also change this line to be a `ClassVar`: https://github.com/python/typeshed/blob/df4618ef2dec1d27e363da0fc4ba8b00a2c811fe/stubs/urllib3/urllib3/connectionpool.pyi#L47

stubs/urllib3/urllib3/contrib/socks.pyi
```diff
@@ -0,0 +1,41 @@
...
+
+class _ConnectionPoolClasses(TypedDict):
+    http: type[SOCKSHTTPSConnectionPool]
```
AlexWaygood, line 26:
```suggestion
    http: type[SOCKSHTTPConnectionPool]
```

stubs/urllib3/urllib3/contrib/socks.pyi
```diff
@@ -0,0 +1,41 @@
...
+    https: type[SOCKSHTTPSConnectionPool]
+
+class SOCKSProxyManager(PoolManager):
+    pool_classes_by_scheme: ClassVar[_ConnectionPoolClasses]
```
AlexWaygood, line 30: This is overridden on instances, so it's wrong to annotate it as a `ClassVar`, despite the fact it has a class-level default: https://github.com/urllib3/urllib3/blob/aa3def7d242525e6e854991247c4b68583d15135/src/urllib3/contrib/socks.py#L216

```suggestion
   # has a class-level default, but is overridden on instances, so not a ClassVar
    pool_classes_by_scheme: _ConnectionPoolClasses
```

AlexWaygood (changes requested):

stubs/urllib3/@tests/requirements-stubtest.txt
AlexWaygood: I like that idea, quite a lot :) though I can't speak for the other maintainers.

(A change like that should obviously be made in a separate PR, though!)

stubs/urllib3/urllib3/contrib/socks.pyi
kkirsche: https://github.com/urllib3/urllib3/blob/1.26.x/src/urllib3/connectionpool.py reviewed with a variety of updates to improve coverage of this file.

If you get stuck on any of the types, comparing the implementation of the 1.26.x branch (above) and the main 2.x branch seen [here](https://github.com/urllib3/urllib3/blob/main/src/urllib3/connectionpool.py) is a great way to verify types for some of the more trickily named parameters or attributes.

stubs/urllib3/urllib3/connectionpool.pyi
```diff
@@ -1,9 +1,14 @@
+from _typeshed import Self
+from collections.abc import Mapping
+from logging import Logger
+from queue import LifoQueue
 from typing import Any
+from typing_extensions import ClassVar, TypeAlias
```
AlexWaygood, line 6:
```suggestion
from typing import Any, ClassVar
from typing_extensions import TypeAlias
```

stubs/urllib3/@tests/requirements-stubtest.txt
kkirsche: https://github.com/python/typeshed/pull/8467 submitted

AlexWaygood: Thank you!

stubs/urllib3/urllib3/contrib/socks.pyi
srittau: We would basically copy the signature of `HTTPConnection.__init__()`. Cumbersome, but possible.

stubs/urllib3/urllib3/contrib/socks.pyi
AlexWaygood: Thanks! That's more than I was expecting you to do in this PR, but that's fine 🙂

Incredibly, the new stubtest failure appears as though it's due to our stdlib stubs not having `abc.ABCMeta` as the metaclass for `io.IOBase` in the stdlib. I've filed #8468 to fix it.

stubs/urllib3/urllib3/contrib/socks.pyi
```diff
@@ -0,0 +1,44 @@
+from collections.abc import Mapping
+from typing_extensions import ClassVar, TypedDict
```
AlexWaygood, line 2:
```suggestion
from typing import ClassVar
from typing_extensions import TypedDict
```

AlexWaygood:
The continued stubtest failure looks like a stubtest bug to me; I've filed https://github.com/python/mypy/issues/13316. We'll get a new mypy release before the end of August; hopefully either I or @hauntsaninja can fix it by then. Until then I recommend that you add an allowlist entry, with a comment linking to the stubtest bug report :)

kkirsche:
> python/mypy#13316

Added via d8d8d0b. Hopefully I'll be able to contribute at some point to `mypy`, currently I'm not comfortable enough with the [visitor design pattern](https://refactoring.guru/design-patterns/visitor) to feel confident beyond small refactoring.

kkirsche:
> Added via [d8d8d0b](https://github.com/python/typeshed/commit/d8d8d0b9c9bb7eec0e4683e59a53c7f16c157313). Hopefully I'll be able to contribute at some point to `mypy`, currently I'm not comfortable enough with the [visitor design pattern](https://refactoring.guru/design-patterns/visitor) to feel confident beyond small refactoring.

OK, I thought I had added it, but apparently I did it wrong? Would someone be able to help me understand what I got wrong with that commit?

AlexWaygood:
> OK, I thought I had added it, but apparently I did it wrong? Would someone be able to help me understand what I got wrong with that commit?

You did it right, but you actually need to add two allowlist entries. Stubtest emits two errors for this class because of the fact that the class is declared in one module, but then imported and re-exported from another module!

AlexWaygood:
> Added via [d8d8d0b](https://github.com/python/typeshed/commit/d8d8d0b9c9bb7eec0e4683e59a53c7f16c157313). Hopefully I'll be able to contribute at some point to `mypy`, currently I'm not comfortable enough with the [visitor design pattern](https://refactoring.guru/design-patterns/visitor) to feel confident beyond small refactoring.

It's a huge and complicated codebase! I've made a few contributions now, but there's still a lot that I don't really understand. Stubtest is a nice place to start contributing, though :)

stubs/urllib3/urllib3/connectionpool.pyi
```diff
@@ -29,48 +34,52 @@ Retry = retry.Retry
...
-    def __enter__(self): ...
-    def __exit__(self, exc_type, exc_val, exc_tb): ...
-    def close(self): ...
+    scheme: ClassVar[str | None]
+    QueueCls: ClassVar[Any]
+    host: str
+    port: int | None
+    def __init__(self, host: str, port: int | None = ...) -> None: ...
+    def __enter__(self: Self) -> Self: ...
+    def __exit__(self, exc_type, exc_val, exc_tb) -> bool: ...
```
AlexWaygood, line 50: The return type here is quite important -- both mypy and pyright do quite heavy special-casing of `__exit__` methods. If they see an `__exit__` method that's annotated as returning `bool`, they'll assume it's an `__exit__` method that can sometimes swallow exceptions. But this method _can't_, because it always returns `False`, never `True`.

```suggestion
    def __exit__(self, exc_type: type[BaseException] | None, exc_val: BaseException | None, exc_tb: TracebackType | None) -> Literal[False]: ...
```

stubs/urllib3/urllib3/connectionpool.pyi
```diff
@@ -1,9 +1,14 @@
-from typing import Any
+from _typeshed import Self
+from collections.abc import Mapping
+from logging import Logger
+from queue import LifoQueue
```
AlexWaygood, line 4: This module imports `LifoQueue` from another submodule of `urllib3`, _not_ the stdlib `queue` module

```suggestion
from .util.queue import LifoQueue
```

stubs/urllib3/urllib3/connectionpool.pyi
```diff
@@ -29,48 +34,52 @@ Retry = retry.Retry
...
-    scheme: Any
-    QueueCls: Any
-    host: Any
-    port: Any
-    def __init__(self, host, port=...) -> None: ...
-    def __enter__(self): ...
-    def __exit__(self, exc_type, exc_val, exc_tb): ...
-    def close(self): ...
+    scheme: ClassVar[str | None]
+    QueueCls: ClassVar[Any]
```
AlexWaygood, line 45:
```suggestion
    QueueCls: ClassVar[queue.Queue]
```

Where `queue` is the stdlib `queue` module

stubs/urllib3/urllib3/connectionpool.pyi
```diff
@@ -29,48 +34,52 @@ Retry = retry.Retry
...
-    timeout: Any
-    retries: Any
-    pool: Any
-    block: Any
-    proxy: Any
-    proxy_headers: Any
-    num_connections: Any
-    num_requests: Any
+    scheme: ClassVar[str]
+    ConnectionCls: ClassVar[type[HTTPConnection]]
```
AlexWaygood, line 55: I believe the `Union` type on the main branch is deliberate here: while `HTTPSConnectionPool`s can only have `HTTPSConnection`s, `HTTPConnectionPool`s can have `HTTPConnection`s or `HTTPSConnection`s: https://github.com/urllib3/urllib3/blob/1e985050bb498f15895478bd61acc002be83ae65/src/urllib3/connectionpool.py#L169

```suggestion
    ConnectionCls: ClassVar[type[HTTPConnection | HTTPSConnection]]
```

stubs/urllib3/urllib3/connectionpool.pyi
```diff
@@ -29,48 +34,52 @@ Retry = retry.Retry
...
+    ResponseCls: ClassVar[type[HTTPResponse]]
+    strict: bool
+    timeout: _Timeout
+    retries: _Retries | None
+    pool: LifoQueue[Any] | None
+    block: bool
+    proxy: Url | None
+    proxy_headers: Mapping[str, str] | None
```
AlexWaygood, line 63: If `None` is passed into the `__init__` parameter, it's set to an empty dictionary: https://github.com/urllib3/urllib3/blob/aa3def7d242525e6e854991247c4b68583d15135/src/urllib3/connectionpool.py#L202

```suggestion
    proxy_headers: Mapping[str, str]
```

stubs/urllib3/urllib3/connectionpool.pyi
```diff
@@ -87,35 +96,33 @@ class HTTPConnectionPool(ConnectionPool, RequestMethods):
...
-    ca_certs: Any
-    ssl_version: Any
-    assert_hostname: Any
-    assert_fingerprint: Any
+    key_file: str | None
+    cert_file: str | None
+    cert_reqs: int | str | None
+    ca_certs: str | None
+    ssl_version: int | str | None
+    assert_hostname: str | bool | None
```
AlexWaygood, line 104: It's `str | Literal[False] | None` on the `main` branch https://github.com/urllib3/urllib3/blob/1e985050bb498f15895478bd61acc002be83ae65/src/urllib3/connectionpool.py#L942

```suggestion
    assert_hostname: str | Literal[False] | None
```

stubs/urllib3/urllib3/connectionpool.pyi
```diff
@@ -87,35 +96,33 @@ class HTTPConnectionPool(ConnectionPool, RequestMethods):
...
+        headers: Mapping[str, str] | None = ...,
+        retries: _Retries | None = ...,
+        _proxy: Url | None = ...,
+        _proxy_headers: Mapping[str, str] | None = ...,
+        key_file: str | None = ...,
+        cert_file: str | None = ...,
+        cert_reqs: int | str | None = ...,
+        ca_certs: str | None = ...,
+        ssl_version: int | str | None = ...,
+        assert_hostname: str | bool | None = ...,
```
AlexWaygood, line 123:
```suggestion
        assert_hostname: str | Literal[False] | None = ...,
```

AlexWaygood (changes requested):

stubs/urllib3/urllib3/connectionpool.pyi
kkirsche: Ah, I had seen https://github.com/python/typeshed/issues/7258 and thought from the title that this was related to literal's in this sense (my bad on not reading the issue, just the title). Sorry about that. I'll fix that in a moment

stubs/urllib3/urllib3/connectionpool.pyi
kkirsche: Ah, sorry. That was a diff between their 1.26 and main branch. I had mixed up which tab was which, as it's LifoQueue from `queue` in `main`:
https://github.com/urllib3/urllib3/blob/main/src/urllib3/connectionpool.py#L3

Since `queue` stubs don't exist, I'll add them since the file is only 22 lines:
https://github.com/urllib3/urllib3/blob/1.26.x/src/urllib3/util/queue.py

stubs/urllib3/urllib3/connectionpool.pyi
kkirsche: Thanks, this was my misunderstanding on the use of Literal again — appreciate the find!

stubs/urllib3/urllib3/connectionpool.pyi
```diff
@@ -1,9 +1,16 @@
-from typing import Any
+from _typeshed import Self
+from collections.abc import Mapping
+from logging import Logger
+from queue import Queue
```
AlexWaygood, line 4: The fact that there are two `queue` modules here is quite confusing. I think it would be more readable if we just did

```python
import queue
import urllib3.util.queue
```

And then referred to `LifoQueue` as `urllib3.util.queue.LifoQueue`, and `Queue` as `queue.Queue`. It's more verbose, but it's a lot less confusing, in my opinion.

stubs/urllib3/urllib3/connectionpool.pyi
Akuli: Their `import queue` imports the standard-library module named `queue`. Anything else would have to be `from urllib.util import queue` or `from .util import queue` or something similar. Here are the stubs for the standard-library `queue` module: https://github.com/python/typeshed/blob/master/stdlib/queue.pyi

stubs/urllib3/urllib3/connectionpool.pyi
kkirsche: That's fixed because that should have been `LifoQueue` in both locations after reviewing 

https://github.com/urllib3/urllib3/blob/1.26.x/src/urllib3/connectionpool.py

stubs/urllib3/urllib3/connectionpool.pyi
```diff
@@ -44,7 +44,7 @@ log: Logger
 
 class ConnectionPool:
     scheme: ClassVar[str | None]
-    QueueCls: ClassVar[Queue[Any]]
+    QueueCls: ClassVar[LifoQueue]
```
AlexWaygood, line 46: No, stubtest was complaining because you had `ClassVar[Queue[Any]]`, when you should have had `ClassVar[type[Queue[Any]]]`

kkirsche: That should be fixed in the latest revision

AlexWaygood: It's not, because (putting aside the disagreement over which queue class it should be), you're saying it's an instance of a class, but it's not, it's the class itself. Even if you want `LifoQueue`, it needs to be `ClassVar[type[LifoQueue]]`, not `ClassVar[LifoQueue]`

AlexWaygood: ^You made that update.

kkirsche: Just a miscommunication caused by using async comments :)

AlexWaygood: Race conditions! Everywhere I look! More race conditions! :)

stubs/urllib3/urllib3/connectionpool.pyi
```diff
@@ -29,48 +35,54 @@ Retry = retry.Retry
...
-    scheme: Any
-    QueueCls: Any
-    host: Any
-    port: Any
-    def __init__(self, host, port=...) -> None: ...
-    def __enter__(self): ...
-    def __exit__(self, exc_type, exc_val, exc_tb): ...
-    def close(self): ...
+    scheme: ClassVar[str | None]
+    QueueCls: ClassVar[type[LifoQueue]]
```
AlexWaygood, line 46: I still think this would be better as the stdlib `Queue[Any]`. It's true that at runtime, `ConnectionPool.QueueCls` is `urllib3.util.queue.LifoQueue`. But `ConnectionPool` is a base class for other `ConnectionPool` classes, and I don't think (from my impression, from reading the code) that it's mandatory for all subclasses to have their `QueueCls` attribute be a subclass of `urllib3.util.queue.LifoQueue`. It seems to me that anything with an interface similar to the stdlib `queue.Queue` class will do. There's evidence for this in the fact that they scrapped the `urllib3.util.queue` submodule in the `main` branch and switched to using the stdlib `queue` module.

AlexWaygood: Anyway, this is minor, other than this it looks good to me now :)

AlexWaygood (approved):

stubs/urllib3/urllib3/connectionpool.pyi
kkirsche: 👍  I think I got mixed up between the `type[_QueueT]` change and the `Queue[Any]` change. I agree with you on that.

If you are comfortable with it, I will update this to: `ClassVar[type[Queue[Any]]]`

and change the imports to be fully qualified to differentiate between the two different queues.

Sound good?

AlexWaygood: That sounds good to me! 👍

kkirsche: That was changed in 3e231f4 (I expect a pre-commit will trigger). The only thing I may have missed was if there was a better way to import the `LifoQueue` to avoid the overlapping module names. As it stands, I've updated it to `import queue` for the standard library instance and usage as `queue.Queue[Any]` and have left the urllib3 `LifoQueue` as `from .util.queue import LifoQueue`.

Hopefully thats correct

AlexWaygood: How about `from .util import queue as urllib3queue`? And then refer to it as `urllib3queue.LifoQueue`

kkirsche: I had thought that was going to trigger a stubtest warning about it not existing at runtime, but I could just be getting confused about what is checked and what isn't. 

That's a good approach, thank you for the suggestion, worst case we can do it as `from .util import queue as _urllib3queue` to avoid the check

AlexWaygood: All imports in stub files are considered "private to the stub" (not re-exported), _unless_ they fall into one of the two categories:

- `import foo as foo` -- the `as` has to be truly redundant; if you do `import foo as bar`, it's still considered private, but `import foo as foo` is considered a public re-export.
- `from foo import *`

AlexWaygood: -- A third exception is if a name is included in `__all__`, it's always considered re-exported, even if it's imported via a means that would usually make it a "private" import.

AlexWaygood (approved):

srittau (approved):
Not an in-depth review, but everything looked fine at first glance. I'll defer to @AlexWaygood's judgement.

AlexWaygood:
Let's land it. Thanks for working on this @kkirsche, this was a tricky one in lots of ways!

kkirsche:
My pleasure, trying to work towards larger and loftier goals, so hopefully experiences like this will help

PR #8460, opened 2022-08-01
Adjust urllib3.util.url

gruebel:
Added the missing type hints in `urllib3.util.url` and changed the attributes of `Url` to reflect the actual attributes, which come from the subclassed `namedtuple`


stubs/urllib3/urllib3/util/url.pyi
```diff
@@ -1,13 +1,17 @@
...
 
 class Url:
-    slots: Any
+    auth: str | None
+    fragment: str | None
+    host: str | None
+    path: str | None
+    port: str | None
+    query: str | None
+    scheme: str | None
```
AlexWaygood, line 14: At runtime, `Url` is a class that inherits from a `NamedTuple` that is also called `Url`. We can't mimic that in the stub exactly, but we can come pretty close by doing something like this:

```python
class _UrlBase(NamedTuple):
    auth: str | None
    fragment: str | None
    host: str | None
    path: str | None
    port: str | None
    query: str | None
    scheme: str | None

class Url(_UrlBase):
    # rest of the class definition as it is in the stub already
```

(GitHub won't let me use the "suggested changes" changes here because this line range includes some deleted lines.)

gruebel: very interesting, thanks for the suggestion 👍

AlexWaygood (approved):
Thanks!

gruebel:
@AlexWaygood just out of curiosity, when do you decide to create and publish a new version of third party stubs?

AlexWaygood:
> @AlexWaygood just out of curiosity, when do you decide to create and publish a new version of third party stubs?

The magic typeshed bots automatically create and publish the stubs packages. I think the workflow runs once every few hours, but I'm not sure :)

gruebel:
> > @AlexWaygood just out of curiosity, when do you decide to create and publish a new version of third party stubs?
> 
> The magic typeshed bots automatically create and publish the stubs packages. I think the workflow runs once every few hours, but I'm not sure :)

Nice, thanks! Looking forward to see the magic 😄

PR #8465, opened 2022-08-02
Remove un-needed `__hash__` methods from `stdlib`

sobolevn:
[While working](https://github.com/typeddjango/django-stubs/pull/689#discussion_r935253811) on stubs for `django` I've noticed that sometimes people redefine `__hash__`.

But, this is not what we do for other methods. For example, if base class `A` defines `foo`, its child class `B` does not need to also define `foo`. Parent's definition will be used by default. We only annotate redefinitions.

So, `__hash__` [is defined](https://github.com/python/typeshed/blob/df4618ef2dec1d27e363da0fc4ba8b00a2c811fe/stdlib/builtins.pyi#L99) in `object`. Child classes may not redefine it, unless:
- `__hash__` returns something different, not `int` (like `enum.Enum` does right now)
- `__hash__` is set to be `ClassVar[None]`
- `__hash__` is reset after parent's `ClassVar[None]` to indicate that subclass is immutable again
- `__hash__` is required to be set explicitly as a part of `@abstractmethod` parent's def, like `numbers.Number` does

Mypy sample: https://mypy-play.net/?mypy=latest&python=3.10&gist=cef2a15be8adfa8d646d260232e6b38a

```python
from typing import Hashable

class MyHash:
    def __hash__(self) -> int:
        ...

class MyClass: ...

def requires_hash(o: Hashable): ...

requires_hash(MyHash())   # ok
requires_hash(MyClass())  # ok
```

What do others think?


srittau (approved):
LGTM, although I'd like another maintainer to confirm.

AlexWaygood (approved):
LGTM too, but I'd like to get @JelleZijlstra's feedback. I think he had some Thoughts when I tried to do something along these lines a while back with some other methods.

sobolevn:
@JelleZijlstra frienly ping 🙂

JelleZijlstra (approved):

PR #8466, opened 2022-08-02
Improve `multiprocessing.context` module

sobolevn:
Two main changes:
1. I think it is better to use assignment in case of alises where possible. For example, `Process` is simply defined as `Process = Process` in [cpython's `DefaultContext`](https://github.com/python/cpython/blob/75a6441718dcbc65d993c9544e67e25bef120e82/Lib/multiprocessing/context.py#L231). This is more accurate: it is set as an explicit classvar and we have full type info (`type[Process]` is wider)
2. There's no need to redefine `SpawnContext` and `SpawnProcess` in `if`. They are the same in case of typing. Only implementation details differ: 
  - https://github.com/python/cpython/blob/75a6441718dcbc65d993c9544e67e25bef120e82/Lib/multiprocessing/context.py#L283-L288
  - https://github.com/python/cpython/blob/75a6441718dcbc65d993c9544e67e25bef120e82/Lib/multiprocessing/context.py#L331-L336

### DIFF
--- a/stdlib/multiprocessing/context.pyi
+++ b/stdlib/multiprocessing/context.pyi
@@ -3,10 +3,9 @@ import sys
 from collections.abc import Callable, Iterable, Sequence
 from ctypes import _CData
 from logging import Logger
-from multiprocessing import queues, synchronize, popen_spawn_win32, popen_forkserver, popen_spawn_posix, popen_fork
+from multiprocessing import popen_fork, popen_forkserver, popen_spawn_posix, popen_spawn_win32, queues, synchronize
 from multiprocessing.connection import _ConnectionBase
 from multiprocessing.managers import SyncManager
-from multiprocessing.popen import Popen
 from multiprocessing.process import BaseProcess
 from multiprocessing.sharedctypes import SynchronizedArray, SynchronizedBase
 from typing import Any, ClassVar, TypeVar, overload
@@ -172,6 +171,7 @@ if sys.platform != "win32":
     class ForkServerContext(BaseContext):
         _name: str
         Process: ClassVar[type[ForkServerProcess]]
+
 else:
     class SpawnProcess(BaseProcess):
         _start_method: str
@@ -181,7 +181,6 @@ else:
     class SpawnContext(BaseContext):
         _name: str
         Process: ClassVar[type[SpawnProcess]]
-
 
 def _force_start_method(method: str) -> None: ...
 def get_spawning_popen() -> Any | None: ...
--- a/stdlib/multiprocessing/context.pyi
+++ b/stdlib/multiprocessing/context.pyi
@@ -1,5 +1,4 @@
 import ctypes
-import multiprocessing
 import sys
 from collections.abc import Callable, Iterable, Sequence
 from ctypes import _CData
@@ -27,11 +26,10 @@ class TimeoutError(ProcessError): ...
 class AuthenticationError(ProcessError): ...
 
 class BaseContext:
-    Process: type[BaseProcess]
-    ProcessError: type[Exception]
-    BufferTooShort: type[Exception]
-    TimeoutError: type[Exception]
-    AuthenticationError: type[Exception]
+    ProcessError = ProcessError
+    BufferTooShort = BufferTooShort
+    TimeoutError = TimeoutError
+    AuthenticationError = AuthenticationError
 
     # N.B. The methods below are applied at runtime to generate
     # multiprocessing.*, so the signatures should be identical (modulo self).
@@ -137,7 +135,6 @@ class Process(BaseProcess):
     def _Popen(process_obj: BaseProcess) -> DefaultContext: ...
 
 class DefaultContext(BaseContext):
-    Process: type[multiprocessing.Process]
     def __init__(self, context: BaseContext) -> None: ...
     def set_start_method(self, method: str | None, force: bool = ...) -> None: ...
     def get_start_method(self, allow_none: bool = ...) -> str: ...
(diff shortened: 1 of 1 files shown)


sobolevn:
Oh, looks like we have a problem with `pyright`:
>  /home/runner/work/typeshed/typeshed/stdlib/multiprocessing/context.pyi:29:20 - error: Type alias "ProcessError" cannot use itself in its definition (reportGeneralTypeIssues)

AlexWaygood:
> 1. This is more accurate: it is set as an explicit classvar

FYI, there's actually an annoying mypy bug that means that assignments inside class bodies are interpreted as instance attributes rather than `ClassVar`s. This bug definitely exists for `.pyi` files, and might also exist for `.py` files. Not sure if there's an open bug report about it or not. 

And yeah, there's the pyright bug w.r.t. variable scope, that you've already identified. We've come across it a few times now (see `multiprocessing.reduction` for some hacky workarounds). Someone should probably file a bug report for that, too 😅

stdlib/multiprocessing/context.pyi
```diff
@@ -27,11 +26,11 @@ class TimeoutError(ProcessError): ...
 class AuthenticationError(ProcessError): ...
 
 class BaseContext:
-    Process: type[BaseProcess]
-    ProcessError: type[Exception]
-    BufferTooShort: type[Exception]
-    TimeoutError: type[Exception]
-    AuthenticationError: type[Exception]
+    Process: ClassVar[type[BaseProcess]]
```
srittau, line 29: I don't see this field, at least in Python 3.9.

stdlib/multiprocessing/context.pyi
```diff
@@ -147,43 +146,33 @@ class DefaultContext(BaseContext):
 
 _default_context: DefaultContext
 
+class SpawnProcess(BaseProcess):
+    _start_method: str
+    @staticmethod
+    def _Popen(process_obj: BaseProcess) -> SpawnProcess: ...
```
srittau, line 152: I know you just moved this, but this returns `multiprocessing.popen.Popen`.

srittau (changes requested):

stdlib/multiprocessing/context.pyi
sobolevn: Hm, it exists here: https://github.com/python/typeshed/blob/df4618ef2dec1d27e363da0fc4ba8b00a2c811fe/stdlib/multiprocessing/context.pyi#L30

I just added a `ClassVar` here 🤔

srittau: Doesn't exist in HEAD either: https://github.com/python/cpython/blob/75a6441718dcbc65d993c9544e67e25bef120e82/Lib/multiprocessing/context.py#L30

Let's just remove it.

stdlib/multiprocessing/context.pyi
```diff
@@ -27,11 +26,11 @@ class TimeoutError(ProcessError): ...
...
-    Process: type[BaseProcess]
-    ProcessError: type[Exception]
-    BufferTooShort: type[Exception]
-    TimeoutError: type[Exception]
-    AuthenticationError: type[Exception]
+    Process: ClassVar[type[BaseProcess]]
+    ProcessError: ClassVar[type[Exception]]
+    BufferTooShort: ClassVar[type[Exception]]
+    TimeoutError: ClassVar[type[Exception]]
+    AuthenticationError: ClassVar[type[Exception]]
```
AlexWaygood, line 33: pyright is fine with this -- the pyright variable-scope bug only exists for assignments, not attribute annotations

```suggestion
    ProcessError: ClassVar[type[ProcessError]]
    BufferTooShort: ClassVar[type[BufferTooShort]]
    TimeoutError: ClassVar[type[TimeoutError]]
    AuthenticationError: ClassVar[type[AuthenticationError]]
```

stdlib/multiprocessing/context.pyi
```diff
@@ -146,33 +145,43 @@ class DefaultContext(BaseContext):
...
         Process: ClassVar[type[ForkServerProcess]]
+else:
+    class SpawnProcess(BaseProcess):
+        _start_method: str
+        @staticmethod
+        def _Popen(process_obj: BaseProcess) -> popen_spawn_win32.Popen: ...
+
+    class SpawnContext(BaseContext):
+        _name: str
+        Process: ClassVar[type[SpawnProcess]]
```
AlexWaygood, line 183: Can we merge these class definitions with the ones in the `!= win32"` branch? Only the return type of `SpawnProcess._Popen` seems to be different; I think we can just do a conditional definition of that one staticmethod inside the class.

stdlib/multiprocessing/context.pyi
```diff
@@ -1,13 +1,11 @@
 import ctypes
-import multiprocessing
 import sys
 from collections.abc import Callable, Iterable, Sequence
 from ctypes import _CData
 from logging import Logger
-from multiprocessing import queues, synchronize
+from multiprocessing import popen_fork, popen_forkserver, popen_spawn_posix, popen_spawn_win32, queues, synchronize
 from multiprocessing.connection import _ConnectionBase
 from multiprocessing.managers import SyncManager
```
AlexWaygood, line 8:
```suggestion
from multiprocessing.managers import SyncManager
from multiprocessing.pool import Pool as _Pool
```

stdlib/multiprocessing/context.pyi
AlexWaygood: But this can wait for another PR.

AlexWaygood (approved):

stdlib/multiprocessing/context.pyi
sobolevn: Yes, I can! I'll send a new one tomorrow, any reasonable working hours are ending in my part of Russia 🙂

srittau (approved):

PR #8468, opened 2022-08-02
`io.IOBase`: correct metaclass

AlexWaygood:
`io.IOBase` has `ABCMeta` as the metaclass at runtime: https://github.com/python/cpython/blob/75a6441718dcbc65d993c9544e67e25bef120e82/Lib/io.py#L71

But not currently in the stub. This is causing an unexpected stubtest failure for a `urllib3` PR here: https://github.com/python/typeshed/runs/7633083101?check_suite_focus=true

### DIFF
--- a/stdlib/io.pyi
+++ b/stdlib/io.pyi
@@ -1,3 +1,4 @@
+import abc
 import builtins
 import codecs
 import sys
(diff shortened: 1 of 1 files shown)


srittau (approved):
I would have expected our tests to explode. Let's wait for a third opinion, but LGTM.

JelleZijlstra (approved):
This isn't part of the hierarchy for `typing.IO`, so makes sense for it to be relatively harmless.

Issue #8082, opened 2022-06-15
`pyvmomi.vmodl`: `ManagedObjectNotFound` should be derived from `Exception`


sodul:
Mypy is raising errors where we use vmodl.fault.ManagedObjectNotFound where an Exception instance is expected.

Python 3.10.5
mypy==0.961
pyvmomi==7.0.3
types-pyvmomi==7.0.7

I tested in the repl and the instances are inheriting from Exception at runtime:
```
>>> isinstance(vmodl.fault.ManagedObjectNotFound(), Exception)
True
```

The pyi file is declaring it as a plain object which is false:
```
class InvalidArgument(Exception): ...

class ManagedObjectNotFound:
    obj: ManagedObject
```

It should be updated with the correct inheritance:
```
>>> vmodl.fault.ManagedObjectNotFound.__bases__
(<class 'pyVmomi.VmomiSupport.vmodl.RuntimeFault'>,)

>>> vmodl.RuntimeFault
<class 'pyVmomi.VmomiSupport.vmodl.RuntimeFault'>

>>> vmodl.RuntimeFault.__bases__
(<class 'pyVmomi.VmomiSupport.vmodl.MethodFault'>,)

>>> vmodl.MethodFault.__bases__
(<class 'pyVmomi.VmomiSupport.vmodl.DynamicData'>, <class 'Exception'>)
```

JelleZijlstra:
PRs are welcome!

PR #8469, opened 2022-08-02
Update pyvmomi to address #8082 and use `Incomplete` type instead of `Any`

kkirsche:
fix: #8082 RE: inherit from `Exception`

Use `Incomplete` instead of `Any` where `# incomplete` comments were in place.


srittau:
I haven't lookup at the pytype error, but LGTM apart from that.

kkirsche:
I'm not honestly sure what's happening in `pytype` to cause this error:

```
  File "/opt/hostedtoolcache/Python/3.10.5/x64/lib/python3.10/site-packages/pytype/load_pytd.py", line 327, in resolve_external_types
    mod_ast = mod_ast.Visit(visitors.LookupExternalTypes(
  File "/opt/hostedtoolcache/Python/3.10.5/x64/lib/python3.10/site-packages/pytype/pytd/parse/node.py", line [99](https://github.com/python/typeshed/runs/7635144002?check_suite_focus=true#step:5:100), in Visit
    return _Visit(self, visitor, *args, **kwargs)
  File "/opt/hostedtoolcache/Python/3.10.5/x64/lib/python3.10/site-packages/pytype/pytd/parse/node.py", line 117, in _Visit
    return _VisitNode(node, visitor, *args, **kwargs)
  File "/opt/hostedtoolcache/Python/3.10.5/x64/lib/python3.10/site-packages/pytype/pytd/parse/node.py", line 222, in _VisitNode
    new_node = visitor.Visit(new_node, *args, **kwargs)
  File "/opt/hostedtoolcache/Python/3.10.5/x64/lib/python3.10/site-packages/pytype/pytd/base_visitor.py", line 203, in Visit
    return self.visit_functions[node.__class__.__name__](
  File "/opt/hostedtoolcache/Python/3.10.5/x64/lib/python3.10/site-packages/pytype/pytd/visitors.py", line 631, in VisitTypeDeclUnit
    new_aliases = self._HandleDuplicates(new_aliases)
  File "/opt/hostedtoolcache/Python/3.10.5/x64/lib/python3.10/site-packages/pytype/pytd/visitors.py", line 592, in _HandleDuplicates
    raise KeyError("Duplicate top level items: %r, %r" % (
KeyError: "Duplicate top level items: None, '_typeshed.Incomplete'"
```

I don't see a duplicate in this, but I'm not clear on the purpose of this test to really investigate it

stubs/pyvmomi/pyVmomi/vmodl/query.pyi
```diff
@@ -13,12 +14,12 @@ class PropertyCollector:
 
     class TraversalSpec:
         def __init__(
-            self, *, path: str = ..., skip: bool = ..., type: type[ContainerView] = ..., **kwargs: Any  # incomplete
+            self, *, path: str = ..., skip: bool = ..., type: type[ContainerView] = ..., **kwargs: Incomplete
```
AlexWaygood, line 17:
```suggestion
            self, *, path: str = ..., skip: bool = ..., type: type[ContainerView] = ..., **kwargs,
```

stubs/pyvmomi/pyVmomi/vmodl/query.pyi
```diff
@@ -38,19 +39,17 @@ class PropertyCollector:
             *,
             propSet: list[PropertyCollector.PropertySpec] = ...,
             objectSet: list[PropertyCollector.ObjectSpec] = ...,
-            **kwargs: Any,  # incomplete
+            **kwargs: Incomplete,
```
AlexWaygood, line 42:
```suggestion
            **kwargs,
```

stubs/pyvmomi/pyVmomi/vmodl/query.pyi
```diff
@@ -38,19 +39,17 @@ class PropertyCollector:
...
         propSet: list[PropertyCollector.PropertySpec]
         objectSet: list[PropertyCollector.ObjectSpec]
-        def __getattr__(self, name: str) -> Any: ...  # incomplete
+        def __getattr__(self, name: str) -> Incomplete: ...
 
     class ObjectContent:
-        def __init__(
-            self, *, obj: ManagedEntity = ..., propSet: list[DynamicProperty] = ..., **kwargs: Any  # incomplete
-        ) -> None: ...
+        def __init__(self, *, obj: ManagedEntity = ..., propSet: list[DynamicProperty] = ..., **kwargs: Incomplete) -> None: ...
```
AlexWaygood, line 49:
```suggestion
        def __init__(self, *, obj: ManagedEntity = ..., propSet: list[DynamicProperty] = ..., **kwargs) -> None: ...
```

AlexWaygood (commented):
Nit: for unknown argument annotations , we generally prefer just to leave them unannotated rather than annotating them with `Incomplete`. It has slightly different semantics for some type checkers, such as pyright.

AlexWaygood:
I'm also not quite sure what the pytype error is trying to tell us. @rchen152, could you possibly take a look? :)

kkirsche:
> Nit: for unknown argument annotations , we generally prefer just to leave them unannotated rather than annotating them with `Incomplete` or `Any`. It has slightly different semantics for some type checkers, such as pyright.

Thanks for the clarification. As I understand it so far from other feedback, `Any` is meant to imply truly anything, but not to mean that it's unused (where Any would still be appropriate).

Where any object can be used due to not being used, `object` is instead used.

`Incomplete` seems to be the confusing bit at the moment as I would expect a resolved Unknown type in pyright to be equivalent to typeshed's  `Incomplete`. Are there any docs about that nuance?

Akuli:
Your understanding is mostly correct.

There are two ways to look at this. From a type checker's point of view:
- `object` means "could be anything, emit errors if the code assumes anything specific"
- `Any` means "do not emit errors"
- `Incomplete` means same as `Any`
- A missing annotation is either same as `Any`, or if it's mypy and there's no annotations of parameters or return type, it the function is considered to be untyped. As far as I understand it, pyright has `Unknown` for this but it behaves just like `Any`.

And from a typeshed maintainer's or contributor's point of view:
- `object` means "could be anything, this function or method doesn't care"
- `Any` means "can't be anything but it's impossible to express the situation properly with the features that type checkers have now"
- `Incomplete` or a missing annotation means "can't be anything but it would be possible to do better than using `Any`". It is basically an implicit TODO comment.

I agree that this should be documented better, probably in `CONTRIBUTING.md` :)

AlexWaygood (approved):
LGTM

srittau:
Off topic, but a few more examples for `object` arguments:

* An argument is just passed to `str()` or `repr()` or another function that accepts `object`.
* Something like this:
```python
def squeakable(it: object) -> bool:
    if not isinstance(o, Parrot):
        return False
    return it.squeak()
```

rchen152:
I have a fix for the pytype issue out for review. If all goes well, I should be able to release a new version of pytype with the fix on Wednesday or Thursday.

PR #8458, opened 2022-08-01
Use typeshed.Incomplete and object instead of Any in tkinter stubs

Akuli:
Created by replacing all `Any`s with `Incomplete` and then reviewing manually with `git add -p`.

Tkinter had quite a lot of `Any` for a few reasons:
- Old versions of stubgen generated `Any | None` (actually `Optional[Any]`) for parameters with a `None` default value.
- I used `Any` in callables before I learned about `Callable[[foo], object]` for ignored return values.
- Things that are impossible to type accurately. I left these as `Any`.

PR #8471, opened 2022-08-02
tkinter: get rid of unnecessary Incomplete

Akuli:
A continuation to #8458


hauntsaninja (commented):
I actually kinda like the explicit `Incomplete`s... I opened https://github.com/python/typeshed/issues/8472 to discuss

AlexWaygood (approved):
I like this PR (but let's continue the broader discussion in #8742 before merging)

hauntsaninja (approved):

PR #8475, opened 2022-08-03
Remove `__str__` methods

sobolevn:
I continue to remove methods that exist on `object`.

There are couple of extras:
- There are `__str__` methods with different signatures
- There are `@abstractmethod` `__str__` methods
- There are `__format__` methods with regular named, non `__fmt`, arguments (I kept them for now)


sobolevn:
Ok, I am wrong about `__format__`. It is required: https://github.com/python/mypy/blob/d2063d260ad3096df78867abf53aa247d67b401e/mypy/checkstrformat.py#L196

srittau (approved):

PR #8476, opened 2022-08-03
`weakref.finalize`: Remove TODO comment

### LINKED ISSUE #8437: `stdlib.weakref:finalize` todo clarification

AlexWaygood:
Closes #8437


AlexWaygood:
Well, seems like there's little enthusiasm for this change :)

PR #8477, opened 2022-08-03
Simplify `Enum` definition

sobolevn:
I removed a bunch of magic methods in `Enum`. Why?
1. `__dir__` is not special to be included, `object` defines `__dir__` as `-> Iterable[str]`, which is good enough (no other type in `typeshed` defines `__dir__`)
2. `__hash__` does simply return `int` (at least at 3.11): https://github.com/python/cpython/blob/89f52293281b6efc4ef666ef25e677683821f4b9/Lib/enum.py#L1184-L1185 And again, `__hash__(self) -> int` is defined in `object`
3. `__format__` is not special. It does not add any new string formatters (at least for 3.11). See https://github.com/python/cpython/blob/89f52293281b6efc4ef666ef25e677683821f4b9/Lib/enum.py#L1181-L1182 Mypy treats classes with custom `__format__` as types that provide new  string formatting patterns: https://github.com/python/mypy/blob/d2063d260ad3096df78867abf53aa247d67b401e/mypy/checkstrformat.py#L367
4. `__reduce_ex__` is defined on `object` and returns the same thing: `str | tuple[Any, ...]`, see: https://github.com/python/cpython/blob/89f52293281b6efc4ef666ef25e677683821f4b9/Lib/enum.py#L1187-L1188 and https://github.com/python/cpython/blob/89f52293281b6efc4ef666ef25e677683821f4b9/Lib/enum.py#L1277-L1292


AlexWaygood (changes requested):
`__format__` and `__reduce_ex__` have positional-only arguments on `object`, but are positional-or-keyword on `Enum`. We should keep the overrides for those methods.

I also think we should keep the `__dir__` override. The only reason why we return `Iterable[str]` from `object.__dir__` is so that the many classes that return tuples from custom `__dir__` methods don't get flagged as violating LSP. But we can be more precise here, so why shouldn't we?

AlexWaygood:
> 1. no other type in `typeshed` defines `__dir__`

There aren't that many classes that define custom `__dir__` methods :)

sobolevn:
Valid points! Updated

AlexWaygood (approved):

AlexWaygood:
Thanks!

sobolevn:
Btw, there are other `__dir__` definitions in cpython: https://cs.github.com/python/cpython?q=def+__dir__

Probably, it is worth adding.

AlexWaygood:
> Btw, there are other `__dir__` definitions in cpython: https://cs.github.com/python/cpython?q=def+__dir__
> 
> Probably, it is worth adding.

I'd be in favour of that!

sobolevn:
Ok, I will do it tomorrow 🙂

PR #8478, opened 2022-08-03
Bump pytype version.

rchen152:
Picks up a fix for the pytype test failure in
https://github.com/python/typeshed/pull/8469.


AlexWaygood (approved):
Thanks for the quick fix, as ever!

PR #8483, opened 2022-08-04
Remove duplicate definitions, step 1

sobolevn:
I've removed all duplicates found by https://github.com/python/mypy/issues/13321 from modules starting with `_` and `a`. There are lots of other modules to work on.
But, I want to split this in parts (I am not able to do it in one PR, too much!).

Except for `__hash__` method, which has its own PR: https://github.com/python/typeshed/pull/8465


srittau (approved):
I did not verify the changes against the implementation, but the changes look reasonable and the lack of stubtest or other errors gives me enough confidence.

stdlib/asyncio/events.pyi
```diff
@@ -91,7 +91,6 @@ class TimerHandle(Handle):
     def __le__(self, other: TimerHandle) -> bool: ...
     def __gt__(self, other: TimerHandle) -> bool: ...
     def __ge__(self, other: TimerHandle) -> bool: ...
-    def __eq__(self, other: object) -> bool: ...
```
sobolevn, line 94: After revisiting some implementation details, this is incorrect and needs to be reverted. Mypy treats `==` with `strict_equliaty` differently with `__eq__` defined in class and in `object` super-class.

So, I will revert all `__eq__` changes.

PR #8484, opened 2022-08-04
fix: add requests.__version__ and fix __init__.pyi

kkirsche:
Focused partial from #8439

Fixes missing __version__ file data from:
https://github.com/psf/requests/blob/main/requests/__version__.py

### DIFF
--- a/stubs/requests/requests/__init__.pyi
+++ b/stubs/requests/requests/__init__.pyi
@@ -1,4 +1,3 @@
-from logging import NullHandler as NullHandler
 
 from .__version__ import (
     __author__ as __author__,
--- a/stubs/requests/requests/__init__.pyi
+++ b/stubs/requests/requests/__init__.pyi
@@ -1,4 +1,3 @@
-
 from .__version__ import (
     __author__ as __author__,
     __author_email__ as __author_email__,
--- a/stubs/requests/requests/__init__.pyi
+++ b/stubs/requests/requests/__init__.pyi
@@ -1,5 +1,4 @@
-import logging
-from typing import Any
+from logging import NullHandler as NullHandler
 
 from .api import (
     delete as delete,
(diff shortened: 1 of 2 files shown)


stubs/requests/requests/__init__.pyi
```diff
@@ -1,6 +1,17 @@
-import logging
-from typing import Any
+from logging import NullHandler as NullHandler
```
AlexWaygood, line 1: I don't think we need this re-export. The only reason `NullHandler` is imported in `requests/__init__.py` is so that they can set up logging for the package when people import `requests`: https://github.com/psf/requests/blob/177dd90f18a8f4dc79a7d2049f0a3f4fcc5932a0/requests/__init__.py#L177. This is an implementation detail that could change at any time. By adding it to the stub, we encourage end users to rely on it being there. They shouldn't rely on it being there; and there's no reason end users should be doing `from requests import NullHandler` anyway.

Please also remember that typeshed's stubs aren't just used for type hints. They're also used for autocompletion suggestions by IDEs like VSCode and PyCharm. Having `NullHandler` pop up in autocomplete suggestions for `requests` wouldn't be useful, and could even be highly confusing for end users.

```suggestion
```

AlexWaygood (commented):
Thanks, this looks good! Just one nit:

AlexWaygood (approved):
Thanks! :)

kkirsche:
Course, thanks for your review and time.

Sorry for my reaction yesterday. It was an overreaction to the information provided and how it was provided. I hadn't provided any form of justification about why I feel implementation details should be included and you all hadn't asked if there was a reason I felt so strongly about it, and as a result I took that to be a comment on the use case and work I do in the security space rather than a comment about the intended use of typeshed. 

Hopefully you can forgive the overreaction, and I'll work to do better in the future.

AlexWaygood:
Thank you for understanding, and apologising :)

I'd like to say that I really appreciate your enthusiasm for contributing here, and that you've made some great contributions!

PR #8485, opened 2022-08-04
refactor: use re-exports in requests rather than assignments

kkirsche:
Partial: #8439

Uses re-exports of various items, adds SOCKSProxyManager and associated requirement.

X-Ref benefits from: #8467 

Source:
https://github.com/psf/requests/blob/main/requests/adapters.py


AlexWaygood (approved):
Thanks, this looks great! I haven't examined in detail whether all of these re-exports are _strictly_ necessary, but this is definitely a big improvement on the existing stub, so I'm happy for it to go in as it is.

I'll wait till the other two `requests` PRs are merged, and then merge `master` in, before merging this, but it looks good.

kkirsche:
Makes sense, I was looking at requests after the last PR from how can I break it up into smaller, easier to review sections. It'll lead to more pull requests, but my hope was that being more focused in the PR should simplify the review and merging process. So some files, like this one, I expect will get touched multiple times as the details get refined

AlexWaygood:
Smaller, more focussed PRs are definitely good!

AlexWaygood:
> 
> ```diff
> + sphinx/builders/linkcheck.py:18: note: ... from here:
> 
> bokeh (https://github.com/bokeh/bokeh)
> + bokeh/_testing/plugins/bokeh_server.py:32: note: ... from here:
> 
> cwltool (https://github.com/common-workflow-language/cwltool)
> + cwltool/docker.py:15: note: ... from here:
> 
> schemathesis (https://github.com/schemathesis/schemathesis)
> + src/schemathesis/cli/__init__.py:15: note: ... from here:
> + src/schemathesis/cli/__init__.py:15: note: ... from here:
> + src/schemathesis/cli/__init__.py:15: note: ... from here:
> - src/schemathesis/service/client.py:7: note: ... from here:
> - src/schemathesis/service/client.py:7: note: ... from here:
> - src/schemathesis/service/client.py:7: note: ... from here:
> + src/schemathesis/cli/__init__.py:15: note: ... from here:
> + src/schemathesis/cli/__init__.py:15: note: ... from here:
> + src/schemathesis/cli/__init__.py:15: note: ... from here:
> + src/schemathesis/cli/__init__.py:15: note: ... from here:
> - src/schemathesis/service/client.py:7: note: ... from here:
> - src/schemathesis/service/client.py:7: note: ... from here:
> - src/schemathesis/runner/events.py:6: note: ... from here:
> 
> schema_salad (https://github.com/common-workflow-language/schema_salad)
> + schema_salad/tests/test_cwl11.py:13: note: ... from here:
> ```

This is weird, but I think it's fine.

JelleZijlstra:
@hauntsaninja do you have any idea what's causing the mypy-primer diff above?

hauntsaninja:
Probably related to switching imports around:
https://github.com/python/mypy/blob/11fe16548c06cc67d6288d08cc6b27379ff974a0/mypy/errors.py#L854

primer is probably also obscuring the real change in message because: https://github.com/hauntsaninja/mypy_primer/pull/12

PR #8486, opened 2022-08-04
feat: add support for requests.help

kkirsche:
Partial: #8439

Adds support for requests.help

### DIFF
--- a/stubs/requests/requests/help.pyi
+++ b/stubs/requests/requests/help.pyi
@@ -1,0 +1,37 @@
+from typing_extensions import TypedDict
+
+# TODO: remove if python/typeshed#8484 is merged
+# from .__version__ import __version__ as requests_version
+
+class _VersionDict(TypedDict):
+    version: str
+
+class _OptionalVersionDict(TypedDict):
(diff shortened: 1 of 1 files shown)


kkirsche:
Source:
https://github.com/psf/requests/blob/main/requests/help.py

stubs/requests/requests/help.pyi
```diff
@@ -0,0 +1,37 @@
+from typing_extensions import TypedDict
+
+# TODO: remove if python/typeshed#8484 is merged
+# from .__version__ import __version__ as requests_version
+
```
AlexWaygood, line 5: I don't think we need this TODO comment. The import of this variable is only so that the `info()` and `main()` functions can have the necessary information to work properly. This is an implementation detail that could change at any time. End users should access the `__version__` attribute from `requests.__version__` or `requests.__init__`.

```suggestion
```

AlexWaygood (commented):
Looks good, just one nit

AlexWaygood (approved):
Thanks!

PR #8487, opened 2022-08-04
Revert `__eq__` removal

sobolevn:
Because `__eq__` is treated specially by mypy when `strict_equality` is turned on.
Without `__eq__` on a type, mypy raises this error:

```python
class A:
    def __eq__(self, o: object) -> bool: ...

class B: ...

if A() == 1:  # ok
    ...

if B() == 1: # error: Non-overlapping equality check (left operand type: "B", right operand type: "Literal[1]")
    ...
```

Refs #8483


srittau (approved):

hauntsaninja:
Thanks for catching this!

PR #8512, opened 2022-08-09
Add missing `TypedDict` methods and `ClassVar`s

AlexWaygood:
Prior to this PR, running mypy on this snippet:

```python
from typing import TypedDict

class Foo(TypedDict):
    a: str
    b: int

reveal_type(Foo.__total__)  # error: "Type[Foo]" has no attribute "__total__"
reveal_type(Foo.__required_keys__)  # error: "Type[Foo]" has no attribute "__required_keys__"
```

With this PR (using `--custom-typeshed-dir`):

```python
from typing import TypedDict

class Foo(TypedDict):
    a: str
    b: int

reveal_type(Foo.__total__)  # Revealed type is "builtins.bool"
reveal_type(Foo.__required_keys__)  # Revealed type is "builtins.frozenset[builtins.str]"
```

Fixes https://github.com/python/mypy/issues/13358


sobolevn (approved):
Thanks! I was going to do the same 🙂

AlexWaygood:
> Thanks! I was going to do the same 🙂

I realised there's also `mypy_extensions._TypedDict`. It's crazy we have three classes that need to be kept in sync, none of them checked by stubtest because they're all fictional :(

sobolevn:
Can we drop `mypy_extensions._TypedDict`?

AlexWaygood:
> Can we drop `mypy_extensions._TypedDict`?

I have no idea. If we can, then that's great, but I don't want to break anybody who's still importing `TypedDict` from `mypy_extensions`. I bet some codebases still do that.

sobolevn:
I think that we can try to at least re-export `typing_extensions` where possible 🤔

AlexWaygood:
> I think that we can try to at least re-export `typing_extensions` where possible 🤔

We can give it a go, but let's leave it to another PR 🙂

stubs/mypy-extensions/mypy_extensions.pyi
```diff
@@ -1,13 +1,19 @@
...
 _T = TypeVar("_T")
 _U = TypeVar("_U")
 
 # Internal mypy fallback type for all typed dicts (does not exist at runtime)
+# N.B. Keep this mostly in sync with typing(_extensions)._TypedDict
+@type_check_only
 class _TypedDict(Mapping[str, object], metaclass=abc.ABCMeta):
+    __total__: ClassVar[bool]
+    # Unlike typing(_extensions).TypedDict,
+    # subclasses of mypy_extensions.TypedDict do NOT have the __required_keys__ and __optional_keys__ ClassVars
```
AlexWaygood, line 16: @sobolevn, this is a good reason why we can't just re-export `typing_extensions._TypedDict` in `mypy_extensions` :(

PR #8516, opened 2022-08-10
Improve `email.headerregistry.HeaderRegistry` definition

sobolevn:
Source: https://github.com/python/cpython/blob/325ae93b6b7113cd4a4c2ce441615ae7def779e2/Lib/email/headerregistry.py#L562-L567
Refs https://github.com/python/mypy/pull/13367#issuecomment-1210311506

### DIFF
--- a/stdlib/email/headerregistry.pyi
+++ b/stdlib/email/headerregistry.pyi
@@ -142,9 +142,9 @@ if sys.version_info >= (3, 8):
         def value_parser(value: str) -> MessageID: ...
 
 class _HeaderParser(Protocol):
-    max_count: int | None
+    max_count: ClassVar[Literal[1] | None]
     @staticmethod
-    def value_parser(value: str) -> Any: ...
+    def value_parser(value: str) -> TokenList: ...
     @classmethod
     def parse(cls, value: str, kwds: dict[str, Any]) -> None: ...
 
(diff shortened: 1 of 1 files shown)


stdlib/email/headerregistry.pyi
```diff
@@ -141,9 +141,20 @@ if sys.version_info >= (3, 8):
         @staticmethod
         def value_parser(value: str) -> MessageID: ...
 
+class _HeaderParser(Protocol):
+    max_count: int | None
+    @staticmethod
+    def value_parser(value: str) -> Any: ...
```
AlexWaygood, line 147: All the `value_parser` methods for all the various `Header` classes return a class that's a subclass of `email._header_value_parser.TokenList`

```suggestion
    def value_parser(value: str) -> TokenList: ...
```

stdlib/email/headerregistry.pyi
```diff
@@ -141,9 +141,20 @@ if sys.version_info >= (3, 8):
...
+    @classmethod
+    def parse(cls, value: str, kwds: dict[str, Any]) -> None: ...
+
 class HeaderRegistry:
+    registry: dict[str, type[_HeaderParser]]
+    base_class: type[BaseHeader]
+    default_class: type[_HeaderParser]
+    use_default_map: bool
```
AlexWaygood, line 155: I don't see this assigned in `__init__`

```suggestion
```

AlexWaygood (approved):

PR #8517, opened 2022-08-10
Improve `dict`'s constructor type

sobolevn:
This is quite hard. So, let me explain.

Right now when doing something like this:

```python
from typing import Iterable, Tuple
x1: Iterable[Tuple[str, bool]]
reveal_type(dict(x1, arg=True))  # N: Revealed type is "builtins.dict[builtins.str, builtins.bool]"

x2: Iterable[Tuple[int, bool]]
reveal_type(dict(x2, arg=True))  # E: Keyword argument only valid with "str" key type in call to "dict"
```

mypy has to special-case this check. Which is not good.

I think this can be easily expressed by the type-system itself.


AlexWaygood:
Please note the comments on lines 1017-1018 :)

AlexWaygood (commented):
Can you add some test cases for the various exotic ways of constructing a `dict` that the stub is trying to account for? 6 overloads is frying my brain a little :(

sobolevn:
Looks like pyright does a lot of quite different things to mypy in terms of `dict` analysis 🤔

sobolevn:
What should we do in this case? All mypy tests pass, but `pyright` fails

test_cases/stdlib/builtins/test_dict.py
```diff
@@ -0,0 +1,37 @@
+# pyright: reportUnnecessaryTypeIgnoreComment=true
+
+from typing import Dict, Generic, Iterable, NoReturn, Tuple, TypeVar
+from typing_extensions import assert_type
+
+# These do follow `__init__` overloads order:
+assert_type(dict(), Dict[NoReturn, NoReturn])
```
AlexWaygood, line 7: This one fails pyright, and also doesn't seem very idiomatic. You'd never deliberately annotate something as `dict[NoReturn, NoReturn]` in real code. Something like this should pass mypy and pyright, and be more idiomatic:

```suggestion
bad = dict()  # type: ignore
good: Dict[str, str] = dict()
assert_type(good, Dict[str, str])
```

test_cases/stdlib/builtins/test_dict.py
```diff
@@ -0,0 +1,37 @@
...
+    def keys(self) -> Iterable[_KT]:
+        ...
+
+    def __getitem__(self, __k: _KT) -> _VT:
+        ...
+
+
+kt1: KeysAndGetItem[int, str] = KeysAndGetItem()
+assert_type(dict(kt1), Dict[int, str])
+dict(kt1, arg="a")  # type: ignore
```
AlexWaygood, line 24: The fact that this test fails pyright is bad. We should figure out why that is. If we can't figure it out, we should comment out this test and leave a TODO comment.

test_cases/stdlib/builtins/test_dict.py
```diff
@@ -0,0 +1,37 @@
...
+
+kt2: KeysAndGetItem[str, int] = KeysAndGetItem()
+assert_type(dict(kt2, arg=1), Dict[str, int])
+
+i1: Iterable[Tuple[int, str]] = [(1, "a")]
+assert_type(dict(i1), Dict[int, str])
```
AlexWaygood, line 30: Does this pass mypy and pyright?

```suggestion
i1: Iterable[Tuple[int, str]] = [(1, "a"), (2, "b")]
assert_type(dict(i1), Dict[int, str])
```

test_cases/stdlib/builtins/test_dict.py
sobolevn: Do you know why I can't press "add suggestion to batch" in this repo? 🤔

sobolevn: Is this feature disabled in the settings or something?

stdlib/collections/__init__.pyi
```diff
@@ -43,7 +43,9 @@ class UserDict(MutableMapping[_KT, _VT], Generic[_KT, _VT]):
     @overload
     def __init__(self, __dict: SupportsKeysAndGetItem[_KT, _VT], **kwargs: _VT) -> None: ...
```
sobolevn, line 44: Hm, the same is true here: `SupportsKeysAndGetItem[str, _VT]` is the only possible case where `**kwargs` will be a good fit.

test_cases/stdlib/builtins/test_dict.py
```diff
@@ -0,0 +1,40 @@
...
+
+kt1: KeysAndGetItem[int, str] = KeysAndGetItem()
+assert_type(dict(kt1), Dict[int, str])
+dict(kt1, arg="a")  # type: ignore
+
+kt2: KeysAndGetItem[str, int] = KeysAndGetItem()
+assert_type(dict(kt2, arg=1), Dict[str, int])
+
+i1: Iterable[Tuple[int, str]] = [(1, "a"), (2, "b")]
+assert_type(dict(i1), Dict[int, str])
```
AlexWaygood: To work around the mypy/pyright disagreement here:

```suggestion
d = dict(i1)
assert_type(next(iter(d.keys())), int)
assert_type(d[1], str)
```

sobolevn: I tried `if MYPY:` hack, it didn't work 😞

sobolevn:
```
  /home/runner/work/typeshed/typeshed/test_cases/stdlib/builtins/test_dict.py:39:13 - error: "assert_type" mismatch: expected "int" but received "Literal[1, 2]"
  /home/runner/work/typeshed/typeshed/test_cases/stdlib/builtins/test_dict.py:40:13 - error: "assert_type" mismatch: expected "str" but received "Literal['a', 'b']"
```

🤣

sobolevn:
I think I finally solved it! 🎉 
(by removing ones that were problematic, but anyways)

I think in the future we can end up doing something like the runtime does:

```python
>>> i = [(1, True)]
>>> dict(i, arg=1.5)
{1: True, 'arg': 1.5}
```

Probably like:

```python
class dict(MutableMapping[_KT, _VT], Generic[_KT, _VT]):
    def __init__(self: dict[_KT | str, _VT | _NVT], __iterable: Iterable[tuple[_KT, _VT]], **kwargs: _NVT) -> None: ...
```

But, this will be very major change!

test_cases/stdlib/builtins/test_dict.py
```diff
@@ -0,0 +1,49 @@
...
+assert_type(dict(kt2, arg=1), Dict[str, int])
+
+i1: Iterable[Tuple[int, str]] = [(1, "a"), (2, "b")]
+# We have to skip this one because of `pyright`. It was failing like:
+# expected "Dict[int, str]" but received "dict[Literal[1, 2], Literal['a', 'b']]"
+# See https://github.com/python/typeshed/pull/8517
+# It was:
+# assert_type(dict(i1), Dict[int, str])
+# So, we only check that it does not fail:
+dict(i1)
```
AlexWaygood, line 42: One last idea:

```suggestion
def test_iterable_tuple_overload(x: Iterable[Tuple[int, str]]) -> Dict[int, str]:
    return dict(x)


i1: Iterable[Tuple[int, str]] = [(1, "a"), (2, "b")]
test_iterable_tuple_overload(i1)
```

test_cases/stdlib/builtins/test_dict.py
```diff
@@ -0,0 +1,49 @@
...
+dict(i1, arg="a")  # type: ignore
+
+i2: Iterable[Tuple[str, int]] = [("a", 1), ("b", 2)]
+assert_type(dict(i2, arg=1), Dict[str, int])
+
+i3: Iterable[str] = ["a"]
```
AlexWaygood, line 48: The generator expression is `string.split(".")`, so let's use a string with a `.` in it 😄

```suggestion
i3: Iterable[str] = ["a.b"]
```

test_cases/stdlib/builtins/test_dict.py
```diff
@@ -0,0 +1,49 @@
...
+assert_type(dict(string.split(".") for string in i3), Dict[str, str])
```
AlexWaygood, line 49:
```suggestion
assert_type(dict(string.split(".") for string in i3), Dict[str, str])
dict(["foo", "bar", "baz"])  # type: ignore
```

sobolevn:
Adding suggetions now work again, I think it is a proper time for an old meme!
index

AlexWaygood (approved):
Hooray!

Issue #8518, opened 2022-08-10
`dataclasses.asdict` definition is not correct


sobolevn:
Example:

```python
from dataclasses import dataclass, asdict

@dataclass
class D:
    x: int

asdict(D(1), dict_factory=dict)  # Argument "dict_factory" to "asdict" has incompatible type "Type[Dict[Any, Any]]"; expected "Callable[[List[Tuple[str, Any]]], Dict[_KT, _VT]]"
```

But, since `dict` is used a default implementation, it should be valid.
Source: https://github.com/python/cpython/blob/325ae93b6b7113cd4a4c2ce441615ae7def779e2/Lib/dataclasses.py#L1254

Found in https://github.com/python/mypy/pull/13367

sobolevn:
Changing `list` to `Iterable` does not work:

```python
from dataclasses import dataclass, astuple, asdict

@dataclass
class D:
    x: int

from typing import Callable, Any, TypeVar, Iterable, Tuple
_T = TypeVar('_T')
def asdict2(obj: Any, *, dict_factory: Callable[[Iterable[Tuple[str, Any]]], _T]) -> _T: ...

asdict2(D(3), dict_factory=dict)  #  error: Argument "dict_factory" to "asdict2" has incompatible type "Type[Dict[Any, Any]]"; expected "Callable[[Iterable[Tuple[str, Any]]], Dict[_KT, _VT]]"
```

5tefan:
As a workaround, I have noticed that annotating the return value will succeed with mypy. 

```python
from typing import Any, Dict
from dataclasses import dataclass, asdict

@dataclass
class D:
    x: int

out: Dict[str, Any] = asdict(D(1), dict_factory=dict)  # Success: no issues found in 1 source file
```

max-muoto:
@sobolevn Both examples work in MyPy/Pyright now, I'm assuming this can be closed out?

PR #8519, opened 2022-08-10
Change `astuple` and `asdict` factories

sobolevn:
There's a problem with `tuple_factory` and `dict_factory` in `dataclasses.pyi`
Why? 
Because the given callback type does not really match `tuple` / `dict` constructors.

Example:

```python
from dataclasses import dataclass, astuple, asdict

@dataclass
class D:
    x: int

astuple(D(1), tuple_factory=tuple)
asdict(D(2), dict_factory=dict)  # Argument "dict_factory" to "asdict" has incompatible type "Type[Dict[Any, Any]]"; expected "Callable[[List[Tuple[str, Any]]], Dict[_KT, _VT]]"
```

Source: https://github.com/python/cpython/blob/main/Lib/dataclasses.py#L1345-L1349

Closes https://github.com/python/typeshed/issues/8518


sobolevn:
I've added some simple test cases for this. Because in my opinion this annotation is very complex.

This should be something like:

```python
def asdict(obj: Any, *, dict_factory: Callable[[Iterable[Tuple[str, Any]]], _T[str, Any]]) -> _T[str, Any]: ...
```

AlexWaygood:
Here's the `homeassistant` code: https://github.com/home-assistant/core/blob/982d197ff3c493024846f38d8688985b6acd3707/homeassistant/components/repairs/websocket_api.py#L66-L78. The primer error is due to parameter types in a `Callable` being contravariant, so `Callable[[list[tuple[Any, Any]]], Any]` is not seen as a valid subtype of `Callable[[Iterable[tuple[Any, Any]]], Any]`. I think they'll just have to update their annotation from `list[tuple[Any, Any]]` to `Iterable[tuple[Any, Any]]`, if we merge this.

sobolevn:
I am not sure that `list` -> `Iterable` is the proper fix.
Because `dict` is still not allowed, see failing tests 😒

Akuli:
Minimal reproducer:

```python3
from typing import *
T = TypeVar("T")

def f(dict_factory: Callable[[Sequence[tuple[str, Any]]], T]) -> T:
    ...

f(dict)
```

Error: `Argument 1 to "f" has incompatible type "Type[Dict[Any, Any]]"; expected "Callable[[Sequence[Tuple[str, Any]]], Dict[_KT, _VT]]"`

Mypy correctly infers `T` to be the return type of `dict`, viewed as a callable, which is `dict[_KT, _VT]`. But it doesn't realize that `_KT` can be solved to a string. Changing `tuple[str, Any]` to `tuple[Any, Any]` helps, and would be a better fix than changing `list`:
- The user doesn't have to be prepared to handle arbitrary iterables. At runtime it is always a list.
- It seems like if the user types their function with `tuple[str, Any]` instead of `tuple[Any, Any]`, it would still work even though `list` is invariant.

We could also try `tuple[str | Any, Any]`.

test_cases/stdlib/test_dataclasses.py
```diff
@@ -0,0 +1,17 @@
...
+from typing_extensions import assert_type
+
+
+@dataclasses.dataclass
+class D:
+    ...
+
+
+assert_type(dataclasses.astuple(D()), Tuple[Any, ...])
+assert_type(dataclasses.astuple(D(), tuple_factory=tuple), Tuple[Any, ...])
```
JelleZijlstra, line 14: I think Eric fixed this recently (Any and Unknown being considered different in assert_type()). Let's upgrade pyright to see if it helps.

AlexWaygood:
Closing for now due to lack of activity and to keep the number of open PRs manageable. Happy to reopen if you're still interested in working on this! :)

sobolevn:
I have 0 ideas on how to fix it.

PR #8520, opened 2022-08-10
All tests in `test_cases` must follow `test_*` convention

sobolevn:
Should we add a note about `test_*` naming to `test_cases/README.md`?


AlexWaygood:
I'm fine with this, but it's just a style nit, right? The naming of the file shouldn't impact whether or not the test is run, with the way we've currently got it set up -- it should be just the location of the file that matters.

sobolevn:
Yes, mypy checks all `*.py` files. We only use it as a convention 🙂

AlexWaygood (approved):
Feel free to add a short note to the README in another PR :)

PR #8521, opened 2022-08-10
Add "Naming convention" to `test_cases/README.md`

sobolevn:
Refs https://github.com/python/typeshed/pull/8520

### DIFF
--- a/test_cases/README.md
+++ b/test_cases/README.md
@@ -56,10 +56,12 @@ provides a useful guide.
 
 Use the same top-level name for the module / package you would like to test.
 Use `test_${thing}.py` naming pattern for individual test files.
-For example:
-- Tests for `gather` defined in `stdlib/asyncio/tasks.pyi` will go into `stdlib/asyncio/test_gather.py`
-- Tests for `ExitStack` defined in `stdlib/contextlib.pyi` will go into `stdlib/test_contextlib.py`
-- Tests for `dict` defined in `stdlib/builtins.pyi` will go into `stdlib/builtins/test_dict.py`
+
+By default, tests go into a test file with the same name as the stub file, prefixed with `test_`.
+For example: `stdlib/test_contextlib.py`.
+
+If that file becomes too big, we instead create a directory with files named after individual objects being tested.
+For example: `stdlib/builtins/test_dict.py`.
 
 ### Differences to the rest of typeshed
 
--- a/test_cases/README.md
+++ b/test_cases/README.md
@@ -52,6 +52,15 @@ annotations,
 [this page](https://typing.readthedocs.io/en/latest/source/quality.html#testing-using-assert-type-and-warn-unused-ignores)
 provides a useful guide.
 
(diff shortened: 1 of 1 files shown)


test_cases/README.md
```diff
@@ -52,6 +52,15 @@ annotations,
 [this page](https://typing.readthedocs.io/en/latest/source/quality.html#testing-using-assert-type-and-warn-unused-ignores)
 provides a useful guide.
 
+### Naming convention
+
+Use the same top-level name for the module / package you would like to test.
+Use `test_${thing}.py` naming pattern for individual test files.
+For example:
+- Tests for `gather` defined in `stdlib/asyncio/tasks.pyi` will go into `stdlib/asyncio/test_gather.py`
```
JelleZijlstra, line 60:
```suggestion
- Tests for `gather` defined in `stdlib/asyncio/tasks.pyi` will go into `stdlib/asyncio/tasks/test_gather.py`
```

test_cases/README.md
```diff
@@ -52,6 +52,15 @@ annotations,
...
+- Tests for `ExitStack` defined in `stdlib/contextlib.pyi` will go into `stdlib/test_contextlib.py`
+- Tests for `dict` defined in `stdlib/builtins.pyi` will go into `stdlib/stdlib/test_dict.py`
```
JelleZijlstra, line 62:
```suggestion
- Tests for `dict` defined in `stdlib/builtins.pyi` will go into `stdlib/builtins/test_dict.py`
```

test_cases/README.md
```diff
@@ -52,6 +52,15 @@ annotations,
 [this page](https://typing.readthedocs.io/en/latest/source/quality.html#testing-using-assert-type-and-warn-unused-ignores)
 provides a useful guide.
 
+### Naming convention
+
+Use the same top-level name for the module / package you would like to test.
+Use `test_${thing}.py` naming pattern for individual test files.
+For example:
+- Tests for `gather` defined in `stdlib/asyncio/tasks.pyi` will go into `stdlib/asyncio/test_gather.py`
+- Tests for `ExitStack` defined in `stdlib/contextlib.pyi` will go into `stdlib/test_contextlib.py`
```
JelleZijlstra, line 61: These examples don't show why ExitStack goes into `test_contextlib` directly, but `pow` doesn't go into `test_builtins`. We should add something like "By default, tests go into a test file with the same name as the stub file, prefixed with `test_`. For example: test_contextlib. If that file becomes too big, we instead create a directory with files named after individual objects being tested. For example: builtins/test_dict."

JelleZijlstra (changes requested):

test_cases/README.md
sobolevn: Should we move it then? https://github.com/python/typeshed/blob/master/test_cases/stdlib/asyncio/test_gather.py

JelleZijlstra (approved):

PR #8523, opened 2022-08-10
Unpin stubtest from 3.10.5

### LINKED ISSUE #8510: Stubtest failures due to 3.10.6
3.10.6 was released a few days ago, and this caused a few stubtest failures on #8507. Unfortunately, GitHub Actions only seems to be using 3.10.6 for the Ubuntu stubtest run; the Windows and MacOS runs are still using 3.10.5. This makes it hard to update our allowlists/stubs.

Should we:

- Temporarily pin to 3.10.5 for all our stubtest runs?
- Leave the CI red for a little bit and hope the Windows and MacOS stubtest runs start using 3.10.6 soon?
- Something else?

AlexWaygood:
Fixes #8510


AlexWaygood:
We now have 3.10.6 on Ubuntu and MacOS, but still not Windows, so this is deferred until we have it on Windows as well.

AlexWaygood:
> We now have 3.10.6 on Ubuntu and MacOS, but still not Windows, so this is deferred until we have it on Windows as well.

Seems like sometimes `setup-python` gives us 3.10.6 on Windows, sometimes 3.10.5 `¯\_(ツ)_/¯`

AlexWaygood:
Looks like we're now getting 3.10.6 consistently on all platforms; I think this is now mergeable!

srittau (approved):

PR #8528, opened 2022-08-11
Additional return types for psycopg2 connections

OldSneerJaw:
Adds correct return types for DB connection and cursor methods that deal with `namedtuple` and `dict`-like cursors. Currently the types are either incomplete (e.g. the connections' `cursor` methods do not specify return types) or wrong (e.g. `DictCursor`'s `fetch*` methods should not return `tuple`s, which is currently inherited from `DictCursorBase`).


stubs/psycopg2/psycopg2/_psycopg.pyi
```diff
@@ -109,7 +109,7 @@ class cursor:
     def __exit__(
         self, type: type[BaseException] | None, value: BaseException | None, traceback: TracebackType | None
     ) -> None: ...
-    def __iter__(self: Self) -> Self: ...
+    def __iter__(self) -> Iterator[tuple[Any, ...]]: ...
```
OldSneerJaw, line 112: Just for reference from the official docs: https://www.psycopg.org/docs/cursor.html#fetch

OldSneerJaw:
I'm not sure of the best approach to solving the mypy errors in CI. The cases where it is reporting that an override's return types are incompatible with the super return types are how it is actually implemented in the underlying library's code (i.e. the stubs I've added merely reflect the reality of the code). Thoughts?

Akuli:
Stubs should reflect what the library actually does. We use `# type: ignore[override]` to silence type checker errors in situations like this.

stubs/psycopg2/psycopg2/extras.pyi
```diff
@@ -28,22 +28,42 @@ from psycopg2._range import (
...
+        name: str | bytes | None = ...,
+        *,
+        cursor_factory: Callable[..., _T_cur],
+        withhold: bool = ...,
+        scrollable: bool | None = ...,
+    ) -> _T_cur: ...
+    @overload
+    def cursor(
+        self, name: str | bytes | None, cursor_factory: Callable[..., _T_cur], withhold: bool = ..., scrollable: bool | None = ...
+    ) -> _T_cur: ...
```
OldSneerJaw, line 56: These overloads are consistent with the base class's definitions of the `cursor` method except that the default `cursor_factory` case has been updated to return `DictCursor`. This comment also applies to `RealDictConnection` and `NamedTupleConnection`.

stubs/psycopg2/psycopg2/extras.pyi
```diff
@@ -28,22 +28,42 @@ from psycopg2._range import (
...
 
 class DictCursor(DictCursorBase):
     def __init__(self, *args, **kwargs) -> None: ...
     index: Any
     def execute(self, query, vars: Any | None = ...): ...
     def callproc(self, procname, vars: Any | None = ...): ...
+    def fetchone(self) -> DictRow | None: ...  # type: ignore[override]
+    def fetchmany(self, size: int | None = ...) -> list[DictRow]: ...  # type: ignore[override]
+    def fetchall(self) -> list[DictRow]: ...  # type: ignore[override]
+    def __next__(self) -> DictRow: ...  # type: ignore[override]
```
OldSneerJaw, line 66: The return types aren't compatible with the base `cursor` class's definition, but that is consistent with the reality of the implementation. This comment also applies to the other modified cursor type defs.

OldSneerJaw:
All right. I have made several changes, including adding `# type: ignore` for the incompatible return types. I believe this PR is ready for review now.

stubs/psycopg2/psycopg2/extras.pyi
```diff
@@ -28,22 +29,42 @@ from psycopg2._range import (
...
+_T_cur = TypeVar("_T_cur", bound=_cursor)
+
 class DictCursorBase(_cursor):
     row_factory: Any
     def __init__(self, *args, **kwargs) -> None: ...
     def fetchone(self) -> tuple[Any, ...] | None: ...
     def fetchmany(self, size: int | None = ...) -> list[tuple[Any, ...]]: ...
     def fetchall(self) -> list[tuple[Any, ...]]: ...
-    def __iter__(self): ...
+    def __next__(self) -> tuple[Any, ...]: ...
```
JelleZijlstra, line 40: What is the reason for this change? Both `__iter__` and `__next__` are inherited from `cursor`, so we don't need them here.

stubs/psycopg2/psycopg2/extras.pyi
```diff
@@ -28,22 +29,42 @@ from psycopg2._range import (
...
+    def cursor(
+        self, name: str | bytes | None, cursor_factory: Callable[..., _T_cur], withhold: bool = ..., scrollable: bool | None = ...
+    ) -> _T_cur: ...
 
 class DictCursor(DictCursorBase):
     def __init__(self, *args, **kwargs) -> None: ...
     index: Any
     def execute(self, query, vars: Any | None = ...): ...
     def callproc(self, procname, vars: Any | None = ...): ...
+    def fetchone(self) -> DictRow | None: ...  # type: ignore[override]
```
JelleZijlstra, line 64: An alternative could be to make the base class generic. But that may cause problems until we have TypeVar defaults.

stubs/psycopg2/psycopg2/extras.pyi
```diff
@@ -58,31 +79,67 @@ class DictRow(list[Any]):
...
     Record: Any
     MAX_CACHE: int
     def execute(self, query, vars: Any | None = ...): ...
     def executemany(self, query, vars): ...
     def callproc(self, procname, vars: Any | None = ...): ...
-    def fetchone(self) -> tuple[Any, ...] | None: ...
-    def fetchmany(self, size: int | None = ...) -> list[tuple[Any, ...]]: ...
-    def fetchall(self) -> list[tuple[Any, ...]]: ...
-    def __iter__(self): ...
+    def fetchone(self) -> NamedTuple | None: ...  # type: ignore[override]
```
JelleZijlstra, line 139: Returning `typing.NamedTuple` is a little weird, but it apparently works with both pyright and mypy, so I suppose it's fine.

JelleZijlstra (approved):

stubs/psycopg2/psycopg2/extras.pyi
OldSneerJaw: Yes, I agree. But at least it offers a hint to the caller that they can use either list indexing or access the fields with dot-notation.

stubs/psycopg2/psycopg2/extras.pyi
OldSneerJaw: Indeed, you're right. in fact `fetchone`, `fetchmany`, `fetchall` and `row_factory` are all superfluous too. I'll clean them up.

OldSneerJaw:
One new commit for review.

PR #8535, opened 2022-08-12
Annotate Model in SQLAlchemy

mano7onam:


srittau (approved):
If I understand correctly, it's possible to pass in a different `Model` class, but I'm fine with this as is for now.

PR #8536, opened 2022-08-12
fix: align set command paramater types for redis

deontpearson:
Proposed change for [#8537](https://github.com/python/typeshed/issues/8537)


AlexWaygood (approved):
Thanks!

PR #8542, opened 2022-08-13
Fix `UDPServer` to correctly inherit from `TCPServer`

PeterJCLaw:
This is documented and implemented as such, and means that the signatures of some of the methods (notably `__init__`) actually differ from those on `BaseServer`.

This also fixes the definition of the (new in Python 3.11, via https://github.com/python/cpython/pull/30072) `allow_reuse_port` attribute on these classes that would otherwise have been in conflict with the inheritance change.


AlexWaygood (approved):
Thanks!

PR #8543, opened 2022-08-16
cryptography: Rename parameter of serialize_key_and_certificates

MaertHaekkinen:
It seems that an invalid parameter name was used for `serialize_key_and_certificates`.
`encryption_algorithm` should be used instead of `enc`.

Documentation: https://cryptography.io/en/latest/hazmat/primitives/asymmetric/serialization/#cryptography.hazmat.primitives.serialization.pkcs12.serialize_key_and_certificates

Source code: https://github.com/pyca/cryptography/blob/4a4f4d94ce5a641de3020042c70c1734af265d5e/src/cryptography/hazmat/primitives/serialization/pkcs12.py#L171


srittau:
It seems that stubtest even noticed that this was a problem, which we overrode for unknown reasons. You need to remove the corresponding line from `stubs/cryptography/@tests/stubtest_allowlist.txt`.

PR #8544, opened 2022-08-16
Remove `@final` from subclassesable types

sobolevn:
I indentified that these types can be subclassed:

```python
>>> from collections import _OrderedDictItemsView, _OrderedDictKeysView, _OrderedDictValuesView
>>> class A(_OrderedDictItemsView): ...
... 
>>> class B(_OrderedDictKeysView): ...
... 
>>> class C(_OrderedDictValuesView): ...
... 
```

Right now our stubtest does not check `@final` declaration on regular types. Maybe we should?


AlexWaygood:
```python
>>> from collections import OrderedDict
>>> class Foo(type(OrderedDict({'foo': 'bar'}).keys())): ...
...
TypeError: type 'odict_keys' is not an acceptable base type
```

The C implementation that's actually used for `OrderedDict.keys()` at runtime differs from the pure-Python implementation. This has come up before; it might be good to add a comment about it to the stub. Or alternatively, maybe we should change `OrderedDict` so that `.keys()` doesn't return `_OrderedDictKeysView`, but returns `odict_keys` or something.

> Right now our stubtest does not check `@final` declaration on regular types. Maybe we should?

This was considered in https://github.com/python/mypy/pull/12091, but rejected due to zero true positives and several false-positives (with the `collections` hits being considered as false-positives): https://github.com/python/mypy/pull/12091#issuecomment-1024887630.

PR #8547, opened 2022-08-17
Add SAFE_RESTARTABLE strategy to ldap3 Connection client_strategy parameter

jakmarkowski:
The ldap3 `Connection` class `client_strategy` parameter is missing `SAFE_RESTARTABLE` value

See the source implementation [here](https://github.com/cannatag/ldap3/blob/dev/ldap3/core/connection.py#L332)


srittau (approved):
Thanks!

PR #8548, opened 2022-08-17
Add stubs for tree-sitter-languages

Akuli:
This project has a tiny public API, documented with a couple examples in their README: https://github.com/grantjenks/py-tree-sitter-languages


stubs/tree-sitter-languages/METADATA.toml
```diff
@@ -0,0 +1,5 @@
+version = "1.2.*"
```
AlexWaygood, line 1:
```suggestion
version = "1.3.*"
```

It was released two days ago ;)

Akuli: Indeed. That release was because I made a couple pull requests there :)

AlexWaygood (approved):

PR #8549, opened 2022-08-17
Normalise use of `Never` vs `NoReturn`

AlexWaygood:
Unblocks https://github.com/PyCQA/flake8-pyi/pull/274


hauntsaninja (approved):
Looks good, assuming that you don't need to explicitly use `typing_extensions.Never` in `typing.py`

AlexWaygood:
> Looks good, assuming that you don't need to explicitly use `typing_extensions.Never` in `typing.py`

Looks like I do :)

stdlib/typing.pyi
```diff
@@ -188,7 +188,7 @@ if sys.version_info >= (3, 8):
 
 if sys.version_info >= (3, 11):
     Self: _SpecialForm
-    Never: _SpecialForm = ...
+    Never: _SpecialForm
```
AlexWaygood, line 191:
```suggestion
    Never: _SpecialForm = ...
```

PR #8550, opened 2022-08-17
Flask-SQLAlchemy: type session as Session

forana:
`SQLAlchemy().session` will always be of type `Session`: https://github.com/pallets-eco/flask-sqlalchemy/blob/2d9257e79b8246632b7518c613fb24bc523f9b1d/src/flask_sqlalchemy/__init__.py#L760 (technically a `ScopedSession`, but just `Session` will go a hell of a long way for type hinting).


AlexWaygood (approved):
Thanks!

kasium:
@forana fyi: We need to use ScopedSession since ScopedSession is not a subtype of Session. I'll push a fix

PR #8552, opened 2022-08-18
Improve classes returned from `OrderedDict.{keys(), items(), values()}`

AlexWaygood:
This is an alternative PR to #8544 (closes #8544). @sobolevn, what do you think? :)


AlexWaygood:
Ping, @sobolevn -- would be interested to hear your thoughts on this :)

sobolevn (approved):
I think that we can still keep `_OrderedDict*` classes, but I am pretty sure that nobody will miss them 😆

AlexWaygood:
> I think that we can still keep `_OrderedDict*` classes, but I am pretty sure that nobody will miss them 😆

Ugh, I guess you're right, _somebody_ somewhere is probably relying on them 😆

sobolevn (approved):

AlexWaygood:
Thanks for the review @sobolevn!

Issue #8538, opened 2022-08-12
Add overloads for `argparse.ArgumentParser.parse_x_args()`


NCPlayz:
An extension of #2566, seems like the typings for `parse_known_args`, `parse_intermixed_args`, `parse_known_intermixed_args` (and the undocumented `_parse_known_args`) of `argparse.ArgumentParser` were overlooked, as they all share the same pattern as `parse_args` for typing the namespace class.

kkirsche:
That file also seems to not exist anymore in that location, but `parse_known_args` and whatnot are documented here it seems:

* `_parse_known_args`: https://github.com/python/typeshed/blob/master/stdlib/argparse.pyi#L221
* `parse_known_args`: https://github.com/python/typeshed/blob/master/stdlib/argparse.pyi#L208
* `parse_interixed_args`: https://github.com/python/typeshed/blob/master/stdlib/argparse.pyi#L214
* `parse_known_intermixed_args`: https://github.com/python/typeshed/blob/master/stdlib/argparse.pyi#L215

So it looks like it's a matter of updating these with the proper overloads

kkirsche:
Corresponding Python source:

* `_parse_known_args`: https://github.com/python/cpython/blob/main/Lib/argparse.py#L1906
* `parse_known_args`: https://github.com/python/cpython/blob/main/Lib/argparse.py#L1868
* `parse_intermixed_args`: https://github.com/python/cpython/blob/main/Lib/argparse.py#L2372
* `parse_known_intermixed_args`: https://github.com/python/cpython/blob/main/Lib/argparse.py#L2379

Due to the argument ordering and the way the methods return (always a list[str] but the namespace can be generic or not, I'm not 100% sure how to write the overloads properly, so I'll have to defer to a typeshed team member to provide guidance or assist in fixing this, if possible.

kkirsche:
The issue I'm not sure how to properly handle in the overloads is related to properly handling having an argument without a default after an argument with a default value as we only need to overload the namespace, based on my review of the source code's behavior (unless I misread it). You're correct about the return type though, I'm just not sure how to represent it within the mypy's rules for argument orders and default values unless it's correct to just leave those out of the overloads.

Akuli:
> an argument without a default after an argument with a default value

It would be tempting to write this:

```python3
@overload
def parse_args(self, args: Iterable[str] | None = ..., namespace: None = ...) -> Namespace: ...
@overload
[...]

PR #8553, opened 2022-08-18
fix: add missing overloads for argparse.*parse_*_args

kkirsche:
fix: #8538

This attempts to address issue #8538. Huge thank you to Akuli for their help in the issue. Hopefully I got this right, apologies if I messed anything up.


kkirsche:
Yikes, I must have messed this up 😕

Akuli:
The overloads are correct (apart from `bound=Namespace` which seems unnecessary to me), but having them here means that everyone who overrides one of these methods in a subclass of `ArgumentParser` would have to copy/paste the overloads into their code.

I'm not sure what the best solution is. We could just leave the stubs as is: they would be incorrect, because the return type shouldn't always have `Namespace`, but that seems to cause less false positives than the overloads. Maybe return `Namespace | Any` so that you can at least do this without mypy errors?

```python3
parsed_args, remaining_args = parser.parse_known_args(namespace=MyNamespace())
assert isinstance(parsed_args, MyNamespace)
```

kkirsche:
My MyPy integration in VSCode was saying that not having it `bound` was an error because of how the functions interact with each other, so I restricted the TypeVar to fix those errors

kkirsche:
Yeah, the error I was seeing was:

> Overloaded function signatures 1 and 2 overlap with incompatible return types

Akuli:
The error is because without a bound, you could use `None` as your namespace. It wouldn't actually work though, because setting any attribute on `None` is an error. I don't think we have a way to express "any object that has at least one settable attribute", so we could just `# type: ignore` that error away if we decide to use overloads.

kkirsche:
Since we are using the variable to represent a namespace, why would we choose to avoid using `bound`? Not opposed to removing it, I just don't understand the motivation

Akuli:
There's no particular reason why the namespace would have to derive from `argparse.Namespace`. It can be any object that you can set attributes on, it's only used with `setattr()`. This is useful if you want to type-check the uses of your namespace.

```python3
class MyNamespace:
    foo: list[str]
    bar: int

args = parser.parse_args(namespace=MyNamespace())
print(args.foobar)  # error
```

If you inherit from `argparse.Namespace` (or just use `argparse.Namespace`), the `def __getattr__(self, name: str) -> Any: ...` we define for it would cause `args.foobar` to not be an error, and have type `Any`.

kkirsche:
Ah, ok, thank you. The namespace concept with argparse has been an abstraction that hadn't made sense to me in the past with argparse, so I appreciate the information about that.

kkirsche:
> I'm not sure what the best solution is. We could just leave the stubs as is: they would be incorrect, because the return type shouldn't always have `Namespace`, but that seems to cause less false positives than the overloads. Maybe return `Namespace | Any` so that you can at least do this without mypy errors?

I think this may be too incorrect of a representation (as I understand it, a lot of people seem to be anti-union returns 😕 ), but I figure I'll ask just to ensure we've considered all options. What about something like this:

```
    def parse_known_args(
        self, args: Sequence[str] | None = ..., namespace: Namespace | _N | None = ...
    ) -> tuple[Namespace | _N, list[str]]: ...
```

kkirsche:
Looks like removing `bound` introduces it's own set of problems, so it sounds like overloads are going to be rather noisy to downstream customers in a way that's (arguably) not going to improve the safety of those systems

Akuli:
Yeah, it seems that we have multiple bad options to choose from. Feel free to try whatever you think might work and see what mypy_primer says :)

Arguably returning the correct type when a namespace is passed isn't very important. Instead of this:

```
args = parser.parse_args(namespace=MyNamespace())
```

You could just do this:

```
args = MyNamespace()
parser.parse_args(namespace=args)
```

Having to rewrite the code only to satisfy type checkers isn't great in general, but maybe it is the least bad option we have.

Akuli:
> a lot of people seem to be anti-union returns

Your suggestion was:

```python3
def parse_known_args(
    self, args: Sequence[str] | None = ..., namespace: Namespace | _N | None = ...
) -> tuple[Namespace | _N, list[str]]: ...
```

Here returning `Namespace | _N`  doesn't have the usual disadvantages of returning a union. Because `Namespace` pretends to have any attribute whatsoever with its `__getattr__`, accessing an attribute of `Foo` on a `Foo | Namespace` shouldn't error. But accessing an attribute that doesn't exist on `Foo` would, because the attribute must exist on both union members. Overall it works a lot as if it was `Foo`, which is good. The one disadvantage is that you can't pass it to a function expecting a `Foo` as an argument, but even then it would be an improvement over the current situation.

There's one other problem with this definition. If you don't specify a `namespace` parameter, or you specify it as a `Namespace` or `None`, then `_N` is used only in the return type, and using a `TypeVar` only once is pointless. For a simpler example, consider this:

```python3
from typing import TypeVar

T = TypeVar("T")
def func(x: str | T) -> T: ...

reveal_type(func(1))     # Revealed type is "builtins.int*"
reveal_type(func("hi"))  # Revealed type is "<nothing>"
```

When you try to assign the result to a variable, mypy will then complain about needing a type annotation for it, as it doesn't know what the type is (it's `T` but there are no clues about what `T` should be solved to):

```python3
x = func("hi")  # error: Need type annotation for "x"
```

We probably don't want to require a type annotation for most uses of `.parse_args()`, it is usually called without specifying a namespace.

Issue #8513, opened 2022-08-09
sys.exit and SystemExit


Akuli:
`sys.exit()` is defined as:

https://github.com/python/typeshed/blob/c68bcc7e69249956f358da7f47028d8fd1f7443e/stdlib/sys.pyi#L224-L225

The argument ends up in the `.code` attribute of a `SystemExit` exception.

```python3
>>> import sys
>>> try: sys.exit()
... except SystemExit as e: print(repr(e.code))
... 
None
>>> try: sys.exit("lol")
... except SystemExit as e: print(repr(e.code))
... 
'lol'
>>> try: sys.exit({"foo": "bar"})
... except SystemExit as e: print(repr(e.code))
... 
{'foo': 'bar'}
```

But that's typed as:

https://github.com/python/typeshed/blob/c68bcc7e69249956f358da7f47028d8fd1f7443e/stdlib/builtins.pyi#L1793-L1794

These should be the same IMO, but what should they be? The interpreter handles uncaught `SystemExit` exceptions like this:
- If the `code` is `None`, exit the process with status zero.
- If the `code` is an integer, exit the process with that status.
- Otherwise, print the `code` (to stderr) and exit the process with status `1`.

I'd suggest using `str | int | None`, where we use `str` because we don't have any way to express "anything reasonably printable" without saying "any object". This way you can use each of the interpreter's behaviors as intended, and if you accidentally pass in e.g. `0.0`, the type checker will warn you instead of making you wonder why your process printed `0.0` and failed.

HeavyLvy:
hi👋

PR #8554, opened 2022-08-18
fix: Update sys.exit and SystemExit exception to have the same types

kkirsche:
fix: #8513

This updates the types of `sys.exit` and `SystemExit` with the recommendations of @Akuli. The issue had received 👍 s from three team members, so I wanted to help move this along for the team.


AlexWaygood (commented):
Thanks, this looks good. The mypy_primer hits are a _little_ unfortunate, but they're arguably "true positives", and I can live with them.

Two things that might be good to change, before merging:

1. Presumably `builtins.quit()` and `builtins.exit()` should be changed in the same way as `sys.exit()`?
2. Can we define a type alias in the `sys` module for `str | int | None`, and then use that for all of these functions, and the `code` attribute of `SystemExit`? I think it would help ensure that they don't drift out of sync in the future.

kkirsche:
Added 🙂

stdlib/sys.pyi
```diff
@@ -222,7 +223,7 @@ if sys.version_info >= (3, 11):
     def exception() -> BaseException | None: ...
 
 # sys.exit() accepts an optional argument of anything printable
```
AlexWaygood, line 225: This comment is now out of date. Can you delete it, and add a comment above the definition of sys._ExitCode, summarising the rationale laid out in https://github.com/python/typeshed/issues/8513#issue-1333671093 for why we're going with `str | int | None` here? (You could maybe just link to that comment in the comment.)

kkirsche: Added a link to the comment as I think to do the full rationale could end up being rather verbose

AlexWaygood: Looks good :)

AlexWaygood (approved):
LGTM, but I'll wait a little while before merging to see if any other maintainers have any thoughts :)

kkirsche:
works for me, thanks for the review

Issue #6886, opened 2022-01-10
xml typing incomplete


surgura:
While using mypy wit typeshed I noticed that some xml.minidom functions return values are annotated as `...`. I am no guru with the typing system, but I think here it is used as a placeholder. E.g. the `createDocument` returns a `Document` instance(I validated this with the std implementation code on my side), but is annotated with an ellipsis.

Are these simply still missing or is this on purpose? Excuse me if there is a typeshed todo list somewhere that I missed.

srittau:
The xml stubs are indeed incomplete. The marker for a missing return type is not `...`, though, but the missing `->` operator. For example:

```python
def createDocumentFragment(self): ...  # no return type
def createDocumentFragment(self) -> Whatever: ...  # return type
```

Any improvements are welcome!

surgura:
I see, that makes sense. I can probably add typing for a few functions, but I am not very experienced with this module.

tungol:
https://github.com/python/typeshed/pull/13349 was accepted, so I believe this ticket can be closed now.

PR #8556, opened 2022-08-18
fix: add missing type stubs for `xml.dom.minidom.DOMImplementation`, where possible

kkirsche:
This adds additional details to `xml.dom.minidom.DOMImplementation` based on what I am able to determine from:
https://github.com/python/cpython/blob/main/Lib/xml/dom/minidom.py

This begins to address #6886


stdlib/xml/dom/minidom.pyi
```diff
@@ -265,10 +265,10 @@ class Notation(Identified, Childless, Node):
     def __init__(self, name, publicId, systemId) -> None: ...
 
 class DOMImplementation(DOMImplementationLS):
-    def hasFeature(self, feature, version) -> bool: ...
-    def createDocument(self, namespaceURI: str | None, qualifiedName: str | None, doctype): ...
-    def createDocumentType(self, qualifiedName: str | None, publicId, systemId): ...
-    def getInterface(self, feature): ...
+    def hasFeature(self, feature: str, version: str | None) -> bool: ...
+    def createDocument(self, namespaceURI: str | None, qualifiedName: str | None, doctype: DocumentType | None) -> Document: ...
```
AlexWaygood, line 269: I couldn't find definitive evidence for this in the source code, but the [documentation for this method](https://docs.python.org/3/library/xml.dom.html#xml.dom.DOMImplementation.createDocument) confirms this type annotation. Looks good.

stdlib/xml/dom/minidom.pyi
```diff
@@ -265,10 +265,10 @@ class Notation(Identified, Childless, Node):
...
+    def createDocumentType(self, qualifiedName: str | None, publicId: Incomplete, systemId: Incomplete) -> Incomplete: ...
```
AlexWaygood, line 270: According to the documentation for this function, `publicId` and `systemId` should both be strings: https://docs.python.org/3/library/xml.dom.html#xml.dom.DOMImplementation.createDocumentType

And we can see from the source that this function returns an instance of `DocumentType`: https://github.com/python/cpython/blob/e0d54a4a799dae4ebdd72a16bcf287ed62ae2972/Lib/xml/dom/minidom.py#L1477

```suggestion
    def createDocumentType(self, qualifiedName: str | None, publicId: str, systemId: str) -> DocumentType: ...
```

kkirsche: Ah, I must have missed that. Thanks for catching this 👍

stdlib/xml/dom/minidom.pyi
kkirsche: I had been trying to use https://cs.github.com/?scopeName=All+repos&scope=&q=repo%3Apython%2Fcpython+path%3ALib%2Fxml%2Fdom+%22.createDocument%28%22 to work backwards through what calls were being used and comparing it to the documentation as my way of trying to find evidence, if that helps

AlexWaygood (approved):
Nice! Always great to get better types for our `xml` stubs — I think we probably all agree it's the weakest area in our stdlib stubs at the moment :)

stdlib/xml/dom/minidom.pyi
AlexWaygood: Oh, that is helpful, thanks! Especially with `xml`, it's sometimes pretty difficult to figure out the types just by studying the source code, so hints as to how you figured out non-obvious types are definitely appreciated :)

kkirsche:
> Nice! Always great to get better types for our `xml` stub


For whatever reason, security standards like OpenSCAP often default to XML standards so I end up in either `xml` or [xsdata](https://xsdata.readthedocs.io/en/latest/) more often than I'd like. I'll need to look at what's there and what's missing to see what's involved with bringing it more inline with other areas of the standard library, in terms of coverage.

AlexWaygood:
You can see here that there are a number of `xml` files that we have to exclude from the stricter pyright settings in our CI:

https://github.com/python/typeshed/blob/fb62ccabf5dfdc1a9d0cda145014cb585c77d59c/pyrightconfig.stricter.json#L9-L95

Ideally there wouldn't be any stdlib files on that list :)

AlexWaygood:
> Ideally there wouldn't be any stdlib files on that list :)

(Having said that, I wouldn't bother with trying to fix the `distutils` or `lib2to3` entries. That feels like it would be somewhat of a waste of time, since those modules are deprecated and slated for removal soon :)

PR #8558, opened 2022-08-18
argparse: type subparser helper's `add_parser`

DMRobertson:
A drive-by. I grepped for TODO and this one seemed easy enough.

### DIFF
--- a/stdlib/argparse.pyi
+++ b/stdlib/argparse.pyi
@@ -460,13 +460,16 @@ class _SubParsersAction(Action, Generic[_ArgumentParserT]):
         metavar: str | tuple[str, ...] | None = ...,
     ) -> None: ...
 
-    # The signature of add_parser matches that of ArgumentParser.__init__, except
-    # add_parser returns an ArgumentParser.
+    # Note: `add_parser` accepts all kwargs of `ArgumentParser.__init__`. It also
+    # accepts its own `help` and `aliases` kwargs.
     if sys.version_info >= (3, 9):
         def add_parser(
             self,
             name: str,
             *,
+            help: str | None = ...,
+            aliases: Sequence[str] = ...,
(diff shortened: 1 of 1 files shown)


DMRobertson:
Oh. Stubtest isn't happy because the actual method takes `**kwargs`. I don't think there's a way to express "this function uses the kwargs of another one" in this case, even with ParamSpec?

stdlib/argparse.pyi
```diff
@@ -458,8 +459,43 @@ class _SubParsersAction(Action, Generic[_ArgumentParserT]):
...
         metavar: str | tuple[str, ...] | None = ...,
     ) -> None: ...
-    # TODO: Type keyword args properly.
-    def add_parser(self, name: str, **kwargs: Any) -> _ArgumentParserT: ...
+
+    # The signature of add_parser matches that of ArgumentParser.__init__, except
+    # add_parser returns an ArgumentParser.
+    if sys.version_info >= (3, 9):
+        def add_parser(
+            self,
```
Akuli, line 467: You are missing the `name` parameter. The rest of the parameters should be also marked as keyword-only, so that `add_parser("asd1", "asd2")` is an error as it should be.

```suggestion
            self,
            name: str,
            *,
```

Also applies to the pre-3.9 definition.

Akuli:
You're right: there's no good way to say "same parameters as this other method" apart from aliasing the methods to each other, as in `alias = original_method` in the class body.

stdlib/argparse.pyi
DMRobertson: Thanks. I wasn't reading the docs carefully enough yesterday.

Akuli (approved):
Thanks! I like how there's a comment in both places that must stay in sync :)

PR #8567, opened 2022-08-19
Improve return type for psycopg2 connect function

OldSneerJaw:
When a `connection_factory` argument is provided to psycopg2's `connect` function, the function's return type now matches that of the factory output class. However, if `cursor_factory` is set and has a non-`None` value and/or `connection_factory` is not set or is `None`, the return type is simply `connection`, as before.

### DIFF
--- a/stubs/psycopg2/psycopg2/__init__.pyi
+++ b/stubs/psycopg2/psycopg2/__init__.pyi
@@ -1,4 +1,5 @@
-from typing import Any
+from collections.abc import Callable
+from typing import Any, TypeVar, overload
 
 # connection and cursor not available at runtime
 from psycopg2._psycopg import (
@@ -32,6 +33,18 @@ from psycopg2._psycopg import (
     threadsafety as threadsafety,
 )
 
+_T_conn = TypeVar("_T_conn", bound=connection)
+
+@overload
+def connect(dsn: str, connection_factory: Callable[..., _T_conn], cursor_factory: None = ..., **kwargs: Any) -> _T_conn: ...
+@overload
 def connect(
-    dsn: Any | None = ..., connection_factory: Any | None = ..., cursor_factory: Any | None = ..., **kwargs
+    dsn: str | None = ..., *, connection_factory: Callable[..., _T_conn], cursor_factory: None = ..., **kwargs: Any
+) -> _T_conn: ...
+@overload
+def connect(
+    dsn: str | None = ...,
+    connection_factory: Callable[..., connection] | None = ...,
+    cursor_factory: Callable[..., cursor] | None = ...,
+    **kwargs: Any,
 ) -> connection: ...


stubs/psycopg2/psycopg2/__init__.pyi
```diff
@@ -32,6 +33,18 @@ from psycopg2._psycopg import (
...
 )
 
+_T_conn = TypeVar("_T_conn", bound=connection)
+
+@overload
+def connect(dsn: str, connection_factory: Callable[..., _T_conn], cursor_factory: None = ..., **kwargs: Any) -> _T_conn: ...
+@overload
+def connect(
+    dsn: str | None = ..., *, connection_factory: Callable[..., _T_conn], cursor_factory: None = ..., **kwargs: Any
+) -> _T_conn: ...
```
OldSneerJaw, line 43: It seems a bit awkward to annotate the `cursor_factory` param with a type that can only ever be `None` for these two overloads, but because of the `**kwargs` param, it was the most reliable way I could find so that these overloads are matched before the catch-all below. Otherwise, if one were to call, for example, `connect(connection_factory=DictConnection, cursor_factory=NamedTupleCursor)`, the return type would be `DictConnection`, which would be somewhat problematic because the actual instance _would_ be a `DictConnection`, but the instance's `cursor` method would return a `NamedTupleCursor` instance by default, while the `cursor` method's return type would indicate it is a `DictCursor` .

An alternative might be to add a couple more overloads above these two, but I thought that might be more confusing/overwhelming for a user. Here's what that might look like:

```python
@overload
def connect(
    dsn: str | None,
    connection_factory: Callable[..., connection] | None,
    cursor_factory: Callable[..., cursor] | None,
    **kwargs: Any,
) -> connection: ...
@overload
def connect(
    dsn: str | None = ...,
    connection_factory: Callable[..., connection] | None = ...,
    *,
    cursor_factory: Callable[..., cursor] | None,
    **kwargs: Any,
) -> connection: ...
@overload
def connect(dsn: str, connection_factory: Callable[..., _T_conn], **kwargs: Any) -> _T_conn: ...
@overload
def connect(dsn: str | None = ..., *, connection_factory: Callable[..., _T_conn], **kwargs: Any) -> _T_conn: ...
@overload
def connect(
    dsn: str | None = ...,
    connection_factory: Callable[..., connection] | None = ...,
    cursor_factory: Callable[..., cursor] | None = ...,
    **kwargs: Any,
) -> connection: ...
```

Thoughts welcome.

stubs/psycopg2/psycopg2/__init__.pyi
```diff
@@ -32,6 +33,18 @@ from psycopg2._psycopg import (
...
+@overload
 def connect(
-    dsn: Any | None = ..., connection_factory: Any | None = ..., cursor_factory: Any | None = ..., **kwargs
+    dsn: str | None = ...,
+    connection_factory: Callable[..., connection] | None = ...,
+    cursor_factory: Callable[..., cursor] | None = ...,
+    **kwargs: Any,
 ) -> connection: ...
```
OldSneerJaw, line 50: I experimented a bit with giving `connection` a generic type param. For example:

```python
_T_conn_cur = TypeVar("_T_conn_cur", bound=cursor)

class connection(Generic[_T_conn_cur]):
    # ...

    @overload
    def cursor(self, name: str | bytes | None = ..., *, withhold: bool = ..., scrollable: bool | None = ...) -> _T_conn_cur: ...

    # ...
```

That way, in this overload of `connect`, the `cursor_factory` param could have a type of `Callable[..., _T_cur]` and its return type could be changed to `connection[_T_cur]`, but this introduces a few problems. The lack of support for a default value in `TypeVar`s (i.e. so that `connection` and `connection[cursor]` would be equivalent) is a nuisance, but the most serious problem is that the other two `connect` overloads would end up accepting, for example, `connection_factory=connection[NamedTupleCursor]`, which would produce a calculated return type that is inaccurate since the underlying code does not recognize the type param.

JelleZijlstra (approved):
Thanks, this seems like an improvement. We can do better if TypeVar defaults become available.

PR #8573, opened 2022-08-20
Improve `redis.cluster` types

sobolevn:
Source: https://github.com/redis/redis-py/blob/771109e74feb3b8f630c047bf4f93da2800ec22e/redis/cluster.py


stubs/redis/redis/cluster.pyi
```diff
@@ -106,47 +106,49 @@ class RedisCluster(AbstractRedisCluster, RedisClusterCommands[_StrType], Generic
...
 class ClusterNode:
-    host: Any
-    port: Any
-    name: Any
-    server_type: Any
-    redis_connection: Any
-    def __init__(self, host, port, server_type: Any | None = ..., redis_connection: Any | None = ...) -> None: ...
-    def __eq__(self, obj): ...
+    host: str
+    port: int
```
AlexWaygood, line 110: Can `port` also be a `str`? In `__init__()`, it's passed to the `get_node_name()` function, and the `port` parameter of `get_node_name()` is annotated as `Union[str, int]`: https://github.com/redis/redis-py/blob/771109e74feb3b8f630c047bf4f93da2800ec22e/redis/cluster.py#L41

(The same question applies to the `port` parameter of `__init__` immediately below.)

stubs/redis/redis/cluster.pyi
```diff
@@ -106,47 +106,49 @@ class RedisCluster(AbstractRedisCluster, RedisClusterCommands[_StrType], Generic
...
+    name: str
+    server_type: str | None
+    redis_connection: Redis[Any] | None
```
AlexWaygood, line 113:
```suggestion
    redis_connection: Redis[Incomplete] | None
```

stubs/redis/redis/cluster.pyi
```diff
@@ -106,47 +106,49 @@ class RedisCluster(AbstractRedisCluster, RedisClusterCommands[_StrType], Generic
...
+    def __init__(
+        self, host: str, port: int, server_type: str | None = ..., redis_connection: Redis[Any] | None = ...
```
AlexWaygood, line 115:
```suggestion
        self, host: str, port: int, server_type: str | None = ..., redis_connection: Redis[Incomplete] | None = ...
```

stubs/redis/redis/cluster.pyi
```diff
@@ -106,47 +106,49 @@ class RedisCluster(AbstractRedisCluster, RedisClusterCommands[_StrType], Generic
...
+    nodes_cache: dict[str, ClusterNode]
+    slots_cache: dict[str, list[ClusterNode]]
+    startup_nodes: dict[str, ClusterNode]
+    default_node: ClusterNode | None
+    from_url: bool
+    connection_kwargs: dict[str, Any]
+    read_load_balancer: LoadBalancer
     def __init__(
         self,
         startup_nodes,
```
AlexWaygood, line 137:
```suggestion
        startup_nodes: Iterable[ClusterNode],
```

stubs/redis/redis/cluster.pyi
```diff
@@ -106,47 +106,49 @@ class RedisCluster(AbstractRedisCluster, RedisClusterCommands[_StrType], Generic
...
     ) -> None: ...
-    def get_node(self, host: Any | None = ..., port: Any | None = ..., node_name: Any | None = ...): ...
-    def update_moved_exception(self, exception) -> None: ...
-    def get_node_from_slot(self, slot, read_from_replicas: bool = ..., server_type: Any | None = ...): ...
-    def get_nodes_by_server_type(self, server_type): ...
-    def populate_startup_nodes(self, nodes) -> None: ...
-    def check_slots_coverage(self, slots_cache): ...
-    def create_redis_connections(self, nodes) -> None: ...
-    def create_redis_node(self, host, port, **kwargs): ...
+    def get_node(self, host: str | None = ..., port: int | None = ..., node_name: str | None = ...) -> ClusterNode | None: ...
```
AlexWaygood, line 144: Same as above

```suggestion
    def get_node(self, host: str | None = ..., port: int | str | None = ..., node_name: str | None = ...) -> ClusterNode | None: ...
```

stubs/redis/redis/cluster.pyi
```diff
@@ -106,47 +106,49 @@ class RedisCluster(AbstractRedisCluster, RedisClusterCommands[_StrType], Generic
...
+    def update_moved_exception(self, exception: MovedError) -> None: ...
+    def get_node_from_slot(self, slot: str, read_from_replicas: bool = ..., server_type: str | None = ...) -> ClusterNode: ...
+    def get_nodes_by_server_type(self, server_type: str) -> list[ClusterNode]: ...
+    def populate_startup_nodes(self, nodes: list[ClusterNode]) -> None: ...
```
AlexWaygood, line 148:
```suggestion
    def populate_startup_nodes(self, nodes: Iterable[ClusterNode]) -> None: ...
```

stubs/redis/redis/cluster.pyi
```diff
@@ -106,47 +106,49 @@ class RedisCluster(AbstractRedisCluster, RedisClusterCommands[_StrType], Generic
...
+    def check_slots_coverage(self, slots_cache: dict[str, list[ClusterNode]]) -> bool: ...
+    def create_redis_connections(self, nodes: Iterable[ClusterNode]) -> None: ...
+    def create_redis_node(self, host: str, port: int, **kwargs: Any) -> Redis[Any]: ...
```
AlexWaygood, line 151:
```suggestion
    def create_redis_node(self, host: str, port: int | str, **kwargs: Any) -> Redis[Incomplete]: ...
```

AlexWaygood (commented):
Mostly looks good! Just a few queries and suggestions :)

stubs/redis/redis/cluster.pyi
sobolevn: It is `Any`, because it can be `str` or `bytes`, but we don't care about it here.

AlexWaygood: Theoretically, couldn't somebody at some point make `ClusterNode` generic, such that whether it is `Redis[str]` or `Redis[bytes]` could be inferred from the type passed into the `redis_connection` parameter of `ClusterNode.__init__`?

stubs/redis/redis/cluster.pyi
sobolevn: I think that it can not, because `node.port` is used in a couple of other places, for example: creating a `Redis(host, port)` instance, which is annotated to have `port` as `int`.

Source: https://github.com/redis/redis-py/blob/771109e74feb3b8f630c047bf4f93da2800ec22e/redis/cluster.py#L1370-L1378
Stub: https://github.com/python/typeshed/blob/c4ca110c5c02699350ca008050c430e23e087b93/stubs/redis/redis/client.pyi#L88

stubs/redis/redis/cluster.pyi
sobolevn: Here `post` can be `str`, because it is only used for string formatting (but, I personally think that it should not be used that away).

stubs/redis/redis/cluster.pyi
sobolevn: I am not sure about it 🤔 
I don't know how redis cluster behaves in terms of encoding.

stubs/redis/redis/cluster.pyi
AlexWaygood: Hmm... I think maybe our stub in `redis/client.pyi` should also be `port: int | str | None`. `redis.client.Redis` passes `port` to the constructor of `redis.connection.ConnectionPool`, which then passes port to the constructor of `redis.connection.Connection`. And `redis.connection.Connection` accepts `str` for the `port` parameter: https://github.com/redis/redis-py/blob/771109e74feb3b8f630c047bf4f93da2800ec22e/redis/connection.py#L539

stubs/redis/redis/cluster.pyi
AlexWaygood: If we don't know, then I'd prefer for us to go with `Incomplete` rather than `Any` ;) I think we should use `Any` only when we know for sure that there's no other viable alternative :)

stubs/redis/redis/cluster.pyi
sobolevn: Let's leave this for the next PR 😉

sobolevn: I will investigate where we can and cannot do it.

AlexWaygood: Sounds good!

AlexWaygood (approved):
Thanks! :D

PR #8579, opened 2022-08-20
Disallow adding stubs for types- packages

hauntsaninja:
This would lead to confusion. The types prefix is special cased in
various ways in stub_uploader

### DIFF
--- a/tests/check_consistent.py
+++ b/tests/check_consistent.py
@@ -46,6 +46,7 @@ def check_stdlib() -> None:
 
 def check_stubs() -> None:
     for distribution in os.listdir("stubs"):
+        assert not distribution.startswith("types-"), f"Distribution not allowed to start with 'types-': {distribution}"
         assert not os.path.isfile(distribution), f"Only directories allowed in stubs, got {distribution}"
         for entry in os.listdir(os.path.join("stubs", distribution)):
             if os.path.isfile(os.path.join("stubs", distribution, entry)):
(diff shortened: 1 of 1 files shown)


AlexWaygood (approved):

tests/check_consistent.py
```diff
@@ -46,6 +46,7 @@ def check_stdlib() -> None:
 
 def check_stubs() -> None:
     for distribution in os.listdir("stubs"):
+        assert not distribution.startswith("types-"), f"Distribution not allowed to start with 'types-': {distribution}"
         assert not os.path.isfile(distribution), f"Only directories allowed in stubs, got {distribution}"
```
hauntsaninja, line 50: Hmm, this check looks buggy. It should be checking `os.path.isfile(os.path.join("stubs", distribution))`. I'll fix in another PR.

tests/check_consistent.py
```diff
@@ -46,6 +46,7 @@ def check_stdlib() -> None:
 
 def check_stubs() -> None:
     for distribution in os.listdir("stubs"):
+        assert not distribution.startswith("types-"), f"Distribution not allowed to start with 'types-': {distribution}"
```
AlexWaygood, line 49:
```suggestion
        assert not distribution.startswith("types-"), f"Directory name cannot start with 'types-': {distribution}"
```

hauntsaninja: The directories in stubs represent distributions and the reason we disallow this is because a distribution named types-something would be confusing (for humans and for our code). So I'd prefer to keep the word "distribution" in the message.

hauntsaninja: I'm going to merge this so I can fix the bug below, but if you still have a viewpoint on the error message I'll change it in the next PR :-)

AlexWaygood: I feel like if anybody runs into this error, it's likely to be a new contributor, who might not understand what "distribution" means in this context. But I don't have a strong opinion :-)

PR #8581, opened 2022-08-20
Improve check_consistent

hauntsaninja:
- Fixes bug in checking directories in stubs/
- Use pathlib for convenience
- Share more logic between check_stubs and check_stdlib
- Better recursive checking, e.g. if a README.md is in the wrong place
- Test distribution names are valid
- Errors now contain full paths
- I believe pathlib normalises separators, but someone with Windows
  might want to double check this continues to work

### DIFF
--- a/tests/check_consistent.py
+++ b/tests/check_consistent.py
@@ -25,6 +25,7 @@ def assert_stubs_only(directory: Path, allowed: set[str]) -> None:
     while contents:
         entry = contents.pop()
         if entry.relative_to(directory) in allowed_paths:
+            # Note if a subdirectory is allowed, we will not check its contents
             continue
         if entry.is_file():
             assert entry.stem.isidentifier(), f"Files must be valid modules, got: {entry}"
@@ -46,7 +47,7 @@ def check_stubs() -> None:
         assert re.fullmatch(
             valid_dist_name, dist.name, re.IGNORECASE
         ), f"Directory name must have valid distribution name: {dist}"
-        assert not dist.name.startswith("types-"), f"Directory not allowed to start with 'types-': {dist}"
+        assert not dist.name.startswith("types-"), f"Directory name not allowed to start with 'types-': {dist}"
 
         allowed = {"METADATA.toml", "README", "README.md", "README.rst", "@tests"}
         assert_stubs_only(dist, allowed)
--- a/tests/check_consistent.py
+++ b/tests/check_consistent.py
@@ -46,7 +46,7 @@ def check_stubs() -> None:
         assert re.fullmatch(
             valid_dist_name, dist.name, re.IGNORECASE
         ), f"Directory name must have valid distribution name: {dist}"
-        assert not dist.name.startswith("types-"), f"Distribution not allowed to start with 'types-': {dist}"
+        assert not dist.name.startswith("types-"), f"Directory not allowed to start with 'types-': {dist}"
 
         allowed = {"METADATA.toml", "README", "README.md", "README.rst", "@tests"}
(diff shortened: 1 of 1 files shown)


JelleZijlstra (approved):

tests/check_consistent.py
```diff
@@ -8,57 +8,48 @@
...
-                    continue
-                assert_stubs_only(os.path.join("stubs", distribution, entry))
+    for dist in Path("stubs").iterdir():
+        assert dist.is_dir(), f"Only directories allowed in stubs, got {dist}"
+
+        valid_dist_name = "^([A-Z0-9]|[A-Z0-9][A-Z0-9._-]*[A-Z0-9])$"  # courtesy of PEP 426
+        assert re.fullmatch(
+            valid_dist_name, dist.name, re.IGNORECASE
+        ), f"Directory name must have valid distribution name: {dist}"
+        assert not dist.name.startswith("types-"), f"Directory not allowed to start with 'types-': {dist}"
```
AlexWaygood, line 49:
```suggestion
        assert not dist.name.startswith("types-"), f"Directory name not allowed to start with 'types-': {dist}"
```

AlexWaygood (approved):
I manually tested it all out on my Windows machine -- everything that previously emitted an error still does so, I think. Looks good!

One optional suggestion :-)

tests/check_consistent.py
```diff
@@ -8,57 +8,48 @@
...
-                else:
-                    assert name.isidentifier(), f"Bad file name '{entry}' in stubs"
-            else:
-                if entry == "@tests":
-                    continue
-                assert_stubs_only(os.path.join("stubs", distribution, entry))
+    for dist in Path("stubs").iterdir():
+        assert dist.is_dir(), f"Only directories allowed in stubs, got {dist}"
+
+        valid_dist_name = "^([A-Z0-9]|[A-Z0-9][A-Z0-9._-]*[A-Z0-9])$"  # courtesy of PEP 426
```
Akuli, line 45: Nit: `^` and `$` are unnecessary because we use `re.fullmatch`.

tests/check_consistent.py
```diff
@@ -8,57 +8,48 @@
...
-            if file in allowed_files:
-                continue
-            name, ext = os.path.splitext(file)
-            assert name.isidentifier(), f"Files must be valid modules, got: {name}"
-            assert ext == ".pyi", f"Only stub flies allowed. Got: {file} in {directory}"
-        for subdir in dirs:
-            assert subdir.isidentifier(), f"Directories must be valid packages, got: {subdir}"
+    allowed_paths = {Path(f) for f in allowed}
+    contents = list(directory.iterdir())
+    while contents:
```
Akuli, line 25: It would be simpler to for loop over `directory.rglob("*")`.

tests/check_consistent.py
Akuli: I like how this allows single-character package names. `pip install 7` lol

tests/check_consistent.py
hauntsaninja: That doesn't quite work, because currently if a subdirectory (e.g. `@tests`) is allowed, we won't check its contents

Akuli: We could do `if entry in allowed_paths or any(parent in allowed_paths for parent in entry.parents)`. A bit of a mess and recurses unnecessarily, but maybe more explicit. I'll let you decide :)

Akuli: Lol, you need to do the whole `relative_to()` thing for that check. It would become a bit complicated, maybe not worth the effort :)

PR #8584, opened 2022-08-21
`cgi.FieldStorage` headers accepts `Message`s

### LINKED ISSUE #8559: `Argument "headers" to "FieldStorage" has incompatible type `
**Bug Report**

Hi,
consider the following examples:

```python
import http.server
from cgi import FieldStorage
from http import HTTPStatus

class Handler(http.server.SimpleHTTPRequestHandler):
    def do_POST(self) -> None:
        if self.path == "/upload":
            result = self.receive_upload()
    def receive_upload(self) -> tuple[HTTPStatus, str]:
        result = (HTTPStatus.INTERNAL_SERVER_ERROR, "Server error")
        form = FieldStorage(fp=self.rfile, headers=self.headers, environ={"REQUEST_METHOD": "POST"})
        return result
```

I get:

```console
main.py:11: error: Argument "headers" to "FieldStorage" has incompatible type "Message"; expected "Optional[Mapping[str, str]]"
```

**Your Environment**

Run on <https://mypy-play.net/?mypy=latest&python=3.10>

Am I missing something?

Thanks

DMRobertson:
Fixes #8559.

(Draft for now to see how stubtest and other CI works)

### DIFF
--- a/stdlib/cgi.pyi
+++ b/stdlib/cgi.pyi
@@ -1,9 +1,8 @@
-from email.message import Message
-
 import sys
 from _typeshed import Self, SupportsGetItem, SupportsItemAccess
 from builtins import list as _list, type as _type
 from collections.abc import Iterable, Iterator, Mapping
+from email.message import Message
 from types import TracebackType
 from typing import IO, Any, Protocol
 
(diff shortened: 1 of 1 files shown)


AlexWaygood (approved):
Thank you!

PR #8586, opened 2022-08-21
Markdown BlockProcessor.detab returns a pair of strings

oprypin:
See https://github.com/Python-Markdown/markdown/blob/eff6c3a0415cb63bc0b38c29b5a7f584619e3c1c/markdown/blockprocessors.py#L94


AlexWaygood (approved):
Thanks!

Issue #8139, opened 2022-06-22
mypy_test: Are `exclude` and `filter` useful?


AlexWaygood:
`mypy_test.py` has two command-line arguments that I find somewhat useless: `filter` and `exclude`. (It looks like they've stayed more-or-less the same since the script's inception 7 years ago.) I'd like to get rid of them and replace them with flags that would allow me to run specific parts of mypy_test locally, instead of running the whole thing at once (e.g. `--stdlib` to only test the stdlib, or `--test-cases` to only test the test cases).

Any objections to that?

AlexWaygood:
Alternatively, we could have a single command-line argument, `--dir`. Running only on the stdlib would be `--dir stdlib`, etc.

AlexWaygood:
I'm actually somewhat inclined to remove the `--verbose` and `--dry-run` options as well. I've never found them useful (but I'm happy to keep them if somebody else has).

srittau:
I'm fine with changing `exclude` and `filter`. The approach I find most intuitive was to just run `mypy_test stdlib/foo.pyi` or `mypy_test stubs/bar` to test a specific file or directory.

`--verbose` seems useful in case you need the full mypy command line for some reason. `--dry-run` on the other hand seems mostly useful when developing mypy_test, so I don't think it's very useful in general.

Unrelated: I would prefer not to test our own scripts in mypy_test. Instead we should have a separate step in CI and just manually call mypy. In my view, testing stubs and testing our support script are two unrelated activities.

AlexWaygood:
> Unrelated: I would prefer not to test our own scripts in mypy_test. Instead we should have a separate step in CI and just manually call mypy. In my view, testing stubs and testing our support script are two unrelated activities.

This makes sense, and @hauntsaninja said something similar to me at PyCon.

AlexWaygood:
> I'm fine with changing `exclude` and `filter`. The approach I find most intuitive was to just run `mypy_test stdlib/foo.pyi` or `mypy_test stubs/bar` to test a specific file or directory.

I generally just invoke mypy directly if I want to test just a specific file or subdirectory. I don't think it makes sense to try to reproduce all of the mypy command-line options in `mypy_test`, so I'd prefer to keep things simple and just remove these options, personally.

PR #8587, opened 2022-08-21
`mypy_test.py`: Move type-checking of our tests and scripts into a different test

AlexWaygood:
As requested in https://github.com/python/typeshed/issues/8139#issuecomment-1166923613

### DIFF
--- a/.github/workflows/mypy_selfcheck.yml
+++ b/.github/workflows/mypy_selfcheck.yml
@@ -33,4 +33,4 @@ jobs:
         with:
           python-version: 3.x
       - run: pip install -r requirements-tests.txt
-      - run: ./tests/test_the_tests.py --platform=${{ matrix.platform }} --python-version=${{ matrix.python-version }}
+      - run: python ./tests/test_the_tests.py --platform=${{ matrix.platform }} --python-version=${{ matrix.python-version }}
--- a/.github/workflows/mypy_selfcheck.yml
+++ b/.github/workflows/mypy_selfcheck.yml
@@ -31,6 +31,6 @@ jobs:
       - uses: actions/checkout@v3
       - uses: actions/setup-python@v4
         with:
-          python-version: ${{ matrix.python-version }}
+          python-version: 3.x
       - run: pip install -r requirements-tests.txt
-      - run: ./tests/test_the_tests.py --platform=${{ matrix.platform }}
+      - run: ./tests/test_the_tests.py --platform=${{ matrix.platform }} --python-version=${{ matrix.python-version }}
--- a/.github/workflows/mypy_selfcheck.yml
+++ b/.github/workflows/mypy_selfcheck.yml
@@ -8,8 +8,9 @@ on:
       - master
   pull_request:
     paths:
-      - scripts
-      - tests
+      - 'scripts/**'
+      - 'tests/**'
+      - '.github/workflows/mypy_primer_comment.yml'
 
 permissions:
   contents: read
--- a/.github/workflows/mypy_selfcheck.yml
+++ b/.github/workflows/mypy_selfcheck.yml
@@ -1,0 +1,35 @@
+name: Test-the-tests
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - main
+      - master
+  pull_request:
+    paths:
+      - scripts
+      - tests
+
+permissions:
+  contents: read
+
+env:
+  PIP_DISABLE_PIP_VERSION_CHECK: 1
+
+jobs:
+  mypy:
+    name: Run mypy against the scripts and tests directories
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        platform: ["linux", "win32", "darwin"]
+        python-version: ["3.9", "3.10", "3.11"]
+      fail-fast: false
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-python@v4
+        with:
+          python-version: ${{ matrix.python-version }}
+      - run: pip install -r requirements-tests.txt
+      - run: ./tests/test_the_tests.py --platform=${{ matrix.platform }}
--- a/.github/workflows/mypy_selfcheck.yml
+++ b/.github/workflows/mypy_selfcheck.yml
@@ -10,7 +10,7 @@ on:
     paths:
       - 'scripts/**'
       - 'tests/**'
-      - '.github/workflows/mypy_primer_comment.yml'
+      - '.github/workflows/mypy_selfcheck.yml'
 
 permissions:
   contents: read
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -61,7 +61,7 @@ jobs:
       - run: ./tests/pytype_test.py --print-stderr
 
   mypy:
-    name: Run mypy against typeshed
+    name: Test the stubs with mypy
     runs-on: ubuntu-latest
     strategy:
       matrix:
@@ -77,7 +77,7 @@ jobs:
       - run: ./tests/mypy_test.py --platform=${{ matrix.platform }} --python-version=${{ matrix.python-version }}
 
   pyright:
-    name: Run pyright against typeshed
+    name: Test the stubs with pyright
     runs-on: ubuntu-latest
     strategy:
       matrix:
--- a/.github/workflows/typecheck_typeshed_code.yml
+++ b/.github/workflows/typecheck_typeshed_code.yml
@@ -1,4 +1,4 @@
-name: Test-the-tests
+name: Typecheck-typeshed-code
 
 on:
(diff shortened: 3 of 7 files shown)


.github/workflows/mypy_selfcheck.yml
```diff
@@ -0,0 +1,35 @@
...
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - main
+      - master
+  pull_request:
+    paths:
+      - scripts
+      - tests
```
AlexWaygood, line 12:
```suggestion
      - 'scripts/**'
      - 'tests/**'
      - '.github/workflows/mypy_primer_comment.yml'
```

.github/workflows/mypy_selfcheck.yml
```diff
@@ -0,0 +1,36 @@
...
+  workflow_dispatch:
+  push:
+    branches:
+      - main
+      - master
+  pull_request:
+    paths:
+      - 'scripts/**'
+      - 'tests/**'
+      - '.github/workflows/mypy_primer_comment.yml'
```
AlexWaygood, line 13:
```suggestion
      - '.github/workflows/mypy_selfcheck.yml'
```

.github/workflows/mypy_selfcheck.yml
```diff
@@ -0,0 +1,36 @@
...
+        platform: ["linux", "win32", "darwin"]
+        python-version: ["3.9", "3.10", "3.11"]
+      fail-fast: false
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-python@v4
+        with:
+          python-version: ${{ matrix.python-version }}
+      - run: pip install -r requirements-tests.txt
+      - run: ./tests/test_the_tests.py --platform=${{ matrix.platform }}
```
AlexWaygood, line 36:
```suggestion
          python-version: 3.x
      - run: pip install -r requirements-tests.txt
      - run: ./tests/test_the_tests.py --platform=${{ matrix.platform }} --python-version=${{ matrix.python-version }}
```

.github/workflows/mypy_selfcheck.yml
```diff
@@ -0,0 +1,36 @@
...
+        platform: ["linux", "win32", "darwin"]
+        python-version: ["3.9", "3.10", "3.11"]
+      fail-fast: false
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-python@v4
+        with:
+          python-version: 3.x
+      - run: pip install -r requirements-tests.txt
+      - run: ./tests/test_the_tests.py --platform=${{ matrix.platform }} --python-version=${{ matrix.python-version }}
```
AlexWaygood, line 36:
```suggestion
      - run: python ./tests/test_the_tests.py --platform=${{ matrix.platform }} --python-version=${{ matrix.python-version }}
```

.github/workflows/tests.yml
```diff
@@ -61,7 +61,7 @@ jobs:
       - run: ./tests/pytype_test.py --print-stderr
 
   mypy:
-    name: Run mypy against typeshed
+    name: Test the stubs with mypy
```
hauntsaninja, line 64: ignorable nit: I prefer "Type check the stubs with mypy" (keeping with my comment about all the use of the word "test")

tests/test_the_tests.py
```diff
@@ -0,0 +1,102 @@
...
+        "-m",
+        "mypy",
+        directory,
+        "--platform",
+        platform,
+        "--python-version",
+        version,
+        "--strict",
+        "--show-traceback",
+        "--warn-incomplete-stub",
```
hauntsaninja, line 57: Don't need `--warn-incomplete-stub`, right?

hauntsaninja (commented):
Thanks for cleaning this up.

I have some minor feedback:
- Calling a workflow `mypy_selfcheck` is very confusing, especially given that typeshed used to have a test called mypy self check where we'd run mypy over mypy using custom typeshed.
- It feels like there are a number of uses of the word "test" floating around. Maybe "typecheck-typeshed-code" is clearer than "test-the-tests" and better covers the fact that we type check scripts?
- This isn't worth 9 CI jobs; the multiple Python versions is overkill. Let's just do 3.9. (Tbh, multiple platforms is also overkill, maybe just make it `--platform win32` and call it a day)

tests/test_the_tests.py
AlexWaygood: Hah, good point

AlexWaygood:
> * Calling a workflow `mypy_selfcheck` is very confusing

Yeah, I had no idea what to call this workflow :)

tests/test_the_tests.py
```diff
@@ -0,0 +1,102 @@
...
+        "--show-error-codes",
+        "--no-error-summary",
+        "--enable-error-code",
+        "ignore-without-code",
+    ]
+    if directory == "tests":
+        command.append("--namespace-packages")
```
hauntsaninja, line 64: Oh let's unconditionally use `--namespace-packages`

.github/workflows/tests.yml
AlexWaygood: Hmm, this one I will ignore, since we still run mypy against the `test_cases` directory in `mypy_test.py`, and that feels more like a "test" than a "type check" :)

tests/test_the_tests.py
AlexWaygood: Sure

.github/workflows/typecheck_typeshed_code.yml
```diff
@@ -0,0 +1,35 @@
...
+  workflow_dispatch:
+  push:
+    branches:
+      - main
+      - master
+  pull_request:
+    paths:
+      - 'scripts/**'
+      - 'tests/**'
+      - '.github/workflows/mypy_selfcheck.yml'
```
AlexWaygood, line 13:
```suggestion
      - '.github/workflows/typecheck_typeshed_code.yml'
```

.github/workflows/typecheck_typeshed_code.yml
```diff
@@ -0,0 +1,35 @@
...
+      matrix:
+        platform: ["linux", "win32"]
+      fail-fast: false
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-python@v4
+        with:
+          python-version: 3.9
+      - run: pip install -r requirements-tests.txt
+      - run: python ./tests/test_the_tests.py --platform=${{ matrix.platform }}
```
AlexWaygood, line 35:
```suggestion
      - run: python ./tests/typecheck_typeshed.py --platform=${{ matrix.platform }}
```

.github/workflows/typecheck_typeshed_code.yml
```diff
@@ -0,0 +1,35 @@
...
+  workflow_dispatch:
+  push:
+    branches:
+      - main
+      - master
+  pull_request:
+    paths:
+      - 'scripts/**'
+      - 'tests/**'
+      - '.github/workflows/typecheck_typeshed_code.yml'
```
AlexWaygood, line 13:
```suggestion
      - '.github/workflows/typecheck_typeshed_code.yml'
      - 'requirements-tests.txt'
```

hauntsaninja (approved):

AlexWaygood:
Thanks for the review! :)

I'll leave this open for a little bit to see if anybody else has any thoughts.

tests/typecheck_typeshed.py
```diff
@@ -0,0 +1,102 @@
...
+        "--no-error-summary",
+        "--enable-error-code",
+        "ignore-without-code",
+        "--namespace-packages",
+    ]
+    if directory == "tests":
+        if platform == "win32":
+            command.extend(["--exclude", "tests/pytype_test.py"])
+    else:
+        command.append("--ignore-missing-imports")
```
srittau, line 67: Side note: We should probably add the types for the imports we use to tests-requirements. (In a different PR.)

srittau (approved):

tests/typecheck_typeshed.py
AlexWaygood: Or maybe `script-requirements`/`stubsabot-requirements`? (Most contributors aren't going to need to run stubsabot manually, after all :)

srittau: I think it's easier to just put all the requirements into one file. It won't hurt much to install these extra requirements, but it's a bit lower of a mental burden for when you need those.

AlexWaygood: Looking at it again, it seems as though stubsabot's additional requirements are pretty minimal anyway, so it probably doesn't hurt -- you're right!

AlexWaygood:
I'll merge this now, but happy to address any further feedback anybody has in a followup PR!

PR #8588, opened 2022-08-21
Add types to tkinter.colorchooser.askcolor

Akuli:
It calls `tk_chooseColor`, because `Chooser` sets `command` to `"tk_chooserColor"` and `Dialog` calls whatever `command` is set to. [Manual page of tk_chooseColor is here.](https://www.tcl.tk/man/tcl8.7/TkCmd/chooseColor.html)


TeamSpen210:
Something else with `askcolor`: As [of 3.9](https://github.com/python/cpython/pull/6578), the return value is actually `tuple[int, int, int]`.

Akuli:
Thanks for pointing that out! I have updated the return type.

srittau (approved):
I didn't double check (because tkinter), but that looks reasonable.

PR #8589, opened 2022-08-22
Fix session type

kasium:
scoped_session is not a subtype of Session, therefore the correct one is scoped_session


AlexWaygood (approved):
Thanks for the fix!

PR #8590, opened 2022-08-22
refactor: expand types in xml module

kkirsche:
1. `encoding` changed from `Any | None` to `str | None` as it's used as a parameter to TextIOWrapper
    * https://github.com/python/cpython/blob/main/Lib/xml/dom/minidom.py#L54
2. `standalone` changed from `Any | None` to `bool | None` per https://docs.python.org/3/library/xml.dom.minidom.html#xml.dom.minidom.Node.writexml
3. `localName` changed from `Any | None` to `str | None` as `Node`'s `localName` property returns `str | None`
4. Comparison operators set to `object` as they use `id` to do the comparisons, not object specific behavior: https://github.com/python/cpython/blob/main/Lib/xml/dom/minidom.py#L534
5. `TypeInfo` partially typed based on https://cs.github.com/python/cpython/blob/5ed584cb6b4e544d307f2f1b6f28667078cc20af/Lib/xml/dom/expatbuilder.py?q=repo%3Apython%2Fcpython+path%3ALib%2Fxml%2F+TypeInfo#L47-L58
6. `get*Item` check with a try / except: https://github.com/python/cpython/blob/main/Lib/xml/dom/minidom.py#L571
7. Various create methods include isinstance type checking to force behaviors: https://github.com/python/cpython/blob/main/Lib/xml/dom/minidom.py#L1676


kkirsche:
Also, some of the parameter names of node are misleading, as the isinstance checks are checking for Attr nodes specifically, which is deceptive on a first pass:
https://github.com/python/cpython/blob/main/Lib/xml/dom/minidom.py#L607-L608

stdlib/xml/dom/minidom.pyi
```diff
@@ -87,31 +90,31 @@ class NamedNodeMap:
...
-    def setNamedItemNS(self, node): ...
-    def __delitem__(self, attname_or_tuple) -> None: ...
+    def __setitem__(self, attname: str, value: Attr | str) -> None: ...
+    def getNamedItem(self, name: str) -> Attr | None: ...
+    def getNamedItemNS(self, namespaceURI: str, localName: str | None) -> Attr | None: ...
+    def removeNamedItem(self, name: str) -> Attr: ...
+    def removeNamedItemNS(self, namespaceURI: str, localName: str | None): ...
+    def setNamedItem(self, node: Attr) -> Attr: ...
+    def setNamedItemNS(self, node: Attr) -> Attr: ...
+    def __delitem__(self, attname_or_tuple: str) -> None: ...
```
JelleZijlstra, line 108: This can also be a tuple, more precisely a `tuple[str, str | None]`. `__getitem__` takes the same type.

stdlib/xml/dom/minidom.pyi
```diff
@@ -290,47 +293,47 @@ class Document(Node, DocumentLS):
...
-    def createDocumentFragment(self): ...
-    def createElement(self, tagName: str): ...
-    def createTextNode(self, data): ...
-    def createCDATASection(self, data): ...
-    def createComment(self, data): ...
+    def createDocumentFragment(self) -> DocumentFragment: ...
+    def createElement(self, tagName: str) -> Element: ...
+    def createTextNode(self, data: str) -> Text: ...
+    def createCDATASection(self, data: str) -> CDATASection: ...
+    def createComment(self, data: Incomplete) -> Comment: ...
```
JelleZijlstra, line 314: Should be `str`. It gets passed to `Comment` and `Comment.writexml` assumes its data is a str: https://github.com/python/cpython/blob/18b1782192f85bd26db89f5bc850f8bee4247c1a/Lib/xml/dom/minidom.py#L1208

JelleZijlstra (changes requested):
Thanks, I have a few comments.

stdlib/xml/dom/minidom.pyi
```diff
@@ -144,7 +147,7 @@ class Element(Node):
     def hasAttributeNS(self, namespaceURI: str, localName) -> bool: ...
     def getElementsByTagName(self, name: str): ...
     def getElementsByTagNameNS(self, namespaceURI: str, localName): ...
-    def writexml(self, writer, indent: str = ..., addindent: str = ..., newl: str = ...) -> None: ...
+    def writexml(self, writer: _Writer, indent: str = ..., addindent: str = ..., newl: str = ...) -> None: ...
```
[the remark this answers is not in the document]
kkirsche: Makes sense, the lack of them being in the typing_extensions as a user-facing protocol I guess is why I assumed the opposite (as I assume users would need the same protocols as typeshed needs in many cases). I'd be curious to know if there's interest in moving some of these types of protocols to be user-facing.

stdlib/xml/dom/minidom.pyi
kkirsche: Sorry about that, I had mixed up which line that was accurate for as one location they're used as arguments to `_attrs` I believe and the other they're used as method arguments

stdlib/xml/dom/minidom.pyi
```diff
@@ -87,31 +87,31 @@ class NamedNodeMap:
...
-    def getNamedItemNS(self, namespaceURI: str, localName): ...
-    def removeNamedItem(self, name): ...
-    def removeNamedItemNS(self, namespaceURI: str, localName): ...
-    def setNamedItem(self, node): ...
-    def setNamedItemNS(self, node): ...
-    def __delitem__(self, attname_or_tuple) -> None: ...
+    def __ge__(self, other: object) -> bool: ...
+    def __gt__(self, other: object) -> bool: ...
+    def __le__(self, other: object) -> bool: ...
+    def __lt__(self, other: object) -> bool: ...
```
AlexWaygood, line 96: This smells a bit off to me. It's true that you _can_ write bizarre code like this, and no exception will be raised at runtime:

```pycon
>>> from xml.dom.minidom import NamedNodeMap
>>> NamedNodeMap('foo', 'bar', 'baz') < object()
False
```

But... that doesn't mean that you necessarily _should_. Code like that just doesn't make much sense.

I feel like `NamedNodeMap`s are _probably_ only meant to be compared to other `NamedNodeMap`s?

stdlib/xml/dom/minidom.pyi
```diff
@@ -70,7 +73,7 @@ class Attr(Node):
     value: str
     prefix: Any
     def __init__(
-        self, qName: str, namespaceURI: str | None = ..., localName: Any | None = ..., prefix: Any | None = ...
+        self, qName: str, namespaceURI: str | None = ..., localName: str | None = ..., prefix: Any | None = ...
```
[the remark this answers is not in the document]
kkirsche: Sure, I'm happy to give that a shot.

Issue:
https://github.com/python/cpython/issues/96175

PR:
https://github.com/python/cpython/pull/96176

JelleZijlstra (approved):
My comments were addressed, thanks. Still need to resolve @AlexWaygood's concern.

stdlib/xml/dom/minidom.pyi
kkirsche: I assumed that, but when I looked at https://docs.python.org/3/library/functions.html#id it mentioned that `id` I saw the CPython implementation detail that it's returning the memory location. As a result, I wasn't sure if it was intended to rely on the implementation details of `id` as a shortcut to inferring details about parsing progression. I'll restrict it though as requested

stdlib/xml/dom/minidom.pyi
```diff
@@ -87,31 +87,31 @@ class NamedNodeMap:
...
-    def __getitem__(self, attname_or_tuple): ...
-    def __setitem__(self, attname, value) -> None: ...
-    def getNamedItem(self, name): ...
-    def getNamedItemNS(self, namespaceURI: str, localName): ...
-    def removeNamedItem(self, name): ...
-    def removeNamedItemNS(self, namespaceURI: str, localName): ...
-    def setNamedItem(self, node): ...
-    def setNamedItemNS(self, node): ...
-    def __delitem__(self, attname_or_tuple) -> None: ...
+    def __eq__(self, other: NamedNodeMap) -> bool: ...
```
AlexWaygood, line 92: `__eq__` methods are a bit special: while it doesn't really make sense for a user to do a test like `if NamedNodeMap('foo', 'bar', 'baz') < object()` (so we'll be doing them a favour by being more restrictive in the type hints than the implementation is), it's perfectly fine for somebody to do a test like `if NamedNodeMap('foo', 'bar', 'baz') == object()`. This one should probably stay as it is:

```suggestion
    def __eq__(self, other: object) -> bool: ...
```

AlexWaygood (approved):
I haven't reviewed in detail, but my concerns have been addressed, and I'll defer to @JelleZijlstra on the rest! :)

PR #8594, opened 2022-08-22
Great cleanup!

sobolevn:
I've searched for duplicate definitions and removed most of them.
This was a manual job with some help of a new `stubtest`.

Rules:
- I haven't touched *any* magic methods (except a couple of identical `__init__`)
- I haven't touched any code with comments
- I haven't touched untyped code
- I haven't touched overloads, because they are too hard to read and verify
- I haven't touched any other advanced stuff


stdlib/xml/dom/minidom.pyi
```diff
@@ -52,8 +52,6 @@ class DocumentFragment(Node):
     nodeName: str
     nodeValue: Incomplete
     attributes: Incomplete
-    parentNode: Incomplete
-    childNodes: Incomplete
```
AlexWaygood, line 56: Can we maybe also leave the ones that are explicitly marked as `Incomplete` as they are? Since we've explicitly marked these as "we haven't figured out what they're meant to be yet", they might plausibly have different types in the subclass than in the superclass at runtime.

sobolevn: I want to cancel all changes in this file :)
Because after my merge conflict I stopped understanding what is going on.

sobolevn:
This error looks like a `stubtest` bug to me:

```
error: io.BufferedRandom.seek is inconsistent, stub argument "__offset" differs from runtime argument "target"
Stub: at line 65
def (self: io.IOBase, builtins.int, builtins.int =) -> builtins.int
Runtime:
def (self, target, whence=0, /)
```

Both types have pos-only parameters, but for some reason they are reported as incompatible.
I will restore these two with a comment for now.

AlexWaygood:
> This error looks like a `stubtest` bug to me:

I believe it's actually stubtest working as intended; see https://github.com/python/mypy/issues/13113. I can see both sides of the argument here :-)

stdlib/importlib/abc.pyi
```diff
@@ -36,24 +36,29 @@ _Path: TypeAlias = bytes | str
 
 class Finder(metaclass=ABCMeta): ...
 
+class Loader(metaclass=ABCMeta):
+    def load_module(self, fullname: str) -> types.ModuleType: ...
+    def module_repr(self, module: types.ModuleType) -> str: ...
+    def create_module(self, spec: ModuleSpec) -> types.ModuleType | None: ...
+    # Not defined on the actual class for backwards-compatibility reasons,
+    # but expected in new code.
+    def exec_module(self, module: types.ModuleType) -> None: ...
```
AlexWaygood, line 45: Given the comment here (this one doesn't actually exist on the class at runtime), maybe we should leave `exec_module` there in subclasses of `Loader`?

sobolevn: I don't really know the machinery in `importlib`, so I am fine with both.

AlexWaygood: Neither do I, so I'd prefer to play it safe and leave it in for now :)

stdlib/importlib/abc.pyi
```diff
@@ -36,24 +36,30 @@ _Path: TypeAlias = bytes | str
...
 class InspectLoader(Loader):
     def is_package(self, fullname: str) -> bool: ...
     def get_code(self, fullname: str) -> types.CodeType | None: ...
-    def load_module(self, fullname: str) -> types.ModuleType: ...
     @abstractmethod
     def get_source(self, fullname: str) -> str | None: ...
-    def exec_module(self, module: types.ModuleType) -> None: ...
     @staticmethod
     def source_to_code(data: bytes | str, path: str = ...) -> types.CodeType: ...
+    def exec_module(self, module: types.ModuleType) -> None: ...
```
AlexWaygood, line 58: Let's move this back to where it was before, to avoid noise in `git blame`

stdlib/importlib/machinery.pyi
```diff
@@ -144,7 +144,6 @@ class ExtensionFileLoader(importlib.abc.ExecutionLoader):
     def get_filename(self, name: str | None = ...) -> importlib.abc._Path: ...
     def get_source(self, fullname: str) -> None: ...
     def create_module(self, spec: ModuleSpec) -> types.ModuleType: ...
-    def exec_module(self, module: types.ModuleType) -> None: ...
-    def is_package(self, fullname: str) -> bool: ...
     def get_code(self, fullname: str) -> None: ...
+    def exec_module(self, module: types.ModuleType) -> None: ...
```
AlexWaygood, line 148: Same here

AlexWaygood (approved):
I haven't rigorously checked the whole thing to see if the signatures on subclasses are identical to the ones on superclasses, but I trust you to have done it properly :)

LGTM, though I'll wait to see if any other maintainers have any thoughts before merging.

sobolevn:
> I trust you to have done it properly

I trust `stubtest` more than myself here 😆

sobolevn:
Any other feedback? :)

I think that this PR can quickly have merge conflicts and this will require some time to re-think.

srittau (approved):
Not checked this in depth, either, but the passing tests give me confidence.

Issue #7813, opened 2022-05-09
Many functions that accept an Iterable should actually accept Iterable SupportsGetItem


matangover:
Many functions in stdlib that are documented to accept an "iterable" are annotated as `Iterable[T]`. This means the arguments must have `__iter__`. However, the actual functions don't require the arguments to have `__iter__`, they also work with arguments that have only `__getitem__`. That's because internally they use [PySequence_Fast](https://docs.python.org/3/c-api/sequence.html#c.PySequence_Fast)

Example functions that have this bug right now:
```
enumerate
str.join
```

For enumerate, a bug was filed [on the mypy repo](https://github.com/python/mypy/issues/9737) but in fact I believe this should be corrected in typeshed, so that other type checkers will have it right too.

Aside: `for` loops also similarly accept 'iterables' that have only __getitem__, but that is implemented inside type checkers themselves. For now mypy [still doesn't](https://github.com/python/mypy/issues/11695) support it properly, but Pyright does.

hauntsaninja:
Related https://github.com/python/mypy/issues/2220

matangover:
I suggest adding to `_typeshed` an alias `AnyIterable: TypeAlias = Union[Iterable | SupportsGetItem]`

matangover:
Or, well, since SupportsGetItem has two type parameters that won't exactly work, but something along those lines.

JelleZijlstra:
This affects every single place where `Iterable` is accepted as an argument, except in the unlikely case that the code directly calls `.__iter__`.

We could add an alias like you suggest and use it everywhere, but that doesn't seem like a great solution. The alias wouldn't get used in code outside typeshed, it would be harder for users to pick up this special case, and type checker error messages may get worse.

matangover:
> The alias wouldn't get used in code outside typeshed, it would be harder for users to pick up this special case

I don't think it's common to call `iter` or `PySequence_Fast` on function arguments in "user" code... At least I never do that in my code. Do you suggest to add a new type to `typing_extensions` or `typing`?

> type checker error messages may get worse

True. But in a minor way. It would probably only change from showing "Iterable" to "SequenceOrIterable". Which is in fact more correct and less misleading.

matangover:
In the end, the job of typeshed is to correctly annotate the stdlib

JelleZijlstra:
[...]

PR #8597, opened 2022-08-23
`__getitem__` iterable support for builtins, collections, itertools

mixilchenko:
Lots of objects/collections/itertation tools use `for` cycle or even `iter` method explicitly.
`iter` method supports not only `Iterable` but also instances of classes with only `__getitem__` method implemented so stubs for commonly used things like `list, `dict`, `filter`, etc could be improved.

In this PR I add this support only for `builtins`, `collections` and `itertools` as I think these modules need this most.


AlexWaygood:
Thanks, but we discussed this in depth in #7813. My opinion has not changed since I wrote https://github.com/python/typeshed/issues/7813#issuecomment-1125172257.

I'm guessing you came here from [the recent discussion on discuss.python.org](https://discuss.python.org/t/deprecate-old-style-iteration-protocol/17863). Thanks for trying to help fix the problem, but the mypy_primer output confirms that trying to fix this in the stubs would be disastrous, as it would mean that all users of these classes would have to update their annotations, and `Iterable` would essentially become useless for type annotations. I don't think fixing this problem justifies such a huge breaking change to the type system.

I think if this is to be fixed, it needs to be done by adding special-casing to type checkers in some way so that they understand that classes with `__getitem__` methods are iterable, even if they don't have `__iter__` methods.

I'm sorry, but I just don't think this PR is the way to go :)

PR #8600, opened 2022-08-23
Update ForwardRef._evaluate for Python 3.11

aronbierbaum:
`recursive_guard` argument was added in https://github.com/python/cpython/pull/21553


sobolevn:
Thank you! You need to remove this line: https://github.com/python/typeshed/blob/2066b9533c5845dbecc5e7841efefaa2974b46fc/tests/stubtest_allowlists/py311.txt#L46

stdlib/typing.pyi
```diff
@@ -820,7 +820,13 @@ class ForwardRef:
     else:
         def __init__(self, arg: str, is_argument: bool = ...) -> None: ...
 
-    def _evaluate(self, globalns: dict[str, Any] | None, localns: dict[str, Any] | None) -> Any | None: ...
+    if sys.version_info >= (3, 11):
```
AlexWaygood, line 823: python/cpython#21553 was backported to 3.9

```suggestion
    if sys.version_info >= (3, 9):
```

aronbierbaum: Done, thanks!

hauntsaninja (approved):

PR #8602, opened 2022-08-24
Fix subprocess.CompletedProcess for 3.8

czardoz:
In 3.8, this is not a generic class: https://github.com/python/cpython/blob/4f100fe9f1c691145e3fa959ef324646e303cdf3/Lib/subprocess.py#L419

Making it generic leads to errors at runtime. (The current workaround is to add `from __future__ import annotations` to the module).

### DIFF
--- a/stdlib/subprocess.pyi
+++ b/stdlib/subprocess.pyi
@@ -102,6 +102,7 @@ if sys.version_info >= (3, 9):
         ) -> None: ...
         def check_returncode(self) -> None: ...
         def __class_getitem__(cls, item: Any) -> GenericAlias: ...
+
 else:
     class CompletedProcess:
         # morally: _CMD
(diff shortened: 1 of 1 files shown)


hauntsaninja:
Thanks, but you should use `from __future__ import annotations`
https://mypy.readthedocs.io/en/stable/runtime_troubles.html#using-classes-that-are-generic-in-stubs-but-not-at-runtime

czardoz:
@hauntsaninja That's a fine workaround, but shouldn't typeshed reflect the actual implementation in the stdlib?

hauntsaninja:
Not in this case — after all, nothing in the standard library even inherits from Generic in the first place. And you can see above^ how this causes worse type checking for users. The sprinkling of `__class_getitem__` in stdlib is one of the few places where stdlib reflects typeshed, not the other way around.

PR #8603, opened 2022-08-24
tqdm: All `leave` arguments are optional

liam-ly:
According to the [docstring](https://github.com/tqdm/tqdm/blob/4f208e72552c4d916aa4fe6a955349ee8b2ed353/tqdm/std.py#L869), the `leave` argument in all tqdm constructors is an optional bool. `None` has a different behavior compared to `False` coming from [this line](https://github.com/tqdm/tqdm/blob/4f208e72552c4d916aa4fe6a955349ee8b2ed353/tqdm/std.py#L1310) originating from [this PR](https://github.com/tqdm/tqdm/issues/230#issuecomment-277364386).

In `typeshed`, `leave` is typed as bool, so this change adds the optionality to the argument.


srittau (approved):
Thanks!

Just a side note on terminology: While the `None` union is named `Optional`, I'd recommend to avoid calling "nonable" types "optional". There's been a lot of confusion in the past over `Optional` (i.e. `Union[..., None]`) and optional arguments, i.e. those with a default value.

PR #8604, opened 2022-08-24
Further enhance `check_consistent.py`

AlexWaygood:
Add tests to make sure that:
- All files in `test_cases` are `.py` files.
- All `.py` files in `test_cases` have names starting with `test_`.
- All requirements listed in `.pre-commit.config.yaml` also appear in `requirements-tests.txt`
- All requirements listed in `.pre-commit.config.yaml` have the same version specifier as the ones in `requirements-tests.txt`

Also rename the `check_same_files()` function to `check_no_symlinks()`, which is more accurate these days.

### DIFF
--- a/tests/check_consistent.py
+++ b/tests/check_consistent.py
@@ -11,15 +11,18 @@ import sys
 from pathlib import Path
 
 import tomli
+import yaml
 from packaging.requirements import Requirement
+from packaging.specifiers import SpecifierSet
 from packaging.version import Version
 
 metadata_keys = {"version", "requires", "extra_description", "obsolete_since", "no_longer_updated", "tool"}
 tool_keys = {"stubtest": {"skip", "apt_dependencies", "extras", "ignore_missing_stub"}}
+extension_descriptions = {".pyi": "stub", ".py": ".py"}
 
 
-def assert_stubs_only(directory: Path, allowed: set[str]) -> None:
-    """Check that given directory contains only valid stub files."""
+def assert_consistent_filetypes(directory: Path, *, kind: str, allowed: set[str]) -> None:
+    """Check that given directory contains only valid Python files of a certain kind."""
     allowed_paths = {Path(f) for f in allowed}
     contents = list(directory.iterdir())
     while contents:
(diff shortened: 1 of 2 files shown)


tests/check_consistent.py
```diff
@@ -126,10 +143,48 @@ def check_metadata() -> None:
...
+
+def get_precommit_requirements() -> dict[str, SpecifierSet]:
+    with open(".pre-commit-config.yaml") as precommit_file:
+        precommit = precommit_file.read()
+    yam = yaml.load(precommit, Loader=yaml.Loader)
+    precommit_requirements = {}
+    for repo in yam["repos"]:
+        hook = repo["hooks"][0]
+        package_name, package_rev = hook["id"], repo["rev"]
+        package_specifier = SpecifierSet(f"=={package_rev.removeprefix('v')}")
```
AlexWaygood, line 161: This feels slightly fragile; not sure if there's a better way

Akuli:
It seems to me that by convention, `test_` means "you can run `pytest` or `unittest` or something over this directory". Should we instead enforce the opposite, no files named `test_blah` in our test cases directory that isn't really tests in the usual sense? Maybe not, because the non-`test_` file names will show up in editors, and the `test_` prefix avoids confusing them with stubs and scripts?

AlexWaygood:
> It seems to me that by convention, `test_` means "you can run `pytest` or `unittest` or something over this directory". Should we instead enforce the opposite, no files named `test_blah` in our test cases directory that isn't really tests in the usual sense? Maybe not, because the non-`test_` file names will show up in editors, and the `test_` prefix avoids confusing them with stubs and scripts?

Hmm, interesting. I'm not wedded to having them be all be called `test_foo.py`. I mainly added the test here because it was easy to test for, and @sobolevn brought it up in #8520 and #8521.

I guess I think the most important thing is for us to be consistent in whether we call them `test_foo.py` or just `foo.py`.

AlexWaygood:
If we wanted to establish a naming convention that distinguishes these files from the rest of typeshed, but also make clear that these aren't "tests" in the conventional sense, that you can run with `pytest`, maybe we could establish a naming convention with _suffixes_ instead. Instead of `test_foo.py`, `foo_tests.py`?

sobolevn:
Or `typecheck_` 🤔

AlexWaygood:
> Or `typecheck_` 🤔

Ooh I quite like that (though I'd go for the slightly less verbose `check_`, e.g. `check_pow.py`, etc.)

tests/check_consistent.py
```diff
@@ -29,14 +32,17 @@ def assert_stubs_only(directory: Path, allowed: set[str]) -> None:
             continue
         if entry.is_file():
             assert entry.stem.isidentifier(), f"Files must be valid modules, got: {entry}"
-            assert entry.suffix == ".pyi", f"Only stub files allowed, got: {entry}"
+            bad_filetype = (
+                f'Only {extension_descriptions[kind]} files allowed in the "{directory}" directory; ' f"got: {entry.suffix!r}"
```
AlexWaygood, line 36:
```suggestion
                f'Only {extension_descriptions[kind]} files allowed in the "{directory}" directory; got: {entry.suffix!r}'
```

tests/check_consistent.py
```diff
@@ -126,10 +143,48 @@ def check_metadata() -> None:
...
+            precommit_requirements[req.name] = req.specifier
+    return precommit_requirements
+
+
+def check_requirements() -> None:
+    requirements_txt_requirements = get_txt_requirements()
+    precommit_requirements = get_precommit_requirements()
+    for requirement, specifier in precommit_requirements.items():
+        no_txt_entry_msg = "All pre-commit requirements must also be listed in `requirements-tests.txt`"
+        assert requirement in requirements_txt_requirements, no_txt_entry_msg
```
JelleZijlstra, line 174: Same here

tests/check_consistent.py
```diff
@@ -126,10 +143,48 @@ def check_metadata() -> None:
...
+        specifier_mismatch = (
+            f"Specifier for {requirement!r} in `.pre-commit-config.yaml` "
```
JelleZijlstra, line 176: Would be good to include the text of the two specifiers here.

JelleZijlstra (approved):

PR #8605, opened 2022-08-24
Add `aiohttp` and `tomlkit` to `requirements-tests.txt`

AlexWaygood:


hauntsaninja (approved):
Looks good — if the idea is to have stubsabot deps here, would you mind removing them from stubsabot.yml as well?

AlexWaygood:
> Looks good — if the idea is to have stubsabot deps here, would you mind removing them from stubsabot.yml as well?

Ooh good point, I'll get on that!

PR #8606, opened 2022-08-24
Delete `scripts/update-stubtest-allowlist.py`

AlexWaygood:
We used to run this in CI, but haven't done for a while. I'm not sure it's useful anymore.


AlexWaygood:
@srittau, I think you originally added this script -- any objections to this? 🙂

srittau (approved):

PR #8609, opened 2022-08-25
Bump protobuf to 3.20.1

nipunn1313:
There are even newer versions available (4.*), but I had found a bug
in it which prevented it from working with mypy-protobuf. For now,
sticking to just 3.20.1 until that stuff is resolved.


nipunn1313:
(converting to draft so I can fix stubtest)

nipunn1313:
I pushed some new code - but the github CI hasn't seemed to rerun yet - not sure what the hold up is. Poking around.

AlexWaygood:
> I pushed some new code - but the github CI hasn't seemed to rerun yet - not sure what the hold up is. Poking around.

I think it did re-run? https://github.com/python/typeshed/runs/8015005080?check_suite_focus=true (the later stubtest run) has different errors to the earlier stubtest run (https://github.com/python/typeshed/runs/8014591866?check_suite_focus=true) 🙂

nipunn1313:
it just took a while to run. I think there's some limitation where it waits for the previous run to fully complete before starting another run.

nipunn1313:
Ok looks like it completed successfully!

scripts/generate_proto_stubs.sh
```diff
@@ -11,7 +11,7 @@ set -ex -o pipefail
 # followed by committing the changes to typeshed
 #
 # Update these two variables when rerunning script
-PROTOBUF_VERSION=3.19.3
+PROTOBUF_VERSION=3.20.1
 MYPY_PROTOBUF_VERSION=v3.2.0
```
AlexWaygood, line 15:
```suggestion
MYPY_PROTOBUF_VERSION=v3.3.0
```

stubs/protobuf/METADATA.toml
```diff
@@ -1,2 +1,2 @@
-version = "3.19.*"
+version = "3.20.*"
 extra_description = "Generated with aid from mypy-protobuf v3.2.0"
```
AlexWaygood, line 2:
```suggestion
extra_description = "Generated with aid from mypy-protobuf v3.3.0"
```

AlexWaygood (commented):
Is it worth bumping the `mypy-protobuf` version being used here?

scripts/generate_proto_stubs.sh
nipunn1313: yes, but I'd like to do it separately - because it ends up having a pretty different set of changes.

I originally was preparing a PR that did both all at once, but it ends up with a bunch of different changes since it changes the code-generator utility.

Would prefer to keep the two PRs isolated.

AlexWaygood: Sure, makes sense!

nipunn1313: I was also planning to test the changes with 3.3.0 and if things don't work out perfectly, I may have to release 3.3.1 (or 3.4.0) to get typeshed to work.

AlexWaygood (approved):
Haven't reviewed this in depth, but nothing sticks out, and you're the expert when it comes to these stubs!

nipunn1313:
sweet! Will send you another one updating mypy-protobuf too!

PR #8610, opened 2022-08-25
Support tests/mypy_test.py [package]

nipunn1313:
Currently, the testsuite early exits if it's filtering on a single
package. With this change, it will correctly complete.

For example `tests/mypy_test.py protobuf`

### DIFF
--- a/tests/mypy_test.py
+++ b/tests/mypy_test.py
@@ -341,6 +341,9 @@ def test_third_party_distribution(distribution: str, args: TestConfig) -> TestRe
     seen_dists: set[str] = set()
     add_third_party_files(distribution, files, args, configurations, seen_dists)
 
+    if not files and args.filter:
+        return TestResults(0, 0)
+
     print(f"testing {distribution} ({len(files)} files)... ", end="")
 
     if not files:


tests/mypy_test.py
```diff
@@ -341,6 +341,9 @@ def test_third_party_distribution(distribution: str, args: TestConfig) -> TestRe
     seen_dists: set[str] = set()
     add_third_party_files(distribution, files, args, configurations, seen_dists)
 
+    if not files and args.filter:
+        return TestResults(0, 0)
+
```
nipunn1313, line 346: previously on line 351 of this file, it would sys.exit after searching the very first package and finding out that none of the files match.

Now it early-skips all the non-matching packages when we're filtering.

nipunn1313: There might be a different or better way to solve this, but this definitely works.

AlexWaygood (approved):
Nice! This makes `filter` actually useful! (I was wondering whether we should get rid of it; see #8139)

nipunn1313:
yep that makes sense. This PR is more of a targeted fix to the flag - but if you want to rethink it - that also sounds reasonable!

PR #8582, opened 2022-08-21
add `_ctypes.pyi`

junkmd:
see #8571

PR #8614, opened 2022-08-25
Add win32-specific constants to `_ctypes`

junkmd:
I added win32-specific constants to `_ctypes`.

(related to #8582 that `stdlib/_ctypes.pyi` was added)


AlexWaygood (approved):
Thanks!

PR #8617, opened 2022-08-25
Fix os.{,p}{read,write}v to take Sequence of ReadableBuffer / WriteableBuffer

### LINKED ISSUE #8615: `os.readv` should take `Sequence[WriteableBuffer]`
`os.readv` is currently defined as:

https://github.com/python/typeshed/blob/179228370af7a0073b90e9b3bf84ecf46950e948/stdlib/os/__init__.pyi#L645

... but the function is documented as supporting any writable buffer:

https://docs.python.org/3/library/os.html#os.readv

>Read from a file descriptor *fd* into a number of mutable [bytes-like objects](https://docs.python.org/3/glossary.html#term-bytes-like-object) *buffers*.

For example, the current definition disallows calling it with a sequence of `memoryview`s.

`writev` could also be changed to take `Sequence[ReadOnlyBuffer]` instead of `Sequence[bytes]`. Of course, `ReadOnlyBuffer` is an alias for `bytes` anyway so this won't change anything except be cosmetically consistent.

### LINKED ISSUE #8616: `os.preadv` and `os.pwritev` are specified incorrectly
https://github.com/python/typeshed/blob/179228370af7a0073b90e9b3bf84ecf46950e948/stdlib/os/__init__.pyi#L627-L628

1. `__buffers` should be a `Sequence`, not just an `Iterable`. This is backed up by both the documentation and the CPython impl.

1. `preadv` should take mutable buffers, not readonly ones. So it should be either `Sequence[bytearray]`, or if #8615 is going to be fixed, `Sequence[WriteableBuffer]`

Arnavion:
Fixes #8615
Fixes #8616

### DIFF
--- a/stdlib/os/__init__.pyi
+++ b/stdlib/os/__init__.pyi
@@ -13,6 +13,7 @@ from _typeshed import (
     Self,
     StrOrBytesPath,
     StrPath,
+    SupportsLenAndGetItem,
     WriteableBuffer,
     structseq,
 )
@@ -626,8 +627,18 @@ if sys.platform != "win32":
     if sys.platform != "darwin":
         if sys.version_info >= (3, 10):
             RWF_APPEND: int  # docs say available on 3.7+, stubtest says otherwise
-        def preadv(__fd: int, __buffers: Sequence[WriteableBuffer], __offset: int, __flags: int = ...) -> int: ...
-        def pwritev(__fd: int, __buffers: Sequence[ReadOnlyBuffer], __offset: int, __flags: int = ...) -> int: ...
+        def preadv(
+            __fd: int,
+            __buffers: SupportsLenAndGetItem[WriteableBuffer],
+            __offset: int,
+            __flags: int = ...,
+        ) -> int: ...
+        def pwritev(
+            __fd: int,
+            __buffers: SupportsLenAndGetItem[ReadOnlyBuffer],
+            __offset: int,
+            __flags: int = ...,
+        ) -> int: ...
         RWF_DSYNC: int
         RWF_SYNC: int
         RWF_HIPRI: int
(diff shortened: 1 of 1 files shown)


Arnavion:
Thinking about it some more, should `writev` and `pwritev` take `ReadableBuffer` instead of `ReadOnlyBuffer` ? The original code was taking `bytes` which is why I replaced it with the equivalent `ReadOnlyBuffer`, but `ReadableBuffer` is more general.

`os.write` takes `bytes` (ie `ReadOnlyBuffer`) too, but the docs say it can take any bytestring and the CPython allows it to be any buffer, so that too sounds like it could be `ReadableBuffer`. `socket.write` does take `ReadableBuffer`

stdlib/os/__init__.pyi
```diff
@@ -624,8 +626,8 @@ if sys.platform != "win32":
     if sys.platform != "darwin":
         if sys.version_info >= (3, 10):
             RWF_APPEND: int  # docs say available on 3.7+, stubtest says otherwise
-        def preadv(__fd: int, __buffers: Iterable[bytes], __offset: int, __flags: int = ...) -> int: ...
-        def pwritev(__fd: int, __buffers: Iterable[bytes], __offset: int, __flags: int = ...) -> int: ...
+        def preadv(__fd: int, __buffers: Sequence[WriteableBuffer], __offset: int, __flags: int = ...) -> int: ...
```
JelleZijlstra, line 629: Looking at the C code, this only needs `__getitem__` and `__len__`, not an object that supports the sequence protocol. Can you use `_typeshed.SupportsLenAndGetItem`?

Arnavion: https://github.com/python/cpython/blob/0319cd6825f3c247a875c91493a38992fb33a5b3/Modules/posixmodule.c#L9749

Doesn't this mean it needs to be a Sequence?

JelleZijlstra: No, PySequence_Check checks only for `__getitem__`. It is not equivalent to Sequence in Python.

Arnavion: Okay. All four functions do the same check, so I've done that for all of them.

What about `ReadOnlyBuffer` vs `ReadableBuffer` in my comment above?

JelleZijlstra: Thanks. You're correct in your comment; the C code indicates they take any readable or writable buffer.

Arnavion: Okay, done that as well.

JelleZijlstra (approved):

PR #8627, opened 2022-08-26
redis: Improve `ClusterPubSub` types

sobolevn:
Source: https://github.com/redis/redis-py/blob/ab5d407a03192bd8ab63a1bfc891db9c04330265/redis/cluster.py#L1558


AlexWaygood (approved):
Thanks! I think some of these `Any`s might be better as `Incomplete`s, but I can see arguments both ways, and it's not a huge deal

Issue #8628, opened 2022-08-26
type annotation for str.translate incomplete


GrozniyToaster:
After reading the docs about [str.translate](https://docs.python.org/3/library/stdtypes.html#str.translate), I wrote a class with `__getitem__`:
```python
class SeparatorTranslator:
    separators_whitelist: Final = {' ', '\n', '\t'}

    def __getitem__(self, char_order_index: int) -> str:
        char = chr(char_order_index)
        if char.isspace() and char not in self.separators_whitelist:
            return ' '
        else:
            return char
```
When I check code like `'asd asd \v\t asd'.translate(separator_translator)` by mypy got an error:

```console
 error: Argument 1 to "translate" of "str" has incompatible type "SeparatorTranslator"; expected "Union[Mapping[int, Union[int, str, None]], Sequence[Union[int, str, None]]]"  [arg-type]
```
Seems `__table` argument  must be `SupportsGetItem[int, int | str | None]` 

Can I fix it?

srittau:
I haven't looked at this, but generally all PRs to fix typing problems are welcome!

PR #8629, opened 2022-08-26
type annotation for str.translate

GrozniyToaster:
by issue [#8628](https://github.com/python/typeshed/issues/8628)


stdlib/builtins.pyi
```diff
@@ -535,7 +538,7 @@ class str(Sequence[str]):
     def title(self: LiteralString) -> LiteralString: ...
     @overload
     def title(self) -> str: ...  # type: ignore[misc]
-    def translate(self, __table: Mapping[int, int | str | None] | Sequence[int | str | None]) -> str: ...
+    def translate(self, __table: _TranslateTable) -> str: ...
```
sobolevn, line 541: Or maybe `_typeshed.SupportsGetItem`? https://github.com/python/typeshed/blob/0259068ad6bc814a510009cade00fd961b5d0355/stdlib/_typeshed/__init__.pyi#L121

Note, that variance there is different.

sobolevn: I had this example in mind:

```python
>>> class Some:
...    def __getitem__(self, key):
...        return MyInt(1)
... 
>>> 'a'.translate(Some())
'\x01'
```

sobolevn: Oh, `SupportsGetItem` is also a `Container`, never mind.

Akuli (approved):
Thanks! The example from your issue now passes the type check, even when modified to return a custom subclass of `int`:

```python
from typing import Final
class MyInt(int): pass

class SeparatorTranslator:
    separators_whitelist: Final = {' ', '\n', '\t'}

    def __getitem__(self, char_order_index: int) -> str | MyInt:
        char = chr(char_order_index)
        if char.isspace() and char not in self.separators_whitelist:
            return MyInt(ord(' '))
        else:
            return char

print('asd asd \v\t asd'.translate(SeparatorTranslator()))
```

Akuli:
Presumably `bytes.translate()` and `bytearray.translate()` could be updated too, but they seem to be different from `str.translate()` and can be done in a separate PR.

PR #8630, opened 2022-08-26
Add datetime to imaplib.Time2Internaldate

jheiss:
Per both the documentation https://docs.python.org/3/library/imaplib.html#imaplib.Time2Internaldate and the implementation at https://github.com/python/cpython/blob/main/Lib/imaplib.py#L1532 this function takes a datetime.


stdlib/imaplib.pyi
```diff
@@ -159,4 +160,4 @@ class _Authenticator:
 def Internaldate2tuple(resp: bytes) -> time.struct_time: ...
 def Int2AP(num: int) -> str: ...
 def ParseFlags(resp: bytes) -> tuple[bytes, ...]: ...
-def Time2Internaldate(date_time: float | time.struct_time | str) -> str: ...
+def Time2Internaldate(date_time: float | time.struct_time | datetime | str) -> str: ...
```
AlexWaygood, line 163:
```suggestion
def Time2Internaldate(date_time: float | time.struct_time | time._TimeTuple | datetime | str) -> str: ...
```

AlexWaygood (approved):
Thanks!

PR #8631, opened 2022-08-27
Fix: `assertRaises` / `assertRaisesRegex` / `assertWarns` / `assertWarnsRegex` take the `msg` argument as keyword-only

MapleCCC:


AlexWaygood (approved):
Thank you!

Issue #8570, opened 2022-08-19
`ipaddress.ip_network` annotations ban tuple arguments


ischeinkman:
Right now the type annotations for the [`ipaddress.ip_network` function](https://github.com/python/typeshed/blob/8a326dc9ece9845c9ab2606c6e7567df2e768f7e/stdlib/ipaddress.pyi#L18) only allow its argument to be one of an `int`, a `str`, `bytes`, an `IPv4Address`/`IPv6Address`, or an existing network. 

However, currently the `ipaddress.ip_network` function also supports a `Tuple[IPv4Address | IPv6Address | str | int | bytes, int]` as an argument, in which case the first element becomes the base address and the second becomes the network prefix length: 

```
Python 3.8.10 (default, Jun 22 2022, 20:18:18) 
[GCC 9.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import ipaddress
>>> ipaddress.ip_network(('192.168.1.0', 24))
IPv4Network('192.168.1.0/24')
>>> ipaddress.ip_network((0, 24))
IPv4Network('0.0.0.0/24')
```

Akuli:
Thanks. PR welcome! :)

PR #8634, opened 2022-08-27
Add `tuple` argument to `ip_network`

sobolevn:
Source: 
- https://github.com/python/cpython/blob/6fbd889d6e937ad255f98b5495b78a06d05640d5/Lib/ipaddress.py#L2090
- https://github.com/python/cpython/blob/6fbd889d6e937ad255f98b5495b78a06d05640d5/Lib/ipaddress.py#L534-L555

Closes https://github.com/python/typeshed/issues/8570


stdlib/ipaddress.pyi
```diff
@@ -15,7 +15,7 @@ _RawIPAddress: TypeAlias = int | str | bytes | IPv4Address | IPv6Address
 _RawNetworkPart: TypeAlias = IPv4Network | IPv6Network | IPv4Interface | IPv6Interface
 
 def ip_address(address: _RawIPAddress) -> IPv4Address | IPv6Address: ...
-def ip_network(address: _RawIPAddress | _RawNetworkPart, strict: bool = ...) -> IPv4Network | IPv6Network: ...
+def ip_network(address: _RawIPAddress | _RawNetworkPart | Tuple[_RawIPAddress], Tuple[_RawIPAddress, int], strict: bool = ...) -> IPv4Network | IPv6Network: ...
```
AlexWaygood, line 18:
```suggestion
def ip_network(address: _RawIPAddress | _RawNetworkPart | tuple[_RawIPAddress] | tuple[_RawIPAddress, int], strict: bool = ...) -> IPv4Network | IPv6Network: ...
```

AlexWaygood (commented):
I'm a bit uneasy about this change. It's true that it works at runtime, but passing a `tuple` in here seems to be entirely undocumented and untested, so I feel like it could change at any time without any backwards-compatibility guarantees.

sobolevn:
I don't have any opinion really

Akuli:
We should in general type things as the implementation does it, not as documented. I think it applies here too. I also don't see any reason why tests for this function should check with all types that `IPv4Network` and `IPv6Network` accept, even though that's the intention; that would make the tests longer without increasing the test coverage.

If someone still feels "uneasy" about this, we could add a test case -- again, nothing too thorough, could be for example the two examples in the issue.

Akuli:
And I just realized that adding a test case to typeshed would be pointless, as those don't run at runtime :) I still think we should support this, because it works.

AlexWaygood:
> And I just realized that adding a test case to typeshed would be pointless, as those don't run at runtime :) I still think we should support this, because it works.

I'm fine with this being merged, just don't necessarily want to be the one to hit merge myself :)

I agree that in general, we should follow the implementation rather than the documentation.

Akuli:
Passing in a tuple seems to be a common pattern: https://grep.app/search?q=ip_network%28%28

AlexWaygood (approved):
I'm convinced.

sobolevn:
🥳

PR #8637, opened 2022-08-28
Use better names for pos-only parameters in `stdlib`

sobolevn:
I understand that this argument is "pos-only" emulation.
But, still `__k` is better for readability, because `__v` is used for values, not keys.


stdlib/builtins.pyi
```diff
@@ -1061,7 +1061,7 @@ class dict(MutableMapping[_KT, _VT], Generic[_KT, _VT]):
     def __len__(self) -> int: ...
     def __getitem__(self, __k: _KT) -> _VT: ...
     def __setitem__(self, __k: _KT, __v: _VT) -> None: ...
-    def __delitem__(self, __v: _KT) -> None: ...
+    def __delitem__(self, __k: _KT) -> None: ...
```
AlexWaygood, line 1064:
```suggestion
    def __delitem__(self, __key: _KT) -> None: ...
```

```pycon
>>> help(dict.__delitem__)
Help on wrapper_descriptor:

__delitem__(self, key, /)
    Delete self[key].
```

stdlib/typing.pyi
```diff
@@ -591,7 +591,7 @@ class MutableMapping(Mapping[_KT, _VT], Generic[_KT, _VT]):
     @abstractmethod
     def __setitem__(self, __k: _KT, __v: _VT) -> None: ...
     @abstractmethod
-    def __delitem__(self, __v: _KT) -> None: ...
+    def __delitem__(self, __k: _KT) -> None: ...
```
AlexWaygood, line 594:
```suggestion
    def __delitem__(self, __key: _KT) -> None: ...
```

```pycon
>>> from collections.abc import MutableMapping
>>> help(MutableMapping.__delitem__)
Help on function __delitem__ in module collections.abc:

__delitem__(self, key)
```

AlexWaygood (changes requested):
If we're going to do this, let's do it properly ;)

stdlib/builtins.pyi
sobolevn: My motivation was to be consistent with `__getitem__` and `__setitem__`, because we work with the same thing - and it should be named the same way.

AlexWaygood: I agree that the current name for this parameter in the stub is bad and confusing! But I don't see why we shouldn't copy the runtime exactly here, since the runtime has a good name for this parameter.

I'd be in favour of updating `dict.__setitem__` and `dict.__getitem__` as well. `help(dict.__getitem__)` doesn't tell you the name of this parameter at runtime:

```pycon
>>> help(dict.__getitem__)
Help on method_descriptor:

__getitem__(...)
    x.__getitem__(y) <==> x[y]
```

But the parameter name given for `dict.__setitem__` at runtime matches the parameter name for `dict.__delitem__`:

```pycon
>>> help(dict.__setitem__)
Help on wrapper_descriptor:

__setitem__(self, key, value, /)
    Set self[key] to value.
```

sobolevn: Related: https://github.com/python/cpython/issues/96364

sobolevn:
Done! 
- `__k` -> `__key`
- `__v` -> `__value`

stdlib/typing.pyi
sobolevn: As far as I remember, `MutableMapping` and its arguments is very complex.
Because of how `dict` is implemented:

```python
>>> help(dict.__delitem__)
Help on wrapper_descriptor:

__delitem__(self, key, /)
    Delete self[key].
```

So, I prefer to not change any semantics here. Only a name that does not affect anything.

AlexWaygood: Correct, we definitely shouldn't make the argument positional-or-keyword! That would break a lot of things :)

AlexWaygood (approved):
Thanks!

Issue #8633, opened 2022-08-27
migrate some stuffs in `ctypes/__init__.pyi` into `_ctypes.pyi`, but circular import occurs


junkmd:
When we submitted PR that adds the `stdlib/_ctypes.pyi`(#8582), imports some stuffs from `ctypes`.

But [the implementation imports some stuffs from `_ctypes` to `ctypes`](https://github.com/python/cpython/blob/0ace820bec8892d621a4aadc1feb6c56e25560bf/Lib/ctypes/__init__.py#L8-L14).
- also related to https://github.com/python/typeshed/issues/8571#issuecomment-1221316973

However, trying to write type stubs as close as possible to the implementation would result in circular imports if we just move definitions around.
- For example, `CDLL` in args of `_CData` methods.
  - In implementation... 
    - `CDLLL` is defined in `ctypes`
    - `_CData`(base class of `_SimpleCData`) is defined in `_ctypes`.
- Some PRs, such as #5350, were merged as "should be fine in `pyi` files" even if they are circular imports.

Any opinions would be appreciated.

hauntsaninja:
You don't really need to worry about circular imports, since stub files are never imported by Python. All type checkers know how to resolve circular imports when resolving references. That said, in this case you could probably move the fake definitions `_CArgObject, _PointerLike` to `_ctypes` and import them in `ctypes`.

junkmd:
When we tried to make the change, we found that the test using [`pytype`](https://github.com/google/pytype) running in CI failed.
Please see #8643.

I will work on adding TODO comments to the stuffs originally defined in `_ctypes` until `pytype` is fixed or better workarounds are found.

junkmd:
I submitted issue to `pytype` google/pytype#1281

junkmd:
The `pytype` problem causes by [its own `_ctypes` stub](https://github.com/google/pytype/blob/ea1bc6734716e30afc4dabff20c059d3807f9965/pytype/stubs/stdlib/_ctypes.pytd).

This will be fixed ["in a few weeks"](https://github.com/google/pytype/issues/1281#issuecomment-1231093224).

junkmd:
## List of the stuffs in `ctypes/__init__.pyi` those what originally defined in `_ctypes`
- `ArgumentError`
- `Array`
- `FormatError`
- `LoadLibrary`
- `POINTER`
- `RTLD_GLOBAL`(TODO marked)
- `RTLD_LOCAL`(TODO marked)
- `Structure`
[...]

Issue #8571, opened 2022-08-20
Add `stdlib/_ctypes`


junkmd:
My projects use some COM-ffi packages.

I use `_ctypes.COMError` or `_ctypes.CopyComPointer` too, but they are not defined in the type stub, so `mypy` type checker raises `Cannot find implementation or library stub for module named "_ctypes"` error.

I would like to add type stubs like this.
```py
from ctypes import _CArgObject, _PointerLike
# are they correct?
_COMError_Details = tuple[str | None, str | None, str | None, int | None, int | None]  # Description, Source, HelpFile, HelpContext, scode

class COMError(Exception):
    """Raised when a COM method call failed."""
    hresult: int
    text: str
    details: _COMError_Details

    def __init__(self, hresult: _COMError_HResult, text: str, details: _COMError_Details) -> None: ...


def CopyComPointer(src: _PointerLike, dst: _PointerLike | _CArgObject) -> int: ...  # `dst` sometimes maybe return value of `ctypes.byref`.
    """CopyComPointer(src, dst) -> HRESULT value"""
```

But other APIs are not clear to me.
- and where the `if sys.platform == "win32":` needed.

Any opinions would be appreciated.

Akuli:
Feel free to make a pull request :) There's already lots of other `_foo.pyi` modules in `stdlib/`.

Some thoughts:
- We could mark `_ctypes` as incomplete. This is explained in CONTRIBUTING.md.
- Com stuff isn't available on my linux system, so presumably everything except the import should go under `if sys.platform == "win32"`.
- Can `_COMError_Details` really be something like `("blah", None, None, 123, None)`? If it's not possible for some but not all elements to be None, it should be `tuple[str, str, str, int, int] | tuple[None, None, None, None, None]`.

AlexWaygood:
Most of the stuff that we currently have in `stdlib/ctypes/__init__.pyi` is actually re-exported from `_ctypes` at runtime. If we add `_ctypes.pyi`, we should eventually move those definitions from `stdlib/ctypes/__init__.pyi` to `_ctypes.pyi, to mimic the runtime more precisely. (That doesn't need to be done in the initial PR adding `_ctypes.pyi, though.)

junkmd:
@Akuli 

Thanks for your response.
I am motivated to contribute to this repository.

> * We could mark `_ctypes` as incomplete. This is explained in CONTRIBUTING.md.

I see. I will have to peruse the `CONTRIBUTING.md`.


[...]

PR #8643, opened 2022-08-29
add and migrate constants from `ctypes/__init__.pyi` to `_ctypes.pyi`

junkmd:
I have moved some constants in `ctypes/__init__.pyi` to `_ctypes.pyi` as defined in the implementation.

I also added some constants defined in `_ctypes`.

I will move other stuffs eventually, but I will try to do it little by little.

- related to #8571 and #8633

### DIFF
--- a/stdlib/ctypes/__init__.pyi
+++ b/stdlib/ctypes/__init__.pyi
@@ -5,7 +5,9 @@ from collections.abc import Callable, Iterable, Iterator, Mapping, Sequence
 from typing import Any, ClassVar, Generic, TypeVar, Union as _UnionT, overload
 from typing_extensions import TypeAlias
 
-from _ctypes import RTLD_GLOBAL as RTLD_GLOBAL, RTLD_LOCAL as RTLD_LOCAL
+# TODO: import these from _ctypes once it no longer breaks pytype
+RTLD_GLOBAL: int
+RTLD_LOCAL: int
 
 if sys.version_info >= (3, 9):
     from types import GenericAlias
--- a/stdlib/ctypes/__init__.pyi
+++ b/stdlib/ctypes/__init__.pyi
@@ -5,6 +5,8 @@ from collections.abc import Callable, Iterable, Iterator, Mapping, Sequence
 from typing import Any, ClassVar, Generic, TypeVar, Union as _UnionT, overload
 from typing_extensions import TypeAlias
 
+from _ctypes import RTLD_GLOBAL as RTLD_GLOBAL, RTLD_LOCAL as RTLD_LOCAL
+
 if sys.version_info >= (3, 9):
     from types import GenericAlias
 
@@ -12,8 +14,6 @@ _T = TypeVar("_T")
 _DLLT = TypeVar("_DLLT", bound=CDLL)
 _CT = TypeVar("_CT", bound=_CData)
 
-RTLD_GLOBAL: int
-RTLD_LOCAL: int
 DEFAULT_MODE: int
 
 class CDLL:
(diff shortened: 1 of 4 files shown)


AlexWaygood (approved):
Great stuff, thank you!

AlexWaygood:
Hmm, I'm not sure what's up with the pytype crash. This _might_ be as a result of the circular import :(

I'll try bumping the pytype version in CI to see if that helps, since I think pytype's done some work on improving their name-resolution logic recently.

AlexWaygood:
> I'll try bumping the pytype version in CI to see if that helps, since I think pytype's done some work on improving their name-resolution logic recently.

No, looks like that didn't fix it ☹️ (cc. @rchen152)

As a workaround for now, maybe you could leave `RTLD_LOCAL` and `RTLD_GLOBAL` as they are in `ctypes/__init__.pyi`, and add a TODO comment saying that ideally we would re-export them from `_ctypes.pyi`?

Akuli:
We could also use `tests/pytype_exclude_list.txt`.

junkmd:
@Akuli 
@AlexWaygood 

It is curious that no other modules using circular imports have this error, such as `importlib`.

I will add workarounds to `tests/pytype_exclude_list.txt` in the hope that someone will do a fix for `pytype`.

Akuli:
Circular imports don't automatically mean that pytype will crash. They only crash pytype when hitting some corner case that wasn't considered before.

junkmd:
@Akuli 
(cc: @AlexWaygood)

I added all the modules in `stdlib/ctypes` to the list and still got the error.

According to the error messages, some errors seem to occur in various modules importing `ctypes`.

Do I have to change the `pyi` file side currently?

Akuli:
Because `_typeshed/__init__.pyi` imports `ctypes`, and because lots of other modules use `_typeshed`, using the exclude list would mean we need to exclude a lot of things. It isn't a good idea after all.

Instead, we can:
- Wait for pytype to get fixed
- Try reverting some of the changes in this PR to make pytype pass (and fix them in a future PR once the pytype bug is fixed)

stdlib/ctypes/__init__.pyi
```diff
@@ -5,15 +5,15 @@ from collections.abc import Callable, Iterable, Iterator, Mapping, Sequence
 from typing import Any, ClassVar, Generic, TypeVar, Union as _UnionT, overload
 from typing_extensions import TypeAlias
 
+from _ctypes import RTLD_GLOBAL as RTLD_GLOBAL, RTLD_LOCAL as RTLD_LOCAL
```
Akuli, line 8:
```suggestion
# TODO: import these from _ctypes once it no longer breaks pytype
RTLD_GLOBAL: int
RTLD_LOCAL: int
```

tests/pytype_exclude_list.txt
```diff
@@ -10,6 +10,12 @@ stubs/mysqlclient/MySQLdb/__init__.pyi
 stubs/mysqlclient/MySQLdb/connections.pyi
 stubs/mysqlclient/MySQLdb/cursors.pyi
 
+# errors about circular imports
+stdlib/ctypes/__init__.pyi
+stdlib/ctypes/util.pyi
+stdlib/ctypes/wintypes.pyi
+stdlib/_ctypes.pyi
+
```
Akuli, line 18:
```suggestion
```

junkmd:
@Akuli 
Thank you for code correction.

AlexWaygood (approved):
LGTM!

PR #8648, opened 2022-08-30
stdlib/_ast.pyi: Fix ast.ImportFrom arg module _identifier to str

iyume:
ast.ImportFrom module field should be `str | None`.

`from ..a.b import f` dumps `ImportFrom(module='a.b', ..., level=2)`

So _identifier typing is incorrect

### DIFF
--- a/stdlib/_ast.pyi
+++ b/stdlib/_ast.pyi
@@ -194,7 +194,7 @@ class Import(stmt):
 class ImportFrom(stmt):
     if sys.version_info >= (3, 10):
         __match_args__ = ("module", "names", "level")
-    module: _Identifier | None
+    module: str | None
     names: list[alias]
     level: int
 


srittau:
`_Identifier` is a type alias for `str`, so this change is unnecessary. That said, the type alias is ancient and I'm not sure it is necessary and useful. But in that case we should remove it altogether and not just in this one instance.

AlexWaygood:
> `_Identifier` is a type alias for `str`, so this change is unnecessary. That said, the type alias is ancient and I'm not sure it is necessary and useful. But in that case we should remove it altogether and not just in this one instance.

I think the idea is to indicate that, for many of the classes defined in `_ast.pyi`, fields don't just have to be `str`s, they also have to be valid Python identifiers. But @iyume is correct that this doesn't apply for `ast.ImportFrom.module`:

```pycon
>>> import ast
>>>  print(ast.dump(ast.parse('from collections.abc import Mapping')))
Module(body=[ImportFrom(module='collections.abc', names=[alias(name='Mapping', asname=None)], level=0)])
>>> 'collections.abc'.isidentifier()
False
```

srittau (approved):

PR #8649, opened 2022-08-30
add `Incomplete` to some untyped parameters in bs4

DetachHead:


srittau:
`Incomplete` is intended for complex cases, where we can't just leave the argument/field unannotated, which we can do here.

Issue #5736, opened 2021-07-07
paramiko stubs seem incorrect for python3


mrodden:
I am getting a bunch of errors from mypy with the `types-paramiko` package.

Some are imports that are not in the stubs for paramiko like `paramiko.PublicBlob` which is an alias import from `paramiko.pkey.PublicBlob` (see https://github.com/paramiko/paramiko/blob/master/paramiko/__init__.py#L65)

Others seem to be issues between python2 and python3 where many of the object stubs are declaring type `str` when in fact paramiko is treating it like a `bytes` (see https://github.com/python/typeshed/blob/master/stubs/paramiko/paramiko/pkey.pyi#L36 and https://github.com/paramiko/paramiko/blob/master/paramiko/pkey.py#L679-L681)

I'm not sure if there is a nice way to resolve this, as it is obviously an `str` in py2 but `bytes` in py3.

I have currently chosen to ignore the `types-paramiko` package in my project, as it causes too many false check errors with mypy.

Akuli:
>str in py2 but bytes in py3

In this case, you should just use `bytes`. In Python 2, `str` and `bytes` are the same, and `unicode` is different.

mrodden:
The issue can be highlight with the following
```
pub_blob = paramiko.pkey.PublicBlob.from_file("/path/to/file")
rsa_key = paramiko.pkey.RSAKey(data=pub_blob.key_blob)
```

The second line fails, since the stubs say that key_blob is a `str` and the `data` kwarg to RSAKey __init__ is a `bytes`.

mrm-david:
Another example : in `exec_command` the parameter `command` is `str` in the stubs; but it can actually accept `byte` as well since it eventually gets passed to `b()`which converts it to `byte` unless it was already `byte`, in which cases it returns it untouched.

Akuli:
PRs to improve the stubs are welcome :)

JelleZijlstra:
In #10109 I fix two of the specific issues mentioned above. Anything else is best handled in a focused issue/PR that describes exactly what type is wrong.

PR #8651, opened 2022-08-30
refactor: add missing paramiko types in init and version

kkirsche:
This begins to work on https://github.com/python/typeshed/issues/5736

This adds missing `__init__` and `_version` variables.

Based on this source code:
https://github.com/paramiko/paramiko/blob/main/paramiko/__init__.py
https://github.com/paramiko/paramiko/blob/main/paramiko/_version.py

I believe `util` in `__all__` is referencing the `paramiko.util`, but it's unclear if this was left out intentionally before.


AlexWaygood (approved):
Thanks!

Issue #7855, opened 2022-05-17
`LoggerAdapter` is typed as a generic but using it as such raises an error at runtime


babolivier:
With the following code:

```python
class ContextLoggingAdapter(logging.LoggerAdapter):
    def process(self, msg: str, kwargs: Any) -> Tuple[str, Any]:
        ...
```

Mypy raises the following `type-arg` error : `error: Missing type parameters for generic type "LoggerAdapter"`

According to [this type stub](https://github.com/python/typeshed/blob/81be86af6e27ac0135b11c7f6697d8ae9623a034/stdlib/logging/__init__.pyi#L413-L414) it's typed as a generic and should be passed the type of the logger it's an adapter for. However, changing the code into this:

```python
class ContextLoggingAdapter(logging.LoggerAdapter[logging.Logger]):
    def process(self, msg: str, kwargs: Any) -> Tuple[str, Any]:
        ...
```

makes mypy happy but raises an error at runtime: `TypeError: 'type' object is not subscriptable`

I'm not sure if this is an issue with the stubs or with mypy (and I'm not fully sure how they interact), so apologies if this issue should have been created on https://github.com/python/mypy instead (feel free to transfer it there if that's the case).

AlexWaygood:
You created the issue in the right place :) and, this is a problem we're aware of.

`LoggerAdapter` will actually be generic at runtime on 3.11: https://github.com/python/cpython/pull/92129

In the meantime, however, @Akuli's advice here is probably the best way around this issue: https://github.com/python/typeshed/pull/5954#issuecomment-1114270968:

```python
import logging
from typing import TYPE_CHECKING

if TYPE_CHECKING:
    _LoggerAdapter = logging.LoggerAdapter[logging.Logger]
else:
    _LoggerAdapter = logging.LoggerAdapter

class MyAdapter(_LoggerAdapter):
    pass
```

There's also some great mypy docs on this problem here: https://mypy.readthedocs.io/en/stable/runtime_troubles.html#using-classes-that-are-generic-in-stubs-but-not-at-runtime

babolivier:
That makes sense, thanks a lot for the details and references :)

PR #8652, opened 2022-08-30
Added stubs for D3DShot

Avasam:
I've filled in the types as best as I could. But I might need help with a few things.
1. I am not sure how to deal with the placeholder `Pointer` types yet. I think I may have to create new classes just to expose their methods when type-checking. Although I think this PR is probably acceptable keeping them as `Incomplete` for now.
2. I created a placeholder `Frame` type. Is it possible to type the `D3DShot` class based on which backend it uses? Otherwise I have to keep casting the `Frame`s when using it:
image


Avasam:
I don't understand the stubtest. Other than that, ready for review.

Avasam:
The Pointer types seemed quite complex for me, hopefully I got them right

.flake8
```diff
@@ -7,6 +7,7 @@
 
 # Some rules are considered irrelevant to stub files:
 #     E701   multiple statements on one line (colon) -- disallows "..." on the same line
+#     E704   multiple statements on one line (def) -- disallows "..." on the same line
```
JelleZijlstra, line 10: Why was this change needed? Seems unrelated to adding new stubs.

stubs/D3DShot/d3dshot/capture_output.pyi
```diff
@@ -0,0 +1,36 @@
...
+    NUMPY_FLOAT: int
+    PYTORCH: int
+    PYTORCH_FLOAT: int
+    PYTORCH_GPU: int
+    PYTORCH_FLOAT_GPU: int
+
+class CaptureOutputError(BaseException): ...
+
+class CaptureOutput:
+    backend: CaptureOutput
```
JelleZijlstra, line 23:
```suggestion
    backend: CaptureOutputs
```
didn't check the code but this would make more sense

stubs/D3DShot/d3dshot/capture_outputs/pil_capture_output.pyi
```diff
@@ -0,0 +1,24 @@
+from ctypes import _CVoidConstPLike
+from typing import Any, TypeVar
+from typing_extensions import TypeAlias
+
+from d3dshot.capture_output import CaptureOutput
+from PIL import Image
+
+_T = TypeVar("_T")
+_Unused: TypeAlias = Any
```
JelleZijlstra, line 9:
```suggestion
_Unused: TypeAlias = object
```
`object` makes more sense for an unused parameter.

stubs/D3DShot/d3dshot/dll/d3d.pyi
```diff
@@ -0,0 +1,219 @@
...
+    ArraySize: wintypes.UINT
+    Format: wintypes.UINT
+    SampleDesc: DXGI_SAMPLE_DESC
+    Usage: wintypes.UINT
+    BindFlags: wintypes.UINT
+    CPUAccessFlags: wintypes.UINT
+    MiscFlags: wintypes.UINT
+
+class ID3D11DeviceChild(_IUnknown):
+    GetDevice: Callable[[], None]
```
JelleZijlstra, line 49: Callable attributes tend to cause problems on mypy; maybe check that these behave as you expect.

JelleZijlstra (commented):
A few comments. The library doesn't import at all on Mac so I didn't play around with it myself, just looked at the stubs. It also pins an ancient version of Pillow which doesn't have 3.9 wheels, so I was only able to install d3dshot on 3.7.

Avasam:
> A few comments. The library doesn't import at all on Mac so I didn't play around with it myself, just looked at the stubs. It also pins an ancient version of Pillow which doesn't have 3.9 wheels, so I was only able to install d3dshot on 3.7.

Indeed it's specific to Windows. For the pinned pillow version issue I've just been doing `pip install D3DShot Pillow>=7.2.0` and it installs without a hitch for me on Python 3.9 and 3.10.

stubs/D3DShot/d3dshot/dll/d3d.pyi
Avasam: This is probably better as a `def`, that's just 3AM melted brain not recognizing that.

.flake8
Avasam: I figured out later what caused E704 to show up in my editor. Should be removed

stubs/D3DShot/d3dshot/capture_output.pyi
Avasam: `CaptureOutput` is right. You pass the `CaptureOutputs` enum to `__init__` and it initializes a child class of `CaptureOutput` and stores it in `backend`.

Here's the source: https://github.com/SerpentAI/D3DShot/blob/dev/d3dshot/capture_output.py#L20

So `CaptureOutput` is both a base class and a wrapper.

AlexWaygood:
I'm slightly concerned about the fact that this library appears to be archived, unmaintained, and only supports 3.6-3.8. Adding any new stubs to typeshed increases our maintenance burden somewhat; is this addition worth the cost?

Arguments in favour of adding these stubs anyway are:
- Lots of people are still on 3.6/3.8, even if these older versions are now considered unsupported or to be in "security-only mode" by CPython.
- It _looks_ like the repo was only archived relatively recently, and the issue count is pretty low.
- The repo has 257 stars in GitHub, indicating it's fairly popular (or at least was at some point).
- The fact that there won't be any new releases means we won't have to constantly update the library when stubtest starts failing, lowering the maintenance burden.

Avasam:
> I'm slightly concerned about the fact that this library appears to be archived
> It looks like the repo was only archived relatively recently

It ironically wasn't when I started this stub ... When I noticed that, I had some concerns whether it would end up accepted.

> only supports 3.6-3.8

Other than having to pin pillow manually (see above comment), I'm using it without issue on a project that's 3.9 - 3.10 only: <https://github.com/Avasam/Auto-Split>

I'll add that I believe this has benefits as it is the only decent screen recording Python library using the DirectX Desktop Duplication API. And most search engine results point back to it.  
<https://github.com/ra1nty/DXcam> is up and coming (first release to PyPI in the last week), and very promising. But I cannot yet confirm whether it is a proper replacement or not.
<https://github.com/Hellikandra/pyDXGID3D> is still in early development, unpublished and the implementation is still incomplete.

I can always make my own stubs package, but discoverability (especially for newer python users) isn't going to be the best.

Ultimately, it's up to you (the team), to see if this belongs in Typeshed.

AlexWaygood:
> Other than having to pin pillow manually (see above comment), I'm using it without issue on a project that's 3.9 - 3.10 only: https://github.com/Avasam/Auto-Split
> 
> I'll add that I believe this has benefits as it is the only decent screen recording Python library using the DirectX Desktop Duplication API. And most search engine results point back to it.

Alright, that all sounds pretty promising. No objections from me to this going in :)

pyrightconfig.stricter.json
```diff
@@ -91,7 +91,8 @@
         "stubs/tzlocal/tzlocal/utils.pyi",
         "stubs/ttkthemes",
         "stubs/urllib3",
-        "stubs/vobject"
+        "stubs/vobject",
+        "stubs/D3DShot"
```
AlexWaygood, line 95: Would be good to keep this list alphabetised :)

stubs/D3DShot/METADATA.toml
```diff
@@ -0,0 +1,5 @@
+version = "0.1.*"
+requires = ["types-Pillow"]
+
+[tool.stubtest]
+skip = true
```
AlexWaygood, line 5: `skip = true` should only be specified under exceptional circumstances (which this is) -- we should add a comment to say why we're disabling stubtest here

```suggestion
# The library only works on Windows; we currently only run stubtest on Ubuntu for third-party stubs in CI.
# See #8660
skip = true
```

stubs/D3DShot/d3dshot/capture_output.pyi
```diff
@@ -0,0 +1,36 @@
+import enum
+from _typeshed import Incomplete
+from ctypes import _CVoidConstPLike
+from typing_extensions import Literal, TypeAlias
+
+from PIL import Image
+
+_Frame: TypeAlias = Image.Image | Incomplete
+# _Frame: TypeAlias = Image.Image | npt.NDArray[np.int32] | npt.NDArray[np.float32] | _Tensor
```
AlexWaygood, line 9: Add a reference to #5768 so it shows up when #5768 is fixed and we start grepping for all the references in the codebase to that issue

```suggestion
# Fix when #5768 is resolved:
# _Frame: TypeAlias = Image.Image | npt.NDArray[np.int32] | npt.NDArray[np.float32] | _Tensor
```

stubs/D3DShot/d3dshot/d3dshot.pyi
```diff
@@ -0,0 +1,52 @@
...
+from collections import deque
+from typing import Any
+from typing_extensions import TypeAlias
+
+from d3dshot.capture_output import CaptureOutput as CaptureOutput, CaptureOutputs as CaptureOutputs
+from d3dshot.display import Display as Display
+from PIL import Image
+
+_Frame: TypeAlias = Image.Image | Incomplete
+# _Frame: TypeAlias = Image.Image | npt.NDArray[np.int32] | npt.NDArray[np.float32] | _Tensor
```
AlexWaygood, line 11:
```suggestion
# Fix when #5768 is resolved:
# _Frame: TypeAlias = Image.Image | npt.NDArray[np.int32] | npt.NDArray[np.float32] | _Tensor
```

stubs/D3DShot/d3dshot/display.pyi
```diff
@@ -0,0 +1,45 @@
+from _typeshed import Incomplete
+from collections.abc import Callable
+from ctypes.wintypes import PFLOAT
+from typing_extensions import Literal, TypeAlias
+
+from PIL import Image
+
+_Frame: TypeAlias = Image.Image | Incomplete
+# _Frame: TypeAlias = Image.Image | npt.NDArray[np.int32] | npt.NDArray[np.float32] | _Tensor
```
AlexWaygood, line 9:
```suggestion
# Fix when #5768 is resolved:
# _Frame: TypeAlias = Image.Image | npt.NDArray[np.int32] | npt.NDArray[np.float32] | _Tensor
```

stubs/D3DShot/d3dshot/dll/d3d.pyi
```diff
@@ -0,0 +1,226 @@
...
+from _typeshed import Incomplete
+from ctypes import Structure, _CArgObject, c_int32, c_uint, c_void_p
+from ctypes.wintypes import FLOAT, UINT
+from typing_extensions import TypeAlias
+
+from d3dshot.dll.dxgi import IDXGIAdapter
+
+# mypy does not support os.name checks, while pyright does https://github.com/python/mypy/issues/13002
+# import os
+# if os.name == "nt":  # noqa: Y002
```
AlexWaygood, line 11: Is there an advantage to using `os.name` instead of `sys.platform` here? If not, I think we can safely do without this comment; I'm not sure it really adds much :)

```suggestion
```

stubs/D3DShot/d3dshot/dll/dxgi.pyi
```diff
@@ -0,0 +1,174 @@
...
+# from torch import Tensor
+# import comtypes
+# import numpy.typing as npt
+# _IUnknown: TypeAlias = Incomplete
+_Frame: TypeAlias = Image.Image | Incomplete
+# _Frame: TypeAlias = Image.Image | npt.NDArray[np.int32] | npt.NDArray[np.float32] | Tensor
+
+# mypy does not support os.name checks, while pyright does https://github.com/python/mypy/issues/13002
+# import os
+# if os.name == "nt":  # noqa: Y002
```
AlexWaygood, line 22:
```suggestion
```

AlexWaygood (commented):
This isn't a proper review, just a few nits about comments etc. :)

stubs/D3DShot/d3dshot/dll/d3d.pyi
Avasam: I'm not sure. It's what `ctypes` does. So I just wanted to mirrored that to stay accurate, then realized mypy doesn't support it.

AlexWaygood: Emulating the runtime as much as possible is Good and Virtuous. But, in this case, I think it's fairly obvious why we're using `sys.platform` instead of mimicking the runtime exactly.

PEP 484 [specifically calls out `sys.platform` checks](https://peps.python.org/pep-0484/#version-and-platform-checking) as something type checkers are supposed to be able to understand. Any other way of checking for platform-specific behaviour is, from the perspective of the `typing` system, undefined behaviour. Pyright has chosen to go above and beyond the spec here in _also_ implementing support for `os.name` checks, which is awesome -- but that's not something that type checkers are _expected_ to necessarily understand.

stubs/D3DShot/d3dshot/dll/d3d.pyi
```diff
@@ -0,0 +1,228 @@
...
+        self,
+        pDstResource: _Pointer[ID3D11Resource],
+        DstSubresource: UINT,
+        DstX: UINT,
+        DstY: UINT,
+        DstZ: UINT,
+        pSrcResource: _Pointer[ID3D11Resource],
+        SrcSubresource: UINT,
+        pSrcBox: _Pointer[D3D11_BOX],
+    ) -> None: ...
```
AlexWaygood, line 112: I don't know anything about ctypes, really. It looks like these methods are all autogenerated using some ctypes metaclass magic or something.

Are these parameters definitely positional-or-keyword? Where do the parameter names come from? https://github.com/SerpentAI/D3DShot/blob/4472120a21a1cb397d4f30d7a36af993e7544bc9/d3dshot/dll/d3d.py#L108-L120

stubs/D3DShot/d3dshot/dll/dxgi.pyi
```diff
@@ -0,0 +1,180 @@
...
+def describe_dxgi_adapter(dxgi_adapter: IDXGIAdapter1) -> Array[WCHAR]: ...
+def discover_dxgi_outputs(dxgi_adapter: IDXGIAdapter) -> list[_Pointer[IDXGIOutput1]]: ...
+def describe_dxgi_output(dxgi_output: IDXGIOutput) -> _DXGIOutput: ...
+def initialize_dxgi_output_duplication(
+    dxgi_output: IDXGIOutput1, d3d_device: _Pointer[ID3D11Device]
+) -> _Pointer[IDXGIOutputDuplication]: ...
+def get_dxgi_output_duplication_frame(
+    dxgi_output_duplication: IDXGIOutputDuplication,
+    d3d_device: ID3D11Device,
+    process_func: Callable[[PFLOAT, int, int, int, _ProcessFuncRegionArg, int], _ProcessFuncReturn] | None = ...,
```
AlexWaygood, line 175: Same as my comment above where there was another function with a `process_func` argument -- I think the `process_func` callable is missing a parameter. (Also, maybe define a private type alias in `dll/__init__.pyi`, and import it in both places where there are `process_func` arguments? It's quite a complicated type :)

```suggestion
    process_func: Callable[[PFLOAT, int, int, int, int, _ProcessFuncRegionArg, int], _ProcessFuncReturn] | None = ...,
```

AlexWaygood (commented):
Okay, here's a proper review!

stubs/D3DShot/d3dshot/capture_output.pyi
```diff
@@ -0,0 +1,37 @@
...
+        pointer: _CVoidConstPLike,
+        pitch: int,
+        size: int,
+        width: int,
+        height: int,
+        region: tuple[int, int, int, int],
+        rotation: int,
+    ) -> _Frame: ...
+    def to_pil(self, frame: _Frame) -> Image.Image: ...
+    def stack(self, frames: list[_Frame], stack_dimension: Literal["first", "last"]) -> _Frame: ...
```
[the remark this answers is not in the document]
Avasam: Not really. I try to watch out for that now, but this is one of the first type stubs I filled in (although I PR'd much later). Didn't catch that when I went back to update.

stubs/D3DShot/d3dshot/d3dshot.pyi
```diff
@@ -0,0 +1,53 @@
...
+from PIL import Image
+
+_Frame: TypeAlias = Image.Image | Incomplete
+# FIXME: When #5768 is resolved:
+# _Frame: TypeAlias = Image.Image | npt.NDArray[np.int32] | npt.NDArray[np.float32] | _Tensor
+
+class Singleton(type):
+    def __call__(cls, *args: Any, **kwargs: Any) -> Any: ...
+
+class D3DShot:
```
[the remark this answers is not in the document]
Avasam: Where does that come from?

stubs/D3DShot/d3dshot/display.pyi
```diff
@@ -0,0 +1,46 @@
...
+    resolution: tuple[int, int]
+    position: dict[Literal["left", "top", "right", "bottom"], int]
+    rotation: int
+    scale_factor: float
+    is_primary: bool
+    hmonitor: int
+    dxgi_output: Incomplete | None
+    dxgi_adapter: Incomplete | None
+    d3d_device: Incomplete
+    d3d_device_context: Incomplete
```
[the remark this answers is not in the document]
Avasam: `self._initialize_dxgi_output_duplication()` is always called at the end of init. Which always sets those values.

I can also complete those types now that I've completed `d3d.pyi` and `dxgi.pyi`

stubs/D3DShot/d3dshot/dll/d3d.pyi
Avasam: Parameter names all come from Microsoft's official docs. For this specific example: https://docs.microsoft.com/en-us/windows/win32/api/d3d11/nf-d3d11-id3d11devicecontext-copysubresourceregion

stubs/D3DShot/d3dshot/dll/dxgi.pyi
Avasam: I miscounted the arguments in both you are right.
I'll do that instead of duplicating the complicated type!

stubs/D3DShot/d3dshot/capture_outputs/pil_capture_output.pyi
```diff
@@ -0,0 +1,24 @@
...
+        pointer: _CVoidConstPLike,
+        pitch: int,
+        size: int,
+        width: int,
+        height: int,
+        region: tuple[int, int, int, int],
+        rotation: int,
+    ) -> Image.Image: ...
+    def to_pil(self, frame: _ImageT) -> _ImageT: ...
+    def stack(self, frames: _ImageT, stack_dimension: _Unused) -> _ImageT: ...  # type: ignore[override]
```
AlexWaygood, line 24: Isn't `frames` supposed to be a `Sequence` here? (I believe you can also get rid of the `_Frames: TypeAlias = Image.Image | Sequence[Incomplete]` alias in the base class, and just use `Sequence[Incomplete]` as the annotation for `frames` in the base class. That should mean you can get rid of the `type: ignore` here.)

```suggestion
    def stack(self, frames: Sequence[_ImageT], stack_dimension: _Unused) -> Sequence[_ImageT]: ...
```

stubs/D3DShot/d3dshot/display.pyi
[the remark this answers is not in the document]
AlexWaygood: Okay, great -- let's add a comment to the stub explaining why they can't be `None`, in that case. It's somewhat confusing code 😄

stubs/D3DShot/d3dshot/dll/d3d.pyi
AlexWaygood: Okay, great. What about the other point, though -- are these parameters definitely positional-or-keyword? I.e., is it okay to do this?

```python
from d3dshot.dll.d3d import ID3D11DeviceContext
x = ID3D11DeviceContext()
x.CopyResource(pDstResource=<whatever>, pSrcResource=<whatever>)
```

Or can these arguments only be used positionally, i.e., you'd have to do this:

```python
from d3dshot.dll.d3d import ID3D11DeviceContext
x = ID3D11DeviceContext()
x.CopyResource(<whatever>, <whatever>)
```

stubs/D3DShot/d3dshot/capture_output.pyi
AlexWaygood: I agree with @Avasam's analysis, but adding a comment here might be good.

AlexWaygood:
Nit: Could you possibly avoid force-pushing for typeshed PRs, where possible (once people have started reviewing)? They interact badly with the GitHub UI, making it harder to see what changed between commits, thus making it harder for us to review :)

stubs/D3DShot/d3dshot/dll/__init__.pyi
```diff
@@ -0,0 +1,10 @@
+from collections.abc import Callable
+from ctypes.wintypes import PFLOAT
+from typing import TypeVar
+from typing_extensions import TypeAlias
+
+from d3dshot.capture_output import _Frame
+
+_ProcessFuncRegionArg = TypeVar("_ProcessFuncRegionArg", tuple[int, int, int, int], None)
+_ProcessFuncReturn = TypeVar("_ProcessFuncReturn", _Frame, None)
+_ProcessFunc: TypeAlias = Callable[[PFLOAT, int, int, int, int, _ProcessFuncRegionArg, int], _ProcessFuncReturn]  # noqa: Y047
```
AlexWaygood, line 10: Add a comment explaining why we're `# noqa`ing it

```suggestion
# The _ProcessFunc alias is used in multiple submodules
_ProcessFunc: TypeAlias = Callable[[PFLOAT, int, int, int, int, _ProcessFuncRegionArg, int], _ProcessFuncReturn]  # noqa: Y047
```

stubs/D3DShot/d3dshot/capture_outputs/pil_capture_output.pyi
Avasam: Humm, might make more sense. The numpy and pytorch implementations return a single item. But those would themselves be a container (ndarray and Tensor)

AlexWaygood:
> On a different note, do you know if it's possible to make `CaptureOutput` generic since `_Frame` depends on the backend? I'm less familiar with Python generic classes. Not that it really matters until #5768 anyway.

We _could_ do, and in some ways it seems like a good candidate for a generic class. _But_, as you say, it doesn't _really_ matter until #5768. And there are also some serious downsides to making classes in the stub generic if they're not generic at runtime -- for example, subclassing the class becomes very annoying (see, e.g., #7855). So, I'm inclined to say, "Let's not overcomplicate it for now, and leave it as it is."

stubs/D3DShot/d3dshot/d3dshot.pyi
```diff
@@ -0,0 +1,44 @@
...
+        numpy_is_available: bool = ...,
+        pytorch_is_available: bool = ...,
+        pytorch_gpu_is_available: bool = ...,
+    ) -> None: ...
+    @property
+    def is_capturing(self) -> bool: ...
+    def get_latest_frame(self) -> _Frame | None: ...
+    def get_frame(self, frame_index: int) -> _Frame | None: ...
+    def get_frames(self, frame_indices: list[int]) -> list[_Frame]: ...
+    def get_frame_stack(self, frame_indices: list[int], stack_dimension: str | None = ...) -> _Frame: ...
```
AlexWaygood, line 31: `list` invariance strikes again :)

```suggestion
    def get_frames(self, frame_indices: Iterable[int]) -> list[_Frame]: ...
    def get_frame_stack(self, frame_indices: Iterable[int], stack_dimension: str | None = ...) -> _Frame: ...
```

stubs/D3DShot/d3dshot/capture_output.pyi
```diff
@@ -0,0 +1,42 @@
...
+    NUMPY_FLOAT: int
+    PYTORCH: int
+    PYTORCH_FLOAT: int
+    PYTORCH_GPU: int
+    PYTORCH_FLOAT_GPU: int
+
+class CaptureOutputError(BaseException): ...
+
+class CaptureOutput:
+    backend: CaptureOutput
```
AlexWaygood, line 29: Would be great to add a comment here explaining what you stated in https://github.com/python/typeshed/pull/8652#discussion_r963191042

stubs/D3DShot/d3dshot/display.pyi
```diff
@@ -0,0 +1,49 @@
...
+
+class _PositionDict(TypedDict):
+    left: int
+    top: int
+    right: int
+    bottom: int
+
+# Note that Display.d3d_device and Display.d3d_device_context can never be None.
+# Despite initially being set to None in __init__,
+# they're always immediatly set in _initialize_dxgi_output_duplication()
```
AlexWaygood, line 16: Extremely picky nitpick: this comment is quite a long way from the attributes it's commenting on; I'd prefer it if it was immediately above those attributes. Also, spelling -- should be "immediately", not "immediatly" :)

AlexWaygood (commented):
We're very nearly there, I think. Thanks for all your work on this, this is a really tough one!

stubs/D3DShot/d3dshot/d3dshot.pyi
Avasam: I think that's the last one

stubs/D3DShot/d3dshot/capture_output.pyi
Avasam: Done. Hopefully it's clear enough.

AlexWaygood: LGTM!

stubs/D3DShot/d3dshot/display.pyi
```diff
@@ -0,0 +1,49 @@
...
+    position: _PositionDict
+    rotation: int
+    scale_factor: float
+    is_primary: bool
+    hmonitor: int
+    dxgi_output: IDXGIOutput1 | None
+    dxgi_adapter: _Pointer[IDXGIAdapter] | None
+    # Note that Display.d3d_device and Display.d3d_device_context can never be None.
+    # Despite initially being set to None in __init__,
+    # they're always immediatley set in _initialize_dxgi_output_duplication()
```
AlexWaygood, line 27:
```suggestion
    # they're always immediately set in _initialize_dxgi_output_duplication()
```

AlexWaygood (approved):
Thanks for your patience! Let's land this.

PR #8653, opened 2022-08-31
`create_baseline_stubs.py`: Fix path separator on windows and ignore root `/out`

Avasam:
On windows, the stub path is generated with a backslash, which of course is not valid when added to the json file.  
`create_baseline_stubs.py` also creates a `out` folder at the root which was not gitignored.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -56,6 +56,7 @@ target/
 
 # Local utility scripts
 analyze.py
+/out/
 
 # Editor backup files
 *~
--- a/.gitignore
+++ b/.gitignore
@@ -56,7 +56,6 @@ target/
 
 # Local utility scripts
 analyze.py
-/out/
 
 # Editor backup files
 *~
--- a/scripts/create_baseline_stubs.py
+++ b/scripts/create_baseline_stubs.py
@@ -167,7 +167,7 @@ def main() -> None:
         sys.exit(1)
     project, version = info
 
-    stub_dir = os.path.join("stubs", project)
+    stub_dir = os.path.join("stubs", project).replace("\\", "/")
     if os.path.exists(stub_dir):
         sys.exit(f"Error: {stub_dir} already exists (delete it first)")
 
--- a/scripts/create_baseline_stubs.py
+++ b/scripts/create_baseline_stubs.py
@@ -90,7 +90,8 @@ def add_pyright_exclusion(stub_dir: str) -> None:
     assert i < len(lines), f"Error parsing {PYRIGHT_CONFIG}"
     while not lines[i].strip().startswith("]"):
         i += 1
-    line_to_add = f'        "{stub_dir}",'
+    # Must use forward slash in the .json file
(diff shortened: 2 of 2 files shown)


scripts/create_baseline_stubs.py
```diff
@@ -167,7 +167,7 @@ def main() -> None:
         sys.exit(1)
     project, version = info
 
-    stub_dir = os.path.join("stubs", project)
+    stub_dir = os.path.join("stubs", project).replace("\\", "/")
```
AlexWaygood, line 170: Can't we just do this?

```suggestion
    stub_dir = f"stubs/{project}"
```

.gitignore
```diff
@@ -56,6 +56,7 @@ target/
 
 # Local utility scripts
 analyze.py
+/out/
```
AlexWaygood, line 59: This feels like we're working around the problem rather than fixing the problem. Wouldn't a better solution be to have `create_baseline_stubs.py` do a better job at cleaning up after the stubs have been generated?

You can change stubgen's output directory by using the `--output` command-line option. I think a better solution to this problem would be to change `create_baseline_stubs.py` so that stubgen outputs the generated stubs directly into `stub_dir`, instead of outputting them into an `out` directory and then copying them across from `out` to `stub_dir`.

(I'm happy to do that fix in a separate PR if you want to limit this PR to just the `pyrightconfig.stricter.json` fix :)

AlexWaygood (commented):
Thanks! I remember being bitten by the path-separator thing as well when I used `create_baseline_stubs.py` a while back on my Windows machine. Can't remember why I didn't file a PR to fix it.

scripts/create_baseline_stubs.py
Akuli: This looks like someone will "refactor"/"clean"/"fix" it to use `os.path.join()` later.

```suggestion
    stub_dir = f"stubs/{project}"  # Must use forward slash in the .json file
```

.gitignore
Avasam: It was an easy enough change

AlexWaygood:
> Maybe we should do a `replace` in `add_pyright_exclusion` and let everything else do normal os.path stuff?

SGTM, though everything seems to work fine on my Windows machine either way :)

Avasam:
> > Maybe we should do a `replace` in `add_pyright_exclusion` and let everything else do normal os.path stuff?
> 
> SGTM, though everything seems to work fine on my Windows machine either way :)

Everything works in windows / powershell for me as well  with either slashes (well except setting the environment variable MYPYPATH since the syntax is different in powershell, but that's unrelated to this work)

AlexWaygood:
> Only changing the slashes in `add_pyright_exclusion` might be more consistent in code, instead of having a single exception to using `path.join`

Yes, let's do that

AlexWaygood (approved):
LGTM!

hauntsaninja (approved):

Issue #8472, opened 2022-08-02
Preferring explicit `Incomplete` for unannotated arguments


hauntsaninja:
This came up in https://github.com/python/typeshed/pull/8458

Currently CONTRIBUTING.md states:

> Included functions and methods must list all arguments, but the arguments can be left unannotated. Do not use Any to mark unannotated arguments or return values.

Leaving things unannotated helps distinguish from cases where we actually need `Any`. But we didn't have `Incomplete` when that was written. `Incomplete` accomplishes the same thing and is easier to grep for, e.g. https://github.com/python/typeshed/issues/4641 mentions grepping for incomplete as a good way to find potential areas to contribute.

hauntsaninja:
Oh, interesting. Agreed that we should prioritise type checker results over usability for ourselves.

That said, there's a quick and easy hack that seems to get us best of both worlds:
```
diff --git a/stdlib/_typeshed/__init__.pyi b/stdlib/_typeshed/__init__.pyi
index 89ca9d81..3a4840b1 100644
--- a/stdlib/_typeshed/__init__.pyi
+++ b/stdlib/_typeshed/__init__.pyi
@@ -34,7 +34,7 @@ AnyStr_co = TypeVar("AnyStr_co", str, bytes, covariant=True)  # noqa: Y001
 # isn't possible or a type is already partially known. In cases like these,
 # use Incomplete instead of Any as a marker. For example, use
 # "Incomplete | None" instead of "Any | None".
-Incomplete: TypeAlias = Any
+from ._does_not_exist import Incomplete as Incomplete  # type: ignore
 
 # stable
 class IdentityFunction(Protocol):
```

<img width="617" alt="Screen Shot 2022-08-02 at 1 12 49 PM" src="https://user-images.githubusercontent.com/12621235/182464669-c0d30f2f-7cf7-4b82-890a-6270bab39a33.png">

Alternatively, we could maybe ask Eric to bless `Incomplete`

Akuli:
Mypy has different behaviour for unannotated arguments too, although it's more complicated:
- If anything at all in a function is annotated (some argument or the return type), the function is "typed" and everything unannotated behaves as if it was `Any`.
- If nothing is annotated, arguments behave as if they were `Any` typed, but the function is also marked as being "untyped", and using it at all will generates errors by default if you use `--strict`.

srittau:
Apart from typechecker issues, I have a slight preference for unannotated items:

[...]

PR #8654, opened 2022-08-31
Added pyautogui stubs

Avasam:


Avasam:
I don't understand the 1 failing test

stubs/PyAutoGUI/pyautogui/__init__.pyi
```diff
@@ -0,0 +1,233 @@
...
+class ImageNotFoundException(PyAutoGUIException): ...
+
+collectionsSequence = collections.abc.Sequence
+
+_P = ParamSpec("_P")
+_R = TypeVar("_R")
+
+# TODO: Complete types with pyscreeze once we can import non-types dependencies
+# See: https://github.com/python/typeshed/issues/5768
+
```
AlexWaygood, line 21: `pyscreeze` isn't currently a `py.typed` package, so resolving #5768 wouldn't actually help us here.

```suggestion
```

stubs/PyAutoGUI/pyautogui/__init__.pyi
```diff
@@ -0,0 +1,233 @@
...
+class _Box(NamedTuple):
```
AlexWaygood, line 22: This looks correct, but it would be good to add a comment here explaining where `_Box` is actually defined. I had to do some hunting to find the definition!

```suggestion
# Actually `pyscreeze.Box`, but typeshed doesn't currently have stubs for pyscreeze
# (and the library doesn't have type annotations either)
class _Box(NamedTuple):
```

stubs/PyAutoGUI/pyautogui/__init__.pyi
```diff
@@ -0,0 +1,233 @@
...
+    width: int
+    height: int
+
+def raisePyAutoGUIImageNotFoundException(wrappedFunction: Callable[_P, _R]) -> Callable[_P, _R]: ...
+def locate(*args: Incomplete, **kwargs: Incomplete) -> _Box | None: ...
+def locateAll(*args: Incomplete, **kwargs: Incomplete) -> Generator[_Box, None, None]: ...
+def locateAllOnScreen(*args: Incomplete, **kwargs: Incomplete) -> Generator[_Box, None, None]: ...
+def locateCenterOnScreen(*args: Incomplete, **kwargs: Incomplete) -> Point | None: ...
+def locateOnScreen(*args: Incomplete, **kwargs: Incomplete) -> _Box | None: ...
+def locateOnWindow(*args: Incomplete, **kwargs: Incomplete) -> _Box | None: ...
```
AlexWaygood, line 34: For parameters where we don't know (or haven't yet figured out) what the annotations should be, we prefer to leave them unannotated rather than using `_typeshed.Incomplete`. See #8472 for some recent discussion as to why.

(This only applies for parameter annotations where the type is completely unknown. For some types, we might know that the type "could be `None`", and in that situation, we might use `Incomplete | None`. But for completely unknown types, we like to leave it unannotated.)

```suggestion
def locate(*args, **kwargs) -> _Box | None: ...
def locateAll(*args, **kwargs) -> Generator[_Box, None, None]: ...
def locateAllOnScreen(*args, **kwargs) -> Generator[_Box, None, None]: ...
def locateCenterOnScreen(*args, **kwargs) -> Point | None: ...
def locateOnScreen(*args, **kwargs) -> _Box | None: ...
def locateOnWindow(*args, **kwargs) -> _Box | None: ...
```

stubs/PyAutoGUI/pyautogui/__init__.pyi
```diff
@@ -0,0 +1,233 @@
...
+class Size(NamedTuple):
+    width: int
+    height: int
+
+def getPointOnLine(x1: float, y1: float, x2: float, y2: float, n: float) -> tuple[float, float]: ...
+def linear(n: float) -> float: ...
+def position(x: int | None = ..., y: int | None = ...) -> Point: ...
+def size() -> Size: ...
+@overload
+def onScreen(xy: tuple[float, float]) -> bool: ...
```
AlexWaygood, line 73: The function does still, technically, work if you explicitly pass `None` in as the second argument:

```suggestion
def onScreen(x: tuple[float, float], y: None = ...) -> bool: ...
```

AlexWaygood (commented):
Thank you! Here's a partial review -- I've got up to line 75 in `__init__.pyi`, will carry on later :)

pyrightconfig.stricter.json
```diff
@@ -90,7 +90,8 @@
         "stubs/tzlocal/tzlocal/utils.pyi",
         "stubs/ttkthemes",
         "stubs/urllib3",
-        "stubs/vobject"
+        "stubs/vobject",
+        "stubs/PyAutoGUI"
```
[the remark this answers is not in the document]
AlexWaygood: > We could edit the stub generating script to keep it alphabetized.

I like that idea :)

Avasam:
I ran stubtest locally so it could tell me about missing class members. However, just like with my D3DShot PR,  I'm not sure how to deal with pointer types. ie `dwExtraInfo: ctypes.POINTER(ctypes.wintypes.ULONG)`

pyrightconfig.stricter.json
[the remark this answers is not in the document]
AlexWaygood: (Actually, it looks like the stub-generating script already has logic in it to keep the list alphabetised. I'm guessing the only reason it wasn't initially alphabetised here is because of the path-separator issue being discussed in #8653.)

stubs/PyAutoGUI/pyautogui/__init__.pyi
```diff
@@ -0,0 +1,230 @@
...
+def getPointOnLine(x1: float, y1: float, x2: float, y2: float, n: float) -> tuple[float, float]: ...
+def linear(n: float) -> float: ...
+def position(x: int | None = ..., y: int | None = ...) -> Point: ...
+def size() -> Size: ...
+@overload
+def onScreen(point: tuple[float, float], _: None = ...) -> bool: ...
+@overload
+def onScreen(x: float, y: float) -> bool: ...
+def mouseDown(
+    x: float | Sequence[int] | None = ...,
```
AlexWaygood, line 74: I think we can be slightly more permissive here; `int()` is called on all elements of the sequence in `_normalizeXYArgs()` if a sequence is passed, so we can allow `Sequence[float]` instead of `Sequence[int]`. The docstring states that `x` can also be a `str`, as well: https://github.com/asweigart/pyautogui/blob/667161b072e89b4a25b806c5f2ca616e6c0aeec3/pyautogui/__init__.py#L895.

```suggestion
    x: float | str | Sequence[float] | None = ...,
```

stubs/PyAutoGUI/pyautogui/__init__.pyi
```diff
@@ -0,0 +1,230 @@
...
+    y: float | Sequence[int] | None = ...,
```
AlexWaygood, line 75: According to the docstring, it's okay to pass a sequence of values in for the `x` argument, but not for the `y` argument: https://github.com/asweigart/pyautogui/blob/667161b072e89b4a25b806c5f2ca616e6c0aeec3/pyautogui/__init__.py#L892-L899. That's also backed up by what happens in the `_normalizeXYArgs()` function.

```suggestion
    y: float | None = ...,
```

stubs/PyAutoGUI/pyautogui/__init__.pyi
```diff
@@ -0,0 +1,230 @@
...
+    button: str = ...,
```
AlexWaygood, line 76: The docstring states that this can also be an `int`: https://github.com/asweigart/pyautogui/blob/667161b072e89b4a25b806c5f2ca616e6c0aeec3/pyautogui/__init__.py#L899. But looking at the `_normalizeButton()` function, that doesn't seem to be right, since `.lower()` is called on the button passed to that function unconditionally: https://github.com/asweigart/pyautogui/blob/667161b072e89b4a25b806c5f2ca616e6c0aeec3/pyautogui/__init__.py#L851. So, let's add a comment:

```suggestion
    # Docstring says `button` can also be `int`, but `.lower()` is called unconditionally in `_normalizeButton()`
    button: str = ...,
```

stubs/PyAutoGUI/pyautogui/__init__.pyi
```diff
@@ -0,0 +1,230 @@
...
+    mouseDownUp: bool = ...,
+) -> None: ...
+
+drag = dragRel
+
+def isValidKey(key: str) -> bool: ...
+def keyDown(key: str, logScreenshot: bool | None = ..., _pause: bool = ...) -> None: ...
+def keyUp(key: str, logScreenshot: bool | None = ..., _pause: bool = ...) -> None: ...
+def press(
+    keys: str | list[str], presses: int = ..., interval: float = ..., logScreenshot: bool | None = ..., _pause: bool = ...
```
AlexWaygood, line 216: The docstring says that `keys` has to be `str` or `list[str]`, but it can actually be any iterable of `str`s. And `list[str]` is quite problematic for parameter annotations due to `list` being invariant.

```suggestion
    keys: str | Iterable[str], presses: int = ..., interval: float = ..., logScreenshot: bool | None = ..., _pause: bool = ...
```

stubs/PyAutoGUI/pyautogui/__init__.pyi
```diff
@@ -0,0 +1,230 @@
...
+) -> None: ...
+def hold(keys: str | list[str], logScreenshot: bool | None = ..., _pause: bool = ...) -> Generator[None, None, None]: ...
+def typewrite(message: str | list[str], interval: float = ..., logScreenshot: bool | None = ..., _pause: bool = ...) -> None: ...
+
+write = typewrite
+
+def hotkey(*args: str, **kwargs: float | bool | None) -> None: ...
```
AlexWaygood, line 223: At runtime, the signature is `*args, **kwargs`, but if we do `*args, **kwargs` in the stub, type checkers won't emit an error if somebody has a typo in their code like `logScreenshot=Falsse`. The only kwargs that would actually be useful here are `logScreenshot` and `interval`:

```suggestion
def hotkey(*args: str, logScreenshot: bool | None = ..., interval: float = ...) -> None: ...
```

stubs/PyAutoGUI/pyautogui/__init__.pyi
```diff
@@ -0,0 +1,230 @@
...
+def failSafeCheck() -> None: ...
+def displayMousePosition(xOffset: float = ..., yOffset: float = ...) -> None: ...
+def sleep(seconds: int) -> None: ...
```
AlexWaygood, line 226:
```suggestion
def sleep(seconds: float) -> None: ...
```

AlexWaygood (commented):
Here's a review for the rest of the PR!

As well as the remarks inline here: we usually only add stubs for a package's public API, unless there's a good reason to add stubs for implementation details. That means we usually omit stubs for private modules like `_pyautogui_osx.pyi`, `_pyautogui_java.pyi`, `_pyautogui_win.pyi` and `_pyautogui_x11.pyi`. Would you mind removing those? (Add an allowlist entry if stubtest complains!)

The remarks I made in relation to the `mouseDown()` signature also apply to `mouseUp()`, `click()`, `leftClick()`, `rightClick()`, `middleClick()`, `doubleClick()`, `tripleClick()`, `scroll()`, `hscroll()`, `vscroll()`, `moveTo()`, `moveRel()`, `dragTo()` and `dragRel()`.

The remarks I made in relation to `press()` also apply to `hold()` and `typewrite()`.

stubs/PyAutoGUI/pyautogui/__init__.pyi
Avasam: I wasn't sure if I should be expanding the kwargs or not. Good to know in cases of specifically known kwargs I'll define them!

stubs/PyAutoGUI/pyautogui/__init__.pyi
```diff
@@ -0,0 +1,230 @@
...
+class Size(NamedTuple):
+    width: int
+    height: int
+
+def getPointOnLine(x1: float, y1: float, x2: float, y2: float, n: float) -> tuple[float, float]: ...
+def linear(n: float) -> float: ...
+def position(x: int | None = ..., y: int | None = ...) -> Point: ...
+def size() -> Size: ...
+@overload
+def onScreen(point: tuple[float, float], _: None = ...) -> bool: ...
```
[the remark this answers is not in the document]
Avasam: Oh that makes sense. Typescript habits 😄

stubs/PyAutoGUI/pyautogui/__init__.pyi
Avasam: Good catch, I totally missed that x could be a string >.<

stubs/PyAutoGUI/pyautogui/__init__.pyi
```diff
@@ -0,0 +1,241 @@
...
+
+drag = dragRel
+
+def isValidKey(key: str) -> bool: ...
+def keyDown(key: str, logScreenshot: bool | None = ..., _pause: bool = ...) -> None: ...
+def keyUp(key: str, logScreenshot: bool | None = ..., _pause: bool = ...) -> None: ...
+def press(
+    keys: str | Iterable[str], presses: int = ..., interval: float = ..., logScreenshot: bool | None = ..., _pause: bool = ...
+) -> None: ...
+def hold(keys: str | Iterable[str], logScreenshot: bool | None = ..., _pause: bool = ...) -> Generator[None, None, None]: ...
```
AlexWaygood, line 227: For this one, the `@contextmanager` decorator changes the signature slightly:

```suggestion
def hold(keys: str | Iterable[str], logScreenshot: bool | None = ..., _pause: bool = ...) -> contextlib._GeneratorContextManager[None]: ...
```

stubs/PyAutoGUI/pyautogui/__init__.pyi
```diff
@@ -0,0 +1,241 @@
+from collections.abc import Callable, Generator, Iterable, Sequence
```
AlexWaygood, line 1:
```suggestion
import contextlib
from collections.abc import Callable, Generator, Iterable, Sequence
```

AlexWaygood (approved):
Thank you!

PR #8656, opened 2022-08-31
Bump flake8-pyi and pytype test dependencies

AlexWaygood:

PR #8657, opened 2022-08-31
In `ctypes/__init__.pyi`, `RTLD_GLOBAL` and `RTLD_LOCAL` are imported from `_ctypes.pyi`

junkmd:
Thanks to google/pytype#1281 and google/pytype#1282, importing `RTLD_GLOBAL` and `RTLD_LOCAL` from `_ctypes` will not fail the `pytype` test on CI if the `pytype==2022.08.30` version is specified.

[Other stuffs that should be defined in `_ctypes`](https://github.com/python/typeshed/issues/8633#issuecomment-1231672437) will be migrated after `pytype` stops using its own `_ctypes` stubs and starts using `typeshed`'s `_ctypes` stubs.

- related to #8571, #8633 and #8643

### DIFF
--- a/stdlib/ctypes/__init__.pyi
+++ b/stdlib/ctypes/__init__.pyi
@@ -5,9 +5,7 @@ from collections.abc import Callable, Iterable, Iterator, Mapping, Sequence
 from typing import Any, ClassVar, Generic, TypeVar, Union as _UnionT, overload
 from typing_extensions import TypeAlias
 
-# TODO: import these from _ctypes once it no longer breaks pytype
-RTLD_GLOBAL: int
-RTLD_LOCAL: int
+from _ctypes import RTLD_GLOBAL as RTLD_GLOBAL, RTLD_LOCAL as RTLD_LOCAL
 
 if sys.version_info >= (3, 9):
     from types import GenericAlias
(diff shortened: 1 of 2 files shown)


AlexWaygood:
Thank you! Unfortunately you were slightly too slow here — I filed #8656 just five minutes before you filed this. Sorry about that :)

junkmd:
@AlexWaygood 
Don't mind!

It was not a big change, so it was not a problem.

stdlib/ctypes/__init__.pyi
```diff
@@ -5,9 +5,7 @@ from collections.abc import Callable, Iterable, Iterator, Mapping, Sequence
 from typing import Any, ClassVar, Generic, TypeVar, Union as _UnionT, overload
 from typing_extensions import TypeAlias
 
-# TODO: import these from _ctypes once it no longer breaks pytype
-RTLD_GLOBAL: int
-RTLD_LOCAL: int
+from _ctypes import RTLD_GLOBAL as RTLD_GLOBAL, RTLD_LOCAL as RTLD_LOCAL
```
Akuli, line 8: I wonder why `isort` didn't move this import near the top of the file.

Akuli, line 8: I wonder why `isort` didn't move this import near the top of the file.

AlexWaygood: Take a look at the second commit in #8656 — there's quite a few private stdlib modules that we have in typeshed that, by default, isort doesn't recognise as being part of the stdlib

PR #8658, opened 2022-08-31
Update `importlib.resources` types

layday:


stdlib/importlib/abc.pyi
```diff
@@ -93,9 +85,9 @@ class FileLoader(ResourceLoader, ExecutionLoader, metaclass=ABCMeta):
 
 class ResourceReader(metaclass=ABCMeta):
     @abstractmethod
-    def open_resource(self, resource: StrOrBytesPath) -> IO[bytes]: ...
+    def open_resource(self, resource: str) -> IO[bytes]: ...
```
AlexWaygood, line 88: [The documentation states](https://docs.python.org/3.11/library/importlib.resources.abc.html#importlib.resources.abc.ResourceReader):

> For any of methods of this class, a resource argument is expected to be a [path-like object](https://docs.python.org/3.11/glossary.html#term-path-like-object) which represents conceptually just a file name.

That would indicate that the methods on this class can indeed accept pathlike objects, and that it's CPython/`importlib_resources` that have incorrect annotations for this class.

stdlib/importlib/abc.pyi
```diff
@@ -115,8 +107,12 @@ if sys.version_info >= (3, 9):
...
         @abstractmethod
         def iterdir(self) -> Iterator[Traversable]: ...
-        @abstractmethod
-        def joinpath(self, child: StrPath) -> Traversable: ...
+        if sys.version_info >= (3, 11):
+            @abstractmethod
+            def joinpath(self, *descendants: str) -> Traversable: ...
+        else:
+            @abstractmethod
+            def joinpath(self, child: str) -> Traversable: ...
```
AlexWaygood, line 115: The annotation for this parameter in CPython is `StrPath` -- why should we do differently here? https://github.com/python/cpython/blob/d4d5e605cd74b198a6e21d5bf8fa053de2069aae/Lib/importlib/resources/abc.py#L99

```suggestion
        if sys.version_info >= (3, 11):
            @abstractmethod
            def joinpath(self, *descendants: StrPath) -> Traversable: ...
        else:
            @abstractmethod
            def joinpath(self, child: StrPath) -> Traversable: ...
```

layday: The annotation in CPython was copied from the typeshed.  A filesystem path is not valid as an abstract resource.  The whole point of `importlib.resources` is to disassociate package resources from the filesystem so that you might be able to load "files" from e.g. memory, a zip file, over the network, an SQL database, or who knows what.  Accepting `os.PathLike` here makes as much sense as defining module names in terms of filesystem paths.

stdlib/importlib/abc.pyi
layday: First of all, if this method should accept bytes, then `TraversableResources` violates the Liskov substitution principle.

The documentation goes on to say:

> This means that no subdirectory paths should be included in the resource argument.

Which is very slightly odd. This method accepts path-like objects but only if they do not represent paths at all. I think it is safer to conceal this misfeature from the typing user.

AlexWaygood: > First of all, if this method should accept bytes, then `TraversableResources` violates the Liskov substitution principle.

Well, it would be _far_ from the first class in the stdlib to do so :)

AlexWaygood: It does seem like the annotations and docstring for `importlib_resources` are completely different, and always have been: https://github.com/python/importlib_resources/blob/main/importlib_resources/abc.py.

AlexWaygood (commented):
Well, I sorta feel like I'm the wrong reviewer for this, since I've never used this module. I was hoping another maintainer might chime in :)

JelleZijlstra (approved):

AlexWaygood:
Seems like the CPython docs are in serious need of some improvement here, if you fancy taking that on :) 

the references to "path-like objects" are... _very_ confusing

PR #8661, opened 2022-09-01
redis: Improve `kwargs` typing in `sentinel`

sobolevn:
1. `ConnectionPool` super-class has better typing of `kwargs`: https://github.com/python/typeshed/blob/master/stubs/redis/redis/connection.pyi#L195 So, I removed it from a child
2. Other changes are trivial: `kwargs` are always `dict[str, *]`

### DIFF
--- a/stubs/redis/redis/sentinel.pyi
+++ b/stubs/redis/redis/sentinel.pyi
@@ -27,7 +27,6 @@ class SentinelManagedSSLConnection(SentinelManagedConnection, SSLConnection): ..
 class SentinelConnectionPool(ConnectionPool):
     is_master: bool
     check_connection: bool
-    connection_kwargs: Any
     service_name: str
     sentinel_manager: Sentinel
     def __init__(self, service_name: str, sentinel_manager: Sentinel, **kwargs) -> None: ...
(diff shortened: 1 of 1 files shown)


stubs/redis/redis/sentinel.pyi
```diff
@@ -37,10 +36,10 @@ class SentinelConnectionPool(ConnectionPool):
     def rotate_slaves(self) -> Iterator[_AddressAndPort]: ...
 
 class Sentinel(SentinelCommands):
-    sentinel_kwargs: Any
+    sentinel_kwargs: dict[str, Any | None]
```
AlexWaygood, line 39: Why `Any | None` here, rather than just `Any`?

sobolevn: To keep it in sync with `__init__`: https://github.com/python/typeshed/pull/8661/files#diff-e6904f5cf832e0aa2d8b88f35c960d621184111abe36685bc4c4f53cb8398c82R47

AlexWaygood: I think we should change the annotation in `__init__` to `dict[str, Any] | None` rather than `Any | None`.

And, if `None` is passed into `__init__`, the `sentinel_kwargs` attribute is set to a different dictionary: https://github.com/redis/redis-py/blob/0b27e3ebdfd7e7dd67222bcb043fcdfcc907bd42/redis/sentinel.py#L179-L182. So the `sentinel_kwargs` attribute can never be `None`. So I think this should just be `dict[str, Any]`

```suggestion
    sentinel_kwargs: dict[str, Any]
```

sobolevn: Good catch.

AlexWaygood (approved):
Thanks! :)

PR #8663, opened 2022-09-01
typed-ast: fix kw_defaults arg

hauntsaninja:
Following the redundant-expr change, mypy self check is broken when run
using Python 3.8 (mypy CI runs self check with 3.7)


AlexWaygood:
This PR has 0 lines changed relative to typeshed master :)

JelleZijlstra:
Yes, we already did this in #8645.

AlexWaygood:
You might need to run `pip install -r test-requirements.txt` again locally in your mypy virtual environment — I bumped the `types-typed-ast` requirement in https://github.com/python/mypy/pull/13547

hauntsaninja:
Sounds like this PR is good to merge then?

PR #8666, opened 2022-09-02
Added type stub for keyboard

Avasam:


stubs/keyboard/keyboard/__init__.pyi
```diff
@@ -0,0 +1,169 @@
...
+version: str
+
+class _Event(_UninterruptibleEvent):
+    def wait(self) -> None: ...  # type: ignore[override]  # Actual implementation
+
+def is_modifier(key: _Key | None) -> bool: ...
+
+class _KeyboardListener(_GenericListener):
+    transition_table: dict[
+        tuple[Literal["free"], Literal["up"], Literal["modifier"]]
```
JelleZijlstra, line 29: This is a little excessive. How do we help users by duplicating this whole table at the type level?

stubs/keyboard/keyboard/__init__.pyi
```diff
@@ -0,0 +1,169 @@
...
+
+press_and_release = send
+
+def press(hotkey: _ParseableHotkey) -> None: ...
+def release(hotkey: _ParseableHotkey) -> None: ...
+
+# is_pressed cannot check multi-step hotkeys, so not using _ParseableHotkey
+
+def is_pressed(hotkey: _Key | _ScanCodeList) -> bool: ...
+def call_later(fn: Callable[..., None], args: _P = ..., delay: float = ...) -> None: ...
```
JelleZijlstra, line 91: This should be expressible with PEP 646's TypeVarTuple, though we can't use that yet because mypy doesn't support it.

stubs/keyboard/keyboard/__init__.pyi
```diff
@@ -0,0 +1,169 @@
...
+def wait(hotkey: _ParseableHotkey | None = ..., suppress: bool = ..., trigger_on_release: bool = ...) -> None: ...
+def get_hotkey_name(names: Iterable[str] | None = ...) -> str: ...
+def read_event(suppress: bool = ...) -> KeyboardEvent: ...
+def read_key(suppress: bool = ...) -> _Key: ...
+def read_hotkey(suppress: bool = ...) -> str: ...
+def get_typed_strings(events: Iterable[KeyboardEvent], allow_backspace: bool = ...) -> Generator[str, None, None]: ...
+def start_recording(
+    recorded_events_queue: Queue[KeyboardEvent] | None = ...,
+) -> tuple[Queue[KeyboardEvent], Callable[[], None]]: ...
+def stop_recording() -> list[deque[KeyboardEvent]]: ...
```
JelleZijlstra, line 153: I think it's just a list of KeyboardEvents.

stubs/keyboard/keyboard/_canonical_names.pyi
```diff
@@ -0,0 +1,6 @@
+basestring = str
```
JelleZijlstra, line 1: I would remove this, it's clearly not meant as a part of the API.

stubs/keyboard/keyboard/_generic.pyi
```diff
@@ -0,0 +1,23 @@
...
+from threading import Lock, Thread
+from typing_extensions import Literal, TypeAlias
+
+from ._keyboard_event import KeyboardEvent
+from ._mouse_event import _MouseEvent
+
+_Event: TypeAlias = KeyboardEvent | _MouseEvent
+
+class GenericListener:
+    lock: Lock
```
JelleZijlstra, line 12:
```suggestion
    lock: ClassVar[Lock]
```

stubs/keyboard/keyboard/mouse.pyi
```diff
@@ -0,0 +1,76 @@
...
+    def listen(self) -> None: ...
+
+def is_pressed(button: _MouseButton = ...): ...
+def press(button: _MouseButton = ...) -> None: ...
+def release(button: _MouseButton = ...) -> None: ...
+def click(button: _MouseButton = ...) -> None: ...
+def double_click(button: _MouseButton = ...) -> None: ...
+def right_click() -> None: ...
+def wheel(delta: int = ...) -> None: ...
+def move(x: int | c_long, y: int | c_long, absolute: bool = ..., duration: float = ...) -> None: ...
```
JelleZijlstra, line 43:
```suggestion
def move(x: SupportsInt, y: SupportsInt, absolute: bool = ..., duration: float = ...) -> None: ...
```
It calls `int()` on x and y. Strictly a few more types are allowed: https://github.com/python/typeshed/blob/8e4b89a707f2508e4938e0691b714e235d1719fd/stdlib/builtins.pyi#L200

stubs/keyboard/keyboard/mouse.pyi
```diff
@@ -0,0 +1,76 @@
...
+def on_middle_click(callback: Callable[..., None], args: _P = ...) -> _Callback: ...
+def wait(button: _MouseButton = ..., target_types: tuple[_MouseEventType] = ...) -> None: ...
+
+if sys.platform == "win32":
+    def get_position() -> tuple[c_long, c_long]: ...
+
+else:
+    def get_position() -> tuple[int, int]: ...
+
+def hook(callback: _Callback) -> _Callback: ...
```
JelleZijlstra, line 63: Could use a bound TypeVar to signify that the exact same type is returned.

JelleZijlstra (commented):
Thanks, left some feedback

stubs/keyboard/keyboard/__init__.pyi
Avasam: Good to know. I hadn't looked into 3.11's variadic generics yet. That's cool. So this would be dependant on https://github.com/python/mypy/issues/12280

stubs/keyboard/keyboard/__init__.pyi
Avasam: You're right, I got confused into thinking it was a list of deque, but it's just converting

stubs/keyboard/keyboard/mouse.pyi
Avasam: Didn't know about that one. Thanks!

JelleZijlstra (approved):

PR #8668, opened 2022-09-02
Removed outdated mypy exceptions

### LINKED ISSUE #8667: flake8 Y022 conflicts with CONTRIBUTING.md doc for arbitrary-length tuples
> use built-in generics (list, dict, tuple, set), instead of importing them from typing, except in type aliases, in base classes, and for arbitrary length tuples (Tuple[int, ...]);

Is the documentation correct? If so, should `Y022` be disabled? Or maybe the doc can mention to ignore this specific error with a `# noqa: Y022` comment when it is relevant to do so.

Avasam:
Closes #8667 . Please do validate this is accurate.


srittau (approved):
Thanks!

PR #8669, opened 2022-09-02
add type hint support for unittest.loader.TestLoader._match_path private method

### LINKED ISSUE #8664: Missing method `TestLoader._match_path` from `unittest.loader.pyi`
The `unittest.loader.TestLoader._match_path` method and other private methods of `TestLoader` are missing from [the unittest.loader stubs](https://github.com/python/typeshed/blob/master/stdlib/unittest/loader.pyi#L14-L25).

While most of the missing private methods are not meant to be overriden, `_match_path` specify in a comment that it might be overidden.

``` python
def _match_path(self, path, full_path, pattern):
  # override this method to use alternative matching strategy
  return fnmatch(path, pattern)
```
When doing so, and calling `super()._match_path` in the child class, mypy will raise a false positive typing error.

It feels that this method should be reflected in the stub.

priandey:
Fixes  #8664


srittau (approved):
Thanks!

PR #8671, opened 2022-09-02
fix: paramiko.win_pageant.pyi stubs

kkirsche:
This fixes the stubs of paramiko's win_pageant.pyi file.

Source code:
https://github.com/paramiko/paramiko/blob/main/paramiko/win_pageant.py

1. `can_talk_to_agent` forces a `bool` return
2. `ULONG_PTR` is used in `COPYDATASTRUCT`'s `_fields_`.
3. `recv` returns `""` if `_response` is `None`, so technically it returns either `str` or `bytes` based on the response slice.


stubs/paramiko/paramiko/win_pageant.pyi
```diff
@@ -3,12 +3,14 @@ import sys
 
 if sys.platform == "win32":
     win32con_WM_COPYDATA: int
-    def can_talk_to_agent(): ...
+    def can_talk_to_agent() -> bool: ...
+
+    ULONG_PTR: ctypes.c_uint64 | ctypes.c_unit32
```
AlexWaygood, line 8:
```suggestion
    ULONG_PTR: ctypes.c_uint64 | ctypes.c_uint32
```

kkirsche: Oops, sorry about that.

AlexWaygood: No need to be sorry, happens to us all :)

stubs/paramiko/paramiko/win_pageant.pyi
```diff
@@ -3,12 +3,14 @@ import sys
...
+
+    ULONG_PTR: ctypes.c_uint64 | ctypes.c_uint32
 
     class COPYDATASTRUCT(ctypes.Structure): ...
 
     class PageantConnection:
         def __init__(self) -> None: ...
         def send(self, data: bytes) -> None: ...
-        def recv(self, n: int) -> bytes: ...
+        def recv(self, n: int) -> str | bytes: ...
```
Akuli, line 15: Presumably the best way to check the return value is `if foo == ""`, rather than `if isinstance(foo, str)`.

```suggestion
        def recv(self, n: int) -> Literal[""] | bytes: ...
```

You will also need to add an import.

stubs/paramiko/paramiko/win_pageant.pyi
```diff
@@ -3,12 +3,14 @@ import sys
 
 if sys.platform == "win32":
     win32con_WM_COPYDATA: int
-    def can_talk_to_agent(): ...
+    def can_talk_to_agent() -> bool: ...
+
+    ULONG_PTR: ctypes.c_uint64 | ctypes.c_uint32
```
JelleZijlstra, line 8: This is a type alias for either uint64 or uint32, not an instance.

stubs/paramiko/paramiko/win_pageant.pyi
JelleZijlstra: I wonder if their code at https://github.com/paramiko/paramiko/blob/1824a27c644132e5d46f2294c1e2fa131c523559/paramiko/win_pageant.py#L136 has a bug and should compare against `b""` instead.

stubs/paramiko/paramiko/win_pageant.pyi
```diff
@@ -1,14 +1,18 @@
...
 import sys
+from typing import Literal
+from typing_extensions import TypeAlias
 
 if sys.platform == "win32":
     win32con_WM_COPYDATA: int
-    def can_talk_to_agent(): ...
+    def can_talk_to_agent() -> bool: ...
+
+    ULONG_PTR: TypeAlias = ctypes.c_uint64 | ctypes.c_unit32
```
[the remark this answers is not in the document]
kkirsche: Oops, not that again. Must have came back when I rebased my local against Github. Sorry about that.

AlexWaygood (approved):

PR #8673, opened 2022-09-02
Fix jsonschema._format.pyi to match upstream

kkirsche:
This updates the types in jsonschema._format.pyi to match the upstream source's type hints:

https://github.com/python-jsonschema/jsonschema/blob/main/jsonschema/_format.py

JSONSchema's typeshed version is 4.15.*:
https://github.com/python/typeshed/blob/master/stubs/jsonschema/METADATA.toml

which was released two days ago:
https://github.com/python-jsonschema/jsonschema/releases/tag/v4.15.0

So the use of master as a reference is not a big concern here.

### DIFF
--- a/stubs/jsonschema/@tests/stubtest_allowlist.txt
+++ b/stubs/jsonschema/@tests/stubtest_allowlist.txt
@@ -1,14 +1,2 @@
-jsonschema._format.is_css21_color
-jsonschema._format.is_css_color_code
-jsonschema._format.is_datetime
-jsonschema._format.is_duration
-jsonschema._format.is_host_name
-jsonschema._format.is_idn_host_name
-jsonschema._format.is_iri
-jsonschema._format.is_iri_reference
-jsonschema._format.is_json_pointer
-jsonschema._format.is_relative_json_pointer
-jsonschema._format.is_time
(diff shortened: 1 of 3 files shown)


AlexWaygood:
The stubtest error is a stubtest bug that's already fixed on mypy master. You'll just need to allowlist it for now, we can remove the allowlist entry as soon as mypy 0.980 is out.

kkirsche:
I've added a fix, but also installed https://github.com/python-jsonschema/jsonschema/blob/main/pyproject.toml#L46 as an extras so that theoretically the other `is_*` calls should not require the stubtest_allowlist. I'll revert that if needed. This was determined by looking at the comments in the stubs which indicate that some of the `is_*` functions are only available when specific dependencies are installed.

kkirsche:
> The stubtest error is a stubtest bug that's already fixed on mypy master. You'll just need to allowlist it for now, we can remove the allowlist entry as soon as mypy 0.980 is out.

Seems like that Mypy behavior ended up being a net positive by reducing the usage of stubtest_allowlist for this package :) Thanks for the feedback on that

AlexWaygood (approved):
Thanks!

PR #8675, opened 2022-09-03
Add more undocumented constants to gzip.pyi

Gobot1234:
https://github.com/python/cpython/blob/837ce6460d6554e9d7a847951ca5108ac6926ee5/Lib/gzip.py#L16

### DIFF
--- a/stdlib/gzip.pyi
+++ b/stdlib/gzip.pyi
@@ -24,7 +24,6 @@ FEXTRA: Literal[4]  # undocumented
 FNAME: Literal[8]  # undocumented
 FCOMMENT: Literal[16]  # undocumented
 
-
 class _ReadableFileobj(Protocol):
     def read(self, __n: int) -> bytes: ...
     def seek(self, __n: int) -> Any: ...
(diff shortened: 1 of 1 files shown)


AlexWaygood (commented):
Would you mind making the new additions `int`, instead of `Literal`s? We agreed to a general pause on using `Literal` for global constants, as it caused false positives for some type checkers. See:

- #7258

stdlib/gzip.pyi
```diff
@@ -15,8 +15,14 @@ _ReadBinaryMode: TypeAlias = Literal["r", "rb"]
 _WriteBinaryMode: TypeAlias = Literal["a", "ab", "w", "wb", "x", "xb"]
 _OpenTextMode: TypeAlias = Literal["rt", "at", "wt", "xt"]
 
-READ: Literal[1]
-WRITE: Literal[2]
+READ: int  # actually Literal[1] # undocumented
+WRITE: int  # actually Literal[2] # undocumented
```
AlexWaygood, line 19: ...But at the same time, I'd rather leave the ones that are _already in the stub_ as they are, unless they're causing a specific problem for somebody :)

```suggestion
READ: Literal[1]  # undocumented
WRITE: Literal[2]  # undocumented
```

AlexWaygood (approved):

PR #8678, opened 2022-09-03
`pyautogui`: just skip stubtest

AlexWaygood:
I realised immediately after merging the pyautogui PR that there's not much point running stubtest on this package, since stubtest isn't able to import any of PyAutoGUI without access to a display screen

### DIFF
--- a/stubs/PyAutoGUI/@tests/stubtest_allowlist.txt
+++ b/stubs/PyAutoGUI/@tests/stubtest_allowlist.txt
@@ -1,3 +1,0 @@
-# pyautogui requires a display, resulting in the following error on the CI:
-# failed to import, KeyError: 'DISPLAY'
-pyautogui
--- a/stubs/PyAutoGUI/METADATA.toml
+++ b/stubs/PyAutoGUI/METADATA.toml
@@ -1,4 +1,6 @@
 version = "0.9.*"
 
 [tool.stubtest]
-ignore_missing_stub = false
+# pyautogui requires a display, resulting in the following error on the CI:
+# failed to import, KeyError: 'DISPLAY'
+skip = true


JelleZijlstra (approved):
I looked into whether we could fudge this with environment variables, but it really does need an actual display it can talk to at import time.

(Also, remind me not to try to install pyautogui on a Mac, it installs like 100 dependencies.)

Akuli:
We could use `xvfb-run` if we want an actual display. I do it for the tests of my GUI programs, e.g. https://github.com/Akuli/porcupine/blob/b4e3e0a0015af661bb4974b1ab200854ce638126/.github/workflows/check.yml#L55-L57

AlexWaygood:
> We could use `xvfb-run` if we want an actual display. I do it for the tests of my GUI programs, e.g. https://github.com/Akuli/porcupine/blob/b4e3e0a0015af661bb4974b1ab200854ce638126/.github/workflows/check.yml#L55-L57

Interesting. Would this slow down running stubtest at all? Would there be a way of specifying in `METADATA.toml` whether a package needs a display, and then only using `xvfb` for packages that need it?

Akuli:
It would slow starting stubtest, but only by about 300ms:

```
$ time xvfb-run sleep 1

real	0m1,341s
user	0m0,019s
sys	0m0,010s
```

Once stubtest is running, it has a `DISPLAY` environment variable but otherwise runs as usual.

AlexWaygood:
Sounds great to me, in that case!

AlexWaygood:
We could maybe unskip `pynput` as well, if we did this: https://github.com/python/typeshed/blob/8e4b89a707f2508e4938e0691b714e235d1719fd/stubs/pynput/METADATA.toml#L4

AlexWaygood:
@Akuli, could you work up a PR for this? (GitHub Actions are... not my strong suit.)

Akuli:
Sure. I'll probably get it done during this weekend :)

PR #8680, opened 2022-09-04
`mypy_test.py`: Use pathlib more

AlexWaygood:


tests/mypy_test.py
```diff
@@ -150,22 +151,22 @@ def parse_version(v_str: str) -> tuple[int, int]:
     return int(m.group(1)), int(m.group(2))
 
 
-def add_files(files: list[str], seen: set[str], root: str, name: str, args: TestConfig) -> None:
+def add_files(files: list[Path], seen: set[str], root: StrPath, name: str, args: TestConfig) -> None:
     """Add all files in package or module represented by 'name' located in 'root'."""
-    full = os.path.join(root, name)
-    mod, ext = os.path.splitext(name)
+    full = Path(root, name)
+    mod, ext = full.stem, full.suffix
```
JelleZijlstra, line 157: You can probably just get rid of these variables.

tests/mypy_test.py
```diff
@@ -150,22 +151,22 @@ def parse_version(v_str: str) -> tuple[int, int]:
...
     if ext in [".pyi", ".py"]:
         if match(full, args):
             seen.add(mod)
             files.append(full)
-    elif os.path.isfile(os.path.join(full, "__init__.pyi")) or os.path.isfile(os.path.join(full, "__init__.py")):
+    elif (full / "__init__.pyi").is_file() or (full / "__init__.py").is_file():
         for r, ds, fs in os.walk(full):
             ds.sort()
```
JelleZijlstra, line 164: Seems like we're not using this, so might as well remove it.

tests/mypy_test.py
```diff
@@ -150,22 +151,22 @@ def parse_version(v_str: str) -> tuple[int, int]:
...
-    mod, ext = os.path.splitext(name)
+    full = Path(root, name)
+    mod, ext = full.stem, full.suffix
     if ext in [".pyi", ".py"]:
         if match(full, args):
             seen.add(mod)
             files.append(full)
-    elif os.path.isfile(os.path.join(full, "__init__.pyi")) or os.path.isfile(os.path.join(full, "__init__.py")):
+    elif (full / "__init__.pyi").is_file() or (full / "__init__.py").is_file():
         for r, ds, fs in os.walk(full):
```
JelleZijlstra, line 163: Maybe we can replace this whole loop with something using `full.glob("**/*.{py,pyi}")`.

AlexWaygood: I'm... actually not quite sure why we're collecting `.py` files here anyway. This part of `mypy_test.py` should probably only care about `.pyi` files!

tests/mypy_test.py
```diff
@@ -151,25 +151,15 @@ def parse_version(v_str: str) -> tuple[int, int]:
...
-                if x in [".pyi", ".py"]:
-                    fn = Path(r, f)
-                    if match(fn, args):
-                        seen.add(mod)
-                        files.append(fn)
+    if module.is_file() and module.suffix == ".pyi":
+        files.append(module)
+        seen.add(module.stem)
+    else:
+        to_add = sorted(module.rglob("*.pyi"))
```
JelleZijlstra, line 160: I believe you need `**/*.pyi`; this will not recurse into subdirectories.

AlexWaygood: I think it will, because we're using `rglob`, not `glob` :) https://docs.python.org/3/library/pathlib.html#pathlib.Path.rglob

I checked locally, and it collects the same number of files (for both the stdlib and the third-party packages) as we currently do in CI

AlexWaygood: I did some more testing -- I'm pretty confident that the code proposed in this PR collects exactly the same list of files as the code on typeshed master. The order of the `files` list appears to be slightly different, but personally I don't think that matters much?

srittau:
Not a review, but +1 on using pathlib here. I had the same thought when looking at the previous PR. This also makes it more clear where a "path" is used and where a "path-like string" is used.

tests/mypy_test.py
JelleZijlstra: Thanks for confirming!

PR #8683, opened 2022-09-04
Enable pyright's `reportUnnecessaryTypeIgnoreComment` setting for the whole `test_cases` directory

### LINKED ISSUE #8506: Figure out how to enable pyright's `reportUnnecessaryTypeIgnoreComment` setting for the whole `test_cases` directory
Currently, we have mypy's `--warn-unused-ignore` setting enabled for the whole `test_cases` directory, but for pyright, we have to manually add `# pyright: reportUnnecessaryTypeIgnoreComment=true` at the top of each file in the directory. These settings are crucial to these tests working properly; none of tests against false-negatives work properly without them.

It would be good if we could work out how to get `# pyright: reportUnnecessaryTypeIgnoreComment=true` enabled for the whole directory. The current system is error-prone: if we forget to add it at the top of a file, any tests against false negatives could fail silently.

AlexWaygood:
Fixes #8506

### DIFF
--- a/test_cases/stdlib/builtins/test_pow.py
+++ b/test_cases/stdlib/builtins/test_pow.py
@@ -1,5 +1,3 @@
-# pyright: reportUnnecessaryTypeIgnoreComment=true
-
 from decimal import Decimal
 from fractions import Fraction
 from typing import Any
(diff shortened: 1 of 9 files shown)


test_cases/stdlib/builtins/test_pow.py
```diff
@@ -16,7 +14,7 @@
 #
 # assert_type(pow(2, 4, 0), NoReturn)
 
-assert_type(pow(2, 4), int)
+assert_type(pow(2, 4), int)  # type: ignore
```
AlexWaygood, line 17:
```suggestion
assert_type(pow(2, 4), int)
```

PR #8684, opened 2022-09-05
Completed pyautogui stubs

Avasam:
Simplified some functions arguments by using SupportsInt, SupportsFloat, SupportsIndex
Typed args and kwargs without depending on pyscreeze


stubs/PyAutoGUI/pyautogui/__init__.pyi
```diff
@@ -1,37 +1,156 @@
...
+# These functions reuse pyscreeze functions directly.
+# TODO: Once pyscreeze is typed, we can alias them to simplify this stub
+
+# _locateAll_opencv
+@overload
+def locate(
+    needleImage,
+    haystackImage,
+    grayscale: bool | None = ...,
+    limit: object | None = ...,
```
Avasam, line 39: This kwarg is always overriden. What's best to do here? `object | None`, `Any`, or just omit it?

AlexWaygood:
Does this PR mean that all parameters in our stubs for `pyautogui` are now annotated? If so, can you remove this line as part of this PR, so that we start testing the stubs with the strictest pyright settings? :) https://github.com/python/typeshed/blob/3a397a225217cc84b2a81b4073ad06a1c3b0c7e3/pyrightconfig.stricter.json#L72

Avasam:
> Does this PR mean that all parameters in our stubs for `pyautogui` are now annotated? If so, can you remove this line as part of this PR, so that we start testing the stubs with the strictest pyright settings? :)
> 
> https://github.com/python/typeshed/blob/3a397a225217cc84b2a81b4073ad06a1c3b0c7e3/pyrightconfig.stricter.json#L72

That's the goal! So I've removed it from the excludes.

stubs/PyAutoGUI/pyautogui/__init__.pyi
Avasam: Answered by https://github.com/python/typeshed/pull/8652#discussion_r963151335
Just assign `object`

stubs/PyAutoGUI/pyautogui/__init__.pyi
```diff
@@ -1,37 +1,177 @@
...
+    region: _Box | None = ...,
+    step: int = ...,
+    confidence: None = ...,
+) -> _Box | None: ...
+
+# _locateAll_opencv
+@overload
+def locateAll(
+    needleImage: str | _Mat,
+    haystackImage: str | _Mat,
```
AlexWaygood, line 62: These can both be `PIL` images -- `_locateAll_opencv` passes the arguments to `_load_cv2`: https://github.com/asweigart/pyscreeze/blob/b693ca9b2c964988a7e924a52f73e15db38511a8/pyscreeze/__init__.py#L201-L203

And `_load_cv2` explicitly accounts for the possibility that these might be `PIL` images: https://github.com/asweigart/pyscreeze/blob/b693ca9b2c964988a7e924a52f73e15db38511a8/pyscreeze/__init__.py#L173-L174

The same point applies for all the other overloaded functions that pass arguments to `locateAll()`

```suggestion
    needleImage: str | Image.Image | _Mat,
    haystackImage: str | Image.Image | _Mat,
```

stubs/PyAutoGUI/pyautogui/__init__.pyi
```diff
@@ -1,37 +1,177 @@
...
+# _locateAll_python / _locateAll_pillow
+@overload
+def locateCenterOnScreen(
+    image: str | Image.Image,
+    minSearchTime: float,
+    grayscale: bool | None = ...,
+    limit: _Unused = ...,
+    region: _Box | None = ...,
+    step: int = ...,
+    confidence: None = ...,
```
AlexWaygood, line 125: ? Should this not be `SupportsFloat`, like all the other `confidence` parameters?

stubs/PyAutoGUI/pyautogui/__init__.pyi
```diff
@@ -1,37 +1,177 @@
...
-# from pyscreeze import Box
+from PIL import Image
 
 class PyAutoGUIException(Exception): ...
 class FailSafeException(PyAutoGUIException): ...
 class ImageNotFoundException(PyAutoGUIException): ...
 
 _P = ParamSpec("_P")
 _R = TypeVar("_R")
+_NormalizeableXArg: TypeAlias = SupportsInt | Sequence[SupportsInt]
```
AlexWaygood, line 16: We covered this in the last PR :)

The `_normalizeXYArgs` fnction explicitly accounts for the fact that the `x` argument might be a string: https://github.com/asweigart/pyautogui/blob/667161b072e89b4a25b806c5f2ca616e6c0aeec3/pyautogui/__init__.py#L664

Unfortunately, `SupportsInt` means "anything that has an `__int__` method, not `anything that can be passed to the `int()` constructor -- hence why we need to explicitly include `str` in this union.

```suggestion
_NormalizeableXArg: TypeAlias = str | SupportsInt | Sequence[SupportsInt]
```

stubs/PyAutoGUI/pyautogui/__init__.pyi
```diff
@@ -72,10 +212,10 @@ def size() -> Size: ...
 @overload
 def onScreen(x: tuple[float, float], y: None = ...) -> bool: ...
 @overload
-def onScreen(x: float, y: float) -> bool: ...
+def onScreen(x: _NormalizeableXArg | None, y: SupportsInt | None) -> bool: ...
```
AlexWaygood, line 215: Hmm, not sure it makes so much sense to use the `_NormalizeableXArg` alias here, since in this case, this is an overloaded function, and this is the second overload. If both `x` and `y` are provided, and `y` is not `None`, then `x` cannot be a `str` or a `Sequence`.

I would just use `SupportsInt` for both arguments here, for the second overload.

AlexWaygood (commented):
Mostly looks good! Just a few things to fix:

stubs/PyAutoGUI/pyautogui/__init__.pyi
Avasam: Totally missed that thanks! So it applies to all the opencv overloads since they all end up calling locateAll (even locate)

stubs/PyAutoGUI/pyautogui/__init__.pyi
Avasam: Not for the `_locateAll_python` implementation. Confidence being anything else than `None` will raise an error https://github.com/asweigart/pyscreeze/blob/b693ca9b2c964988a7e924a52f73e15db38511a8/pyscreeze/__init__.py#L246

stubs/PyAutoGUI/pyautogui/__init__.pyi
Avasam: Looking back at this. This overload just seems wrong.
First it should be more flexible than tuples, even the docstring suggest it's meant to be useable with a list.
Second, it's directly calling `_normalizeXYArgs`, so why not just do like everywhere else and just have `x: _NormalizeableXArg | None, y: SupportsInt | None` ?

Only thing is, the docstring doesn't suggest this is meant to be used to find whether a window is on screen (by passing a string as the first arg), but I could totally see users do that.

Due to this incertainty, I've asked the question on pyautogui, hopefully this'll get an answer: https://github.com/asweigart/pyautogui/issues/729

In the mean time, I'd heir on the side of caution (and simplicity), following current implementation. And just do:
`def onScreen(x: _NormalizeableXArg | None, y: SupportsInt | None) -> bool: ...` without any overload.

stubs/PyAutoGUI/pyautogui/__init__.pyi
AlexWaygood: Got it, thanks!

stubs/PyAutoGUI/pyautogui/__init__.pyi
AlexWaygood: The new signature you've given it looks fine to me -- we just need to make the `y` argument optional:

```suggestion
def onScreen(x: _NormalizeableXArg | None, y: SupportsInt | None = ...) -> bool: ...
```

AlexWaygood (approved):
Thank you!

PR #8685, opened 2022-09-05
`__isabstractmethod__` is a class-level variable

sobolevn:
Source: https://github.com/python/cpython/blob/a9d58feccfd956dc99195af6872b06446738d7db/Lib/abc.py#L41

### DIFF
--- a/stdlib/abc.pyi
+++ b/stdlib/abc.pyi
@@ -1,7 +1,7 @@
 import sys
 from _typeshed import Self, SupportsWrite
 from collections.abc import Callable
-from typing import Any, Generic, TypeVar
+from typing import Any, ClassVar, Generic, TypeVar
 from typing_extensions import Literal
 
 _T = TypeVar("_T")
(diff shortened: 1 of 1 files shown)


sobolevn:
I am not sure what to do here 🤔

srittau:
I think you just need to update `stdlib/abc.pyi:40` as well.

sobolevn:
@srittau this is about https://github.com/python/typeshed/blob/b6d28acb2368cdd8c87554e01e22e134061997d6/stdlib/builtins.pyi#L1179

Why am I hesitant? Because of how `property.__isabstractmethod__` is defined:

```python
>>> property
<class 'property'>
>>> property.__isabstractmethod__
<attribute '__isabstractmethod__' of 'property' objects>
```

AlexWaygood:
`builtins.property.__isabstractmethod__` should definitely be an instance attribute — most properties are concrete, but you can have abstract properties as well. The same goes for `builtins.classmethod` and `builtins.staticmethod`.

Tbh, I'm not entirely sure whether it's necessary to make any change to the classes in `abc.pyi` here. Is it correct to think of `abc.abstractproperty.__isabstractmethod__` as a `ClassVar`? Or is it more accurate to think of it as an instance attribute that has a class-level default?

I'm inclined to say the latter — I'm not sure it ever makes sense to access `__isabstractmethod__` directly from the class in the case of `abc.abstractproperty` (or indeed with the other two classes).

sobolevn:
This turned out to be more complex than I initially thought. I don't see any potential gain in changing more types. Ok, let it be!

PR #8687, opened 2022-09-05
[bs4] More precise types for bs4.Tag.select(_one)

Gobot1234:
These methods annoy pyright without the annotations


JelleZijlstra:
We prefer to leave parameters for which we don't have precise types yet unannotated. What is the pyright error you're seeing?

Gobot1234:
Oh I didn't realise they were unannotated but, I'm seeing 
```
Type of "select" is partially unknown
  Type of "select" is "(selector: str, namespaces: Any | None = ..., limit: int | None = ..., **kwargs: Unknown) -> ResultSet[Tag]"Pylance[reportUnknownMemberType]
```
I'm happy to update this PR with the correct types which I think are
```py
def select(self, selector: str, namespaces: Any | None = ..., limit: int | None = ..., flags: int = 0, *, custom: dict[str, str] | None = ...) -> ResultSet[Tag]: ...
def select_one(self, selector: str, namespaces: Any | None = ..., flags: int = 0, *, custom: dict[str, str] | None = ...) -> ResultSet[Tag]: ...
```

JelleZijlstra:
Precise types would be even better!

stubs/beautifulsoup4/bs4/element.pyi
```diff
@@ -329,8 +329,8 @@ class Tag(PageElement):
     def children(self) -> Iterable[PageElement]: ...
     @property
     def descendants(self) -> Iterable[PageElement]: ...
-    def select_one(self, selector: str, namespaces: Any | None = ..., **kwargs: Any) -> Tag | None: ...
-    def select(self, selector: str, namespaces: Any | None = ..., limit: int | None = ..., **kwargs: Any) -> ResultSet[Tag]: ...
+    def select_one(self, selector: str, namespaces: Any | None = ..., flags: int = 0, *, custom: dict[str, str] | None = ...) -> Tag | None: ...
+    def select(self, selector: str, namespaces: Any | None = ..., limit: int | None = ..., flags: int = 0, *, custom: dict[str, str] | None = ...) -> ResultSet[Tag]: ...
```
JelleZijlstra:
```suggestion
    def select(self, selector: str, namespaces: Any | None = ..., limit: int | None = ..., *, flags: int = ..., *, custom: dict[str, str] | None = ...) -> ResultSet[Tag]: ...
```

`flags` is also kwonly, and defaults should always be `...`.

stubs/beautifulsoup4/bs4/element.pyi
```diff
@@ -329,8 +329,8 @@ class Tag(PageElement):
     def children(self) -> Iterable[PageElement]: ...
     @property
     def descendants(self) -> Iterable[PageElement]: ...
-    def select_one(self, selector: str, namespaces: Any | None = ..., **kwargs: Any) -> Tag | None: ...
-    def select(self, selector: str, namespaces: Any | None = ..., limit: int | None = ..., **kwargs: Any) -> ResultSet[Tag]: ...
+    def select_one(self, selector: str, namespaces: Any | None = ..., flags: int = 0, *, custom: dict[str, str] | None = ...) -> Tag | None: ...
```
JelleZijlstra:
```suggestion
    def select_one(self, selector: str, namespaces: Any | None = ..., *, flags: int = ..., custom: dict[str, str] | None = ...) -> Tag | None: ...
```

JelleZijlstra (commented):
In theory it accepts additional kwargs: they get passed to `soupsieve.compile`, which takes `**kwargs` and ignores them. It seems more useful for type checkers to warn about additional kwargs.

JelleZijlstra (approved):

Gobot1234:
> In theory it accepts additional kwargs: they get passed to `soupsieve.compile`, which takes `**kwargs` and ignores them. It seems more useful for type checkers to warn about additional kwargs, so I think it's fine to specify only the kwargs that actually do something.

Yeah, I wasn't entirely sure what to do about that but I agree and since it doesn't affect the runtime I don't think it matters anyway.

PR #8688, opened 2022-09-05
Rename the files in the `test_cases` directory; add further clarifications to the README

AlexWaygood:
As noted by @Akuli in https://github.com/python/typeshed/pull/8604#issuecomment-1226148929, using the `test_` prefix for the files in the `test_cases` directory is actually quite confusing, as this is the convention commonly used for tests that are executed using some kind of test runner, such as pytest or unittest, and that's not what these files are actually for. Nonetheless, it's good to have some kind of consistent naming convention, so that the file with the test cases for `codecs` can't be confused with the actual stub for `codecs` in the `stdlib` directory. As such, this PR proposed renaming the files in the `test_cases` directory so that they use a `check_` prefix instead of a `test_` prefix.

I've also tried to reword the README a little:

- I've tried to more consistently use the term "test _cases_" rather than "_tests_", since, as noted above, these aren't really _tests_ in the conventional sense of the term.
- I've tried to clarify that the code in this directory is never actually executed; it's whether the code type-checks or not that really matters.

@kkirsche, I'd appreciate your feedback on my edits to the README, if you have the time :)


test_cases/README.md
```diff
@@ -1,31 +1,35 @@
 ## Regression tests for typeshed
 
-This directory contains regression tests for the stubs found elsewhere in the
-typeshed repo.
+This directory contains code samples that act as a regression test for the
+standard-library stubs found elsewhere in the typeshed repo.
```
srittau, line 4:
```suggestion
standard library stubs found elsewhere in the typeshed repo.
```

test_cases/README.md
```diff
@@ -1,31 +1,35 @@
...
+Different test cases in this directory serve different purposes. For some stubs in
 typeshed, the type annotations are complex enough that it's useful to have
 basic sanity checks that test whether a type checker understands the intent of
 the annotations correctly. Examples of tests like these are
 `stdlib/builtins/test_pow.py` and `stdlib/asyncio/test_gather.py`.
 
-Other tests, such as the tests for `ExitStack` in `stdlib/test_contextlib.py`
-and the tests for `LogRecord` in `stdlib/test_logging.py`, do not relate to
+Other test cases, such as the samples for `ExitStack` in `stdlib/test_contextlib.py`
+and the samples for `LogRecord` in `stdlib/test_logging.py`, do not relate to
```
srittau, line 21: `test_` -> `check_`

Might be others in the file.

test_cases/README.md
```diff
@@ -1,31 +1,35 @@
...
 stubs where the annotations are particularly complex, but they *do* relate to
 stubs where decisions have been taken that might be slightly unusual. These
-tests serve a different purpose: to check that type checkers do not emit
+test cases serve a different purpose: to check that type checkers do not emit
 false-positive errors for idiomatic usage of these classes.
 
 ### How the tests work
 
+None of the code in this directory is actually executed in typeshed's CI.
```
srittau, line 29: I think this sentence is a bit confusing. Maybe something along the lines of "The code in this directory is not meant to be executed."

srittau (changes requested):

test_cases/README.md
AlexWaygood: Whoops, yeah, there's a few. Thanks.

test_cases/README.md
AlexWaygood: Not sure about this one. "Standard library" is a compound noun here; without the hyphen, you could misread it as "library stubs that are standard" rather than "stubs for the standard library".

test_cases/README.md
```diff
@@ -1,31 +1,35 @@
...
+and the samples for `LogRecord` in `stdlib/check_logging.py`, do not relate to
 stubs where the annotations are particularly complex, but they *do* relate to
 stubs where decisions have been taken that might be slightly unusual. These
-tests serve a different purpose: to check that type checkers do not emit
+test cases serve a different purpose: to check that type checkers do not emit
 false-positive errors for idiomatic usage of these classes.
 
 ### How the tests work
 
+The code in this directory are not intended to be directly executed. Instead,
```
AlexWaygood, line 29:
```suggestion
The code in this directory is not intended to be directly executed. Instead,
```

test_cases/README.md
srittau: Well, it's called "The Python Standard Library" and it's consistently spelled "standard library" in the Python docs.

AlexWaygood: Meh, we can agree to disagree; I'll drop the hyphen

srittau (approved):

sobolevn (approved):

test_cases/README.md
```diff
@@ -1,31 +1,35 @@
 ## Regression tests for typeshed
 
-This directory contains regression tests for the stubs found elsewhere in the
-typeshed repo.
+This directory contains code samples that act as a regression test for the
+standard library stubs found elsewhere in the typeshed repo.
 
-**This directory should *only* contain tests for functions and classes which
+**This directory should *only* contain test cases for functions and classes which
 are known to have caused problems in the past, where the stubs are difficult to
```
kkirsche, line 7: Is there, or should there be, a way to map the tests to when they've caused issues in the past (e.g. the GitHub issue or merge request(s))? Not something that would make sense to retroactively hunt down, but may be useful when including new test cases.

test_cases/README.md
```diff
@@ -1,31 +1,35 @@
...
 get right.** 100% test coverage for typeshed is neither necessary nor
 desirable, as it would lead to code duplication. Moreover, typeshed has
 multiple other mechanisms for spotting errors in the stubs.
 
 ### The purpose of these tests
 
-Different tests in this directory serve different purposes. For some stubs in
+Different test cases in this directory serve different purposes. For some stubs in
 typeshed, the type annotations are complex enough that it's useful to have
 basic sanity checks that test whether a type checker understands the intent of
```
kkirsche, line 16: > basic sanity checks

I'd remove `basic` from this sentence. While "Sanity Check" is not in Merriam Webster's dictionary, Wikipedia says on [Sanity Check](https://en.wikipedia.org/wiki/Sanity_check)'s page that a sanity check is "...a basic test to quickly evaluate", which would make `basic` a redundant adjective.

test_cases/README.md
```diff
@@ -1,31 +1,35 @@
...
 the annotations correctly. Examples of tests like these are
-`stdlib/builtins/test_pow.py` and `stdlib/asyncio/test_gather.py`.
+`stdlib/builtins/check_pow.py` and `stdlib/asyncio/check_gather.py`.
 
-Other tests, such as the tests for `ExitStack` in `stdlib/test_contextlib.py`
-and the tests for `LogRecord` in `stdlib/test_logging.py`, do not relate to
+Other test cases, such as the samples for `ExitStack` in `stdlib/check_contextlib.py`
+and the samples for `LogRecord` in `stdlib/check_logging.py`, do not relate to
```
kkirsche, line 21: I think this section flows very nicely now, great word and file name choice.

test_cases/README.md
```diff
@@ -1,31 +1,35 @@
...
 stubs where the annotations are particularly complex, but they *do* relate to
 stubs where decisions have been taken that might be slightly unusual. These
-tests serve a different purpose: to check that type checkers do not emit
+test cases serve a different purpose: to check that type checkers do not emit
 false-positive errors for idiomatic usage of these classes.
 
 ### How the tests work
 
+The code in this directory is not intended to be directly executed. Instead,
+type checkers are run on the code, in order to check whether typing errors are
```
kkirsche, line 30: Is the intent to check for where errors are emitted or is it to verify the type checker's behavior both when errors are omitted but also when they are not? I'm not sure if the distinction matters, but wanted to ask in case it is a useful distinction. If it's superfluous though, feel free to just resolve this.

AlexWaygood: It's both; and that's probably worth clarifying, thanks.

test_cases/README.md
kkirsche: The reason I ask was not due to quantity of relationships, but accessibility, primarily. I don't have a problem with this if everyone is happy with it, but I think it's worth considering if being more deliberate about using this information to group test cases and provide feedback in test failures about what behavior regressed could be a useful longer term goal.

test_cases/README.md
AlexWaygood: I think I'll change this wording to `to check whether typing errors are emitted at the correct places`. It's slightly vaguer, but it's clarified in the following paragraphs, and it's more accurate.

test_cases/README.md
AlexWaygood: That's an interesting point, for sure. There are things we could consider there (like maybe using an actual test runner, such as pytest) -- but I do also value simplicity of implementation, and readability. So there's trade-offs here: making the output from the test more readable might make the implementation of the test more complex and less readable.

One thing we could consider: `assert_type` is a very new function. Perhaps we could make it more user-friendly: maybe there could be an optional `msg` argument that the type-checker could report if the "static assertion" failed.

Issue #8378, opened 2022-07-24
Typing the ast.AST subclass constructors


multimeric:
We currently have static types for the fields of most (all?) `ast` classes, but none of these have typed constructors, e.g.:
https://github.com/python/typeshed/blob/62cde013659b4714b6eefe14cecb4b6752b33570/stdlib/_ast.pyi#L79-L86

I think technically this might be because none of these subclasses actually have a unique constructor, but that shouldn't stop us from *typing* each of the constructors, since in practise the constructor arguments must correspond to the class fields.

Before I do this, though, I'm firstly wondering if an easy solution here would be to apply the [`dataclass_transform`](https://peps.python.org/pep-0681/) decorator (note: not the same as the `dataclass` decorator). This would simply tell the type checker that all the class fields can and should be provided in the constructor, which is broadly correct. However I don't have a deep understanding of how the `ast.AST` constructor works, so this might not be the correct behaviour. For `ClassDef` this might look like:
```python
from typing_extensions import dataclass_transform

@dataclass_transform
class ClassDef(stmt): 
    if sys.version_info >= (3, 10): 
        __match_args__ = ("name", "bases", "keywords", "body", "decorator_list") 
    name: _Identifier 
    bases: list[expr] 
    keywords: list[keyword] 
    body: list[stmt] 
    decorator_list: list[expr] 
```

If this isn't sufficient, I propose that we simply add an `__init__()` stub to each subclass. For instance, for the `ClassDef` above, this might look like:

```python
class ClassDef(stmt): 
    if sys.version_info >= (3, 10): 
        __match_args__ = ("name", "bases", "keywords", "body", "decorator_list") 
    name: _Identifier 
    bases: list[expr] 
    keywords: list[keyword] 
    body: list[stmt] 
    decorator_list: list[expr] 

    def __init__(
        name: _Identifier 
        bases: list[expr] 
        keywords: list[keyword] 
        body: list[stmt] 
        decorator_list: list[expr] 
    ):
        ...
```

hauntsaninja:
I'd be in favour of adding `__init__` stubs to each subclass. PR welcome!

multimeric:
Thoughts on the `dataclass_transform` idea?

hauntsaninja:
I would prefer explicit `__init__`s, since typically typeshed prefers to mirror closely what's happening at the runtime, rather than getting fancy and lying. Also not sure how well supported PEP 681 is by type checkers — last I checked mypy doesn't yet support it.
[...]

PR #8690, opened 2022-09-06
_ast: Change ast.ImportFrom.level to int | None

iyume:
See: https://github.com/python/cpython/issues/57645

### DIFF
--- a/stdlib/_ast.pyi
+++ b/stdlib/_ast.pyi
@@ -196,7 +196,7 @@ class ImportFrom(stmt):
         __match_args__ = ("module", "names", "level")
     module: str | None
     names: list[alias]
-    level: int
+    level: int | None
 
 class Global(stmt):
     if sys.version_info >= (3, 10):


hauntsaninja (commented):
(author of recent PRs in this space to CPython)
Note that level is None can only happen on artificially constructed nodes, so if it comes to it, I'd prefer errors for people doing that over false positives for everyone else. But mypy_primer seems okay, so maybe it's fine.

I'm curious what your interest in this change is? Does it affect a project of yours? If so, is that project open source?

iyume:
@hauntsaninja The idea is followed the ASDL and docs.

A similar issue https://github.com/python/cpython/issues/90447 changed ASDL to disallow optional.

But ast.ImportFrom still allow optional...

hauntsaninja:
Yes, I'm aware, see for instance https://github.com/python/cpython/pull/92987

I was just wondering if this affected code you're using or if this was just a theoretical thing :-)

iyume:
> Yes, I'm aware, see for instance [python/cpython#92987](https://github.com/python/cpython/pull/92987)
> 
> I was just wondering if this affected code you're using or if this was just a theoretical thing :-)

Here is: https://github.com/nonebot/nb-autodoc/blob/3786ea2e9a39e69551264b86c2ad1b755389d181/nb_autodoc/utils.py#L64-L69

But I will change `level` to `level or 0` to adapt the ASDL behaviour...

hauntsaninja:
Okay, thank you!

I'm -0.5 on this change, so I don't plan on hitting merge myself. I'll leave this PR to other maintainers :-)

This is basically a bug in CPython's ASDL, like https://github.com/python/cpython/pull/30467 which got fixed in 3.11. Unfortunately, this particular bug got half fixed in the wrong direction (making `compile` more permissive) and CPython is now hesitant to remove it for backward compatibility reasons.

It is only possible to have None here with artificially constructed nodes. When opening https://github.com/python/cpython/pull/92987 I grepped through a lot of real world code and couldn't find any cases where people relied on the behaviour with None when constructing nodes. And I don't like making changes that cause extra checks for things that never happen. Note that if we type AST constructors (https://github.com/python/typeshed/issues/8378) correspondingly, in fully typed programs we can guarantee there won't be any issue.

But shouldn't we type things close to the implementation? Yes, but we've repeatedly chosen not to change this, e.g. in https://github.com/python/typeshed/pull/7877. Implementations can be very flexible, especially for "plain old data" classes that can store basically anything. Plenty of functions allow ints when we type them as taking bools. Anyway, I recognise that this is an opinionated opinion (as opinions tend to be) and that mypy_primer currently claims no coverage of this, so 🤷

AlexWaygood:
I'm also not a massive fan of this change. There's nothing on mypy_primer that would currently be affected by this change, _but_:

1. I could quite easily see myself doing something with flake8-pyi at some point that might be affected by this change.
2. We should probably add flake8-pyi to mypy_primer!

AlexWaygood:
We have two maintainers who are both -1 on this, so I'm going to go ahead and close this for now. But thank you for the PR; sorry that we disagree on this one :)

Akuli:
I agree that type checking should prevent doing potentially bad things that CPython allows for historical reasons.

PR #8691, opened 2022-09-06
Use `ClassVar` for class-level attributes in `calendar`

sobolevn:
Source: https://github.com/python/cpython/blob/f177f6f29b069f522a0b3ba44eaae19852b6d2b0/Lib/calendar.py#L418-L436


AlexWaygood (approved):
The docs confirm that you're not really meant to override these attributes on instances; you should subclass these classes if you want to override these attributes: https://docs.python.org/3/library/calendar.html#calendar.HTMLCalendar

PR #8692, opened 2022-09-06
Use `ClassVar` for class-level attributes in `cmd`

sobolevn:
Source: https://github.com/python/cpython/blob/f177f6f29b069f522a0b3ba44eaae19852b6d2b0/Lib/cmd.py#L64-L74
Docs: https://docs.python.org/3/library/cmd.html

### DIFF
--- a/stdlib/cmd.pyi
+++ b/stdlib/cmd.pyi
@@ -8,7 +8,6 @@ PROMPT: Literal["(Cmd) "]
 IDENTCHARS: str  # Too big to be `Literal`
 
 class Cmd:
-    prompt: ClassVar[str]
     identchars: ClassVar[str]
     ruler: ClassVar[str]
     doc_leader: ClassVar[str]
@@ -17,6 +16,7 @@ class Cmd:
     undoc_header: ClassVar[str]
     nohelp: ClassVar[str]
     use_rawinput: ClassVar[bool]
+    prompt: str
     lastcmd: str
     intro: Any | None
     stdin: IO[str]
--- a/stdlib/cmd.pyi
+++ b/stdlib/cmd.pyi
@@ -11,14 +11,14 @@ class Cmd:
     prompt: ClassVar[str]
     identchars: ClassVar[str]
     ruler: ClassVar[str]
-    lastcmd: ClassVar[str]
-    intro: Any | None
     doc_leader: ClassVar[str]
     doc_header: ClassVar[str]
     misc_header: ClassVar[str]
     undoc_header: ClassVar[str]
     nohelp: ClassVar[str]
     use_rawinput: ClassVar[bool]
+    lastcmd: str
+    intro: Any | None
     stdin: IO[str]
     stdout: IO[str]
     cmdqueue: list[str]
--- a/stdlib/cmd.pyi
+++ b/stdlib/cmd.pyi
@@ -12,7 +12,7 @@ class Cmd:
     identchars: ClassVar[str]
     ruler: ClassVar[str]
     lastcmd: ClassVar[str]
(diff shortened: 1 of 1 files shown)


AlexWaygood (changes requested):
I'm not sure about this one. `intro` is clearly overridden on instances. Are we sure it's incorrect to do the same with the other attributes?  https://github.com/python/cpython/blob/f177f6f29b069f522a0b3ba44eaae19852b6d2b0/Lib/cmd.py#L116

It seems to me like these could easily be instance attributes with class-level defaults.

The docs also refer to these as "public instance variables", and it seems like this change breaks at least one project included in mypy_primer

sobolevn:
@AlexWaygood ok, I agree about `intro`. It is special due to `cmdloop` assign.

Here's what's wrong wrong with `jax`:
- [`cli_debugger.Cmd`](https://github.com/google/jax/blob/main/jax/_src/debugger/cli_debugger.py#L30) is somewhat incorrect: `use_rawinput: bool = False`. It is used as a class-level variable, but annotated as instance level one
- `self.intro = 'Entering jdb:'` is also used like a class-level variable, but for some reason is put into `self`
- The same for `webdebugger`: `use_rawinput: bool = False` https://github.com/google/jax/blob/main/jax/_src/debugger/web_debugger.py#L37

sobolevn:
I will send a PR to `jax` later today.

AlexWaygood (dismissed):
`lastcmd` is also definitely an instance attribute: https://github.com/python/cpython/blob/f177f6f29b069f522a0b3ba44eaae19852b6d2b0/Lib/cmd.py#L207

AlexWaygood (commented):
Are we _sure_ we need to disallow overriding `prompt` on instances? That's not how `ansible` uses this class: https://github.com/ansible/ansible/blob/223e05e0125de7a699e2a9ecc910981960a6452c/lib/ansible/cli/console.py#L141-L151

stdlib/cmd.pyi
```diff
@@ -8,23 +8,23 @@ PROMPT: Literal["(Cmd) "]
...
 class Cmd:
+    identchars: ClassVar[str]
+    ruler: ClassVar[str]
+    doc_leader: ClassVar[str]
+    doc_header: ClassVar[str]
+    misc_header: ClassVar[str]
+    undoc_header: ClassVar[str]
+    nohelp: ClassVar[str]
+    use_rawinput: ClassVar[bool]
     prompt: str
```
AlexWaygood, line 19:
```suggestion
    # "prompt", "lastcmd" and "intro" all have class-level default values,
    # but are commonly overridden on instances, so shouldn't be ClassVars
    prompt: str
```

AlexWaygood (commented):
I think this is okay now, but I still feel slightly nervous about accidentally causing a regression for somebody. We've identified that three attributes on this class are actually instance variables, even though they have class-level defaults, and that's a fairly common pattern for classes in the stdlib, most of which predate Python-typing. By annotating an attribute as a `ClassVar`, we're forbidding anybody to override the attribute on instances; are we _sure_ we need to do that for all of these attributes?

I'd like to get an opinion from another maintainer before merging.

JelleZijlstra:
What is the upside of making this change? I see a risk of regressions for legitimate users of this module, but little benefit to making the change.

sobolevn:
Ok, I see the confusion of how actually `Cmd` is documented: attributes are documented to be instance-level, but generally used a class-level. For example, `Cmd.__init__` is not suited for overriding these values. 

Probably, this is not something we need to push. Old python modules are quite specific :)

PR #8695, opened 2022-09-06
Update traceback.clear_frames() to accept None

jack1142:
`Exception.__traceback__` can potentially be None so for convenience
(and because the simplest way of implementing it will allow it anyway),
traceback.clear_frames() allows None.


hauntsaninja (approved):

hauntsaninja:
Thanks!

PR #8696, opened 2022-09-06
Add SearchableAPIResource and SearchResultObject for types-stripe

yejia-stripe:
Add type annotations for `SearchableAPIResource` and `SearchResultObject` for the stripe library stubs, and add `SearchableAPIResource` inheritance for relevant classes.

Fixes stripe/stripe-python#871


stubs/stripe/stripe/api_resources/search_result_object.pyi
```diff
@@ -0,0 +1,32 @@
+from _typeshed import Self
+from collections.abc import Iterator
+from typing import Any, Generic, Literal, TypeVar
```
AlexWaygood, line 3: Doesn't exist on 3.7

```suggestion
from typing import Any, Generic, TypeVar
from typing_extensions import Literal
```

yejia-stripe:
@AlexWaygood The overall feature isn't new to v4, but the way I'm typing it is not _technically_ accurate? The `search` and `search_auto_paging_iter` don't actually exist on `SearchableAPIResult`, but I didn't want to type each subclass separately. They will be defined after I release https://github.com/stripe/stripe-python/pull/873, but that would be for v4.

I would like to make this change without updating to v4. Would I be able to add an exception for this for now?

AlexWaygood:
> I would like to make this change without updating to v4. Would I be able to add an exception for this for now? Or should I add types for each subclass?

Yeah that should be fine! Sorry for the delayed response!

We do that for a few other classes, e.g.:

https://github.com/python/typeshed/blob/b4d91054e32b57698d5480abae77124a81dc0ff6/tests/stubtest_allowlists/py3_common.txt#L101-L102

You'll just need to add some entries to https://github.com/python/typeshed/blob/master/stubs/stripe/%40tests/stubtest_allowlist.txt to get the CI happy with it :)

stubs/stripe/stripe/api_resources/abstract/searchable_api_resource.pyi
```diff
@@ -0,0 +1,20 @@
...
+class SearchableAPIResource(APIResource):
+    @classmethod
+    def _search(
+        cls: type[Self],
+        search_url: str,
+        api_key: str | None = ...,
+        stripe_version: str | None = ...,
+        stripe_account: str | None = ...,
+        **params,
+    ) -> SearchResultObject[Self]: ...
```
AlexWaygood, line 16: Hmm, is this method public-API? We usually prefer to omit private methods (unless there's a good reason to make an exception) -- including them increases our maintenance burden, and users _probably_ shouldn't be using them anyway :)

stubs/stripe/stripe/api_resources/search_result_object.pyi
```diff
@@ -0,0 +1,33 @@
...
+from collections.abc import Iterator
+from typing import Any, Generic, TypeVar
+from typing_extensions import Literal
+
+from stripe.stripe_object import StripeObject
+
+_T = TypeVar("_T")
+
+class SearchResultObject(StripeObject, Generic[_T]):
+    OBJECT_NAME: Literal["search_result"] = ...
```
AlexWaygood, line 11: (You'll need to import `ClassVar` from `typing`)

```suggestion
    OBJECT_NAME: ClassVar[Literal["search_result"]]
```

stubs/stripe/stripe/api_resources/search_result_object.pyi
```diff
@@ -0,0 +1,33 @@
...
+    url: str
+    has_more: bool
+    data: list[_T]
+    next_page: str
+    total_count: int
```
AlexWaygood, line 16: Where are these attributes set?

AlexWaygood (commented):
Mostly looks good! Just one suggestion, and two questions

stubs/stripe/stripe/api_resources/abstract/searchable_api_resource.pyi
yejia-stripe: No this is private. Got it, will remove, thanks!

stubs/stripe/stripe/api_resources/search_result_object.pyi
yejia-stripe: Sorry, these aren't documented in the library itself, but are in the API reference. These attributes should match the "Search response format" described [here](https://stripe.com/docs/api/pagination/search).

[This](https://github.com/stripe/stripe-python/blob/1ae42227/stripe/util.py#L169) is where a `SearchResultObject` would be instantiated, which eventually sets those values [here](https://github.com/stripe/stripe-python/blob/fa26fa57ff980eee12ab6b770df601e6dd7bef34/stripe/stripe_object.py#L222).

AlexWaygood: Can't say I really follow the code here, but I'll take your word for it 😄

stubs/stripe/stripe/api_resources/search_result_object.pyi
```diff
@@ -0,0 +1,33 @@
...
+from collections.abc import Iterator
+from typing import Any, ClassVar, Generic, TypeVar
+from typing_extensions import Literal
+
+from stripe.stripe_object import StripeObject
+
+_T = TypeVar("_T")
+
+class SearchResultObject(StripeObject, Generic[_T]):
+    OBJECT_NAME: ClassVar[Literal["search_result"]] = ...
```
AlexWaygood, line 11:
```suggestion
    OBJECT_NAME: ClassVar[Literal["search_result"]]
```

AlexWaygood (approved):
Thanks!!

stubs/stripe/stripe/api_resources/search_result_object.pyi
yejia-stripe: Ah my bad, ty :)

yejia-stripe:
Thank you for the help!! :)

PR #8698, opened 2022-09-06
Require AsyncGenerator for asynccontextmanager

hauntsaninja:
I'm against merging #7430 because:
- Generator is not an ergonomic type
- It's very widely breaking
- Safety could easily be enforced by a linter that ensures use of `yield from x` instead of `return x` in an @contextmanager

To expand on that a little:
- I care about the typing system being accessible. Generator with its three type vars is not a friendly type. I see enough confusion about Iterable, Iterator, Generator as it is.
- Maintaining a high signal to noise / effort ratio is an important part of making typing accessible. Breaking changes that seem arbitrary to most users reinforce existing negative impressions of typing that hurt the ecosystem as a whole.
- In all the years of typing, this has come up basically never (reported twice by asottile, none of the outlinks from the issue contain the bug, issue itself is not popular), so I think this is 99.99% noise. Between typeshed and mypy and gitter and discord and work, I've seen a lot of typing issue reports.

But I'm willing to admit that the considerations are slightly different for asynccontextmanager:
- async is less widely used than sync, so this is less widely breaking (let's at least see primer)
- async is typically used by more experienced Python users, so there's already an accessibility floor here
- There is no equivalent to `yield from` in async that a linter could enforce, so this genuinely can only be solved in type checkers
- The issue that led to #7430 was with asynccontextmanager
- We've gotten away with some pedantic narrowings of async types before

This is just an experiment, I'm not sure whether I'm actually in favour of this change, but I'm a lot more willing to consider it. Note that I'm not open to slippery slope consistency arguments here; if a constraint is that we do both async and sync or neither, then I'm firmly in the neither camp.

### DIFF
--- a/stdlib/contextlib.pyi
+++ b/stdlib/contextlib.pyi
@@ -2,7 +2,7 @@ import abc
 import sys
 from _typeshed import Self, StrOrBytesPath
 from abc import abstractmethod
-from collections.abc import AsyncGenerator, AsyncIterator, Awaitable, Callable, Generator, Iterator
+from collections.abc import AsyncGenerator, Awaitable, Callable, Generator, Iterator
 from types import TracebackType
 from typing import IO, Any, Generic, Protocol, TypeVar, overload, runtime_checkable
 from typing_extensions import ParamSpec, TypeAlias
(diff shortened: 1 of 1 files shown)


AlexWaygood:
FWIW I'm pretty neutral on this:

- Unlike [#7430](https://github.com/python/typeshed/pull/7430), I think the mypy_primer fallout is acceptable
- `AsyncGenerator` doesn't seem to me to be _as_ horrifyingly ugly a type as `Generator` (only two type parameters instead of three)
- As you say, there's no `async` version of `yield from`, so the workarounds for `@contextmanager` don't apply for `@asynccontextmanager`

All that said:

- For all the months that PR was open, I still don't feel like I've seen a complete repro of the actual problem that led to [#7430](https://github.com/python/typeshed/pull/7430) being opened
- I'm still not persuaded that the problem that led to [#7430](https://github.com/python/typeshed/pull/7430) being opened is all that common
- This would still be at least somewhat disruptive
- `AsyncGenerator` _is_ still uglier as a type annotation than `AsyncIterator`
- It _would_ be slightly weird to do one thing for the sync version but another thing for the async version

hauntsaninja:
Okay, no one seems bullish on this and no one from the original PR has shown up in favour, so I'm happy to let status quo win. I remain potentially open to this change.

srittau:
FWIW, I'm in favor of every step in the right direction. And the sooner we make a change, the less problems we cause. I still think not biting the bullet in #7430 was a major mistake.

PR #8700, opened 2022-09-07
Add infrastructure allowing for test cases for third-party stubs

### LINKED ISSUE #8569: Figure out a way to have test cases for third-party stubs
We currently only have test cases for our stdlib stubs, and don't currently have the infrastructure set up to allow us to add test cases for any of our third-party stubs. This has come up a few times now, so it would be good to figure out a way to get this to work.

(This issue relates to the test cases in our `test_cases` directory. Currently we test these samples with mypy and pyright.)

AlexWaygood:
This PR:

- Moves the logic for running mypy on the test cases from `tests/mypy_test.py` to a separate script, `tests/regr_test.py`.
- Adds the necessary logic in order to be able to have test cases for third-party stubs.
- Moves logic common to `tests/mypy_test.py` and `tests/regr_test.py` into `tests/colors.py`, and renames `tests/colors.py` to `tests/utils.py`.
- Adds a new check to `tests/check_consistent.py`, to enforce the use of `# pyright: reportUnnecessaryTypeIgnoreComment=true` comments in third-party test cases. These are essential if we want to have our tests against false-negatives work with pyright.
- Updates the relevant documentation to account for the new test file.
- Adds a new job to the `tests.yml` GitHub workflow, to run the new test in CI.
- Adds a simple proof-of-concept test case for `requests`, as a regression test for #7998.
- Closes #8569


srittau:
Not reviewed yet, but the list above sounds great!

.github/workflows/tests.yml
```diff
@@ -80,6 +81,17 @@ jobs:
...
       - run: ./tests/mypy_test.py --platform=${{ matrix.platform }} --python-version=${{ matrix.python-version }}
 
+  regression-tests:
+    name: Run mypy on the test cases
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-python@v4
+        with:
+          python-version: 3.10
```
AlexWaygood, line 91:
```suggestion
          python-version: "3.10"
```

.github/workflows/tests.yml
```diff
@@ -80,6 +81,17 @@ jobs:
       - run: pip install -r requirements-tests.txt
       - run: ./tests/mypy_test.py --platform=${{ matrix.platform }} --python-version=${{ matrix.python-version }}
 
+  regression-tests:
+    name: Run mypy on the test cases
+    runs-on: ubuntu-latest
```
AlexWaygood, line 86:
```suggestion
    runs-on: windows-latest
```

AlexWaygood:
Okay, this should now be ready for review! Getting this to work both locally and in CI was a bit of a wild ride.

The second commit deliberately introduces errors in stdlib test cases and third-party test cases, to show what it looks like in CI when mypy and pyright fail on those test cases. Unfortunately, it looks like if there are failures in the stdlib test cases, pyright won't then go on to check the third-party test cases. I'm not sure that's fixable without rethinking how we run pyright in CI:

- pyright run on commit 2: https://github.com/python/typeshed/runs/8234439208?check_suite_focus=true
- mypy run on commit 2: https://github.com/python/typeshed/runs/8234442216?check_suite_focus=true

The third commit fixes the errors in the stdlib test cases, so we can see what it looks like when there are only errors in the third-party test cases:

- pyright run on commit 3: https://github.com/python/typeshed/runs/8234492908?check_suite_focus=true
- mypy run on commit 3: https://github.com/python/typeshed/runs/8234496168?check_suite_focus=true

The fourth commit fixes the errors in the third-party test cases, so we're back to how it was in commit 1.

tests/regr_test.py
```diff
@@ -0,0 +1,176 @@
...
+    # SO, to work around this, we copy the test_cases directory into a TemporaryDirectory.
+    with tempfile.TemporaryDirectory() as td:
+        td_path = Path(td)
+        new_test_case_dir = td_path / "test_cases"
+        shutil.copytree(test_case_directory, new_test_case_dir)
+        env_vars = dict(os.environ)
+        if IS_STDLIB:
+            flags.extend(["--custom-typeshed-dir", str(Path(__file__).parent.parent)])
+        else:
+            new_typeshed = td_path / "typeshed"
```
AlexWaygood, line 121:
```suggestion
            # HACK: we want to run these test cases in an isolated environment --
            # we want mypy to see all stub packages listed in the "requires" field of METADATA.toml
            # (and all stub packages required by those stub packages, etc. etc.),
            # but none of the other stubs in typeshed.
            #
            # The best way of doing that without stopping --warn-unused-ignore from working
            # seems to be to create a "new typeshed" directory in a tempdir
            # that has only the required stubs copied over.
            new_typeshed = td_path / "typeshed"
```

tests/regr_test.py
```diff
@@ -0,0 +1,184 @@
...
+def get_recursive_requirements(package_name: str, seen: set[str] | None = None) -> list[str]:
+    seen = seen if seen is not None else {package_name}
+    for dependency in filterfalse(seen.__contains__, read_dependencies(package_name)):
+        seen.update(get_recursive_requirements(dependency, seen))
+    return sorted(seen | {package_name})
+
+
+def test_testcase_directory(package: PackageInfo, version: str, platform: str) -> ReturnCode:
+    package_name, test_case_directory = package
+    IS_STDLIB = package_name == "stdlib"
```
JelleZijlstra, line 93: shouldn't be all-caps

tests/regr_test.py
```diff
@@ -0,0 +1,184 @@
...
+            # The best way of doing that without stopping --warn-unused-ignore from working
+            # seems to be to create a "new typeshed" directory in a tempdir
+            # that has only the required stubs copied over.
+            new_typeshed = td_path / "typeshed"
+            os.mkdir(new_typeshed)
+            shutil.copytree(Path("stdlib"), new_typeshed / "stdlib")
+            requirements = get_recursive_requirements(package_name)
+            for requirement in requirements:
+                shutil.copytree(Path("stubs", requirement), new_typeshed / "stubs" / requirement)
+            sep = ";" if sys.platform == "win32" else ":"
```
JelleZijlstra, line 135: This is `os.pathsep` (not to be confused with `os.path.sep`).

tests/regr_test.py
```diff
@@ -0,0 +1,184 @@
...
+            env_vars["MYPYPATH"] = sep.join(map(str, new_typeshed.glob("stubs/*")))
+            flags.extend(["--custom-typeshed-dir", str(td_path / "typeshed")])
+        result = subprocess.run([sys.executable, "-m", "mypy", new_test_case_dir, *flags], capture_output=True, env=env_vars)
+
+    stdout, stderr, returncode = result.stdout, result.stderr, result.returncode
```
JelleZijlstra, line 140: Can skip these variables

tests/utils.py
```diff
@@ -0,0 +1,79 @@
...
+@lru_cache()
+def read_dependencies(distribution: str) -> tuple[str, ...]:
+    with Path("stubs", distribution, "METADATA.toml").open("rb") as f:
+        data = tomli.load(f)
+    requires = data.get("requires", [])
+    assert isinstance(requires, list)
+    dependencies = []
+    for dependency in requires:
+        assert isinstance(dependency, str)
+        assert dependency.startswith("types-")
```
JelleZijlstra, line 54:
```suggestion
        assert dependency.startswith("types-"), f"unrecognized dependency {dependency}"
```

JelleZijlstra (commented):
Thanks! A few nits.

tests/regr_test.py
AlexWaygood: Huh, TIL!

JelleZijlstra (approved):

test_cases/README.md
```diff
@@ -1,14 +1,22 @@
 ## Regression tests for typeshed
 
 This directory contains code samples that act as a regression test for the
```
srittau, line 3:
```suggestion
This directory contains code samples that act as a regression test for
```

tests/regr_test.py
```diff
@@ -0,0 +1,182 @@
...
+    print_error,
+    print_success_msg,
+    read_dependencies,
+    testcase_dir_from_package_name,
+)
+
+ReturnCode: TypeAlias = int
+
+SUPPORTED_PLATFORMS = ["linux", "darwin", "win32"]
+SUPPORTED_VERSIONS = ["3.11", "3.10", "3.9", "3.8", "3.7"]
```
srittau, line 29: Maybe we should move this into `utils.py` as well? Or maybe add a `config.py`?

tests/utils.py
```diff
@@ -0,0 +1,79 @@
...
+
+import os
+from functools import lru_cache
+from pathlib import Path
+from typing import TYPE_CHECKING, NamedTuple
+
+import tomli
+
+# ====================================================================
+# Some simple hacks so we don't have to install types-termcolor in CI
```
srittau, line 11: Also to make termcolor itself optional.

srittau (approved):
LGTM, although I skipped some of the code for time constraints. One optional remark and some comment fluff.

tests/regr_test.py
AlexWaygood: I think it might be better to keep this per-test. We already have one test that supports a narrower version range than this:

https://github.com/python/typeshed/blob/e4d0d3d9d7b7366096d95ad10dada1638ceebf82/tests/typecheck_typeshed.py#L16

And we started testing 3.11 with `mypy_test.py` much earlier than we started testing it with `stubtest_stdlib.py`.

AlexWaygood:
Thanks for the reviews! Going to land this now, but happy to address any further feedback anybody has in follow-up PRs :)

PR #8701, opened 2022-09-07
Added _ExtendAction to argparse

jspaezp:
_ExtendAction was added to argparse since python 3.8. I am adding it here, since it is missing

Reference of the addition to the standard library: python/cpython@aa32a7e

### DIFF
--- a/stdlib/argparse.pyi
+++ b/stdlib/argparse.pyi
@@ -400,6 +400,9 @@ class _StoreFalseAction(_StoreConstAction):
 class _AppendAction(Action): ...
 
 # undocumented
+class _ExtendAction(_AppendAction): ...
+
+# undocumented
 class _AppendConstAction(Action):
     if sys.version_info >= (3, 11):
         def __init__(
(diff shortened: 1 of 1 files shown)


stdlib/argparse.pyi
```diff
@@ -399,6 +399,9 @@ class _StoreFalseAction(_StoreConstAction):
 # undocumented
 class _AppendAction(Action): ...
 
+# undocumented
+class _ExtendAction(_AppendAction): ...
```
AlexWaygood, line 403:
```suggestion
if sys.version_info >= (3, 8):
    class _ExtendAction(_AppendAction): ...
```

AlexWaygood (changes requested):
Looks good! We need to use a `sys.version_info` guard, though, since it's not available on 3.7:

jspaezp:
Thanks for the quick review!
Accepted the change!
Thanks, since it is my first contribution here I did not know we had to support py3.7

AlexWaygood (approved):
Thanks! And welcome to typeshed!

PR #8702, opened 2022-09-07
Added PyInstaller stubs for all documented modules & packages

Avasam:
`stubgen` was timing out (https://github.com/python/mypy/issues/9103), so I had to generate these stubs using Pyright rather than `create_baseline_stubs.py`. Hopefully I didn't miss anything.

`ignore_missing_stub` set to `true` because most modules are not meant to be used, yet are not marked as private.  
Documented modules:
- https://pyinstaller.org/en/stable/usage.html  
- https://pyinstaller.org/en/stable/hooks.html  
- https://pyinstaller.org/en/stable/hooks-config.html
- https://pyinstaller.org/en/stable/advanced-topics.html  
- https://github.com/pyinstaller/pyinstaller/blob/master/PyInstaller/fake-modules/pyi_splash.py
- All `__all__` from documented modules & packages

`pyi_splash` is a fake module that only exists once the app is frozen. It is meant to manipulate the splash screen.
image


stubs/pyinstaller/METADATA.toml
```diff
@@ -0,0 +1,6 @@
+version = "5.3.*"
+requires = ["types-setuptools"]
+
+[tool.stubtest]
+# Most modules are not meant to be used, yet are not marked as private
+ignore_missing_stub = true
```
Avasam, line 6: Or should I add said modules to `stubtest_allowlist.txt` ?

AlexWaygood: Definitely — though, it might be worth making the scale of the problem clear in the comment (138!!)

stubs/pyinstaller/PyInstaller/building/build_main.pyi
```diff
@@ -0,0 +1,28 @@
+# Referenced in: https://pyinstaller.org/en/stable/hooks.html?highlight=get_hook_config#PyInstaller.utils.hooks.get_hook_config
+# Not to be imported during runtime, but is the type reference for hooks and analysis configuration
+
+from _typeshed import StrOrBytesPath
+from collections.abc import Iterable
+
+from PyInstaller.building.datastruct import Target  # type: ignore[import]
```
AlexWaygood, line 7: Why are all the `pyinstaller` imports `type: ignore`d in this PR?

Avasam: I was getting this error for PyInstaller.building and PyInstaller.depend
```
stubs\pyinstaller\PyInstaller\depend\imphookapi.pyi:6: error: Skipping analyzing "PyInstaller.building.build_main": module is installed, but missing library stubs or py.typed marker
stubs\pyinstaller\PyInstaller\building\build_main.pyi:7: error: Skipping analyzing "PyInstaller.building.datastruct": module is installed, but missing library stubs or py.typed marker
stubs\pyinstaller\PyInstaller\building\build_main.pyi:7: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
stubs\pyinstaller\PyInstaller\utils\hooks\__init__.pyi:11: error: Skipping analyzing "PyInstaller.depend.imphookapi": module is installed, but missing library stubs or py.typed marker
```

I think I was just missing some `__init__.pyi` files.

AlexWaygood: Oh, that's a classic! Yeah, mypy struggles with namespace packages unless you have the `--namespace-packages` flag specified (which should be the default, see https://github.com/python/mypy/pull/9636).

stubs/pyinstaller/PyInstaller/building/build_main.pyi
```diff
@@ -0,0 +1,28 @@
...
+        hookspath=...,
+        hooksconfig: dict[str, dict[str, object]] | None = ...,
+        excludes=...,
+        runtime_hooks=...,
+        cipher=...,
+        win_no_prefer_redirects: bool = ...,
+        win_private_assemblies: bool = ...,
+        noarchive: bool = ...,
+        module_collection_mode=...,
+    ) -> None: ...
```
AlexWaygood, line 28: I'm confused about why pyright isn't emitting an error here. There's no entry in the `pyrightconfig.stricter.json` exclude-list, so it should be erroring for any parameters that don't have a type annotation.

stubs/pyinstaller/pyi_splash/__init__.pyi
```diff
@@ -0,0 +1,12 @@
...
+from typing_extensions import Literal
+
+__all__ = ["CLOSE_CONNECTION", "FLUSH_CHARACTER", "is_alive", "close", "update_text"]
+
+def is_alive() -> bool: ...
+def update_text(msg: str) -> None: ...
+def close() -> None: ...
+
+CLOSE_CONNECTION: Literal[b"\u0004"]
+FLUSH_CHARACTER: Literal[b"\r"]
```
AlexWaygood, line 12: Is this module in the wrong directory? Shouldn't it be `stubs/pyinstaller/PyInstaller/pyi_splash/__init__.pyi`, rather than `stubs/pyinstaller/pyi_splash/__init__.pyi`?

Avasam: You use it as such: `import pyi_splash` (see PR description for a code snippet). Which means it's a different `*-stubs` than PyInstaller. (which I've also tested by manually adding `pyi_splash-stubs` to my site-packages folder)

My current understanding is that with this setup, typeshed will publish a single `pyinstaller-types` package to PyPI. Which when installed will add both `PyInstaller-stubs` and `pyi_splash-stubs`.

AlexWaygood: Huh. I've never seen it done before. I'd want to make sure the typeshed bots would do the right thing when uploading the package to PyPI.

Sadly I know piteously little about how the logic in https://github.com/typeshed-internal/stub_uploader works. Can @hauntsaninja or @srittau help, maybe? :)

stubs/pyinstaller/PyInstaller/building/build_main.pyi
Avasam: Just a guess:
`pyrightconfig.stricter.json` has `"typeCheckingMode": "basic"`
and doesn't configure `reportMissingParameterType`. The default value for that setting being `None`

AlexWaygood: No idea -- anyway, it's not your fault, and we certainly don't have a requirement that all parameters need to be annotated when first adding stubs :) I'll look into it and try to work it out...

stubs/pyinstaller/pyi_splash/__init__.pyi
hauntsaninja: Should work just fine!

hauntsaninja: https://github.com/typeshed-internal/stub_uploader/blob/a4aba444e8bc45eb583efaeff268f04e4e976a19/stub_uploader/build_wheel.py#L154

AlexWaygood: Awesome, thank you!

stubs/pyinstaller/PyInstaller/depend/analysis.pyi
```diff
@@ -0,0 +1,19 @@
+# https://pyinstaller.org/en/stable/hooks.html#the-pre-safe-import-module-psim-api-method
+
+# The documentation explicitely mentions that "Normally you do not need to know about the module-graph."
+# However, some PyiModuleGraph typed class attributes are still documented as existing in imphookapi.
+from _typeshed import Incomplete
+from typing_extensions import LiteralString
+
+class PyiModuleGraph:  # incomplete
```
Avasam, line 8: Given the reason I haven't completed this stub (see comment at the top of the file). Does the `# incomplete` comment apply here?
I have the same question for similar reasons in `build_main.pyi`.

stubs/pyinstaller/PyInstaller/building/build_main.pyi
```diff
@@ -0,0 +1,28 @@
+# Referenced in: https://pyinstaller.org/en/stable/hooks.html?highlight=get_hook_config#PyInstaller.utils.hooks.get_hook_config
+# Not to be imported during runtime, but is the type reference for hooks and analysis configuration
+
+from _typeshed import StrOrBytesPath
+from collections.abc import Iterable
+
+from PyInstaller.building.datastruct import Target
+
+class Analysis(Target):
```
Avasam, line 9: Given the reason I haven't completed this stub (see comment at the top of the file). Should I add a `# incomplete` comment here?
I have the same question for similar reasons in `depend/analysis.pyi`.

stubs/pyinstaller/PyInstaller/utils/hooks/__init__.pyi
```diff
@@ -0,0 +1,79 @@
...
+from typing import Any, AnyStr
+from typing_extensions import Literal, TypeAlias
+
+import pkg_resources
+from PyInstaller import HOMEPATH as HOMEPATH, isolated
+from PyInstaller.depend.imphookapi import PostGraphAPI
+from PyInstaller.utils.hooks import conda as conda_support
+from PyInstaller.utils.hooks.win32 import get_pywin32_module_file_attribute as get_pywin32_module_file_attribute
+
+_Environ: TypeAlias = SupportsKeysAndGetItem[str, str] | Iterable[tuple[str, str]] | os._Environ[str]
```
AlexWaygood, line 15: Ultimately it looks like the `env` arguments are passed to `__exec_python_cmd`, and `__exec_python_cmd` uses it to update a deepcopy of `os.environ`. So I think any mapping will do here, not just instances of `os._Environ` (which are pretty hard to come by :)

https://github.com/pyinstaller/pyinstaller/blob/36aa18acf2a04a4f28e45f9e805480f0df0c577a/PyInstaller/utils/hooks/__init__.py#L55-L59

```suggestion
_Environ: TypeAlias = SupportsKeysAndGetItem[str, str] | Iterable[tuple[str, str]] | Mapping[str, str]
```

stubs/pyinstaller/PyInstaller/utils/hooks/__init__.pyi
```diff
@@ -0,0 +1,79 @@
...
+def exec_statement(statement: str) -> str | int: ...
+def exec_statement_rc(statement: str) -> str | int: ...
+def exec_script(script_filename: os.PathLike[AnyStr] | AnyOrLiteralStr, *args: str, env: _Environ | None = ...) -> str | int: ...
+def exec_script_rc(
+    script_filename: os.PathLike[AnyStr] | AnyOrLiteralStr, *args: str, env: _Environ | None = ...
+) -> str | int: ...
+def eval_statement(statement: str) -> Any | Literal[""]: ...
+def eval_script(
+    scriptfilename: os.PathLike[AnyStr] | AnyOrLiteralStr, *args: str, env: _Environ | None = ...
+) -> Any | Literal[""]: ...
```
AlexWaygood, line 29: I'm not sure there's any way for the `AnyStr` or `AnyOrLiteralStr` type variables to be solveable in these functions. I think maybe unions might be better than `TypeVar`s here.

stubs/pyinstaller/PyInstaller/utils/hooks/__init__.pyi
```diff
@@ -0,0 +1,79 @@
...
+@isolated.decorate
```
AlexWaygood, line 30: We usually omit decorators in stub files -- there are very limited exceptions, such as `@classmethod`, `@staticmethod`, and `@dataclass`, but outside of those, it's generally considered undefined behaviour in a stub.

```suggestion
```

stubs/pyinstaller/PyInstaller/utils/hooks/__init__.pyi
```diff
@@ -0,0 +1,79 @@
...
+def get_pyextension_imports(module_name: str) -> list[str]: ...
+def get_homebrew_path(formula: str = ...) -> str | None: ...
+def remove_prefix(string: str, prefix: str) -> str: ...
+def remove_suffix(string: str, suffix: str) -> str: ...
+def remove_file_extension(filename: str) -> str: ...
+@isolated.decorate
```
AlexWaygood, line 36:
```suggestion
```

stubs/pyinstaller/PyInstaller/utils/hooks/__init__.pyi
```diff
@@ -0,0 +1,79 @@
...
+def package_base_path(package_path: str, package: str) -> str: ...
+def get_package_paths(package: str) -> tuple[str, str]: ...
+def collect_submodules(
+    package: str, filter: Callable[[str], bool] = ..., on_error: Literal["ignore", "warn once", "warn", "raise"] = ...
+) -> list[str]: ...
+def is_module_or_submodule(name: str, mod_or_submod: str) -> bool: ...
+
+PY_DYLIB_PATTERNS: list[str]
+
+def collect_dynamic_libs(package: str, destdir: object | None = ...) -> list[tuple[str, str]]: ...
```
AlexWaygood, line 56: `None` is redundant in a union with `object`

```suggestion
def collect_dynamic_libs(package: str, destdir: object = ...) -> list[tuple[str, str]]: ...
```

stubs/pyinstaller/PyInstaller/utils/hooks/conda.pyi
```diff
@@ -0,0 +1,53 @@
...
+    _PackagePath: TypeAlias = Any
+
+CONDA_ROOT: Path
+CONDA_META_DIR: Path
+PYTHONPATH_PREFIXES: list[Path]
+
+class _RawDict(TypedDict):
+    name: str
+    version: str
+    files: list[StrOrBytesPath | PurePath]
```
AlexWaygood, line 22: `PurePath` is redundant in this union, since `PurePath` inherits from `PathLike[str]`, and `StrOrBytesPath` is just `str | bytes | PathLike[str] | PathLike[bytes]`

```suggestion
    files: list[StrOrBytesPath]
```

stubs/pyinstaller/PyInstaller/utils/hooks/conda.pyi
```diff
@@ -0,0 +1,53 @@
...
+    depends: list[str]
+
+class Distribution:
+    raw: _RawDict
+    name: str
+    version: str
+    files: list[PackagePath]
+    dependencies: list[str]
+    packages: list[str | None]
```
AlexWaygood, line 31: Not sure the items in `self.packages` can be `None`: `self.packages` is set [here](https://github.com/pyinstaller/pyinstaller/blob/36aa18acf2a04a4f28e45f9e805480f0df0c577a/PyInstaller/utils/hooks/conda.py#L103), and there's a check to make sure that `None` items are not appended to the list [here](https://github.com/pyinstaller/pyinstaller/blob/36aa18acf2a04a4f28e45f9e805480f0df0c577a/PyInstaller/utils/hooks/conda.py#L137-L138).

```suggestion
    packages: list[str]
```

stubs/pyinstaller/PyInstaller/utils/hooks/conda.pyi
```diff
@@ -0,0 +1,53 @@
...
+    def from_package_name(cls, name: str) -> Distribution: ...
+
+# distribution and package_distribution are meant to be used and are not internal helpers
+distribution = Distribution.from_name
+package_distribution = Distribution.from_package_name
+
+class PackagePath(_PackagePath):
+    def locate(self) -> Path: ...
+
+def walk_dependency_tree(initial: str, excludes: Sequence[str] | None = ...) -> dict[str, Distribution]: ...
```
AlexWaygood, line 45: I like the more permissive annotation in the source code here :)

```suggestion
def walk_dependency_tree(initial: str, excludes: Iterable[str] | None = ...) -> dict[str, Distribution]: ...
```

stubs/pyinstaller/PyInstaller/utils/hooks/conda.pyi
```diff
@@ -0,0 +1,53 @@
...
+def requires(name: str, strip_versions: bool = ...) -> list[str]: ...
+def files(name: str, dependencies: bool = ..., excludes: Sequence[str] | None = ...) -> list[PackagePath]: ...
+def collect_dynamic_libs(
+    name: str, dest: str = ..., dependencies: bool = ..., excludes: Sequence[str] | None = ...
```
AlexWaygood, line 49:
```suggestion
    name: str, dest: str = ..., dependencies: bool = ..., excludes: Iterable[str] | None = ...
```

AlexWaygood (commented):
Thanks! Mostly looks good, but a few points below :) all are fairly minor!

My main concern is using `LiteralString` for parameter annotations, which seems quite restrictive. Not sure about that.

stubs/pyinstaller/PyInstaller/compat.pyi
```diff
@@ -0,0 +1,82 @@
...
+is_solar: bool
+is_aix: bool
+is_freebsd: bool
+is_openbsd: bool
+is_hpux: bool
+is_unix: bool
+is_musl: bool
+is_macos_11_compat: tuple[int, ...] | bool | None
+is_macos_11_native: tuple[int, ...] | bool | None
+is_macos_11: tuple[int, ...] | bool | None
```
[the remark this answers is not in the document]
Avasam: They are in PyInstaller's develop branch because I did a PR over there to fix that. But as of right now with the version on PyPI this is accurate.
For instance, changing them to bool will raise the following in stubtest:
```
error: PyInstaller.compat.is_macos_11 variable differs from runtime type None
Stub: at line 33 in file stubs\pyinstaller\PyInstaller\compat.pyi
builtins.bool
Runtime:
None

error: PyInstaller.compat.is_macos_11_compat variable differs from runtime type None
Stub: at line 31 in file stubs\pyinstaller\PyInstaller\compat.pyi
builtins.bool
Runtime:
None

error: PyInstaller.compat.is_macos_11_native variable differs from runtime type None
Stub: at line 32 in file stubs\pyinstaller\PyInstaller\compat.pyi
builtins.bool
Runtime:
None
```

stubs/pyinstaller/PyInstaller/building/datastruct.pyi
```diff
@@ -0,0 +1,30 @@
+# https://pyinstaller.org/en/stable/advanced-topics.html#the-toc-and-tree-classes
```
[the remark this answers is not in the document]
Avasam: I figured I'd get a lot of questions about why a module was or wasn't included 😃

stubs/pyinstaller/PyInstaller/depend/analysis.pyi
```diff
@@ -0,0 +1,19 @@
...
+    def __init__(
+        self,
+        pyi_homepath: LiteralString,
```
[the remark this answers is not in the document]
Avasam: Here's my reasoning behind those (for anything in `imphookapi.pyi`, `analysis.pyi`, `build_main.pyi`, and `modulegraph.pyi`):
None of those are meant to be initialized by a dev using PyInstaller. Anythime you have to deal with one of those classes, it's because you were given an instance of it. The string class attributes are paths and package names found on the machine (and not something using user input), so I believe they are "safe literals".

However, some of those attributes are actually directly taken from __init__ parameters, so I made those LiteralStrings as well to stay consistent (I shouldn't assign a `str` to a `LiteralString`. As I mentioned, the dev has no reason to initialize those classes (ie: `PostGraphAPI(...)`), so it doesn't matter much if it's too restrictive. In fact, the only reason I'm defining the __init__ method is so that stubtest doesn't whine about missing parameters, or pyright about missing parameter types.

If I had a way to not expose the __init__ of those classes in the first place, I'd prefer that.

If you think we should just simplify it all to use `str`, I can understand that as well, I don't have an example of a use-case where those class attributes *need* to be literal.

stubs/pyinstaller/PyInstaller/depend/imphookapi.pyi
```diff
@@ -0,0 +1,76 @@
...
+    @property
+    def node(self) -> Package: ...
+    @property
+    def imports(self) -> Generator[Package, None, None]: ...
+    def add_imports(self, *module_names: str) -> None: ...
+    def del_imports(self, *module_names: str) -> None: ...
+    def add_binaries(self, list_of_tuples: TOC | Iterable[tuple[StrOrBytesPath, StrOrBytesPath]]) -> None: ...
+    def add_datas(self, list_of_tuples: TOC | Iterable[tuple[StrOrBytesPath, StrOrBytesPath]]) -> None: ...
+    def set_module_collection_mode(
+        self, name: str | None, mode: Literal["pyz", "pyc", "py", "pyz+py", "py+pyz"] | None
```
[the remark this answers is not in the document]
Avasam: Oh neat. There's one in compat.pyi as well.

stubs/pyinstaller/PyInstaller/utils/hooks/__init__.pyi
Avasam: Is there an easier/shorter way to define "anything that can be used to update a `dict[str, str]` ?

stubs/pyinstaller/PyInstaller/utils/hooks/conda.pyi
Avasam: Because excludes is changed to a set, right

Avasam:
Here's a batch of fixes from your comments, and a few discussion points.

stubs/pyinstaller/METADATA.toml
Avasam: I just realized I can use wildcards in the allowlist. Doing so I was able to bring it just under 40 items. Is this more acceptable or should I keep `ignore_missing_stub = true` ? 
```
# fake module, only exists once the app is frozen
pyi_splash
# undocumented and clearly not meant to be exposed
PyInstaller.__main__.generate_parser
PyInstaller.__main__.run_build
PyInstaller.__main__.run_makespec
PyInstaller.utils.hooks.conda.lib_dir
# A mix of modules meant to be private, and shallow incomplete type references for other modules
PyInstaller.building.*
PyInstaller.depend.analysis.*
# Most modules are not meant to be used, yet are not marked as private
PyInstaller.isolated._parent.*
PyInstaller.archive.*
PyInstaller.config
PyInstaller.configure
PyInstaller.depend.bindepend
PyInstaller.depend.bytecode
PyInstaller.depend.dylib
PyInstaller.depend.imphook
PyInstaller.depend.utils
PyInstaller.exceptions
PyInstaller.hooks.*
PyInstaller.lib.*
PyInstaller.loader.*
PyInstaller.log
PyInstaller.utils.cliutils.*
PyInstaller.utils.conftest
PyInstaller.utils.git
PyInstaller.utils.hooks.django
PyInstaller.utils.hooks.gi
PyInstaller.utils.hooks.qt
PyInstaller.utils.hooks.subproc.*
PyInstaller.utils.hooks.tcl_tk
PyInstaller.utils.misc
PyInstaller.utils.osx
PyInstaller.utils.run_tests
PyInstaller.utils.tests
PyInstaller.utils.win32.*
```

stubs/pyinstaller/PyInstaller/utils/hooks/__init__.pyi
AlexWaygood: Sadly no, not that I'm aware of :(

stubs/pyinstaller/METADATA.toml
AlexWaygood: Ah! Let's go with that then, please :)

If possible, it's definitely best not to have to enable `--ignore-missing-stub`. With stubtest on maximum strictness, our CI starts to fail when new methods are added to classes (etc.), so it's a very helpful way of ensuring that our stubs don't get too out of date.

stubs/pyinstaller/PyInstaller/depend/analysis.pyi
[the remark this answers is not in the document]
AlexWaygood: I think I'd prefer it if we went with `str` for now, if that's okay. `LiteralString` is a very new type, and I don't feel like I have a good idea about how using it this widely might affect usability of the package :)

stubs/pyinstaller/PyInstaller/lib/modulegraph/modulegraph.pyi
```diff
@@ -0,0 +1,39 @@
...
+    def __init__(self, identifier: LiteralString) -> None: ...
+    def is_global_attr(self, attr_name: str) -> bool: ...
+    def is_submodule(self, submodule_basename: str) -> bool: ...
+    def add_global_attr(self, attr_name: str) -> None: ...
+    def add_global_attrs_from_module(self, target_module: Node) -> None: ...
+    def add_submodule(self, submodule_basename: str, submodule_node: Node) -> None: ...
+    def get_submodule(self, submodule_basename: str) -> Node: ...
+    def get_submodule_or_none(self, submodule_basename: str) -> Node | None: ...
+    def remove_global_attr_if_found(self, attr_name: str) -> None: ...
+    def __lt__(self, other: _SupportsGraphident) -> bool: ...
```
AlexWaygood, line 26: It's useful to include `__eq__` and `__ne__` overrides, as mypy special-cases them for things like its `--strict-equality` check (included by default if you use `mypy --strict`)

```suggestion
    def __eq__(self, other: object) -> bool: ...
    def __ne__(self, other: object) -> bool: ...
    def __lt__(self, other: _SupportsGraphident) -> bool: ...
```

stubs/pyinstaller/PyInstaller/utils/hooks/conda.pyi
AlexWaygood: Yup!

stubs/pyinstaller/@tests/stubtest_allowlist.txt
```diff
@@ -0,0 +1,38 @@
...
+PyInstaller.utils.hooks.django
+PyInstaller.utils.hooks.gi
+PyInstaller.utils.hooks.qt
+PyInstaller.utils.hooks.subproc.*
+PyInstaller.utils.hooks.tcl_tk
+PyInstaller.utils.misc
+PyInstaller.utils.osx
+PyInstaller.utils.run_tests
+PyInstaller.utils.tests
+PyInstaller.utils.win32.*
```
AlexWaygood, line 38:
```suggestion
PyInstaller.utils.win32.*
# Explicitly private implementation details
PyInstaller\._*
PyInstaller.isolated._child
PyInstaller.utils._git_revision
```

AlexWaygood (approved):
LGTM! Just needs a couple more allowlist entries

AlexWaygood:
Thanks @Avasam! :D

PR #8705, opened 2022-09-08
[protobuf] Optional MessageToJson indent parameter

alkasm:
See https://github.com/protocolbuffers/protobuf/issues/10518


srittau (approved):
Thanks!

PR #8706, opened 2022-09-08
Make the return type of multiprocessing.connection.Pipe more precise

JukkaL:
The precise return type depends on the platform. Link to implementation: https://github.com/python/cpython/blob/68fb03249f3b17146db42b00a75718b823a2280c/Lib/multiprocessing/connection.py#L518

Now users no longer need to use the internal-looking class `_ConnectionBase` in annotations, at least in code that doesn't need to be cross-platform.


AlexWaygood (changes requested):
- We should update `multiprocessing.context.BaseContext.Pipe` as well (and probably add comments to say that they should be kept in sync).
- Can we add a comment to say that users can use `_ConnectionBase` if they want to write cross-platform code?

stdlib/multiprocessing/connection.pyi
```diff
@@ -58,4 +58,12 @@ def wait(
     object_list: Iterable[Connection | socket.socket | int], timeout: float | None = ...
 ) -> list[Connection | socket.socket | int]: ...
 def Client(address: _Address, family: str | None = ..., authkey: bytes | None = ...) -> Connection: ...
-def Pipe(duplex: bool = ...) -> tuple[_ConnectionBase, _ConnectionBase]: ...
+
+# N.B. Keep this in sync with multiprocessing.context.BaseContext.Pipe.
+# _BaseConnection is the common base class of Connection and PipeConnection
```
AlexWaygood, line 63:
```suggestion
# _ConnectionBase is the common base class of Connection and PipeConnection
```

stdlib/multiprocessing/context.pyi
```diff
@@ -43,7 +47,15 @@ class BaseContext:
     def active_children() -> list[BaseProcess]: ...
     def cpu_count(self) -> int: ...
     def Manager(self) -> SyncManager: ...
-    def Pipe(self, duplex: bool = ...) -> tuple[_ConnectionBase, _ConnectionBase]: ...
+
+    # N.B. Keep this in sync with multiprocessing.connection.Pipe.
+    # _BaseConnection is the common base class of Connection and PipeConnection
```
AlexWaygood, line 52:
```suggestion
    # _ConnectionBase is the common base class of Connection and PipeConnection
```

AlexWaygood (approved):
Some typos in the comments, but otherwise LGTM!

PR #8711, opened 2022-09-08
`mypy_test.py`: rework `filter`, `exclude` and `--dir` arguments

### LINKED ISSUE #8139: mypy_test: Are `exclude` and `filter` useful?

AlexWaygood:
(1) I accidentally disabled the `filter` and `exclude` command-line arguments in #8680 (oops). Get those working again.

(2) Rework `filter` and `exclude` so they use paths instead of regexes.

Currently the behaviour is such that if you do `python tests/mypy_test.py typing`, you'll pick up `stdlib/typing.pyi` and `stdlib/typing_extensions.pyi` (likely the files you wanted to test), but you'll also pick up random files from third-party stubs, like `redis/typing.pyi`. It doesn't make much sense to me. With this PR, you'll be able to do (for example):

- `python tests/mypy_test.py stdlib/typing.pyi` to test just the stdlib `typing` module
- `python tests/mypy_test.py stdlib/asyncio stdlib/distutils` to test the `stdlib/asyncio` and `stdlib/distutils` directories.
- `python tests/mypy_test.py stubs/protobuf` to test just the protobuf stubs
- `python tests/mypy_test.py stubs --exclude stubs/SQLAlchemy` to test all third-party stubs, excluding SQLALchemy
- `python tests/mypy_test.py stdlib` to test all stdlib stubs
- etc. etc.

(3) Get rid of the `--dir` argument. With these changes, it's no longer needed.

Closes #8139


tests/mypy_test.py
```diff
@@ -21,53 +21,55 @@
 from typing_extensions import Annotated, TypeAlias
 
 import tomli
-from utils import colored, print_error, print_success_msg, read_dependencies
+from utils import VERSIONS_RE as VERSION_LINE_RE, colored, print_error, print_success_msg, read_dependencies, strip_comments
 
-SUPPORTED_VERSIONS = [(3, 11), (3, 10), (3, 9), (3, 8), (3, 7)]
+SUPPORTED_VERSIONS = ["3.11", "3.10", "3.9", "3.8", "3.7"]
```
JelleZijlstra, line 26: Any reason not to centralize this in utils.py? It would be nice to minimize the number of places we have to change when 3.12 comes around or when we drop 3.7.

JelleZijlstra (approved):

tests/mypy_test.py
AlexWaygood: Yup, definitely. I'd like to do something like that in a follow-up PR, though — I have a few more ideas about things I'd like to do in that vein :)

PR #8716, opened 2022-09-09
Add missing attributes for docutils.io

amyreese:
`error_string` and `ErrorOutput` were added in `v0.19`. Fills out some missing pieces of `Input` and `FileInput` as well.


AlexWaygood (changes requested):
Thanks! A few things to fix, see my comments below :)

amyreese:
@AlexWaygood Thank you for the pointers! All have been applied now.

stubs/docutils/docutils/io.pyi
```diff
@@ -9,12 +10,17 @@ class InputError(IOError): ...
...
 
 class Input(TransformSpec):
     component_type: ClassVar[str]
     default_source_path: ClassVar[str | None]
     def read(self) -> Any: ...
-    def __getattr__(self, name: str) -> Any: ...  # incomplete
+    def decode(self, data: str | bytes) -> str: ...
+    coding_slug: ClassVar[Pattern[bytes]]
+    byte_order_marks: ClassVar[tuple[tuple[bytes, str], ...]]
+    def determine_encoding_from_data(self, data: str | bytes) -> str: ...
```
AlexWaygood, line 22: The function can return `None` (I mentioned this in https://github.com/python/typeshed/pull/8716#discussion_r966976730): https://sourceforge.net/p/docutils/code/HEAD/tree/trunk/docutils/docutils/io.py#l205

```suggestion
    def determine_encoding_from_data(self, data: str | bytes) -> str | None: ...
```

stubs/docutils/docutils/io.pyi
```diff
@@ -1,5 +1,6 @@
-from _typeshed import OpenBinaryModeReading, OpenBinaryModeWriting, OpenTextModeReading, OpenTextModeWriting
-from typing import Any, ClassVar
+from _typeshed import OpenBinaryModeReading, OpenBinaryModeWriting, OpenTextModeReading, OpenTextModeWriting, SupportsWrite
+from typing import Any, ClassVar, Pattern
```
AlexWaygood, line 2: Style nit: we prefer to import `Pattern` from `re` rather than `typing` in stub files :)

```suggestion
from re import Pattern
from typing import Any, ClassVar
```

AlexWaygood (approved):
Thanks @amyreese! I pushed two more additional fixes to this PR, everything looks good now 🎉

PR #8718, opened 2022-09-10
Fix `TODO` item in `sqlalchemy/engine/base.pyi`

sobolevn:
Looks like https://github.com/python/typeshed/pull/8415 did the same, but for a different class.


AlexWaygood (approved):
LGTM -- thanks!

PR #8720, opened 2022-09-10
`stubtest_third_party.py`: Make it work on Windows

AlexWaygood:


Akuli:
Is there a reason to run the script locally?

AlexWaygood:
> Is there a reason to run the script locally?

I've never felt a need to. However, here's a few reasons why I think this change might be good:

- `tests/README.md` advertises it as a test that _can_ be run locally. If we don't want to fix it so it's runnable on Windows, we should probably amend the docs to clarify that it can only be run locally if you're on a non-Windows platform.
- If you're contributing stubs for a new package (something I've only done once), you might plausibly have a good reason for running this script locally before filing a PR. The "create a virtual environment for testing, then automatically clean up the virtual environment after you've finished testing" feature is quite useful.
- If we merge #8652, we'll probably want to add a way to stubtest at least _some_ third-party packages on Windows in CI, since D3DShot is a Windows-only package. This change will be necessary if we want to start stubtesting third-party packages on Windows on CI.

Akuli (approved):
This function has quite a few local variables, but refactoring it is beyond the scope of this PR.

PR #8724, opened 2022-09-11
Use `ClassVar` in `string.Template`

sobolevn:
They are all used as `cls.ATTR_NAME`: https://github.com/python/cpython/blame/a36235d5c7863a85fa323b2048d3d254116a958e/Lib/string.py#L69-L85


AlexWaygood (approved):
The docs also describe these as "class attributes", and recommend subclassing the class if you want to override them: https://docs.python.org/3/library/string.html#string.Template.

Looks good!

PR #8726, opened 2022-09-11
Use `@abstractmethod` for abstract properties in `typing`

sobolevn:
Source: https://github.com/python/cpython/blame/53a54b781d1f05f2d0b40ce88b3da92d5d23e9d2/Lib/typing.py#L3167-L3297


sobolevn:
I've opened a new bug for `stubtest`: https://github.com/python/mypy/issues/13646

srittau:
I mostly consider `IO`, `TextIO` and `BytesIO` a lost cause. There too many nits and incompatibilities to use or type them properly.

sobolevn:
This is too disruptive. Let's try again later some time :)

Issue #8727, opened 2022-09-12
Update *func* parameter annotation for sqlite3's create_function().


shawnbrown:
The *func* parameter type for sqlite3's `create_function()` is defined as `callback | None` (see [docs](https://docs.python.org/3/library/sqlite3.html#sqlite3.Connection.create_function)) but the typeshed stub does not include `None` as a valid type. This is a long-undocumented feature that was only recently added to the docs (see related issue python/cpython#95235 and commit python/cpython@5012bed).

In **stdlib/sqlite3/dbapi2.pyi**, the annotation for this parameter currently reads ``func: Callable[..., _SqliteData]`` ([link](https://github.com/python/typeshed/blob/1ec57121e3570cf8d10c00dc418e043a35996b2e/stdlib/sqlite3/dbapi2.pyi#L319-L324)) but it should be changed to ``func: Callable[..., _SqliteData] | None``:

```python
if sys.version_info >= (3, 8):
    def create_function(
        self, name: str, narg: int, func: Callable[..., _SqliteData] | None, *, deterministic: bool = ...
    ) -> None: ...
else:
    def create_function(
        self, name: str, num_params: int, func: Callable[..., _SqliteData] | None
    ) -> None: ...
```

srittau:
Fixed in #8728.

PR #8728, opened 2022-09-12
sqlite3: update func parameter for create_function

shawnbrown:
By setting `func=None`, user-defined functions can be removed.
This long-undocumented behavior was only clarified July 26,
2022 (python/cpython@5012bed).

This addresses python/typeshed#8727.


Akuli (approved):
`None` doesn't cause errors to me:

```
>>> import sqlite3
>>> c=sqlite3.connect(':memory:')
>>> c.create_function('foo', 3, print)
>>> c.create_function('foo', 3, None)
```

I also found some documentation that mentions passing in NULLs: https://www.sqlite.org/c3ref/create_function.html

PR #8729, opened 2022-09-12
usjon: Add default keyword argument

antoniobusrod:
Add missing `default` keyword argument to `dump` and `dumps` functions.

Introduced since https://github.com/ultrajson/ultrajson/releases/tag/4.2.0

Relates to https://github.com/ultrajson/ultrajson/pull/470

### DIFF
--- a/stubs/ujson/ujson.pyi
+++ b/stubs/ujson/ujson.pyi
@@ -10,6 +10,7 @@ def encode(
     escape_forward_slashes: bool = ...,
     sort_keys: bool = ...,
     indent: int = ...,
+    default: Any = ...,
 ) -> str: ...
 def dumps(
     obj: Any,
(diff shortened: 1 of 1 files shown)


Akuli (commented):
The value is called with one argument to get the default value, so we can use `Callable[[Any], Any]` instead of just `Any`.

```
>>> ujson.encode({'a': object()}, default=(lambda x: "lol"))
'{"a":"lol"}'
```

We should also add this parameter to `dump()`.

```
>>> ujson.dump({'a': object()}, sys.stdout, default=(lambda x: "lol"))
{"a":"lol"}>>> 
```

Akuli:
It also seems like most of the arguments are keyword-only, even though they aren't marked as keyword-only in the stub:

```
>>> ujson.dump({'a': object()}, sys.stdout, True, 1, True, True, True, 1, (lambda x: "lol"))
TypeError: function takes exactly 2 arguments (9 given)
```

We can fix this in a separate pull request.

JelleZijlstra:
Could you address @Akuli's feedback?

AlexWaygood:
Closing due to lack of response from OP

PR #8733, opened 2022-09-13
Add new `sys` functions introduced to combat CVE-2020-10735

### LINKED ISSUE #8732: Stubtest failed on Tue Sep 13 2022

AlexWaygood:
Fixes #8732


stdlib/sys.pyi
```diff
@@ -328,3 +332,8 @@ if sys.version_info < (3, 8):
     _CoroWrapper: TypeAlias = Callable[[Coroutine[Any, Any, Any]], Any]
     def set_coroutine_wrapper(__wrapper: _CoroWrapper) -> None: ...
     def get_coroutine_wrapper() -> _CoroWrapper: ...
+
+# The following two functions were added in 3.11.0, 3.10.7, 3.9.14, 3.8.14, & 3.7.14,
+# as part of the response to CVE-2020-10735
+def set_int_max_str_digits(__n: int) -> None: ...
```
Akuli, line 338:
```suggestion
def set_int_max_str_digits(maxdigits: int) -> None: ...
```

Akuli (commented):
This might get reverted because it's pretty controversial, but if that happens we can just revert in typeshed :)

stdlib/sys.pyi
AlexWaygood: Argh, why does the `sys` documentation here differ from the output of `help()`?? https://docs.python.org/3/library/sys.html#sys.set_int_max_str_digits

Akuli (approved):

PR #8736, opened 2022-09-14
Add type stubs for cffi

freundTech:
I'm currently working on creating type stubs for cffi.
cffi is a popular library, which generates python-to-c-bindings. Sadly neither cffi itself, nor its generated bindings support type hints.

This PR adds stubs for cffi itself (cffi.api.FFI) as well as the part of the generated bindings that are always the same (_cffi_backend.FFI).
I'm currently also working on a project, which generated stubs for cffi generated bindings (stubgen can't deal with cffi generated bindings), which I will publish in a separate repository.

These stubs have been generated using `create_baseline_stubs.py`.
The stubs in for cffi.api.FFI and _cffi_backend.FFI have been manually refined and verified. Everything else has only been verified by the typeshed tests.


stubs/cffi/_cffi_backend.pyi
```diff
@@ -49,10 +50,86 @@ class CType:
...
+    def __gt__(self, other): ...
+    def __le__(self, other): ...
+    def __len__(self): ...
+    def __lt__(self, other): ...
+    def __ne__(self, other): ...
+    def __setitem__(self, index, object): ...
+
 class FFI:
     CData: TypeAlias = _CDataBase
     CType: TypeAlias = CType
```
AlexWaygood, line 131: The pyright errors along the lines of `Type alias "CType" cannot use itself in its definition (reportGeneralTypeIssues)` are quite annoying (someone should probably file an issue about it). See https://github.com/python/typeshed/blob/42c044eb6772ee985fccc584f039998131bd5216/stdlib/multiprocessing/reduction.pyi#L55-L78 for examples of some hacky workarounds we use elsewhere.

freundTech:
Is there any experience on how to best deal with classes that can't be subclassed at runtime, but have subclasses in the stubs?

If I don't mark them as final stubtest complains. If I mark them as final mypy and pyright complain when I subclass them in the stubs. I added `type: ignore[misc]` to silence mypy, but pyright still complains

AlexWaygood:
> I added `type: ignore[misc]` to silence mypy, but pyright still complains

It looks like you might just have the `type: ignore` on the wrong line. Try moving it onto the line of the class definition, rather than the line of the `@final` decorator? (Or, if that doesn't work -- try adding two `type: ignore`s, one to satisfy mypy, the other pyright?)

AlexWaygood:
The "type ignore your problems away" approach works pretty well for these classes, which have the same issue:

https://github.com/python/typeshed/blob/42c044eb6772ee985fccc584f039998131bd5216/stdlib/collections/__init__.pyi#L327-L340

It _should_ work here, unless something weird is going on 🤔

stubs/cffi/_cffi_backend.pyi
```diff
@@ -0,0 +1,290 @@
...
+    length: Incomplete
+    relements: Incomplete
+    result: Incomplete
+    def __dir__(self): ...
+
+@final
+class Lib:
+    def __delattr__(self, name): ...
+    def __dir__(self): ...
+    def __setattr__(self, name, value): ...
```
AlexWaygood, line 57: `__delattr__` and `__setattr__` have special meaning for type checkers if you override them. Both mypy and pyright will assume that any arbitrary attributes can be set on instances of the class if you override `__setattr__`. Pyright applies similar special-casing to `__delattr__`, though mypy doesn't yet.

Are these special semantics intended here? If not, it might be best to delete them:

```suggestion
    def __dir__(self): ...
```

stubs/cffi/_cffi_backend.pyi
```diff
@@ -0,0 +1,290 @@
...
+
+@final
+class _CDataBase:
+    __name__: ClassVar[str] = ...
```
AlexWaygood, line 61: Style nit: the `= ...` is usually unnecessary in stub files, and we prefer to omit it where possible

```suggestion
    __name__: ClassVar[str]
```

stubs/cffi/_cffi_backend.pyi
```diff
@@ -0,0 +1,290 @@
...
+    def __add__(self, other): ...
+    def __bool__(self): ...
+    def __call__(self, *args, **kwargs): ...
+    def __complex__(self): ...
+    def __delattr__(self, name): ...
```
AlexWaygood, line 66:
```suggestion
```

stubs/cffi/_cffi_backend.pyi
```diff
@@ -0,0 +1,290 @@
...
+    def __hash__(self): ...
+    def __int__(self): ...
+    def __iter__(self): ...
+    def __le__(self, other): ...
+    def __len__(self): ...
+    def __lt__(self, other): ...
+    def __ne__(self, other): ...
+    def __radd__(self, other): ...
+    def __rsub__(self, other): ...
+    def __setattr__(self, name, value): ...
```
AlexWaygood, line 85:
```suggestion
```

stubs/cffi/_cffi_backend.pyi
```diff
@@ -0,0 +1,290 @@
...
+
+    class error(Exception): ...
+    NULL: ClassVar[CData] = ...
+    RTLD_DEEPBIND: ClassVar[int] = ...
+    RTLD_GLOBAL: ClassVar[int] = ...
+    RTLD_LAZY: ClassVar[int] = ...
+    RTLD_LOCAL: ClassVar[int] = ...
+    RTLD_NODELETE: ClassVar[int] = ...
+    RTLD_NOLOAD: ClassVar[int] = ...
+    RTLD_NOW: ClassVar[int] = ...
```
AlexWaygood, line 148:
```suggestion
    NULL: ClassVar[CData]
    RTLD_DEEPBIND: ClassVar[int]
    RTLD_GLOBAL: ClassVar[int]
    RTLD_LAZY: ClassVar[int]
    RTLD_LOCAL: ClassVar[int]
    RTLD_NODELETE: ClassVar[int]
    RTLD_NOLOAD: ClassVar[int]
    RTLD_NOW: ClassVar[int]
```

stubs/cffi/cffi/backend_ctypes.pyi
```diff
@@ -0,0 +1,81 @@
...
+    __metaclass__: Incomplete
+    __name__: str
+    def __init__(self, *args) -> None: ...
+    def __iter__(self): ...
+    __eq__: Incomplete
+    __ne__: Incomplete
+    __lt__: Incomplete
+    __le__: Incomplete
+    __gt__: Incomplete
+    __ge__: Incomplete
```
AlexWaygood, line 22: These look like they probably should be typed as methods rather than variables

stubs/cffi/cffi/backend_ctypes.pyi
```diff
@@ -0,0 +1,81 @@
...
+class CTypesGenericPrimitive(CTypesData):
+    def __hash__(self): ...
+
+class CTypesGenericArray(CTypesData):
+    def __iter__(self): ...
+
+class CTypesGenericPtr(CTypesData):
+    kind: str
+    def __nonzero__(self): ...
+    __bool__: Incomplete
```
AlexWaygood, line 35: At a guess...

```suggestion
    def __bool__(self) -> bool: ...
```

AlexWaygood (commented):
Thanks for this! Here's a few initial remarks from skimming through. (I haven't taken a detailed look at any of the annotations yet.)

stubs/cffi/_cffi_backend.pyi
```diff
@@ -0,0 +1,286 @@
...
+
+FFI_CDECL: int
+FFI_DEFAULT_ABI: int
+RTLD_DEEPBIND: int
+RTLD_GLOBAL: int
+RTLD_LAZY: int
+RTLD_LOCAL: int
+RTLD_NODELETE: int
+RTLD_NOLOAD: int
+RTLD_NOW: int
```
AlexWaygood, line 21: Some of these look like they're platform-specific. On my Windows machine:

```pycon
>>> import _cffi_backend as c
>>> c.FFI_CDECL
1
>>> c.FFI_DEFAULT_ABI
1
>>> c.RTLD_DEEPBIND
AttributeError: module '_cffi_backend' has no attribute 'RTLD_DEEPBIND'
>>> c.RTLD_GLOBAL
0
>>> c.RTLD_LAZY
0
>>> c.RTLD_LOCAL
0
>>> c.RTLD_NODELETE
AttributeError: module '_cffi_backend' has no attribute 'RTLD_NODELETE'
>>> c.RTLD_NOLOAD
AttributeError: module '_cffi_backend' has no attribute 'RTLD_NOLOAD'. Did you mean: 'RTLD_GLOBAL'?
>>> c.RTLD_NOW
0
```

stubs/cffi/_cffi_backend.pyi
```diff
@@ -0,0 +1,286 @@
...
+    def __len__(self): ...
+    def __lt__(self, other): ...
+    def __ne__(self, other): ...
+    def __radd__(self, other): ...
+    def __rsub__(self, other): ...
+    def __setitem__(self, index, object): ...
+    def __sub__(self, other): ...
+
+@final
+class __CDataFromBuf(_CDataBase): ...  # type: ignore[misc]
```
AlexWaygood, line 86: We _generally_ don't include stubs for private things unless there's a specific reason to make an exception. This is to lower the maintenance burden for us, and because a variety of things in our test suite assume that class starting with private names are things that shouldn't be tested, since they're usually "private to the stub" things like `TypedDict`s, `Protocol`s and type aliases.

I know very little about `cffi` -- is there a reason why it's important to include stubs for `__CDataFromBuf` (etc.)? (`_CData` is fine because we need it in the stub in order to type the `FFI` class correctly.)

stubs/cffi/_cffi_backend.pyi
```diff
@@ -0,0 +1,286 @@
...
+
+    class error(Exception): ...
+    NULL: ClassVar[CData]
+    RTLD_DEEPBIND: ClassVar[int]
+    RTLD_GLOBAL: ClassVar[int]
+    RTLD_LAZY: ClassVar[int]
+    RTLD_LOCAL: ClassVar[int]
+    RTLD_NODELETE: ClassVar[int]
+    RTLD_NOLOAD: ClassVar[int]
+    RTLD_NOW: ClassVar[int]
```
AlexWaygood, line 144: Looks like some of these may also be platform-specific. On my Windows machine:

```pycon
>>> import _cffi_backend as c
>>> c.FFI.RTLD_DEEPBIND
AttributeError: type object '_cffi_backend.FFI' has no attribute 'RTLD_DEEPBIND'
>>> c.FFI.RTLD_GLOBAL
0
>>> c.FFI.RTLD_LAZY
0
>>> c.FFI.RTLD_LOCAL
0
>>> c.FFI.RTLD_NODELETE
AttributeError: type object '_cffi_backend.FFI' has no attribute 'RTLD_NODELETE'
>>> c.FFI.RTLD_NOLOAD
AttributeError: type object '_cffi_backend.FFI' has no attribute 'RTLD_NOLOAD'. Did you mean: 'RTLD_GLOBAL'?
>>> c.FFI.RTLD_NOW
0
```

stubs/cffi/cffi/cffi_opcode.pyi
```diff
@@ -0,0 +1,94 @@
+from _typeshed import Incomplete
+
+from .error import VerificationError as VerificationError
```
AlexWaygood, line 3: Is there a reason to publicly re-export this here? (Same question for most of the `from .foo import bar as bar` imports in the following files as well -- stubgen can get a big over-eager in assuming every import is a public re-export)

```suggestion
```

AlexWaygood (commented):
Some more small comments

stubs/cffi/_cffi_backend.pyi
freundTech: I just checked the source of the C module and there it's defined with a preprocessor macro like this:
```
#ifdef RTLD_DEEPBIND
    { "RTLD_DEEPBIND", RTLD_DEEPBIND },
#endif
```
I'll need to research which operating systems expose which flags in order to correctly map this to `if sys.platform` conditions.

stubs/cffi/_cffi_backend.pyi
freundTech: I think I can delete all of those except for _CDataBase.

stubs/cffi/_cffi_backend.pyi
freundTech: Same as above. Will need to do more research

stubs/cffi/_cffi_backend.pyi
AlexWaygood: We could put the ones that aren't on my Windows machine behind `if sys.platform != "win32"` guards for now, since we can probably be confident that they don't exist on Windows. Or we can leave it to a future PR. Up to you; I'm easy either way :)

stubs/cffi/cffi/cffi_opcode.pyi
freundTech: No. I just checked all the files and I think all reexports except for in `__init__.py` can be removed.

However: Two reexports cause stubtest errors when removed (`from .commontypes import COMMON_TYPES` in `cparser.pyi` and `from . import __version_verifier_modules__` in `verifier.pyi`).
I can either

1. Keep the reexports for them, even though they aren't needed
2. Turn them into normal imports, which makes pyright complain about non-accessed import, which I would have to silence or
3. Don't import them, which makes stubtest complain, which I could silence in `stubtest_allowlist.txt`

AlexWaygood: Yup. Stubtest is pretty good at recognising when a class or a function comes from another module (and generally won't complain about missing re-exports in those situations), but struggles a little more (for obvious reasons) to distinguish which constants are defined inside modules, and which are being re-exported from other modules. 

Option (3) is best — just allowlist them, and add a comment to explain why (needless re-exports).

AlexWaygood (approved):
Thank you! I haven't manually verified every one of these signatures, but I can't see any more red flags, and fixes can always be made in future PRs if there any issues. I'm happy for this to go in :)

freundTech:
Thanks for reviewing this so quickly!

PR #8741, opened 2022-09-15
stubsabot: refuse to run if uncommitted changes are present locally

AlexWaygood:
I keep on doing this thing where I'll:

1. make experimental changes to the stubsabot script
2. then run stubsabot to check if the changes work
3. find out that my changes to the stubsabot script were committed to one of the random stubsabot branches that the script creates locally, because I forgot to commit my changes to the stubsabot script before running stubsabot.

This PR makes it so that stubsabot just refuses to run if you forget to commit your changes locally before running it.

### DIFF
--- a/scripts/stubsabot.py
+++ b/scripts/stubsabot.py
@@ -398,6 +398,19 @@ async def main() -> None:
     )
     args = parser.parse_args()
 
+    if args.action_level > ActionLevel.nothing:
+        subprocess.run(["git", "update-index", "--refresh"], capture_output=True)
+        diff_result = subprocess.run(["git", "diff-index", "HEAD", "--name-only"], text=True, capture_output=True)
+        if diff_result.returncode:
+            print("Unexpected exception!")
+            print(diff_result.stdout)
+            print(diff_result.stderr)
+            sys.exit(diff_result.returncode)
+        if diff_result.stdout:
+            changed_files = ", ".join(repr(line) for line in diff_result.stdout.split("\n") if line)
+            print(f"Cannot run stubsabot, as uncommitted changes are present in {changed_files}!")
+            sys.exit(1)
+
     if args.action_level > ActionLevel.fork:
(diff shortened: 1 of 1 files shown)


scripts/stubsabot.py
```diff
@@ -398,6 +398,19 @@ async def main() -> None:
     )
     args = parser.parse_args()
 
+    if args.action_level > ActionLevel.nothing:
+        subprocess.run(["git", "update-index", "--refresh"], capture_output=True)
```
hauntsaninja, line 402: what does update-index do?

the way i've done this kind of thing is use `git diff-index HEAD --quiet` which sets exit code to 1 if things are dirty and exit code 0 if things are clean

AlexWaygood: It's the first time I've done this kind of thing, and I was principally working off [this StackOverflow answer](https://stackoverflow.com/a/3879077/13990016) :)

I think the reasoning in the answer sounds pretty good, though. According to the answer:

> if there are files that have been touched, but whose contents are the same as in the index, you'll need to run `git update-index --refresh` before `git diff-index`, otherwise `diff-index` will incorrectly report that the tree is dirty

> the way i've done this kind of thing is use `git diff-index HEAD --quiet` which sets exit code to 1 if things are dirty and exit code 0 if things are clean

I considered that, but playing around with it I decided I liked having a precise error message that would tell me exactly which files had uncommitted changes in them. Don't feel too strongly about it, though :)

AlexWaygood: > the way i've done this kind of thing is use `git diff-index HEAD --quiet` which sets exit code to 1 if things are dirty and exit code 0 if things are clean

I considered that, but playing around with it I decided I liked having a precise error message that would tell me exactly which files had uncommitted changes in them. Don't feel too strongly about it, though :)

hauntsaninja: cool, TIL!

hauntsaninja (approved):

PR #8744, opened 2022-09-15
stubsabot PR messages: link to diff between releases on GitHub, where possible

AlexWaygood:


scripts/stubsabot.py
```diff
@@ -183,6 +183,70 @@ def get_updated_version_spec(spec: str, version: packaging.version.Version) -> s
     return _check_spec(".".join(rounded_version) + ".*", version)
 
 
+@functools.cache
+def get_github_api_headers() -> dict[str, str]:
```
hauntsaninja, line 187: nit: if we're caching, should type as Mapping

scripts/stubsabot.py
```diff
@@ -183,6 +183,70 @@ def get_updated_version_spec(spec: str, version: packaging.version.Version) -> s
...
+    """
+    github_info = await get_github_repo_info(session, pypi_info)
+    if github_info is None:
+        return None
+    tags_to_versions = {tag["name"]: packaging.version.Version(tag["name"]) for tag in github_info.tags}
+    curr_release = packaging.version.Version(stub_info.version_spec.removesuffix(".*")).release
+
+    try:
+        new_tag = next(tag for tag, version in tags_to_versions.items() if version == pypi_info.version)
+        old_tag = next(tag for tag, version in tags_to_versions.items() if version.release[: len(curr_release)] == curr_release)
```
hauntsaninja, line 239: maybe slightly more accurate as the equivalent of:
old_version = max(v for v in versions if v in packaging.specifiers.SpecifierSet(f"=={stub_info.version_spec}"))

hauntsaninja (commented):
Thanks, this is useful!

scripts/stubsabot.py
AlexWaygood: Ah, good point!

hauntsaninja (approved):

AlexWaygood:
Thanks @hauntsaninja! :D

Issue #8686, opened 2022-09-05
Single script to run PR checks locally


Avasam:
After running `scripts/create_baseline_stubs.py`, we are informed of the following
```
Suggested next steps:
1. Manually review the generated stubs in {stub_dir}
2. Run "MYPYPATH={stub_dir} python3 -m mypy.stubtest {package}" to check the stubs against runtime
3. Run "mypy {stub_dir}" to check for errors
4. Run "black {stub_dir}" and "isort {stub_dir}" (if you\'ve made code changes)
5. Run "flake8 {stub_dir}" to check for e.g. unused imports
6. Commit the changes on a new branch and create a typeshed PR
```

There's a few issues with running those commands as such:
1. Setting the `MYPYPATH` environment variable and running a script afterward is done differently per shell (think of powershell for example, where its `$Env:MYPYPATH="{stub_dir}"; python3 -m mypy.stubtest {package}`)
2. These will ignore some configurations such as `METADATA.toml`, `stubtest_allowlist.txt`, etc.
3. Does not include other checks defined in https://github.com/python/typeshed/blob/master/tests/README.md
4. It takes multiple commands to autofix and check everything
5. Overall, the results will be different from what's on the CI

For those reasons, I suggest adding a new script under `scripts` that a dev who's contributing to typeshed can run locally to easily and completely check their stubs. The documentation in [CONTRIBUTING.md](https://github.com/python/typeshed/blob/master/CONTRIBUTING.md) and the output of `scripts/create_baseline_stubs.py` should then be updated accordingly.

Akuli:
I would prefer:
```
1. Manually review the generated stubs in {stub_dir}
2. Run tests locally if you prefer (see CONTRIBUTING.md)
3. Commit the changes on a new branch and create a typeshed PR
```

The CI will check everything anyway, and is set up so that you don't have to run anything locally. This would also be consistent with CONTRIBUTING.md.

Avasam:
@Akuli I'm mainly asking this because of the constant back and forth this produces with pushing changes, waiting on the CI (which has additional overhead), opening the errors page and dropdowns, then searching for the error. I'd much rather be able to quickly run all autofixes and checks locally.

My editor being VSCode, I could just setup a task for this project to run everything with the click of a single button. (but that's only gonna be beneficial to me)

[...]

PR #8769, opened 2022-09-19
Support for npx 8 and pyright PyPI package

Avasam:
Fix the npx command for npx 8:
`'pyright@1.1.270' is not recognized as an internal or external command,`

Also ensures that the right version of pyright will be used if https://pypi.org/project/pyright/ is installed

Found these issues while investigating the work needed for #8686 and #8770

### DIFF
--- a/tests/pyright_test.py
+++ b/tests/pyright_test.py
@@ -1,5 +1,6 @@
 #!/usr/bin/env python3
 
+import os
 import shutil
 import subprocess
 import sys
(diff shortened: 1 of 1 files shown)


AlexWaygood:
I've never run this test locally -- @jakebailey and/or @erictraut, could one of you possibly take a look? :)

erictraut:
This change looks good to me.

AlexWaygood:
> This change looks good to me.

Thank you! And thanks, @Avasam!

PR #8770, opened 2022-09-19
Add reportMissingParameterType to `pyrightconfig.stricter.json`

Avasam:
This was noticed in https://github.com/python/typeshed/pull/8702#discussion_r970089760
Completes the types in keyboard and PyInstaller (which I've both added recently) so that this option can be turned on.


stubs/pyinstaller/PyInstaller/depend/analysis.pyi
```diff
@@ -2,17 +2,25 @@
...
     def __init__(
         self,
         pyi_homepath: str,
-        user_hook_dirs=...,
-        excludes=...,
-        path: Incomplete | None = ...,
-        replace_paths=...,
-        implies=...,
-        graph: Incomplete | None = ...,
+        user_hook_dirs: Iterable[StrOrBytesPath] = ...,
```
JelleZijlstra, line 19: Bytes paths don't work: `PyInstaller.depend.analysis.PyiModuleGraph("x", [b"y"])` throws a TypeError.

stubs/pyinstaller/PyInstaller/depend/analysis.pyi
```diff
@@ -2,17 +2,25 @@
...
+        excludes: Iterable[str] = ...,
+        path: Iterable[str] | None = ...,
```
JelleZijlstra, line 21: These and further args are keyword-only, since they're forwarded through `**kwargs`.

stubs/pyinstaller/PyInstaller/depend/analysis.pyi
Avasam: > TypeError: Can't mix strings and bytes in path components

This is probably the case in other parts of the stubs as well 😓 I'll go back through other `StrOrBytesPath` and make a PR to fix them

stubs/pyinstaller/PyInstaller/depend/analysis.pyi
Avasam: Didn't know [keyword-only arguments](https://peps.python.org/pep-3102/) were a thing. Thanks!

JelleZijlstra (approved):
Looks good now, thanks!

Your other PR introduced a few conflicts, I hope I got the merge right.

PR #8771, opened 2022-09-20
sendmsg can accept None for the address

ccotter:
Per the Python docs and socketmodule.c source, the address argument can be None.


AlexWaygood (approved):
Thanks! I haven't manually verified, but I'll trust the docs and your reading of the source code.

https://docs.python.org/3/library/socket.html#socket.socket.sendmsg

PR #8773, opened 2022-09-20
fix: pillow.ImageDraw.textlength may return float

OmmyZhang:
https://github.com/python-pillow/Pillow/blob/87ecd01fc042bb67b11bfcff3db99fc2b61e702d/src/PIL/ImageDraw.py#L650

I don't think it's necessary, but pillow do return float sometimes.


AlexWaygood (approved):
Thanks! The docs confirm that this is sometimes a `float`: https://pillow.readthedocs.io/en/stable/reference/ImageDraw.html#PIL.ImageDraw.ImageDraw.textlength.

From digging around in the `Pillow` source code, it looks like several other `Pillow` types that we currently have as `int` might be better as `float`. But that's beyond the scope of this PR.

PR #8777, opened 2022-09-21
`pygments.lexers.guess_lexer_for_filename` cannot return `None`

chriskuehl:
This function can raise `ClassNotFound` if it can't find a matching lexer, but it never returns `None`:
https://pypi-browser.org/package/pygments/Pygments-2.13.0-py3-none-any.whl/pygments/lexers/__init__.py#L234


srittau (approved):
Thanks!

PR #8780, opened 2022-09-21
Pyinstaller: use `StrPath` over `StrOrBytesPath`

Avasam:
Changed StrOrBytesPath to StrPath when used with strings.
`TypeError: Can't mix strings and bytes in path components` https://github.com/python/typeshed/pull/8770#discussion_r975969682


JelleZijlstra:
Thanks, confirmed that none of these work with bytes paths.

PR #8785, opened 2022-09-23
Fix intransitive subtyping issue with SupportsGetItem

hauntsaninja:
See https://github.com/python/mypy/issues/13713 for details

### DIFF
--- a/stdlib/_typeshed/__init__.pyi
+++ b/stdlib/_typeshed/__init__.pyi
@@ -7,7 +7,7 @@ import ctypes
 import mmap
 import pickle
 import sys
-from collections.abc import Awaitable, Callable, Container, Iterable, Set as AbstractSet
+from collections.abc import Awaitable, Callable, Iterable, Set as AbstractSet
 from os import PathLike
 from types import FrameType, TracebackType
 from typing import Any, AnyStr, Generic, Protocol, TypeVar, Union
--- a/stdlib/_typeshed/__init__.pyi
+++ b/stdlib/_typeshed/__init__.pyi
@@ -118,7 +118,8 @@ class SupportsKeysAndGetItem(Protocol[_KT, _VT_co]):
     def __getitem__(self, __key: _KT) -> _VT_co: ...
 
 # stable
-class SupportsGetItem(Container[_KT_contra], Protocol[_KT_contra, _VT_co]):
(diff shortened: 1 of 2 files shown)


AlexWaygood:
... why is `typing.Container` even generic in the first place??

hauntsaninja:
Maybe it once had a strict annotation for `__contains__`? Maybe it just felt aesthetically wrong to have a container that wasn't generic. Didn't have time to spelunk. In an ideal world it would be removed, but probably too breaking to touch.

AlexWaygood:
I guess it is pretty nice to say something's a `Container[str]` if it contains strings, even if it is meaningless to the type checker

AlexWaygood:
It might be nice to add a comment to `typing.pyi` noting the oddity, though

hauntsaninja:
I did a spelunk, and it's always been this way. From before typeshed:
https://github.com/python/typeshed/commit/337abed05a35f0c5b67c02d42bca0af1e2adc76e#diff-009a41a8f985ef690b94fb876cfcd4d00fff765eb3f41468a1d816c555bc338bR91
And from the earliest days of mypy:
https://github.com/python/mypy/commit/05f70ca9d99c05a3876488cba255e3d38b9d1da2#diff-e114935a16df41cf681c93cd8be4e3b681194568d45c54f4df7985e149395dbcR265

AlexWaygood:
> I did a spelunk, and it's always been this way. From before typeshed: [337abed#diff-009a41a8f985ef690b94fb876cfcd4d00fff765eb3f41468a1d816c555bc338bR91](https://github.com/python/typeshed/commit/337abed05a35f0c5b67c02d42bca0af1e2adc76e#diff-009a41a8f985ef690b94fb876cfcd4d00fff765eb3f41468a1d816c555bc338bR91) And from the earliest days of mypy: [python/mypy@05f70ca#diff-e114935a16df41cf681c93cd8be4e3b681194568d45c54f4df7985e149395dbcR265](https://github.com/python/mypy/commit/05f70ca9d99c05a3876488cba255e3d38b9d1da2#diff-e114935a16df41cf681c93cd8be4e3b681194568d45c54f4df7985e149395dbcR265)

We did simultaneous spelunking, I arrived at the same conclusion :-)

AlexWaygood (approved):
Looks good!

PR #8787, opened 2022-09-23
Mark object.__module__ as ClassVar

hauntsaninja:
The reported use case for this isn't particularly compelling, so if this causes too many errors I will close.

### DIFF
--- a/stdlib/builtins.pyi
+++ b/stdlib/builtins.pyi
@@ -80,7 +80,7 @@ _AwaitableT_co = TypeVar("_AwaitableT_co", bound=Awaitable[Any], covariant=True)
 class object:
     __doc__: str | None
     __dict__: dict[str, Any]
-    __module__: str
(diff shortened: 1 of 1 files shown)


hauntsaninja:
I'm fine with taking the primer hits, but I'm a little worried about all the stubs that have this.

I'm happy to have `object.__module__` exist, it's in line with several things we do. There's a small chance removing it wouldn't be the worst, though, since we still have `__module__` on `type`.

AlexWaygood:
> I'm happy to have `object.__module__` exist, it's in line with several things we do.

I know, but I'm not sure I'm super jazzed about those things either `¯\_(ツ)_/¯`

AlexWaygood:
> There's a small chance removing it wouldn't be the worst, though, since we still have `__module__` on `type`.

#8789 went down like a bag of sick, so if we want to fix this, this does indeed look like the better solution, at least for now :)

PR #8788, opened 2022-09-24
gdb: add missing automatic imports

pmderodat:
When GDB has just started, several `gdb` submodules are automatically loaded, so user code does not have to manually import them (for instance `import gdb.events`). Reflect that in `gdb` stubs.

### DIFF
--- a/stubs/gdb/gdb/__init__.pyi
+++ b/stubs/gdb/gdb/__init__.pyi
@@ -8,12 +8,13 @@ from contextlib import AbstractContextManager
 from typing import Protocol, overload
 from typing_extensions import TypeAlias
 
-# The following submodules are automatically imported
-import gdb.events as events
-import gdb.printing as printing
-import gdb.prompt as prompt
 import gdb.types
-import gdb.types as types
+
+# The following submodules are automatically imported
+from . import events as events
+from . import printing as printing
+from . import prompt as prompt
+from . import types as types
 
 # Basic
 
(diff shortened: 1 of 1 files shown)


AlexWaygood (commented):
In a stub file, all imports are considered "private to the stub" rather than re-exported, _unless_ you use a "redundant" alias, e.g. `import gdb.events as gdb.events`. See https://peps.python.org/pep-0484/#stub-files. So you probably want to update these imports to use that format, like we do e.g. here:

https://github.com/python/typeshed/blob/c1692130fdd8c74d29f387ef973744728bf27bd0/stdlib/socket.pyi#L13-L15

pmderodat:
I wasn’t aware of this rule: this is now fixed, thanks! For the record I tested locally with mypy 0.971 before submitting this patch: mypy was happy with simple `import gdb.X`, I don’t know if that works as expected…

stubs/gdb/gdb/__init__.pyi
```diff
@@ -8,7 +8,12 @@ from contextlib import AbstractContextManager
 from typing import Protocol, overload
 from typing_extensions import TypeAlias
 
+# The following submodules are automatically imported
+import gdb.events as events
```
AlexWaygood, line 12: This still isn't *quite* right; the name following the `as` needs to be identical to the name before the `as` in order for it to be considered a public re-export. Something like this should work:

```suggestion
from . import events as events
```

Not sure why mypy isn't emitting an error for you locally (might be something to do with config settings or something), but this is how it's *meant* to work :)

pmderodat: Ah right, I should have read the PEP more carefully. This should be fixed now, thank you.

AlexWaygood (approved):
Thanks!

pmderodat:
Great, thank you!

PR #8789, opened 2022-09-24
Delete `object.__module__`

AlexWaygood:
An alternative approach to #8787


AlexWaygood:
Hmm, 1868 stubtest errors

AlexWaygood:
Seems like mypy struggles to infer that the `__module__` attribute is inherited from `type` for class objects

PR #8792, opened 2022-09-26
boto: drop dependency on types-six

hauntsaninja:
For https://github.com/typeshed-internal/stub_uploader/pull/61#discussion_r979327370

### DIFF
--- a/stubs/boto/METADATA.toml
+++ b/stubs/boto/METADATA.toml
@@ -1,2 +1,2 @@
 version = "2.49.*"
-requires = ["types-six"]
+requires = []
--- a/stubs/boto/boto/connection.pyi
+++ b/stubs/boto/boto/connection.pyi
@@ -1,6 +1,5 @@
+import http.client
 from typing import Any
-
-from six.moves import http_client
 
 HAVE_HTTPS_CONNECTION: bool
 ON_APP_ENGINE: Any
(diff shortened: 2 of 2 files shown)


stubs/boto/METADATA.toml
```diff
@@ -1,2 +1,2 @@
 version = "2.49.*"
-requires = ["types-six"]
+requires = []
```
AlexWaygood, line 2:
```suggestion
```

AlexWaygood (approved):
Just to state explicitly what's going on here:

1. `boto` at runtime does, in fact, still use `six` internally 
2. But typeshed only supports Python 3, and `six.http_client` is the same as `http.client` on Python 3
3. So that we can allow non-types dependencies without introducing a security hole, it's useful to ban stubs from having any dependencies if the runtime package they're stubs for doesn't also have that dependency. This applies to stub dependencies as well, since they're also now potentially dangerous given that stubs can now have non-stubs dependencies.
4. `boto` doesn't actually, officially, depend on `six` at runtime, it just vendors `six`. This means that the machinery in the stub-uploader can't detect that `boto` uses `six` at runtime, ergo `types-boto` also cannot depend on `types-six`. And because of point (2) above, it's not a big deal to remove the dependency.

PR #8793, opened 2022-09-26
google-cloud-ndb: drop dependency on types-six

hauntsaninja:
For https://github.com/typeshed-internal/stub_uploader/pull/61#discussion_r979327370


AlexWaygood (approved):
Maybe we could have some kind of lint detecting entirely unused dependencies

PR #8794, opened 2022-09-26
Run stub_uploader tests in CI

hauntsaninja:
stub_uploader is developing a bunch of validation, a lot of which must live in that repo for security reasons. We don't want to find out about this validation failing when stub uploading fails.

We also run it in the daily test since there are some checks that (unfortunately, albeit unlikely) can be broken by new upstream releases.

This is blocked on https://github.com/typeshed-internal/stub_uploader/pull/60


AlexWaygood (approved):

.github/workflows/daily.yml
```diff
@@ -58,12 +58,34 @@ jobs:
...
+          path: typeshed
+      - name: Checkout stub_uploader
+        uses: actions/checkout@v3
+        with:
+          repository: typeshed-internal/stub_uploader
+          path: stub_uploader
+      - uses: actions/setup-python@v4
+        with:
+          python-version: "3.10"
+      - run: |
```
AlexWaygood, line 78:
```suggestion
      - name: Run tests
        run: |
```

.github/workflows/tests.yml
```diff
@@ -125,6 +125,28 @@ jobs:
...
+          path: typeshed
+      - name: Checkout stub_uploader
+        uses: actions/checkout@v3
+        with:
+          repository: typeshed-internal/stub_uploader
+          path: stub_uploader
+      - uses: actions/setup-python@v4
+        with:
+          python-version: "3.10"
+      - run: |
```
AlexWaygood, line 145:
```suggestion
      - name: Run tests
        run: |
```

AlexWaygood (commented):
(Very minor: The name of the part of the GH Actions job where we actually run the tests is currently `cd stub_uploader`)

PR #8796, opened 2022-09-26
Bump mypy to 0.981

AlexWaygood:
Once I've got stdlib stubtest passing, I'll hit this PR with third-party stubtest as well, since there have been a fair few stubtest changes in this release.


stdlib/builtins.pyi
```diff
@@ -1952,7 +1952,7 @@ if sys.version_info >= (3, 11):
         @overload
         def subgroup(self: Self, __condition: Callable[[_BaseExceptionT_co], bool]) -> Self | None: ...
         @overload
-        def split(
+        def split(  # type: ignore[misc]  # complaints about overlapping overloads
```
AlexWaygood, line 1955: I believe these new mypy errors are already fixed on mypy master, by https://github.com/python/mypy/pull/13461. Unfortunately, it doesn't look like that PR made it into 0.981 (cc. @hauntsaninja, @ilevkivskyi)

AlexWaygood:
This PR should now be ready for review!

stdlib/importlib/metadata/__init__.pyi
```diff
@@ -41,6 +41,9 @@ class _EntryPointBase(NamedTuple):
 
 class EntryPoint(_EntryPointBase):
     pattern: ClassVar[Pattern[str]]
+    if sys.version_info >= (3, 11):
+        def __init__(self, name: str, value: str, group: str) -> None: ...
+
```
AlexWaygood, line 46: The allowlist entry had a comment saying that it was to do with `Protocol.__init__`, so I thought I could safely remove the allowlist entry now that we were bumping to mypy 0.981. Actually the comment was wrong, the allowlist entry had nothing to do with `Protocol.__init__`. But I figured fixing the allowlist entry was just as easy as correcting the comment next to the entry.

stubs/html5lib/html5lib/_inputstream.pyi
```diff
@@ -72,7 +72,7 @@ class HTMLBinaryInputStream(HTMLUnicodeInputStream):
     def detectEncodingMeta(self): ...
 
 class EncodingBytes(bytes):
-    def __new__(cls, value): ...
+    def __new__(self, value): ...
```
AlexWaygood, line 75: This change isn't strictly related to the mypy bump, but it was a very easy allowlist entry to fix, and it means that we can delete the whole allowlist for `html5lib` since the other allowlist entry is now unused as of mypy 0.981.

JelleZijlstra (approved):

stdlib/builtins.pyi
ilevkivskyi: cc @jhance if people will complain about this, we will need to CP it and release 0.982.

AlexWaygood: @Molkree thanks!

AlexWaygood: Mypy was successfully bumped to 0.982 in #8831; the `type ignore`s are gone again 🎉

Issue #8797, opened 2022-09-27
`tests/mypy_test.py` fails locally with stubs that have `requires`


Avasam:
Running the following locally: `py .\tests\mypy_test.py stubs/D3DShot`

Results in:
```
failure

C:\Users\Avasam\AppData\Local\Temp\tmpiqiqpwq3: No [mypy] section in config file

stubs\D3DShot\d3dshot\capture_output.pyi:7: error: Cannot find implementation or library stub for module named "PIL"  [import]
stubs\D3DShot\d3dshot\capture_output.pyi:7: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
stubs\D3DShot\d3dshot\capture_outputs\pytorch_capture_output.pyi:7: error: Cannot find implementation or library stub for module named "PIL"  [import]
stubs\D3DShot\d3dshot\capture_outputs\pil_capture_output.pyi:7: error: Cannot find implementation or library stub for module named "PIL"  [import]
stubs\D3DShot\d3dshot\capture_outputs\numpy_capture_output.pyi:7: error: Cannot find implementation or library stub for module named "PIL"  [import]
```

PR #8798, opened 2022-09-27
Script to run all checks locally

### LINKED ISSUE #8686: Single script to run PR checks locally

Avasam:
Closes #8686

There's still an issue that needs to be fixed with the mypy test (see #8797), and I would like to be able to seamlessly run the pytype test on windows, if that's even feasible.

Summary output: 
image

Please let me know how to better word the "next steps" in `create_baseline_stubs`.

Notes of things to do in follow-up PR(s):
- [ ] Different return codes for different errors in pyright_test and regr_test (https://github.com/python/typeshed/pull/8798#discussion_r983451283)
- [ ] Run pytype_test using WSL on Windows if a distribution is installed (https://github.com/python/typeshed/pull/8798#discussion_r981690202)
- [ ] Run tests for more than 1 stubs (https://github.com/python/typeshed/pull/8798#discussion_r982222296)
- [ ] Improve windows docs (https://github.com/python/typeshed/pull/8798#discussion_r985272898)
- [ ] Consider adding an optional CLI argument to automatically run or skip third-party stubtest (https://github.com/python/typeshed/pull/8798#discussion_r985284856)


scripts/create_baseline_stubs.py
```diff
@@ -169,11 +169,10 @@ def main() -> None:
...
     print(f" 1. Manually review the generated stubs in {stub_dir}")
-    print(f' 2. Run "MYPYPATH={stub_dir} python3 -m mypy.stubtest {package}" to check the stubs against runtime')
-    print(f' 3. Run "mypy {stub_dir}" to check for errors')
-    print(f' 4. Run "black {stub_dir}" and "isort {stub_dir}" (if you\'ve made code changes)')
-    print(f' 5. Run "flake8 {stub_dir}" to check for e.g. unused imports')
-    print(" 6. Commit the changes on a new branch and create a typeshed PR")
+    print(" 2. See CONTRIBUTING.md for all the best practices and tips to improve your stub")
+    print(' 3. Run autofixes and tests locally using "python3 -m ./scripts/run_all_checks.py"')
+    print("    See tests/README.md to run individual tests")
+    print(" 4. Commit the changes on a new branch and create a typeshed PR (don't force-push!)")
```
AlexWaygood, line 175: I still don't think we should be telling contributors that they _have_ to run the tests prior to submitting a PR -- if they _want_ to run the tests first, that's fine, but I think it should also be absolutely fine if they just want to rely on CI :)

Also, I'd rather keep all our documentation on tests in one place -- would you mind adding a short description of the new script to `tests/README.md`?

```suggestion
    print(" 2. Optionally run tests and autofixes (see tests/README.md for details")
    print(" 3. Commit the changes on a new branch and create a typeshed PR (don't force-push!)")
```

scripts/run_all_checks.py
```diff
@@ -0,0 +1,63 @@
...
+        subprocess.run(["python3", "tests/regr_test.py", "stdlib"])
+    else:
+        subprocess.run(["python3", "tests/regr_test.py", stub])
+
+    # If mypy failed, this will fail without any helpful error
+    if mypy_result.returncode == 0:
+        if folder == "stdlib":
+            subprocess.run(["python3", "tests/stubtest_stdlib.py", stub])
+        else:
+            subprocess.run(["python3", "tests/stubtest_third_party.py", stub])
```
AlexWaygood, line 63: It would be good to have a `0` exit code if all checks succeeded, and a `1` exit code if one or more checks failed.

AlexWaygood (changes requested):
Thanks! This looks like a great start.

I have a few questions, a few suggestions, and a few nits ;)

scripts/run_all_checks.py
```diff
@@ -0,0 +1,63 @@
...
+    assert folder in {"stdlib", "stubs"}, "Only the 'stdlib' and 'stubs' folders are supported"
+
+    subprocess.run(["python3", "-m", "black", path])
+    subprocess.run(["python3", "-m", "isort", path])
+    subprocess.run(["python3", "-m", "flake8", path, "--extend-ignore", _extend_ignore])
+
+    subprocess.run(["python3", "tests/check_consistent.py"])
+    subprocess.run(["python3", "tests/check_new_syntax.py"])
+
+    subprocess.run(["python3", "tests/pyright_test.py", path] + _get_strict_params(path))
```
AlexWaygood, line 44: I'd prefer it if this didn't fail if pyright's not installed. E.g. I don't have pyright installed locally in my typeshed environment -- I've tried once or twice, but it didn't work immediately, I didn't bother to investigate why, and relying on pyright in CI is good enough for me ;)

scripts/run_all_checks.py
```diff
@@ -0,0 +1,63 @@
+import json
```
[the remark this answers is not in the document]
Avasam: It makes sense to me to place scripts that I use for a repo in a scripts folder, as to not clutter the root any more than it usually is with project-wide config files. It also makes them easier to find.
But that's just how I do it in *my* projects. I can move it to wherever works best for typeshed.

I'll change the name, it's more concise. I don't tend to mind long names because of tab-completion.

scripts/run_all_checks.py
```diff
@@ -0,0 +1,63 @@
...
+
+    subprocess.run(["python3", "tests/check_consistent.py"])
+    subprocess.run(["python3", "tests/check_new_syntax.py"])
+
+    subprocess.run(["python3", "tests/pyright_test.py", path] + _get_strict_params(path))
+
+    # FIXME: Fails to see requires
+    mypy_result = subprocess.run(["python3", "tests/mypy_test.py", path])
+
+    # TODO: Run in WSL if available
```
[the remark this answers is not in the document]
Avasam: Just confirmed it says "linux"

scripts/run_all_checks.py
Avasam: Are you asking this *specifically* because pyright may be slightly different than others? (uses node and is installed in the global `node_modules` instead of python in a virtual env).

If pyright fails to run (not installed, npx not found, etc.), you already get a proper message from `pyright_test.py` and it doesn't stop the execution.

See how I can't run mypy locally for example (because of the mentioned issue), but this entire script still completes.

I don't really have a way of differentiating between "failed pyright test" and "pyright failed to run" that I can think of. `shutil.which("pyright")` doesn't make sense if ran through npx, and I don't think a `shutil.which("npx")` check would change anything for you.

AlexWaygood: > Are you asking this _specifically_ because pyright may be slightly different than others?

Yeah, pretty much — if you run `pip install -r requirements-tests.txt`, it's the only one that won't _necessarily_ be installed locally.

> I don't really have a way of differentiating between "failed pyright test" and "pyright failed to run" that I can think of. `shutil.which("pyright")` doesn't make sense if ran through npx, and I don't think a `shutil.which("npx")` check would change anything for you.

Fair enough :) I haven't actually tried running this script yet — I'll give it a go soon and see if I have any specific suggestions

scripts/run_all_checks.py
```diff
@@ -0,0 +1,63 @@
...
+    subprocess.run(["python3", "-m", "black", path])
+    subprocess.run(["python3", "-m", "isort", path])
+    subprocess.run(["python3", "-m", "flake8", path, "--extend-ignore", _extend_ignore])
+
+    subprocess.run(["python3", "tests/check_consistent.py"])
+    subprocess.run(["python3", "tests/check_new_syntax.py"])
+
+    subprocess.run(["python3", "tests/pyright_test.py", path] + _get_strict_params(path))
+
+    # FIXME: Fails to see requires
```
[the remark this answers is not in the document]
AlexWaygood: Nope, this is a "Yet again, Alex broke typeshed's test suite when refactoring stuff" problem :)

scripts/run_all_checks.py
Avasam: Not only that, but I've added a summary (see image in original comment) of which test passed, which didn't, and which were completely skipped.
By "skipped" I mean the test wasn't even run. Mostly because we know it's not possible to(pytype, or stubtest when mypy). So we can't know the true result.

Something like "no regression test for this stub" is considered a success, as it shows it will pass on the CI. (and "0 out of  0" is still 0 failures)

scripts/run_all_checks.py
Avasam: Can you give me the output (stderr) you get when running `pyright_test.py`? Since I now show which tests are skipped because they can't be ran locally, (and I learned how to pipe the output), I can check for npx and whatever error you get.

AlexWaygood: The output printed to stderr on my machine is:

```
error finding npx; is Node.js installed?\n
```

scripts/run_all_checks.py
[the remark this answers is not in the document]
AlexWaygood: We can safely remove this TODO, in that case — `pytype_test.py` should be run under this script if you're running under WSL

```suggestion
```

scripts/run_all_checks.py
AlexWaygood: Nice!

scripts/run_all_checks.py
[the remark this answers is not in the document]
Avasam: I was more thinking of calling wsl and passing it the command:
`wsl --list` (check if output it starts with `Windows Subsystem for Linux Distributions:` or a help message.
Then `wsl --cd current/typeshed/directory python3 test/pytype_test.py`.

But that could be done separately. Maybe even as part of the pytype script for ease of use on windows.
A nice to have, but probably not worth the trouble for this PR.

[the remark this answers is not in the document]
AlexWaygood: Oh, interesting idea! But yeah, as you say — better to fix that in `pytype test.py` itself rather than in this script :)

scripts/create_baseline_stubs.py
```diff
@@ -169,11 +169,8 @@ def main() -> None:
 
     print("\nDone!\n\nSuggested next steps:")
     print(f" 1. Manually review the generated stubs in {stub_dir}")
-    print(f' 2. Run "MYPYPATH={stub_dir} python3 -m mypy.stubtest {package}" to check the stubs against runtime')
-    print(f' 3. Run "mypy {stub_dir}" to check for errors')
-    print(f' 4. Run "black {stub_dir}" and "isort {stub_dir}" (if you\'ve made code changes)')
-    print(f' 5. Run "flake8 {stub_dir}" to check for e.g. unused imports')
-    print(" 6. Commit the changes on a new branch and create a typeshed PR")
+    print(" 2. Optionally run tests and autofixes (see tests/README.md for details")
```
AlexWaygood, line 172: Still need to add a short description of this script to the `tests/README.md` file :)

scripts/runtests.py
```diff
@@ -0,0 +1,128 @@
...
+        path = sys.argv[1]
+    except IndexError:
+        print("Missing path argument in format <folder>/<stub>", file=sys.stderr)
+        sys.exit(1)
+    path_tokens = Path(path).parts
+    assert len(path_tokens) == 2, "Path argument should be in format <folder>/<stub>"
+    folder, stub = path_tokens
+    assert folder in {"stdlib", "stubs"}, "Only the 'stdlib' and 'stubs' folders are supported"
+    stubtest_result: Optional[subprocess.CompletedProcess[bytes]] = None
+    pytype_result: Optional[subprocess.CompletedProcess[bytes]] = None
```
AlexWaygood, line 50: If you add `from __future__ import annotations` at the top of the file, we can use PEP 604 here:

```suggestion
    stubtest_result: subprocess.CompletedProcess[bytes] | None = None
    pytype_result: subprocess.CompletedProcess[bytes] | None = None
```

scripts/runtests.py
```diff
@@ -0,0 +1,128 @@
...
+    with open(_STRICTER_CONFIG_FILE) as file:
+        data = json.loads(_parse_jsonc(file.read()))
+    if stub_path in data["exclude"]:
+        return []
+    return ["-p", _STRICTER_CONFIG_FILE]
+
+
+if __name__ == "__main__":
+    try:
+        path = sys.argv[1]
```
AlexWaygood, line 41: In future PRs, we could consider adding more command-line arguments. It might be nice to have an argument to run *all the tests*, not just the tests for a specific stubs directory. That would have been useful for a PR like #8796, which involved changes across a wide swathe of typeshed.

But I think we should keep this PR simple, and not do that for now :)

scripts/runtests.py
```diff
@@ -0,0 +1,128 @@
...
+    subprocess.run(["python3", "-m", "pycln", path, "--all"])
+    subprocess.run(["python3", "-m", "isort", path])
+    subprocess.run(["python3", "-m", "black", path])
+
+    flake8_result = subprocess.run(["python3", "-m", "flake8", path])
+
+    check_consistent_result = subprocess.run(["python3", "tests/check_consistent.py"])
+    check_new_syntax_result = subprocess.run(["python3", "tests/check_new_syntax.py"])
+
+    pyright_result = subprocess.run(["python3", "tests/pyright_test.py", path] + _get_strict_params(path), stderr=subprocess.PIPE)
```
AlexWaygood, line 62: I just tried running the script locally in my virtual environment, and this was the output I got:

```console
(.venv) C:\Users\alexw\coding\typeshed>python scripts/runtests.py stubs/sqlalchemy
Python was not found; run without arguments to install from the Microsoft Store, or disable this shortcut from Settings > Manage App Execution Aliases.
Python was not found; run without arguments to install from the Microsoft Store, or disable this shortcut from Settings > Manage App Execution Aliases.
Python was not found; run without arguments to install from the Microsoft Store, or disable this shortcut from Settings > Manage App Execution Aliases.
Python was not found; run without arguments to install from the Microsoft Store, or disable this shortcut from Settings > Manage App Execution Aliases.
Python was not found; run without arguments to install from the Microsoft Store, or disable this shortcut from Settings > Manage App Execution Aliases.
Python was not found; run without arguments to install from the Microsoft Store, or disable this shortcut from Settings > Manage App Execution Aliases.
Python was not found; run without arguments to install from the Microsoft Store, or disable this shortcut from Settings > Manage App Execution Aliases.

Python was not found; run without arguments to install from the Microsoft Store, or disable this shortcut from Settings > Manage App Execution Aliases.
Skipping stubtest since mypy failed.
Skipping pytype on Windows. You can run the test with WSL.
Python was not found; run without arguments to install from the Microsoft Store, or disable this shortcut from Settings > Manage App Execution Aliases.

One or more tests failed. See above for details.
flake8: Failed
Check consistent: Failed
Check new syntax: Failed
pyright: Failed
mypy: Failed
stubtest: Skipped
pytype: Skipped
Regression test: Failed
```

You need to use `sys.executable` instead of hardcoding `python3` in these `subprocess.run()` calls:

```suggestion
    subprocess.run([sys.executable, "-m", "pycln", path, "--all"])
    subprocess.run([sys.executable, "-m", "isort", path])
    subprocess.run([sys.executable, "-m", "black", path])

    flake8_result = subprocess.run([sys.executable, "-m", "flake8", path])

    check_consistent_result = subprocess.run([sys.executable, "tests/check_consistent.py"])
    check_new_syntax_result = subprocess.run([sys.executable, "tests/check_new_syntax.py"])

    pyright_result = subprocess.run([sys.executable, "tests/pyright_test.py", path] + _get_strict_params(path), stderr=subprocess.PIPE)
```

scripts/runtests.py
```diff
@@ -0,0 +1,128 @@
...
+    print(pyright_result.stderr.decode())
+    if b"error running npx" in pyright_result.stderr:
+        pyright_returncode = 0
+        pyright_skipped = True
+    else:
+        pyright_returncode = pyright_result.returncode
+        pyright_skipped = False
+
+    mypy_result = subprocess.run(["python3", "tests/mypy_test.py", path])
```
AlexWaygood, line 71:
```suggestion
    mypy_result = subprocess.run([sys.executable, "tests/mypy_test.py", path])
```

scripts/runtests.py
```diff
@@ -0,0 +1,128 @@
...
+    # If mypy failed, stubtest will fail without any helpful error
+    if mypy_result.returncode == 0:
+        if folder == "stdlib":
+            stubtest_result = subprocess.run(["python3", "tests/stubtest_stdlib.py", stub])
+        else:
+            stubtest_result = subprocess.run(["python3", "tests/stubtest_third_party.py", stub])
```
AlexWaygood, line 77:
```suggestion
            stubtest_result = subprocess.run([sys.executable, "tests/stubtest_stdlib.py", stub])
        else:
            stubtest_result = subprocess.run([sys.executable, "tests/stubtest_third_party.py", stub])
```

scripts/runtests.py
```diff
@@ -0,0 +1,128 @@
...
+
+    if sys.platform == "win32":
+        print("Skipping pytype on Windows. You can run the test with WSL.")
+    else:
+        pytype_result = subprocess.run(["python3", "tests/pytype_test.py", path])
+
+    if folder == "stdlib":
+        regr_test_result = subprocess.run(["python3", "tests/regr_test.py", "stdlib"], stderr=subprocess.PIPE)
+    else:
+        regr_test_result = subprocess.run(["python3", "tests/regr_test.py", stub], stderr=subprocess.PIPE)
```
AlexWaygood, line 89:
```suggestion
        pytype_result = subprocess.run([sys.executable, "tests/pytype_test.py", path])

    if folder == "stdlib":
        regr_test_result = subprocess.run([sys.executable, "tests/regr_test.py", "stdlib"], stderr=subprocess.PIPE)
    else:
        regr_test_result = subprocess.run([sys.executable, "tests/regr_test.py", stub], stderr=subprocess.PIPE)
```

scripts/runtests.py
```diff
@@ -0,0 +1,128 @@
...
+        flake8_result.returncode
+        + check_consistent_result.returncode
+        + check_new_syntax_result.returncode
+        + pyright_returncode
+        + mypy_result.returncode
+        + (stubtest_result.returncode if stubtest_result else 0)
+        + (pytype_result.returncode if pytype_result else 0)
+        + regr_test_returncode
+        > 0
+    )
```
AlexWaygood, line 104:
```suggestion
    any_failure = any(
        [
            flake8_result.returncode,
            check_consistent_result.returncode,
            check_new_syntax_result.returncode,
            pyright_returncode,
            mypy_result.returncode,
            getattr(stubtest_result, "returncode", 0),
            getattr(pytype_result, "returncode", 0),
            regr_test_returncode
        ]
    )
```

AlexWaygood (changes requested):

scripts/runtests.py
```diff
@@ -0,0 +1,128 @@
...
+        print("Skipping pytype on Windows. You can run the test with WSL.")
+    else:
+        pytype_result = subprocess.run(["python3", "tests/pytype_test.py", path])
+
+    if folder == "stdlib":
+        regr_test_result = subprocess.run(["python3", "tests/regr_test.py", "stdlib"], stderr=subprocess.PIPE)
+    else:
+        regr_test_result = subprocess.run(["python3", "tests/regr_test.py", stub], stderr=subprocess.PIPE)
+    print(regr_test_result.stderr.decode())
+    # No test means they all ran successfully (0 out of 0). Very few 3rd part stub have regression tests.
```
AlexWaygood, line 91: This comment feels at risk of getting out of date :)

```suggestion
    # No test means they all ran successfully (0 out of 0). Not all 3rd-party stubs have regression tests.
```

scripts/runtests.py
Avasam: I forgot Microsoft did that weird thing where python3 is just an executable that references/launches the microsoft store... (Until you install python from the store).

Wouldn't that happen in other scripts as well? It doesn't happen for me because I overwrote that shortcut.

AlexWaygood: > Wouldn't that happen in other scripts as well?

Yup, that's why the other scripts in typeshed mostly use `sys.executable` in `subprocess.run()` calls :)

I guess our docs aren't ideal for Windows users, given that they recommend running the scripts with `python3`. But docs get really long and convoluted if you spell out the platform-specific ways of doing each and every thing ://

scripts/runtests.py
Avasam: Definitely better.

scripts/create_baseline_stubs.py
Avasam: Done! As usual I'm open to suggestion to improve the wording.

AlexWaygood:
<details>
<summary>Here's the output currently if I try running `python scripts/runtests.py redis`:</summary>

```console
(.venv) C:\Users\alexw\coding\typeshed>python scripts/runtests.py stubs/redis
Looks good! ✨ 🍰 ✨
50 files left unchanged.
All done! ✨ 🍰 ✨
50 files left unchanged.
error finding npx; is Node.js installed?

Running mypy --python-version 3.11 --show-traceback --warn-incomplete-stub --show-error-codes --no-error-summary --platform win32 --no-site-packages --custom-typeshed-dir C:\Users\alexw\coding\typeshed --no-implicit-optional --disallow-untyped-decorators --disallow-any-generics --strict-equality --enable-error-code ignore-without-code --config-file /tmp/...
testing redis (50 files)... success

*** Testing Python 3.10 on win32
Running mypy --python-version 3.10 --show-traceback --warn-incomplete-stub --show-error-codes --no-error-summary --platform win32 --no-site-packages --custom-typeshed-dir C:\Users\alexw\coding\typeshed --no-implicit-optional --disallow-untyped-decorators --disallow-any-generics --strict-equality --enable-error-code ignore-without-code --config-file /tmp/...
testing redis (50 files)... success

*** Testing Python 3.9 on win32
Running mypy --python-version 3.9 --show-traceback --warn-incomplete-stub --show-error-codes --no-error-summary --platform win32 --no-site-packages --custom-typeshed-dir C:\Users\alexw\coding\typeshed --no-implicit-optional --disallow-untyped-decorators --disallow-any-generics --strict-equality --enable-error-code ignore-without-code --config-file /tmp/...
testing redis (50 files)... success

*** Testing Python 3.8 on win32
Running mypy --python-version 3.8 --show-traceback --warn-incomplete-stub --show-error-codes --no-error-summary --platform win32 --no-site-packages --custom-typeshed-dir C:\Users\alexw\coding\typeshed --no-implicit-optional --disallow-untyped-decorators --disallow-any-generics --strict-equality --enable-error-code ignore-without-code --config-file /tmp/...
testing redis (50 files)... success

*** Testing Python 3.7 on win32
Running mypy --python-version 3.7 --show-traceback --warn-incomplete-stub --show-error-codes --no-error-summary --platform win32 --no-site-packages --custom-typeshed-dir C:\Users\alexw\coding\typeshed --no-implicit-optional --disallow-untyped-decorators --disallow-any-generics --strict-equality --enable-error-code ignore-without-code --config-file /tmp/...
testing redis (50 files)... success

--- success, 250 files checked ---
redis... success
Skipping pytype on Windows. You can run the test with WSL.
usage: regr_test.py [-h] [--all] [--platform [{linux,darwin,win32} ...]] [-p [{3.11,3.10,3.9,3.8,3.7} ...]] [packages_to_test ...]
regr_test.py: error: argument packages_to_test: No test cases found for 'redis'!

One or more tests failed. See above for details.
flake8: Success
Check consistent: Success
Check new syntax: Success
pyright: Failed
mypy: Success
stubtest: Success
pytype: Skipped
Regression test: Success
```

</details>

Some notes:

- It's not obvious currently which tests are running when. `pycln` and `black` have very similar "success!!1!" messages, for example, and `isort` doesn't have one at all. It might be nice to add some `print()` statements before each `subprocess.run()` call: "Running black...", etc.
- It looks like `pyright_test.py` is still marked as "Failed" if npx is not available; it should be marked as "skipped".
- `regr_test.py` is correctly marked as "Skipped", but the error message from argparse is still printed to the terminal, which makes it look like it failed. If we use `capture_output=True` in the `subprocess.run()` call, we can stop any output being printed to the screen. We can then manually print the captured output in `stdout` and `stderr`, if and only if test cases were found.

tests/README.md
```diff
@@ -19,6 +19,15 @@ in the `tests` and `scripts` directories.
...
 in the `CONTRIBUTING.md` document. In particular, we recommend running with Python 3.9+.
 
+## Run all tests for a specific stub
+
+Run using:
+```
+(.venv3)$ python3 scripts/runtests.py <stdlib-or-stubs>/<stub-to-test>
+```
+
+This script will run all tests below for a specific stub with a summary of the results. You must provide a single argument which is a path to the stub to test, like so: `stblib/os` or `stubs/requests`.
```
AlexWaygood, line 29:
```suggestion
This script will run all tests below for a specific typeshed directory. A summary of the results will be printed to the terminal.

You must provide a single argument which is a path to the stubs to test, like so: `stdlib/os` or `stubs/requests`.
```

scripts/runtests.py
```diff
@@ -0,0 +1,132 @@
...
+    flake8_result = subprocess.run([sys.executable, "-m", "flake8", path])
+
+    check_consistent_result = subprocess.run([sys.executable, "tests/check_consistent.py"])
+    check_new_syntax_result = subprocess.run([sys.executable, "tests/check_new_syntax.py"])
+
+    pyright_result = subprocess.run(
+        [sys.executable, "tests/pyright_test.py", path] + _get_strict_params(path), stderr=subprocess.PIPE
+    )
+    print(pyright_result.stderr.decode())
+    if b"error running npx" in pyright_result.stderr:
```
AlexWaygood, line 67:
```suggestion
    if b"error finding npx" in pyright_result.stderr:
```

scripts/runtests.py
```diff
@@ -0,0 +1,132 @@
...
+
+    if sys.platform == "win32":
+        print("Skipping pytype on Windows. You can run the test with WSL.")
+    else:
+        pytype_result = subprocess.run([sys.executable, "tests/pytype_test.py", path])
+
+    if folder == "stdlib":
+        regr_test_result = subprocess.run([sys.executable, "tests/regr_test.py", "stdlib"], stderr=subprocess.PIPE)
+    else:
+        regr_test_result = subprocess.run([sys.executable, "tests/regr_test.py", stub], stderr=subprocess.PIPE)
```
AlexWaygood, line 92: Question: is the goal for this script to be a "convenience script", to run the most important checks? Or is it to be an "exhaustive script": to run _all_ the checks?

If it's a "convenience script", then we maybe don't need `mypy_test.py` to run on _all_ Python versions -- that takes forever, and usually isn't necessary. We could run `python tests/mypy_test.py -p <version-I'm-running>`.

If it's an "exhaustive script", then we should run `python tests/regr_test.py --all`, so it runs on _all_ platforms/python versions.

Avasam: I started this script with the idea that if it passes, the CI should pass. It's the convenience of not having to think about how each individual test expect their arguments, and not having to wait on the CI to tell you it failed many minutes later.

It's a valid question and I don't have a definite answer for it. It could be both, but then we have to decide which goal is a priority (and doesn't require an extra argument, wanting to keep this as dead simple to use as possible).

scripts/runtests.py
Avasam: Ah! there's two different error messages! I swear I kept reading it as the same.

scripts/runtests.py
```diff
@@ -0,0 +1,132 @@
...
+    print(regr_test_result.stderr.decode())
+    # No test means they all ran successfully (0 out of 0). Not all 3rd-party stubs have regression tests.
+    regr_test_returncode = 0 if b"No test cases found" in regr_test_result.stderr else regr_test_result.returncode
```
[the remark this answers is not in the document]
Avasam: Since it's the 3rd thing we say we could do in a follow-up PR, I've noted them in the PR description, so we can remember to create appropriate issues/PRs for them.

scripts/runtests.py
AlexWaygood: Note also that by default `mypy_test.py` only runs with `--platform <platform-I'm- running`, so if we want to make this script truly exhaustive, we should iterate through all platforms on `mypy_test.py` as well :)

Personally, I'd go for making this a convenience script. It will still pick up on 95% of errors that will be flagged by CI. And emulating _all_ parts of our CI locally just isn't possible (unless you want to write a script to spin up a virtual machine to run `stubtest_stdlib.py` on linux, Windows and macOS, as well as whatever system you're running on locally ;).

Avasam: Food for thought: stubtest may be run in a different environment than the CI. Which means the results won't always 100% match up (but may allows the user to complete non-linux os-specific stubs that the CI may fail on, like I did with PyAutogui). That being said, this probably won't be true long-term.

As for mypy, I was gonna make a similar comment as you did. Should we still let it run on all python versions? It doesn't take *that* long. If we choose a specific version, either the current one being execute, or 3.7 since it has the most errors (mainly 3.10 union syntax in type alias)

AlexWaygood: My preference would be to run `mypy_test.py`, `regr_test.py` and `pyright_test.py` all with the same Python version -- I think that's much easier for users to remember, increasing the simplicity of the script.

And good shout on using 3.7! We should add a comment to note why we're using 3.7 as opposed to any other Python version.

scripts/runtests.py
```diff
@@ -0,0 +1,149 @@
...
+    mypy_result = subprocess.run([sys.executable, "tests/mypy_test.py", path])
+    # If mypy failed, stubtest will fail without any helpful error
+    if mypy_result.returncode == 0:
+        print("\nRunning stubtest...")
+        if folder == "stdlib":
+            stubtest_result = subprocess.run([sys.executable, "tests/stubtest_stdlib.py", stub])
+        else:
+            stubtest_result = subprocess.run([sys.executable, "tests/stubtest_third_party.py", stub])
+    else:
+        print("Skipping stubtest since mypy failed.")
```
AlexWaygood, line 94: I wonder if this is possibly a little dangerous. `stubtest_third_party.py` installs packages into an isolated virtual environment, but it's still executing arbitrary code that's been downloaded from PyPI. That's not really an issue when you're doing it in CI (we can set appropriate permissions that mean that there's only so much damage it can do). But maybe it shouldn't be run by default in this script, given that it's the recommended "easy way" for new contributors to run all checks locally, if that's something they want to do.

@hauntsaninja, what do you think?

scripts/runtests.py
```diff
@@ -0,0 +1,149 @@
...
+            getattr(stubtest_result, "returncode", 0),
+            getattr(pytype_result, "returncode", 0),
+            regr_test_returncode,
+        ]
+    )
+
+    if any_failure:
+        print("\nOne or more tests failed. See above for details.")
+    else:
+        print("\nAll tests passed!")
```
AlexWaygood, line 130: Makes it a little more obvious where the output from the various scripts ends, and where the summary of all the tests begins.

```suggestion
    if any_failure:
        print(colored("\n\n--- TEST SUMMARY: One or more tests failed. See above for details. ---\n\n", "red"))
    else:
        print(colored("\n\n--- TEST SUMMARY: All tests passed! ---\n\n", "green"))
```

scripts/runtests.py
AlexWaygood: The same issue doesn't really apply to `stubtest_stdlib.py` — we can be reasonably confident that stubtest can import and inspect the stdlib without there being any adverse side effects on the computer stubtest is running on.

scripts/runtests.py
Avasam: Note added to the readme that the oldest version supported by typeshed will be used. The script also shows it to the user. And added a comment in the script explaining why (without mentioning any specific issue so that the comment doesn't get outdated)

Avasam:
> (Nit): we currently get a bit of an ugly traceback if the script is interrupted halfway through:
> [...]

A traceback can still appear, but it's clearer why.

image

Unless maybe we start piping the output of all tests, checking for a KeyboardInterrupt and *then* printing the output. I don't think that's really worth it.  
As a user, you *did* interrupt a script mid-execution after all.

scripts/runtests.py
Avasam: Btw I'm having issues running stubtest in a virtual environment (".venv"), so I'm already having to run it separately already. I'll raise an issue for that.
Sidenote, there's no mention of such a concern in `tests/README.md`, so if it's valid, it should probably be added there as well.

AlexWaygood: > there's no mention of such a concern in `tests/README.md`, so if it's valid, it should probably be added there as well.

Agreed!

Avasam: I found the issue with running stubtest on windows, it's a venv bug: https://github.com/python/cpython/issues/87915
Basically you HAVE to run using `python` as the command, because:
- `py.exe` does not exists, so that always calls the global installation.
- `python3.exe` is not copied properly in `venv.create`, resulting in the above bug.

AlexWaygood:
> Unless maybe we start piping the output of all tests, checking for a KeyboardInterrupt and _then_ printing the output. I don't think that's really worth it.

Agreed, it looks good now!

tests/README.md
```diff
@@ -114,6 +126,7 @@ Run using
 ```
 (.venv3)$ python3 tests/stubtest_third_party.py
 ```
+*Note: Due to [a bug in venv](https://github.com/python/cpython/issues/87915), on Windows you **have to** use `python` as your executable. `python3` will **not** work.*
```
AlexWaygood, line 129: I'm not sure this note belongs here: it seems to apply to all our tests:

```pycon
(.venv) C:\Users\alexw\coding\typeshed>python3 tests/check_new_syntax.py
Python was not found; run without arguments to install from the Microsoft Store, or disable this shortcut from Settings > Manage App Execution Aliases.
```

Can we take out this note for now, and maybe work on improving the docs for those on Windows in (yet) another PR?

scripts/runtests.py
AlexWaygood: > I found the issue with running stubtest on windows, it's a venv bug: [python/cpython#87915](https://github.com/python/cpython/issues/87915)

Ew, what a pain.

scripts/runtests.py
```diff
@@ -0,0 +1,165 @@
...
+    print(f"\nRunning regression tests for Python {_PYTHON_VERSION}...")
+    regr_test_result = subprocess.run(
+        [sys.executable, "tests/regr_test.py", "stdlib" if folder == "stdlib" else stub, "--python-version", _PYTHON_VERSION],
+        stderr=subprocess.PIPE,
+        text=True,
+    )
+    # No test means they all ran successfully (0 out of 0). Not all 3rd-party stubs have regression tests.
+    if "No test cases found" in regr_test_result.stderr:
+        regr_test_returncode = 0
+        print(colored(f"\nNo test cases found for '{stub}'!"), "green")
```
AlexWaygood, line 117: Currently this prints:

```
No test cases found for 'redis'! green
```

```suggestion
        print(colored(f"\nNo test cases found for {stub!r}!", "green"))
```

scripts/runtests.py
```diff
@@ -0,0 +1,165 @@
...
+    mypy_result = subprocess.run([sys.executable, "tests/mypy_test.py", path, "--python-version", _PYTHON_VERSION])
+    # If mypy failed, stubtest will fail without any helpful error
+    if mypy_result.returncode == 0:
+        print("\nRunning stubtest...")
+        if folder == "stdlib":
+            stubtest_result = subprocess.run([sys.executable, "tests/stubtest_stdlib.py", stub])
+        else:
+            stubtest_result = subprocess.run([sys.executable, "tests/stubtest_third_party.py", stub])
+    else:
+        print("\nSkipping stubtest since mypy failed.")
```
AlexWaygood, line 100:
```suggestion
        print(colored("\nSkipping stubtest since mypy failed.", "yellow"))
```

scripts/runtests.py
```diff
@@ -0,0 +1,165 @@
...
+
+    if sys.platform == "win32":
+        print("\nSkipping pytype on Windows. You can run the test with WSL.")
```
AlexWaygood, line 103:
```suggestion
        print(colored("\nSkipping pytype on Windows. You can run the test with WSL.", "yellow"))
```

scripts/runtests.py
```diff
@@ -0,0 +1,165 @@
...
+    check_new_syntax_result = subprocess.run([sys.executable, "tests/check_new_syntax.py"])
+
+    print(f"\nRunning Pyright on Python {_PYTHON_VERSION}...")
+    pyright_result = subprocess.run(
+        [sys.executable, "tests/pyright_test.py", path, "--pythonversion", _PYTHON_VERSION] + _get_strict_params(path),
+        stderr=subprocess.PIPE,
+        text=True,
+    )
+    if re.match(r"error (runn|find)ing npx", pyright_result.stderr):
+        print("\nSkipping Pyright tests: npx is not installed or can't be run!")
```
AlexWaygood, line 82:
```suggestion
        print(colored("\nSkipping Pyright tests: npx is not installed or can't be run!", "yellow"))
```

AlexWaygood (commented):
Some very minor suggestions about some of the terminal output

scripts/runtests.py
AlexWaygood: Something like this patch might be a good way of getting round security concerns:

```diff
diff --git a/scripts/runtests.py b/scripts/runtests.py
index 0ee847b9..96b7f73c 100644
--- a/scripts/runtests.py
+++ b/scripts/runtests.py
@@ -91,11 +91,23 @@ def main() -> None:
     mypy_result = subprocess.run([sys.executable, "tests/mypy_test.py", path, "--python-version", _PYTHON_VERSION])
     # If mypy failed, stubtest will fail without any helpful error
     if mypy_result.returncode == 0:
-        print("\nRunning stubtest...")
         if folder == "stdlib":
+            print("\nRunning stubtest...")
             stubtest_result = subprocess.run([sys.executable, "tests/stubtest_stdlib.py", stub])
         else:
-            stubtest_result = subprocess.run([sys.executable, "tests/stubtest_third_party.py", stub])
+            run_stubtest_query = (
+                f"Run stubtest for {stub!r} (Y/N)?\n\n"
+                "NOTE: Running third-party stubtest involves downloading and executing arbitrary code from PyPI.\n"
+                f"Only run stubtest if you trust the {stub!r} package.\n"
+            )
+            run_stubtest_answer = input(colored(run_stubtest_query, "yellow")).lower()
+            while run_stubtest_answer not in {"yes", "no", "y", "n"}:
+                run_stubtest_answer = input(colored("Invalid response; please try again.\n", "red")).lower()
+            if run_stubtest_answer in {"yes", "y"}:
+                print("\nRunning stubtest...")
+                stubtest_result = subprocess.run([sys.executable, "tests/stubtest_third_party.py", stub])
+            else:
+                print(f"Skipping stubtest for {stub!r}...")
```

But maybe having user input halfway through the script would be annoying for your use case in VSCode?

tests/README.md
Avasam: Sure. Btw just FYI, you don't have to install it from the store (which I've had very bad experience with as well). You can disable that store link by going under settings > App execution aliases, and disabling it
image

AlexWaygood: I've got used to typing "python" (which is one fewer character anyway) whenever instructions on how to run a script say "python3", so it doesn't matter too much to me :)

scripts/runtests.py
Avasam: In VSCode I'm just loading up a powershell command prompt, not any different than anyone else running the script through their favorite terminal really.

It's more that having the prompt everytime might get annoying and tedious. Could we have a way to auto-accept or auto-deny? Like a command-line argument, or an Env variable? Where you'd have to have read the documentation to even know it's there. Assuring the user still at least (probably) read and accepted the risk.
Should the base `stubtest_third_party` script contain this warning, or is it really not as much of a concern?

AlexWaygood: > Could we have a way to auto-accept or auto-deny? Like a command-line argument, or an Env variable?

That would also be fine!

> Should the base stubtest_third_party script contain this warning, or is it really not as much of a concern?

We should add a note to `tests/README.md` about this, so that contributors know about it.

tests/README.md
```diff
@@ -19,6 +19,17 @@ in the `tests` and `scripts` directories.
...
+## Run all tests for a specific stub
+
+Run using:
+```
+(.venv3)$ python3 scripts/runtests.py <stdlib-or-stubs>/<stub-to-test>
+```
+
+This script will run all tests below for a specific typeshed directory. If a test supports multiple python versions, the oldest supported by typeshed will be selected. A summary of the results will be printed to the terminal.
+
+You must provide a single argument which is a path to the stubs to test, like so: `stdlib/os` or `stubs/requests`.
```
AlexWaygood, line 31: (Wrap to ~80 characters per line where possible).

```suggestion
This script will run all tests below for a specific typeshed directory. If a
test supports multiple python versions, the oldest supported by typeshed will
be selected. A summary of the results will be printed to the terminal.

You must provide a single argument which is a path to the stubs to test, like
so: `stdlib/os` or `stubs/requests`.
```

AlexWaygood (approved):
Awesome, thank you! Just one last comment, but other than that -- I've run out of things to say!

I'll leave this open for a day or two in case any other maintainers want to take a look, but this looks great to me :)

scripts/runtests.py
AlexWaygood: I've filed #8822 to add a warning about the security aspects of running `stubtest_third_party.py` locally, and also improve the docs on a few other things regarding running stubtest locally.

JelleZijlstra (approved):

PR #8801, opened 2022-09-28
Add __init__ definition for AttributeError

klaseskilson:
Since Python 3.10, [AttributeError supports keyword arguments for name and obj](https://docs.python.org/3/library/exceptions.html?highlight=attributeerror#AttributeError).

Because of this, doing something like this will cause issues with Mypy:

```py
def do_something(something):
    if not something['field']:
        # Mypy gives you `error: Unexpected keyword argument "obj" for "AttributeError" [call-arg]`
        raise AttributeError(obj=something, name='field')
    ...
```

Seems like this was only due to a missing `__init__` definition.

----- 

First contribution here. Sorry if I missed a step or made some mistake.


stdlib/builtins.pyi
```diff
@@ -1825,6 +1825,7 @@ class AssertionError(_StandardError): ...
 
 class AttributeError(_StandardError):
     if sys.version_info >= (3, 10):
+        def __init__(self, *args: object, name: str | None = ..., obj: object | None = ...) -> None: ...
```
AlexWaygood, line 1828: `None` is redundant in the union `object | None`:

```suggestion
        def __init__(self, *args: object, name: str | None = ..., obj: object = ...) -> None: ...
```

AlexWaygood (commented):
Looks good! Just one small thing to fix :)

AlexWaygood (approved):
Thank you! And welcome to typeshed!

PR #8804, opened 2022-09-28
`check_consistent.py`: ignore `.gitignore`d files

AlexWaygood:
I often have `.mypy_cache` folders hanging around in odd places in my local typeshed clone. It's annoying that `check_consistent.py` starts failing locally whenever I have one of these folders in the `stubs/` or `stdlib/` directories. Having these folders around locally is harmless; since they're globally `gitignore`d, they'll never be committed as part of a PR.

This PR uses the `pathspec` third-party library to skip checking all `gitignore`d files in the `check_consistent.py` functions `assert_consistent_filetypes()` and `check_stubs()`. It also reuses this logic to get rid of some existing hacks in `mypy_test.py`.

`pathspec` is the library `black` uses to identify which files it should skip when it does its autoformatting. This means that we all already have it installed in our virtual environments already.

### DIFF
--- a/requirements-tests.txt
+++ b/requirements-tests.txt
@@ -7,6 +7,7 @@ flake8-pyi==22.8.2      # must match .pre-commit-config.yaml
 isort==5.10.1           # must match .pre-commit-config.yaml
 mypy==0.981
 packaging==21.3
+pathspec>=0.10.1
 pycln==2.1.1            # must match .pre-commit-config.yaml
 pyyaml==6.0
 pytype==2022.08.30; platform_system != "Windows"
--- a/requirements-tests.txt
+++ b/requirements-tests.txt
@@ -7,7 +7,7 @@ flake8-pyi==22.8.2      # must match .pre-commit-config.yaml
 isort==5.10.1           # must match .pre-commit-config.yaml
 mypy==0.981
 packaging==21.3
-pathspec>=0.10.1
+pathspec
 pycln==2.1.1            # must match .pre-commit-config.yaml
 pyyaml==6.0
 pytype==2022.08.30; platform_system != "Windows"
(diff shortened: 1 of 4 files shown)


requirements-tests.txt
```diff
@@ -7,6 +7,7 @@ flake8-pyi==22.8.2      # must match .pre-commit-config.yaml
 isort==5.10.1           # must match .pre-commit-config.yaml
 mypy==0.981
 packaging==21.3
+pathspec>=0.10.1
```
AlexWaygood, line 10:
```suggestion
pathspec
```

tests/check_consistent.py
```diff
@@ -5,11 +5,13 @@
 # verify these constraints.
 from __future__ import annotations
 
+import functools
 import os
 import re
 import sys
 from pathlib import Path
 
+import pathspec
```
AlexWaygood, line 14:
```suggestion
import pathspec  # type: ignore[import]
```

tests/check_consistent.py
```diff
@@ -22,12 +24,28 @@
...
+def get_gitignore_spec() -> pathspec.PathSpec:
+    with open(".gitignore") as f:
+        return pathspec.PathSpec.from_lines("gitwildmatch", f.readlines())
+
+
+def _spec_matches_path(spec: pathspec.PathSpec, path: Path) -> bool:
+    normalized_path = path.as_posix()
+    if path.is_dir():
+        normalized_path += "/"
+    return spec.match_file(normalized_path)
```
AlexWaygood, line 37:
```suggestion
    return spec.match_file(normalized_path)  # type: ignore[no-any-return]
```

srittau (approved):

PR #8807, opened 2022-09-29
Fix installing dependencies for 3rd party stubtest

### LINKED ISSUE #8805: Stubtest failed on Thu Sep 29 2022

Akuli:
Fixes #8805

I believe it's fine to just install all dependencies, because stubtest uses venvs that don't "inherit" the dependencies installed here. If this isn't correct, we can grep the dependencies we actually need (I actually implemented and reverted this already).

### DIFF
--- a/.github/workflows/daily.yml
+++ b/.github/workflows/daily.yml
@@ -50,7 +50,7 @@ jobs:
         with:
           python-version: "3.9"
       - name: Install dependencies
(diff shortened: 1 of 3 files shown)


srittau (approved):

Akuli:
CI passed with a temporary commit that changed a third-party stub.

AlexWaygood (approved):
Thanks!

The other fix I was wondering about was to do a lazy import of `pathspec` in `tests/utils.py`. But that would be really icky with type hints, so I prefer this solution.

PR #8809, opened 2022-09-29
Allow `create_baseline_stubs.py` to add packages to existing stub

Avasam:
I was looking into pywin32 stubs, which has multiple packages. And I couldn't specify multiple packages to `create_baseline_stubs.py`, or run it multiple times with different packages.

This adds support for that, but still ensures we don't overwrite existing stubs.

An example of existing stub that has multiple packages is PyInstaller.

I could also make the `pyrightconfig.stricter.json` "exclude" more granular by specifying the package as well. I feel this would be especially important when adding a package to an existing stub, and help with coverage by not having too wide of an exclude.

### DIFF
--- a/scripts/create_baseline_stubs.py
+++ b/scripts/create_baseline_stubs.py
@@ -62,15 +62,16 @@ def run_isort(stub_dir: str) -> None:
 
 def create_metadata(stub_dir: str, version: str) -> None:
     """Create a METADATA.toml file."""
-    m = re.match(r"[0-9]+.[0-9]+", version)
-    if m is None:
+    match = re.match(r"[0-9]+.[0-9]+", version)
+    if match is None:
         sys.exit(f"Error: Cannot parse version number: {version}")
-    fnam = os.path.join(stub_dir, "METADATA.toml")
-    version = m.group(0)
-    assert not os.path.exists(fnam)
-    print(f"Writing {fnam}")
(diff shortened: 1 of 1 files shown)


scripts/create_baseline_stubs.py
```diff
@@ -62,15 +62,16 @@ def run_isort(stub_dir: str) -> None:
...
-    fnam = os.path.join(stub_dir, "METADATA.toml")
-    version = m.group(0)
-    assert not os.path.exists(fnam)
-    print(f"Writing {fnam}")
-    with open(fnam, "w") as f:
-        f.write(
+    filename = os.path.join(stub_dir, "METADATA.toml")
+    version = match.group(0)
+    if os.path.exists(filename):
+        return
```
AlexWaygood, line 71: Just to check: this is the only "substantive" change in this function, correct? The others are all readability improvements?

(In general, I don't mind readability improvements, but it's easier to review if they're in separate PRs -- it's a bit difficult to see what the substantive changes are here as opposed to the readability improvements :)

Avasam: Indeed. The only "substantive" change is to silently pass instead of erroring out if METADATA.tom already exists.
The only other changes in this function are non-shortened variable names for readability.

AlexWaygood (approved):

PR #8816, opened 2022-09-30
Remove empty `__init__` methods

sobolevn:
I've removed all `def __init__(self)` definitions if they are:
- Defined in a class that inherits directly from `object`, like `TaskGroup`
- Or defined in a class that inherits from other classes with no custom `__init__` set, like `PidfdChildWatcher` 
- Is not used as a part of `@overload`

They are not needed, because `object` already provides this method.
Keeping the stubs minimal is better for future maintenance.

There are some complex cases that I've left for now like `stdlib/xml/dom/minidom.pyi`, because it is not fully annotated yet.

Refs #8594


sobolevn:
Stubtest fails with:
```
note: unused allowlist entry _csv.Dialect.__init__
note: unused allowlist entry queue.SimpleQueue.__init__
note: unused allowlist entry select.kqueue.__init__
```

Looking into it.

sobolevn:
Ok, I removed from this PR all three cases that are listed in allowlists.

AlexWaygood (approved):

PR #8817, opened 2022-09-30
Stubsabot: Add analysis of the diff to the PR body

AlexWaygood:
This PR adds an analysis of the diff between two releases to stubsabot's PR messages, where possible. The analysis includes the following pieces of information:

- The total number of lines of Python code added between the two releases
- The total number of lines of Python code deleted between the two releases
- The total number of public Python files added between the two releases. (If this number is <5, the filenames are listed.)
- The total number of Python files included in typeshed's stubs that have been renamed or modified between the two releases. (If this number is <=5, the filenames are listed.)
- The total number of Python files included in typeshed's stubs that have been deleted between the two releases. (If this number is <=5, the filenames are listed.)

If the project isn't hosted on GitHub, no analysis is added. If the project has a directory structure that doesn't match typeshed's (e.g. the files are in an `src/` folder at runtime), only the first two pieces of information are added to the PR body.


scripts/stubsabot.py
```diff
@@ -281,7 +289,118 @@ async def get_diff_url(
...
+    def total_lines_deleted(self) -> int:
+        return sum(file["deletions"] for file in self.py_files)
+
+    @staticmethod
+    def _describe_files(*, verb: str, filenames: Sequence[str]) -> str:
+        num_files = len(filenames)
+        if num_files > 1:
+            description = f"have been {verb}"
+            # Don't list the filenames if there are *loads* of files
+            if num_files <= 5:
```
JelleZijlstra, line 349: Maybe make this a global constant? Also we could probably make it a bit higher than 5 before it becomes overwhelming.

JelleZijlstra (approved):
Looks good, just a nit

scripts/stubsabot.py
AlexWaygood: I made it a `ClassVar` constant, and bumped it up to 7.

PR #8822, opened 2022-10-02
Improve docs on running stubtest locally

AlexWaygood:


tests/README.md
```diff
@@ -88,28 +88,27 @@ Run using
 This test compares the stdlib stubs against the objects at runtime. Because of
 this, the output depends on which version of Python and on what kind of system
 it is run.
-Thus the easiest way to run this test is via Github Actions on your fork;
-if you run it locally, it'll likely complain about system-specific
-differences (in e.g, `socket`) that the type system cannot capture.
+As such, if you run this test locally, it *may* complain about system-specific
```
hauntsaninja, line 91: The original text is more prescriptive about how you should run stubtest, which I think is valuable (I don't think it's a particularly worthwhile or even feasible endeavour to make test_stubtest pass cleanly on all possible configurations of python and system). Last time I checked it was "likely" that you'd get system-specific complaints. What is the goal of this change?

tests/README.md
```diff
@@ -88,28 +88,27 @@ Run using
...
+that exactly matches typeshed's CI.
+
 If you need a specific version of Python to repro a CI failure,
-[pyenv](https://github.com/pyenv/pyenv) can help.
+[pyenv](https://github.com/pyenv/pyenv) can also help.
 
-Due to its dynamic nature, you may run into false positives. In this case, you
-can add to the allowlists for each affected Python version in
-`tests/stubtest_allowlists`. Please file issues for stubtest false positives
+Stubtest is an imperfect tool, and sometimes produces false-positive errors.
```
hauntsaninja, line 101: This implies that false positives are generally due to stubtest imperfections. With the original text I was going for "things are dynamic so we can't get always get things right, it's fine to allowlist and move on" as opposed to "file bugs on stubtest if you get some false positive" (or "give up and be confused"). Again, not quite sure what the change in wording is meant to help with.

hauntsaninja (commented):
Removing the example and the other new text is fine, but not clear to me what we're attempting to clarify in the changed text.

tests/README.md
AlexWaygood: Well, I've run `stubtest_stdlib.py` many times locally, and never once run into an issue where stubtest complained about something locally but didn't in CI. So whether or not it's "likely" that you'll get system-specific complaints would appear itself to be system-specific :)

AlexWaygood: > the type system genuinely "cannot capture" most extant system-specific differences.

I don't think there are _that_ many allowlist entries left that are due to platform-specific differences. And I think there's a solid chance that we could remove a lot of the remaining ones, if somebody has the inclination to put in the necessary work.

tests/README.md
AlexWaygood: This was meant to just be some incidental wordsmithing, but you raise good points; the changes here aren't very well motivated. I'll revert the changes to this section for now.

tests/README.md
hauntsaninja: > whether or not it's "likely"

Oh okay, yeah that makes sense. For reference, when I run `stubtest_stdlib.py` I get: `Found 139 errors (checked 485 modules)`.

> I don't think there are that many allowlist entries left 

The presence of allowlist entries is fine — by definition those don't cause `stubtest_stdlib.py` to fail. The issue is all the random ways in which local system and local Python can differ from CI, causing `stubtest_stdlib.py` to fail when run locally.

These can be worked around by adding regexes to stubtest allowlists (which is what we ended up doing with `socket`) and using the allow-unused-empty-regex trick, but I don't think this work is particularly worthwhile and can result in false negatives. For a dumb example, if someone has a Python build that doesn't include tkinter, there just isn't a great way to avoid stubtest tkinter errors

More generally, there's a norm that if you checkout a project and run its tests, and those tests fail, something unusual is happening. This just isn't the case with stubtest. Fwiw, the wording here was introduced because this confused some contributors, e.g. https://github.com/python/typeshed/issues/3782#issuecomment-592438089

AlexWaygood: > which is what we ended up doing with `socket`

Only on linux these days ;)

Cool, this all makes sense, I'll push a commit to improve this bit

AlexWaygood: > For a dumb example, if someone has a Python build that doesn't include tkinter, there just isn't a great way to avoid stubtest tkinter errors

Yeah, this is definitely true. I guess what I'm objecting to here a little bit is "the _type_ system cannot capture" this stuff. For a lot of this stuff, it's not really the _type_ system (we could probably describe differences between various linux systems if we were prepared to do stuff like `if sys.platform == "freebsd"` or whatever). For that kind of thing, it's more "we don't want to get into all of that because it's a fool's errand to try to capture platform-specific minutiae for little-used and unusual platforms".

JelleZijlstra (approved):

tests/README.md
```diff
@@ -88,28 +88,27 @@ Run using
...
 it is run.
-Thus the easiest way to run this test is via Github Actions on your fork;
-if you run it locally, it'll likely complain about system-specific
-differences (in e.g, `socket`) that the type system cannot capture.
+As such, if you run this test locally, it may complain about system-specific
+differences (in e.g, `socket`) that the type system and/or our stubtest settings
+in CI cannot capture. If you run into this issue, you can consider
+[running the test via Github Actions](https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow#running-a-workflow)
+on your typeshed fork. This ensures that the test is run in an environment
+that exactly matches typeshed's CI.
```
hauntsaninja, line 96:
```suggestion
As such, if you run this test locally, it may complain about system-specific
differences (in e.g, `socket`) that the type system cannot capture or our stubtest settings
in CI do not account for. If you run into this issue, consider opening a draft PR and letting CI
test it automatically (or
[running the test via Github Actions](https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow#running-a-workflow)
on your typeshed fork).
```

hauntsaninja (approved):
Thanks for working on this! Made a suggestion with a wording nit / explicitly suggesting that opening a PR is really fine, but feel free to ignore and merge.

tests/README.md
AlexWaygood: Nice

AlexWaygood:
Thanks! :-)

PR #8823, opened 2022-10-02
Add pyscreeze stubs

Avasam:
A lot of it comes from https://github.com/python/typeshed/blob/master/stubs/PyAutoGUI/pyautogui/__init__.pyi

Predecessor to #

### DIFF
--- a/stubs/PyScreeze/pyscreeze/__init__.pyi
+++ b/stubs/PyScreeze/pyscreeze/__init__.pyi
@@ -1,0 +1,196 @@
+import sys
+from _typeshed import Incomplete, StrOrBytesPath
+from collections.abc import Callable, Generator
+from typing import NamedTuple, SupportsFloat, TypeVar, overload
+from typing_extensions import Literal, ParamSpec, SupportsIndex, TypeAlias
+
+from PIL import Image
+
+_Unused: TypeAlias = object
+_P = ParamSpec("_P")
+_R = TypeVar("_R")
+# TODO: cv2.Mat is not available as a type yet: https://github.com/microsoft/python-type-stubs/issues/211
+# cv2.Mat is just an alias for a numpy NDArray, but can't import that either.
+_Mat: TypeAlias = Incomplete
+
+useOpenCV: bool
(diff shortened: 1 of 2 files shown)


stubs/PyScreeze/pyscreeze/__init__.pyi
```diff
@@ -0,0 +1,196 @@
...
+    image: str | Image.Image,
+    title: str,
+    *,
+    grayscale: bool | None = ...,
+    limit: _Unused = ...,
+    step: int = ...,
+    confidence: None = ...,
+) -> Box | None: ...
+def showRegionOnScreen(region: tuple[int, int, int, int], outlineColor: str = ..., filename: str = ...) -> None: ...
+def center(coords: Box) -> Point: ...
```
JelleZijlstra, line 166:
```suggestion
def center(coords: tuple[int, int, int, int]) -> Point: ...
```
The docstring just says that it needs a tuple.

stubs/PyScreeze/pyscreeze/__init__.pyi
```diff
@@ -0,0 +1,196 @@
...
+def pixelMatchesColor(
+    x: int, y: int, expectedRGBColor: tuple[int, int, int] | tuple[int, int, int, int], tolerance: int = ...
+) -> bool: ...
+def pixel(x: int, y: int) -> tuple[int, int, int]: ...
+def screenshot(imageFilename: StrOrBytesPath | None = ..., region: tuple[int, int, int, int] | None = ...) -> Image.Image: ...
+
+grab = screenshot
+# _locateAll_opencv
+@overload
```
JelleZijlstra, line 175: This is a strange use of overloads, since it looks like either one variant or the other is used depending on whether cv2 is present. Overloads are normally used when there are complicated relationships between different parameters and/or the return type. On the other hand, I can't think of any concrete ill effect of using overloads here, so I'm fine with keeping them.

Avasam: Yeah, since we can't represent that condition in typing, overloads are the next best thing so that both possible signatures are accurate.

PR #8825, opened 2022-10-02
Add `pywin32` type stubs from microsoft/python-type-stubs and mhammond/pywin32 

Avasam:
Helps with https://github.com/microsoft/python-type-stubs/issues/211

This PR does not intend to complete the types for pywin32. Due to the size of this stub, it will be done in multiple steps:
1. (this) Make the stubs at least as good as https://github.com/microsoft/python-type-stubs and https://github.com/mhammond/pywin32 combined.
    - Add everything from both stubs
    - Upon duplicates, use the most complete version of any function, method, class, type, etc.
    - Remove all explicit `Any` types (or use `Incomplete` when that was not possible)
    - Incomplete methods from https://github.com/microsoft/python-type-stubs have been updated with `(*args, **kwargs): ...  # incomplete`
    - Incomplete modules detected by stubtest (on Windows) have `def __getattr__(self, __name: str) -> Any: ...  # incomplete` added to them
    - Complete any type needed for strict pyright checking in https://github.com/Avasam/Auto-Split
    - Fixed the basetype for error classes
    - Fix anything found by typeshed's tooling
2. (future PR, or update to this PR after initial review) Complete existing modules and add all missing (non-private) modules with the help of stubgen
3. Let the community work on completing stubs to eventually turn on pyright strict typing.


Avasam:
About that primer error:
`pythoncom.py` is a module that lives directly in `sites-package`, I tried with both `pythoncom.pyi` and `pythoncom/__init__.pyi` in typeshed without success, not sure how to fix that.

stubs/pywin32/_win32typing.pyi
```diff
@@ -0,0 +1,6661 @@
+# Not available at runtime. Contains type definitions that are otherwise not exposed
+from _typeshed import Incomplete
+from typing_extensions import final
+
+class PyDSCAPSType:
+    def __new__(cls): ...
```
JelleZijlstra, line 6: What's the point of these `__new__` methods in the stubs? Wouldn't the stub behave the same without them?

AlexWaygood:
> About that primer error:
> `pythoncom.py` is a module that lives directly in `sites-package`, I tried with both `pythoncom.pyi` and `pythoncom/__init__.pyi` in typeshed without success, not sure how to fix that.

The primer output is good! It shows that mypy currently emits an error when checking `comtypes` due to the lack of types for pywin32. But this PR would (if they started using these stubs) fix that error!

Avasam:
Oh I get it now, it's saying those errors would not be in the output anymore! I'll revert back to a single module since that's more accurate.

stubs/pywin32/win32comext/axdebug/axdebug.pyi
```diff
@@ -0,0 +1,66 @@
+APPBREAKFLAG_DEBUGGER_BLOCK = ...
```
JelleZijlstra, line 1: Are these all ints? Can we use `: int` for them?

JelleZijlstra (commented):
Thanks, this is great! I have a few comments but I can't review this in detail, both because of its size and because I don't have access to Windows :)

I say we merge it (after giving other maintainers a chance to have their say) and handle any further improvements in separate PRs.

stubs/pywin32/_win32typing.pyi
Avasam: Pretty sure there's not point and they're just leftovers from copying the original stubs

PR #8827, opened 2022-10-03
[redis] Use `object` as return type when result is ignored

sobolevn:
Source: https://github.com/redis/redis-py/blob/b0883b791f95a595fae70bcedf3ad0f73c00e258/redis/asyncio/retry.py#L62


stubs/redis/redis/asyncio/retry.pyi
```diff
@@ -9,4 +9,4 @@ _T = TypeVar("_T")
 class Retry:
     def __init__(self, backoff: AbstractBackoff, retries: int, supported_errors: tuple[type[RedisError], ...] = ...) -> None: ...
     def update_supported_errors(self, specified_errors: Iterable[type[RedisError]]) -> None: ...
-    async def call_with_retry(self, do: Callable[[], Awaitable[_T]], fail: Callable[[RedisError], Any]) -> _T: ...
+    async def call_with_retry(self, do: Callable[[], Awaitable[_T]], fail: Callable[[RedisError], object]) -> _T: ...
```
JelleZijlstra, line 12:
```suggestion
    async def call_with_retry(self, do: Callable[[], Awaitable[_T]], fail: Callable[[RedisError], Awaitable[object]]) -> _T: ...
```
It awaits the return value of the callback.

sobolevn:
Thanks for catching this :)

PR #8828, opened 2022-10-03
`invoke`: Add return type for `Context.run` and `Context.sudo`

heni123321:


stubs/invoke/invoke/context.pyi
```diff
@@ -1,15 +1,16 @@
...
+from .runners import Result
 
 class Context(DataProxy):
     def __init__(self, config: Config | None = ...) -> None: ...
     @property
     def config(self) -> Config: ...
     @config.setter
     def config(self, value: Config) -> None: ...
-    def run(self, command: str, **kwargs): ...
-    def sudo(self, command: str, *, password: str = ..., user: str = ..., **kwargs): ...
```
sobolevn, line 12: Looks like these two kw-only params actually exist here: https://github.com/pyinvoke/invoke/blob/main/invoke/context.py#L179-L180

No need to remove them.

stubs/invoke/invoke/context.pyi
```diff
@@ -1,15 +1,16 @@
...
+    def run(self, command: str, **kwargs) -> Result: ...
+    def sudo(self, command: str, **kwargs) -> Result: ...
```
sobolevn, line 13: Both `run` and `sudo` call `Runner.run`:
- https://github.com/pyinvoke/invoke/blob/ce85bdf11ae9a88e18e5d669310ff0bfd4338725/invoke/context.py#L219
- https://github.com/pyinvoke/invoke/blob/ce85bdf11ae9a88e18e5d669310ff0bfd4338725/invoke/context.py#L102

`Runner.run` returns different types in different cases: https://github.com/python/typeshed/blob/6950c3237065e6e2a9b64810765fec716252d52a/stubs/invoke/invoke/runners.pyi#L18-L120

We decide what to return based on `disown: bool` and `asynchronous` args. How can we be sure that it is always `Result` here?

heni123321: In __init__, I saw that run and sudo returns Result. Promise is also a Result. What makes this different?

sobolevn: It also returns `None` and `Any`

heni123321: Yes but here it's only returning result https://github.com/python/typeshed/blob/6950c3237065e6e2a9b64810765fec716252d52a/stubs/invoke/invoke/__init__.pyi#L36-L37

AlexWaygood:
Closing due to lack of response by OP. Feel free to open a new PR if you have the time and are able to address @JelleZijlstra's comments, @heni123321! :)

PR #8834, opened 2022-10-03
Add auto-generated peewee stubs

Akuli:
#4776 has been open and inactive for a very long time now. I'm planning to copy parts of it into smaller, easily reviewable PRs later. Other contributors should feel free to do that too in case I get distracted :)

This PR contains the output of `scripts/create_baseline_stubs.py` and whatever is needed to make the CI pass.


AlexWaygood (approved):
There's some obvious things to fix here, but they can all be addressed in follow-up PRs :)

srittau (approved):

PR #8838, opened 2022-10-04
Update pyOpenSSL stubs to 22.1.*

srittau:
Cf.  #8810


srittau:
I *assume* that the `errors` argument should be `list[str]`, because the only initialization is here: https://github.com/pyca/pyopenssl/blob/7abe40cbff2b30f6b5d59b5a1106b7a6a237df86/src/OpenSSL/crypto.py#L1876-L1899

But I'm not 100% sure what those methods return. (Could be `bytes`.)

srittau:
Actually it seems that `errors` is initialized with [int, int, str]. Changed `Incomplete` to `Any`.

stubs/pyOpenSSL/OpenSSL/crypto.pyi
```diff
@@ -143,8 +143,9 @@ class X509StoreContext:
     def verify_certificate(self) -> None: ...
 
 class X509StoreContextError(Exception):
+    errors: list[Any]
     certificate: X509
-    def __init__(self, message: str | bytes, certificate: X509) -> None: ...
+    def __init__(self, message: str | bytes, errors: list[Any], certificate: X509) -> None: ...
```
JelleZijlstra, line 148: OpenSSL's inline annotations have `message` as just `str`. Not sure where we got `bytes` from. That's a separate issue from this PR though.

JelleZijlstra (approved):

stubs/pyOpenSSL/OpenSSL/crypto.pyi
srittau: I changed it while I'm here.

Issue #8841, opened 2022-10-04
Add unit tests for scripts in `tests/` and `scripts/`


AlexWaygood:
Some of our scripts in the `tests/` and `scripts/` directories are increasingly subtle. It's becoming increasingly hard to refactor and improve some of them without breaking them in unexpected ways.

I'd like to add some unit tests for some of these scripts, to help address this. Does that sound like a good idea?

PR #8800, opened 2022-09-27
`mypy_test.py`: Always add dependencies of stubs to the files to test

AlexWaygood:
Fixes #8797 (cc. @Avasam)

PR #8842, opened 2022-10-04
`mypy_test.py`: restore previous behaviour where all third-party stubs are tested by default

AlexWaygood:
Accidentally broken in #8800 (#8841 would help prevent this kind of thing!)


JelleZijlstra (approved):

AlexWaygood:
Stubtest seems stuck, but I'm pretty confident this won't break stubtest :)

PR #8845, opened 2022-10-04
Run stubsabot daily, not weekly

AlexWaygood:
I think stubsabot has been a really big success. Let's change it to a daily workflow, so that we don't get an avalanche of PRs every Friday, and so that we can see more quickly if a PR changing stubsabot has accidentally broken the script. (If that _does_ happen, it should be pretty easy to just revert the commit that accidentally broke stubsabot, and go back to a version that works pretty well.)


JelleZijlstra (approved):

hauntsaninja (approved):
Sure. I typically manually trigger a stubsabot run after merging changes.

PR #8846, opened 2022-10-04
`mypy_test.py`: Remove unused `seen` variable/parameters

AlexWaygood:
I think we used to use this variable in a previous version of this script, but it's now unused.


JelleZijlstra (approved):

AlexWaygood:
Stubtest is stuck again, but I'm pretty sure this won't break stubtest :)

PR #8854, opened 2022-10-05
stubsabot: only update existing branches if the diff relative to `master` would be different

AlexWaygood:
Currently, if there are no non-stubsabot commits on a stubsabot branch associated with an open PR, stubsabot will force-push to that branch on a daily basis to update the PR.

This PR proposes tweaking that logic so that stubsabot will only update the PR branch if it would change the diff between the PR and master.

### DIFF
--- a/scripts/stubsabot.py
+++ b/scripts/stubsabot.py
@@ -523,6 +523,22 @@ def has_non_stubsabot_commits(branch: str) -> bool:
         return False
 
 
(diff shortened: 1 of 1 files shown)


JelleZijlstra (approved):

AlexWaygood:
Confirmed that stubsabot is still updating open PRs where necessary: see #8849 for an example.

PR #8855, opened 2022-10-06
`fpdf2`: complete stubs

AlexWaygood:


srittau (approved):
Did not compare with the implementation, but looks plausible.

PR #8856, opened 2022-10-06
`ZoneInfo.clear_cache` can have `Optional` input

sobolevn:
Source Python: https://github.com/python/cpython/blob/0d68879104dfb392d31e52e25dcb0661801a0249/Lib/zoneinfo/_zoneinfo.py#L92
Source C: https://github.com/python/cpython/blob/0d68879104dfb392d31e52e25dcb0661801a0249/Modules/_zoneinfo.c#L413


sobolevn:
Re-triggering

sobolevn:
We need to re-run the CI later, looks like GitHub has some issues.

srittau (approved):

PR #8861, opened 2022-10-07
Use general type ignore for asyncio.Task

cdce8p:
Discovered while working on https://github.com/python/mypy/pull/13831

Mypy does also complain about the covariance introduced in #8781.
To reproduce a simple test case from mypy
```py
# test.py

from asyncio import Future, wait
from typing import List

async def foo() -> None:
    f = []  # type: List[Future[None]]
    await wait(f)
```

```
mypy --no-silence-site-packages test.py
```

```
mypy/typeshed/stdlib/asyncio/tasks.pyi:273: error: Variance of TypeVar "_T_co" incompatible with variance in parent type  [type-var]
```

/CC: @hauntsaninja

### DIFF
--- a/stdlib/asyncio/tasks.pyi
+++ b/stdlib/asyncio/tasks.pyi
@@ -266,11 +266,11 @@ else:
     ) -> tuple[set[Task[_T]], set[Task[_T]]]: ...
     async def wait_for(fut: _FutureLike[_T], timeout: float | None, *, loop: AbstractEventLoop | None = ...) -> _T: ...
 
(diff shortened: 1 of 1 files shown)


JelleZijlstra:
I think we just merged the mypy change that checks for this.

JelleZijlstra (approved):

AlexWaygood (approved):
Thanks! I think this is because mypy complains about this on the master branch, but not in the release of mypy we use in CI (the check was only recently added)

cdce8p:
Found it: https://github.com/python/mypy/pull/13714

PR #8868, opened 2022-10-07
Add MYPYPATH To Stubtest Documentation

HassanAbouelela:
MYPYPATH must be set to the stub directory when running subtest on third party packages, otherwise the stubs won't be found. This is already done in scripts, but not mentioned in the relevant documentation.


tests/README.md
```diff
@@ -116,7 +116,7 @@ at [mypy](https://github.com/python/mypy/issues).
 To run stubtest against third party stubs, it's easiest to use stubtest
```
hauntsaninja, line 116: Want to move this paragraph into the stubtest_third_party section where you made the other change? (I think this text was written before we tested third party packages with stubtest, which is how this text ended up in the stdlib section)

hauntsaninja (commented):
Thanks for updating this, one comment

AlexWaygood:
This PR conflicts a little with #8822. Not sure how we want to resolve that.

tests/README.md
HassanAbouelela: That sounds reasonable, I'll see about doing that pending figuring out the solution with the other PR.

HassanAbouelela:
Ah I should've done a better job looking at existing PRs. If this section is removed anyway, I'm happy to just close this PR.

I do hope some mention of it is kept, as it is still convenient to run stubtest with other options (such as concise), which isn't as easy with the script. However, toggling the ignore in metadata is something I wasn't aware of and is a good substitute.

AlexWaygood:
> I do hope some mention of it is kept, as it is still convenient to run stubtest with other options (such as concise), which isn't as easy with the script.

That's very fair! (I'm also very happy for this to be merged before mine btw).

Maybe we should meek the example (with your revisions) and move it down to the section on `stubtest_third_party.py`? I feel like it doesn't make much sense in the current section, but your rationale for keeping it somewhere makes a lot of sense.

hauntsaninja:
Let's merge #8822 first and once it's merged, move the example down like Alex suggests :-)

hauntsaninja:
Okay @HassanAbouelela , just merged the other PR!

HassanAbouelela:
Thanks, I've merged in the other changes now

tests/README.md
HassanAbouelela: Resolved in 6d2bd8e

tests/README.md
```diff
@@ -136,6 +136,14 @@ check on the command line:
 (.venv3)$ python3 tests/stubtest_third_party.py Pillow toml  # check stubs/Pillow and stubs/toml
 ```
 
+To run stubtest against third party stubs, it's easiest to use stubtest
```
AlexWaygood, line 139:
```suggestion
If you have the runtime package installed in your local virtual environment, you can also run stubtest
```

hauntsaninja (approved):
Thanks again, looks great!

PR #8870, opened 2022-10-08
[redis] Type `redis.commands.parser`

sobolevn:
Source: https://github.com/redis/redis-py/blob/master/redis/commands/parser.py

This is my 200th PR 🎉


stubs/redis/redis/commands/parser.pyi
```diff
@@ -1,8 +1,9 @@
...
+from redis.client import AbstractRedis
+from redis.typing import EncodableT
 
 class CommandsParser:
-    initialized: bool
-    commands: Any
-    def __init__(self, redis_connection) -> None: ...
-    def initialize(self, r) -> None: ...
-    def get_keys(self, redis_conn, *args): ...
+    initialized: bool  # will be removed in the future
```
JelleZijlstra, line 5: It's already gone I think, I don't see it in 4.3.4.

JelleZijlstra:
```suggestion
```

PR #8873, opened 2022-10-09
Cancel old `mypy_primer` builds 

sobolevn:
Several changes:
1. Now old `mypy_primer` builds are auto-canceled on new commits
2. `mypy_primer_comment` is only executed if `mypy_primer` succeeded
3. `pre-commit.ci` can run `mypy_primer` now, because it cancels the previous build
4. I moved `permissions` on the top level for better visibility

Docs: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#running-a-workflow-based-on-the-conclusion-of-another-workflow

Refs https://github.com/python/mypy/pull/13846
CC @hauntsaninja

### DIFF
--- a/.github/workflows/mypy_primer.yml
+++ b/.github/workflows/mypy_primer.yml
@@ -9,13 +9,17 @@ on:
     - '.github/workflows/mypy_primer.yml'
     - '.github/workflows/mypy_primer_comment.yml'
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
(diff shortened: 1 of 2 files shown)


sobolevn:
CI is canceled: https://github.com/python/typeshed/actions/runs/3213617778/jobs/5253418027 by https://github.com/python/typeshed/pull/8873/commits/97f6cdcd98ffdc66d50dc9e7cee90b8462029e64

hauntsaninja (approved):
Thanks!

PR #8875, opened 2022-10-10
stat_result: add docstrings

simon04:
For better IDE integration in VS Code.

Ref: https://docs.python.org/3.11/library/os.html#os.stat_result


JelleZijlstra:
Our current policy is to not have docstrings in stubs. I've heard before that it would be useful for IDEs if stubs included docstrings and I'd be open to add docstrings in order to support IDEs, but so far nobody has come to ask to change the policy.

AlexWaygood:
I'd also be open to allowing docstrings in stubs. But it would be a pretty major policy change that would require some discussion first (if we go for this, should we add all docstrings to stubs? Just the ones requested by users? Only if they're under a certain length?). And I think that discussion should take place in an issue, rather than a PR thread. (As you can see by the failing CI, we'd also either have to update flake8-pyi or ignore that error code in CI.)

I'm going to close this for now, but feel free to open an issue to discuss our policy on this if you feel strongly about it :)

PR #8877, opened 2022-10-10
extend unquote type hint and replace _Str with AnyStr

gruebel:
- extended the `string` parameter in `unquote()` to allow both `str` and `bytes` which was added in Python 3.9
- replace the `TypeAlis` `_Str` with `AnyStr` which feels for me the right thing to do, but if there is a specific reason to keep this Alias, then I can revert that specific change


JelleZijlstra:
The AnyStr change looks incorrect. TypeVars are not equivalent to unions; they are usually meaningful only when they are used more than once in a function definition.

srittau:
In addition to what @JelleZijlstra wrote: I think you are right about `_Str` (probably a holdover from Python 2), but we should just use `str | bytes`.

srittau:
I would guess either "historic" reasons or a mistake. Or maybe all the `Any`s in `urlencode()` should actually be `AnyStr`s. 🤷

PR #8878, opened 2022-10-10
Fix paho-mqtt typing errors

JonathanPlasse:
Hi,
I replaced `TypeVar` with `Union` because [an attribute should not have a type unbound](https://github.com/python/mypy/issues/6520).
And fixed minor type inconsistencies.


JonathanPlasse:
I finished fixing the typing errors.

JonathanPlasse:
Is it normal that the new version has not been published on PyPI?

srittau:
Yes, the stubs are updated a few times a day. It might take a few hours.

PR #8879, opened 2022-10-11
opencv-python stubs

Avasam:
***ready for review, but do not merge until #5768 . numpy is too fundamental for this stub, and merging as-is will cause lots of issues when trying to use it with pyright strict typing***
(I could make an ndarray protocol with all of its properties and methods, but I'd rather wait)

- I first generated the stubs using stubgen
  - Including manually re-running stubgen for dynamically generated modules that it wasn't picking up recursively (https://github.com/python/mypy/issues/13904)
- also generated the stubs with pyright since it catches more return types (but can't do C module)
- satisfied pyright strict typing for all files but `cv2/cv2.pyi`.
- then added everything from https://github.com/microsoft/python-type-stubs/tree/main/cv2
- used the existing docstrings to complete signatures and overloads
- Included different pending fixes/adjustments:
  - https://github.com/microsoft/python-type-stubs/pull/227
  - https://github.com/microsoft/python-type-stubs/pull/219
  - https://github.com/microsoft/python-type-stubs/pull/232
- Fixed misnamed parameters `dts` to `dst`
- Changed `list` params to `Sequence` and `tuple` params to aliases that include `Sequence`
- Removed docstrings (see #4881)
- Fixed overloads missing `self` parameter
- Validated against https://github.com/Avasam/Auto-Split

As noted in `METADATA.toml`, the stub is stuck on opencv-python v4.5 until https://github.com/opencv/opencv-python/issues/676 is fixed (`import cv2.cv2` causes an error).
`cv2/cv2.pyi` is incomplete, massive and you might just wanna skim over it. But everything else should get a proper look and is meant to be complete.

For later / Not in this PR:
- ~~Fix unspecified encoding causing `UnicodeDecodeError` with utf-8 chars on Windows in `check_new_syntax.py` (#8882) and `Flake8-pyi` (https://github.com/PyCQA/flake8-pyi/pull/296)~~
- Complete `# incomplete` signatures using docstrings from the `help()` method.
- Complete return types enough to remove all temporary aliases.


stubs/opencv-python/cv2/__init__.pyi
```diff
@@ -0,0 +1,6 @@
+from cv2 import data as data, gapi as gapi, mat_wrapper as mat_wrapper, misc as misc, utils as utils, version as version
+from cv2.cv2 import *
+
+__all__: list[str] = []
```
AlexWaygood, line 4: This is a bit weird. `__all__` is empty, but there's nonetheless a bunch of stuff that should be considered re-exported from the module? 🧐

stubs/opencv-python/cv2/cv2.pyi
```diff
@@ -0,0 +1,4623 @@
...
+
+# #5768
+# import numpy
+_Mat: TypeAlias = Incomplete  # numpy.ndarray[int, np.dtype[np.generic]]
+_MatF: TypeAlias = Incomplete  # numpy.ndarray[float, np.dtype[np.generic]]
+
+# Function argument types
+_NumericScalar: TypeAlias = float | bool | None
+_Scalar: TypeAlias = _Mat | _NumericScalar | Sequence[_NumericScalar]
+_Point: TypeAlias = Union[tuple[int, int], Sequence[int]]  # noqa: Y047
```
AlexWaygood, line 16: Can you add a comment to explain why there's all these `noqa`s?

stubs/opencv-python/cv2/cv2.pyi
```diff
@@ -0,0 +1,4623 @@
...
+_Range: TypeAlias = Union[tuple[int, int], Sequence[int]]  # noqa: Y047
+_PointFloat: TypeAlias = Union[tuple[float, float], Sequence[float]]  # noqa: Y047
+_SizeFloat: TypeAlias = Union[tuple[float, float], Sequence[float]]  # noqa: Y047
+_Rect: TypeAlias = Union[tuple[int, int, int, int], Sequence[int]]  # noqa: Y047
+_Boolean: TypeAlias = bool | int | None  # noqa: Y047
+# _UMat also covers InputArray and InputOutputArray
+_UMat: TypeAlias = UMat | _Mat | _NumericScalar
+_UMatF: TypeAlias = UMat | _MatF | _NumericScalar
+
+_TUMat = TypeVar("_TUMat", bound=_UMat)  # noqa: Y018
```
AlexWaygood, line 27: Remind me which one Y018 is (😅) and why we're `noqa`ing it?

stubs/opencv-python/cv2/cv2.pyi
```diff
@@ -0,0 +1,4623 @@
...
+WINDOW_NORMAL: int
+WINDOW_OPENGL: int
+WND_PROP_ASPECT_RATIO: int
+WND_PROP_AUTOSIZE: int
+WND_PROP_FULLSCREEN: int
+WND_PROP_OPENGL: int
+WND_PROP_TOPMOST: int
+WND_PROP_VISIBLE: int
+WND_PROP_VSYNC: int
+_INPUT_ARRAY_CUDA_GPU_MAT: int
```
AlexWaygood, line 1797: Hmm, do we need all these private variables in the stub?

stubs/opencv-python/cv2/cv2.pyi
```diff
@@ -0,0 +1,4623 @@
...
+    def __init__(self, *args, **kwargs) -> None: ...  # incomplete
+    def create(self, *args, **kwargs) -> Any: ...  # incomplete
+    def getDefaultName(self, *args, **kwargs) -> Any: ...  # incomplete
+    def getDescriptorChannels(self, *args, **kwargs) -> Any: ...  # incomplete
+    def getDescriptorSize(self, *args, **kwargs) -> Any: ...  # incomplete
+    def getDescriptorType(self, *args, **kwargs) -> Any: ...  # incomplete
+    def getDiffusivity(self, *args, **kwargs) -> Any: ...  # incomplete
+    def getNOctaveLayers(self, *args, **kwargs) -> Any: ...  # incomplete
+    def getNOctaves(self, *args, **kwargs) -> Any: ...  # incomplete
+    def getThreshold(self, *args, **kwargs) -> Any: ...  # incomplete
```
AlexWaygood, line 1873: Should these return `Incomplete` if they're incomplete? (Same question for a lot of the other functions with the `# incomplete` comment.)

stubs/opencv-python/cv2/gapi/__init__.pyi
```diff
@@ -0,0 +1,107 @@
...
+def register(mname: str) -> Callable[[_F], _F]: ...
+def networks(*args: gapi_ie_PyParams) -> gapi_GNetPackage: ...
+def compile_args(*args: gapi_GKernelPackage | gapi_GNetPackage | queue_capacity) -> list[GCompileArg]: ...
+def GIn(*args: _A) -> list[_A]: ...
+def GOut(*args: _A) -> list[_A]: ...
+def gin(*args: _A) -> list[_A]: ...
+def descr_of(*args: _A) -> list[_A]: ...
+
+class GOpaque:
+    def __new__(cls, argtype: int) -> GOpaqueT: ...  # type: ignore[misc]
```
AlexWaygood, line 26: What are all the `type: ignore`s here for?

AlexWaygood (commented):
Thanks! I can see a lot of work's gone into this.

Just a few questions from skimming through:

stubs/opencv-python/cv2/__init__.pyi
Avasam: `__all__` is also empty in the source code and at runtime.
If `__all__` is empty, should it be omitted?

Btw, cv2 adds a bunch of code and modules dynamically, not sure if that's why they have an empty `__all__` there..

stubs/opencv-python/cv2/cv2.pyi
Avasam: I'm not sure whether they're meant to be private or if it's just because the c equivalent is named the same.
But they would be easy to add back if a use-case is found.

stubs/opencv-python/cv2/cv2.pyi
Avasam: I wrote it similar to how we'd write:
`def __getattr__(name: str) -> Any: ...  # incomplete` (taken from CONTRIBUTING.md)

Just that in this case we know that name of the attribute, and that it's a function. But have no idea about the arguments.

To me, it makes it more clear that that args are not actually `*args, **kwargs`

`def getThreshold(self, *args, **kwargs): ... # incomplete`
or
`def getThreshold(self, *args, **kwargs) -> Incomplete: ...`
could work as well (although we tend to not use `Incomplete` by itself as a return type)

stubs/opencv-python/cv2/gapi/__init__.pyi
Avasam: mypy is not happy that `__new__` doesn't return a subtype of `GOpaque`.
I'm adding the following comment from source to make it clearer why it's like this:
```py
# NB: Inheritance from c++ class cause segfault.
# So just aggregate cv.GOpaqueT instead of inheritance
```

stubs/opencv-python/cv2/cv2.pyi
Avasam: Y047 flags the unused TypeAlias, and Y018 the unused TypeVar. In both cases they're reused elsewhere in the stub. And they will certainly be used in this file once enough types are filled in.

Why here instead of `cv2/__init__.pyi`? Because `cv2` is just a re-export of `cv2.cv2` for convenience.

stubs/opencv-python/cv2/cv2.pyi
JelleZijlstra: In that case, you can just omit the return types instead of writing "-> Any". (Any should generally mean "we've looked at this function but we can't express its return type in the type system".)

Avasam: Done!

stubs/opencv-python/cv2/__init__.pyi
```diff
@@ -0,0 +1,21 @@
...
+    Error as Error,
+    data as data,
+    gapi as gapi,
+    mat_wrapper as mat_wrapper,
+    misc as misc,
+    utils as utils,
+    version as version,
+)
+from cv2.cv2 import *
+from cv2.mat_wrapper import Mat as WrappedMat, _NDArray
```
JelleZijlstra, line 13: `cv2.WrappedMat` doesn't exist for me (on Mac, 3.10, with opencv 4.5 or 4.6)

stubs/opencv-python/cv2/cv2.pyi
```diff
@@ -0,0 +1,4551 @@
...
+    def __init__(self, *args, **kwargs) -> None: ...  # incomplete
+    def calculateShift(self, *args, **kwargs): ...  # incomplete
+    def computeBitmaps(self, *args, **kwargs): ...  # incomplete
+    def getCut(self, *args, **kwargs): ...  # incomplete
+    def getExcludeRange(self, *args, **kwargs): ...  # incomplete
+    def getMaxBits(self, *args, **kwargs): ...  # incomplete
+    @overload
+    def process(self, src, dst, times, response) -> None: ...
+    @overload
+    def process(src, dst) -> None: ...
```
JelleZijlstra, line 1857:
```suggestion
    def process(self, src, dst) -> None: ...
```

Verified the requirement argument count in the REPL

stubs/opencv-python/cv2/cv2.pyi
```diff
@@ -0,0 +1,4551 @@
...
+    def descriptorSize(self, *args, **kwargs): ...  # incomplete
+    def descriptorType(self, *args, **kwargs): ...  # incomplete
+    def detect(self, *args, **kwargs): ...  # incomplete
+    def detectAndCompute(self, *args, **kwargs): ...  # incomplete
+    def empty(self, *args, **kwargs): ...  # incomplete
+    def getDefaultName(self, *args, **kwargs): ...  # incomplete
+    @overload
+    def read(self, fileName) -> None: ...
+    @overload
+    def read(arg1) -> None: ...
```
JelleZijlstra, line 2133: This one might also need a self arg, but...
```
In [18]: f = cv2.Feature2D()

In [19]: f.read??
Docstring:
read(fileName) -> None
.   



read(arg1) -> None
.
Type:      builtin_function_or_method

In [20]: f.read("x")
zsh: segmentation fault  python -m IPython
```


If you pass an int it tells you what it expects

```
In [4]: f.read(1)
error                                     Traceback (most recent call last)
<ipython-input-4-5a354f952aa4> in <module>
----> 1 f.read(1)

error: OpenCV(4.5.5) :-1: error: (-5:Bad argument) in function 'read'
> Overload resolution failed:
>  - Can't convert object to 'str' for 'fileName'
>  - Expected cv::FileNode for argument 'arg1'
```

stubs/opencv-python/cv2/cv2.pyi
```diff
@@ -0,0 +1,4551 @@
...
+    def __init__(self, *args, **kwargs) -> None: ...  # incomplete
+
+class GStreamingCompiled:
+    def __init__(self, *args, **kwargs) -> None: ...  # incomplete
+    def pull(self, *args, **kwargs): ...  # incomplete
+    def running(self, *args, **kwargs): ...  # incomplete
+    @overload
+    def setSource(self, callback) -> None: ...
+    @overload
+    def setSource(self): ...
```
JelleZijlstra, line 2264:
```suggestion
    def setSource(self, callback) -> None: ...
```

`.setSource()` with no arg just fails.

stubs/opencv-python/cv2/cv2.pyi
```diff
@@ -0,0 +1,4551 @@
...
+    def channels(self, *args, **kwargs): ...  # incomplete
+    def clone(self, *args, **kwargs): ...  # incomplete
+    def col(self, *args, **kwargs): ...  # incomplete
+    def colRange(self, *args, **kwargs): ...  # incomplete
+    def convertTo(self, *args, **kwargs): ...  # incomplete
+    def copyTo(self, *args, **kwargs): ...  # incomplete
+    @overload
+    def create(self, rows, cols, type) -> None: ...
+    @overload
+    def create(size, type) -> None: ...
```
JelleZijlstra, line 2871:
```suggestion
    def create(self, size, type) -> None: ...
```

stubs/opencv-python/cv2/cv2.pyi
```diff
@@ -0,0 +1,4551 @@
...
+    def setTo(self, *args, **kwargs): ...  # incomplete
+    def size(self, *args, **kwargs): ...  # incomplete
+    def step1(self, *args, **kwargs): ...  # incomplete
+    def swap(self, mat) -> None: ...
+    def type(self, *args, **kwargs): ...  # incomplete
+    def updateContinuityFlag(self) -> None: ...
+    @overload
+    def upload(self, arr) -> None: ...
+    @overload
+    def upload(arr, stream) -> None: ...
```
JelleZijlstra, line 2894:
```suggestion
    def upload(self, arr, stream) -> None: ...
```

stubs/opencv-python/cv2/cv2.pyi
```diff
@@ -0,0 +1,4551 @@
...
+    def apply(self, index, corner, image, mask) -> _image: ...
+    def getBlockSize(self, *args, **kwargs): ...  # incomplete
+    def getMatGains(self, *args, **kwargs): ...  # incomplete
+    def getNrFeeds(self, *args, **kwargs): ...  # incomplete
+    def getNrGainsFilteringIterations(self, *args, **kwargs): ...  # incomplete
+    def getSimilarityThreshold(self, *args, **kwargs): ...  # incomplete
+    @overload
+    def setBlockSize(self, width, height) -> None: ...
+    @overload
+    def setBlockSize(size) -> None: ...
```
JelleZijlstra, line 2976:
```suggestion
    def setBlockSize(self, size) -> None: ...
```
I also found out that `size` has to be a length-two sequence. If you pass a 1-sequence it throws an exception, but with a 2-sequence it segfaults.

stubs/opencv-python/cv2/cv2.pyi
```diff
@@ -0,0 +1,4551 @@
...
+def AffineFeature_create(*args, **kwargs): ...  # incomplete
+def AgastFeatureDetector_create(threshold=..., nonmaxSuppression=..., type=...): ...
+def BFMatcher_create(normType: int = ..., crossCheck=...): ...
+@overload
+def BRISK_create(thresh=..., octaves=..., patternScale=...): ...
+@overload
+def BRISK_create(radiusList, numberList, dMax=..., dMin=..., indexChange=...): ...
+@overload
+def BRISK_create(thresh, octaves, radiusList, numberList, dMax=..., dMin=..., indexChange=...): ...
+def CamShift(probImage, window, criteria) -> tuple[tuple[Incomplete, _window]]: ...
```
JelleZijlstra, line 3731:
```suggestion
def CamShift(probImage, window, criteria) -> tuple[Incomplete, _window]: ...
```

```
In [23]: cv2.CamShift(1, [1, 2, 3, 4], [1, 2, 3])
Out[23]: (((0.5, 1.0), (0.0, 0.0), 90.0), (0, 0, 1, 2))
```

stubs/opencv-python/cv2/gapi/streaming.pyi
```diff
@@ -0,0 +1,16 @@
...
+sync_policy_drop: int
+
+queue_capacity = gapi_streaming_queue_capacity
+
+def desync(g: GMat) -> GMat: ...
+def seqNo(arg1: GMat) -> GOpaqueT: ...
+def seq_id(arg1: GMat) -> GOpaqueT: ...
+
+# "src" and "r" are both valid named first parameter, but the overload resolution always fails
+def size(_src: GMat) -> GOpaqueT: ...
```
JelleZijlstra, line 15:
```suggestion
def size(src: GMat) -> GOpaqueT: ...
```
works for me as a named parameter

```
In [42]: cv2.gapi.streaming.size(src=cv2.GMat())
Out[42]: <GOpaqueT 0x1050d56f0>
```

JelleZijlstra (commented):
I read over most of it and noticed a few things. The tool you're using to generate the stubs may have a bug where it omits the `self` argument to the second overload of a method.

stubs/opencv-python/cv2/gapi/__init__.pyi
AlexWaygood: I think it might be slightly clearer if you added a comment to the effect of "the `type: ignore`s are because mypy doesn't like `__new__` returning something that isn't a subtype of the class" as well :)

stubs/opencv-python/cv2/__init__.pyi
Avasam: Should've been a private alias. I'm surprised stubtest didn't catch that.

stubs/opencv-python/cv2/cv2.pyi
Avasam: > The tool you're using to generate the stubs may have a bug where it omits the self argument to the second overload of a method.

Indeed! I hadn't noticed that. I used mypy.stubgen. But I think it' more an issue in the docstring (which to my understanding is used to generate some signatures):
```py
>>> from cv2.cv2 import * 
>>> help(AlignMTB.process) 
Help on method_descriptor:

process(...)
    process(src, dst, times, response) -> None
    .



    process(src, dst) -> None
    .   @brief Short version of process, that doesn't take extra arguments.
    .
    .       @param src vector of input images
    .       @param dst vector of aligned images
```

stubs/opencv-python/cv2/cv2.pyi
Avasam: All `Point`, `Size`, `Range`, `Rect` (and probably more) are like that.
Since there's no way that I know of to specify an exact number of arguments for sequences, I've created aliases that are a union of sequence and tuples with the right length

AlexWaygood:
Adding the "DO-NOT-MERGE" label, as I just noticed the note at the top of the PR description :)

Avasam:
Some update: work as started again on opencv stub generation. Hopefully this here PR never gets to be merged, or a slightly different version of it for older openCV support (as I've already mentioned, some people, myself included, are stuck on 4.5 for different reasons).

Keeping this here as a draft as it is a useful comparison point to eventually get the wip official generated stubs as accurate as possible.

Avasam:
Given the recent issues with SQLAlchemy2, I no longer think it would be wise to backport cv2 types, even if they're available right now from a github install.

And thanks to opencv-python being updated before opencv 4.8, it means users will already have access to types (even if early versions) in opencv4.7 by installing from github.

PR #8882, opened 2022-10-11
Specify file encoding

Avasam:
See https://peps.python.org/pep-0597/
This also happened to me on Windows in #8879 because of mathematical unicode characters as part of the comments:
```py
(.venv) PS C:\Users\Avasam\Documents\Git\typeshed> python .\tests\check_new_syntax.py
  File "C:\Users\Avasam\Documents\Git\typeshed\tests\check_new_syntax.py", line 110, in <module>
    main()
  File "C:\Users\Avasam\Documents\Git\typeshed\tests\check_new_syntax.py", line 100, in main
    stub = f.read()
  File "C:\Program Files\Python39\lib\encodings\cp1252.py", line 23, in decode 
    return codecs.charmap_decode(input,self.errors,decoding_table)[0]
UnicodeDecodeError: 'charmap' codec can't decode byte 0x90 in position 176166: 
character maps to <undefined>
```

There are linters to catch this, namely [pylint (unspecified-encoding)](https://pylint.pycqa.org/en/latest/user_guide/messages/warning/unspecified-encoding.html), [flake8-encodings](https://pypi.org/project/flake8-encodings/) and [flake8-file-encoding](https://pypi.org/project/flake8-file-encoding/). I have tried both flake8 plugins mentioned and they gave me the same results, other than slightly different error messages:
image

### DIFF
--- a/stubs/requests/@tests/test_cases/check_post.py
+++ b/stubs/requests/@tests/test_cases/check_post.py
@@ -44,8 +44,8 @@ requests.post("http://httpbin.org/anything", data=b"foobar").json()["data"]
 requests.post("http://httpbin.org/anything", data="foobar").json()["data"]
 
 # Files
-requests.post("http://httpbin.org/anything", data=open("/tmp/foobar", "rb")).json()["data"]
-requests.post("http://httpbin.org/anything", data=open("/tmp/foobar", "r")).json()["data"]
+requests.post("http://httpbin.org/anything", data=open("/tmp/foobar", "rb", encoding="UTF-8")).json()["data"]
+requests.post("http://httpbin.org/anything", data=open("/tmp/foobar", "r", encoding="UTF-8")).json()["data"]
 
 # Mappings
 requests.post("http://httpbin.org/anything", data={b"foo": b"bar"}).json()["form"]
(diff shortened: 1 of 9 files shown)


JelleZijlstra:
Thanks! This is generally an issue on Windows only since Unix defaults to utf-8.

stubs/requests/@tests/test_cases/check_post.py
```diff
@@ -44,8 +44,8 @@ def gen() -> Iterable[bytes]:
 requests.post("http://httpbin.org/anything", data="foobar").json()["data"]
 
 # Files
-requests.post("http://httpbin.org/anything", data=open("/tmp/foobar", "rb")).json()["data"]
-requests.post("http://httpbin.org/anything", data=open("/tmp/foobar", "r")).json()["data"]
+requests.post("http://httpbin.org/anything", data=open("/tmp/foobar", "rb", encoding="UTF-8")).json()["data"]
+requests.post("http://httpbin.org/anything", data=open("/tmp/foobar", "r", encoding="UTF-8")).json()["data"]
```
AlexWaygood, line 48: The code in this file is never actually run (it's a test-case), but I guess there's no way of telling the flake8 plugin that, and I don't think this change does any actual harm :)

AlexWaygood (approved):
Thanks!

PR #8883, opened 2022-10-11
Mark `atomicwrites` as complete

sobolevn:
Btw, this package was deprecated: https://github.com/untitaker/python-atomicwrites

Let's see if stubtest is happy about it.


hauntsaninja (approved):
Looks great, thanks!

PR #8884, opened 2022-10-11
Mark `chevron` as completed

sobolevn:
Source: https://github.com/noahmorrison/chevron


stubs/chevron/chevron/main.pyi
```diff
@@ -2,6 +2,8 @@ from _typeshed import StrOrBytesPath
 from typing import Any
 from typing_extensions import TypeAlias
 
+version: str
+
```
AlexWaygood, line 6: I'm going to be picky here: I think it would be better if this was allowlisted. The fact that this constant is imported from `.metadata` in this file is an implementation detail; if users are accessing the constant, they should import it from `chevron.metadata` instead of importing it from `chevron.main`.

stubs/chevron/chevron/renderer.pyi
```diff
@@ -1,9 +1,13 @@
 from _typeshed import StrPath, SupportsRead
 from collections.abc import MutableSequence, Sequence
 from typing import Any
+from typing_extensions import Literal
 
 g_token_cache: dict[str, list[tuple[str, str]]]  # undocumented
+linesep: str
```
AlexWaygood, line 7: This is just a re-export from `os`; we shouldn't include it in the stub, I don't think. It would be a code smell for people to do `from chevron.renderer import linesep` instead of `from os import linesep`, and it would be weird for `linesep` to start popping up in IDE autocompletes. This should just be allowlisted.

stubs/chevron/chevron/main.pyi
sobolevn: There's something strange. Why is `version` reported and `render` is not? They are imported together: https://github.com/noahmorrison/chevron/blob/5e1c12827b7fc3db30cb3b24cae9a7ee3092822b/chevron/main.py#L7-L8

hauntsaninja: stubtest is smart enough to know that `render` comes from another module because it's a function. But unfortunately, `version` is just a str, so Python doesn't track which module it comes from.

AlexWaygood (approved):
Thanks!

PR #8888, opened 2022-10-13
Fix password hashes being non-optional for verify()

Ovsyanka83:
passlib allows hash argument to be None for convenience but the stubs do not indicate that. 

```python
if hash is None:
    # convenience feature -- let apps pass in hash=None when user
    # isn't found / has no hash; useful because it invokes dummy_verify()
    self.dummy_verify()
    return False
```


srittau (approved):
Thanks!

hauntsaninja (commented):
Is this definitely the case? It's documented as being only unicode or bytes here: https://foss.heptapod.net/python-libs/passlib/-/blob/branch/stable/passlib/context.py#L2280 

And elsewhere in the file they seem to note where things can be None, e.g. https://foss.heptapod.net/python-libs/passlib/-/blob/branch/stable/passlib/context.py#L2087

Ovsyanka83:
Yes, it is definitely the case:
verify: https://foss.heptapod.net/python-libs/passlib/-/blob/branch/stable/passlib/context.py#L2338
verify_and_update: https://foss.heptapod.net/python-libs/passlib/-/blob/branch/stable/passlib/context.py#L2417

srittau:
@hauntsaninja: https://foss.heptapod.net/python-libs/passlib/-/blob/branch/stable/passlib/context.py#L2272 says:

```python
        :type hash: unicode or bytes
        :arg hash:
            hash string to compare to

            if ``None`` is passed in, this will be treated as "never verifying"
```

And there is an explicit `None` check.

(Same for `verify_and_update()`.)

hauntsaninja:
I was definitely looking at the `secret` argument :-) Sorry for the noise, this is what I get for reviewing a PR before bedtime.

Ovsyanka83:
@srittau @hauntsaninja that was the fastest review/merge of my pull request I've ever seen in an open source project. 

You guys rock!

PR #8890, opened 2022-10-13
asyncio: fix SubprocessTransport.send_signal type

allisonkarlitskaya:
See https://github.com/python/mypy/issues/13885

This method returns `None`, not `int`.


srittau (approved):
Thanks!

PR #8891, opened 2022-10-13
[fpdf2] [drawing.pyi] Return types added

darshanip:
This PR adds return types in fpdf2. Please verify my initial commit, if it looks good I'll complete this file in same pattern.
@AlexWaygood


stubs/fpdf2/fpdf/drawing.pyi
```diff
@@ -11,8 +11,8 @@ from .syntax import Name, Raw
 
 __pdoc__: dict[str, bool]
 
-def force_nodocument(item): ...
-def force_document(item): ...
+def force_nodocument(item) -> item: ...
+def force_document(item) -> item: ...
```
AlexWaygood, line 15: The names we use in type annotations need to be names that are already defined in the global namespace. `item` here is not defined in the global namespace.

To say "this function returns the same thing that we put into the function", we can use a `TypeVar` (_type variable_). The docs for `TypeVar` are here: https://docs.python.org/3/library/typing.html#typing.TypeVar.

For example, in this case we'd use `TypeVar` like this:

```python
_T = TypeVar("_T")
def force_document(item: _T) -> _T: ...
```

The fact that the `item` parameter and the return type are annotated with the same type variable indicates to a static type checker that the same type is returned as the type that was passed in.

You'll need to add `from typing import TypeVar` at the top of the file.

AlexWaygood (commented):
Thanks! :) There's one thing that's not quite right -- see my comment below.

stubs/fpdf2/fpdf/drawing.pyi
darshanip: @AlexWaygood Yups thanks, it makes sense. 👍

darshanip:
@AlexWaygood, Can I get highlights on issues in code and also by which pre-commit.ci is failing.

stubs/fpdf2/fpdf/drawing.pyi
```diff
@@ -305,42 +307,44 @@ class PaintedPath:
...
+    def paint_rule(self: Self) -> Self: ...
     @paint_rule.setter
     def paint_rule(self, style) -> None: ...
     @property
-    def clipping_path(self): ...
+    def clipping_path(self: Self) -> Self: ...
     @clipping_path.setter
     def clipping_path(self, new_clipath) -> None: ...
+    @contextmanager
     def transform_group(self, transform) -> _GeneratorContextManager[Incomplete]: ...
```
Akuli, line 328: This should no longer be `_GeneratorContextManager`. See e.g. https://stackoverflow.com/q/49733699
```suggestion
    def transform_group(self, transform) -> Iterator[Incomplete]: ...
```

stubs/fpdf2/fpdf/drawing.pyi
```diff
@@ -4,15 +4,17 @@ from collections import OrderedDict
 from collections.abc import Generator
 from contextlib import _GeneratorContextManager
```
Akuli, line 5:
```suggestion
```

stubs/fpdf2/fpdf/drawing.pyi
```diff
@@ -4,15 +4,17 @@ from collections import OrderedDict
...
 from re import Pattern
-from typing import ClassVar, NamedTuple
+from typing import ClassVar, NamedTuple, TypeVar
```
Akuli, line 7:
```suggestion
from typing import ClassVar, NamedTuple, TypeVar, Iterator
```

stubs/fpdf2/fpdf/drawing.pyi
```diff
@@ -2,17 +2,18 @@ import decimal
 from _typeshed import Incomplete, Self
 from collections import OrderedDict
 from collections.abc import Generator
-from contextlib import _GeneratorContextManager
 from re import Pattern
```
Akuli, line 5:
```suggestion
from contextlib import contextmanager
from re import Pattern
```

stubs/fpdf2/fpdf/drawing.pyi
```diff
@@ -1,18 +1,20 @@
...
+from typing import ClassVar, NamedTuple, TypeVar
 from typing_extensions import TypeAlias
 
 from .syntax import Name, Raw
 
 __pdoc__: dict[str, bool]
 
-def force_nodocument(item): ...
-def force_document(item): ...
+_T = TypeVar("_T")
```
Akuli, line 14: The docstring says "class or method. We shouldn't just allow everything: if you somehow end up doing `force_nodocument("blah blah")`, that will fail. Because classes and methods are both callable, we can restrict `_T` like this:

```suggestion
_T = TypeVar("_T", bound=Callable[Any])
```

You will need to import `Callable` and `Any` from `typing`.

stubs/fpdf2/fpdf/drawing.pyi
```diff
@@ -23,10 +25,10 @@ STR_ESC: Pattern[str]
 STR_ESC_MAP: dict[str, str]
 
 class GraphicsStateDictRegistry(OrderedDict[Raw, Name]):
-    def register_style(self, style: GraphicsStyle): ...
+    def register_style(self, style: GraphicsStyle) -> Name | None: ...
 
-def number_to_str(number): ...
-def render_pdf_primitive(primitive): ...
+def number_to_str(number) -> str: ...
+def render_pdf_primitive(primitive) -> str: ...
```
Akuli, line 31:
```suggestion
def render_pdf_primitive(primitive) -> Raw: ...
```

stubs/fpdf2/fpdf/drawing.pyi
```diff
@@ -62,13 +64,21 @@ class Point(NamedTuple):
...
     def __neg__(self): ...
+    @force_document
     def __mul__(self, other): ...
+    @force_document
     def __rmul__(self, other): ...
+    @force_document
     def __truediv__(self, other): ...
+    @force_document
     def __floordiv__(self, other): ...
+    @force_document
```
Akuli, line 81: `@force_document` and `@force_nodocument` do nothing in the stub file, because these decorators return the same type `_T` that is passed to them as input. I suggest deleting them, unless you have a specific reason to keep them.

stubs/fpdf2/fpdf/drawing.pyi
```diff
@@ -322,25 +362,26 @@ class PaintedPath:
     def clipping_path(self): ...
     @clipping_path.setter
     def clipping_path(self, new_clipath) -> None: ...
-    def transform_group(self, transform) -> _GeneratorContextManager[Incomplete]: ...
+    @contextmanager
+    def transform_group(self, transform) -> Iterator[Incomplete]: ...
```
Akuli, line 366: It yields `self`, and not using `Any` in this method will probably help a lot in type checking. You shouldn't lose autocompletions, type checker errors etc only because you decided to apply a transform :)

```suggestion
    def transform_group(self: Self, transform) -> Iterator[Self]: ...
```

stubs/fpdf2/fpdf/drawing.pyi
```diff
@@ -361,15 +402,16 @@ class GraphicsContext:
     def __init__(self) -> None: ...
     def __deepcopy__(self: Self, memo) -> Self: ...
     @property
-    def transform(self): ...
+    def transform(self: Self) -> Self: ...
```
Akuli, line 405:
```suggestion
    def transform(self) -> Transform | None: ...
```

stubs/fpdf2/fpdf/drawing.pyi
```diff
@@ -361,15 +402,16 @@ class GraphicsContext:
...
     @transform.setter
     def transform(self, tf) -> None: ...
     @property
-    def clipping_path(self): ...
+    def clipping_path(self: Self) -> Self: ...
```
Akuli, line 409:
```suggestion
    def clipping_path(self) -> ClippingPath | None: ...
```

stubs/fpdf2/fpdf/drawing.pyi
darshanip: Sure, I'll remove it.

stubs/fpdf2/fpdf/drawing.pyi
```diff
@@ -1,18 +1,20 @@
...
+from typing import ClassVar, NamedTuple, TypeVar
 from typing_extensions import TypeAlias
 
 from .syntax import Name, Raw
 
 __pdoc__: dict[str, bool]
 
-def force_nodocument(item): ...
-def force_document(item): ...
+_T = TypeVar("_T", bound=Callable[Any])
```
Akuli, line 14: Maybe we should rename this to something like `_CallableT`. A common convention (at least within typeshed) is that `_T` can be anything, and bound types have the bound in the name.

darshanip: I've updated this.

stubs/fpdf2/fpdf/drawing.pyi
```diff
@@ -1,18 +1,20 @@
...
+from typing import ClassVar, NamedTuple, TypeVar
 from typing_extensions import TypeAlias
 
 from .syntax import Name, Raw
 
 __pdoc__: dict[str, bool]
 
-def force_nodocument(item): ...
-def force_document(item): ...
+_CallableT = TypeVar("_CallableT", bound=Callable[Any])
```
Akuli, line 14: Just realized that `Callable` takes two things, the argument types (or `...` to denote that anything is fine) and the return type. Searching the repository (e.g. `git grep bound=Callable`) shows plenty of examples.

```suggestion
_CallableT = TypeVar("_CallableT", bound=Callable[..., Any])
```

stubs/fpdf2/fpdf/drawing.pyi
```diff
@@ -1,18 +1,20 @@
 import decimal
 from _typeshed import Incomplete, Self
 from collections import OrderedDict
-from collections.abc import Generator
-from contextlib import _GeneratorContextManager
+from collections.abc import Generator, Iterator
```
Akuli, line 4:
```suggestion
from collections.abc import Generator, Iterator, Callable
```

stubs/fpdf2/fpdf/drawing.pyi
```diff
@@ -1,18 +1,20 @@
...
+from contextlib import contextmanager
 from re import Pattern
-from typing import ClassVar, NamedTuple
+from typing import ClassVar, NamedTuple, TypeVar
```
Akuli, line 7:
```suggestion
from typing import ClassVar, NamedTuple, TypeVar, Any
```

Akuli (approved):
Thanks!

darshanip:
> 

Thanks helping out in very first PR to Open Source.

PR #8895, opened 2022-10-13
PyYaml missed `*` re-exports

Avasam:
Noticed here: https://github.com/aws/aws-sam-cli/blob/develop/samcli/yamlhelper.py#L27

### DIFF
--- a/stubs/PyYAML/yaml/__init__.pyi
+++ b/stubs/PyYAML/yaml/__init__.pyi
@@ -3,19 +3,19 @@ from re import Pattern
 from typing import Any, TypeVar, overload
 from typing_extensions import TypeAlias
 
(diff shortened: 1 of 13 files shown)


JelleZijlstra:
Are all these star imports really intended to be part of the library's API? We may be serving users better if we have them import all of these objects from the files they're defined in.

Avasam:
> Are all these star imports really intended to be part of the library's API? We may be serving users better if we have them import all of these objects from the files they're defined in.

That's fair. I just figured "since it's technically valid". But also it's not as if these imports are unavailable otherwise.
I don't mind if this is the decision is to close this PR.

JelleZijlstra:
Closing then, feel free to reopen if you disagree.

PR #8898, opened 2022-10-14
Fix jsonschema's RefResolver context manager return types

kaisen:
jsonschema's `RefResolver` class has [2 methods](https://github.com/python-jsonschema/jsonschema/blob/8fd12e29d8ad6e44ab48e0cd54c87ffe015b6c2a/jsonschema/validators.py#L808) that are decorated with `@contextmanager` but the existing stubs have `None` as the return type.

This sets the return type as `contextlib.AbstractContextManager` ([per PEP585](https://peps.python.org/pep-0585/#implementation)) so code using the methods with `with` pass mypy checks.

Verified that `scripts/run_tests.py` passes.


stubs/jsonschema/jsonschema/validators.pyi
```diff
@@ -86,8 +87,8 @@ class RefResolver:
     def resolution_scope(self): ...
     @property
     def base_uri(self): ...
-    def in_scope(self, scope) -> None: ...
-    def resolving(self, ref) -> None: ...
+    def in_scope(self, scope) -> AbstractContextManager[None]: ...
+    def resolving(self, ref) -> AbstractContextManager[Any]: ...
```
Akuli, line 91: - You can apparently use `@contextmanager` in stubs, just like you would use it in a `.py` file. See e.g. https://stackoverflow.com/a/49736916
- We would prefer using `Incomplete` (imported from `_typeshed`) instead of `Any` in the return type of `resolving()`. We try to use `Any` only when it isn't possible to express the correct type.

Akuli (approved):
Thanks!

PR #8902, opened 2022-10-15
Mark `flake8-2020` as complete

sobolevn:
Source: https://github.com/asottile/flake8-2020


sobolevn:
This is really weird! GitHub is buggy right now.

stubs/flake8-2020/@tests/stubtest_allowlist.txt
```diff
@@ -0,0 +1,5 @@
+# We don't need violation definitions:
+flake8_2020.YTT\d+
```
AlexWaygood, line 2: Why not?

stubs/flake8-2020/@tests/stubtest_allowlist.txt
```diff
@@ -0,0 +1,5 @@
...
+
+# Has many implementation details:
+flake8_2020.Visitor
```
AlexWaygood, line 5: I don't think this actually has any public methods that aren't already included in typeshed's stub for the base class, `ast.NodeVisitor`: https://github.com/python/typeshed/blob/38d0f691e6afba8e88347726b7d56ed361ffd7bf/stdlib/ast.pyi#L37

AlexWaygood (approved):
Thanks!

sobolevn:
@AlexWaygood shouldn't it be `def __getattr__(name: str) -> Incomplete: ...`?
I think it should. I will send a PR for this.

PR #8903, opened 2022-10-15
Use `Incomplete` instead of `Any` in `__getattr__`

sobolevn:
I've only touched things that were explicit about being incomplete.
All other `__getattr__` items are untouched.
Refs https://github.com/python/typeshed/pull/8902


stdlib/types.pyi
```diff
@@ -591,7 +591,7 @@ if sys.version_info >= (3, 9):
             @property
             def __typing_unpacked_tuple_args__(self) -> tuple[Any, ...] | None: ...
 
-        def __getattr__(self, name: str) -> Any: ...  # incomplete
+        def __getattr__(self, name: str) -> Incomplete: ...
```
JelleZijlstra, line 594: I think this should actually be Any, as GenericAlias delegates attr lookup to its origin.

JelleZijlstra:
```suggestion
        def __getattr__(self, name: str) -> Any: ...
```

AlexWaygood: Yeah, this one really is inexpressible in the stub

stdlib/__main__.pyi
```diff
@@ -1,3 +1,3 @@
-from typing import Any
+from _typeshed import Incomplete
 
-def __getattr__(name: str) -> Any: ...
+def __getattr__(name: str) -> Incomplete: ...
```
AlexWaygood, line 3: We want to use the `Incomplete` symbol as a signal for contributors: if a function has `Incomplete` as a parameter or return type, it's us saying "Hey! We'd welcome a PR to fix this thing!"

I'm not sure it's useful to use `Incomplete` here -- is there any way a contributor would be able to improve the stubs for `stdlib/__main__.pyi`?

AlexWaygood (approved):

AlexWaygood:
Thanks!

PR #8904, opened 2022-10-15
Mark `hdbcli` as complete

sobolevn:
Docs: https://help.sap.com/docs/SAP_HANA_CLIENT/f1b440ded6144a54ada97ff95dac7adf/f3b8fabf34324302b123297cdbe710f0.html?locale=en-US


JelleZijlstra:
stubtest finds a few missing objects.

sobolevn:
This was quite hard.
I've used https://help.sap.com/docs/SAP_HANA_CLIENT/f1b440ded6144a54ada97ff95dac7adf/d0c010fa5c4e4e4cba7b791cd1c9479b.html?locale=en-US to get at least some information.
I've also best-guessed some methods. 

But, most of missing methods are just marked as `Incompelete`. Because I cannot find the source code, I cannot make it work locally, it is shipped as `.so` file with no `__text_signature__` props.

It is just like a well-forgotten dive into corporate-software development, not a big fan.

sobolevn:
Any updates? :)

PR #8908, opened 2022-10-16
`datetime.tzinfo` is abtract

sobolevn:
This class is not marked as abstract, but it is documented: 

> Abstract base class for time zone info classes.
> Subclasses must override the name(), utcoffset() and dst() methods.

Source: https://github.com/python/cpython/blob/f6b1e4048dc353aecfbfbae07de8212900632098/Lib/datetime.py#L1222-L1243


sobolevn:
I was following return types from https://docs.python.org/3/library/datetime.html#datetime.timezone

sobolevn:
`ZoneInfo` can return `None` because of https://github.com/python/cpython/blob/f6b1e4048dc353aecfbfbae07de8212900632098/Lib/zoneinfo/_zoneinfo.py#L418

sobolevn:
`dateutil` source: https://github.com/dateutil/dateutil/blob/7c5c98959bd0e9eea3c55fdf78bef3ae5b8c995e/src/dateutil/tz/_common.py#L149

sobolevn:
Example in Sphinx is technically correct:

```python

class LocalTimeZone(tzinfo):
    def __init__(self, *args: Any, **kwargs: Any) -> None:
        super().__init__(*args, **kwargs)
        self.tzdelta = tzdelta

    def utcoffset(self, dt: Optional[datetime]) -> timedelta:
        return self.tzdelta

    def dst(self, dt: Optional[datetime]) -> timedelta:
        return timedelta(0)


ltz = LocalTimeZone()
```

`LocalTimeZone` misses `tzname` method definition.
Link: https://github.com/sphinx-doc/sphinx/blob/c1f2a52d5d97f874d84e4d5421c99a566242af6b/sphinx/builders/gettext.py#L191

I will send a PR.

hauntsaninja (commented):
Do we need to add the metaclass, since runtime doesn't have it? mypy allows for checking abstractmethod without metaclass

JelleZijlstra (approved):

sobolevn:
> Do we need to add the metaclass, since runtime doesn't have it? mypy allows for checking abstractmethod without metaclass

Yes, we sometimes need this.

Please, take a look at this error:

```
error: not checking stubs due to mypy build errors:
stdlib/zoneinfo/__init__.pyi:12: error: Class zoneinfo.ZoneInfo has abstract attributes "dst", "tzname", "utcoffset"
stdlib/zoneinfo/__init__.pyi:12: note: If it is meant to be abstract, add 'abc.ABCMeta' as an explicit metaclass
```

sobolevn:
Basically, `ABCMeta` metaclass is an explicit marker that some class is indeed abstract. That it is not just a missing method by accident.

hauntsaninja:
Oh interesting, didn't realise there was this stub-only check: https://github.com/python/mypy/blob/5e1e26eba15b43449bdce4010bf16d160d3ec505/mypy/semanal_classprop.py#L97

AlexWaygood:
We need the metaclass change in `zoneinfo.ZoneInfo`, but I don't think we need the metaclass change for `datetime.tzinfo` itself

PR #8910, opened 2022-10-17
Remove `ABCMeta` from `tzinfo`

sobolevn:
Refs https://github.com/python/typeshed/pull/8908#issuecomment-1280377066


AlexWaygood (approved):
Thank you! 😊

PR #8914, opened 2022-10-17
Fix stubtest failures on `SQLAlchemy`

### LINKED ISSUE #8909: Daily tests failed on Mon Oct 17 2022

AlexWaygood:
And pin `SQLAlchemy` to a micro version, so that it stops breaking typeshed's CI every other week.

Fixes #8909.

### DIFF
--- a/stubs/SQLAlchemy/METADATA.toml
+++ b/stubs/SQLAlchemy/METADATA.toml
@@ -1,4 +1,4 @@
-version = "1.4.*"
+version = "1.4.42"
 extra_description = """\
   The `sqlalchemy-stubs` package is an alternative to this package and also \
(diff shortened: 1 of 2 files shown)


AlexWaygood:
The runtime appears to have just deleted the problematic attribute in the latest release, so that's why I'm doing the same here: https://github.com/sqlalchemy/sqlalchemy/compare/rel_1_4_41...rel_1_4_42

srittau (approved):

hauntsaninja:
Confirmed that stub_uploader uploaded the package and the new versioning logic did the right thing here. And once sqlalchemy releases 1.4.54, we should be back in "expected" version land, where stub_uploader versions correspond to upstream versions.

PR #8918, opened 2022-10-18
Add types to invoke.Task.__call__

### LINKED ISSUE #8915: invoke stubs lead to Untyped decorator
When type checking the below code I get an error:

```py
from invoke import task, Context

@task
def other(context: Context)-> None:
    pass

@task(other)
def that(context: Context) -> None:
    pass
```

Then execute `mypy foo.py --strict`

```
foo.py:7: error: Untyped decorator makes function "that" untyped
```

My question is: how can resolve this. It seems that mypy can't figure out if overload 1 or 2 should be used of `invoke/tasks,pyi->task`. Or is this a mypy bug/limitation?

kasium:
Closes #8915


JelleZijlstra:
Adding `Any` doesn't really make things better. Can this class be made generic over a ParamSpec instead? `ParamSpec` looks like the right tool but I haven't checked in detail.

AlexWaygood:
I agree that making the class generic around `ParamSpec` is probably the way to go here. You probably want to do something like this:

```diff
@@ -1,17 +1,20 @@
  from _typeshed import Self
  from collections.abc import Callable, Iterable
- from typing import Any, TypeVar, overload
+ from typing import Any, Generic, TypeVar, overload
+ from typing_extensions import ParamSpec
  
  from .config import Config
  from .context import Context
  from .parser import Argument
  
- _TaskT = TypeVar("_TaskT", bound=Task)
+ _P = ParamSpec("_P")
+ _R = TypeVar("_R", covariant=True)
+ _TaskT = TypeVar("_TaskT", bound=Task[..., Any])
  
  NO_DEFAULT: object
  
- class Task:
-     body: Callable[..., Any]
+ class Task(Generic[_P, _R]):
+     body: Callable[_P, _R]
      __doc__: str | None
      __name__: str
      __module__: str
@@ -29,7 +32,7 @@ class Task:
      autoprint: bool
      def __init__(
          self,
-         body: Callable[..., Any],
+         body: Callable[_P, _R],
          name: str | None = ...,
          aliases: tuple[str, ...] = ...,
          positional: Iterable[str] | None = ...,
@@ -47,7 +50,7 @@ class Task:
      def name(self): ...
      def __eq__(self, other: Task) -> bool: ...  # type: ignore[override]
      def __hash__(self) -> int: ...
-     def __call__(self, *args: Any, **kwargs: Any) -> Any: ...
+     def __call__(self, *args: _P.args, **kwargs: _P.kwargs) -> _R: ...
      @property
      def called(self) -> bool: ...
      def argspec(self, body): ...
```

You might need to update some other parts of the file as well, though.

hauntsaninja (commented):
Lol, it looks like `task` will literally error out if you use it to decorate a function with annotations (the example in #8915 fails for me at runtime). So questionable whether the generic is worth it... I guess it would preserve arity, but this does seem like a pretty hostile environment for accurate typing

kasium:
@hauntsaninja I'm aware of this. invoke still suffers from some limitations which can still be worked around. Since we have the types here, I would still vote to keep this

AlexWaygood (approved):
This LGTM, and @kasium is the main contributor to these stubs, so I'm going to merge. We can always revert if this causes problems.

PR #8919, opened 2022-10-18
Mark `mock` as completed

sobolevn:
Source: https://github.com/testing-cabal/mock

Should I also mark it as `no_longer_updated = true`?


srittau:
While there was *some* activity in February, I think we can remove this in foreseeable future. So: yes, in my opinion.

stubs/mock/mock/backports.pyi
```diff
@@ -0,0 +1,11 @@
+import sys
+
+if sys.version_info[:2] >= (3, 8):
+    from asyncio import iscoroutinefunction as iscoroutinefunction  # pyright: ignore[reportGeneralTypeIssues]
```
JelleZijlstra, line 4: What is pyright unhappy about here?

JelleZijlstra (approved):
Thanks, just one question

stubs/mock/mock/backports.pyi
sobolevn: Arg names / signatures do not match: https://github.com/python/typeshed/actions/runs/3272486451/jobs/5383555539

JelleZijlstra: It shouldn't matter because they're in different version branches. The problem is that you have a `[:2]` on the version check, that shouldn't be there. I'm surprised flake8-pyi doesn't complain.

stubs/mock/mock/backports.pyi
```diff
@@ -0,0 +1,11 @@
...
+    from unittest import IsolatedAsyncioTestCase as IsolatedAsyncioTestCase  # pyright: ignore[reportGeneralTypeIssues]
```
JelleZijlstra, line 5:
```suggestion
if sys.version_info >= (3, 8):
    from asyncio import iscoroutinefunction as iscoroutinefunction
    from unittest import IsolatedAsyncioTestCase as IsolatedAsyncioTestCase
```

JelleZijlstra (approved):

AlexWaygood:
> Should I also mark it as `no_longer_updated = true`?

FWIW this feels a tiny bit premature to me — the runtime package describes itself as a "rolling backport" of CPython `main` for Python 3.6+, and there's nothing explicitly stating that it's no longer maintained. The lack of recent activity might just be because there haven't been any significant new features to backport recently.

But I also don't feel too strongly about it, and don't want to block this PR if everybody else agrees with marking the stubs as `no_longer_updated` 🙂

AlexWaygood (approved):
Thanks!

PR #8921, opened 2022-10-18
Add pathlib.Path.walk

Ovsyanka83:


stdlib/pathlib.pyi
```diff
@@ -198,6 +200,10 @@ class Path(PurePath):
         def write_text(self, data: str, encoding: str | None = ..., errors: str | None = ...) -> int: ...
     if sys.version_info >= (3, 8):
         def link_to(self, target: StrPath | bytes) -> None: ...
+    if sys.version_info >= (3, 12):
+        def walk(
+            self: Self, top_down: bool = ..., on_error: _OnError | None = ..., follow_symlinks: bool = ...
```
AlexWaygood, line 205: The `_OnError` `TypeAlias` is only used once in this file (unlike in our stubs for `os`, where the alias is reused multiple times in the file). I don't think there's any reason to define it at all here in `pathlib.pyi`; let's just use `Callable[[OSError], object]` directly:

```suggestion
            self: Self, top_down: bool = ..., on_error: Callable[[OSError], object] | None = ..., follow_symlinks: bool = ...
```

stdlib/pathlib.pyi
```diff
@@ -198,6 +200,10 @@ class Path(PurePath):
...
+        ) -> Iterator[tuple[Self, list[AnyStr], list[AnyStr]]]: ...
```
AlexWaygood, line 206: Since `pathlib` only supports string paths (not bytes paths), I think these are guaranteed to be lists of `str`s. This is different to `os.walk`, which supports string paths and bytes paths.

```suggestion
        ) -> Iterator[tuple[Self, list[str], list[str]]]: ...
```

AlexWaygood (commented):
Thank you! (And, thanks for adding the method to CPython!) A few comments below:

AlexWaygood (approved):
Thank you!

Issue #8660, opened 2022-09-01
Run stubtest under different OS environments


Avasam:
This is an issue I'm having while adding D3DShot #8652, ~~PyAutoGUI #8654~~ and pywin32 stubs. Some modules are highly dependent on the OS, and as such, I have to add them to the `stubtest_allowlist`, even though I can run it just fine locally.

For this reason, and to help with stub completeness (#8481), I suggest that stubtest should ~~by default be run under Windows, Linux and Darwin (just like mypy and pyright already are)~~ be configurable to run under Windows and/or Linux and/or Darwin.

This also means we should be able to specify the OS in stubtest_allowlist. Whether it's by having a specific file (`stubtest_allowlist.linux.txt`), prefixing the line (`linux:pyautogui._pyautogui_osx`) or any other proposals.

AlexWaygood:
We already do this in CI for our stdlib stubs, and it works pretty well. For most of our third-party stubs, it might be a bit overkill; but on the other hand, it probably wouldn't slow down CI that much to have three stubtest runs instead of one on a PR relating to third-party stubs.

I'm curious for @hauntsaninja's thoughts :)

AlexWaygood:
> We ran into CI limits in the past and we don't need to waste them on what is inconsequential for most stubs.

It's an unrelated change, but we could consider doing something similar to https://github.com/python/mypy/pull/13249. That would reduce our risk of running into CI limits.

hauntsaninja:
I don't think this is worth it for third party stubs. In practice, the only OS dependent things I remember in third party stubs are implementation details that we usually would avoid adding types for.

If the concern is something something fiddliness with the allowlists, consider using the empty regex trick, where if you do `(entry)?` stubtest will not complain about that allowlist entry being unused.

Avasam:
@hauntsaninja The concern is more that for libraries with OS-specific support and modules, stubtest is pretty much useless in detecting missing or invalid types. This is especially prevalent with I/O (mouse, keyboard, displays, ...) where the API changes wildly between OSes (so it is wrong to assume that the only difference is implementation details).

I can fully understand concerns about CI limits however, especially when this scenario doesn't apply for most stubs..

JelleZijlstra:
Perhaps we could run OS-specific stubtest only for specific packages where we know there are differences.

Avasam:
I just stumbled upon a new issue relating to this when working on pywin32 stubs:
[...]

PR #8923, opened 2022-10-18
Run stubtest on stubs on different platforms

sobolevn:
Open problems:
- [ ] How to install deps on other platforms. `brew` works well on `macos`, but no idea how to use windows (I've never touched it since 2008)
- [x] Listing platform specific packages. Which ones should we add?
- [x] I still need to change `daily.yml` to do the same thing for all stubs
- [x] How to test this? Not quite clear for now. I will need to tweak some values 

Refs https://github.com/python/typeshed/issues/8660


sobolevn:
I have no idea how to fix this Windows problem:

```

Run STUBS=$(
ParserError: D:\a\_temp\4a20ead6-174f-4640-96c9-b47d87c97584.ps1:8
Line |
   8 |  if [ -n "$STUBS" ]; then
     |    ~
     | Missing '(' after 'if' in if statement.
```

It is valid `sh`, what more do you need? 🤣 
Probably we need two different test runners: one for `win` and one for `posix`

sobolevn:
Looks like it works! https://github.com/python/typeshed/actions/runs/3273273447/jobs/5385277526

Avasam:
> It is valid sh, what more do you need?

But not valid powershell! Although you already figured out that you can specify the default shell in github actions. Which is what I would've recommended over:
> Can we use WSL in GitHub Actions? That might be an easy workaround.

> How to install deps on other platforms
Chocolatey is what I see most often (at least when developing with Node and Python).

> Listing platform specific packages. Which ones should we add?
Whatever's needed as you develop/test this I guess?  
Taking a look at `stubtest_allowlist.txt` files seem to be a good starting point to get an idea of which stubs have os-specific needs.

tests/stubtest_third_party.py
```diff
@@ -34,6 +34,11 @@ def run_stubtest(dist: Path, *, verbose: bool = False) -> bool:
         print(colored("skipping", "yellow"))
         return True
 
+    platform = stubtest_meta.get("platform", None)
+    if platform is not None and sys.platform not in platform:
```
AlexWaygood, line 38: Wouldn't it make more sense just to have this field default to "only test on linux" for now?

```suggestion
    platforms_to_test = stubtest_meta.get("platform", ["linux"])
    if sys.platform not in platforms_to_test:
```

sobolevn: No, because we want more coverage for more packages. Most of the packages do not rely on platform, so we can be quite safe.

Akuli:
You can use `shell: bash` on Windows github actions.

sobolevn:
I already do that, thanks!

tests/stubtest_third_party.py
AlexWaygood: I agree that increasing coverage in the long term is an important goal. However, the consensus in #8660 was that it isn't necessary for most stubs to run stubtest on multiple platforms, so it would just slow down CI.

I'd prefer it if this PR focused on adding the necessary infrastructure, without changing the default for all stubs. Then, after this PR is merged, we can experiment with changing the default for all stubs, and evaluate whether it significantly slows down CI (or leads to widespread CI failures on stubs that haven't been stubtested on Windows before).

AlexWaygood:
Could you make this change to the `METADATA.toml` file for D3DShot as part of this PR, to test that the Windows part of this is working?

```diff
  [tool.stubtest]
- # The library only works on Windows; we currently only run stubtest on Ubuntu for third-party stubs in CI.
- # See #8660
- skip = true
+ platform = ["win32"]
```

Akuli:
Nit: I would prefer if the metadata key was named `platforms` instead of `platform`.

.github/workflows/stubtest_stubs.yml
```diff
@@ -1,4 +1,4 @@
-name: Test
+name: Run stubtest on stubs
```
AlexWaygood, line 1:
```suggestion
name: Third-party stubtest
```

Could you also change this line to something like "Stdlib stubtest", so that it's easy to distinguish the two workflows:

https://github.com/python/typeshed/blob/f8ae34c34e01736ed7f38c35ed068420ee74fc71/.github/workflows/stubtest_stdlib.yml#L1

It would also be good to change the name of this workflow file so that it has "third-party" in it

AlexWaygood:
You might have to add a `@tests/requirements-stubtest.txt` file to `D3DShot`, and pin `Pillow` to a specific revision, in order to get it to install.

sobolevn:
Output from windows daily run: https://github.com/sobolevn/typeshed/actions/runs/3280905801/jobs/5402262235

sobolevn:
Since we now have platform-specific `stubtest` failures: https://github.com/sobolevn/typeshed/actions/runs/3283716889/jobs/5408788907

I will introduce platform-specific allowlists. Something similar to what we have in `stdlib/`

stubs/D3DShot/@tests/requirements-stubtest.txt
```diff
@@ -0,0 +1 @@
+pillow
```
JelleZijlstra, line 1: If I recall correctly this needs to be an older version of Pillow, the newest release doesn't work.

sobolevn: Yes, they have some complicated setup: https://pillow.readthedocs.io/en/latest/installation.html
Снимок экрана 2022-10-19 в 21 10 15

I will try different ones.

Avasam: There's a weird dependency conflict here.
IIRC, D3DShot 0.13 didn't lock the pillow version, but 0.15 did.
And it locked it to 7.1.2 which is not 3.9 compatible.

There was a PR to fix that, but the repository got archived not long ago (most of the forks are just that change). Even more recently, most of SerpentAI's repos have been archived as well. The timing with the addition to Typeshed is unfortunate, (the stub was started before the repo got archived). Using D3DShot now is somewhat of a stopgap to let upcoming replacements mature. With an already planned removal (#8929), I think it may be acceptable to just keep `skip = true` in this stub, unless you can let pip ignore the requirement, or let it build from source with zlib.

sobolevn: It is said that we need `Pillow>=7.2.0`: https://github.com/python/typeshed/pull/8652
But, we already have `Pillow>=7.2.0`, looks like we need an upper bound as well.
@Avasam, can you please take a look?

Avasam: What I didn't realize when I wrote that comment (https://github.com/python/typeshed/pull/8652#issuecomment-1237571064) is that requirering `Pillow>=7.2.0` will install D3DShot 0.1.3 .

https://github.com/SerpentAI/D3DShot/commit/a37abe4b6845246fe3437a9557f073b1b19f1963#diff-60f61ab7a8d1910d86d9fda2261620314edcae5894d5aaa236b821c7256badd7L18
Because pip's dependency resolver will see that 0.1.3 has `Pillow>=5.0.0` (which any newer version matches) while 0.1.4 is pinned at `Pillow~=7.1.2`

sobolevn: No, it still does not work: https://github.com/sobolevn/typeshed/actions/runs/3284090937/jobs/5409620328

I think that existing deps can be re-installed here: https://github.com/python/typeshed/blob/master/tests/stubtest_third_party.py#L74

AlexWaygood: Okay, I endorse @Avasam's suggestion of leaving D3DShot as `skip = true` for now -- this is all slightly tangential to the main point of this PR, and we also have `pywin32` to test that this PR works correctly on Windows. We can try to figure out how to get stubtest to install this package after this PR is merged :)

sobolevn:
Daily tests also pass: https://github.com/sobolevn/typeshed/actions/runs/3285152647/jobs/5411937127

stubs/pywin32/@tests/stubtest_allowlist.txt
```diff
@@ -1,75 +0,0 @@
-# Not available at runtime. Contains type definitions that are otherwise not exposed
```
Avasam, line 1: Let's not forget to revert this before merging

sobolevn:
I have no idea what is going on with review requests! Sorry!

Avasam:
> I have no idea what is going on with review requests! Sorry!

Github only let's you request a single reviewer >.<

sobolevn:
We have a problem:

```
The conflict is caused by:
    pycln 2.1.1 depends on libcst<0.5.0 and >=0.3.10
    pytype 2022.8.30 depends on libcst>=0.4.5

To fix this you could try to:
1. loosen the range of package versions you've specified
2. remove package versions to allow pip attempt to solve the dependency conflict

ERROR: ResolutionImpossible: for help visit https://pip.pypa.io/en/latest/topics/dependency-resolution/#dealing-with-dependency-conflicts

```

sobolevn:
Me neither, `pip` has some known problems with its resolver. It might be the reason.

Avasam:
Weird, there's something very wrong with the dependency resolution on that run.
```
ERROR: Could not find a version that satisfies the requirement flake8-noqa==1.2.9 (from versions: 1.0.0, 1.0.1, 1.0.2, 1.0.3, 1.0.4, 1.0.5, 1.1.0, 1.2.0, 1.2.1)
ERROR: No matching distribution found for flake8-noqa==1.2.9
```

Tested this branch on a fork and it worked: https://github.com/Avasam/typeshed/pull/6

Maybe try to run again? Hopefully it's just a fluke. Maybe some bad cache, idk.

CONTRIBUTING.md
```diff
@@ -192,6 +192,9 @@ This has the following keys:
 * `apt_dependencies` (default: `[]`): A list of Ubuntu APT packages
   that need to be installed for stubtest to run successfully. These are
   usually packages needed to pip install the implementation distribution.
+* `platforms` (default: `["linux"]`): A list of supported OSes.
+  Can contain `win32`, `linux`, and `darwin` values.
+  If not specified, only `linux` is considered supported.
```
JelleZijlstra, line 197: We should specify that this should be set only if there's some specific reason to test on multiple OSes, because we don't want useless duplicate stubtest runs for all pure-Python packages.

.github/workflows/stubtest_third_party.yml
```diff
@@ -0,0 +1,71 @@
...
+              xvfb-run python tests/stubtest_third_party.py $STUBS
+            fi
+
+            if [ "${{ matrix.os }}" = "macos-latest" ]; then
+              # TODO: install brew packages
+              python tests/stubtest_third_party.py $STUBS
+            fi
+
+            if [ "${{ matrix.os }}" = "windows-latest" ]; then
+              # TODO: support windows
```
JelleZijlstra, line 66: Looks like it's supported, what's left to do?

stubs/pywin32/METADATA.toml
```diff
@@ -1,7 +1,6 @@
 version = "304.*"
+
 [tool.stubtest]
-# The library only works on Windows; we currently only run stubtest on Ubuntu for third-party stubs in CI.
-# See #8660
+platforms = ["win32"]
 # `ignore_missing_stub = false` crashes stubtest. This should be fixed in next mypy update
```
JelleZijlstra, line 5: Any reason to keep this comment?

tests/stubtest_third_party.py
```diff
@@ -34,6 +34,12 @@ def run_stubtest(dist: Path, *, verbose: bool = False) -> bool:
         print(colored("skipping", "yellow"))
         return True
 
+    # TODO: enable for `win32` and `darwin` by default?
```
JelleZijlstra, line 37: I think we don't want to do that.

stubs/pywin32/METADATA.toml
Avasam: This comment is accurate. `ignore_missing_stub = false` should be here. Along with the previous content of `stubtest_allowlist.txt`. However doing so will crash stubtest until the next mypy version, as the comment mentions.

So either we need `skip=true` here until then. Or revert the `stubtest_allowlist.txt` changes when mypy is updated.

CONTRIBUTING.md
```diff
@@ -195,6 +195,8 @@ This has the following keys:
 * `platforms` (default: `["linux"]`): A list of supported OSes.
   Can contain `win32`, `linux`, and `darwin` values.
   If not specified, only `linux` is considered supported.
+  Only add extra OSes to the test
+  if there is platform-specific code in a package.
```
Avasam, line 199: Just a suggestion: I'd specify platform-specific stubs. PyAutoGUI is a good example that has platform-specific code, but the public module (the one we stub), has platform-agnostic typing. (hence it doesn't need to be stub-tested on multiple platforms).

vobject, cffi, setuptools, colorama, etc. have specific allowlist entries because of windows, these are example of stubs that would benefit from running on multiple OSes. (less exclusions, more completion)

Avasam: Maybe something around the lines of `if there are platform-specific [type hints/typing] in a package's stub`
or `if a stub contains platform-specific types`

sobolevn: cc @AlexWaygood as some who is great with wording :)

AlexWaygood: Maybe `if there is platform-specific code in a stubs package` or `if there are platform-specific branches in a stubs package`?

CONTRIBUTING.md
```diff
@@ -192,6 +192,11 @@ This has the following keys:
 * `apt_dependencies` (default: `[]`): A list of Ubuntu APT packages
   that need to be installed for stubtest to run successfully. These are
   usually packages needed to pip install the implementation distribution.
+* `platforms` (default: `["linux"]`): A list of supported OSes.
+  Can contain `win32`, `linux`, and `darwin` values.
+  If not specified, only `linux` is considered supported.
+  Only add extra OSes to the test
+  if there is platform-specific code in a package.
```
sobolevn, line 199:
```suggestion
  Only add extra OSes to the test
  if there are platform-specific branches in a stubs package.
```

Avasam:
mypy 0.9.90 is released. stubtest no longer crashes on dll files. This PR is unblocked as soon as #9123 is merged!
Although stubtest did find a handful of new issues now that it doesn't crash midway, I'll go ahead and update those.

sobolevn:
I've never used `win32` package in my life, but I already don't like it 😆

Avasam:
> I've never used `win32` package in my life, but I already don't like it 😆

It's quite useful, but the documentation is... lacking to say the least. (I mostly look at the official microsoft doc and guess the equivalent python binding, it'd be even more of a pain without stubs...).

Anyway, for your allowlist issue, you might wanna base yourself on https://github.com/python/typeshed/pull/9124 (and/or wait/push for it to be merged first)

tests/get_packages.py
```diff
@@ -0,0 +1,23 @@
...
+platform = sys.platform
+distributions = sys.argv[1:]
+if not distributions:
+    distributions = os.listdir("stubs")
+
+metadata_mapping = {
+    "linux": "apt_dependencies",
+    # TODO: support others:
+    # "darwin": "brew_dependencies",
+    # "win32": "choco_dependencies",
```
AlexWaygood, line 16: I like that you've done this in an extensible way, but this doesn't need to be a TODO comment for now; we don't have any stubs that currently need installations using `brew` or `choco`

```suggestion
    # We could add others here if we run into stubs that need it:
    # "darwin": "brew_dependencies",
    # "win32": "choco_dependencies",
```

sobolevn:
Finally 🎉

stubs/pywin32/@tests/stubtest_allowlist_win32.txt
```diff
@@ -65,3 +65,15 @@ win32com(ext)?.axdebug.documents
...
+pythoncom.internal_error
+pythoncom.new
+pywintypes.TimeType
+pywintypes.WAVE_FORMAT_PCM
+win32.lib.pywintypes.TimeType
+win32.lib.pywintypes.WAVE_FORMAT_PCM
+win32.winxpgui.GetConsoleWindow
+win32.winxpgui.GetWindowRgnBox
+winxpgui.GetConsoleWindow
+winxpgui.GetWindowRgnBox
```
AlexWaygood, line 79:
```suggestion
```

AlexWaygood (approved):
Looks great to me! Thanks for all the work you've done on this!

CONTRIBUTING.md
```diff
@@ -192,6 +192,11 @@ This has the following keys:
 * `apt_dependencies` (default: `[]`): A list of Ubuntu APT packages
   that need to be installed for stubtest to run successfully. These are
   usually packages needed to pip install the implementation distribution.
+* `platforms` (default: `["linux"]`): A list of supported OSes.
+  Can contain `win32`, `linux`, and `darwin` values.
+  If not specified, only `linux` is considered supported.
```
JelleZijlstra, line 197:
```suggestion
* `platforms` (default: `["linux"]`): A list of OSes on which to run stubtest.
  Can contain `win32`, `linux`, and `darwin` values.
  If not specified, tests are run only on `linux`.
```

JelleZijlstra (commented):
Slight wording suggestion because many stubs may be "supported" on other platforms, we just don't need to run stubtest multiple times.

.github/workflows/stubtest_third_party.yml
```diff
@@ -0,0 +1,70 @@
...
+          python-version: "3.9"
+      - name: Install dependencies
+        run: pip install -r requirements-tests.txt
+      - name: Run stubtest
+        shell: bash
+        run: |
+          STUBS=$(
+            git diff --name-only origin/${{ github.base_ref }} HEAD |
+            # Uncomment the following to (very slowly) run on all third party stubs:
+            # git ls-files |
```
AlexWaygood, line 44: This comment is somewhat out of date now; the `daily.yml` workflow is a much faster way of accomplishing the same thing

```suggestion
            # Use the daily.yml workflow to run stubtest on all third party stubs
```

.github/workflows/stubtest_third_party.yml
```diff
@@ -0,0 +1,70 @@
...
+            if [ "${{ matrix.os }}" = "ubuntu-latest" ]; then
+              if [ -n "$PACKAGES" ]; then
+                echo "Installing apt packages: $PACKAGES"
+                sudo apt update && sudo apt install -y $PACKAGES
+              fi
+              xvfb-run python tests/stubtest_third_party.py $STUBS
+            fi
+
+            if [ "${{ matrix.os }}" = "macos-latest" ]; then
+              # TODO: install brew packages
```
AlexWaygood, line 61:
```suggestion
              # Could install brew packages here if we run into stubs that need it
```

.github/workflows/stubtest_third_party.yml
```diff
@@ -0,0 +1,70 @@
...
+              python tests/stubtest_third_party.py $STUBS
+            fi
+
+            if [ "${{ matrix.os }}" = "windows-latest" ]; then
+              python tests/stubtest_third_party.py $STUBS
```
AlexWaygood, line 66:
```suggestion
              # Could install choco packages here if we run into stubs that need it
              python tests/stubtest_third_party.py $STUBS
```

CONTRIBUTING.md
```diff
@@ -192,6 +192,11 @@ This has the following keys:
 * `apt_dependencies` (default: `[]`): A list of Ubuntu APT packages
   that need to be installed for stubtest to run successfully. These are
   usually packages needed to pip install the implementation distribution.
+* `platforms` (default: `["linux"]`): A list of OSes on which to run stubtest.
+  Can contain `win32`, `linux`, and `darwin` values.
+  If not specified, tests are run only on `linux`.
```
AlexWaygood, line 197:
```suggestion
  If not specified, stubtest is run only on `linux`.
```

AlexWaygood (commented):
A few more tiny nits I spotted while looking over again

stubs/D3DShot/METADATA.toml
```diff
@@ -2,6 +2,6 @@ version = "0.1.*"
 requires = ["types-Pillow"]
 
 [tool.stubtest]
-# The library only works on Windows; we currently only run stubtest on Ubuntu for third-party stubs in CI.
-# See #8660
+# TODO: re-enable
```
sobolevn, line 5:
```suggestion
# TODO: re-enable
```

AlexWaygood: Wait, this TODO comment I _did_ like -- it would be good if we could figure out how to enable stubtest for this package!

stubs/D3DShot/METADATA.toml
```diff
@@ -2,6 +2,5 @@ version = "0.1.*"
 requires = ["types-Pillow"]
 
 [tool.stubtest]
-# The library only works on Windows; we currently only run stubtest on Ubuntu for third-party stubs in CI.
-# See #8660
 skip = true
```
AlexWaygood, line 5:
```suggestion
# TODO: figure out how to run stubtest for this package
# (the package pins Pillow in a problematic way)
skip = true
```

AlexWaygood:
Time for this to go in.

Avasam:
@sobolevn & @AlexWaygood There's an unforeseen issue: Now I can't run stubtest locally on any stub that doesn't explicitely specify my non-linux OS!

If no platform is specified in `METADATA.toml`, it should be **allowed** to be run on any OS, but the CI should only run linux.

sobolevn:
I think we can add `--local-run` flag to be sure that it is indeed runing on a local env.

Avasam:
Proposed solution: https://github.com/python/typeshed/pull/9173

Issue #8922, opened 2022-10-18
`bytes` constructor does not permit `ctypes.Struture`


iujakchu:
```python
from ctypes import Structure

class A (Structure):
    pass


a = A()
b = bytes(a)
```
the snippet works but linter raises.

JelleZijlstra:
Duplicate of #8889 (well, at least the fix is the same).

PR #8924, opened 2022-10-18
bytes, bytearray constructors: support buffers

### LINKED ISSUE #8889: `bytes` constructor does not permit `PickleBuffer`
Currently it appears `PickleBuffer` is not included in `SupportsBytes`.

As a result, linting tools may complain if a user writes a script like the one below.

```python
# script2.py


from pickle import PickleBuffer

b1 = b"abc"
pb = PickleBuffer(b1)
b2 = bytes(pb)         # <- linter raises here
print(b2)
```

Executing the script above `print`s out `b'abc'` as expected, but linters don't expect `PickleBuffer` to be a valid input to `bytes`.

If possible, it would be helpful if `PickleBuffer` was added to `SupportsBytes`.

xref: https://github.com/microsoft/pyright/issues/4043

JelleZijlstra:
Also combine some overloads.

Fixes #8889.


JelleZijlstra:
Also confirmed that both `pickle.PickleBuffer` (from #8889) and `ctypes.Structure` (from #8922) are already included in `ReadableBuffer`.

AlexWaygood:
I assume the overloads are currently the way they are is to mimic the output of `help()` at runtime:

```pycon
C:\Users\alexw\coding>python
Python 3.10.7 (tags/v3.10.7:6cc6b13, Sep  5 2022, 14:08:36) [MSC v.1933 64 bit (AMD64)] on win32
>>> help(bytes)
class bytes(object)
 |  bytes(iterable_of_ints) -> bytes
 |  bytes(string, encoding[, errors]) -> bytes
 |  bytes(bytes_or_buffer) -> immutable copy of bytes_or_buffer
 |  bytes(int) -> bytes object of size given by the parameter initialized with null bytes
 |  bytes() -> empty bytes object
 |
 |  Construct an immutable array of bytes from:
 |    - an iterable yielding integers in range(256)
 |    - a text string encoded using the specified encoding
 |    - any object implementing the buffer API.
 |    - an integer
```

But, I agree with combining them in typeshed!

AlexWaygood (approved):

PR #8929, opened 2022-10-19
[April 2023] `D3DShot` is an archived repo

sobolevn:
Link: https://github.com/SerpentAI/D3DShot

### DIFF
--- a/stubs/D3DShot/METADATA.toml
+++ b/stubs/D3DShot/METADATA.toml
@@ -1,5 +1,6 @@
 version = "0.1.*"
 requires = ["types-Pillow"]
+no_longer_updated = true
 
 [tool.stubtest]
 # The library only works on Windows; we currently only run stubtest on Ubuntu for third-party stubs in CI.


srittau (approved):
Thanks!

AlexWaygood:
This one's a bit of a special case -- we only added these stubs a few weeks ago, and we knew it was archived at the time (https://github.com/python/typeshed/pull/8652#issuecomment-1237727896, and following discussion).

It feels a bit harsh on @Avasam to delete the stubs from typeshed only a month after he put a lot of work into adding them!

srittau:
Maybe we should delete them after six months if there haven't been major updates to the stubs since then? I'd expect possible kinks having been worked out by then, so the last published version on PyPi of the types should be in a good state by then.

AlexWaygood:
> Maybe we should delete them after six months if there haven't been major updates to the stubs since then? I'd expect possible kinks having been worked out by then, so the last published version on PyPi of the types should be in a good state by then.

Yeah, that sounds like a good plan to me!

PR #8932, opened 2022-10-19
`psutil`: Fix `win_service_iter()` return type to allow iterating over its result.

happy-monk:


JelleZijlstra (approved):
Confirmed at https://github.com/giampaolo/psutil/blob/669b672dc7acc696376e5ca8104cf01086fe6f8c/psutil/_pswindows.py#L475

PR #8937, opened 2022-10-20
Bump the rev of protobuf supported to 4.21.8

nipunn1313:
Bumping rev of mypy-protobuf incoming next

This change is fairly minor - just some comment changes in the generated stubs.

### DIFF
--- a/stubs/protobuf/METADATA.toml
+++ b/stubs/protobuf/METADATA.toml
@@ -1,2 +1,2 @@
-version = "3.20.*"
+version = "4.21.*"
 extra_description = "Generated with aid from mypy-protobuf v3.3.0"
--- a/stubs/protobuf/METADATA.toml
+++ b/stubs/protobuf/METADATA.toml
@@ -1,2 +1,2 @@
-version = "3.21.*"
+version = "3.20.*"
 extra_description = "Generated with aid from mypy-protobuf v3.3.0"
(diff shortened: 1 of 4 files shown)


AlexWaygood:
We were wondering about removing `google/__init__.pyi` in our next major version bump for our protobuf stubs. Would you be interested in making that change as part of this PR? And if so -- would you be okay if we held off merging this until mypy 0.990 is released (should be in the next few weeks)? `--namespace-packages` will be the default in mypy 0.990, so it will make removing `google/__init__.pyi` much less painful for users (but we'd still rather make the change along with a major version bump, just to be sure).

nipunn1313:
Sure that works.

One idea could be to keep this at `3.21.8` until then - just to unblock/decouple this work from the `__init__` work - since this piece is not actually a breaking change.

stubs/protobuf/METADATA.toml
```diff
@@ -1,2 +1,2 @@
-version = "3.20.*"
+version = "4.21.*"
```
nipunn1313, line 1:
```suggestion
version = "3.21.*"
```

Sort of a holdover in the meantime

nipunn1313: (if you like the idea, feel free to commit it and merge - otherwise we can just wait.)

AlexWaygood: Sure, let's do it!

stubs/protobuf/METADATA.toml
```diff
@@ -1,2 +1,2 @@
-version = "3.20.*"
+version = "3.21.*"
```
AlexWaygood, line 1:
```suggestion
version = "3.20.*"
```

PR #8917, opened 2022-10-18
Add stubs for python-cronlog

kasium:

PR #8938, opened 2022-10-20
Mark `python-crontab` as completed

sobolevn:
It was just added in https://github.com/python/typeshed/pull/8917

### DIFF
--- a/stubs/python-crontab/@tests/stubtest_allowlist.txt
+++ b/stubs/python-crontab/@tests/stubtest_allowlist.txt
@@ -1,0 +1,2 @@
+# Re-export:
+crontabs.X_OK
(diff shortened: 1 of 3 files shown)


sobolevn:
I will try to check for possible stubtest changes to fix constants re-export.

AlexWaygood (approved):
Nice

AlexWaygood:
> I will try to check for possible stubtest changes to fix constants re-export.

You could potentially try to get the source code with `inspect.getsource()`, and then do AST analysis to examine which symbols are exported from other modules.

AlexWaygood:
Give me a ping if you want a hand/second look on anything :D

PR #8939, opened 2022-10-20
Remove `dj-database-url` from `pyright`s exclude

sobolevn:


sobolevn:
I will try to check for possible `stubtest` changes to fix constants re-export.

hauntsaninja (approved):
Nice!

PR #8944, opened 2022-10-20
Remove `Send2Trash` from `pyright`'s exclude

sobolevn:


stubs/Send2Trash/send2trash/util.pyi
```diff
@@ -1 +1,3 @@
-def preprocess_paths(paths): ...
+from _typeshed import Incomplete
+
+def preprocess_paths(paths: Incomplete) -> Incomplete: ...
```
JelleZijlstra, line 3:
```suggestion
def preprocess_paths(paths: StrOrBytesPath | Sequence[StrOrBytesPath]) -> list[str | bytes]: ...
```

stubs/Send2Trash/send2trash/util.pyi
```diff
@@ -1 +1,4 @@
-def preprocess_paths(paths): ...
+from _typeshed import StrOrBytesPath
+from collections.abc import Sequence
+
+def preprocess_paths(paths: StrOrBytesPath | Sequence[StrOrBytesPath]) -> list[str | bytes]: ...
```
Akuli, line 4: It checks for `list` with isinstance(), so we need `list[StrOrBytesPath]`. Except that it doesn't work because of invariance, so we need `list[Any]`. I think it's more important to guard against passing a non-`list` sequence than passing a list with the wrong element type.

In any case, this should be made consistent with `__init__.pyi`.

JelleZijlstra: I went with Sequence to avoid the invariance, but don't have a strong opinion on that vs. `list[Any]`.

sobolevn:
Done!

Akuli (approved):

sobolevn:
Friendly ping to get this merged :)

PR #8945, opened 2022-10-20
Remove `simplejson` from `pyright`'s exclude

sobolevn:


stubs/simplejson/simplejson/decoder.pyi
```diff
@@ -3,5 +3,5 @@ from typing import Any
 
 class JSONDecoder:
     def __init__(self, **kwargs: Any) -> None: ...
-    def decode(self, s: str, _w: Match[str], _PY3: bool): ...
-    def raw_decode(self, s: str, idx: int, _w: Match[str], _PY3: bool): ...
+    def decode(self, s: str, _w: Match[str], _PY3: bool) -> Any: ...
+    def raw_decode(self, s: str, idx: int, _w: Match[str], _PY3: bool) -> tuple[bool | None, int]: ...
```
AlexWaygood, line 7: The first element of the tuple can also be a `float`: https://github.com/simplejson/simplejson/blob/43645252224d4e529c22cb78ac42f241d6426073/simplejson/scanner.py#L64

```suggestion
    def raw_decode(self, s: str, idx: int, _w: Match[str], _PY3: bool) -> tuple[Any, int]: ...
```

stubs/simplejson/simplejson/encoder.pyi
```diff
@@ -1,9 +1,9 @@
-from typing import Any
+from typing import Any, Iterator, NoReturn
```
AlexWaygood, line 1:
```suggestion
from collections.abc import Iterator
from typing import Any, NoReturn
```

AlexWaygood (approved):
Thanks!

PR #8946, opened 2022-10-20
Remove `tree-sitter` from `pyright`'s exclude

sobolevn:
Source: https://github.com/tree-sitter/py-tree-sitter


stubs/tree-sitter/tree_sitter/__init__.pyi
```diff
@@ -13,5 +13,5 @@ class Language:
     language_id: int
     # library_path is passed into ctypes LoadLibrary
     def __init__(self, library_path: str, name: str) -> None: ...
-    def field_id_for_name(self, name): ...
-    def query(self, source): ...
+    def field_id_for_name(self, name: str) -> int | None: ...
+    def query(self, source: str) -> Incomplete: ...
```
JelleZijlstra, line 17: It returns a `tree_sitter.binding.Query` object from https://github.com/tree-sitter/py-tree-sitter/blob/master/tree_sitter/binding.c#L1389

sobolevn:
Looks like we need help from @Akuli :)

sobolevn:
I don't want to fix this signature completely right now, it will take quite some time for me to understand what is going on there (I don't know anything about this package).

So, I believe that `node` and `TODO` is enought.

stubs/tree-sitter/tree_sitter/binding.pyi
```diff
@@ -72,7 +72,9 @@ class Parser:
 @final
 class Query:
     # start_point and end_point arguments don't seem to do anything
-    def captures(self) -> list[tuple[Node, str]]: ...
+    # TODO: sync with
+    # https://github.com/tree-sitter/py-tree-sitter/blob/d3016edac2c33ce647653d896fbfb435ac2a6245/tree_sitter/binding.c#L1304
+    def captures(self, node: Node) -> list[tuple[Node, str]]: ...
```
Akuli, line 77: Spent some time figuring out how the code you linked works:
```suggestion
    def captures(self, node: Node, start_point: tuple[int, int] = ..., end_point: tuple[int, int] = ..., start_byte: int = ..., end_byte: int = ...) -> list[tuple[Node, str]]: ...
```

Akuli (approved):
I added the types for remaining arguments of the `captures()` method, but then saw my comment about how the arguments seem to be broken and reverted the additions. I might try the arguments again later. If I find that they work, I will make a PR that adds them back.

sobolevn:
Fixed conflicts

PR #8947, opened 2022-10-20
Remove `tzlocal` from `pyright`'s exclude

sobolevn:
Source: https://github.com/regebro/tzlocal/blob/master/tzlocal/utils.py


hauntsaninja:
Thanks, closing in favour of #8969

PR #8948, opened 2022-10-20
Remove some packages from `pyright`'s exclude: they are correct

sobolevn:
Sorry for the noise lately! 
This is the last PR, I promise :)


AlexWaygood (approved):
🥳

JelleZijlstra:
Thanks @sobolevn for all the work improving our packages!

srittau:
@sobolevn Even if I'm currently time constrained, so the burden to review those isn't on me: This is the good kind of noise!

PR #8950, opened 2022-10-20
Fix `invoke` regression

### LINKED ISSUE #8936: Invoke stubs fail with latest changes
Hi,

with the latest changes added to the invoke-stubs, mypy fails

```py
from invoke import task, Context

@task
def docker_build(context: Context) -> None:
    pass

@task(docker_build)
def docker_push(context: Context) -> None:
    pass
```

```
foo.py:9:2: error: Argument 1 to "__call__" of "Task" has incompatible type "Callable[[Context], None]"; expected "Context"  [arg-type]
    @task(docker_build)
```

As soon as I add any argument to `@task(docker_build)` like `@task(docker_build, name="fooo")` it works. I guess mypy fails to distinguish task overload 1 and 2. Do you have any idea how to solve this?

AlexWaygood:
Refs #8936.

This implements @Akuli's suggested idea in https://github.com/python/typeshed/issues/8936#issuecomment-1285748131. I've added a test case (@kasium's snippet in https://github.com/python/typeshed/issues/8936#issue-1416013788), which fails without this change.

@kasium, are you able to do any additional testing of this PR, to check that this doesn't have any unintended side effects?

### DIFF
--- a/stubs/invoke/@tests/test_cases/check_task.py
+++ b/stubs/invoke/@tests/test_cases/check_task.py
@@ -6,6 +6,7 @@ from invoke import Context, task
 # This snippet is a regression test for #8936
 # ===========================================
 
+
 @task
 def docker_build(context: Context) -> None:
     pass
--- a/stubs/invoke/@tests/test_cases/check_task.py
+++ b/stubs/invoke/@tests/test_cases/check_task.py
@@ -1,0 +1,11 @@
+# pyright: reportUnnecessaryTypeIgnoreComment=true
+
+from invoke import task, Context
+
+@task
+def docker_build(context: Context) -> None:
+    pass
+
+@task(docker_build)
+def docker_push(context: Context) -> None:
+    pass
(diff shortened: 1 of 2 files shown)


stubs/invoke/@tests/test_cases/check_task.py
```diff
@@ -0,0 +1,13 @@
+# pyright: reportUnnecessaryTypeIgnoreComment=true
+
+from invoke import Context, task
+
```
AlexWaygood, line 4:
```suggestion

# ===========================================
# This snippet is a regression test for #8936
# ===========================================
```

JelleZijlstra (approved):

kasium:
@AlexWaygood thanks for the fix. Looks fine to me and I couldn't find an issue with it. However, the issue is still not close, since I found another regression which is not fixed by this change (see the issue for details)

AlexWaygood:
> @AlexWaygood thanks for the fix. Looks fine to me and I couldn't find an issue with it. However, the issue is still not close, since I found another regression which is not fixed by this change (see the issue for details)

Okay, I'll merge this but leave the issue open.

PR #8951, opened 2022-10-21
Bump mypy-protobuf to 3.4.0

nipunn1313:
Use latest version of mypy-protobuf.


stubs/protobuf/METADATA.toml
```diff
@@ -1,2 +1,2 @@
 version = "3.20.*"
```
nipunn1313, line 1: this wants to update to `4.21.*` but we're waiting a bit for a new version of mypy to release

hauntsaninja (approved):
Thank you for this, and all of your work on these stubs!

PR #8953, opened 2022-10-21
`ValuesView` is a `Collection`, not `Iterable`

sobolevn:
See https://github.com/python/cpython/issues/98500 and https://github.com/python/cpython/blob/49f61068f49747164988ffc5a442d2a63874fc17/Lib/_collections_abc.py#L893


srittau (approved):
No primer warnings gives me confidence.

PR #8958, opened 2022-10-21
Remove parts of `tkinter` from `pyright` exclude

sobolevn:


Akuli (approved):
I actually started doing this for a few other files. I think I still have a branch about it on my other computer :)

PR #8963, opened 2022-10-21
Improve flask-cors stubs

kasium:


kasium:
Mhm, I thought it's now possible to include 3rd party packages

AlexWaygood:
> Mhm, I thought it's now possible to include 3rd party packages

We implemented the necessary changes in the stub-uploader repo, but there are some further changes required to typeshed's CI before this will be possible (as you can see).

kasium:
Thanks for the update. Let me close the PR for now

AlexWaygood:
#5768 is the issue to track this :)

PR #8965, opened 2022-10-21
Add a couple of missing type hints in pyyaml.constructor

gruebel:
Added some missing type hints in `pyyaml`, but there are still many left.


gruebel:
@srittau or @JelleZijlstra  would be great if one of you could take a look :)

srittau (approved):
I didn't verify against the implementation,  but the types look reasonable.

PR #8967, opened 2022-10-22
Add stubs for flask-migrate

kasium:
As discussed in https://github.com/miguelgrinberg/Flask-Migrate/issues/492 types should be added to typeshed.


kasium:
So stubstest fails since the cli functions are not functions but click.Command or click.Group. click has inline types so we cannot use it here as a dependency right now. Any clue how to solve this?

```
error: flask_migrate.cli.branches is not a function
Stub: at line [35](https://github.com/python/typeshed/actions/runs/3302880777/jobs/5450175814#step:5:36) in file /home/runner/work/typeshed/typeshed/stubs/Flask-Migrate/flask_migrate/cli.pyi
def (directory: Union[builtins.str, None], verbose: builtins.bool)
Runtime:
<Command branches>
```

kasium:
Anybody any idea?

JelleZijlstra:
Should we even include the cli functions in the stub? I imagine they're not supposed to be called directly.

AlexWaygood:
> Should we even include the cli functions in the stub? I imagine they're not supposed to be called directly.

I agree; we can probably just omit that file for now. Add an allowlist entry if stubtest complains.

kasium:
@JelleZijlstra @AlexWaygood thanks for the support

stubs/Flask-Migrate/flask_migrate/__init__.pyi
```diff
@@ -0,0 +1,105 @@
...
+
+class _MigrateConfig:
+    migrate: Migrate
+    db: SQLAlchemy | None
+    directory: str
+    configure_args: dict[str, Any]
+    def __init__(self, migrate: Migrate, db: SQLAlchemy | None, **kwargs) -> None: ...
+    @property
+    def metadata(self) -> MetaData: ...
+
```
AlexWaygood, line 24: We usually try to avoid including private things in the stubs, unless there's a specific reason to make an exception. Is there a specific reason why this needs to be included?

```suggestion
```

stubs/Flask-Migrate/flask_migrate/__init__.pyi
```diff
@@ -0,0 +1,105 @@
...
+class Config:  # should inherit from alembic.config.Config which is not possible yet
+    template_directory: str | None
+    def __init__(self, *args, **kwargs) -> None: ...
+    def get_template_directory(self) -> str: ...
+
+class Migrate:
+    configure_callbacks: list[Callable[[Config], None]]
```
AlexWaygood, line 31: I'm not sure it's true that "configure callbacks" have to return `None`. That would make these functions a little odd, which return whatever the callback functions return:

- https://github.com/miguelgrinberg/Flask-Migrate/blob/7cb4236327ea04fc6be8a17bbfadae6de7bfbc8b/src/flask_migrate/__init__.py#L79-L82
- https://github.com/miguelgrinberg/Flask-Migrate/blob/7cb4236327ea04fc6be8a17bbfadae6de7bfbc8b/src/flask_migrate/__init__.py#L104

I'm not sure what they _should_ return, so (unless you have a better idea than me), maybe we should go with `Callable[[Config], Incomplete]` for now.

Also: It would be good to define a `_ConfigureCallback: TypeAlias = Callable[[Config], Incomplete]` alias somewhere, so that the type isn't repeated numerous times across the file.

stubs/Flask-Migrate/flask_migrate/__init__.pyi
```diff
@@ -0,0 +1,105 @@
...
+        db: SQLAlchemy | None = ...,
+        directory: str | None = ...,
+        command: str | None = ...,
+        compare_type: bool | None = ...,
+        render_as_batch: bool | None = ...,
+        **kwargs,
+    ) -> None: ...
+    def configure(self, f: Callable[[Config], None]) -> Callable[[Config], None]: ...
+    def call_configure_callbacks(self, config: Config): ...
+    def get_config(self, directory: str | None = ..., x_arg: tuple[str] | None = ..., opts: list[str] | None = ...): ...
```
AlexWaygood, line 57: Looks to me like `x_arg` can be a number of different things: https://github.com/miguelgrinberg/Flask-Migrate/blob/7cb4236327ea04fc6be8a17bbfadae6de7bfbc8b/src/flask_migrate/__init__.py#L95-L103

And I think `opts` can be any `Iterable`, it doesn't have to be a `list`: https://github.com/miguelgrinberg/Flask-Migrate/blob/7cb4236327ea04fc6be8a17bbfadae6de7bfbc8b/src/flask_migrate/__init__.py#L92-L93

```suggestion
    def get_config(self, directory: str | None = ..., x_arg: str | Sequence[str] | None = ..., opts: Iterable[str] | None = ...): ...
```

stubs/Flask-Migrate/flask_migrate/__init__.pyi
```diff
@@ -0,0 +1,105 @@
...
+def migrate(
+    directory: str | None = ...,
+    message: str | None = ...,
+    sql: bool = ...,
+    head: str = ...,
+    splice: bool = ...,
+    branch_label: str | None = ...,
+    version_path: str | None = ...,
+    rev_id: str | None = ...,
+    x_arg: tuple[str] | None = ...,
```
AlexWaygood, line 82:
```suggestion
    x_arg: str | Sequence[str] | None = ...,
```

stubs/Flask-Migrate/flask_migrate/__init__.pyi
```diff
@@ -0,0 +1,105 @@
...
+def edit(directory: str | None = ..., revision: str = ...) -> None: ...
+def merge(
+    directory: str | None = ...,
+    revisions: str = ...,
+    message: str | None = ...,
+    branch_label: str | None = ...,
+    rev_id: str | None = ...,
+) -> None: ...
+def upgrade(
+    directory: str | None = ..., revision: str = ..., sql: bool = ..., tag: str | None = ..., x_arg: tuple[str] | None = ...
```
AlexWaygood, line 93:
```suggestion
    directory: str | None = ..., revision: str = ..., sql: bool = ..., tag: str | None = ..., x_arg: str | Sequence[str] | None = ...
```

stubs/Flask-Migrate/flask_migrate/__init__.pyi
```diff
@@ -0,0 +1,105 @@
...
+) -> None: ...
+def downgrade(
+    directory: str | None = ..., revision: str = ..., sql: bool = ..., tag: str | None = ..., x_arg: tuple[str] | None = ...
```
AlexWaygood, line 96:
```suggestion
    directory: str | None = ..., revision: str = ..., sql: bool = ..., tag: str | None = ..., x_arg: str | Sequence[str] | None = ...
```

AlexWaygood (commented):
Mostly looks good -- a few comments below

stubs/Flask-Migrate/flask_migrate/__init__.pyi
```diff
@@ -0,0 +1,105 @@
+from collections.abc import Callable
```
AlexWaygood, line 1:
```suggestion
from collections.abc import Callable, Iterable, Sequence
```

AlexWaygood (approved):

PR #8972, opened 2022-10-23
`datetime.timedelta`: Remove explicit inheritance from `SupportsAbs`

### LINKED ISSUE #8971: Possible bug in definition of datetime.timedelta
```
from datetime import timedelta

class Too(timedelta):
    def __abs__(self) -> Too:
        return Too()

t = Too()
t1 = abs(t)
t2 = t.__abs__()
reveal_type(t1)
reveal_type(t2)
```
Here the revealed types (with pyright and mypy) of `t1` and `t2` are `timedelta` and `Too`, respectively, but the type of `t1` should be `Too`.

`pyright` claims this is a bug in typeshed.  See https://github.com/microsoft/pyright/issues/4082#issuecomment-1288206453

AlexWaygood:
Fixes #8971

### DIFF
--- a/stdlib/datetime.pyi
+++ b/stdlib/datetime.pyi
@@ -2,7 +2,7 @@ import sys
 from _typeshed import Self
 from abc import abstractmethod
 from time import struct_time
-from typing import ClassVar, NamedTuple, NoReturn, SupportsAbs, TypeVar, overload
+from typing import ClassVar, NamedTuple, NoReturn, TypeVar, overload
 from typing_extensions import Literal, TypeAlias, final
 
 if sys.version_info >= (3, 11):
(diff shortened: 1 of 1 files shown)


JelleZijlstra (approved):
I checked the implementation and it indeed returns an exact timedelta, not an instance of a subclass.

I was amused by this comment in the `__pos__` implementation:
```
    /* Could optimize this (by returning self) if this isn't a
     * subclass -- but who uses unary + ?  Approximately nobody.
     */
```

AlexWaygood:
> I checked the implementation and it indeed returns an exact timedelta, not an instance of a subclass.

I also verified by playing around in the REPL before filing the PR :)

PR #8973, opened 2022-10-23
PyYAML: Permit `width: float` for pure-Python `dump(...)`

kmurphy4:
To prevent `PyYAML` from wrapping *any* lines, it's possible to pass `width=float("inf")`, but the current type hints don't like that.  This works at runtime:

```python
>>> s = yaml.dump({"foo": "bar" * 1000}, width=float("inf"))
>>> print(s)
foo: barbarbarbarbarbarbarbarbarbarbarbarbarbarbarbarbarbar...
```

but `mypy` says

```python
floatwidth.py:2: error: No overload variant of "dump" matches argument types "Dict[str, str]", "float"
floatwidth.py:2: note: Possible overload variants:
floatwidth.py:2: note:     def dump(data: Any, stream: _WriteStream[Any], Dumper: Any = ..., *, default_style: Optional[str] = ..., default_flow_style: Optional[bool] = ..., canonical: Optional[bool] = ..., indent: Optional[int] = ..., width: Optional[int] = ..., allow_unicode: Optional[bool] = ..., line_break: Optional[str] = ..., encoding: Optional[str] = ..., explicit_start: Optional[bool] = ..., explicit_end: Optional[bool] = ..., version: Optional[Tuple[int, int]] = ..., tags: Optional[Mapping[str, str]] = ..., sort_keys: bool = ...) -> None
floatwidth.py:2: note:     def dump(data: Any, stream: None = ..., Dumper: Any = ..., *, default_style: Optional[str] = ..., default_flow_style: Optional[bool] = ..., canonical: Optional[bool] = ..., indent: Optional[int] = ..., width: Optional[int] = ..., allow_unicode: Optional[bool] = ..., line_break: Optional[str] = ..., encoding: Optional[str] = ..., explicit_start: Optional[bool] = ..., explicit_end: Optional[bool] = ..., version: Optional[Tuple[int, int]] = ..., tags: Optional[Mapping[str, str]] = ..., sort_keys: bool = ...) -> Any
```

Poking through the `PyYAML` source, it looks the `width` parameter could probably be anything "comparable", as [it's only compared via the `<` operator](https://github.com/yaml/pyyaml/blob/8cdff2c80573b8be8e8ad28929264a913a63aa33/lib/yaml/emitter.py#L89).
    
For the LibYAML implementation, however, we have to use `int`s:

```python    
>>> stream = io.StringIO()
>>> dumper = yaml.CDumper(stream, width=float("inf"))
  File "/usr/lib/python3/dist-packages/yaml/cyaml.py", line 81, in __init__
    version=version, tags=tags)
  File "ext/_yaml.pyx", line 973, in _yaml.CEmitter.__init__ (ext/_yaml.c:14797)
OverflowError: cannot convert float infinity to integer
```
    
I'm not familiar with Cython, though I assume this is because of [the definition of `yaml_emitter_set_width`](https://github.com/yaml/pyyaml/blob/8cdff2c80573b8be8e8ad28929264a913a63aa33/yaml/_yaml.pxd#L252).


Avasam:
Ref https://github.com/python/typing/issues/1160

srittau (dismissed):
I'm sorry for the late response. The change looks good functionally, but i would prefer if we would define a type alias `_Inf = float` or similar (with an explanatory comment) and then use `int | _Inf | None`. While functionally identically, it makes the intent clearer and would make it easier to potentially change it to `Literal['inf']` (or whatever) in the future.

AlexWaygood:
flake8-pyi is pretty dumb about these things — it'll detect `int | float` and yell at you, but if you do `_Inf: TypeAlias = float`, it'll be absolutely cool with you using `int | _Inf` :)

stubs/PyYAML/yaml/__init__.pyi
```diff
@@ -7,6 +7,7 @@ from . import resolver as resolver  # Help mypy a bit; this is implied by loader
 from .constructor import BaseConstructor
 from .cyaml import *
 from .dumper import *
+from .dumper import _Inf
```
kmurphy4, line 10: I'm not sure what the best place to define the alias is.. this feels a little awkward.

AlexWaygood: I think this is fine!

AlexWaygood (approved):
Thanks!

PR #8974, opened 2022-10-24
Initial tensorflow stubs

hmc-cs-mdrissi:
# Summary
Initial tensorflow stubs to start with. These are very incomplete given that tensorflow's api is very large. I'm reviving this now as supporting [non-types dependencies](https://github.com/python/typeshed/issues/5768) has made progress and is a requirement for these stubs as they strongly rely on numpy. Similar to opencv stubs this should not merge until non type dependencies work in CI.

These are all manually made based on tensorflow documentation, testing in repl, and usage in large codebase reliant on tensorflow. mypy stubgen last I tried, crashes on tensorflow's codebase and pyright works but produces a large number of false positives. Both mypy and pyright share a weakness in that tensorflow relies on a lot of dynamic module magic with imports. It uses a [decorator](https://github.com/tensorflow/tensorflow/blob/359c3cdfc5fabac82b3c70b3b6de2b0a8c16874f/tensorflow/python/framework/type_spec.py#L52-L604)(tf_export) to re-export class in a different module sometimes even though module may not exist by itself. A secondary issue is most tensorflow classes are defined in internal locations without any private prefixing, and re-exported elsewhere that's documented. And some core tensorflow classes rely on dynamic patching of methods. As an example [tf.Tensor](https://github.com/tensorflow/tensorflow/blob/359c3cdfc5fabac82b3c70b3b6de2b0a8c16874f/tensorflow/python/framework/ops.py#L293) class one of the most commonly used types in tensorflow many of it's basic methods like `__add__` and `__mul__` are dynamically patched on.

I consider this ready for discussion as there are a couple questions I have before moving forward.

1. What would be best approach for review? I currently have roughly ~3k lines of incomplete stubs for tensorflow and this pr adds some of the basic pieces. Would you prefer a large pr with all I currently have or chop it up into dozen smaller prs of a few hundred lines each? I'm using `getattr(name: str) -> Incomplete` pattern to minimize false positives users would experience. If I do make a pr with full stubs I have currently I'll need to be careful of that as my current local stubs often missed that because at work I followed a philosophy of false positives are better then missing an error and fix stub as needed. A full version of what I have would look roughly like this [pr](https://github.com/microsoft/python-type-stubs/pull/220).
2. Is it fine to leave default values for functions in stubs? My current stubs include default values and they are mostly int/str/None. I find them useful for IDE usage, but I'm ok dropping them if that's preferred for consistency with most other typeshed stubs. If including defaults is ok, where should I adjust mypy/pyright/etc settings to not produce an error of `Default values in stub files should be specified as "..."`
3. Tensor related classes probably should be generic over data type and shape. I'm avoiding that as both would add a lot more complexity to stubs. Shapes especially as typevartuple support is in progress and a lot of shape types may require further peps. DTypes are mostly possible now just may add a lot more overloads/verbosity to track. I could make class generic today but mostly use Any for the type arguments? Not sure on recommended practice here.

The files/classes added here are a couple of the most commonly used ones. If small pr approach is good I'll likely continue by filling in more of `tensorflow/__init__.pyi` and `tensorflow/math.pyi` followed by adding in other commonly used files.


hauntsaninja:
You're missing a layer of nesting. It should be `stubs/distribution_name/package_name`. This explains many of the mypy_primer errors you're seeing.

For the questions you ask, my non-authoritative opinions are:
1. I'm mostly agnostic. If there's stuff that you expect to be controversial, I would separate it from stuff that is non-controversial. Note that while typeshed's philosophy is that false negatives are better than false positives, it's not the biggest deal if we have some false positives as we work through your PRs. As long as we do this quickly, it won't hurt users, since there are currently zero users of types-tensorflow.
2. I would prefer omitting default values from stubs.
3. Let's stick with the simple non-generic stuff for now.

gramster:
> 2. I would prefer omitting default values from stubs.

An advantage of including default values is that tools like pyright or pylance will use stubs (if available) for the signatures they show to users when hovering over a function call, and users want to see the default values. I'd like to understand your point of view though as to why you would prefer them removed?

hmc-cs-mdrissi:
I took a look at primer hits.

The jax hit mypy's redefinition rule and they have code like,

```python
try:
  import tensorflow
except:
  tensorflow = None
```

I'm less sure about what streamlit hit means about note ... here. The code they have is,

```python
try:
    import tensorflow as tf
except ImportError:
    pass
```

The remaining errors look to be expected from lack of numpy in environment.

hauntsaninja:
@gramster I'm personally okay in theory with default values in stubs if we restrict to simple literals, but it should be its own discussion — my comment here mostly just reflected the status quo, changing of which would need discussion. I also figured we're not really losing that much work on this PR: simple literal default values should be easy to automatically reintroduce to all non-overload stubs and we'd need to do so for all existing stubs anyway.

hauntsaninja:
Anyway, opened #8988 to discuss default values

AlexWaygood:
@hmc-cs-mdrissi, it's now possible to declare a dependency on `numpy`, if you're still interested in working on this!

hmc-cs-mdrissi:
I am interested. Pyright now passes. mypy primer looks reasonable and is mainly patterns like,

```python
try:
    import tensorflow as tf
except ImportError:
    pass
```

Next on mypy test, one error I'm unsure how to deal with is,

```
/tmp/tmpi6bf00j9/.venv--5516585432835101127/lib/python3.10/site-packages/numpy/__init__.pyi:641: error: Positional-only parameters are only supported in Python 3.8 and greater  [syntax]
```

As tensorflow depends on numpy and numpy already uses positional only syntax. Can I skip mypy 3.7 runs for tensorflow? Does Metadata.toml for typeshed packages have a way to indicate minimum python version?

From the defaults issue [discussion](https://github.com/python/typeshed/issues/8988) sounds like including simple defaults is fine now. flake8 fails but I can make a separate small pr to drop that rule.

The other errors I'll tackle fixing and make sense to me.

hmc-cs-mdrissi:
@rchen152 Any recommendations on pytype error? Looks like it crashes handling numpy.

```
ValueError: Unreplaced NamedType: 'np.ndarray'

The above exception was the direct cause of the following exception:

  File "/opt/hostedtoolcache/Python/3.10.9/x64/lib/python3.10/site-packages/pytype/load_pytd.py", line 588, in finish_and_verify_ast
    self._resolver.verify(mod_ast)
  File "/opt/hostedtoolcache/Python/3.10.9/x64/lib/python3.10/site-packages/pytype/load_pytd.py", line 290, in verify
    raise BadDependencyError(str(e), name) from e
pytype.load_pytd.BadDependencyError: Unreplaced NamedType: 'np.ndarray', referenced from 'tensorflow'

The above exception was the direct cause of the following exception:

  File "/home/runner/work/typeshed/typeshed/./tests/pytype_test.py", line 74, in run_pytype
  File "/opt/hostedtoolcache/Python/3.10.9/x64/lib/python3.10/site-packages/pytype/load_pytd.py", line 602, in finish_and_verify_ast
    raise (
pytype.load_pytd.BadDependencyError: Can't find pyi for 'np', referenced from 'tensorflow'
```

stubs/tensorflow/tensorflow/__init__.pyi
```diff
@@ -0,0 +1,195 @@
...
+
+_FloatDataSequence: TypeAlias = Sequence[float] | Sequence[_FloatDataSequence]
+_StrDataSequence: TypeAlias = Sequence[str] | Sequence[_StrDataSequence]
+_ScalarTensorCompatible: TypeAlias = Tensor | str | float | np.ndarray[Any, Any] | np.number[Any]
+_TensorCompatible: TypeAlias = _ScalarTensorCompatible | Sequence[_TensorCompatible]
+_ShapeLike: TypeAlias = TensorShape | Iterable[_ScalarTensorCompatible | None] | int | Tensor
+_DTypeLike: TypeAlias = DType | str | np.dtype[Any]
+
+class Tensor:
+    def __init__(self, op: Operation, value_index: int, dtype: DType) -> None: ...
```
hmc-cs-mdrissi, line 32: I'm not sure why I need return type for this `__init__` (and a couple others). Without it pyright in CI shows an error message of unknown return type. Locally I can't reproduce that error.

hmc-cs-mdrissi:
I think this is ready for review. CI now almost all passes. Only failures left I'm unsure how to deal with.

1. Mypy 3.7 fails as numpy uses 3.8 syntax (positional arguments)
2. Pytype crashes with numpy usage.

stubs/tensorflow/tensorflow/__init__.pyi
AlexWaygood: We have strict pyright settings in CI that disallow leaving any parameters or returns without type annotations. You can switch off the strict settings by adding `stubs/tensorflow` to the excludelist here:

https://github.com/python/typeshed/blob/5eb6a692f5e52ad374b8ab5434a351a149cf02ce/pyrightconfig.stricter.json#L8

But we prefer it if as many stubs as possible are tested with the stricter settings!

Avasam:
> pytype.load_pytd.BadDependencyError: Can't find pyi for 'np', referenced from 'tensorflow'

I wonder if it's caused by aliasing a top-level import (`import numpy as np`).
`pytype_test` uses the list of external `requires` found in METADATA.toml to feed pytype modules to ignore.
But if pytype sees the module as `np`, and ignoring `numpy`, then it won't match.

Avasam:
> Mypy 3.7 fails as numpy uses 3.8 syntax (positional arguments)

Tensorflow supports 3.7, so you can't really blindly typeguard the entire stub.

Could you try the following?
```python
import sys
from typing import Any

# numpy uses PEP 570 style Positional-Only Parameters. Which breaks mypy in Python 3.7
    import numpy
else:
    numpy: Any

_: numpy.dtype[Any]
```

Avasam:
@AlexWaygood Alternatively you could try pinning numpy to "<=1.21" until mypy fixes this bug or python 3.7 support is dropped.

AlexWaygood:
Okay, I fixed the mypy issue! It required actually running `mypy_test.py` _on_ Python 3.7 in CI (rather than just running it with the `--python-version=3.7` flag). Running it _on_ Python 3.7 means pip's dependency resolver will install an older, 3.7-compatible version of numpy when 3.7 is selected from the version matrix in CI.

Avasam:
> Okay, I fixed the mypy issue! It required actually running `mypy_test.py` _on_ Python 3.7 in CI (rather than just running it with the `--python-version=3.7` flag). Running it _on_ Python 3.7 means pip's dependency resolver will install an older, 3.7-compatible version of numpy when 3.7 is selected from the version matrix in CI.

Or that. Nice :)

AlexWaygood:
> @AlexWaygood Alternatively you could try pinning numpy to "<=1.21" until mypy fixes this bug or python 3.7 support is dropped.

Yup, but upper bounds are very bad and best avoided where possible in Python packaging 🙂 See https://iscinumpy.dev/post/bound-version-constraints/ for the very long version.

Avasam:
> Yup, but upper bounds are (in my opinion) very bad and best avoided where possible in Python packaging 🙂 See [iscinumpy.dev/post/bound-version-constraints](https://iscinumpy.dev/post/bound-version-constraints/) for the very long version.

I edited my previous comment instead. But yeah, I remembered after writing this that `requires` isn't used only for typeshed and the version restriction would propagate to users of `types-tensorflow`

AlexWaygood:
I'll separate out the infrastructure changes into another PR (it was useful to edit this PR directly so I could test whether the changes fixed the problem 🙂)

AlexWaygood:
> I'll separate out the infrastructure changes into another PR (it was useful to edit this PR directly so I could test whether the changes fixed the problem 🙂)

- #9499

stubs/tensorflow/tensorflow/__init__.pyi
hmc-cs-mdrissi: Keeping it to pyright strict sounds good to me.

hmc-cs-mdrissi:
> > pytype.load_pytd.BadDependencyError: Can't find pyi for 'np', referenced from 'tensorflow'
> 
> I wonder if it's caused by aliasing a top-level import (`import numpy as np`). `pytype_test` uses the list of external `requires` found in METADATA.toml to feed pytype modules to ignore. But if pytype sees the module as `np`, and ignoring `numpy`, then it won't match.

Thank you that fix works. Not using alias for numpy fixes the error and pytype now passes. All checks are now good (one mypy 3.8 error in last run is transient error).

AlexWaygood:
(Marking with do-not-merge for now, as I'd like a decision to be made on #9499 first :)

stubs/tensorflow/tensorflow/__init__.pyi
```diff
@@ -0,0 +1,195 @@
...
+    AUTO = 0
+    NONE = 1
+    ON_WRITE = 2
+    ON_READ = 3
+
+class VariableAggregation(Enum):
+    AUTO = 0
+    NONE = 1
+    ON_WRITE = 2
+    ON_READ = 3
```
AlexWaygood, line 89: Generally speaking, we do our best to reproduce the structure of the runtime source code in typeshed's stubs. I don't see that much value in doing so with `tensorflow`, since I doubt you're meant to be importing these enums with `from tensorflow.python.ops.variables import VariableSynchronization`. But maybe we could add a comment, where possible, to indicate where these classes are actually defined at runtime?

stubs/tensorflow/tensorflow/dtypes.pyi
```diff
@@ -0,0 +1,55 @@
...
+uint8: DType = ...
+uint16: DType = ...
+uint32: DType = ...
+uint64: DType = ...
+qint8: DType = ...
+qint16: DType = ...
+qint32: DType = ...
+quint8: DType = ...
+quint16: DType = ...
+string: DType = ...
```
AlexWaygood, line 52: Style nit: we prefer to omit the `= ...`s where they're not necessary

```suggestion
bool: DType
complex128: DType
complex64: DType
bfloat16: DType
float16: DType
half: DType
float32: DType
float64: DType
double: DType
int8: DType
int16: DType
int32: DType
int64: DType
uint8: DType
uint16: DType
uint32: DType
uint64: DType
qint8: DType
qint16: DType
qint32: DType
quint8: DType
quint16: DType
string: DType
```

stubs/tensorflow/tensorflow/__init__.pyi
hmc-cs-mdrissi: I'm hesitant as it's true for almost every single tensorflow class/function. Variable/Tensor are not defined in same file. Math functions are mostly split across files too. Similarly dtypes are not actually defined in that file. My guess is 95%+ of functions/classes the real definition is internal file and then re-exported elsewhere.

I can add a top level docstring to the file describing for where to find it? I usually find runtime class by going to [documentation](https://www.tensorflow.org/api_docs/python/tf/dtypes/DType) and clicking view source. 

A related thing is I'm not currently mimicking the full class inheritance. Mainly as most parent classes are undocumented internal classes. Metaclasses I only added for stubtest.

AlexWaygood: Yeah that's all very fair. I guess this is just something I'll have to mentally adjust to with our TensorFlow stubs — "nothing will be where it seems" :)

> I can add a top level docstring to the file describing for where to find it? I usually find runtime class by going to [documentation](https://www.tensorflow.org/api_docs/python/tf/dtypes/DType) and clicking view source.

Good to know that it's easy to find the source from the docs! In that case, we can probably do without the docstrings in the stub.

.github/workflows/tests.yml
```diff
@@ -94,7 +94,7 @@ jobs:
           python-version: ${{ matrix.python-version }}
           cache: pip
           cache-dependency-path: requirements-tests.txt
-      - run: pip install mypy packaging pathspec termcolor tomli
+      - run: pip install $(grep mypy== requirements-tests.txt) packaging pathspec termcolor tomli typing-extensions
```
Avasam, line 97: Flake8 is the problem right?
Can't you add a python version environment marker to the requirements file?

`flake8>=6 ; python_version>="3.8"

AlexWaygood: Good point! Wanna make a PR?

hmc-cs-mdrissi:
Anything left to merge this? @AlexWaygood After this is merged I'll continue to make tensorflow stub prs of a similar size (few hundred lines) until all the ones I have accumulated are landed.

AlexWaygood:
> Anything left to merge this?

This all looks good to me. I don't really feel very qualified to review stubs outside of my pure-Python comfort zone, but if there are any issues here, I think they can be resolved in follow-up PRs as and when they're discovered. This PR has been open long enough now :)

> After this is merged I'll continue to make tensorflow stub prs of a similar size (few hundred lines) until all the ones I have accumulated are landed.

That sounds like a great strategy, and will make it much easier to review!

AlexWaygood (approved):
Thank you!

PR #8975, opened 2022-10-24
Fix `pkg_resources.split_sections`.

HexDecimal:
This function is currently documented as returning a list of lines as the 2nd pair value.  It does not return plain `str` here.

Also there's a broken link at https://pypi.org/project/types-pkg-resources/

https://github.com/python/typeshed/tree/master/stubs/pkg_resources
should be:
https://github.com/python/typeshed/tree/master/stubs/setuptools/pkg_resources


hauntsaninja:
You shouldn't be using types-pkg-resources; use types-setuptools instead. Maybe we can manually make a release to mark those as obsolete? cc @srittau @JelleZijlstra

hauntsaninja (approved):
Thanks!

HexDecimal:
> You shouldn't be using types-pkg-resources; use types-setuptools instead.

I couldn't tell the difference. I actually have both installed, so they're probably causing conflicts now.

The `types-pkg-resources` package should probably point users to `types-setuptools`.

PR #2658, opened 2018-11-30
Stop inheriting Sized in collection ABCs

gvanrossum:
Instead define abstract __len__ in affected classes.

Fixes #2655 without breaking
https://github.com/rominf/ordered-set-stubs/issues/1

PR #8977, opened 2022-10-24
`Collection` is `Sized`

sobolevn:
See https://github.com/python/cpython/blob/ad1dc3ebb6aadaeeeacde13d4ed2d62bf302bf62/Lib/_collections_abc.py#L431﻿

### DIFF
--- a/stdlib/builtins.pyi
+++ b/stdlib/builtins.pyi
@@ -805,7 +805,7 @@ class bytearray(MutableSequence[int], ByteString):
     def __alloc__(self) -> int: ...
 
 @final
-class memoryview(Sized, Sequence[int]):
+class memoryview(Sequence[int]):
     @property
     def format(self) -> str: ...
     @property
--- a/stdlib/typing.pyi
+++ b/stdlib/typing.pyi
@@ -452,8 +452,7 @@ class Container(Protocol[_T_co]):
     def __contains__(self, __x: object) -> bool: ...
 
 @runtime_checkable
-class Collection(Iterable[_T_co], Container[_T_co], Protocol[_T_co]):
-    # Implement Sized (but don't have it as a base class).
+class Collection(Sized, Iterable[_T_co], Container[_T_co], Protocol[_T_co]):
     @abstractmethod
     def __len__(self) -> int: ...
 
--- a/test_cases/stdlib/typing/check_aliases.py
+++ b/test_cases/stdlib/typing/check_aliases.py
@@ -1,4 +1,5 @@
 from typing import Collection
 from typing_extensions import assert_type
 
+# See https://github.com/python/typeshed/pull/8977
 assert_type(['a', {'b'}], list[Collection[str]])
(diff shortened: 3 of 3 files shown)


hauntsaninja (commented):
Hm, the decision not to implement Sized in Collection seems pretty intentional.

Looks like this goes back to https://github.com/python/typeshed/pull/2658 , and was maybe done for Python 2 reasons?

srittau (approved):
Considering that primer does not complain and the change back in 2018 alludes to Python 2/3 problems or problems that may be fixed in mypy 0.650,  this looks good to me. But maybe @gvanrossum has some recollection about this?

sobolevn:
Or @JukkaL :)

JukkaL:
This change seems fine, since the original motivation was Python 2/3 compatibility concerns, and the mypy primer output is clean.

sobolevn:
I think that it is a good idea to copy original samples to our test cases.
Please, wait a moment :)

sobolevn:
pyright has a different opinion:    

```
/home/runner/work/typeshed/typeshed/test_cases/stdlib/typing/check_aliases.py:4:13 - error: "assert_type" mismatch: expected "list[Collection[str]]" but received "list[str | set[str]]"
```

Ok, a test case won't quite work.

But, here's my manual test case for this:

```
(.venv) ~/Desktop/typeshed  collection-is-sized ✔                                    
» mypy -c "reveal_type(['a', ['b']])" --custom-typeshed-dir .
<string>:1: note: Revealed type is "builtins.list[typing.Sequence[builtins.str]]"
```

hauntsaninja:
Looks like this change breaks mypy: https://github.com/python/mypy/pull/13977

Reproduce with e.g. `mypy -c 'reveal_type(5)' --no-silence-site-packages --custom-typeshed-dir ~/dev/typeshed`

The two quickest fixes are either:
```
diff --git a/stdlib/typing.pyi b/stdlib/typing.pyi
index d03a3eab..c9cfe304 100644
--- a/stdlib/typing.pyi
+++ b/stdlib/typing.pyi
@@ -325,7 +325,7 @@ class SupportsRound(Protocol[_T_co]):
     def __round__(self, __ndigits: int) -> _T_co: ...
 
 @runtime_checkable
-class Sized(Protocol, metaclass=ABCMeta):
+class Sized(Protocol):
     @abstractmethod
     def __len__(self) -> int: ...
 ```
or:
```
diff --git a/stdlib/typing.pyi b/stdlib/typing.pyi
index d03a3eab..bfb7cfa7 100644
--- a/stdlib/typing.pyi
+++ b/stdlib/typing.pyi
@@ -452,7 +452,9 @@ class Container(Protocol[_T_co]):
     def __contains__(self, __x: object) -> bool: ...
 
 @runtime_checkable
-class Collection(Sized, Iterable[_T_co], Container[_T_co], Protocol[_T_co]): ...
+class Collection(Iterable[_T_co], Container[_T_co], Protocol[_T_co]):
+    @abstractmethod
 
 class Sequence(Collection[_T_co], Reversible[_T_co], Generic[_T_co]):
     @overload
```

Which do people prefer?

JelleZijlstra:
I'd prefer removing the metaclass, I don't think we need it there.

jab:
I think this change caused a false positive error that I'm seeing. More info in  https://github.com/python/typeshed/issues/9296.

PR #8982, opened 2022-10-25
google.protobuf: Add an __init__ stub to fix type checking

BrendanGraham14:
`google.protobuf.descriptor.Descriptor` was missing a type hint for its `__init__` function. 

This causes `typing` to look at its parent, `_NestedDescriptorBase`'s `__init__` function, which requires a different set of arguments and causes failures.

### DIFF
--- a/stubs/protobuf/google/protobuf/descriptor.pyi
+++ b/stubs/protobuf/google/protobuf/descriptor.pyi
@@ -60,6 +60,27 @@ class Descriptor(_NestedDescriptorBase):
     oneofs: Any
     oneofs_by_name: Any
     syntax: Any
+    def __init__(
+        self,
+        name: str,
+        full_name: str,
+        filename: Any,
+        containing_type: Descriptor | None,
+        fields: list[FieldDescriptor],
+        nested_types: list[FieldDescriptor],
+        enum_types: list[EnumDescriptor],
+        extensions: list[FieldDescriptor],
+        options: Any | None = ...,
+        serialized_options: Any | None = ...,
+        is_extendable: bool | None = ...,
+        extension_ranges: Any | None = ...,
+        oneofs: list[OneofDescriptor] | None = ...,
+        file: FileDescriptor | None = ...,
+        serialized_start: Any | None = ...,
+        serialized_end: Any | None = ...,
+        syntax: str | None = ...,
+        create_key: Any | None = ...,
+    ): ...
     def EnumValueName(self, enum, value): ...
     def CopyToProto(self, proto): ...
     def GetOptions(self) -> MessageOptions: ...
@@ -227,6 +248,19 @@ class ServiceDescriptor(_NestedDescriptorBase):
     index: Any
     methods: Any
     methods_by_name: Any
+    def __init__(
+        self,
+        name: str,
+        full_name: str,
+        index: int,
+        methods: list[MethodDescriptor],
+        options: ServiceOptions | None = ...,
+        serialized_options: Any | None = ...,
+        file: FileDescriptor | None = ...,
+        serialized_start: Any | None = ...,
+        serialized_end: Any | None = ...,
+        create_key: Any | None = ...,
+    ): ...
     def FindMethodByName(self, name): ...
     def CopyToProto(self, proto): ...
     def GetOptions(self) -> ServiceOptions: ...
--- a/stubs/protobuf/@tests/stubtest_allowlist.txt
+++ b/stubs/protobuf/@tests/stubtest_allowlist.txt
@@ -27,9 +27,7 @@ google.protobuf.service.Service.GetDescriptor
 # These are deliberately omitted in the stub.
 # The classes can't be constructed directly anyway,
 # so the signatures of their constructors are somewhat irrelevant.
-google.protobuf.descriptor.Descriptor.__init__
 google.protobuf.descriptor.Descriptor.__new__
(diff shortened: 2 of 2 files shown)


JelleZijlstra:
cc @nipunn1313 for protobuf.

stubs/protobuf/google/protobuf/descriptor.pyi
```diff
@@ -60,6 +60,27 @@ class Descriptor(_NestedDescriptorBase):
...
+        serialized_options: Any | None = ...,
+        is_extendable: bool = ...,
+        extension_ranges: Any = ...,
+        oneofs: list[OneofDescriptor] = ...,
+        file: FileDescriptor = ...,
+        serialized_start: Any = ...,
+        serialized_end: Any = ...,
+        syntax: str = ...,
+        create_key: Any = ...,
+    ): ...
```
nipunn1313, line 83: this makes sense. In the future, would appreciate a link to the original source to help double check. I ended up finding it here

https://github.com/protocolbuffers/protobuf/blob/main/python/google/protobuf/descriptor.py#L319

stubs/protobuf/google/protobuf/descriptor.pyi
```diff
@@ -60,6 +60,27 @@ class Descriptor(_NestedDescriptorBase):
...
+        containing_type: Descriptor,
+        fields: list[FieldDescriptor],
+        nested_types: list[FieldDescriptor],
+        enum_types: list[EnumDescriptor],
+        extensions: list[FieldDescriptor],
+        options: Any | None = ...,
+        serialized_options: Any | None = ...,
+        is_extendable: bool = ...,
+        extension_ranges: Any = ...,
+        oneofs: list[OneofDescriptor] = ...,
```
nipunn1313, line 77:
```suggestion
        oneofs: None | list[OneofDescriptor] = ...,
```

stubs/protobuf/google/protobuf/descriptor.pyi
```diff
@@ -60,6 +60,27 @@ class Descriptor(_NestedDescriptorBase):
...
+        file: FileDescriptor = ...,
```
nipunn1313, line 78:
```suggestion
        file: None | FileDescriptor = ...,
```

stubs/protobuf/google/protobuf/descriptor.pyi
```diff
@@ -60,6 +60,27 @@ class Descriptor(_NestedDescriptorBase):
...
+        serialized_start: Any = ...,
+        serialized_end: Any = ...,
+        syntax: str = ...,
```
nipunn1313, line 81:
```suggestion
        syntax: None | str = ...,
```

nipunn1313 (approved):
Appears to be an improvement to before.
Many of the `Any`s could be improved, but not going to block this improvement on that.

@JelleZijlstra - any idea why stubcheck didn't catch this proactively?

stubs/protobuf/google/protobuf/descriptor.pyi
```diff
@@ -60,6 +60,27 @@ class Descriptor(_NestedDescriptorBase):
     oneofs: Any
     oneofs_by_name: Any
     syntax: Any
+    def __init__(
+        self,
+        name: str,
+        full_name: str,
+        filename: Any,
+        containing_type: Descriptor,
```
JelleZijlstra, line 68:
```suggestion
        containing_type: Descriptor | None,
```
I see `containing_type=None` in our internal _pb2 files.

nipunn1313 (changes requested):
Actually - I'd better request changes - because some of the args need to be optional - to match the original source.

hauntsaninja:
@nipunn1313 stubtest did catch this https://github.com/python/typeshed/blob/8ca0394eadaf1f7f75906d8951f66c8a8792dba9/stubs/protobuf/%40tests/stubtest_allowlist.txt#L27

nipunn1313:
aha I see. you're right @hauntsaninja 

@BrendanGraham14 - in what context are you directly instantiating a `Descriptor` and getting this compiler error? I think we have to go one of two routes

1) Leave it out - arguing that it's not really on the "public" API and shouldn't be exposed.
2) Update the stubtest_allowlist and add all the `__init__` methods that are missing.

By chesterton's fence principle, I think the onus is on us now to prove (2) is required, otherwise we should default to (1).

nipunn1313:
If there are legitimate use cases under which we'd want to be instantiating Descriptors directly - then this seems like a good route to go.

JelleZijlstra:
We saw erroors internally from typechecking a generated `_pb2.py` file using pyanalyze. Probably not the most useful thing to do, but the _pb2 is within a package we otherwise ttypecheck.

BrendanGraham14:
@nipunn1313, `Descriptor`s are constructed directly for protocol message types in generated python protobuf code

nipunn1313:
I wouldn't expect the generated `_pb2` files to typecheck in general - as they're full of crazy generated code stuff.

I think you're expected to use tools like protobuf's (new) stub emission (`protoc`'s new arg - ` --pyi_out=OUT_DIR.`), or [mypy-protobuf](https://github.com/nipunn1313/mypy-protobuf)

JelleZijlstra:
> I wouldn't expect the generated `_pb2` files to typecheck in general - as they're full of crazy generated code stuff.

There don't seem to be too many issues other than the signature of Descriptor. I guess if we run into more problems we could just suppress type checking for these files.

PR #8983, opened 2022-10-25
Use `3.11` instead of `3.11-dev`

sobolevn:
It was just released 🎉


JelleZijlstra:
You may have to fix #8984 :)

sobolevn:
```
Run actions/setup-python@v4
Version 3.11 was not found in the local cache
Error: Version 3.11 with arch x64 not found
The list of all available versions can be found here: https://raw.githubusercontent.com/actions/python-versions/main/versions-manifest.json
```

srittau:
Well, it seems macos is lagging behind.

sobolevn:
Yes, we will have to wait for `3.11` on macos for a bit more.

PR #8989, opened 2022-10-26
requests: type RequestException members (not Any)

ErikBjare:
Thought it was weird that `mypy` didn't catch the following:

```py
import requests

try:
    r = requests.get("http://localhost:666")
except requests.RequestException as e:
    # e.response can be None for some subtypes of RequestException (like ConnectionError, Timeout, etc.)
    # basically for all except HTTPError (I think?)
    # Therefore, it crashes with:
    #   AttributeError: 'NoneType' object has no attribute 'status_code'
    if e.response.status_code == 500:
        pass
```

Hopefully this is a reasonable change.


ErikBjare:
Looks like this triggered a bunch of issues in upstream repos, should be possible to make more precise by specifying that HTTPError always has them set (not sure about this assumption, though).

ErikBjare:
If https://github.com/psf/requests/pull/6270 is considered mergeable/correct, that means we can further specify `ConnectionError.request` to not be None, which would get rid of the final complaints from `mypy_primer`.

JelleZijlstra:
Thanks! Let's wait for resolution on the requests PR so that we can merge this without the mypy-primer regression.

intgr:
> Let's wait for resolution on the requests PR

> Oct 26, 2022

image

PR #8996, opened 2022-10-27
lzma: Takes ReadableBuffer, not bytes

JelleZijlstra:
e.g. https://github.com/python/cpython/blob/723ebe76e787cfa6b08cc9587dd679f3234a1025/Modules/_lzmamodule.c#L596


srittau (approved):
Import is missing, but LGTM.

stdlib/lzma.pyi
```diff
@@ -189,7 +189,9 @@ def open(
     newline: str | None = ...,
 ) -> LZMAFile | TextIO: ...
 def compress(
-    data: bytes, format: int = ..., check: int = ..., preset: int | None = ..., filters: _FilterChain | None = ...
+    data: _ReadableBuffer, format: int = ..., check: int = ..., preset: int | None = ..., filters: _FilterChain | None = ...
+) -> bytes: ...
+def decompress(
+    data: _ReadableBuffer, format: int = ..., memlimit: int | None = ..., filters: _FilterChain | None = ...
```
JelleZijlstra, line 195:
```suggestion
    data: ReadableBuffer, format: int = ..., check: int = ..., preset: int | None = ..., filters: _FilterChain | None = ...
) -> bytes: ...
def decompress(
    data: ReadableBuffer, format: int = ..., memlimit: int | None = ..., filters: _FilterChain | None = ...
```

PR #9002, opened 2022-10-27
ast: parse takes a ReadableBuffer

JelleZijlstra:
parse() is basically just a wrapper around compile(), which we
already have as taking ReadableBuffer.


hauntsaninja (approved):
Great!

PR #9003, opened 2022-10-27
_codecs: Fix buffer parameters

JelleZijlstra:


stdlib/_codecs.pyi
```diff
@@ -64,66 +65,66 @@ def decode(
...
 def charmap_build(__map: str) -> _CharMap: ...
-def ascii_decode(__data: bytes, __errors: str | None = ...) -> tuple[str, int]: ...
+def ascii_decode(__data: ReadableBuffer, __errors: str | None = ...) -> tuple[str, int]: ...
 def ascii_encode(__str: str, __errors: str | None = ...) -> tuple[bytes, int]: ...
-def charmap_decode(__data: bytes, __errors: str | None = ..., __mapping: _CharMap | None = ...) -> tuple[str, int]: ...
+def charmap_decode(__data: ReadableBuffer, __errors: str | None = ..., __mapping: _CharMap | None = ...) -> tuple[str, int]: ...
 def charmap_encode(__str: str, __errors: str | None = ..., __mapping: _CharMap | None = ...) -> tuple[bytes, int]: ...
-def escape_decode(__data: str | bytes, __errors: str | None = ...) -> tuple[str, int]: ...
+def escape_decode(__data: str | ReadableBuffer, __errors: str | None = ...) -> tuple[str, int]: ...
 def escape_encode(__data: bytes, __errors: str | None = ...) -> tuple[bytes, int]: ...
```
JelleZijlstra, line 76: This one is really just bytes:

```
>>> _codecs.escape_encode(bytearray(b"x"))
TypeError: escape_encode() argument 1 must be bytes, not bytearray
```

srittau (approved):
I've only found this in the Python docs in regards to whether codecs should be able to handle any buffer:

https://docs.python.org/3/library/codecs.html?highlight=codec#codecs.Codec.decode

> Codec.decode(input, errors='strict')
> [...]
> For text encodings and bytes-to-bytes codecs, input must be a bytes object or one which provides the read-only buffer interface – for example, buffer objects and memory mapped files.

But I guess this kind of implies that generally buffer protocol objects are okay where bytes are allowed.

PR #9005, opened 2022-10-28
CONTRIBUTING: Note no bytearray promotion

### LINKED ISSUE #9001: Assume there is no bytes/bytearray/memoryview implicit promotion
Currently, we treat `bytearray` and `memoryview` as implicit subclasses of `bytes`, following the CPython documentation. In PEP 688, I propose to remove this implicit promotion. Discussion is still ongoing, but I'd like to make sure typeshed is ready for this change by reviewing existing usage of `bytes`.

To make that easier, I propose that we *write stubs as if there is no implicit promotion between `bytes`, `bytearray`, and `memoryview`*. Thus, if a function accepts either `bytes` or `bytearray`, we should write `bytes | bytearray`, not just `bytes`.

The only cost here should be that certain argument annotations become longer; everything should still work fine for type checkers that assume the implicit promotion.

If we agree on this rule, I can review existing `bytes` annotations for correctness. In my experience, most usages of `bytes` in argument positions should be `ReadableBuffer`.

JelleZijlstra:
Closes #9001

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -293,7 +293,7 @@ Typeshed stubs should be written assuming that this proposal
 is accepted, so a parameter that accepts either `bytes` or
 `bytearray` should be typed as `bytes | bytearray`.
 Often one of the aliases from `_typeshed`, such as
-`_typeshed.ReadableBuffer` can be used instead.
+`_typeshed.ReadableBuffer`, can be used instead.
 
 ### What to include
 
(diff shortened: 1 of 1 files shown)


hauntsaninja (approved):
Looks good, worth mentioning the _typeshed buffer types?

CONTRIBUTING.md
```diff
@@ -286,6 +286,15 @@ Two exceptions are `Protocol` and `runtime_checkable`: although
...
 of Python version.
 
+[PEP 688](https://www.python.org/dev/peps/pep-0688/), which is
+currently a draft, removes the implicit promotion of the
+`bytearray` and `memoryview` classes to `bytes`.
+Typeshed stubs should be written assuming that this proposal
+is accepted, so a parameter that accepts either `bytes` or
+`bytearray` should be typed as `bytes | bytearray`.
+Often one of the aliases from `_typeshed`, such as
+`_typeshed.ReadableBuffer` can be used instead.
```
JelleZijlstra, line 296:
```suggestion
`_typeshed.ReadableBuffer`, can be used instead.
```

PR #9008, opened 2022-10-28
_msi: improve types

JelleZijlstra:


stdlib/_msi.pyi
```diff
@@ -12,11 +12,11 @@ if sys.platform == "win32":
...
         __init__: None  # type: ignore[assignment]
-    # Actual typename Summary, not exposed by the implementation
-    class _Summary:
-        def GetProperty(self, propid: int) -> str | bytes | None: ...
+    # Actual typename SummaryInformation, not exposed by the implementation
+    class _SummaryInformation:
+        def GetProperty(self, field: int) -> int | bytes | None: ...
         def GetPropertyCount(self) -> int: ...
-        def SetProperty(self, propid: int, value: str | bytes) -> None: ...
+        def SetProperty(self, field: int, value: int | bytes) -> None: ...
```
srittau, line 19: It uses `PyUnicode_Check`, so this must be a `str`, even if this is asymmetric with `GetProperty`, right? (My Python C API knowledge is a bit rusty.)

https://github.com/python/cpython/blob/4114bcc9ef7595a07196bcecf9c7d6d39f57f64d/PC/_msi.c#L752-L783

```suggestion
        def SetProperty(self, field: int, value: int | str) -> None: ...
```

srittau (changes requested):

srittau (approved):

PR #9009, opened 2022-10-28
_posixsubprocess: improve types

JelleZijlstra:
Based on the code on CPython's 3.11 branch. This is an internal module
and frequently gets new parameters, so keeping it precise and up to
date is probably not high value.


hauntsaninja (approved):
Lgtm

Elno5 (approved):

PR #8997, opened 2022-10-27
ssl, socket, array: Improve bytes handling

JelleZijlstra:

PR #9013, opened 2022-10-28
asyncio: improve bytes handling

JelleZijlstra:
This depends partly on #8997 and should be merged after that PR.

A lot of asyncio code is hard to type because it consists of generic protocols without clear indications about expected types. Where necessary, I went with types from `isinstance()` checks in concrete classes.


stdlib/asyncio/streams.pyi
```diff
@@ -160,9 +160,9 @@ class StreamReader(AsyncIterator[bytes]):
     def set_transport(self, transport: transports.BaseTransport) -> None: ...
     def feed_eof(self) -> None: ...
     def at_eof(self) -> bool: ...
-    def feed_data(self, data: bytes) -> None: ...
+    def feed_data(self, data: Iterable[SupportsIndex]) -> None: ...
     async def readline(self) -> bytes: ...
-    async def readuntil(self, separator: bytes = ...) -> bytes: ...
+    async def readuntil(self, separator: ReadableBuffer = ...) -> bytes: ...
```
srittau, line 165: This code is using `len(separator)`. Is it guaranteed that all buffer protocol objects support `len()`? While buffer objects have fields like `len` and `itemsize`, they refer to the in-memory presentation as far as I can see.

Unfortunately, https://docs.python.org/3/c-api/buffer.html only describes the C side of things, but not how a buffer protocol object looks on the Python side, so it's a bit hard for me to judges buffer uses on the Python-side.

stdlib/asyncio/windows_events.pyi
```diff
@@ -50,10 +50,10 @@ if sys.platform == "win32":
         def recv_into(self, conn: socket.socket, buf: WriteableBuffer, flags: int = ...) -> futures.Future[Any]: ...
         def send(self, conn: socket.socket, buf: WriteableBuffer, flags: int = ...) -> futures.Future[Any]: ...
         def accept(self, listener: socket.socket) -> futures.Future[Any]: ...
-        def connect(self, conn: socket.socket, address: bytes) -> futures.Future[Any]: ...
+        def connect(self, conn: socket.socket, address: tuple[Any, Any] | tuple[Any, Any, Any, Any]) -> futures.Future[Any]: ...
```
srittau, line 53: Use `Incomplete` instead of `Any`? I feel we could tighten this.

srittau (approved):
Looks mostly good to me, but I have issues below where I'm unsure.

stdlib/asyncio/streams.pyi
JelleZijlstra: Good catch, the buffer protocol does not require `len()`. My favorite bare-bones buffer, `pickle.PickleBuffer`, has no len(). I'll just change it to `bytes | bytearray | memoryview` with a comment; after PEP 688 is available we could use a protocol with `__buffer__` and `__len__`.

stdlib/asyncio/windows_events.pyi
JelleZijlstra: Yes, this can definitely be more precise.

srittau (approved):

PR #9015, opened 2022-10-28
bz2: fix bytes parameters

JelleZijlstra:


stdlib/bz2.pyi
```diff
@@ -132,12 +132,12 @@ class BZ2File(BaseStream, IO[bytes]):
...
 class BZ2Compressor:
     def __init__(self, compresslevel: int = ...) -> None: ...
-    def compress(self, __data: bytes) -> bytes: ...
+    def compress(self, __data: ReadableBuffer) -> bytes: ...
     def flush(self) -> bytes: ...
 
 @final
 class BZ2Decompressor:
-    def decompress(self, data: bytes, max_length: int = ...) -> bytes: ...
+    def decompress(self, data: ReadableBuffer, max_length: int = ...) -> bytes: ...
```
srittau, line 140: I was a bit concerned here, as decompress uses `while data:` and I'm not sure if all buffers support that, but it turns out that `data` is overwritten with `BZ2Decompressor.unused_data` at the end of the first loop iteration, which appears to be `bytes`. So if push comes to shove, an empty bytes string is pushed into `BZ2Decompressor.decompress()` once (which has no effect), and then - since `unused_data` is now an empty `bytes`, it will exit the loop.

srittau (approved):

Issue #9006, opened 2022-10-28
Track review of `bytes` types


JelleZijlstra:
I'll use this issue to track reviewing all usage of `bytes` types to see if they should be changed. The below list is from `git grep -c '\bbytes\b'`.

My plan is to do look at each file one by one, reviewing `bytes` occurrences. I'll generally assume occurrences in return types or `IO[bytes]` are correct, but check occurrences in argument types to see if they should be `ReadableBuffer`, `bytes`, or similar.

- [ ] stdlib/_ast.pyi (2 usages)
- [ ] stdlib/_codecs.pyi (48 usages)
- [ ] stdlib/_compression.pyi (2 usages)
- [ ] stdlib/_curses.pyi (19 usages)
- [ ] stdlib/_imp.pyi (1 usages)
- [ ] stdlib/_msi.pyi (2 usages)
- [ ] stdlib/_posixsubprocess.pyi (2 usages)
- [ ] stdlib/_socket.pyi (17 usages)
- [ ] stdlib/_tkinter.pyi (1 usages)
- [ ] stdlib/_typeshed/__init__.pyi (6 usages)
- [ ] stdlib/_typeshed/wsgi.pyi (8 usages)
- [ ] stdlib/_winapi.pyi (6 usages)
- [ ] stdlib/aifc.pyi (15 usages)
- [ ] stdlib/antigravity.pyi (1 usages)
- [ ] stdlib/array.pyi (9 usages)
- [ ] stdlib/ast.pyi (2 usages)
- [ ] stdlib/asynchat.pyi (6 usages)
- [ ] stdlib/asyncio/base_events.pyi (8 usages)
- [ ] stdlib/asyncio/base_subprocess.pyi (3 usages)
- [ ] stdlib/asyncio/events.pyi (8 usages)
- [ ] stdlib/asyncio/exceptions.pyi (2 usages)
- [ ] stdlib/asyncio/protocols.pyi (3 usages)
- [ ] stdlib/asyncio/sslproto.pyi (8 usages)
- [ ] stdlib/asyncio/streams.pyi (11 usages)
- [ ] stdlib/asyncio/subprocess.pyi (5 usages)
- [ ] stdlib/asyncio/transports.pyi (3 usages)
- [ ] stdlib/asyncio/trsock.pyi (17 usages)
- [ ] stdlib/asyncio/windows_events.pyi (4 usages)
- [ ] stdlib/asyncore.pyi (8 usages)
[...]

PR #9016, opened 2022-10-28
Improve types and `bytes` usage of `pathlib`

sobolevn:
Notes:
- `link_to` is removed in 3.12: https://github.com/python/cpython/blob/3.11/Lib/pathlib.py#L1222-L1225
- `samefile` uses the same args as `Path.__new__`

Source: https://github.com/python/cpython/blob/main/Lib/pathlib.py
Refs https://github.com/python/typeshed/issues/9006


stdlib/pathlib.pyi
```diff
@@ -188,16 +189,16 @@ class Path(PurePath):
     def expanduser(self: Self) -> Self: ...
     def read_bytes(self) -> bytes: ...
     def read_text(self, encoding: str | None = ..., errors: str | None = ...) -> str: ...
-    def samefile(self, other_path: str | bytes | int | Path) -> bool: ...
-    def write_bytes(self, data: bytes) -> int: ...
+    def samefile(self, other_path: StrPath) -> bool: ...
```
srittau, line 192: This just calls `other_path.stat()`, which should also work with Path[bytes]. In case it's not a `Path`, it's passed on to `os.stat()`, which uses `int | StrOrBytesPath`. I therefore think the following would be correct:

```suggestion
    def samefile(self, other_path: StrOrBytesPath | int) -> bool: ...
```

stdlib/pathlib.pyi
```diff
@@ -188,16 +189,16 @@ class Path(PurePath):
...
+    def write_bytes(self, data: ReadableBuffer) -> int: ...
     if sys.version_info >= (3, 10):
         def write_text(
             self, data: str, encoding: str | None = ..., errors: str | None = ..., newline: str | None = ...
         ) -> int: ...
     else:
         def write_text(self, data: str, encoding: str | None = ..., errors: str | None = ...) -> int: ...
-    if sys.version_info >= (3, 8):
-        def link_to(self, target: StrPath | bytes) -> None: ...
+    if (3, 8) <= sys.version_info < (3, 12):
```
AlexWaygood, line 200: I don't think type checkers are able to handle complex `sys.version_info` guards like this

```suggestion
    if sys.version_info >= (3, 8) and sys.version_info < (3, 12):
```

stdlib/pathlib.pyi
```diff
@@ -188,16 +189,16 @@ class Path(PurePath):
...
+        def link_to(self, target: StrPath) -> None: ...
```
srittau, line 201: Similarly, this is passed to `os.link()`, so:

```suggestion
        def link_to(self, target: StrOrBytesPath) -> None: ...
```

srittau (changes requested):

stdlib/pathlib.pyi
AlexWaygood: `Path[bytes]` isn't a thing — `Path` isn't a generic class. pathlib only works with string paths, unlike `os.path` or functions in `os`

srittau: I always mix up `Path` and `PathLike`. The annotation should be correct nevertheless.

sobolevn: Here' how this method is implemented:

```python
    def samefile(self, other_path):
        st = self.stat()
        try:
            other_st = other_path.stat()
        except AttributeError:
            other_st = self.__class__(other_path).stat()
        return os.path.samestat(st, other_st)
```

Som `other_path` can either be:
- Something with `.stat()` method (assuming `Path`, it can be a new protocol, but I don't think it is worth it right now)
- Something that `self.__class__.__new__` accepts: `StrPath` (== `str | PathLike[str]`)

But, `Path` is a part of `StrPath`, because `Path` is `PathLike[str]`.
That's why `StrPath` seems correct to me. Am I missing something?

stdlib/pathlib.pyi
sobolevn: Yes, the same as `os.link` 👍

stdlib/pathlib.pyi
srittau: Ah interesting, I was looking at the 3.9 version of the implementation:

https://github.com/python/cpython/blob/b43496c01a554cf41ae654a0379efae18609ad39/Lib/pathlib.py#L1145-L1154

```python
    def samefile(self, other_path):
        """Return whether other_path is the same or not as this file
        (as returned by os.path.samefile()).
        """
        st = self.stat()
        try:
            other_st = other_path.stat()
        except AttributeError:
            other_st = self._accessor.stat(other_path)
        return os.path.samestat(st, other_st)
```

No need to version guard this. Let's just use the latest version (yours).

srittau (approved):

PR #9017, opened 2022-10-28
pyexpat: fix bytes usage

hauntsaninja:


JelleZijlstra (approved):
Thanks, confirmed in the C code.

PR #9019, opened 2022-10-28
typed-ast: fix get_docstring, allow buffer

hauntsaninja:


stubs/typed-ast/typed_ast/ast27.pyi
```diff
@@ -9,11 +10,11 @@ class NodeVisitor:
 class NodeTransformer(NodeVisitor):
     def generic_visit(self, node: AST) -> None: ...
 
-def parse(source: str | bytes, filename: str | bytes = ..., mode: str = ...) -> AST: ...
+def parse(source: str | ReadableBuffer, filename: str | ReadableBuffer = ..., mode: str = ...) -> AST: ...
 def copy_location(new_node: AST, old_node: AST) -> AST: ...
 def dump(node: AST, annotate_fields: bool = ..., include_attributes: bool = ...) -> str: ...
 def fix_missing_locations(node: AST) -> AST: ...
-def get_docstring(node: AST, clean: bool = ...) -> bytes | None: ...
+def get_docstring(node: AST, clean: bool = ...) -> str | bytes | None: ...
```
JelleZijlstra, line 17: I was confused about how this could be bytes, but then I remembered Python 2.

JelleZijlstra (approved):
thanks!

stubs/typed-ast/typed_ast/ast27.pyi
hauntsaninja: `ast27.get_docstring(ast27.parse("""def f(): 'hi'""").body[0])` does crash at runtime (but it works with clean=False)

PR #9021, opened 2022-10-28
imaplib: fix bytes usage

sobolevn:
- `send` calls `_socket.sendall`
- `append` calls `MapCRLF.sub(CRLF, message)`

```python
>>> import imaplib
>>> imaplib.ParseFlags(bytearray())
()
>>> imaplib.Internaldate2tuple(bytearray())
>>> 
```

Source: https://github.com/python/cpython/blob/7ee3aca00a0dffc9cb5906a24005a8c61d5bce14/Lib/imaplib.py
Refs https://github.com/python/typeshed/issues/9006


stdlib/imaplib.pyi
```diff
@@ -157,7 +157,7 @@ class _Authenticator:
     def encode(self, inp: bytes) -> str: ...
     def decode(self, inp: str) -> bytes: ...
 
-def Internaldate2tuple(resp: bytes) -> time.struct_time: ...
+def Internaldate2tuple(resp: ReadableBuffer) -> time.struct_time | None: ...
 def Int2AP(num: int) -> str: ...
```
JelleZijlstra, line 161:
```suggestion
def Int2AP(num: int) -> bytes: ...
```
Not directly related but this one definitely return bytes. We could also change the arg type to `SupportsAbs` I think.

sobolevn: Yes, sure! It is `num = int(abs(num))`: `SupportsAbs[SupportsInt]` 😱

JelleZijlstra: Still got to fix the return type :)

JelleZijlstra (approved):

stdlib/imaplib.pyi
AlexWaygood: > Yes, sure! It is `num = int(abs(num))`: `SupportsAbs[SupportsInt]` 😱

Could be worse:

https://github.com/python/typeshed/blob/b5c4580d529c03efe7b4546ed9d3632befd5fb49/stdlib/unittest/case.pyi#L219

PR #9022, opened 2022-10-28
array: one more bad bytes type

JelleZijlstra:


stdlib/array.pyi
```diff
@@ -56,7 +56,7 @@ class array(MutableSequence[_T], Generic[_T]):
     def tolist(self) -> list[_T]: ...
     def tounicode(self) -> str: ...
     if sys.version_info < (3, 9):
-        def fromstring(self, __buffer: bytes) -> None: ...
+        def fromstring(self, __buffer: str | ReadableBuffer) -> None: ...
```
JelleZijlstra, line 59: proof (on 3.6):

```
>>> array.array("b", []).fromstring("x")
>>> array.array("b", []).fromstring(42)
TypeError: a bytes-like object is required, not 'int'
>>> array.array("b", []).fromstring(b"x")
>>> 
```

hauntsaninja (approved):
Lg

PR #9029, opened 2022-10-29
ctypes: improve bytes handling

JelleZijlstra:


stdlib/ctypes/__init__.pyi
```diff
@@ -194,7 +194,7 @@ class _SimpleCData(Generic[_T], _CData):
 class c_byte(_SimpleCData[int]): ...
 
 class c_char(_SimpleCData[bytes]):
-    def __init__(self, value: int | bytes = ...) -> None: ...
+    def __init__(self, value: int | bytes | bytearray = ...) -> None: ...
```
srittau, line 197: I failed to trace through the code for this, but I trust this is correct.

tests/stubtest_allowlists/py3_common.txt
```diff
@@ -367,6 +367,8 @@ _ctypes.sizeof
 # Allowlist entries that cannot or should not be fixed
 # ==========
 
+ctypes.Array.raw  # exists but stubtest can't see it
```
srittau, line 370: Nit: Maybe copy the comment from the stub?

srittau (approved):

tests/stubtest_allowlists/py3_common.txt
JelleZijlstra, line 370:
```suggestion
ctypes.Array.raw  # exists but stubtest can't see it; only available if _CT == c_char
```

stdlib/ctypes/__init__.pyi
JelleZijlstra: I didn't find the code either but tried it in the REPL. This one allows bytearray but `c_char_p` doesn't.

PR #9030, opened 2022-10-29
dbm: fix bytes handling

JelleZijlstra:


stdlib/dbm/__init__.pyi
```diff
@@ -6,7 +6,7 @@ from typing_extensions import Literal, TypeAlias
 __all__ = ["open", "whichdb", "error"]
 
 _KeyType: TypeAlias = str | bytes
-_ValueType: TypeAlias = str | bytes
+_ValueType: TypeAlias = str | bytes | bytearray
```
srittau, line 9: This alias is only used in `_Database.__setitem__()` as far as I can see (unless the definition is imported elsewhere).

`value` is parsed using `s#`, so I think this should be `ReadOnlyBuffer`:

https://github.com/python/cpython/blob/c1c3be0f9dc414bfae9a5718451ca217751ac687/Modules/_dbmmodule.c#L196

Alternatively, if we want to use the same type for all database types (which is reasonable), I think we should define `_ValueType` centrally and reuse it.

srittau (changes requested):

stdlib/dbm/__init__.pyi
JelleZijlstra: It's a bit more complicated than that. The C code you point to is for `dbm.ndbm`. The class at issue in this file, `dbm._Database`, doesn't exist at runtime. We use it as the return type for `dbm.open`, which can return an instance of any of the database classes in the `dbm` package. I think the intent is that `dbm._Database` reflects the common interface of all three. Maybe we should use a Union return type instead?

I think for now I'll switch it to `str | bytes`, which are the types that all three dbm implementations support. (`dbm.dumb` is `str | bytes | bytearray`, the other two are `str | ReadOnlyBuffer`).

srittau (approved):

PR #9033, opened 2022-10-29
`asyncore` and `asynchat` will be removed in 3.12

sobolevn:
See https://github.com/python/cpython/pull/96580


stdlib/VERSIONS
```diff
@@ -61,7 +61,7 @@ antigravity: 2.7-
 argparse: 2.7-
 array: 2.7-
 ast: 2.7-
-asynchat: 2.7-
+asynchat: 2.7-3.12
```
AlexWaygood, line 64:
```suggestion
asynchat: 2.7-3.11
```

stdlib/VERSIONS
```diff
@@ -72,7 +72,7 @@ asyncio.taskgroups: 3.11-
 asyncio.threads: 3.9-
 asyncio.timeouts: 3.11-
 asyncio.trsock: 3.8-
-asyncore: 2.7-
+asyncore: 2.7-3.12
```
AlexWaygood, line 75:
```suggestion
asyncore: 2.7-3.11
```

AlexWaygood (dismissed):
The last version in the "version ranges" in this file is the "last version when the module was available" :)

AlexWaygood:
I'm also not sure if we should do this before the runtime PR's been merged? But don't have a strong opinion either way on that.

hauntsaninja:
I think stubtest might not complain about entirely missing stdlib modules, so this is best done when betas are released (in case the removal is undone)

AlexWaygood:
> I think stubtest might not complain about missing stubs for stdlib modules

That's correct, it doesn't

sobolevn:
Beta 1 is out! 👍

AlexWaygood (approved):

PR #9034, opened 2022-10-29
asyncore: improve bytes handling

sobolevn:
Source: https://github.com/python/cpython/blob/7ea10567afe38e7770e82b4642c0b01659acaad5/Lib/asyncore.py

`asyncore.send` calls `socket.send`: https://github.com/python/cpython/blob/7ea10567afe38e7770e82b4642c0b01659acaad5/Lib/asyncore.py#L363


AlexWaygood:
I don't really want to spend my time reviewing a PR like this for a module that's long-deprecated and about to be removed :(

sobolevn:
Yeah, I can totally understand, sometimes I feel this too!
(I even have a backlog of things to review) 😆 

But, since typeshed still supports >=3.7, people are and will still use these modules.
This is why I even bothered creating this PR.

Maybe someone else will review it! If you feel like it, closing these two PRs is always an option.

AlexWaygood:
Yeah, to be clear — no objections to other maintainers reviewing and merging this if they want to :)

PR #9035, opened 2022-10-29
binhex: improve bytes handling

sobolevn:
Source: https://github.com/python/cpython/blob/85f88f63d96b07208c98a725391af7cb710fe06b/Lib/binhex.py#L216-L234

Both `write` and `write_rsrc` calls `io.open(..., 'wb').write(data)`


hauntsaninja (changes requested):
Looks like it calls `len`. I'd say we should add a `_BufferWithLen: TypeAlias = ReadableBuffer` to `_typeshed` (or at least add a comment). This would have been useful in at least one other place

hauntsaninja (approved):
Thanks!

AlexWaygood:
> Looks like it calls `len`. I'd say we should add a `_BufferWithLen: TypeAlias = ReadableBuffer` to `_typeshed` (or at least add a comment). This would have been useful in at least one other place

I agree with adding this, but why the leading underscore, out of interest?

hauntsaninja:
Oh, I don't remember exactly what rules we have for `_typeshed`, but I wanted to mark as "private / experimental / do not use"... The type doesn't really correspond to its name, it's currently incorrectly typed (e.g. since `ReadableBuffer` includes `pickle.PickleBuffer`), we don't know the fate of PEP 688, etc.

sobolevn:
> but I wanted to mark as "private / experimental / do not use"...

+1

AlexWaygood:
Cool, that makes sense!

PR #9037, opened 2022-10-29
gzip: improve bytes handling

sobolevn:
`compress` calls `zlib.compress`: https://github.com/python/cpython/blob/7ea10567afe38e7770e82b4642c0b01659acaad5/Lib/gzip.py#L578-L580

`decompress` calls `BytesIO(data)`: https://github.com/python/cpython/blob/7ea10567afe38e7770e82b4642c0b01659acaad5/Lib/gzip.py#L592

I am done with `bytes` for today! 😅


stdlib/gzip.pyi
```diff
@@ -159,9 +159,9 @@ class _GzipReader(_compression.DecompressReader):
     def __init__(self, fp: _ReadableFileobj) -> None: ...
 
 if sys.version_info >= (3, 8):
-    def compress(data: bytes, compresslevel: int = ..., *, mtime: float | None = ...) -> bytes: ...
+    def compress(data: ReadableBuffer, compresslevel: int = ..., *, mtime: float | None = ...) -> bytes: ...
```
JelleZijlstra, line 162: This also does `len(data)` in one branch, so we should use `_BufferWithLen`.

JelleZijlstra (commented):
Thanks for all the bytes PRs!

stdlib/gzip.pyi
sobolevn: Done!

JelleZijlstra (approved):

PR #9040, opened 2022-10-30
pywin32 stubs update

Avasam:
Small update to the pywin32 stubs as I was working on adding basic type-checking support to https://github.com/Kalmat/PyWinCtl

- Validated and updated boolean methods (`def Is...()`)
- Added `win32gui_struct`. A helper module I initially thought was a test helper.
- Updated the following functions: `MonitorFromRect`, `FindWindow`, `FindWindowEx`, `GetWindowPlacement`, `GetMenuItemRect`*

\*`GetMenuItemRect` actually returns a 2-items list (a return code and a Rect). But since lists don't support positional typing, it would make it harder to use:
`return_code, rect = GetMenuItemRect(0, 0, 0)`


stubs/pywin32/pythoncom.pyi
```diff
@@ -63,7 +63,7 @@ def GetRunningObjectTable(reserved: int = ...) -> _win32typing.PyIRunningObjectT
 def GetScodeString(scode) -> str: ...
 def GetScodeRangeString(scode) -> str: ...
 def GetSeverityString(scode) -> str: ...
-def IsGatewayRegistered(iid: _win32typing.PyIID) -> bool: ...
+def IsGatewayRegistered(_iid: _win32typing.PyIID | None) -> int: ...
```
JelleZijlstra, line 66: Did you mean to make the param positional-only? That needs two underscores not one.

JelleZijlstra (approved):

stubs/pywin32/pythoncom.pyi
Avasam: I did! Fixed.
I also think all of the C-extension modules don't support keyword-arguments. But it would be too much of a chore to validate them all.

JelleZijlstra: Thanks!

> all methods in C-extension modules don't support keyword-arguments 

Might be true for pywin32, but it's possible to support kwargs by using `PyArg_ParseTupleAndKeywords`.

JelleZijlstra (approved):
Didn't check all of these against the implementation, but no red flags.

PR #9042, opened 2022-10-30
json: improve bytes handling

sobolevn:
`dumps` and `detect_encoding` support `bytearray`:

```python
>>> import json

>>> json.loads(bytearray(b'false'))
False

>>> json.detect_encoding(bytearray(b''))
'utf-8'
```

But, not `memoryview`:

```python
>>> json.loads(memoryview(b'false'))
  File "/Users/sobolev/.pyenv/versions/3.10.0/lib/python3.10/json/__init__.py", line 339, in loads
    raise TypeError(f'the JSON object must be str, bytes or bytearray, '
TypeError: the JSON object must be str, bytes or bytearray, not memoryview
```

### DIFF
--- a/stdlib/json/__init__.pyi
+++ b/stdlib/json/__init__.pyi
@@ -37,7 +37,7 @@ def dump(
     **kwds: Any,
 ) -> None: ...
 def loads(
-    s: str | bytes,
+    s: str | bytes | bytearray,
     *,
     cls: type[JSONDecoder] | None = ...,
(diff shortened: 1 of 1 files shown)


hauntsaninja (approved):
Lgtm

JelleZijlstra (approved):
`loads` explicitly checks for `str | bytes | bytearray`. `detect_encoding` does `.startswith` which doesn't exist on any other well-known buffer type.

Issue #8956, opened 2022-10-21
"master" branch renamed to "main"


srittau:
I don't want to go into the political reasons for renaming the branch, but in my opinion there is a clear push to standardize on the new name `main`. I believe there was also a SC decision that all projects under the python org should eventually change to that name, although we haven't been approached about this. But CPython has switched (and the stub_uploader always used `main`). Selfishly, I also renamed the branches in most of my projects, with typeshed being one of the last still using `master`.

AlexWaygood:
I support doing this.

srittau:
From what I can see, the following files need to be updated after the transition:

* `.github/workflows/mypy_primer.yml`
* `README.md`
* `scripts/stubsabot.py`

srittau:
Also in the stub_uploader:

* `stub_uploader/build_wheel.py`
* `stub_uploader/get_version.py`

srittau:
I have renamed the branch now. Contributors will need to run the following commands locally to update their local copy:

```
git branch -m master main
git fetch origin
git branch -u origin/main main
git remote set-head origin -a
```

PR #9045, opened 2022-10-31
'master' branch was updated to 'main'

srittau:
Cf. #8956


AlexWaygood (approved):
We can probably delete these lines as well, since I doubt we'll be going back:

https://github.com/python/typeshed/blob/a7cf6144deeb3f99570f8b97fb5f5ecf16f20c01/.github/workflows/stubtest_stdlib.yml#L8

https://github.com/python/typeshed/blob/a7cf6144deeb3f99570f8b97fb5f5ecf16f20c01/.github/workflows/tests.yml#L8

https://github.com/python/typeshed/blob/a7cf6144deeb3f99570f8b97fb5f5ecf16f20c01/.github/workflows/typecheck_typeshed_code.yml#L8

PR #9047, opened 2022-10-31
marshal: improve bytes handling

sobolevn:
Examples:

```python
>>> import marshal
>>> marshal.loads(b'\xe9{\x00\x00\x00')
123
>>> marshal.loads(bytearray(b'\xe9{\x00\x00\x00'))
123
>>> marshal.loads(memoryview(b'\xe9{\x00\x00\x00'))
123
```


hauntsaninja (approved):
Lgtm

stdlib/marshal.pyi
```diff
@@ -1,8 +1,9 @@
+from _typeshed import ReadableBuffer
 from typing import IO, Any
 
 version: int
 
 def dump(__value: Any, __file: IO[Any], __version: int = ...) -> None: ...
 def load(__file: IO[Any]) -> Any: ...
```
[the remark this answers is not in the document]
sobolevn: I am not double-checking, just commiting 😆

stdlib/marshal.pyi
```diff
@@ -1,5 +1,5 @@
 from _typeshed import ReadableBuffer
```
JelleZijlstra:
```suggestion
from _typeshed import ReadableBuffer, SupportsRead, SupportsWrite
```

stdlib/marshal.pyi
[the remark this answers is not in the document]
JelleZijlstra: For the record the relevant code is here https://github.com/python/cpython/blob/main/Python/marshal.c#L1685

stdlib/marshal.pyi
JelleZijlstra: Oh you just did that, thanks! GitHub's UI doesn't handle this well: the "Commit" button still shows up on the suggestion but clicking it does nothing. :)

sobolevn: Already done ;)

JelleZijlstra (approved):

PR #9056, opened 2022-11-01
imaplib: _Authenticator also works with bytearray

JelleZijlstra:


stdlib/imaplib.pyi
```diff
@@ -151,10 +151,10 @@ class IMAP4_stream(IMAP4):
         def open(self, host: str | None = ..., port: int | None = ...) -> None: ...
 
 class _Authenticator:
-    mech: Callable[[bytes], bytes]
-    def __init__(self, mechinst: Callable[[bytes], bytes]) -> None: ...
+    mech: Callable[[bytes], bytes | bytearray | None]
+    def __init__(self, mechinst: Callable[[bytes], bytes | bytearray | None]) -> None: ...
     def process(self, data: str) -> str: ...
-    def encode(self, inp: bytes) -> str: ...
+    def encode(self, inp: bytes | bytearray) -> str: ...
```
hauntsaninja, line 157: I think this is `str | (Buffer & Sequence)`, e.g. don't see anything stopping `memoryview` from working here

stdlib/imaplib.pyi
```diff
@@ -151,10 +151,10 @@ class IMAP4_stream(IMAP4):
...
     def decode(self, inp: str) -> bytes: ...
```
hauntsaninja, line 158: I think this is `str | _BufferWithLen`

stdlib/imaplib.pyi
```diff
@@ -151,10 +151,10 @@ class IMAP4_stream(IMAP4):
         def open(self, host: str | None = ..., port: int | None = ...) -> None: ...
 
 class _Authenticator:
-    mech: Callable[[bytes], bytes]
-    def __init__(self, mechinst: Callable[[bytes], bytes]) -> None: ...
+    mech: Callable[[bytes], bytes | bytearray | None]
```
hauntsaninja, line 154: I think this can return str as well

hauntsaninja (changes requested):

hauntsaninja (approved):
Thanks!

Issue #8998, opened 2022-10-27
Should we use `_ProtocolMeta` in `typing.py`?


sobolevn:
Right now `_ProtocolMeta` is never used:

https://github.com/python/typeshed/blob/739711e94725d7720d5ae923f6ff44c8aeb5f915/stdlib/typing.pyi#L282

But, it is a base metaclass for `Protocol` type in `typing.py`: https://github.com/python/cpython/blame/723ebe76e787cfa6b08cc9587dd679f3234a1025/Lib/typing.py#L1965

It is also used for all `Protocol` types (we currently just use `ABCMeta` for them`): 

https://github.com/python/typeshed/blob/739711e94725d7720d5ae923f6ff44c8aeb5f915/stdlib/typing.pyi#L287-L305

CPython: https://github.com/python/cpython/blame/723ebe76e787cfa6b08cc9587dd679f3234a1025/Lib/typing.py#L2745

So, the question is:
1. Should we use for `Protocol` class? It is rather complex, because right now `Protocol` is `_SpecialForm`
2. Should we use it for `Protocol` types in `typing.pyi`?

JelleZijlstra:
typing.pyi is central to type checking and typecheckers need to heavily special case for the contents of the stub. Therefore, I think we should only make changes here if they fix a concrete problem for users, not out of mere perfectionism.

PR #9058, opened 2022-11-01
typing: remove metaclass from Sized

hauntsaninja:
as per https://github.com/python/typeshed/pull/8977#issuecomment-1297872394

### DIFF
--- a/stdlib/typing.pyi
+++ b/stdlib/typing.pyi
@@ -325,7 +325,7 @@ class SupportsRound(Protocol[_T_co]):
     def __round__(self, __ndigits: int) -> _T_co: ...
 
 @runtime_checkable
-class Sized(Protocol, metaclass=ABCMeta):
+class Sized(Protocol):
     @abstractmethod
     def __len__(self) -> int: ...
 


sobolevn (commented):
Can we please add a regression test? This seems quite important to catch.

hauntsaninja:
The existing mypy test will catch it if you run it with latest mypy. The change in behaviour on mypy side traces back to https://github.com/python/mypy/pull/13579

srittau:
Wouldn't it make more sense to fix this in mypy?

sobolevn:
I agree, this looks like a mypy bug, not a typeshed bug.

JukkaL:
I agree that this looks like a mypy bug, but the metaclass declaration doesn't seem correct so this still PR seems reasonable:
```py
>>> type(Protocol)
<class 'typing._ProtocolMeta'>
```

AlexWaygood:
> but the metaclass declaration doesn't seem correct

It does feel `more` correct than _not_ having the metaclass declaration, though, since `_ProtocolMeta` is a subclass of `ABCMeta`

srittau:
Also, the behavior is very ABCMeta-like:

```
>>> from typing import Sized
>>> Sized()
  File "/usr/lib/python3.9/typing.py", line 670, in __call__
    result = self.__origin__(*args, **kwargs)
TypeError: Can't instantiate abstract class Sized with abstract method __len__
```

That said, if this is not an easy fix in mypy, I'm fine with reverting for now.

AlexWaygood:
I've just realised that at runtime, all classes directly inheriting from `Protocol` automatically have `_ProtocolMeta` as their metaclass. Maybe ideally type checkers should understand that all protocols have [a subclass of] ABCMeta as the metaclass? Which would mean that we wouldn't need any of these `metaclass=ABCMeta` declarations in `typing.pyi`.

But yeah, I agree with @srittau — whatever the case, it's not really a big deal, so I'm fine with this being merged for now.

JukkaL:
A subclass of Protocol is implicitly an ABC, so explicitly giving the metaclass seems redundant, and we don't do it consistently in typeshed right now -- many protocols don't have an explicit metaclass. 

Relevant example:
```py
import abc
from typing import Protocol

class P(Protocol):
    @abc.abstractmethod
    def f(self) -> None: pass

class C(P):
    pass

C()  # TypeError: Can't instantiate abstract class C with abstract method f
```
I think that type checkers should infer the correct metaclass automatically. Mypy doesn't seem to do it right now. I've filed an issue about this: https://github.com/python/mypy/issues/13979

AlexWaygood:
Probably the reason why we specify the metaclass so consistently in `typing.pyi` is because at runtime these classes aren't actually protocols (they're ABCs, defined in `_collections_abc.py` rather than `typing.py`). It looks like laboriously specifying the metaclass in `typing.pyi` is a holdover from before these classes were made protocols in the stub: https://github.com/python/typeshed/commit/ec2cb8e44fb924f32c5d925942c5818495400b57

JelleZijlstra:
Why does the metaclass cause problems for `Sized` but not for `Hashable` and `SupportsIndex`, which look the same in the stub?

sobolevn:
Related https://github.com/python/typeshed/issues/8998

erictraut:
Many classes in typeshed claim to derive from `Protocol` but do not actually derive from `Protocol` at runtime. For example, according to the typeshed stubs, `str` derives from `Sequence` which derives from `Collection` which derives from `Protocol`. If a type checker were to assume that `Protocol` has a metaclass of `_ProtocolMeta`, an attempt to subclass from `str` and `Enum` will fail at type checking time but succeed at runtime. Unless typeshed can distinguish between a "real" `Protocol` and a "fake" `Protocol`, I don't think type checkers can properly detect when the introduction of a custom metaclass will generate a runtime exception.

JukkaL:
To summarize the discussion above, this works around a bug in mypy and the original ABCMeta metaclass was correct, since `Sized` isn't a `Protocol` subclass at runtime. After python/mypy#13986 is fixed, this change can be reverted.

PR #9065, opened 2022-11-02
codecs: improve bytes handling

sobolevn:
Source code: https://github.com/python/cpython/blob/0689b99bb8c4f6058af43a52effaa8a25609dbed/Lib/codecs.py#L312-L322

You can use `+` with `bytes`:

```python
>>> b'' + bytearray()
b''
>>> b'' + memoryview(b'a')
b'a'
>>> b'' + memoryview(b'a')
b'a'

>>> import array
>>> b'1' + array.array('b')
b'1'
```

Refs https://github.com/python/typeshed/pull/9059


sobolevn:
Mypy correctly raises: 

```
 stdlib/io.pyi:193: error: Argument 1 of "decode" is incompatible with supertype "IncrementalDecoder"; supertype defines the argument type as "Union[bytes, Union[bytearray, memoryview, array[Any], mmap, _CData, PickleBuffer]]"
```

It needs changes from https://github.com/python/typeshed/pull/9059 to be correct.
I will merge these two prs into https://github.com/python/typeshed/pull/9059

PR #9066, opened 2022-11-02
Bump jsonschema to 4.17.*

sobolevn:
Technically it also can be `Unset`: https://github.com/python-jsonschema/jsonschema/blob/d93b9670c60d108e63bce4a4bf72cc26d9fad21f/jsonschema/validators.py#L218

Closes https://github.com/python/typeshed/pull/9062


AlexWaygood (approved):
ty ty

PR #9068, opened 2022-11-02
Bump various test dependencies; pin Python to <3.11 in CI to workaround pytype incompatibility

### LINKED ISSUE #9067: A lot of CI failures bacause of `No matching distribution found for pytype==2022.08.30`
Example: https://github.com/python/typeshed/actions/runs/3379717081/jobs/5611639599

I am not sure we can anything from our side, except trying to clear cache: https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#managing-caches

I cannot do this myself, can someone else please try it?

AlexWaygood:
Fixes #9067 (I've run CI several times on my fork, to check)

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -28,7 +28,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: actions/setup-python@v4
         with:
-          python-version: "3.9 - 3.10"
+          python-version: "3.10"
       - run: pip install -r requirements-tests.txt
       - run: ./tests/check_consistent.py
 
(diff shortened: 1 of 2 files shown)


.github/workflows/tests.yml
```diff
@@ -28,7 +28,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: actions/setup-python@v4
         with:
-          python-version: 3.x
+          python-version: "3.9 - 3.10"
```
JelleZijlstra, line 31: Why not just 3.10?

AlexWaygood: I thought it might make it faster if we specified "use Python 3.9 or 3.10, I don't care". But actually, both should be in the runner's tool cache anyway. So it shouldn't matter. I'll change it so it's just 3.10 :)

JelleZijlstra (approved):

AlexWaygood:
Shoot, screwed up the commit message. Ah well.

PR #9070, opened 2022-11-03
importlib: improve bytes handling

JelleZijlstra:


stdlib/importlib/abc.pyi
```diff
@@ -70,17 +77,17 @@ class MetaPathFinder(Finder):
...
     # Not defined on the actual class, but expected to exist.
     def find_spec(self, fullname: str, target: types.ModuleType | None = ...) -> ModuleSpec | None: ...
 
 class FileLoader(ResourceLoader, ExecutionLoader, metaclass=ABCMeta):
     name: str
-    path: _Path
-    def __init__(self, fullname: str, path: _Path) -> None: ...
-    def get_data(self, path: _Path) -> bytes: ...
-    def get_filename(self, name: str | None = ...) -> _Path: ...
+    path: str
```
JelleZijlstra, line 87: This class has no implementation, but from `_bootstrap_external.FileLoader` it seems clear the path is not intended to be bytes: https://github.com/python/cpython/blob/2cfcaf5af602b297fc90086de4d8ac980c7891e2/Lib/importlib/_bootstrap_external.py#L1195

stdlib/importlib/machinery.pyi
```diff
@@ -113,10 +114,10 @@ class PathFinder:
 
     @classmethod
     def find_spec(
-        cls, fullname: str, path: Sequence[bytes | str] | None = ..., target: types.ModuleType | None = ...
+        cls, fullname: str, path: Sequence[str] | None = ..., target: types.ModuleType | None = ...
```
JelleZijlstra, line 117: The path gets passed to `_get_spec`, which ignores any non-str entries in the path: https://github.com/python/cpython/blob/2cfcaf5af602b297fc90086de4d8ac980c7891e2/Lib/importlib/_bootstrap_external.py#L1525

stdlib/importlib/abc.pyi
```diff
@@ -47,40 +52,40 @@ class InspectLoader(Loader):
     def get_source(self, fullname: str) -> str | None: ...
     def exec_module(self, module: types.ModuleType) -> None: ...
     @staticmethod
-    def source_to_code(data: bytes | str, path: str = ...) -> types.CodeType: ...
+    def source_to_code(data: ReadableBuffer | str, path: str = ...) -> types.CodeType: ...
```
srittau, line 55: Could also be `AST`, but I don't think that's too important.

srittau (approved):

PR #9072, opened 2022-11-03
os: improve bytes handling

JelleZijlstra:


stdlib/os/__init__.pyi
```diff
@@ -503,11 +503,14 @@ if sys.platform != "win32":
     def getenvb(key: bytes) -> bytes | None: ...
     @overload
     def getenvb(key: bytes, default: _T) -> bytes | _T: ...
+    def putenv(__name: StrOrBytesPath, __value: StrOrBytesPath) -> None: ...
+    def unsetenv(__name: StopAsyncIteration) -> None: ...
```
hauntsaninja, line 507: Where are you getting StopAsyncIteration from?

JelleZijlstra: lol, autocomplete
```suggestion
    def unsetenv(__name: StrOrBytesPath) -> None: ...
```
I'm impressed this had no mypy-primer hits.

hauntsaninja (approved):
Everything else in here looks good!

stdlib/os/__init__.pyi
hauntsaninja: tsk, symbolic autocomplete. copilot would never

PR #9073, opened 2022-11-03
pickletools: accepts bytearray

JelleZijlstra:


stdlib/pickle.pyi
```diff
@@ -98,7 +98,7 @@ class _ReadableFileobj(Protocol):
     def readline(self) -> bytes: ...
 
 class _WritableFileobj(Protocol):
-    def write(self, __b: bytes) -> Any: ...
+    def write(self, __b: bytes) -> object: ...
```
AlexWaygood, line 101: Isn't this Protocol identical to `_typeshed.SupportsWrite[bytes]`?

srittau (approved):
LGTM, optionally what Alex said.

stdlib/pickle.pyi
JelleZijlstra: Good catch, removed it

srittau (approved):

PR #9075, opened 2022-11-03
http.server.SimpleHTTPRequestHandler: Fix version branching

AlexWaygood:
The stub currently implies that the index_pages parameter has been removed in 3.12. Actually, it was only added in 3.12 :)

See https://github.com/python/cpython/pull/31985. (This is a follow-up to https://github.com/python/typeshed/pull/9055 — cc. @JelleZijlstra.)


srittau (approved):
Oops, sorry.

JelleZijlstra:
Thanks for catching this!

PR #9076, opened 2022-11-03
tarfile: improve bytes handling

sobolevn:
It works with `bytearray` as well:

```
>>> tarfile.TarInfo.frombuf(b, 'ascii', 'strict')
<TarInfo 'ustar/conttype' at 0x105e87af0>
>>> tarfile.TarInfo.frombuf(bytearray(b), 'ascii', 'strict')
<TarInfo 'ustar/conttype' at 0x105e87bd0>
>>> tarfile.TarInfo.frombuf(memoryview(b), 'ascii', 'strict')
  File "/Users/sobolev/Desktop/cpython/Lib/tarfile.py", line 1064, in frombuf
    if buf.count(NUL) == BLOCKSIZE:
       ^^^^^^^^^
AttributeError: 'memoryview' object has no attribute 'count'
```

Source: https://github.com/python/cpython/blob/e9ac890c0273aee413aa528cc202c3efa29f1d7a/Lib/tarfile.py#L1057


JelleZijlstra (approved):
The metthod uses `.count()`, `len()`, and slicing, so it won't work with any other well-known buffer types.

PR #9078, opened 2022-11-03
`beautifulsoup`: change `Tag` to accept `dict[str, Any]` for the `attrs` parameter

### LINKED ISSUE #8755: [bs4] Incorrect type hints for `.__getitem__()` and `.get()`?
`Tag.attrs` is type-hinted as `dict[str, str]` as expected. However, `Tag.__getitem__()` is type hinted as returning `str | list[str]`, and `Tag.get()` is type hinted as returning `str | list[str] | None`. Why is that the case?

AmberAnsari89:
Fixes #8755


AmberAnsari89:
Thank you for your guidance. I noticed it and was trying to figure out why
it is happening.you guidance really help me. I will do the needful.

PR #9083, opened 2022-11-04
mailbox: improve bytes handling

JelleZijlstra:
- The _MessageData alias refers to objects handled in the
  _dump_message method: https://github.com/python/cpython/blob/016c7d37b6acfe2203542a2655080c6402b3be1f/Lib/mailbox.py#L210
- The path passed to __init__ should be a str path because there are
  many places in the file (search for self._path) where we join it
  with a str, which won't work with a bytes path.


stdlib/mailbox.pyi
```diff
@@ -32,7 +33,9 @@ __all__ = [
 
 _T = TypeVar("_T")
 _MessageT = TypeVar("_MessageT", bound=Message)
-_MessageData: TypeAlias = email.message.Message | bytes | str | IO[str] | IO[bytes]
+_MessageData: TypeAlias = email.message.Message | bytes | str | io.StringIO | _SupportsReadAndReadline
+
+class _SupportsReadAndReadline(SupportsRead[bytes], SupportsNoArgReadline[bytes], Protocol): ...
```
AlexWaygood, line 38: Nit: It's slightly confusing to have the alias defined before the protocol, since the alias depends on the protocol

```suggestion

class _SupportsReadAndReadline(SupportsRead[bytes], SupportsNoArgReadline[bytes], Protocol): ...

_MessageData: TypeAlias = email.message.Message | bytes | str | io.StringIO | _SupportsReadAndReadline
```

AlexWaygood (approved):
Looks great!

stdlib/mailbox.pyi
JelleZijlstra: Thanks!

PR #9084, opened 2022-11-04
typing: accept buffers in typing.IO.write

### LINKED ISSUE #9082: Support buffers in `IO[bytes].write`
Trying to annotate psycopg code in order to pass mypy check with disabled bytearray/memoryview promotion to bytes [see branch mypy-0990](https://github.com/psycopg/psycopg/tree/mypy-0990).

Everything was smooth, except one questionable point:

```python
from typing import IO, Union

Buffer: TypeAlias = Union[bytes, bytearray, memoryview]

def write1(f: IO[bytes], data: Buffer) -> None:
    f.write(data)
```

Mypy 0.990 reports:

```
error: Argument 1 to "write" of "IO" has incompatible type "Union[bytes, bytearray, memoryview]"; expected "bytes"  [arg-type]
```

Same using `BinaryIO` instead of `IO[bytes]`.

As far as I can see, a binary file supports any object supporting the buffer protocol (or maybe less than that? However, the three types in the union tested are definitely supported).

Declaring `f: IO[Buffer]` almost works: `write()` doesn't complain anymore, but `IO` does: *Invalid type argument value for "IO"*.

Maybe something could be relaxed here?

JelleZijlstra:
Fixes #9082


JelleZijlstra:
The mypy failures are a bug, python/mypy#14002.

srittau:
This looks quite disruptive to me and I'm not sure whether it's worth it, considering the slightly legacy nature of `IO`.

JelleZijlstra:
Worth noting that most of the mypy-primer fallout is due to a mypy bug (python/mypy#14002).

hauntsaninja:
This seems useful, but given that I think I have a mypy fix, let's wait until it's released: https://github.com/python/mypy/pull/14017

stdlib/codecs.pyi
```diff
@@ -239,7 +239,7 @@ class StreamReaderWriter(TextIO):
     def __next__(self) -> str: ...
     def __iter__(self: Self) -> Self: ...
     def write(self, data: str) -> None: ...  # type: ignore[override]
-    def writelines(self, list: Iterable[str]) -> None: ...
+    def writelines(self, list: Iterable[str]) -> None: ...  # type: ignore[override]  # https://github.com/python/mypy/issues/14002
```
AlexWaygood, line 242: We probably don't need this `type: ignore` anymore (as well as several others), with the latest mypy release.

stdlib/http/client.pyi
```diff
@@ -101,7 +101,7 @@ class HTTPMessage(email.message.Message):
 
 def parse_headers(fp: io.BufferedIOBase, _class: Callable[[], email.message.Message] = ...) -> HTTPMessage: ...
 
-class HTTPResponse(io.BufferedIOBase, BinaryIO):
+class HTTPResponse(io.BufferedIOBase, BinaryIO):  # type: ignore[misc]  # https://github.com/python/mypy/issues/14002
```
AlexWaygood, line 104: We do still need this `type: ignore`, but not because of the mypy issue linked (which is now closed) -- we need it because the definitions of `writelines()` in the two base classes are incompatible. The same goes for all the `type: ignore`s in `io.pyi` and the one in `lzma.pyi`.

stdlib/tempfile.pyi
```diff
@@ -215,8 +215,14 @@ class _TemporaryFileWrapper(Generic[AnyStr], IO[AnyStr]):
...
-    def write(self, s: AnyStr) -> int: ...
-    def writelines(self, lines: Iterable[AnyStr]) -> None: ...
+    @overload
+    def write(self: _TemporaryFileWrapper[str], s: str) -> int: ...
+    @overload
+    def write(self: _TemporaryFileWrapper[bytes], s: ReadableBuffer) -> int: ...
+    @overload
+    def writelines(self: _TemporaryFileWrapper[str], lines: Iterable[str]) -> None: ...
+    @overload
+    def writelines(self: _TemporaryFileWrapper[bytes], lines: Iterable[ReadableBuffer]) -> None: ...
```
AlexWaygood, line 225: These overloads might not work very well with pyright, which has different handling of self types to mypy (#9591). The same goes for the overloads on lines 401-408, and the overloads in `typing.pyi`. You might want to instead do something similar to what I did with `re.Pattern` and `re.Match` in #9592.

AlexWaygood (commented):
Looks like we can go back to this now, with the latest mypy release!

AlexWaygood:
Ping @JelleZijlstra, have you had a chance to take a look at my review feedback? :)

stdlib/tempfile.pyi
JelleZijlstra: Changed, and added some testcases for `IO.write`.

stdlib/codecs.pyi
JelleZijlstra: Cleaned up all comments referencing that issue.

stdlib/http/client.pyi
JelleZijlstra: Updated these comments.

JelleZijlstra:
hm, the new way to write the overloads makes subclassing much worse. Let me see if I can fix that.

JelleZijlstra:
Seems like we have a choice between:

1. One overload for AnyStr and another for Buffer: pyright accepts `IO.write(AnyStr)`, but mypy hits python/mypy#14866 and subclasses of IO classes that override the `write`/`writelines` method start complaining.
2. One overload for str and another for bytes + Buffer: mypy is fine with more subclasses, but pyright rejects `IO.write(AnyStr)`. 

I'm inclined to go with (1) because mypy's behavior is more clearly buggy than pyright's, but ideally we'd fix the mypy bug first.

AlexWaygood:
Now that https://github.com/python/mypy/pull/14882 has been merged, I guess we have two options:

- we can either merge #9861 now (and simplify the overloads after mypy 1.2 comes out, perhaps);
- or we close #9861, wait for mypy 1.2, and then merge this.

I'm easy either way :)

PR #9085, opened 2022-11-04
multiprocessing: improve bytes handling

JelleZijlstra:


sobolevn (approved):
It is bassed to `memoryview(buf)`: https://github.com/python/cpython/blob/016c7d37b6acfe2203542a2655080c6402b3be1f/Lib/multiprocessing/connection.py#L185 which works with `ReadableBuffer`

Issue #9061, opened 2022-11-01
Add stubs for pycocotools


hoel-bagard:
Hi,

I would be interested in adding stubs for [pycocotools](https://github.com/ppwwyyxx/cocoapi) to typeshed (original repo [here](https://github.com/cocodataset/cocoapi), on pypi [here](https://pypi.org/project/pycocotools/))

If there isn't any issue with that, I would like to start working on a PR.

srittau:
We welcome all type stubs for PyPI packages that don't ship with type information here on typeshed! I'm going to close this issue for now per our policy, but please free to open a PR. It's also okay to open a draft PR while working on it if you want early feedback.

PR #9086, opened 2022-11-04
Add stubs for pycocotools

hoel-bagard:
This PR adds stubs for pycocotools (see #9061).


AlexWaygood:
Unfortunately you won't be able to use `numpy` as a dependency just yet. This should hopefully be changing soon; see #5768.

For now, as a workaround, I recommend you do something similar to what we do here:

https://github.com/python/typeshed/blob/3ac147cf673ecb2b915611a3ac45d691360b32f6/stubs/D3DShot/d3dshot/capture_outputs/numpy_capture_output.pyi#L14

hoel-bagard:
@AlexWaygood Thank you for the workaround. I've removed the use of numpy, but if possible I'd like to leave the numpy typing information in case it can be used in the future. I'm not sure if it's better leave all the original lines as comments like [here](https://github.com/hoel-bagard/typeshed/blob/main/stubs/pycocotools/pycocotools/cocoeval.pyi) or to create aliases for every type I'm using like [here](https://github.com/hoel-bagard/typeshed/blob/main/stubs/pycocotools/pycocotools/mask.pyi).

AlexWaygood:
You can probably actually just replace the `import` line with a `TypeAlias` to `Any`:

```diff
- import numpy as np
- import numpy.typing as not
+ # TODO when #5768 is resolved
+ # import numpy as np
+ # import numpy.typing as npt
+ np: TypeAlias = Incomplete
+ npt: TypeAlias = Incomplete
```

And then have the rest of the stub as it would be if you were *actually* importing `numpy as np`. I'm not 100% sure if that'll work, but might be worth a shot?

(`TypeAlias` needs to be imported from `typing_extensions`, BTW. And you'll need to use `_typeshed.Self` instead of `typing(_extensions).Self` for now, since mypy doesn't support `typing(_extensions).Self` yet. Note that `_typeshed.Self` is just a normal TypeVar, so it works slightly differently to `typing(_extensions).Self`.)

hoel-bagard:
I've tried replacing the `import` line with a `TypeAlias` to `Incomplete`, but it doesn't work. I get errors like these:
```
stubs/pycocotools/pycocotools/coco.pyi:41: error: Name "npt.NDArray" is not defined
stubs/pycocotools/pycocotools/coco.pyi:41: error: Name "np.float64" is not defined
```

hoel-bagard:
I've tried replacing the `import` line with a `TypeAlias` to `Incomplete`, but it doesn't work. I get errors like these:
```
stubs/pycocotools/pycocotools/coco.pyi:41: error: Name "npt.NDArray" is not defined
stubs/pycocotools/pycocotools/coco.pyi:41: error: Name "np.float64" is not defined
```

AlexWaygood:
> I've tried replacing the `import` line with a `TypeAlias` to `Incomplete`, but it doesn't work. I get errors like these:
> 
> ```
> stubs/pycocotools/pycocotools/coco.pyi:41: error: Name "npt.NDArray" is not defined
> stubs/pycocotools/pycocotools/coco.pyi:41: error: Name "np.float64" is not defined
> ```

Okay, what you've done looks fine, in that case. It would be good to include a reference to #5768 in a comment somewhere, though — that way we'll find it when we're grepping through the codebase after #5768 is solved

pyrightconfig.stricter.json
```diff
@@ -59,6 +59,7 @@
         "stubs/peewee",
         "stubs/psutil",
         "stubs/psycopg2",
+        "stubs/pycocotools",
```
AlexWaygood, line 62: This is automatically added by the stub generation script on the assumption that most contributed stubs will start off being somewhat incomplete. But I think you might have annotated everything, so I'm not sure it's actually needed!

```suggestion
```

stubs/pycocotools/pycocotools/mask.pyi
```diff
@@ -0,0 +1,29 @@
...
+_NDArrayFloat64: TypeAlias = Incomplete  # npt.NDArray[np.float64]
+
+def iou(
+    dt: _NDArrayUInt32 | list[float] | list[_EncodedRLE],
+    gt: _NDArrayUInt32 | list[float] | list[_EncodedRLE],
+    pyiscrowd: list[int] | _NDArrayUInt8,
+) -> list[Any] | _NDArrayFloat64: ...
+def merge(rleObjs: list[_EncodedRLE], intersect: int = ...) -> _EncodedRLE: ...
+@overload
+def frPyObjects(pyobj: _NDArrayUInt32 | list[list[int]] | list[_EncodedRLE], h: int, w: int) -> list[_EncodedRLE]: ...
```
AlexWaygood, line 23:
```suggestion
# ignore an "overlapping overloads" error due to _NDArrayInt32 being an alias for `Incomplete` for now
@overload
def frPyObjects(pyobj: _NDArrayUInt32 | list[list[int]] | list[_EncodedRLE], h: int, w: int) -> list[_EncodedRLE]: ...  # type: ignore[misc]
```

hoel-bagard:
@AlexWaygood Two checks are still failing, but I don't understand why or how to fix it. Would you have time to look at it ?

AlexWaygood:
Sure, I'll take a look!

stubs/pycocotools/pycocotools/__init__.pyi
```diff
@@ -0,0 +1,3 @@
+from .coco import *
+from .cocoeval import *
+from .mask import *
```
AlexWaygood, line 3: It looks like none of these `from .foo import *` imports exist at runtime, and that's causing most of the stubtest failures: at runtime `__init__.py` is just an empty file (other than `__author__`): https://github.com/ppwwyyxx/cocoapi/blob/71e284ef862300e4319aacd523a64c7f24750178/PythonAPI/pycocotools/__init__.py#L1

stubs/pycocotools/pycocotools/coco_types.pyi
```diff
@@ -0,0 +1,72 @@
+from _typeshed import Incomplete
```
AlexWaygood, line 1: This module doesn't exist at runtime, and that's causing another stubtest failure. You should delete this file.

For these "private-to-the-stub" types like the `_Image` `TypedDict`:
- If a type is only used in one file, you should put the definition in that file (that will also make flake8-pyi happier with you).
- If a type is used in multiple files, you should put it in `__init__.py`, and import it from there. You may have to add a few `# noqa` comments if you do that, as flake8-pyi will believe that the private `TypedDict`s (etc) are unused; it won't be able to tell that they're imported in other files.

AlexWaygood (commented):
Stubtest is a tool that compares the stub to the runtime, and verifies that the two are consistent with each other. Currently stubtest is unhappy about a few imports that don't look like they exist at runtime, and a file that doesn't exist at runtime :)

hoel-bagard:
Thank you!
I've done as you instructed and now all the checks are passing. Is there anything else that needs to be modified ?

AlexWaygood:
Thank you! I still need to do a manual review, but it looks like we're nearly there :)

stubs/pycocotools/pycocotools/coco.pyi
```diff
@@ -0,0 +1,89 @@
...
+    dataset: _Dataset
+    cats: dict[int, _Category]
+    imgs: dict[int, _Image]
+    imgToAnns: dict[int, list[_Annotation]]
+    catToImgs: dict[int, list[int]]
+    def __init__(self, annotation_file: str | Path | None = ...) -> None: ...
+    def createIndex(self) -> None: ...
+    def info(self) -> None: ...
+    def getAnnIds(
+        self, imgIds: list[int] | int = ..., catIds: list[int] | int = ..., areaRng: list[float] = ..., iscrowd: bool | None = ...
```
AlexWaygood, line 68: It's generally best to avoid using `list[T]` for parameter annotations. It can lead to "annoying" errors from a type checker, due to the fact that `list` is invariant. In this case, for the first two parameters, the method will accept an `int`, or [any object that passes the `_isArrayLike` function](https://github.com/ppwwyyxx/cocoapi/blob/71e284ef862300e4319aacd523a64c7f24750178/PythonAPI/pycocotools/coco.py#L136-L137). `_isArrayLike` [checks for whether an object is iterable and sized](https://github.com/ppwwyyxx/cocoapi/blob/71e284ef862300e4319aacd523a64c7f24750178/PythonAPI/pycocotools/coco.py#L63-L64), so `Collection[int]` is a good fit here. (You'll need to import `Collection` from `collections.abc`.)

The third parameter must support indexing, so `Sequence` is a better fit for that one.

```suggestion
        self, imgIds: Collection[int] | int = ..., catIds: Collection[int] | int = ..., areaRng: Sequence[float] = ..., iscrowd: bool | None = ...
```

stubs/pycocotools/pycocotools/coco.pyi
```diff
@@ -0,0 +1,89 @@
...
+    ) -> list[int]: ...
+    def getCatIds(
+        self, catNms: list[str] | str = ..., supNms: list[str] | str = ..., catIds: list[int] | int = ...
```
AlexWaygood, line 71:
```suggestion
        self, catNms: Collection[str] | str = ..., supNms: Collection[str] | str = ..., catIds: Collection[int] | int = ...
```

stubs/pycocotools/pycocotools/coco.pyi
```diff
@@ -0,0 +1,89 @@
...
+    ) -> list[int]: ...
+    def getImgIds(self, imgIds: list[int] | int = ..., catIds: list[int] | int = ...) -> list[int]: ...
+    def loadAnns(self, ids: list[int] | int = ...) -> list[_Annotation]: ...
+    def loadCats(self, ids: list[int] | int = ...) -> list[_Category]: ...
+    def loadImgs(self, ids: list[int] | int = ...) -> list[_Image]: ...
```
AlexWaygood, line 76:
```suggestion
    def getImgIds(self, imgIds: Collection[int] | int = ..., catIds: list[int] | int = ...) -> list[int]: ...
    def loadAnns(self, ids: Collection[int] | int = ...) -> list[_Annotation]: ...
    def loadCats(self, ids: Collection[int] | int = ...) -> list[_Category]: ...
    def loadImgs(self, ids: Collection[int] | int = ...) -> list[_Image]: ...
```

stubs/pycocotools/pycocotools/coco.pyi
```diff
@@ -0,0 +1,89 @@
...
+    def showAnns(self, anns: list[_Annotation], draw_bbox: bool = ...) -> None: ...
```
AlexWaygood, line 77:
```suggestion
    def showAnns(self, anns: Sequence[_Annotation], draw_bbox: bool = ...) -> None: ...
```

stubs/pycocotools/pycocotools/coco.pyi
```diff
@@ -0,0 +1,89 @@
...
+    def loadRes(self: Self, resFile: str) -> Self: ...
```
AlexWaygood, line 78: This method [always returns an instance of `COCO`](https://github.com/ppwwyyxx/cocoapi/blob/71e284ef862300e4319aacd523a64c7f24750178/PythonAPI/pycocotools/coco.py#L313), even if it's called on a subclass, so using `Self` here isn't correct.

To illustrate, if you had the following code:

```python
class Foo(COCO): ...
bar = Foo().loadRes('baz')
```

Then, with your current annotation, mypy/pyright would infer the type of `bar` to be `Foo`. But that would be incorrect: the type of `bar` would be `COCO`.

```suggestion
    def loadRes(self, resFile: str) -> COCO: ...
```

stubs/pycocotools/pycocotools/coco.pyi
```diff
@@ -0,0 +1,89 @@
...
+    def download(self, tarDir: str | None = ..., imgIds: list[int] = ...) -> Literal[-1] | None: ...
```
AlexWaygood, line 79:
```suggestion
    def download(self, tarDir: str | None = ..., imgIds: Collection[int] = ...) -> Literal[-1] | None: ...
```

stubs/pycocotools/pycocotools/mask.pyi
```diff
@@ -0,0 +1,31 @@
...
+_NPUInt32: TypeAlias = Incomplete  # np.uint32
+_NDArrayUInt8: TypeAlias = Incomplete  # npt.NDArray[np.uint8]
+_NDArrayUInt32: TypeAlias = Incomplete  # npt.NDArray[np.uint32]
+_NDArrayFloat64: TypeAlias = Incomplete  # npt.NDArray[np.float64]
+
+def iou(
+    dt: _NDArrayUInt32 | list[float] | list[_EncodedRLE],
+    gt: _NDArrayUInt32 | list[float] | list[_EncodedRLE],
+    pyiscrowd: list[int] | _NDArrayUInt8,
+) -> list[Any] | _NDArrayFloat64: ...
```
AlexWaygood, line 20: (No change requested, just pointing out to clarify:) This function _should_ have the parameters annotated as `list`, since the code explicitly forbids other kinds of sequences: https://github.com/ppwwyyxx/cocoapi/blob/71e284ef862300e4319aacd523a64c7f24750178/PythonAPI/pycocotools/_mask.pyx#L182

AlexWaygood (commented):
Thanks for your patience, @hoel-bagard, and I'm sorry I've taken so long in getting to this. I've been busier than I expected the last few weeks :(

This mostly looks good, but I've left a few review comments. In general, we need to be quite cautious about using concrete types for parameter annotations. Abstract types are usually better -- they mean we don't run into annoying issues around invariance. There's some documentation on this here:

- https://mypy.readthedocs.io/en/stable/common_issues.html#invariance-vs-covariance.
- https://mypy.readthedocs.io/en/stable/generics.html#variance-of-generics

AlexWaygood (approved):
Thank you! Sorry again about the wait :)

hoel-bagard:
Thank you for the great review!

PR #9089, opened 2022-11-04
Updated some pywin32 unknown types

Avasam:
I've updated a few Unknown types, namedtuples in `win32gui_struct` and some comments.

The namedtuples in `win32gui_struct` are not directly accessible. They are dynamically generated by `_MakeResult` whenever a function is called.


JelleZijlstra (approved):
Didn't review against the implementation, but looks reasonable.

PR #9091, opened 2022-11-04
Add new functions for Babel 2.11.0

### LINKED ISSUE #9090: Support types for `format_compact_decimal` (Babel 2.11.0)
I am getting an error using `format_compact_decimal` with Pyright and I was told the issue is that it's missing from the typeshed stubs here.

https://github.com/microsoft/pyright/issues/4138

(This will probably also apply to `format_compact_currency`, `Format.compact_decimal` and `Format.compact currency` which will be in the next Babel release (2.12.0))

DenverCoder1:
* Adds new functions added to Babel 2.11.0 in https://github.com/python-babel/babel/pull/909

* Fixes #9090


stubs/babel/babel/numbers.pyi
```diff
@@ -34,6 +34,7 @@ def get_decimal_quantum(precision): ...
 def format_decimal(
     number, format: Any | None = ..., locale=..., decimal_quantization: bool = ..., group_separator: bool = ...
 ): ...
+def format_compact_decimal(number, *, format_type: str = ..., locale=..., fraction_digits: int = ...): ...
```
AlexWaygood, line 37: - Looks like `locale` can be anything that can be passed to `babel.core.Locale.parse`, which means it should be `Locale | str | None`
- Looks like the function returns whatever `babel.NumberPattern.apply` returns, and I think that only ever returns a `str`

```suggestion
def format_compact_decimal(number, *, format_type: str = ..., locale: Locale | str | None =..., fraction_digits: int = ...) -> str: ...
```

AlexWaygood (commented):
Thanks (and welcome to typeshed)! I'm not too familiar with babel, but here's two suggestions from quickly skimming the source code:

stubs/babel/babel/numbers.pyi
DenverCoder1: I had left it without to be consistent with the functions around it as I'm not sure I have the time to fix annotations for the entire library.

I suppose it makes sense to annotate new functions accurately regardless, though.

AlexWaygood: Stubs are generally incomplete when they're first contributed to typeshed, which is fine (it's a lot of work to contribute stubs for a new package!) — but we reserve the right to be slightly pickier for smaller PRs like this, which only touch a few functions ;)

AlexWaygood:
Btw, we have pre-commit.ci's "autofix" feature enabled for this repo. It looks like you didn't tick the "allow maintainers to edit this PR" box when you filed this PR — I don't think there's any way for you to change it now, but if you had ticked that box, some bots would have already automatically fixed the `black`/`isort` formatting issues for you. Something to bear in mind for future PRs :)

AlexWaygood (approved):
Thanks!

DenverCoder1:
> Btw, we have pre-commit.ci's "autofix" feature enabled for this repo. It looks like you didn't tick the "allow maintainers to edit this PR" box when you filed this PR — I don't think there's any way for you to change it now, but if you had ticked that box, some bots would have already automatically fixed the `black`/`isort` formatting issues for you. Something to bear in mind for future PRs :)

I'm aware of the issue and unfortunately I don't have the option since I created the fork in an Organization account.

I was having a hard time getting pre-commit to run locally and it turns out the `language_version: python3.9` in the `.pre-commit-config.yaml` was what was causing to fail with `RuntimeError: failed to find interpreter for Builtin discover of python_spec='python3.9'`. I believe I've gotten that resolved now by removing it locally.

PR #9094, opened 2022-11-05
smtplib: Improve bytes handling

JelleZijlstra:


stdlib/smtplib.pyi
```diff
@@ -29,7 +29,7 @@ __all__ = [
 _Reply: TypeAlias = tuple[int, bytes]
 _SendErrs: TypeAlias = dict[str, _Reply]
 # Should match source_address for socket.create_connection
```
AlexWaygood, line 31: This comment looks out of date — looks like the alias has diverged somewhat from https://github.com/python/typeshed/blob/9f0d0b9af7d4980431bbd62ca38497cf33308093/stdlib/_socket.pyi#L20

JelleZijlstra: We should just reuse that alias, not sure why I missed that. Source address just gets passed to `socket.create_connection`.

stdlib/smtplib.pyi
```diff
@@ -9,6 +9,8 @@ from types import TracebackType
 from typing import Any, Protocol, overload
 from typing_extensions import TypeAlias
 
+from _socket import _Address
```
AlexWaygood, line 12: To keep it clear what the `_Address` alias is referring to, maybe:

```suggestion
from _socket import _Address as _SocketAddress
```

Or

```suggestion
from _socket import _Address as _SourceAddress
```

stdlib/smtplib.pyi
```diff
@@ -9,6 +9,8 @@ from types import TracebackType
 from typing import Any, Protocol, overload
 from typing_extensions import TypeAlias
 
+from _socket import _Address as _SourceAddress
```
AlexWaygood, line 12: Oh, we should add `_socket` to this list, so isort recognises it as part of the stdlib:

https://github.com/python/typeshed/blob/9f0d0b9af7d4980431bbd62ca38497cf33308093/pyproject.toml#L15

JelleZijlstra: I can do that in a separate PR.

stdlib/smtplib.pyi
```diff
@@ -125,16 +125,16 @@ class SMTP:
     @overload
     def auth_cram_md5(self, challenge: None = ...) -> None: ...
     @overload
-    def auth_cram_md5(self, challenge: bytes) -> str: ...
-    def auth_plain(self, challenge: bytes | None = ...) -> str: ...
-    def auth_login(self, challenge: bytes | None = ...) -> str: ...
+    def auth_cram_md5(self, challenge: ReadableBuffer) -> str: ...
+    def auth_plain(self, challenge: ReadableBuffer | None = ...) -> str: ...
```
AlexWaygood, line 129: The `challenge` parameter doesn't look like it's actually used for this one

AlexWaygood (approved):

stdlib/smtplib.pyi
JelleZijlstra: The next one doesn't really use it either, I figured I should be consistent with `auth_cram_md5`.

AlexWaygood: Yeah, that makes sense, I figured as much!

stdlib/smtplib.pyi
JelleZijlstra: #9098

PR #9095, opened 2022-11-05
socket: constructor accepts bytes on Windows

JelleZijlstra:
https://github.com/python/cpython/blob/d04899abb0da4cc7c0a390f5bc52d57ace8955fd/Modules/socketmodule.c#L5318

This is used in `socket.fromshare`. It must be bytes, not some other buffer.


AlexWaygood:
It took me a while, but I finally managed to find a REPL snippet that demonstrates that this is indeed the case on my Windows machine (there might be a simpler way of demonstrating this, idk):

```pycon
>>> import socket, multiprocessing
>>> sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
>>> p = multiprocessing.Process()
>>> foo = sock.share(p._parent_pid)
>>> type(foo)
<class 'bytes'>
>>> socket.socket(fileno=foo)
<socket.socket fd=168, family=AddressFamily.AF_INET, type=SocketKind.SOCK_STREAM, proto=6>
```

AlexWaygood (approved):

PR #9096, opened 2022-11-05
socketserver: improve bytes handling

JelleZijlstra:


JelleZijlstra:
hm, since these are mostly for subclasses, making the address types precise is probably mostly a hassle for subclassers.

stdlib/socketserver.pyi
```diff
@@ -1,8 +1,8 @@
 import sys
 import types
-from _typeshed import Self
+from _typeshed import ReadableBuffer, Self
 from collections.abc import Callable
-from socket import socket as _socket
+from socket import _RetAddress, socket as _socket
```
AlexWaygood, line 5: Nit: can we import `_RetAddress` from `_socket` rather than `socket`, so I only have to open one file to find out what the alias is, instead of opening up `socket.pyi`, finding it's actually re-exported from `_socket.pyi`, then opening `_socket.pyi`? 🙃

stdlib/socketserver.pyi
```diff
@@ -53,14 +54,14 @@ class BaseServer:
     def serve_forever(self, poll_interval: float = ...) -> None: ...
     def shutdown(self) -> None: ...
     def server_close(self) -> None: ...
-    def finish_request(self, request: _RequestType, client_address: _AddressType) -> None: ...
+    def finish_request(self, request: _RequestType, client_address: _RetAddress) -> None: ...
```
AlexWaygood, line 57: Should we also use the `_RetAddress` alias for the second parameter of `RequestHandlerClass` on lines 45, 49, and 51? (A few other places in this file as well.)

stdlib/socketserver.pyi
```diff
@@ -74,29 +75,29 @@ class TCPServer(BaseServer):
         allow_reuse_port: bool
```
AlexWaygood, line 75: If we're updating the parameter annotation in `__init__`, we need to update the variable annotation as well

```suggestion
        allow_reuse_port: bool
   sever_address: _AfInetAddress
```

stdlib/socketserver.pyi
```diff
@@ -74,29 +75,29 @@ class TCPServer(BaseServer):
...
+    def get_request(self) -> tuple[_socket, _RetAddress]: ...
 
 class UDPServer(TCPServer):
     max_packet_size: ClassVar[int]
-    def get_request(self) -> tuple[tuple[bytes, _socket], Any]: ...  # type: ignore[override]
+    def get_request(self) -> tuple[tuple[bytes, _socket], _RetAddress]: ...  # type: ignore[override]
 
 if sys.platform != "win32":
     class UnixStreamServer(BaseServer):
         def __init__(
```
AlexWaygood, line 90:
```suggestion
        server_address: _AfUnixAddress
        def __init__(
```

stdlib/socketserver.pyi
```diff
@@ -74,29 +75,29 @@ class TCPServer(BaseServer):
...
             self: Self,
-            server_address: str | bytes,
+            server_address: _AfUnixAddress,
             RequestHandlerClass: Callable[[Any, Any, Self], BaseRequestHandler],
             bind_and_activate: bool = ...,
         ) -> None: ...
 
     class UnixDatagramServer(BaseServer):
         def __init__(
```
AlexWaygood, line 98:
```suggestion
        server_address: _AfUnixAddress
        def __init__(
```

AlexWaygood (changes requested):

stdlib/socketserver.pyi
JelleZijlstra: Done

stdlib/socketserver.pyi
JelleZijlstra: Yes, good point

AlexWaygood (commented):
Looks good now, just one more question: Should the `BaseServer.server_address` attribute (on line 39) also be changed to be `_socket._Address`, to match the type passed in `__init__`?

PR #9098, opened 2022-11-05
isort: Add more to extra_standard_library

JelleZijlstra:
Got all underscored modules from VERSIONS.


hauntsaninja (approved):
gotta catch em all

PR #9101, opened 2022-11-05
termios: improve comments, add more constants

JelleZijlstra:
I just got the list in termios.c of all constants that can potentially exist.

### DIFF
--- a/stdlib/termios.pyi
+++ b/stdlib/termios.pyi
@@ -9,45 +9,33 @@ if sys.platform != "win32":
 
     B0: int
     B1000000: int
-    B1000000: int
     B110: int
     B115200: int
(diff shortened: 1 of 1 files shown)


AlexWaygood:
Is there any chance we can reduce the amount we allowlist this module? See also #9150.

JelleZijlstra:
Other than your PR? I think it's hard because the set of constants that exists varies so much by exact OS version.

AlexWaygood:
> Other than your PR? I think it's hard because the set of constants that exists varies so much by exact OS version.

Yeah, fair enough :)

AlexWaygood (approved):
Reviewed by reading the source code — looks good!

PR #9103, opened 2022-11-05
urllib.parse: improve bytes handling

sobolevn:
Source: https://github.com/python/cpython/blob/d04899abb0da4cc7c0a390f5bc52d57ace8955fd/Lib/urllib/parse.py

## quote, quote_from_bytes, quote_plus

- `quote_from_bytes` is always called in the end of all three cases
- `quote_from_bytes` handles `safe` arg: it can be `ReadableBuffer`: https://github.com/python/cpython/blob/d04899abb0da4cc7c0a390f5bc52d57ace8955fd/Lib/urllib/parse.py#L900-L905

```python
>>> parse.quote(b'a')
'a'
>>> parse.quote(bytearray(b'a'))
'a'
>>> parse.quote(memoryview(b'a'))
  File "/Users/sobolev/.pyenv/versions/3.10.0/lib/python3.10/urllib/parse.py", line 870, in quote
    return quote_from_bytes(string, safe)
  File "/Users/sobolev/.pyenv/versions/3.10.0/lib/python3.10/urllib/parse.py", line 895, in quote_from_bytes
    raise TypeError("quote_from_bytes() expected bytes")
TypeError: quote_from_bytes() expected bytes
```

```python
>>> parse.quote(bytearray(b'a'), bytearray(b'/'))
'a'
>>> parse.quote(bytearray(b'a'), memoryview(b'/'))
'a'
```

## urlencode

Passes `safe` to `quote` or `quote_plus`

## urlparse

```python
>>> parse.urlparse(b'http://ya.ru')
ParseResultBytes(scheme=b'http', netloc=b'ya.ru', path=b'', params=b'', query=b'', fragment=b'')
>>> parse.urlparse(bytearray(b'http://ya.ru'))
ParseResultBytes(scheme=b'http', netloc=b'ya.ru', path=b'', params=b'', query=b'', fragment=b'')
```

## urlsplit

```python
>>> parse.urlsplit(bytearray(b'http://ya.ru'))
SplitResultBytes(scheme=b'http', netloc=b'ya.ru', path=b'', query=b'', fragment=b'')
```

## Final note

However, some functions work with empty `memoryview` objects. But fail with non-empty ones: 

```python
>>> parse.urlparse(memoryview(b''))
ParseResultBytes(scheme=b'', netloc=b'', path=b'', params=b'', query=b'', fragment=b'')
>>> parse.urlparse(memoryview(b'http://ya.ru'))
  File "/Users/sobolev/.pyenv/versions/3.10.0/lib/python3.10/urllib/parse.py", line 392, in urlparse
    url, scheme, _coerce_result = _coerce_args(url, scheme)
  File "/Users/sobolev/.pyenv/versions/3.10.0/lib/python3.10/urllib/parse.py", line 128, in _coerce_args
    return _decode_args(args) + (_encode_result,)
  File "/Users/sobolev/.pyenv/versions/3.10.0/lib/python3.10/urllib/parse.py", line 112, in _decode_args
    return tuple(x.decode(encoding, errors) if x else '' for x in args)
  File "/Users/sobolev/.pyenv/versions/3.10.0/lib/python3.10/urllib/parse.py", line 112, in <genexpr>
    return tuple(x.decode(encoding, errors) if x else '' for x in args)
AttributeError: 'memoryview' object has no attribute 'decode'
```

I've decided not to add `memoryview` in cases like this.


sobolevn:
One unrelated failure: https://github.com/python/typeshed/actions/runs/3399278092/jobs/5652918608

I don't have "restart this job" button and I don't want to retrigger the whole ci.

stdlib/urllib/parse.pyi
```diff
@@ -132,14 +133,14 @@ def parse_qsl(
     separator: str = ...,
 ) -> list[tuple[AnyStr, AnyStr]]: ...
 @overload
-def quote(string: str, safe: str | bytes = ..., encoding: str | None = ..., errors: str | None = ...) -> str: ...
+def quote(string: str, safe: str | ReadableBuffer = ..., encoding: str | None = ..., errors: str | None = ...) -> str: ...
```
JelleZijlstra, line 136: From the link you posted, `safe` is `str | Iterable[int]` instead. The code doesn't use the buffer protocol. For example, `urllib.parse.quote("x", safe=pickle.PickleBuffer(b"x"))` gives `TypeError: 'pickle.PickleBuffer' object is not iterable`.

stdlib/urllib/parse.pyi
```diff
@@ -152,24 +153,24 @@ def unquote_plus(string: str, encoding: str = ..., errors: str = ...) -> str: ..
...
+def urldefrag(url: bytes | bytearray | None) -> DefragResultBytes: ...
 def urlencode(
     query: Mapping[Any, Any] | Mapping[Any, Sequence[Any]] | Sequence[tuple[Any, Any]] | Sequence[tuple[Any, Sequence[Any]]],
     doseq: bool = ...,
-    safe: str | bytes = ...,
+    safe: str | ReadableBuffer = ...,
     encoding: str = ...,
     errors: str = ...,
-    quote_via: Callable[[AnyStr, str | bytes, str, str], str] = ...,
+    quote_via: Callable[[AnyStr, str | ReadableBuffer, str, str], str] = ...,
```
JelleZijlstra, line 163: This is dangerous because it makes the type more restrictive for callers: the callback now has to accept any ReadableBuffer. Looking at the code, it just gets passed whatever `safe` is, so we could make the function generic over a TypeVar bound to `str | Iterable[int]`, and use that for both `safe` and this argument to `quote_via`.

sobolevn: 👍

sobolevn:
@JelleZijlstra friendly ping :)

stdlib/urllib/parse.pyi
JelleZijlstra: This wasn't addressed, the callback still takes a very wide type.

sobolevn: Done, sorry, I've missed this comment.

JelleZijlstra:
Thank you!

PR #9106, opened 2022-11-05
unicodedata: accepts ReadOnlyBuffer

JelleZijlstra:
https://github.com/python/cpython/blob/b5f711185bd11819566068ddf2a74a1402340e2d/Modules/unicodedata.c#L1385


hauntsaninja (approved):
Nice

PR #9107, opened 2022-11-05
unittest: cannot use bytes regexes

JelleZijlstra:
```pycon
>>> from unittest.case import TestCase
>>> c = TestCase()
>>> with c.assertRaisesRegex(Exception, b"x"): 1/0
...
ZeroDivisionError: division by zero

During handling of the above exception, another exception occurred:

  File "/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/unittest/case.py", line 274, in __exit__
    if not expected_regex.search(str(exc_value)):
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: cannot use a bytes pattern on a string-like object
```

### DIFF
--- a/stdlib/unittest/case.pyi
+++ b/stdlib/unittest/case.pyi
@@ -157,7 +157,7 @@ class TestCase:
     def assertRaisesRegex(  # type: ignore[misc]
         self,
         expected_exception: type[BaseException] | tuple[type[BaseException], ...],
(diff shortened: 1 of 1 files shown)


hauntsaninja (approved):

AlexWaygood:
sad, i always wanted to

PR #9108, opened 2022-11-05
winsound: accepts buffers

JelleZijlstra:
https://github.com/python/cpython/blob/b5f711185bd11819566068ddf2a74a1402340e2d/PC/winsound.c#L93


hauntsaninja (approved):
Lg

PR #9109, opened 2022-11-05
zipfile: improve bytes handling

JelleZijlstra:
Technically some uses of `pwd` accept more types, but others
use `isinstance()` to enforce only `bytes`. It seems better to
keep the same type throughout the module.


hauntsaninja (approved):
Looks good!

PR #9110, opened 2022-11-05
xml: improve bytes types

JelleZijlstra:


stdlib/xml/dom/expatbuilder.pyi
```diff
@@ -30,8 +30,8 @@ class ExpatBuilder:
     def getParser(self): ...
     def reset(self) -> None: ...
     def install(self, parser) -> None: ...
-    def parseFile(self, file) -> Document: ...
-    def parseString(self, string: str) -> Document: ...
+    def parseFile(self, file: str | SupportsRead[ReadableBuffer | str]) -> Document: ...
```
hauntsaninja, line 33:
```suggestion
    def parseFile(self, file: SupportsRead[ReadableBuffer | str]) -> Document: ...
```
I don't think this can be str

stdlib/xml/etree/ElementTree.pyi
```diff
@@ -313,9 +313,11 @@ class TreeBuilder:
         def __init__(self, element_factory: _ElementFactory | None = ...) -> None: ...
 
     def close(self) -> Element: ...
-    def data(self, __data: str | bytes) -> None: ...
-    def start(self, __tag: str | bytes, __attrs: dict[str | bytes, str | bytes]) -> Element: ...
-    def end(self, __tag: str | bytes) -> Element: ...
+    def data(self, __data: str) -> None: ...
+    # tag and attrs are passed to the element_factory, so they could be anything
+    # depending on what the particular factory supports.
+    def start(self, __tag: Any, __attrs: dict[Any, Any]) -> Element: ...
```
hauntsaninja, line 319: might be worth keeping the comment, but only changing the type if someone actually complains?

stdlib/xml/dom/expatbuilder.pyi
JelleZijlstra: You're right, it does `file.read`.

stdlib/xml/etree/ElementTree.pyi
JelleZijlstra: The existing type is pretty hard to achieve due to invariance, so I'd rather not keep it.

hauntsaninja (approved):

PR #9112, opened 2022-11-06
`PyInstaller`: Fix `DeprecationWarning` when parsing the stub using `ast.parse()`

AlexWaygood:
```pycon
>>> import ast, warnings
>>> warnings.filterwarnings("always")
>>> with open("typeshed/stubs/pyinstaller/pyi_splash/__init__.pyi", encoding="utf-8") as file:
...     source = file.read()
...
>>> ast.parse(source)
<unknown>:11: DeprecationWarning: invalid escape sequence '\u'
<ast.Module object at 0x0000027EAF6D3AF0>
```


AlexWaygood:
Don't ask me how I came across this

stubs/pyinstaller/pyi_splash/__init__.pyi
```diff
@@ -8,5 +8,5 @@ def is_alive() -> bool: ...
 def update_text(msg: str) -> None: ...
 def close() -> None: ...
 
-CLOSE_CONNECTION: Literal[b"\u0004"]
+CLOSE_CONNECTION: Literal[rb"\u0004"]
```
JelleZijlstra, line 11:
```suggestion
CLOSE_CONNECTION: Literal[b"\x04"]
```
I think what you have would be the wrong byte string. It's just `b'\x04'`:
https://github.com/pyinstaller/pyinstaller/blob/develop/PyInstaller/fake-modules/pyi_splash.py#L69

AlexWaygood: Huh, I'm surprised stubtest didn't pick up on that.

hauntsaninja:
@AlexWaygood stubtest should catch this just fine:
```
λ cat hm.pyi
from typing import Literal

CLOSE_CONNECTION: Literal[rb"\u0004"]
λ cat hm.py 
CLOSE_CONNECTION = b'\x04'
λ stubtest hm
error: hm.CLOSE_CONNECTION variable differs from runtime type Literal[b'\\x04']
Stub: at line 3 in file /Users/shantanu/dev/mypy/hm.pyi
Literal[b'\\\\u0004']
Runtime:
b'\x04'

Found 1 error (checked 1 module)
```
So worth looking into what's gotten broken in typeshed's testing setup

AlexWaygood:
> So I agree, worth looking into what's gotten broken in typeshed's testing setup

Oh, could it be because this file isn't actually "in" the `PyInstaller` package directory — it's in a `pyi_splash` package directory — ?

(I believe `pyi_splash` is a separate package bundled with PyInstaller — @Avasam is that dict?)

JelleZijlstra:
I think `pyi_splash` is only available inside a PyInstaller-bundled executable or something like that. There is a comment about it at the top of the file. It confused me for a while too, I tried to install PyInstaller locally and couldn't find this module.

Avasam:
@AlexWaygood and @JelleZijlstra , you are correct. Feel free to update the comment if you can make it clearer.
`pyi_splash` is a module that's injected in frozen apps. It contains a few methods to manipulate an optional splash-screen. Here's the source code: https://github.com/pyinstaller/pyinstaller/blob/develop/PyInstaller/fake-modules/pyi_splash.py

FWIW, pywin32 does something very similar with it's top-level modules.

JelleZijlstra:
Thanks, I think the comment is very clear, I just had to click a few times to find it from this PR's diff view :)

AlexWaygood:
Thanks @Avasam! It now makes sense that stubtest isn't able to provide much verification for this file :)

PR #9115, opened 2022-11-06
Suggestion: SliceableBuffer type alias

Avasam:
While adding types for the Xlib library, I noticed that it was often slicing what would be `ReadableBuffer`. However `ctypes._CData` and `pickle.PickleBuffer` are not typed as sliceable. Making a sliceable buffer annoying to type and not possible using existing buffer aliases.

For this reason, assuming `ctypes._CData` and `pickle.PickleBuffer` are not simply mistyped, I would like to suggest a new buffer type alias: The Sliceable Buffer.

Not sure if I should split it into a `WriteableSliceableBuffer` and have `WriteableBuffer` use that as part of its union.


hauntsaninja (commented):
It might be the case that `Sequence[int]` is the type you want

Avasam:
Note that `Sequence[int]` would exclude mmap, as it is not a valid Sequence (at least according to pyright)

JelleZijlstra:
`Sequence` is not a protocol and it defines rather a lot of methods, so it's pretty different from what we want here.

It might be a useful exercise to write out what this type would look like post-PEP 688; we can write that in a comment and switch to it when we're ready. Is this what you had in mind:

```python
class SliceableBuffer(Protocol):
    def __buffer__(self, __flags: int) -> memoryview: ...
    def __getitem__(self, __slice: slice) -> Sequence[int]: ...
```

Avasam:
> `Sequence` is not a protocol and it defines rather a lot of methods, so it's pretty different from what we want here.
> 
> It might be a useful exercise to write out what this type would look like post-PEP 688; we can write that in a comment and switch to it when we're ready. Is this what you had in mind:
> 
> ```python
> class SliceableBuffer(Protocol):
>     def __buffer__(self, __flags: int) -> memoryview: ...
>     def __getitem__(self, __slice: slice) -> Sequence[int]: ...
> ```

Sounds about right.

Once again using mmap as an example:
```py
>>> mmap(0, 5)[0:4] 
b'\x00\x00\x00\x00'
>>> mmap(0, 5)[0:4][0]
0
# bytes being a valid sequence of int, this works. Since a sliced mmap returns bytes
```

In my case I would also need a buffer type that I can index (which, if I understand correctly, would look like below). Then post PEP 688 it should be easy for me to use `SliceableBuffer`, `IndexableBuffer`, or a union of both whenever necessary.

```
class SliceableBuffer(Protocol):
    def __buffer__(self, __flags: int) -> memoryview: ...
    def __getitem__(self, __i: int) -> int: ...
```

Avasam:
Example of what it would look like. I realized making a type that supports both slice and index isn't as simple as making a Union (that's either, not both), or extending both base classes (__getitem__ isn't overloaded).  So I added a `SupportsGetItemBuffer` that does both (and __contains__).

stdlib/_typeshed/__init__.pyi
```diff
@@ -236,6 +236,29 @@ else:
 ReadableBuffer: TypeAlias = ReadOnlyBuffer | WriteableBuffer  # stable
 _BufferWithLen: TypeAlias = ReadableBuffer  # not stable  # noqa: Y047
 
+# Anything that implements the read-write buffer interface, and can be sliced/indexed.
+SliceableBuffer: TypeAlias = bytes | bytearray | memoryview | array.array[Any] | mmap.mmap
+IndexableBuffer: TypeAlias = bytes | bytearray | memoryview | array.array[Any] | mmap.mmap
+# https://github.com/python/typeshed/pull/9115#issuecomment-1304905864
+# Post PEP 668, they should be rewritten as such:
```
JelleZijlstra, line 243:
```suggestion
# Post PEP 688, they should be rewritten as such:
```

PR #9116, opened 2022-11-06
Remove `ssl_match_hostname` from pyright exclude

sobolevn:


stubs/backports.ssl_match_hostname/backports/ssl_match_hostname/__init__.pyi
```diff
@@ -1,3 +1,5 @@
+from ssl import _PeerCertRetDictType
```
sobolevn, line 1: I had two options:
1. Replicating the very same complex type
2. Using protected type alias from stdlib

I went with the `2`.

hauntsaninja (approved):
Thanks! Importing the type seems fine, especially for a backport thing. And CI will break if the private type goes away (although users could still end up with version drift).

PR #9120, opened 2022-11-07
Mark `vobject` as complete

sobolevn:
Source: https://github.com/eventable/vobject


AlexWaygood (commented):
I think maybe we should just allowlist all of these things. All of them look like internal implementation details (all hacks to get around Python 2/3 compatibility challenges); none of them look like public API. It feels like it would be weird, and potentially confusing, to have `vobject.base.str_` pop up in autocomplete suggestions.

sobolevn:
Ok, why not :)

AlexWaygood (approved):
Thanks!

PR #9123, opened 2022-11-07
Upgrade mypy to 0.990

AlexWaygood:


JelleZijlstra:
It can wait for another PR but we should turn on the option @hauntsaninja added to disable bytes/bytearray/memoryview promotion in our CI and mypy-primer. I'll look into doing that tonight.

tests/stubtest_allowlists/py3_common.txt
```diff
@@ -438,6 +438,17 @@ types.WrapperDescriptorType.__get__
 typing.[A-Z]\w+
 typing_extensions\..*
 
+# These are abstract properties at runtime,
+# but marking them as such in the stub breaks half the the typed-Python ecosystem
```
AlexWaygood, line 442:
```suggestion
# but marking them as such in the stub breaks half the the typed-Python ecosystem (see #8726)
```

JelleZijlstra (approved):

PR #9130, opened 2022-11-08
Improve `ast` types; revert several "redundant numeric union" changes from #7906

### LINKED ISSUE #9129: Adapt ast types for mypy 0.990
With mypy 0.990 the complex/int/float promotion was changed (see https://github.com/python/mypy/issues/14030).
Therefore checks like `isinstance(node, ast.Num) and isinstance(node.n, int)` fails with strict mode enabled.

The recommendation by mypy is to change complex to `Union[int, float, complex]`. Are there any objections to do so in the ast module?

kasium:
Since mypy  0.990 type promotions was limited.
This means that complex is not longer promoted to int/float, therefore we should adapt the types to list all possible types

Closes #9129


kasium:
Mypy flake8 doesn't like it
```
./stdlib/_ast.pyi:332:12: Y041 Use "complex" instead of "float | complex" (see "The numeric tower" in PEP 484)
```

ilevkivskyi:
Well I guess we can exclude `Y041` (it is not very useful IMO), or if people still like it, you can put a `# noqa: Y041` on those lines.

kasium:
@ilevkivskyi I added a noqa for now and opened an issue at flake8-pyi

AlexWaygood (changes requested):
I'd like to discuss this a little more before we go ahead with this. We may want to do a complete revert of #7906 if we need to go down this route.

.flake8
```diff
@@ -35,7 +37,7 @@ per-file-ignores =
   # Unfortunately, flake8 does not allow to "noqa" just a specific error inside the file itself.
   # https://github.com/PyCQA/flake8/issues/1079
   #     F811 redefinition of unused '...'
-  stdlib/typing.pyi: B, E301, E302, E305, E501, E701, E741, NQA102, F401, F403, F405, F811, F822, Y037
+  stdlib/typing.pyi: B, E301, E302, E305, E501, E701, E741, NQA102, F401, F403, F405, F811, F822, Y037, Y041
```
AlexWaygood, line 40: You'd need to add Y041 to lines 35 and 42 in this file, as well. Currently you're only ignoring Y041 in `stdlib/typing.pyi` specifically, not in any other stubs

ilevkivskyi:
I don't think we need to do a full revert. It needs to be decided on case by case basis. For example, in function argument types using promotions is OK (unless function behaves differently for `int` and `float`).

hauntsaninja (approved):
Changes to stubs look good!

kasium:
Can this please be merged?

kasium:
Hi,
can we please get this merged? Do you know when mypy will pick this up?

AlexWaygood:
Hi @kasium! The discussion in the mypy issue means that it looks like the behaviour change in mypy is about to be reverted: https://github.com/python/mypy/pull/14077. So I don't think there's a massive rush to get this merged :)

Having said that, I'm happy for the changes to the `ast` and `typed-ast` stubs to go in. The typing specs don't precisely specify how type checkers should deal with implicit promotions with respect to attribute annotations. As such, we should err on the side of  fully specified behaviour, and _not_ rely on implicit promotions for these cases over at typeshed. So the changes to the `ast` and `typed-ast` stubs look good to me.

I'd rather fix flake8-pyi first, however, so that it only emits Y041 for parameter annotations. Then we won't need to switch off Y041 in typeshed's `.flake8` config file, and this PR can be focused solely on the changes to the `ast` and `typed-ast` stub. I should be able to create a flake8-pyi fix and release at some point in the next few days.

AlexWaygood:
I've asked if there's any chance of getting a mypy patch release with python/mypy#14077 included: https://github.com/python/mypy/issues/13871#issuecomment-1313250893.

AlexWaygood:
Update: mypy doesn't want to include https://github.com/python/mypy/pull/14077 in a patch release. So let's merge this as a short-term fix.

AlexWaygood (approved):

AlexWaygood:
Thanks for the PR @kasium!

kasium:
Thanks for the fast work!

AlexWaygood:
@kasium — looks like the `ast` fixes will be included in the next mypy patch release, possibly coming out later today: https://github.com/python/mypy/issues/13871#issuecomment-1313680647

PR #9131, opened 2022-11-08
Annotate known magic method return types

sobolevn:
I was pretty careful to include only 100% ones.


srittau (approved):
There might be some cases, where the dunder method does something strange in the implementation, but as long as primer passes, I think this situation is better than the alternative and potential problems can be fixed when they come up.

AlexWaygood (approved):
This is great, thank you!

sobolevn:
Looks like I need to fix metaclasses of many projects.

@srittau can you please trigger `daily` task? I don't have permissions in this repo to do that.

AlexWaygood:
@sobolevn it already failed last night: #9127

PR #9132, opened 2022-11-08
Run stubtest on all third party stubs if requirements.txt is changed

### LINKED ISSUE #9128: Run stubtest on all third party stubs if requirements.txt is changed
See https://github.com/python/typeshed/issues/9127 / this has happened before

sobolevn:
Closes #9128

Let's test it, plan:
1. Now `daily` must not run
2. Next, I will add something to the requirements
3. It must run 🎉 
4. I will add `daily.yml` itself to `paths:`, because when it is changed, we need to validate the correctness of the change

### DIFF
--- a/.github/workflows/stubtest_stdlib.yml
+++ b/.github/workflows/stubtest_stdlib.yml
@@ -10,7 +10,7 @@ on:
       - 'stdlib/**'
       - '.github/workflows/stubtest_stdlib.yml'
       - 'tests/**'
-      - 'requirements-tests.txt'
+      # When requirements change, we run `daily.yml`
 
 permissions:
   contents: read
--- a/.github/workflows/stubtest_stdlib.yml
+++ b/.github/workflows/stubtest_stdlib.yml
@@ -10,7 +10,7 @@ on:
       - 'stdlib/**'
       - '.github/workflows/stubtest_stdlib.yml'
       - 'tests/**'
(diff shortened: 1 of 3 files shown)


sobolevn:
Ok, here we go: changing requirements.

AlexWaygood:
You should remove this line, or we'll get a double run of stdlib stubtest when we tinker with `requirements-tests.txt`:

https://github.com/python/typeshed/blob/4f381af4c5d85f7fd41681231c535da6b5a74f44/.github/workflows/stubtest_stdlib.yml#L13

.github/workflows/stubtest_stdlib.yml
```diff
@@ -10,7 +10,7 @@ on:
       - 'stdlib/**'
       - '.github/workflows/stubtest_stdlib.yml'
       - 'tests/**'
-      - 'requirements-tests.txt'
+      # When requirements change, we run `daily.yml`
```
AlexWaygood, line 13:
```suggestion
      # When requirements.txt changes, we run `daily.yml`, which includes stdlib stubtest
```

AlexWaygood (approved):
Thanks!

sobolevn:
Thanks for the review. Next I propose to merge https://github.com/python/typeshed/pull/8923

I will fix merge conflicts in a moment.

PR #9134, opened 2022-11-08
Use recursive type aliases in builtins and `_typeshed`

AlexWaygood:
Cc. @ilevkivskyi -- is there a risk that this could cause performance regressions for mypy? If so, then it may not be worth it.


AlexWaygood:
> I think it should be OK. Real problem is big unions, not recursive types per se, recursive types just exaggerate problems with big unions.

I was worried about `isinstance()`/`issubclass()`, since they're such fundamental functions :) Thanks for taking a look!

hauntsaninja:
@ilevkivskyi this PR causes mypy to crash with RecursionError when using `--any-exprs-report`, want to take a look? https://github.com/python/mypy/pull/14130

ilevkivskyi:
OK, I can take a look

PR #9135, opened 2022-11-08
add default for simplejson.JSONEncoder.iterencode

asottile-sentry:
reference: https://github.com/simplejson/simplejson/blob/43645252224d4e529c22cb78ac42f241d6426073/simplejson/encoder.py#L304


AlexWaygood (approved):
Thanks!

PR #9136, opened 2022-11-08
Fix and allow classes with missing metaclasses

sobolevn:
Refs https://github.com/python/typeshed/issues/9127
It does not really close this issue, because I still have to fix a lot of metaclasses :)


AlexWaygood (approved):
Thanks! I'm happy to merge this as-is so we can unbreak CI :)

PR #9026, opened 2022-10-29
stripe: test clocks resource

sbdchd:
based off: https://github.com/stripe/stripe-python/blob/16812ae9ff5b78d8f103c09da461e5569890d05f/stripe/api_resources/test_helpers/test_clock.py

and https://stripe.com/docs/api/test_clocks/advance?lang=python

PR #9137, opened 2022-11-09
stripe: fix test clocks imports

sbdchd:
follow up to https://github.com/python/typeshed/pull/9026 which fixes some import related issues so the following now works:

```python
import stripe

stripe.test_helpers.TestClock.create()
```

Before the `test_helpers` module wasn't imported/exported properly


sbdchd:
hmm 
image

srittau (approved):

PR #9139, opened 2022-11-09
All `__nonzero__` are methods that return `bool`

sobolevn:
I know that `__nonzero__` methods are not used anymore, I've even created several PRs to delete them from existing projects.

But, if we still have them in our stubs today, they at least should be correct.


AlexWaygood (approved):

AlexWaygood:
Nearly all of these are `__nonzero__ = __bool__` at runtime, but it doesn't really make any difference in the stub

Issue #9138, opened 2022-11-09
add overload to check for __ceil__ method


programmerjake:
currently `ceil` is defined as:
```python
def ceil(__x: _SupportsFloatOrIndex) -> int: ...
```

`ceil` is documented to call `__x.__ceil__()`, therefore I think an overload should be added to check for that.

example:
```python
import math

class C:
    def __ceil__(self) -> int:
        return 1

print(math.ceil(C()))  # shouldn't error, but does
```

in pyright, I get the error:
```
Argument of type "C" cannot be assigned to parameter "__x" of type "_SupportsFloatOrIndex" in function "ceil"
  "C" is incompatible with protocol "_SupportsFloatOrIndex"
    "__float__" is not present
```

Obviously, all this applies to `round`, `floor`, and `trunc` too.

sobolevn:
Yes, this is right. Here's the C code for this logic:

```c
static PyObject *
math_ceil(PyObject *module, PyObject *number)
{
    if (!PyFloat_CheckExact(number)) {
        math_module_state *state = get_math_module_state(module);
        PyObject *method = _PyObject_LookupSpecial(number, state->str___ceil__);
        if (method != NULL) {
            PyObject *result = _PyObject_CallNoArgs(method);
            Py_DECREF(method);
            return result;
        }
```

PR #9141, opened 2022-11-09
Improve `math.{ceil,floor,trunc}`

sobolevn:
Closes https://github.com/python/typeshed/issues/9138


stdlib/math.pyi
```diff
@@ -26,7 +26,10 @@ def atanh(__x: _SupportsFloatOrIndex) -> float: ...
 if sys.version_info >= (3, 11):
     def cbrt(__x: _SupportsFloatOrIndex) -> float: ...
 
-def ceil(__x: _SupportsFloatOrIndex) -> int: ...
+class _SupportsCeil(Protocol):
+    def __ceil__(self) -> int: ...
+
+def ceil(__x: _SupportsFloatOrIndex | _SupportsCeil) -> int: ...
```
AlexWaygood, line 32: You can implement wacky `__ceil__` methods, and `math.ceil` won't complain:

```pycon
>>> class Foo:
...     def __ceil__(self):
...         return 'foo'
...
>>> f = Foo()
>>> import math
>>> math.ceil(f)
'foo'
```

The docs say that `__ceil__` should generally return an integral, but they don't say that it has to be an `int`. Presumably, it would also be okay to return an instance of `numpy.uint32` or whatever. 

Maybe we should do something like this:

```suggestion
class _SupportsCeil(Protocol[_T_co]):
    def __ceil__(self) -> _T_co: ...

@overload
def ceil(__x: _SupportsCeil[_T]) -> _T: ...
@overload
def ceil(__x: _SupportsFloatOrIndex) -> int: ...
```

sobolevn: Well, this is a classic "documentation" vs "runtime" thing.
It is documented that: https://docs.python.org/3/reference/datamodel.html#object.__ceil__

> Called to implement the built-in function [round()](https://docs.python.org/3/library/functions.html#round) and [math](https://docs.python.org/3/library/math.html#module-math) functions [trunc()](https://docs.python.org/3/library/math.html#math.trunc), [floor()](https://docs.python.org/3/library/math.html#math.floor) and [ceil()](https://docs.python.org/3/library/math.html#math.ceil). Unless ndigits is passed to __round__() all these methods should return the value of the object truncated to an [Integral](https://docs.python.org/3/library/numbers.html#numbers.Integral) (typically an [int](https://docs.python.org/3/library/functions.html#int)).

I went with "the docs path". Do you think non-int cases are common enough?
Maybe some middle ground like `TypeVar("_T", bound=int)`?

JelleZijlstra:
The issue mentions that `floor`, `round`, and `trunc` have a similar problem, would be good to fix those in the same PR if you're up for it.

stdlib/math.pyi
sobolevn: Here are current results:

```
» ag 'def __ceil__'
stdlib/builtins.pyi
285:    def __ceil__(self) -> int: ...
343:        def __ceil__(self) -> int: ...

stdlib/fractions.pyi
130:    def __ceil__(a) -> int: ...

stdlib/_decimal.pyi
118:    def __ceil__(self) -> int: ...

stdlib/numbers.pyi
60:    def __ceil__(self) -> int: ...
```

sobolevn: Ok, sounds fair! I hope people won't overuse this great power 🪄

stdlib/builtins.pyi
```diff
@@ -1619,12 +1619,7 @@ class reversed(Iterator[_T], Generic[_T]):
     def __length_hint__(self) -> int: ...
 
 def repr(__obj: object) -> str: ...
-@overload
-def round(number: SupportsRound[Any]) -> int: ...
-@overload
-def round(number: SupportsRound[Any], ndigits: None) -> int: ...
-@overload
-def round(number: SupportsRound[_T], ndigits: SupportsIndex) -> _T: ...
+def round(number: SupportsRound[_T], ndigits: SupportsIndex | None = ...) -> _T: ...
```
AlexWaygood, line 1622: This needs to be an overloaded method so that mypy can infer that `round(3.14) -> int`, but `round(3.14, ndigits=1) -> float`. That's causing the vast majority of the primer hits.

sobolevn: Again, first arg can be like this:

```python
>>> class R:
...       def __round__(self): 
...           return 'a'
... 
>>> round(R())
'a'
```

JelleZijlstra: The implementation does indeed just return whatever `__round__` returns. Maybe we need two protocols, one for zero-argument `__round__` and one for one-argument  `__round__`.

sobolevn: Just to be clear, two argument `round` is the same: it can return anything.

```python
>>> class R2:
...    def __round__(self, n):
...       return 'a'
... 
>>> round(R2(), 5)
'a'
```

programmerjake:
note that you can pass arbitrary types for both arguments of `round`, the arguments are passed through unmodified to `__round__`, so imho the second argument shouldn't be restricted to `int` or `SupportsIndex`

sobolevn:
Primer analysis:

- `homeassistant/components/nissan_leaf/sensor.py:129` is correct, https://github.com/home-assistant/core/blob/874ece195e6064bd8f4585e7b936cd5ee1c3f673/homeassistant/components/nissan_leaf/sensor.py#L119 has type `Any`, because `Leaf` is imported from untyped lib
- `homeassistant/components/elgato/light.py:95` is correct, it returns `Any`, because `.brigtness` is `Any` (imported from untyped lib): https://github.com/home-assistant/core/blob/874ece195e6064bd8f4585e7b936cd5ee1c3f673/homeassistant/components/elgato/light.py#L95
- `homeassistant/components/demo/cover.py:199` is not correct
- `homeassistant/components/demo/cover.py:211` is not correct

sobolevn:
Some more cases:

- `homeassistant/components/fritzbox/light.py:157` is correct, because `round` withs with `Any` here: https://github.com/home-assistant/core/blob/9bd676aff644aab1419458f80ace8bc11a806b8a/homeassistant/components/fritzbox/light.py#L140
- `homeassistant/components/accuweather/sensor.py:256` is correct, because it works with `Any`: https://github.com/home-assistant/core/blob/9bd676aff644aab1419458f80ace8bc11a806b8a/homeassistant/components/accuweather/sensor.py#L58

sobolevn:
I consider this complete 🎉

AlexWaygood:
The changes to the `math` functions look great. The changes to `builtins.round` look interesting, but more complicated.

Could we possibly split the changes to `builtins.round()` off into a separate PR? (I know this is contrary to what @JelleZijlstra said earlier, sorry to be annoying!)

AlexWaygood (approved):

JelleZijlstra:
> Could we possibly split the changes to `builtins.round()` off into a separate PR? (I know this is contrary to what @JelleZijlstra said earlier, sorry to be annoying!)

Good call, when I wrote that I thought there was a separate `math.round`.

PR #9148, opened 2022-11-10
Remove `sqlite3/dbapi2.pyi` from pyright exclude

sobolevn:


JelleZijlstra:
Seems like we need a few more annotations

sobolevn:
I went with `Incomplete`, because there are a lot of C logic involved. I don't want to really waste a lot of time precisely annotating outdated APIs :(

AlexWaygood (approved):
In general I'm not really a fan of removing items from the excludelist if it means having to add loads of `Incomplete` annotations. But since these classes only exist on Python <3.8, this seems like a net win in this case

stdlib/sqlite3/dbapi2.pyi
```diff
@@ -227,9 +227,9 @@ else:
 
 if sys.version_info < (3, 8):
     class Cache:
-        def __init__(self, *args, **kwargs) -> None: ...
-        def display(self, *args, **kwargs) -> None: ...
-        def get(self, *args, **kwargs) -> None: ...
+        def __init__(self, *args: Incomplete, **kwargs: object) -> None: ...
```
sobolevn, line 230: Here's `kwargs` is `object` because `kwargs` is not used in `Cache` constructor: https://github.com/python/cpython/blob/b0b590be9597fd5919228d251812dd54145f70a7/Modules/_sqlite/cache.c#L57-L87

PR #9151, opened 2022-11-10
Refactor `round` and add tests

sobolevn:
Refs #9141


stdlib/builtins.pyi
```diff
@@ -1619,12 +1618,27 @@ class reversed(Iterator[_T], Generic[_T]):
...
+def round(number: int, ndigits: SupportsIndex) -> int: ...
 @overload
-def round(number: SupportsRound[Any]) -> int: ...
+def round(number: float, ndigits: SupportsIndex) -> float: ...
 @overload
-def round(number: SupportsRound[Any], ndigits: None) -> int: ...
+def round(number: _SupportsRound1[_T], ndigits: None = ...) -> _T: ...
 @overload
-def round(number: SupportsRound[_T], ndigits: SupportsIndex) -> _T: ...
+def round(number: _SupportsRound2[_T], ndigits: SupportsIndex) -> _T: ...
```
programmerjake, line 1641: note that you can pass arbitrary types for both arguments of round, the arguments are passed through unmodified to __round__, so imho the second argument shouldn't be restricted to int or SupportsIndex
see https://github.com/python/cpython/blob/231d83b72435d61e929d6d62f2be7305e7b5b62b/Python/bltinmodule.c#L2305-L2331

stdlib/builtins.pyi
```diff
@@ -1619,12 +1618,27 @@ class reversed(Iterator[_T], Generic[_T]):
...
+
+class _SupportsRound1(Protocol[_T_co]):
+    def __round__(self) -> _T_co: ...
+
+class _SupportsRound2(Protocol[_T_co]):
+    def __round__(self, __ndigits: int) -> _T_co: ...
+
+# We need these overloads to replicate how stdlib works
+@overload
+def round(number: float, ndigits: None = ...) -> int: ...
```
JelleZijlstra, line 1633: I haven't looked at the code deeply, but I feel like we shouldn't need these if `float.__round__` is defined correctly in our stubs. I believe the C code for `builtins.round` doesn't actually treat `float` specially.

stdlib/builtins.pyi
sobolevn: Technically, yes. But, again: this is quite against its main purpose: `round(1.55, 'a')` must be reported. This is a clear error.

I don't think people are basing their API around non-`int` arguments to `round`

AlexWaygood: Agree with @sobolevn. Practicality beats purity here; it doesn't serve users well to allow a `str` argument (or whatever) to `ndigits`

programmerjake: `round(1.55, 'a')` would be a type error because `float.__round__` doesn't allow a string. imho there are valid use-cases for non-int types, e.g. passing an object that is a symbolic variable (like how sympy has symbolic variables): `repr(round(Symbol('x'), Symbol('y'))) == 'round(x, y)'`. imho it should be more like:
```python
class _SupportsRound2(Protocol[_T, _R]): # add co/contravariance as necessary
    def __round__(self, __ndigits: _T) -> _R: ...
@overload
def round(__x: _SupportsRound2[_T, _R], ndigits: _T) -> _R: ...
```

AlexWaygood: Interesting. Would the `Symbol` type in that example not have an `__index__` method @programmerjake? (I haven't really used sympy.)

AlexWaygood: The [docs for `round()`](https://docs.python.org/3/library/functions.html#round) state:

> Any integer value is valid for _ndigits_ (positive, zero, or negative).

`SupportsIndex` reflects that quite well — it allows `int`, but also allows other alternative integer types that might exist, as long as they have an `__index__` method

programmerjake: note, idk if sympy actually does that, i just want it to be valid to write a `Symbol` class that can do that. `Symbol.__index__` if it existed would probably return a symbolic expression, not an `int`.

AlexWaygood: I'm sympathetic, but I still think we probably shouldn't support it since, according to the Python docs, `round()` just isn't meant to be used with a non-integer passed to the second argument :)

programmerjake: > The [docs for `round()`](https://docs.python.org/3/library/functions.html#round) state:
> 
> > Any integer value is valid for _ndigits_ (positive, zero, or negative).

i interpret that to be for builtin self types only, since it's in the paragraph starting with "For the built-in types supporting round()". It states in a separate paragraph: "For a general Python object number, round delegates to number.__round__" -- imho this implies that `__round__`'s type should be used, rather than forcing integer-like types (SupportsIndex).

AlexWaygood: I see what you mean, but it still says: "For a general Python object _number_" — implying `round()` is generally meant to be used with numeric types. Would you consider `Symbol` in your example to be a "numeric" type?

sobolevn: I've just tested proposed `_SupportsRound2[_T, _T_co]` protocol. It does not work, tests fail:

```
test_cases/stdlib/builtins/check_round.py:13: error: Expression is of type "Any", not "float"  [assert-type]
test_cases/stdlib/builtins/check_round.py:13: error: Cannot infer type argument 1 of "round"  [misc]
test_cases/stdlib/builtins/check_round.py:14: error: Expression is of type "Any", not "float"  [assert-type]
test_cases/stdlib/builtins/check_round.py:14: error: Cannot infer type argument 1 of "round"  [misc]
test_cases/stdlib/builtins/check_round.py:15: error: Expression is of type "Any", not "float"  [assert-type]
test_cases/stdlib/builtins/check_round.py:15: error: Cannot infer type argument 1 of "round"  [misc]
test_cases/stdlib/builtins/check_round.py:16: error: Expression is of type "Any", not "float"  [assert-type]
test_cases/stdlib/builtins/check_round.py:16: error: Cannot infer type argument 1 of "round"  [misc]
test_cases/stdlib/builtins/check_round.py:46: error: Expression is of type "Any", not "str"  [assert-type]
test_cases/stdlib/builtins/check_round.py:46: error: Cannot infer type argument 1 of "round"  [misc]
```

Source:

```python
class _SupportsRound1(Protocol[_T_co]):
    def __round__(self) -> _T_co: ...

class _SupportsRound2(Protocol[_T, _T_co]):
    def __round__(self, __ndigits: _T) -> _T_co: ...

@overload
def round(number: _SupportsRound1[_T], ndigits: None = ...) -> _T: ...
@overload
def round(number: _SupportsRound2[_S, _T], ndigits: _S) -> _T: ...
```

programmerjake: > I see what you mean, but it still says: "For a general Python object _number_" (emphasis mine) — implying `round()` is generally meant to be used with numeric types. Would you consider `Symbol` in your example to be a "numeric" type?

yes, because it represents an arbitrary-ish number. an example where round is used with a symbol: https://www.wolframalpha.com/input?i=round%28x%29-floor%28x%29

programmerjake: > I've just tested proposed `_SupportsRound2[_T, _T_co]` protocol. It does not work, tests fail:

i tried some variations...it appears as though neither mypy nor pyright can correctly deduce the types, their type solvers are just not powerful enough...i guess I'll have to give up for now.

JelleZijlstra:
Took a look at a few of the mypy-primer results and they seem right. homeassistant is calling round() on a bunch of numbers coming from untyped third-party packages. This change makes `round(Any)` return `Any`, so no `mypy --warn-return-any` complains.

> 
> ```diff
> + homeassistant/components/dsmr/sensor.py:572: error: Incompatible types in assignment (expression has type "int", variable has type "Optional[str]")  [assignment]
Seems correct. The variable is changed to a new type. (https://github.com/home-assistant/core/blob/6517e3e21e5828a317ea8b814d0345f2536b9170/homeassistant/components/dsmr/sensor.py#L572)

> + homeassistant/components/dsmr/sensor.py:579: error: Statement is unreachable  [unreachable]
Also correct. The value can never be None at this point.

> + homeassistant/components/nissan_leaf/sensor.py:129: error: Returning Any from function declared to return "Optional[float]"  [no-any-return]
Haven't fully tracked this but seems probably right. https://github.com/home-assistant/core/blob/6517e3e21e5828a317ea8b814d0345f2536b9170/homeassistant/components/nissan_leaf/sensor.py#L129

> + homeassistant/components/elgato/light.py:95: error: Returning Any from function declared to return "Optional[int]"  [no-any-return]
Seems right, `elgato` probably has no types. https://github.com/home-assistant/core/blob/6517e3e21e5828a317ea8b814d0345f2536b9170/homeassistant/components/elgato/light.py#L95

> + homeassistant/components/tradfri/sensor.py:67: error: Returning Any from function declared to return "int"  [no-any-return]
> + homeassistant/components/tplink/light.py:272: error: Returning Any from function declared to return "Optional[int]"  [no-any-return]
> + homeassistant/components/fritzbox/light.py:157: error: Returning Any from function declared to return "Union[SupportsDunderLT[Any], SupportsDunderGT[Any]]"  [no-any-return]
> + homeassistant/components/shelly/light.py:198: error: Returning Any from function declared to return "int"  [no-any-return]
> + homeassistant/components/accuweather/sensor.py:256: error: Returning Any from function declared to return "Union[None, str, int, float]"  [no-any-return]
> + homeassistant/components/esphome/light.py:267: error: Returning Any from function declared to return "Optional[int]"  [no-any-return]
> + homeassistant/components/esphome/cover.py:93: error: Returning Any from function declared to return "Optional[int]"  [no-any-return]
> + homeassistant/components/esphome/cover.py:101: error: Returning Any from function declared to return "Optional[int]"  [no-any-return]
> + homeassistant/components/zwave_js/light.py:181: error: Returning Any from function declared to return "Optional[int]"  [no-any-return]
> + homeassistant/components/zwave_js/cover.py:135: error: Returning Any from function declared to return "Optional[int]"  [no-any-return]
> + homeassistant/components/isy994/light.py:73: error: Returning Any from function declared to return "Optional[int]"  [no-any-return]
> + homeassistant/components/isy994/cover.py:61: error: Returning Any from function declared to return "Optional[int]"  [no-any-return]
> ```

AlexWaygood (commented):
Can you add a test for a custom class that has an overloaded `__round__` method, similar to `builtins.float`?

test_cases/stdlib/builtins/check_round.py
```diff
@@ -0,0 +1,68 @@
+from typing import overload
```
AlexWaygood, line 1:
```suggestion
from __future__ import annotations
from typing import overload
```

JelleZijlstra (approved):

test_cases/stdlib/builtins/check_round.py
```diff
@@ -0,0 +1,70 @@
...
+assert_type(round(WithCustomRound2(), 1), str)
+assert_type(round(WithCustomRound2(), CustomIndex()), str)
+# Errors:
+round(WithCustomRound2(), None)  # type: ignore
+round(WithCustomRound2())  # type: ignore
+
+
+class WithOverloadedRound:
+    @overload
+    def __round__(self) -> str:
```
AlexWaygood, line 57:
```suggestion
    def __round__(self, ndigits: None = ...) -> str:
```

AlexWaygood (approved):
Looks good!

PR #9152, opened 2022-11-10
Fix usage of byte2int with bytes

### LINKED ISSUE #9145: Issue with annotation six.byte2int starting 1.16.21.2
Hi,
Starting 1.16.21.2 we have an issue with mypy failing for annotations of `six.byte2int`.
We had this method:

```python 
def byte2int(c: Union[bytes, int]) -> int:
    if isinstance(c, bytes):
        return six.byte2int(c)
    return c
```

which was called the following way:
`pp_fingerprint = ":".join("{:02x}".format(byte2int(c)) for c in fingerprint).upper()`

This used to work until now, but now we get this failure:
`Argument 1 to "byte2int" has incompatible type "bytes"; expected "SupportsGetItem[int, <nothing>]"
`
Is this an issue with this release? Or are we supposed to fix our annotations in some way?

Thanks in advance!

Avasam:
Fixes #9145 . Restores functionality specifically with bytes.
Works around a mypy bug that thinks `SupportsGetItem[int, <nothing>]` is expected when passed bytes. ~~May be https://github.com/python/mypy/issues/14032~~

`indexbytes` seems unnafected


Avasam:
I was going to but I can't replicate it anymore O_o

```py
from typing import Any, Protocol, Union

from _typeshed import _T, _KT_contra, _VT_co

class SupportsGetItem(Protocol[_KT_contra, _VT_co]):
    def __contains__(self, __x: Any) -> bool:
        ...

    def __getitem__(self, __key: _KT_contra) -> _VT_co:
        ...

###

def foo(obj: SupportsGetItem[int, _T]) -> _T:
    return obj[0]

bar: int = foo(b"1")

###

import six

six.byte2int(b"1")

###

def byte2int(c: Union[bytes, int]) -> int:
    if isinstance(c, bytes):
        return six.byte2int(c)
    return c


bbyte2int(b"1")
```

mypy 0.990 (compiled: yes)
Python 3.9.13

JelleZijlstra:
Interesting, different mypy version?

Avasam:
Actually it differs whether I point to my local typeshed repo or not... Even if I revert the change to `byte2int`.

The other culprit could be `SupportsGetItem`. In my example above, I redefined it, and it's the same in my local repo. But the typeshed stubs shipped with mypy still has `def __contains__(self, __x: object) -> bool: ...`.

Avasam:
For the record, having to use `Any` instead of `object` or `_VT_co` sounds an awful lot like the comment in `Mapping` which points to https://github.com/python/typing/pull/273

AlexWaygood (commented):
A better solution might be to just copy the new version of `SupportsGetItem` into the stubs for `six`, and then use that directly instead of importing it from `_typeshed` (with a TODO comment saying we should switch to the `_typeshed` version once mypy updates its vendored copy of typeshed and makes a new release). Then we wouldn't have to make the function an overloaded function.

AlexWaygood (approved):
Thanks @Avasam!

PR #9154, opened 2022-11-10
fix tqdm: allow `disable=None` in `tqdm.__init__`

tconbeer:
similar to #8603

According to the [docstring](https://github.com/tqdm/tqdm/blob/6791e8c5b3d6c30bdd2060c346996bfb5a6f10d1/tqdm/std.py#L901), you can initialize tqdm with `disable=None`, which provides different behavior than True or False. This PR updates the types of the arguments to `__init__` accordingly.

All other flavors of the progress bar (async, notebook, etc.) are derived from the base `tqdm` class and initialize with `super`, so accept the same type.


JelleZijlstra (approved):

hauntsaninja (approved):

tconbeer:
You guys are fast! Thanks so much!

PR #9157, opened 2022-11-11
PyYAML: Add missing types to _N TypeVar

arareko:
In https://github.com/python/typeshed/blob/294b03f75bd23d95cf4709ba14547cc4190aacdd/stubs/PyYAML/yaml/constructor.pyi the signature for `construct_sequence()` defines  `SequenceNode` as the type for **node** (which is correct), however the `_N` TypeVar used in the signature for `add_constructor()` only allows for `Node`.

It would seem that `MappingNode`, `ScalarNode` and `SequenceNode` are missing from the TypeVar definition since `add_constructor()` does not consider any of them as compatible. For example:

```bash
error: Argument 2 to "add_constructor" has incompatible type "Callable[[Loader, SequenceNode], str]"; expected "Callable[[Union[Loader, FullLoader, UnsafeLoader], Node], Any]"
```

This PR addresses the above.

### DIFF
--- a/stubs/PyYAML/yaml/constructor.pyi
+++ b/stubs/PyYAML/yaml/constructor.pyi
@@ -9,7 +9,7 @@ from yaml.loader import BaseLoader, FullLoader, Loader, SafeLoader, UnsafeLoader
 from yaml.nodes import MappingNode, Node, ScalarNode, SequenceNode
 
 _L = TypeVar("_L", bound=Loader | BaseLoader | FullLoader | SafeLoader | UnsafeLoader)
-_N = TypeVar("_N", bound=Node)
+_N = TypeVar("_N", bound=Node | MappingNode | ScalarNode | SequenceNode)
 
 _Scalar: TypeAlias = str | int | float | bool | None
 


JelleZijlstra:
These are all subclasses of `Node`, so adding them to the bound shouldn't do anything. Can you share a code sample on which you are getting an error?

arareko:
@JelleZijlstra sure thing:

Save this as `example.py`:
```
import os

import yaml


def load_config(config_path: str) -> dict:
    # define custom tag handler
    def _join(loader: yaml.Loader, node: yaml.SequenceNode) -> str:
        seq = loader.construct_sequence(node)
        return os.path.join(*[str(i) for i in seq])

    # register the tag handler
    yaml.add_constructor("!join", _join)

    # load config
    with open(config_path, "r") as ymlfile:
        config = yaml.load(ymlfile, Loader=yaml.Loader)

    return config

```

Then check with `mypy`:
```
❯ mypy example.py
example.py:13: error: Argument 2 to "add_constructor" has incompatible type "Callable[[Loader, SequenceNode], str]"; expected "Callable[[Union[Loader, FullLoader, UnsafeLoader], Node], Any]"  [arg-type]
```

JelleZijlstra:
Thanks for the code sample! Our stub for `add_constructor` is

```
@overload
def add_constructor(
    tag: str, constructor: Callable[[Loader | FullLoader | UnsafeLoader, Node], Any], Loader: None = ...
) -> None: ...
@overload
def add_constructor(tag: str, constructor: Callable[[_Constructor, Node], Any], Loader: type[_Constructor]) -> None: ...
```
`_N` isn't involved, so your PR is unlikely to help.

Maybe the `Node` type in the second argument to the constructor should be something else, but I'd have to look at the yaml code to verify that.

arareko:
@JelleZijlstra I see what you mean. So basically, it won't work because what is defined in the `constructor.pyi` stub is being overloaded in `__init__.pyi`.

arareko:
@JelleZijlstra `add_constructor()` in https://github.com/yaml/pyyaml/blob/957ae4d495cf8fcb5475c6c2f1bce801096b68a5/lib/yaml/__init__.py#L302 supports `Loader`, `FullLoader` and `UnsafeLoader`, which are in sync with the `__init__.pyi` stub.

Those loaders make use of `Constructor` or `FullConstructor`, which inherit from [`BaseConstructor`](https://github.com/yaml/pyyaml/blob/957ae4d495cf8fcb5475c6c2f1bce801096b68a5/lib/yaml/constructor.py#L19) and which handles `MappingNode`, `ScalarNode` and `SequenceNode` and not `Node` per se. Those three node types actually are subclasses of `Node`.

This would suggest that the allowed node types for `add_constructor()` in the stub should actually be:
```python
@overload
def add_constructor(
    tag: str, constructor: Callable[[Loader | FullLoader | UnsafeLoader, MappingNode | ScalarNode | SequenceNode], Any], Loader: None = ...
) -> None: ...
@overload
def add_constructor(tag: str, constructor: Callable[[_Constructor, MappingNode | ScalarNode | SequenceNode], Any], Loader: type[_Constructor]) -> None: ...
```

What do you think?

JelleZijlstra:
These proposed types imply that the constructor callable must accept all of `MappingNode`, `ScalarNode`, and `SequenceNode` (because Callable is contravariant in the parameter types). So if nothing else, that doesn't seem like it would make your original code typecheck.

AlexWaygood:
I haven't looked into how to fix your problem, but I concur with @JelleZijlstra that this PR won't fix anything, unfortunately. The proposed union would be redundant, since `MappingNode`, `ScalarNode`, `SequenceNode` are all subclasses of `Node`.

If the problem is still present, please open an issue, and we can try to figure out how to fix it.

arareko:
@AlexWaygood I've been intending to look into this issue further and potentially finding the right fix, however, life and work have gotten in the way. I'll probably re-open this whenever I manage to work on it.

AlexWaygood:
@arareko no worries! Feel free to reopen, or open a new issue/PR at any time :)

PR #8256, opened 2022-07-07
Third-party stubs: enforce CamelCase for type alias names

AlexWaygood:
Unblocks https://github.com/PyCQA/flake8-pyi/pull/243

PR #9158, opened 2022-11-11
Use `TypeAlias` for types in `sqlite3.dbapi2`

sobolevn:


sobolevn:
Btw, I've found *many* more missing type aliases in other modules via my custom query.

Are you interested in annotated them?

AlexWaygood:
What benefits does this PR bring?

There are some differing opinions on exactly when `TypeAlias` should be used in stubs -- e.g. https://github.com/python/typeshed/pull/8256#issuecomment-1177966512

sobolevn:
I've just seen @srittau's comment on his vision of `TypeAlias`. Mine is completely different.

I see `TypeAlias` as a way to say: these things are both types and they are identical.

Consider these lines:

```python
x = 1
def some(): ...
class T: ...

y = x
other = some
T1 = T
T2: TypeAlias = T
```

Here we can notice that `y`,` other`, and `T1` are assignments. But, we cannot be sure what their values are (until we go and see, it is even more complicated when different name schemes are used like `error = _SomeError`).

But, `TypeAlias` gives us strong guarantee: these two are just different names for a single type.

But, feel free to close if this is not something you want, I don't have a strong opinion.

JelleZijlstra:
I'd also be hesitant to make a change here unless there's a clear problem being fixed.

AlexWaygood:
Let's close this for now

PR #9159, opened 2022-11-11
Fix CI failure when no third-party stubs were updated

sobolevn:
Testing the CI after https://github.com/python/typeshed/actions/runs/3445520575/jobs/5749332275


AlexWaygood:
One weird trick to get CI to pass: make sure you update at least one line of code in the `stubs/` directory in every PR

sobolevn:
It should be `{ egrep '^stubs/' || echo '' }`

sobolevn:
So, github tried to substitute `{ egrep '^stubs/' || echo '' }` with some actions specific logic. I had to use function instead.

JelleZijlstra (approved):
Using a function is a good idea, thanks!

PR #9160, opened 2022-11-11
Do not run `stubtest_third_party` on requirements change

sobolevn:
Similar to https://github.com/python/typeshed/blob/main/.github/workflows/stubtest_stdlib.yml

### DIFF
--- a/.github/workflows/stubtest_third_party.yml
+++ b/.github/workflows/stubtest_third_party.yml
@@ -5,6 +5,8 @@ on:
     paths-ignore:
       - '**/*.md'
       - 'scripts/**'
+      # When requirements.txt changes, we run `daily.yml`, which includes third-party stubtest
+      - 'requirements-tests.txt'
 
 permissions:
   contents: read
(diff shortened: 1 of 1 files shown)


sobolevn:
Right now on `requirements-tests.txt` change it will run all the machinery, find 0 stubs change and exit with `0` code.

.github/workflows/stubtest_third_party.yml
```diff
@@ -5,6 +5,8 @@ on:
     paths-ignore:
       - '**/*.md'
       - 'scripts/**'
+      # When requirements.txt changes, we run `daily.yml`, which includes third-party stubtest
+      - 'requirements-tests.txt'
```
AlexWaygood, line 9: Why don't we just do this, now that (following #8923), this job now has its own workflow file?

```suggestion
    paths:
      - 'stubs/**'
      - '.github/workflows/stubtest_third_party.yml'
      - 'tests/**'
      # When requirements.txt changes, we run `daily.yml`, which includes third-party stubtest
```

sobolevn:
Yes, this is even better!

AlexWaygood (approved):
Thanks!

PR #9162, opened 2022-11-11
Add None as a valid type in SAXParseException's second argument

### LINKED ISSUE #9161: Second argument to xml.sax.SAXParseException.__init__ incorrectly omits union with None
This just bit me: when creating an `xml.sax.SAXParseException`, the second argument is allowed to be `None`, but this isn't reflected in the typings available here.

[Current typeshed typing for `xml.sax.SAXParseException`](https://github.com/python/typeshed/blob/70c4af4b73faf1b1c2fe9e2292c6c6da7f9614c4/stdlib/xml/sax/__init__.pyi#L17-L18):

```python
class SAXParseException(SAXException):
    def __init__(self, msg: str, exception: Exception, locator: Locator) -> None: ...
```

You can see in the CPython source, however, that `__init__` does nothing with the `exception` argument other than pass it up to its superclass (`SAXException`), which *does* allow it to be `None`.

[Current typeshed typing for `xml.sax.SAXException`](https://github.com/python/typeshed/blob/70c4af4b73faf1b1c2fe9e2292c6c6da7f9614c4/stdlib/xml/sax/__init__.pyi#L11-L12):

```python
class SAXException(Exception):
    def __init__(self, msg: str, exception: Exception | None = ...) -> None: ...
```

The CPython source for `xml.sax.SAXParseException.__init__` is identical in [3.7.15](https://github.com/python/cpython/blob/9de83ce9704ba065626f483abf6c21b7b829104e/Lib/xml/sax/_exceptions.py#L59-L70), [3.8.15](https://github.com/python/cpython/blob/44adf8a80a6e27a394094c99ec0fba8e7124bd7e/Lib/xml/sax/_exceptions.py#L59-L70), [3.9.15](https://github.com/python/cpython/blob/7e2815419663a2f82abf6c5e1f848fd1de7c8033/Lib/xml/sax/_exceptions.py#L59-L70), [3.10.8](https://github.com/python/cpython/blob/aaaf5174241496afca7ce4d4584570190ff972fe/Lib/xml/sax/_exceptions.py#L59-L70), and [3.11.0](https://github.com/python/cpython/blob/deaf509e8fc6e0363bd6f26d52ad42f976ec42f2/Lib/xml/sax/_exceptions.py#L59-L70):

```python
    def __init__(self, msg, exception, locator):
        "Creates the exception. The exception parameter is allowed to be None."
        SAXException.__init__(self, msg, exception)
        self._locator = locator


        # We need to cache this stuff at construction time.
        # If this exception is raised, the objects through which we must
        # traverse to get this information may be deleted by the time
        # it gets caught.
        self._systemId = self._locator.getSystemId()
        self._colnum = self._locator.getColumnNumber()
        self._linenum = self._locator.getLineNumber()
```

benblank:
This closes #9161 by adding the omitted union with `None`.


hauntsaninja (approved):
Thank you!

PR #9163, opened 2022-11-11
Bump paramiko to 2.12.*

sobolevn:
Docs: https://github.com/paramiko/paramiko/blob/bbb8d57e917f5a9fefd25a85f3d10cd0c1a52ef5/paramiko/client.py#L318-L323
Closes https://github.com/python/typeshed/pull/9093

### DIFF
--- a/stubs/paramiko/paramiko/client.pyi
+++ b/stubs/paramiko/paramiko/client.pyi
@@ -12,7 +12,7 @@ from .transport import _SocketLike
 
 class _TransportFactory(Protocol):
     def __call__(
-        self, sock: _SocketLike, gss_kex: bool, gss_deleg_creds: bool, disabled_algorithms: dict[str, Iterable[str]] | None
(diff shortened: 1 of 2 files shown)


stubs/paramiko/paramiko/client.pyi
```diff
@@ -10,6 +10,11 @@ from paramiko.util import ClosingContextManager
 
 from .transport import _SocketLike
 
+class _TransportFactory(Protocol):
+    def __call__(
+        self, sock: _SocketLike, gss_kex: bool, gss_deleg_creds: bool, disabled_algorithms: dict[str, Iterable[str]] | None
```
JelleZijlstra, line 15:
```suggestion
        self, __sock: _SocketLike, *, gss_kex: bool, gss_deleg_creds: bool, disabled_algorithms: dict[str, Iterable[str]] | None
```
This param is passed positionally, the others as keywords.

sobolevn: I copied args from `Transport.__init__`. But yes, this one can be pos-only in the callback.

JelleZijlstra (approved):

PR #9165, opened 2022-11-12
marshal: specify allowed types

JelleZijlstra:
See w_object and w_complex_object in https://github.com/python/cpython/blob/main/Python/marshal.c

### DIFF
--- a/stdlib/marshal.pyi
+++ b/stdlib/marshal.pyi
@@ -7,23 +7,23 @@ from typing_extensions import TypeAlias
 version: int
 
 _Marshallable: TypeAlias = Union[
-    int
-    , float
-    , complex
-    , bytes
-    , str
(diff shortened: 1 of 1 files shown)


hauntsaninja (commented):
NotImplemented is not marshallable. Also why list bytes explicitly, but not bytearray?
nit: Could also be nice to order it the same as in the documentation

JelleZijlstra:
> NotImplemented is not marshallable. 

Oops, not sure where I got that from.

> Also why list bytes explicitly, but not bytearray? 

bytes has its own code path, bytearray doesn't.

> nit: Could also be nice to order it the same as in the documentation

I ordered them the way they are handled in the code, since I believe the code more than the documentation.

hauntsaninja (approved):
Looks great, thanks!

PR #9166, opened 2022-11-12
xmlrpc: improve bytes handling

JelleZijlstra:


stdlib/xmlrpc/client.pyi
```diff
@@ -15,7 +15,19 @@ class _SupportsTimeTuple(Protocol):
...
+    bool
+    | int
+    | float
+    | str
+    | bytes
+    | bytearray
+    | None
+    | tuple[Any, ...]
+    | list[Any]
+    | dict[Any, Any]
```
AlexWaygood, line 27: There's a description of marshallable types here, below the paragraph that starts with "Types that are conformable" -- it seems it should be a recursive type: https://docs.python.org/3/library/xmlrpc.client.html#xmlrpc.client.ServerProxy

```suggestion
    | tuple[_Marshallable, ...]
    | list[_Marshallable]
    | dict[str, _Marshallable]
```

AlexWaygood (commented):
This all looks correct, though the mypy_primer hit makes me pause slightly, given that the `Transport` class is explicitly advertised in the docs as being one designed for subclassing. But pip is already doing `if TYPE_CHECKING` imports, so maybe they won't mind too much about having to import `ReadableBuffer` from `_typeshed`. (The error message would be a lot nicer if https://github.com/python/mypy/pull/12835 were merged.)

stdlib/xmlrpc/client.pyi
JelleZijlstra: Using that exact type would be impractical because of invariance. However, we can change tuple, and you're right that the dict keys must be str.
```suggestion
    | tuple[_Marshallable, ...]
    | list[Any]
    | dict[str, Any]
```

AlexWaygood (approved):
Happy for this to be merged, with my slight reservation about ease-of-subclassing

stdlib/xmlrpc/client.pyi
```diff
@@ -15,7 +15,19 @@ class _SupportsTimeTuple(Protocol):
 
 _DateTimeComparable: TypeAlias = DateTime | datetime | str | _SupportsTimeTuple
 _Marshallable: TypeAlias = (
```
AlexWaygood, line 17:
```suggestion
# Ideally this would be a recursive type alias, but dict/list invariance makes that impractical
_Marshallable: TypeAlias = (
```

stdlib/xmlrpc/client.pyi
```diff
@@ -15,7 +15,19 @@ class _SupportsTimeTuple(Protocol):
...
-    bool | int | float | str | bytes | None | tuple[Any, ...] | list[Any] | dict[Any, Any] | datetime | DateTime | Binary
+    bool
+    | int
+    | float
+    | str
+    | bytes
+    | bytearray
+    | None
+    | tuple[_Marshallable, ...]
+    | list[Any]
```
JelleZijlstra, line 26:
```suggestion
    # Ideally we'd use _Marshallable for list and dict, but invariance makes that impractical
    | list[Any]
```

stdlib/xmlrpc/client.pyi
JelleZijlstra: That's not quite right because the alias *is* recursive (what is "fully recursive"?), so I put a more specific comment next to list and dict.

AlexWaygood: Yeah, your wording is better!

PR #9168, opened 2022-11-12
Run stubtest on windows for colorama

Avasam:
One of the stubs that benefit from #8923
In fact, stubtest found 2 missing methods on windows!


stubs/colorama/colorama/win32.pyi
```diff
@@ -25,6 +25,8 @@ if sys.platform == "win32":
     def FillConsoleOutputCharacter(stream_id: int, char: str, length: int, start: COORD) -> int: ...
     def FillConsoleOutputAttribute(stream_id: int, attr: int, length: int, start: COORD) -> wintypes.BOOL: ...
     def SetConsoleTitle(title: str) -> wintypes.BOOL: ...
+    def GetConsoleMode(handle: int) -> int: ...
+    def SetConsoleMode(handle: int, mode: int) -> None: ...
```
Akuli, line 29: Can you temporarily delete these methods and then add them back? I would like to make sure that the CI fails when these are missing.

Avasam: There you go, fails on windows now :)

Akuli (approved):

PR #9170, opened 2022-11-12
Use `TypeAlias` in builtins and complete a `TODO` item

sobolevn:
From https://github.com/python/typeshed/pull/9158 I understood that explicit type aliases are not a goal. But, this one is special. It has a TODO item and some history to it.

I am also curious about whether mypy now supports it.


sobolevn:
>  stdlib/builtins.pyi:195: error: Unsupported left operand type for | ("object")  [operator]

Not today!

AlexWaygood:
> I am also curious about whether mypy now supports it.

I haven't checked yet, but I wonder if https://github.com/python/mypy/pull/14008 might have fixed this bug (not yet released, will be released as part of mypy 1.0).

PR #9171, opened 2022-11-12
Add types to `multiprocessing/reduction.pyi`

sobolevn:
Source: https://github.com/python/cpython/blob/main/Lib/multiprocessing/reduction.py


stdlib/multiprocessing/reduction.pyi
```diff
@@ -47,10 +54,10 @@ else:
         ACKNOWLEDGE: Literal[False]
 
     def recvfds(sock: socket, size: int) -> list[int]: ...
-    def send_handle(conn, handle, destination_pid) -> None: ...
+    def send_handle(conn: HasFileno, handle: int, destination_pid: object) -> None: ...
```
AlexWaygood, line 57:
```suggestion
    # destination_pid is unused
    def send_handle(conn: HasFileno, handle: int, destination_pid: object) -> None: ...
```

stdlib/multiprocessing/reduction.pyi
```diff
@@ -84,4 +91,4 @@ class AbstractReducer(metaclass=ABCMeta):
         sendfds = _sendfds
         recvfds = _recvfds
         DupFd = _DupFd
-    def __init__(self, *args) -> None: ...
+    def __init__(self, *args: object) -> None: ...
```
AlexWaygood, line 94:
```suggestion
    # *args are unused
    def __init__(self, *args: object) -> None: ...
```

AlexWaygood (commented):
Looks good, just a few tiny nits

AlexWaygood (approved):
Thanks!

PR #9172, opened 2022-11-12
types: add `__getitem__` on `GenericAlias`

iyume:
Make `types.GenericAlias` subscriptable just like `typing._GenericAlias`.

This is useful when:

```python
alias = dict[int, T]

if isinstance(alias, GenericAlias):
    alias[int]
```


stdlib/types.pyi
```diff
@@ -585,6 +585,7 @@ if sys.version_info >= (3, 9):
         @property
         def __parameters__(self) -> tuple[Any, ...]: ...
         def __init__(self, origin: type, args: Any) -> None: ...
+        def __getitem__(self, typeargs: Any) -> Any: ...
```
JelleZijlstra, line 588:
```suggestion
        def __getitem__(self, typeargs: Any) -> GenericAlias: ...
```
It always returns a GenericAlias (see the `ga_getitem` function in CPython).

iyume: Right. `__getitem__` only replace the concrete type and not change itself. But why `typing._Alias` returns `Any`?

https://github.com/python/typeshed/blob/1aab9d136a7f9fa090e40e34a41142c25d8933e9/stdlib/typing.pyi#L261

stdlib/types.pyi
```diff
@@ -585,6 +585,7 @@ if sys.version_info >= (3, 9):
         @property
         def __parameters__(self) -> tuple[Any, ...]: ...
         def __init__(self, origin: type, args: Any) -> None: ...
+        def __getitem__(self, typeargs: Any) -> GenericAlias: ...
```
JelleZijlstra, line 588:
```suggestion
        def __getitem__(self, __typeargs: Any) -> GenericAlias: ...
```
It's positional-only.

stdlib/types.pyi
JelleZijlstra: That class doesn't even exist at runtime and type checkers special case it, so we probably didn't think too hard about it.

JelleZijlstra (approved):

PR #9173, opened 2022-11-12
stubtest_third_party.py: Allow running non-listed platforms locally

Avasam:
Here is my proposed solution for: https://github.com/python/typeshed/pull/8923#issuecomment-1312528733

This should keep the same behaviour on the CI (although with a more accurate skip message when no platform are listed).

While still allowing to run stubtest locally.

I am not immediatly skipping if platforms are listed but is ran on an "unsupported" platform locally, because of the following edge case:
Say a stub is the same on darwin and linux, but different on windows. You don't want to run the CI for all three, so you may specify `platforms = ["linux", "win32"]`. But it's still fine to run locally on macos.
image

In this solution, the behaviour is affected by the "CI" environment variable: https://docs.github.com/en/actions/learn-github-actions/environment-variables#default-environment-variables . But it could be something else like a CLI flag passed by the github action.


AlexWaygood:
Hmm, I'm not a massive fan of the script magically doing something different in CI due to the presence of an environment variable. I like the idea of a CLI flag more.

Avasam:
Failure should be fixed by #9176

sobolevn (commented):
Another idea.

Let's say you work on a module that has several supported platforms like `psutil`. And you want to run several tests locally: for `win32`, `linux`, and `macos`.

Right now there's no way to do it.

With mypy / stubtest you can pass `--platform=X` as a CLI argument.
Why won't we have the same thing here?

So, if no `--platform` is passed: we just skip unmatching platforms.
If it is, we use whatever is passed (with a warning that the results might not be accurate).

AlexWaygood:
> With mypy / stubtest you can pass `--platform=X` as a CLI argument.
> Why won't we have the same thing here?

It doesn't make much sense to pass `--platform` to stubtest. The whole point of stubtest is that it compares the stub to what's happening at runtime. And stubtest has no idea what the runtime would be like on linux if it's running on Windows.

I have a few nits to sort out (I'll review soon), but I broadly like the design of this PR now.

sobolevn:
Yes, it does not make much sense. This is why I've added this part:

> If it is, we use whatever is passed (with a warning that the results might not be accurate).

But, it is sometimes better than not being able to run stubtest at all.

tests/stubtest_third_party.py
```diff
@@ -34,10 +34,24 @@ def run_stubtest(dist: Path, *, verbose: bool = False) -> bool:
...
+            return True
+
+    elif platforms_to_test and sys.platform not in platforms_to_test:
+        print(
+            colored(
+                f"warning: your platform, {sys.platform}, is not in: {platforms_to_test}. "
+                + "It may be unsupported or deemed redundant to test on the CI",
+                "yellow",
+            )
+        )
```
AlexWaygood, line 54: The current output in CI is a bit too verbose imo. I think the new warning can also be a bit more concise.

```suggestion
    platforms_to_test = stubtest_meta.get("platforms", ["linux"])
    if sys.platform not in platforms_to_test:
        if ci_run:
            print(colored(f"skipping (unsupported platform)", "yellow"))
            return True
        print(
            colored(
                f"Note: stubtest is not currently tested on {sys.platform} in typeshed's CI.",
                "yellow",
            )
        )
```

tests/stubtest_third_party.py
```diff
@@ -167,6 +181,7 @@ def main() -> NoReturn:
     parser.add_argument("-v", "--verbose", action="store_true", help="verbose output")
     parser.add_argument("--num-shards", type=int, default=1)
     parser.add_argument("--shard-index", type=int, default=0)
+    parser.add_argument("--ci-run", action="store_true")
```
AlexWaygood, line 184: It would be nice to add a short description of what the flag does here, using the `help=` keyword argument

AlexWaygood (commented):
Thanks! I like the design of this PR, as I said. I agree that the least surprising behaviour is if `python tests/stubtest_third_party.py foo` tests the stubs for `foo` locally on whatever system you're running on. That's the behaviour we had prior to #8923, and it's the behaviour we should go back to; it's the CI run that should require additional command-line flags.

I think the logic at line 37 can be simplified a bit, and the messages printed can be less verbose -- have offered a suggestion there. I'm also not keen on the name of the flag -- `--ci-run` isn't very descriptive about what the flag actually _does_. Maybe `--supported-stubs-only`?

tests/stubtest_third_party.py
Avasam: This looks reasonable

tests/stubtest_third_party.py
Avasam: I agree on the flag name. It's a principle I need to remind myself more often:
"Don't make code that's environment-specific. Instead give options and let the environment decide".
So instead of having a "ci" flag, I should have a flag that describes this option.

I went with "specified" instead of "supported". For the same reason as the multi-platform PR by sobolevn: It's not necessarily unsupported. It's just not tested (it could simply be redundant).

tests/stubtest_third_party.py
```diff
@@ -36,10 +36,10 @@ def run_stubtest(dist: Path, *, verbose: bool = False, ci_run: bool = False) ->
 
     platforms_to_test = stubtest_meta.get("platforms", ["linux"])
     if sys.platform not in platforms_to_test:
-        if ci_run:
-            print(colored(f"skipping (unsupported platform)", "yellow"))
+        if specified_stubs_only:
+            print(colored("skipping (unspecified platform)", "yellow"))
```
AlexWaygood, line 40:
```suggestion
            print(colored("skipping (platform not specified in METADATA.toml)", "yellow"))
```

tests/stubtest_third_party.py
AlexWaygood: "Specified" is a great choice!

AlexWaygood (approved):
Nice — thank you!

PR #9174, opened 2022-11-12
Remove `setuptools._distutils.command.bdst_msi`

Avasam:
I was looking into if this stub needed to be ran on multiple platforms.
Turns out `setuptools._distutils.command.bdst_msi` has been deprecated since Python 3.9 (https://docs.python.org/3/whatsnew/3.9.html#deprecated). And recent versions of setuptools don't even include it.


AlexWaygood (approved):

AlexWaygood:
I've discovered a few fictional classes in typeshed before, but that's the first time I've come across an entire module that's fictional :)

PR #9175, opened 2022-11-12
Explain why `vobject` is not tested on Windows

Avasam:
A very simple comment update to mention that `vobject.win32tz` is broken even on Windows on Python 3. Hence there's no point in testing this stub on multiple platforms.

Technically that makes this module Python2-only. Should it be removed or is there value to keeping it?

Rel: https://github.com/eventable/vobject/pull/187


JelleZijlstra (approved):
Thanks! Seems OK to keep the stub in the expectation that the runtime will be fixed.

PR #9177, opened 2022-11-12
Update asyncio subprocess optional **kwargs.

socketpair:
Made strongly according to documentation (possible Sequence vs Iterable vs Collection bug).

But did not check.


socketpair:
Please fix someone `typing.Iterable` -> `Collections.abc.Iterable`. I don't know how to fix from github web interface...

socketpair:
@JelleZijlstra

JelleZijlstra (commented):
These kwargs are delegated to `subprocess.Popen`, so it would be good to match the types exactly. It seems that 3.11 added another new kwarg, `process_group: int | None = ...`, so we should add a 3.11 branch to these stubs too.

`pass_fds` is Any in `stdlib/subprocess.pyi`, but from looking at the code `Iterable[int]` does appear to be the right type. Feel free to fix that type in this PR too.

socketpair:
@JelleZijlstra done

socketpair:
subprocess.py:
```Python
                    fds_to_keep = set(pass_fds)
                    fds_to_keep.add(errpipe_write)
                    self.pid = _posixsubprocess.fork_exec(
                            args, executable_list,
                            close_fds, tuple(sorted(map(int, fds_to_keep))),
```

Yes, it's allowed to pass non-int. But I think it's a bug to pass non-int.

socketpair:
Since `bool(Iterable)`, and code contains `if pass_fds` then it should be either Collection or Sequence. I don't know what is better. Please advise.

JelleZijlstra:
The new error seems correct, though maybe mypy should listen to the type ignore: https://github.com/pyppeteer/pyppeteer/blob/7245ee0bccd97d43e45b801319dd6cbae860292a/pyppeteer/launcher.py#L148

stdlib/asyncio/subprocess.pyi
```diff
@@ -65,7 +65,13 @@ if sys.version_info >= (3, 10):
         creationflags: int = ...,
         restore_signals: bool = ...,
         start_new_session: bool = ...,
-        pass_fds: Any = ...,
+        pass_fds: Collection[int] = ...,
+        group: None | str | int = ...,
+        extra_groups: None | Collection[str | int] = ...,
+        user: None | str | int = ...,
+        umask: int = ...,
+        process_group: str | int | None = ...,
```
JelleZijlstra, line 73: subprocess has this as just `int | None`, which looks correct based on the C code.

socketpair: I dont' understand what you wanted to say.

socketpair:
What should I do to move this PR to merging ?

stdlib/asyncio/subprocess.pyi
JelleZijlstra: In `subprocess.pyi`, the type of the `process_group` parameter is `int | None`, not `str | int | None`. Based on my reading of the relevant C code, `int | None` appears to be correct.

socketpair:
@JelleZijlstra done

hauntsaninja:
I'm not seeing the change, did you forget to push?

socketpair:
@JelleZijlstra sorry for calling again. What should I change (what is wrong) in current PR? Please mark these changes somehow in order to distinguish from just speaking. Now, I will just rebase, no other changes.

stdlib/asyncio/subprocess.pyi
AlexWaygood: @socketpair, you haven't addressed this comment here. Jelle is asking you to change the annotation for the `process_group` parameter (both for this function and all other functions that have this parameter) to be `int | None` instead of `str | int | None`. You need to either make the change or explain why you think that this would be incorrect.

AlexWaygood (changes requested):
Requesting changes until https://github.com/python/typeshed/pull/9177#discussion_r1045269090 is addressed

stdlib/asyncio/subprocess.pyi
socketpair: Really. it's my bug. Will fix in next commit.

socketpair:
I don't understand new Github review semantics. I just changed what you requested. I don't know if I need to tap something in GH

PR #9181, opened 2022-11-12
Run stubtest on Windows for waitress

Avasam:
Yet another stub that benefits from #8923 . Fixing stubtest failure on Windows and offering better coverage.

I did not add `darwin` to `platforms` because it would be redundant with `linux`. But I still added a `stubtest_allowlist_darwin.txt` for those who would like to run it locally on macos. (This is the edge case mentioned in #9173 )


stubs/waitress/waitress/wasyncore.pyi
```diff
@@ -77,20 +78,21 @@ class dispatcher_with_send(dispatcher):
...
-    def __init__(self, fd: BytesIO) -> None: ...
-    def __del__(self) -> None: ...
-    def recv(self, *args: Any) -> bytes: ...
-    def send(self, *args: Any) -> bytes: ...
-    def getsockopt(self, level: int, optname: int, buflen: bool | None = ...) -> int: ...
-    read: Callable[..., bytes] = ...
-    write: Callable[..., bytes] = ...
-    def close(self) -> None: ...
-    def fileno(self) -> BytesIO: ...
+if sys.platform == "linux" or sys.platform == "darwin":
```
JelleZijlstra, line 81:
```suggestion
if sys.platform != "win32":
```
Consistent with the other check you added.

Avasam: When the original code does `if os.name == "posix"`, the stub did `if sys.platform == "linux" or sys.platform == "darwin"` (there's another platform check like this in `waitress.trigger`).

I added `if sys.platform != "win32"` to be consistent with other `if hasattr(socket, "AF_UNIX")` checks.

But I can change them all to be either if you'd like.

JelleZijlstra: I think the runtime checks are pretty much equivalent, and in stubs we have tended to assume all non-Windows OSes are Unix.

Avasam: Sounds good, I'll make all 3 `!= "win32`

JelleZijlstra (approved):

PR #9184, opened 2022-11-13
Complete stubtest for `aiofiles` + fix on Windows

Avasam:
Set `ignore_missing_stub = false`, added a missing method and property, and fixed stubtest runs on Windows.

I did not add `darwin` to `platforms` because it would be redundant with `linux`. But I still added a `stubtest_allowlist_darwin.txt` for those who would like to run it locally on macos. (This is the edge case mentioned in #9173 )


stubs/aiofiles/aiofiles/tempfile/temptypes.pyi
```diff
@@ -17,6 +17,7 @@ _T = TypeVar("_T")
 class AsyncSpooledTemporaryFile(AsyncBase[_T]):
     def fileno(self) -> Generator[Incomplete, Incomplete, Incomplete]: ...
     def rollover(self) -> Generator[Incomplete, Incomplete, Incomplete]: ...
+    async def close(self, *args: Incomplete, **kwargs: Incomplete) -> None: ...
```
JelleZijlstra, line 20:
```suggestion
    async def close(self) -> None: ...
```
This gets delegated to the underlying file's close() method.

Avasam: Thanks, I wasn't sure of the exact signature. I'll add an allowlist entry.

PR #9187, opened 2022-11-13
Add support for Homebrew and Chocolatey

Avasam:
I have added support to specify packages to install using Homebrew on MacOS and Chocolatey on Windows. Both come pre-install on the Github Action environments.

~~Fixed all affected stubs  (`JACK-Client`, `pyaudio`, and `PycURL`)~~
~~Added `ignore_missing_stub = false` and completed said stubs~~
Edit: #9205, #9206 and #9207
Locally validated the installs on both MacOS and Windows.

~~JACK-Client should be the same on all platforms. I'm only running it on all platforms to prove that the brew and choco installs work. Before merging it should not have any platforms specified.~~
Edit: Done!

Special notes about pycurl:
The pycurl install on Windows is currently broken as it depends on a version of setuptools that's too old.
That's fixed in their master branch, but still requires to be built from source and does not provide wheels (yet).

The pycurl install on MacOS is too complicated for the CI and does not work with stubtest (unless I'm doing it wrong):
> % brew install openssl
> % export LDFLAGS="-L/usr/local/opt/openssl@3/lib"
> % export CPPFLAGS="-I/usr/local/opt/openssl@3/include"
> % pip install --compile --install-option="--with-openssl" pycurl


stubs/pycurl/METADATA.toml
```diff
@@ -1,5 +1,14 @@
...
+# Install on Windows requires building PycURL from source
+#
+# Install on MacOS is too complicated for the CI and does not work with stubtest:
+# % brew install openssl
+# % export LDFLAGS="-L/usr/local/opt/openssl@3/lib"
+# % export CPPFLAGS="-I/usr/local/opt/openssl@3/include"
+# % pip install --compile --install-option="--with-openssl" pycurl
 platforms = ["linux"]
 apt_dependencies = ["libcurl4-openssl-dev"]
+brew_dependencies = ["openssl"]
```
sobolevn, line 14: Do we need this line if we don't run `macos` in CI?

Avasam: Likely not. Especially since atm the CI will install everything even if stubtest isn't run.

AlexWaygood:
There's quite a lot going on in this PR all at once, which  makes it a little hard to review:

- `Any` -> `Incomplete` for several annotations
- missing symbols added to several stubs packages so that we can enable `ignore_missing_stub = false`
- New "infrastructure" changes so that we can install dependencies via homebrew or chocolatey when required.

Many of these look like good changes, but would you mind splitting it up into a few smaller PRs, so that we can consider them in isolation? :)

Avasam:
> There's quite a lot going on in this PR all at once, which makes it a little hard to review:
> [...] Many of these look like good changes, but would you mind splitting it up into a few smaller PRs, so that we can consider them in isolation? :)

Absolutely! #9205, #9206 and #9207 . Tackling these first will reduce the changes in this PR.

AlexWaygood:
> > There's quite a lot going on in this PR all at once, which makes it a little hard to review:
> > [...] Many of these look like good changes, but would you mind splitting it up into a few smaller PRs, so that we can consider them in isolation? :)
> 
> Absolutely! #9205, #9206 and #9207 . Tackling these first will reduce the changes in this PR.

Thanks so much! The changes in this PR are _much_ more manageable now :)

stubs/pycurl/METADATA.toml
AlexWaygood: Let's leave it out in that case (you can add a comment if you like) -- it'll only slow down the CI unnecessarily

AlexWaygood (commented):
LGTM other than @sobolevn's point about the unnecessary `brew_dependency` entry for `pycurl`

AlexWaygood:
(The CI failure will be fixed by #9243.)

stubs/pycurl/METADATA.toml
Avasam: Yep, that's my plan. Now that it's proven the install works on CI, I've commented out unneeded installs and removed redundant platforms.

AlexWaygood:
Ah, looks like we now have a new CI failure due to the fact that 0 packages are being passed into the `choco install` step.

.github/workflows/daily.yml
```diff
@@ -69,12 +69,21 @@ jobs:
...
-            sudo apt install -y $(python tests/get_packages.py)
+            sudo apt update && sudo apt install -y $PACKAGES
 
             xvfb-run python tests/stubtest_third_party.py --specified-stubs-only --num-shards 4 --shard-index ${{ matrix.shard-index }}
           else
+            if [ "${{ matrix.os }}" = "macos-latest" ]; then
+              NONINTERACTIVE=1 brew install $PACKAGES
+            fi
+
+            if [ "${{ matrix.os }}" = "windows-latest" ]; then
```
sobolevn, line 83:
```suggestion
            if [ ! -z "$PACKAGES" ] && [ "${{ matrix.os }}" = "windows-latest" ]; then
```

sobolevn: @AlexWaygood can you please apply this change? :)

This will fix the CI and we can merge it.

AlexWaygood: It's probably worth making this change to the `apt_dependencies` and `brew_dependencies` branches of this script as well, right? What if we remove stubs for `JACK-Client` and `pyaudio` at some point in the future, and then the CI randomly starts failing because suddenly we're passing in 0 packages to the `brew install` part of this workflow?

Anyway, is there a massive rush? It's @Avasam's PR :)

Avasam: No rush on my side. I simply wanted to complete all the stubs that benefited from sobolevn's work with multi-platform support. And this is the last PR in that series.

Whilst more unlikely, I've added the check for apt installs as well. It doesn't hurt.

I've also changed the os condition to use `runner.os` instead of `matrix.os`. This way the exact os version can be updated without braking the condition.

sobolevn (approved):
Thanks!

AlexWaygood (approved):
Thanks @Avasam! 🎉

Issue #9189, opened 2022-11-13
Tests failing due to pip install connection issues


Avasam:
I've seen ***a lot*** of tests failing during the install step lately due to `ERROR: Could not install packages due to an OSError: HTTPSConnectionPool`.

The python action offers a way to cache pip dependency and specify a differently named `requirements.txt` to base itself off: https://github.com/actions/setup-python/blob/main/docs/advanced-usage.md#caching-packages
Maybe this could help?

AlexWaygood:
Caching pip requirements seems like a good idea to me. I've no idea whether it'll solve this problem, but it might do. And I've seen it speed up CI on other projects pretty substantially, so there's no real downside AFAICS.

Avasam:
Completed in #9192

PR #9192, opened 2022-11-13
Cache pip downloads in CI

Avasam:
Addresses #9189

- Let `actions/setup-python` know to cache pip.
- Jobs that don't run `pip install` have no cache specified.
- I've specified the relevant `cache-dependency-path` for every job.
  - `mypy_primer` is a bit of a special case, but it does so much downloading it should still be beneficial. Edit: To be reconsidered, we don't have a great way of specifying a cache key to invalidate.

Note that this only caches the **download** of distributions, not the **install** (for that, see: https://github.com/actions/setup-python/issues/330 )
This means that any change in one of the files specified for `cache-dependency-path` will result in a different cache hash and all dependencies will be redownloaded (current behaviour).

Further improvements may include:
- Caching of machine dependencies (apt, Homebrew, Choco).
- Extracting the "skip" logic from stubtest_third_party so that the job can skip installs if stubtest is going to be skipped.

I wanna keep this PR simple and just targeting pip dependencies. I can create new issues for those.


.github/workflows/daily.yml
```diff
@@ -55,6 +58,10 @@ jobs:
       - uses: actions/setup-python@v4
         with:
           python-version: "3.9"
+          cache: 'pip'
+          cache-dependency-path: |
+            requirements-tests.txt
```
sobolevn, line 63: Should this contain `METADATA.toml` files? 🤔

.github/workflows/daily.yml
```diff
@@ -34,7 +34,10 @@ jobs:
         uses: actions/setup-python@v4
         with:
           python-version: ${{ matrix.python-version }}
+          cache: 'pip'
+          cache-dependency-path: requirements-tests.txt
       - name: Update pip
+        if: matrix.os == "macos-11"
```
sobolevn, line 40: Why do we need this line? I usually update pip on all envs.

.github/workflows/typecheck_typeshed_code.yml
```diff
@@ -34,6 +34,8 @@ jobs:
       - uses: actions/checkout@v3
       - uses: actions/setup-python@v4
         with:
-          python-version: 3.9
+          python-version: "3.9"
+          cache: 'pip'
+          cache-dependency-path: requirements-tests.txt
```
sobolevn, line 39:
```yaml
          python-version: "3.9"
          cache: 'pip'
```
Consider this as a rant about `yaml`: these three lines have three different quotes 😒

.github/workflows/mypy_primer.yml
```diff
@@ -32,6 +32,11 @@ jobs:
       - uses: actions/setup-python@v4
         with:
           python-version: "3.10"
+          cache: 'pip'
+          cache-dependency-path: |
+            typeshed_to_test/requirements-tests.txt
+            typeshed_to_test/stubs/**/@tests/requirements-stubtest.txt
+            typeshed_to_test/stubs/**/METADATA.toml
```
AlexWaygood, line 39: I'm not sure this will actually speed anything up. The only thing this workflow directly pip-installs is `mypy_primer`, which has 0 dependencies. `mypy_primer` will then dynamically install loads of things during the course of the script's execution, but I don't think it'll ever install anything from typeshed's `requirements-tests.txt` file, any `METADATA.toml` files, or any `requirements-stubtest.txt` files.

Maybe just get rid of the changes to this workflow file, for now?

.github/workflows/daily.yml
Avasam: > Usually, pip is always updated and included in [runner images](https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2204-Readme.md?rgh-link-date=2022-09-29T16%3A30%3A56Z). Because of this, the latest version of pip is usually installed already.

https://github.com/actions/setup-python/issues/506#issuecomment-1262522726

I can only assume `python -m pip install -U pip` was done because of `"macos-11"`. As the only times that command is ran is when also using `"macos-11`. So only running it on that OS is a small optimization for the others.

.github/workflows/daily.yml
Avasam: When we'll be able to specify non-types dependencies, are they going to be specified in `METADATA.toml` ? If so, then yes `cache-dependency-path` should contain `METADATA.toml` for stubtest jobs.

.github/workflows/daily.yml
AlexWaygood: Given that we have `PIP_DISABLE_PIP_VERSION_CHECK: 1` set as an environment variable in this workflow, we probably don't need this step at all. It usually isn't essential to have the very latest version of pip installed; the only significant benefit from upgrading pip in CI is that you don't get the annoying "WARNING: a newer version of pip is available, you should upgrade NOW" message with every CI job.

.github/workflows/daily.yml
AlexWaygood: > When we'll be able to specify non-types dependencies, are they going to be specified in `METADATA.toml` ?

I think that's the idea, yup!

.github/workflows/mypy_primer.yml
Avasam: It would be more sensible to base the cache key on `mypy_primer.py`. But I can't really specify that. mypy_primer downloads so much it would make sense to cache those. But then how do we specify that the cache should be invalidated? Here I tried to target everything that may suggest a need to invalidate (ie version bump or dependency change of the stub).

Anyway, this can be reconsidered, I'll revert for now.

.github/workflows/typecheck_typeshed_code.yml
Avasam: 😅 What's the prefered quote style between `"` and `'` here? Normally I use the [`redhat.vscode-yaml`](https://marketplace.visualstudio.com/items?itemName=redhat.vscode-yaml) plugin which formats to `"` (but I've disabled all my auto-formatters on non-python files in this project).

AlexWaygood: > What's the preferred quote style between `"` and `'` here?

I don't think we have one, and I for one don't care at all, as long as it works :)

.github/workflows/daily.yml
Avasam: I didn't even notice `PIP_DISABLE_PIP_VERSION_CHECK`. Should I remove the step? Leave it as it was before?

AlexWaygood: I reckon you can remove it. There's other workflows we have in this repo where we don't have it as a step, and it fits with the theme of this PR ("speed up CI by getting pip to do less work").

AlexWaygood (approved):
LGTM!

PR #9194, opened 2022-11-14
`DecimalTuple` can have string exponent

### LINKED ISSUE #9191: DecimalTuple definition is incorrect
Current `DecimalTuple` definition is:
```py
class DecimalTuple(NamedTuple):
    sign: int
    digits: tuple[int, ...]
    exponent: int
```
source: https://github.com/python/typeshed/blob/main/stdlib/_decimal.pyi#L16

But if a decimal value is `NaN` the exponent is of type `str`.  The following example illustrate that:
```py
>>> from decimal import Decimal
>>> exponent = Decimal('NaN').as_tuple().exponent
>>> print(exponent)
n
>>> print(type(exponent))
<class 'str'>
>>>
```

It seems like the correct definition should be like this:
```py
class DecimalTuple(NamedTuple):
    sign: int
    digits: tuple[int, ...]
    exponent: Union[int, str]
```

sobolevn:
Source code: https://github.com/python/cpython/blob/db115682bd639a2642c617f0b7d5b30cd7d7f472/Lib/_pydecimal.py#L1049-L1053

Closes #9191

### DIFF
--- a/stdlib/_decimal.pyi
+++ b/stdlib/_decimal.pyi
@@ -4,7 +4,7 @@ from _typeshed import Self
 from collections.abc import Container, Sequence
 from types import TracebackType
 from typing import Any, ClassVar, NamedTuple, Union, overload
-from typing_extensions import TypeAlias
+from typing_extensions import Literal, TypeAlias
 
 _Decimal: TypeAlias = Decimal | int
 _DecimalNew: TypeAlias = Union[Decimal, float, str, tuple[int, Sequence[int], int]]
@@ -16,7 +16,7 @@ __libmpdec_version__: str
 class DecimalTuple(NamedTuple):
     sign: int
     digits: tuple[int, ...]
-    exponent: int
+    exponent: int | Literal["n", "N", "F"]
 
 ROUND_DOWN: str
 ROUND_HALF_UP: str


sobolevn:
This failure is not related: https://github.com/python/typeshed/actions/runs/3460013243/jobs/5776061994 I cannot restart it.

Looking at the primer results ⌛

sobolevn:
`mongo`'s code is correct. But, we cannot narrow `exponent` based on other calls: https://github.com/mongodb/mongo-python-driver/blob/fcb11514506acddbd50b8ec13e76a7c34d336aac/bson/decimal128.py#L81-L83

```python
    sign, digits, exponent = value.as_tuple()

    if value.is_nan():
```

`pyspark`'s code is broken: https://github.com/apache/spark/blob/632784dec8a0c682a85a888c66e83f709e402e18/python/pyspark/sql/connect/column.py#L114-L115

It is calling `abs()` on a possibly string. They even state that `raise ValueError("Python Decimal not supported.")`

`pydantic` has correct runtime code: https://github.com/pydantic/pydantic/blob/13ad034e4400baf5a875767e363aba879501065f/pydantic/_internal/_validators.py

```python
            _, digit_tuple, exponent = value.as_tuple()
            if not self.allow_inf_nan and exponent in {'F', 'n', 'N'}:
                raise PydanticKnownError('finite_number')

            if self.check_digits:
                if exponent >= 0:
```

But, looks like we cannot narrow `exponent` to be `int` in this case :(

Code in `psycopg` is actually correct (it has `type: ignore` that will be removed): https://github.com/psycopg/psycopg/blob/a8168869b762c821e7f07d1fc52d3642f899711e/psycopg/psycopg/types/numeric.py#L393-L397

srittau (approved):

PR #9197, opened 2022-11-14
Bump test dependencies; improve `pycln` config

AlexWaygood:
- Bump mypy to 0.991
- Bump pytype to 2022.11.10
- Bump pycln to 2.1.2
- Enable the new `disable_all_dunder_policy` flag for pycln.

  This flag means that pycln treats `__init__.py(i)` files just like normal `.py(i)` files. By default, pycln is much more cautious when removing imports from `__init__.py(i)` files due to the fact that many projects use "redundant" imports in `__init__.py` files for implicit reexports. But it doesn't make sense for pycln to apply the same level of caution to `__init__.pyi` files in stubs, where all imports are "private to the stub" except imports using the `import foo as foo` idiom. So it makes sense to enable this flag for typeshed.
- Put all pycln config in the `pyproject.toml` file instead of hardcoding it into the `.pre-commit-config.yaml` file.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,9 +1,9 @@
 repos:
   - repo: https://github.com/hadialqattan/pycln
-    rev: v2.1.1 # must match requirements-tests.txt
+    rev: v2.1.2 # must match requirements-tests.txt
     hooks:
       - id: pycln
-        args: [--all, stubs, stdlib, tests, scripts]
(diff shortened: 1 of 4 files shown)


AlexWaygood:
PR to my typeshed fork showing that pycln still works fine with the new config setup: https://github.com/AlexWaygood/typeshed/pull/7

JelleZijlstra (approved):

Avasam:
@AlexWaygood I think you missed a `--all` in `runtests.py`

PR #9203, opened 2022-11-15
`check_consistent.py`: Add a check guarding against typos for the new `platforms` key

AlexWaygood:


tests/check_consistent.py
```diff
@@ -151,6 +151,12 @@ def check_metadata() -> None:
             for key in data.get("tool", {}).get(tool, {}):
                 assert key in tk, f"Unrecognised {tool} key {key} for {distribution}"
 
+        stubtest_platforms = set(data.get("tool", {}).get("stubtest", {}).get("platforms", []))
+        supported_platforms = {"win32", "linux", "darwin"}
```
srittau, line 155: I'd consider moving the supported platforms to the top of the file.

srittau (approved):
LGTM, one optional remark.

tests/check_consistent.py
AlexWaygood: Yes, that makes sense. At some point I'd like to centralise that information between tests as well (but not in this PR).

PR #9205, opened 2022-11-15
Add missing symbols in `JACK-Client` stubs

Avasam:
Split off from #9187
Turned off `ignore_missing_stub` for JACK-Client stubs.
Ensured it's valid for Windows and MacOS.

### DIFF
--- a/stubs/JACK-Client/METADATA.toml
+++ b/stubs/JACK-Client/METADATA.toml
@@ -1,5 +1,6 @@
 version = "0.5.*"
 
 [tool.stubtest]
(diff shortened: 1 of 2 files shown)


JelleZijlstra (approved):
Didn't actually manage to install locally to verify, but looks reasonable.

Avasam:
> Didn't actually manage to install locally to verify, but looks reasonable.

Just FYI, you'd have to install `jack` first (I did with Chocolatey on Windows or Homebrew on MacOS)

PR #9206, opened 2022-11-15
Add missing symbols in `pyaudio` stubs

Avasam:
Split off from #9187
Turned off `ignore_missing_stub` for pyaudio stubs.
Ensured it's valid for Windows and MacOS.

### DIFF
--- a/stubs/pyaudio/pyaudio.pyi
+++ b/stubs/pyaudio/pyaudio.pyi
@@ -1,5 +1,8 @@
+import sys
 from collections.abc import Callable, Mapping, Sequence
-from typing_extensions import Final, TypeAlias
+from typing_extensions import Final, TypeAlias, final
+
+__docformat__: str
 
 paFloat32: Final[int]
 paInt32: Final[int]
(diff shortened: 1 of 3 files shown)


stubs/pyaudio/pyaudio.pyi
```diff
@@ -64,7 +67,45 @@ paOutputUnderflow: Final[int]
...
+        paMacCoreConversionQualityLow: Final[int]
+        paMacCoreConversionQualityHigh: Final[int]
+        paMacCoreConversionQualityMax: Final[int]
+        paMacCorePlayNice: Final[int]
+        paMacCorePro: Final[int]
+        paMacCoreMinimizeCPUButPlayNice: Final[int]
+        paMacCoreMinimizeCPU: Final[int]
+        flags: Final[int]
+        channel_map: Final[_ChannelMap | None]
+        def __init__(self, flags: int | None = ..., channel_map: _ChannelMap | None = ...) -> None: ...
```
JelleZijlstra, line 88:
```suggestion
        def __init__(self, flags: int = ..., channel_map: _ChannelMap = ...) -> None: ...
```
Tried it locally, passing None doesn't work.

Avasam: You're right. Since I'm moving these around, may as well fix them. I've also double-checked that `PaMacCoreStreamInfo.__init__` is correct.

PR #9207, opened 2022-11-15
Add missing symbols in `pycurl` stubs

Avasam:
Split off from #9187
Turned off `ignore_missing_stub` for pycurl stubs.
Ensured it's valid for MacOS.

Use of `Incomplete` for incomplete types rather than `Any`.

Couldn't test on Windows because the pycurl install on Windows is currently broken as it depends on a version of setuptools that's too old.
That's fixed in their master branch, but still requires to be built from source and does not provide wheels (yet).


stubs/pycurl/pycurl.pyi
```diff
@@ -150,8 +169,8 @@ HEADER_SIZE: int
 HEADER_UNIFIED: int
 HTTP200ALIASES: int
 HTTPAUTH: int
-HTTPAUTH_ANY: int
-HTTPAUTH_ANYSAFE: int
+HTTPAUTH_Incomplete: int
```
AlexWaygood, line 172: 🤨 find-and-replace error? 😉

AlexWaygood:
`¯\_(ツ)_/¯`

Tried for a while to install it, gave up. All looks plausible, though!

PR #9208, opened 2022-11-16
Fix urllib3 Url.port type

pquentin:
I maintain urllib3 and can confirm the [port is an int](https://github.com/urllib3/urllib3/blob/8b8e4b5a148d0eb706daf5ac48b4423b434495f5/src/urllib3/util/url.py#L381).

Thankfully, the upcoming 2.0 release is fully type annotated!


srittau (approved):
Thanks!

PR #9209, opened 2022-11-16
Refine types for asyncio transports

### LINKED ISSUE #9199: loop.create_connection should return Transport not BaseTransport
According to [the docs](https://docs.python.org/3/library/asyncio-protocol.html?highlight=transport#asyncio.Transport), Transport is the class created by `loop.create_connection`, but the annotations have it returning a BaseTransport. This causes a false positive when the returned transport is passed to a StreamWriter, which as of #7611 expects a WriteTransport (a base class of Transport but not of BaseTransport).

bmerry:
- Change the return type of create_connection, start_tls, connect_accepted_socket, create_unix_connection to Transport rather than BaseTransport (closes #9199).
- Change the return type of create_datagram_endpoint to DatagramTransport rather than BaseTransport.
- Change the argument of sendfile to WriteTransport rather than BaseTransport.

I considered also changing the argument of start_tls to Transport, but I think that will give false positives for code that implements a custom transport class that inherits from both ReadTransport and WriteTransport but not from Transport, and I'm not sure if typing has a way to express an intersection of types. Since users are not normally expected to implement transports that may be overthinking things.

### DIFF
--- a/stdlib/asyncio/base_events.pyi
+++ b/stdlib/asyncio/base_events.pyi
@@ -5,7 +5,7 @@ from asyncio.events import AbstractEventLoop, AbstractServer, Handle, TimerHandl
 from asyncio.futures import Future
 from asyncio.protocols import BaseProtocol
 from asyncio.tasks import Task
-from asyncio.transports import BaseTransport, ReadTransport, SubprocessTransport, WriteTransport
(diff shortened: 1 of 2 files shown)


bmerry:
mypy-primer results look pretty promising, telling us that we're fixing things that people have had to work around.

PR #9211, opened 2022-11-16
Mark simplejson as completed

sobolevn:
Source: https://github.com/simplejson/simplejson


stubs/simplejson/simplejson/__init__.pyi
```diff
@@ -7,8 +9,83 @@ from simplejson.raw_json import RawJSON as RawJSON
...
+    namedtuple_as_object: bool = ...,
+    tuple_as_array: bool = ...,
+    bigint_as_string: bool = ...,
+    sort_keys: bool = ...,
+    item_sort_key: Callable[[Any], SupportsRichComparison] | None = ...,
+    for_json: bool = ...,
+    ignore_nan: bool = ...,
+    int_as_string_bitcount: int | None = ...,
+    iterable_as_array: bool = ...,
+    # Technically: you cannot pass extra `kw` without passing custom `cls`
```
JelleZijlstra, line 35: Might be useful to handle this with overloads. Being able to remove `**kwargs` from the stub is useful because it means type checkers can tell you if you get a parameter name wrong.

stubs/simplejson/simplejson/encoder.pyi
```diff
@@ -1,10 +1,61 @@
...
+        iterable_as_array: bool = ...,
+    ) -> None: ...
     def encode(self, o: Any) -> str: ...
     def default(self, o: Any) -> NoReturn: ...
     def iterencode(self, o: Any, _one_shot: bool = ...) -> Iterator[str]: ...
 
 class JSONEncoderForHTML(JSONEncoder): ...
+
+def encode_basestring(s: Any, _PY3: Literal[True] = ..., _q: str = ...) -> str: ...
+def py_encode_basestring_ascii(s: Any, _PY3: Literal[True] = ...) -> str: ...
```
JelleZijlstra, line 61:
```suggestion
def encode_basestring(s: str | bytes, _PY3: Literal[True] = ..., _q: str = ...) -> str: ...
def py_encode_basestring_ascii(s: str | bytes, _PY3: Literal[True] = ...) -> str: ...
```
These work only on str (including subclasses) and bytes.

stubs/simplejson/simplejson/__init__.pyi
sobolevn: Done.

stubs/simplejson/simplejson/__init__.pyi
```diff
@@ -32,17 +32,42 @@ def dumps(
     ignore_nan: bool = ...,
     int_as_string_bitcount: int | None = ...,
     iterable_as_array: bool = ...,
+    *,
```
JelleZijlstra, line 35: Thanks for adding the overloads!

I think we need to make everything from `indent` onwards keyword-only, because otherwise we'd incorrectly type calls with 6+ positional arguments.

stubs/simplejson/simplejson/__init__.pyi
AlexWaygood: This comment is no longer necessary now that we have the overloads, correct?
```suggestion
```

sobolevn:
@AlexWaygood I fixed the CI, would you have to time to take a look?

JelleZijlstra (approved):
Not sure why this got stuck. Looks like CI should succeed and this should be good to go.

Issue #9122, opened 2022-11-07
dateutil.parser.parse: Add precise types for **kwargs


JelleZijlstra:
Our stub for dateutil.parser.parse currently takes **kwargs: https://github.com/python/typeshed/blob/11be6effd3363fabf995da14f88586334e668e6e/stubs/python-dateutil/dateutil/parser/__init__.pyi#L44. But looking at the implementation and docs, we should be able to get more precise types. I can do this if nobody beats me to it.

PR #9212, opened 2022-11-16
Add exact `**kwargs` types to `dateutil.parser.parse`

sobolevn:
Sorry, @JelleZijlstra for taking this over, but I had a similar task: converting `python-dateutil` to a completed package.

Source https://github.com/dateutil/dateutil/blob/322fa34bc5fafb2e1c725e97909bf490a9dec1c8/src/dateutil/parser/_parser.py

Closes https://github.com/python/typeshed/issues/9122

### DIFF
--- a/stubs/python-dateutil/dateutil/parser/__init__.pyi
+++ b/stubs/python-dateutil/dateutil/parser/__init__.pyi
@@ -43,7 +43,6 @@ class parser:
         yearfirst: bool | None = ...,
         fuzzy: bool = ...,
         fuzzy_with_tokens: bool = ...,
-        **kwargs: Any,
     ) -> datetime: ...
 
 DEFAULTPARSER: parser
@@ -59,7 +58,6 @@ def parse(
     fuzzy_with_tokens: bool = ...,
     default: datetime | None = ...,
     tzinfos: _TzInfo | None = ...,
-    **kwargs: Any,
 ) -> datetime: ...
 
 class _tzparser: ...
--- a/stubs/python-dateutil/dateutil/parser/__init__.pyi
+++ b/stubs/python-dateutil/dateutil/parser/__init__.pyi
@@ -1,4 +1,4 @@
-from collections.abc import Mapping
+from collections.abc import Callable, Mapping
 from datetime import datetime, tzinfo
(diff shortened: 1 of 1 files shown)


stubs/python-dateutil/dateutil/parser/__init__.pyi
```diff
@@ -35,13 +37,30 @@ class parser:
...
         default: datetime | None = ...,
         ignoretz: bool = ...,
-        tzinfos: Mapping[str, tzinfo] | None = ...,
+        tzinfos: _TzInfo | None = ...,
+        *,
+        dayfirst: bool | None = ...,
+        yearfirst: bool | None = ...,
+        fuzzy: bool = ...,
+        fuzzy_with_tokens: bool = ...,
         **kwargs: Any,
```
JelleZijlstra, line 46:
```suggestion
```
We can remove this because we've covered all `**kwargs` that it accepts, and any other kwargs will error out.

stubs/python-dateutil/dateutil/parser/__init__.pyi
```diff
@@ -35,13 +37,30 @@ class parser:
...
+    parserinfo: parserinfo | None = ...,
+    *,
+    dayfirst: bool | None = ...,
+    yearfirst: bool | None = ...,
+    ignoretz: bool = ...,
+    fuzzy: bool = ...,
+    fuzzy_with_tokens: bool = ...,
+    default: datetime | None = ...,
+    tzinfos: _TzInfo | None = ...,
+    **kwargs: Any,
```
JelleZijlstra, line 62:
```suggestion
```

JelleZijlstra (commented):
Thanks!

stubs/python-dateutil/dateutil/parser/__init__.pyi
sobolevn: (Technically someone might change what `DEFAULTPARSER` is and that's why we might need `**kwargs`, but we can ignore it probably). https://github.com/dateutil/dateutil/blob/master/src/dateutil/parser/_parser.py#L1368

JelleZijlstra (approved):

PR #9213, opened 2022-11-16
Always use `bool` and `Literal` for Python compat code

sobolevn:
When values used for python compat are defined, they are always `bool`. 
But, we can also highlight that some versions are not supported at all: by marking them as `Literal[False]`.

### DIFF
--- a/stubs/SQLAlchemy/sqlalchemy/util/compat.pyi
+++ b/stubs/SQLAlchemy/sqlalchemy/util/compat.pyi
@@ -12,6 +12,7 @@ from io import BytesIO as BytesIO, StringIO as StringIO
 from itertools import zip_longest as zip_longest
 from time import perf_counter as perf_counter
 from typing import TYPE_CHECKING as TYPE_CHECKING, Any, NamedTuple
+from typing_extensions import Literal
 from urllib.parse import (
     parse_qsl as parse_qsl,
     quote as quote,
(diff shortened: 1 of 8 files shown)


AlexWaygood:
Some of these go against the policy we established in #7258

sobolevn:
@AlexWaygood I don't think it is:
1. There are exactly the same annotations in multiple other places
2. These are real `Literal` constants that affect code reachability, that's their main goal

AlexWaygood:
> 1. There are exactly the same annotations we have in multiple other places

We agreed in #7258 that we'd pause converting global constants to `Literal` types. We didn't say that we wanted to revert all of the global constants that already use `Literal` types. That's going to necessarily lead to a certain amount of inconsistency.

> 2. These are real `Literal` constants that affect code reachability, that's their main goal

I don't have a problem with these constants being included in the stubs, since it's plausible that some users might be using them and/or find them useful, and I don't think they do much harm. But I think they're mostly _intended_ as internal implementation details rather than things that are meant to be publicly exported.

sobolevn:
> But I think they're mostly intended as internal implementation details rather than things that are meant to be publicly exported.

It depends, at least 3rd-party plugins for sqlalchemy and requests might be using them to have similar features support for different languages.

AlexWaygood:
> > But I think they're mostly intended as internal implementation details rather than things that are meant to be publicly exported.
> 
> It depends, at least 3rd-party plugins for sqlalchemy and requests might be using them to have similar features support for different languages.

That's an interesting point. Are 3rd-party plugins for `requests` and `sqlalchemy` a thing?

srittau (approved):
I don't think the reasoning from #7258 applies here. From rereading that issue, the main problem was "enum-like" literals. When assigning such a literal to a field or variable and later re-assigning the variable, type checkers will complain. I can't really imagine situations where you would want to assign these "flag-like" literals and later change the flag.

AlexWaygood:
Alrighty, let's do it then

PR #9214, opened 2022-11-16
Mark regex as completed

sobolevn:


JelleZijlstra:
Stubtest finds a bunch of missing names, could you take a look at the CI?

sobolevn:
Yes, turns out this is quite complex. This package had a recent rewrite. I am working on fixing stubs right now.

sobolevn:
@JelleZijlstra friendly ping :)

Looks like this is a real problem in `paroxythin`: https://github.com/laowantong/paroxython/blob/a6d45829dd34f046d20e5bae780fbf7af59429cb/paroxython/map_taxonomy.py#L171-L178

The type annotation uses `typing.Pattern` and the value uses `regex.Pattern`.
I don't think we should support their mixing.

stubs/regex/regex/regex.pyi
```diff
@@ -309,6 +315,336 @@ def cache_all(value: bool = ...) -> None: ...
 @overload
 def cache_all(value: None) -> bool: ...
 def escape(pattern: AnyStr, special_only: bool = ..., literal_spaces: bool = ...) -> AnyStr: ...
-def template(pattern: AnyStr | _regex.Pattern[AnyStr], flags: int = ...) -> _regex.Pattern[AnyStr]: ...
+def template(pattern: AnyStr | Pattern[AnyStr], flags: int = ...) -> Pattern[AnyStr]: ...
 
 Regex = compile
+
+@final
+class Pattern(Generic[AnyStr]):
```
JelleZijlstra, line 323: At runtime this class is defined in `regex._regex`, so why did you move it?

sobolevn: It is not exported from `C` module, but redefined here: https://github.com/mrabarnett/mrab-regex/blob/cc392629b28ed0aeb8bf6d46f826ff9e87aac551/regex_3/regex.py#L732-L733

AlexWaygood: It still ultimately _comes_ from `regex._regex`, though. The fact that the author of the package chooses to expose the type through the `type(compile())` call rather than through a direct import doesn't seem to me to be _that_ relevant. I think it makes more sense to leave the class defined in `_regex.pyi` :)

sobolevn: This is why I don't agree:

```
>>> from regex._regex import Pattern
ImportError: cannot import name 'Pattern' from 'regex._regex' (/Users/sobolev/Desktop/typeshed/.venv/lib/python3.10/site-packages/regex/_regex.cpython-310-darwin.so)
>>> from regex.regex import Pattern
```

I think that `__qualname__` is less important than import here.

AlexWaygood: Oh, I see. In that case, the move does seem like a good idea to me. But maybe worth adding a comment saying why we have it in this file rather than `_regex.pyi`?

sobolevn: Sure!

PR #9215, opened 2022-11-16
builtins: make type.__call__ return object

JelleZijlstra:
Helps with python/mypy#10482


JelleZijlstra:
This is too disruptive, mypy will need an alternative solution.

PR #9217, opened 2022-11-16
Fix SimplePath protocol

layday:
`pathlib.Path` nor `zipfile.Path` were assignable to `SimplePath`.


layday:
Does mypy not support `Self` or what's the issue here?

sobolevn:
> Does mypy not support Self or what's the issue here?

Not in `0.99x`

layday:
I thought this would be a quick fix.

I was sorely mistaken.

layday:
If the `PackagePath` inherits from `pathlib.PurePosixPath`, I have to wonder why `SimplePath` exists at all, or how are you gonna get `PurePosixPath`s from arbitrary `SimplePath`s?

stdlib/importlib/metadata/__init__.pyi
```diff
@@ -123,7 +129,7 @@ class Distribution:
     @abc.abstractmethod
     def read_text(self, filename: str) -> str | None: ...
     @abc.abstractmethod
-    def locate_file(self, path: StrPath) -> PathLike[str]: ...
+    def locate_file(self, path: _TSimplePath) -> _TSimplePath: ...
```
JelleZijlstra, line 132: This would imply a guarantee that it returns the exact same type it was passed, which I don't think makes sense.

stdlib/importlib/metadata/__init__.pyi
```diff
@@ -170,15 +176,15 @@ class DistributionFinder(MetaPathFinder):
...
         def invalidate_caches(cls) -> None: ...
 
-class PathDistribution(Distribution):
-    def __init__(self, path: Path) -> None: ...
-    def read_text(self, filename: StrPath) -> str: ...
-    def locate_file(self, path: StrPath) -> PathLike[str]: ...
+class PathDistribution(Distribution, Generic[_TSimplePath]):
+    def __init__(self, path: _TSimplePath) -> None: ...
+    def read_text(self, filename: str) -> str: ...
+    def locate_file(self, path: _TSimplePath) -> _TSimplePath: ...  # type: ignore[override]
```
JelleZijlstra, line 187: This seems incorrect. It implies that a PathDistribution constructed with a Path as the argument to `__init__` will only accept a Path as the argument to `locate_file`, but that's not right:

```
In [47]: p = importlib.metadata.PathDistribution(Path("x"))

In [48]: p.locate_file("y")
Out[48]: PosixPath('y')
```

stdlib/importlib/metadata/_meta.pyi
```diff
@@ -12,11 +13,11 @@ class PackageMetadata(Protocol):
...
+# to common path-likes like `pathlib.Path` and `zipfile.Path`.
 class SimplePath(Protocol):
-    def joinpath(self) -> SimplePath: ...
-    def parent(self) -> SimplePath: ...
-    def read_text(self) -> str: ...
-    # There was a bug in `SimplePath` definition in cpython, see #8451
-    #  Strictly speaking `__div__` was defined in 3.10, not __truediv__,
-    # but it should have always been `__truediv__`.
-    def __truediv__(self) -> SimplePath: ...
+    def joinpath(self: Self, *other: str | Self) -> Self: ...
```
JelleZijlstra, line 19: I would not use `Self` here: that makes the protocol stricter than necessary. Instead, I would only require that it returns an instance of `SimplePath`.

stdlib/importlib/metadata/_meta.pyi
```diff
@@ -12,11 +13,11 @@ class PackageMetadata(Protocol):
     @property
     def json(self) -> dict[str, str | list[str]]: ...
 
+# Protocol deviates from runtime implementation.  The latter is not assignable
```
JelleZijlstra, line 16: Could you add test cases to show that this Protocol is indeed compatible with types like `pathlib.Path`?

https://github.com/python/typeshed/tree/main/test_cases

stdlib/importlib/metadata/__init__.pyi
JelleZijlstra: I'm referring to the fact that I passed a string, not the Path/PosixPath change.

stdlib/importlib/metadata/__init__.pyi
layday: That's true, the `path` should be the same type as passed to the constructor but there's no guarantee that the return value would be of the same type, e.g. if the distribution would extract the contents of a zip file.

stdlib/importlib/metadata/__init__.pyi
layday: whoops, didn't see you replied and deleted the previous message. Yes, I realised afterwards 😅

hauntsaninja:
Looks like https://github.com/Rapptz has been deleted, let me replace it with a fork...

layday:
The tests are failing because `zipfile.Path` is typed as accepting and returning file system paths.

layday:
The `filename` property of `zipfile.Path`, besides being [undocumented](https://docs.python.org/3/library/zipfile.html#path-objects), actually returns a file system path: https://github.com/python/cpython/blob/56a517e9e854559e2f0c23c0e5adcaf846a2e0c3/Lib/zipfile.py#L2420.

JelleZijlstra:
This has a lot of merge conflicts now. I created a simpler alternate PR at #11436. We can apply the rest of the changes in this PR separately later.

layday:
Thank you, closing this.

PR #9220, opened 2022-11-17
click_spinner: make arguments optional

hsheth2:
Matches the code, where these arguments are not required: https://github.com/click-contrib/click-spinner/blob/master/click_spinner/__init__.py.


AlexWaygood (approved):
Thanks!

hauntsaninja (approved):

PR #9223, opened 2022-11-18
Narrower return type for `east_asian_width()`

dahlia:
According to [UAX #11], there are 6 kinds of possible values for property `East_Asian_Width`:

- `F`: East Asian Fullwidth
- `H`: East Asian Halfwidth
- `W`: East Asian Wide
- `Na`: East Asian Narrow
- `A`: East Asian Ambiguous
- `N`: Neutral (Not East Asian)

This patch narrows the return type of `east_asian_width()` using `Literal[...]`.

[UAX #11]: https://www.unicode.org/reports/tr11/#Definitions


stdlib/unicodedata.pyi
```diff
@@ -17,7 +17,10 @@ def combining(__chr: str) -> int: ...
 def decimal(__chr: str, __default: _T = ...) -> int | _T: ...
 def decomposition(__chr: str) -> str: ...
 def digit(__chr: str, __default: _T = ...) -> int | _T: ...
-def east_asian_width(__chr: str) -> str: ...
+
+EastAsianWidth: TypeAlias = Literal["F", "H", "W", "Na", "A", "N"]
```
AlexWaygood, line 21: It needs to have a name starting with a leading underscore, or stubtest will complain about the object not existing at runtime :)

```suggestion
_EastAsianWidth: TypeAlias = Literal["F", "H", "W", "Na", "A", "N"]
```

dahlia: Thank you for guiding me.  I amended the commit.  Could you give it a look again?

hauntsaninja (approved):
Thanks!

PR #9225, opened 2022-11-19
python-slugify: use explicit re-exports

AlexWaygood:
Fixes https://github.com/python/typeshed/pull/9224#discussion_r1027053364


hauntsaninja:
Thanks for fixing!

PR #9226, opened 2022-11-19
Mark `python-nmap` as completed

sobolevn:
Source: https://bitbucket.org/xael/python-nmap/src/master/nmap/nmap.py


AlexWaygood (approved):
One more down ✅️

PR #9228, opened 2022-11-20
Add type-hints for _vformat in string.Formatter

ischaojie:
Add type-hints for `_vformat` method in `string.Formatter`.

When add type-hints for
https://github.com/pypa/hatch/blob/b0911bb0eaa8d331c24eda940b97bf244ecd5ac3/backend/src/hatchling/utils/context.py#LL119 part,  mypy report `Cannot access member "_vformat" for type` error.

But `_vformat` not appear in API document, I'm not sure should add this.


stdlib/string.pyi
```diff
@@ -64,6 +64,15 @@ class Formatter:
...
     def vformat(self, format_string: str, args: Sequence[Any], kwargs: Mapping[str, Any]) -> str: ...
+    def _vformat(
+        self,
+        format_string: str,
+        args: Sequence[Any],
+        kwargs: Mapping[str, Any],
+        used_args: Sequence[int | str],
+        recursion_depth: int,
+        auto_arg_index: int = ...,
+    ) -> str | bool: ...
```
JelleZijlstra, line 75:
```suggestion
    ) -> str: ...
```
Don't see anything in the code that returns a bool.

stdlib/string.pyi
```diff
@@ -64,6 +64,15 @@ class Formatter:
     ) -> LiteralString: ...
     @overload
     def vformat(self, format_string: str, args: Sequence[Any], kwargs: Mapping[str, Any]) -> str: ...
+    def _vformat(
+        self,
+        format_string: str,
+        args: Sequence[Any],
+        kwargs: Mapping[str, Any],
+        used_args: Sequence[int | str],
```
JelleZijlstra, line 72:
```suggestion
        used_args: set[int | str],
```
It does `used_args.add`.

stdlib/string.pyi
```diff
@@ -64,6 +64,15 @@ class Formatter:
     ) -> LiteralString: ...
     @overload
     def vformat(self, format_string: str, args: Sequence[Any], kwargs: Mapping[str, Any]) -> str: ...
+    def _vformat(
```
JelleZijlstra, line 67:
```suggestion
    def _vformat(  # undocumented
```
Following our convention.

stdlib/string.pyi
ischaojie: I notice `self.check_unused_args(used_args, args, kwargs)` in method `vformat`, but :
```pyi
def check_unused_args(self, used_args: Sequence[int | str], args: Sequence[Any], kwargs: Mapping[str, Any]) -> None: ...
```
Maybe need add `@overload` for `check_unused_args` stub?

stdlib/string.pyi
ischaojie: Thanks, I misread.

stdlib/string.pyi
JelleZijlstra: I think that annotation is just wrong. It should also be `set[int | str]`.

stdlib/string.pyi
ischaojie: resolved.

stdlib/string.pyi
ischaojie: resolved.

JelleZijlstra (approved):

PR #9234, opened 2022-11-21
colorama: re-export just_fix_windows_console at top level

njsmith:
Added here: https://github.com/tartley/colorama/pull/352

Re-export can be seen here:

https://github.com/tartley/colorama/blob/21c4b94fe21ce29c85c896ace828da24b7527641/colorama/__init__.py#L2

I don't know why all the exports are "X as X" instead of just "X", but I followed the same style for consistency.


AlexWaygood (approved):
Thanks!

> I don't know why all the exports are "X as X" instead of just "X", but I followed the same style for consistency.

In stubs, all imports are generally considered by type checkers to be "private to the stub" (I.e., they generally _don't_ represent a symbol that's being exported, that an end user of the stubs would be able to access). Using the "redundant" `from foo import bar as bar` idiom is one of several ways PEP 484 lays out to indicate to a type checker that _these_ imports _are_ in fact public re-exports that an end user would be able to use at runtime.

PR #9235, opened 2022-11-21
Mark pynput as completed

sobolevn:
Source: https://github.com/moses-palmer/pynput/tree/master/lib/pynput/_util


sobolevn:
Looks like it might be a good idea to run this package on all platforms 🤔

sobolevn:
I've sent https://github.com/moses-palmer/pynput/pull/519 to `pynput` so we can later delete `stubtest_allowlist_*.txt` files with `__init__` special casing.

PR #9237, opened 2022-11-21
Remove `slumber` from pyright exclude

sobolevn:
This code looks quite outdated. But, it still works.
Source: https://github.com/samgiles/slumber


stubs/slumber/slumber/__init__.pyi
```diff
@@ -1,30 +1,39 @@
...
-    def as_raw(self): ...
-    def get(self, **kwargs): ...
-    def options(self, **kwargs): ...
-    def head(self, **kwargs): ...
-    def post(self, data: Any | None = ..., files: Any | None = ..., **kwargs): ...
-    def patch(self, data: Any | None = ..., files: Any | None = ..., **kwargs): ...
-    def put(self, data: Any | None = ..., files: Any | None = ..., **kwargs): ...
-    def delete(self, **kwargs): ...
-    def url(self): ...
+    def __init__(self, *args: Any, **kwargs: Any) -> None: ...
```
srittau, line 16: Since `args` is ignored:

```suggestion
    def __init__(self, *args: object, **kwargs: Any) -> None: ...
```

srittau (changes requested):
Thanks, one nit below.

stubs/slumber/slumber/__init__.pyi
sobolevn: This is intentional. This is designed to be subclassed, as far as I understand. And it can be used in subclasses. 

It is easier to redefine `*Any, **Any` than `*object, **Any` from my perspective.

PR #9230, opened 2022-11-20
Fix typings of `ExceptionGroup` and `BaseExceptionGroup`

sobolevn:
There are a lot of things going on in this PR:
1. Methods no longer return `Self` type, because of how `C` implementation works
2. `split` and `subgroup` require `Self` argument to be in callable, because of how `C` implementation works
3. Some casts from `BaseExceptionGroup` to and from `ExceptionGroup` are supported. But not all of them

This is a very complex PR. I would love to provide any of my help to the reviewers.

PR #9238, opened 2022-11-21
Allow adding test cases testing version-dependent Python features

AlexWaygood:
It's currently impossible to add a test case for `ExceptionGroup` or `BaseExceptionGroup`: see #9230. You'd think that you'd be able to just put the test behind an `if sys.version_info >= (3, 11)` guard, and pyright is indeed fine with that. Mypy, however, is not: when you run the test case with `--python-version 3.10`, it complains about many "unused type ignore" comments (it detects the `type: ignore` comments behind the `if sys.version_info >= (3, 11)` guard, even though it considers the whole block unreachable with `--python-version 3.10`, meaning it considers all `type: ignore`s within that block to be "unused").

This PR works around that mypy annoyance. If this PR is merged, the test-case file can simply be renamed to be `test_cases/stdlib/builtins/check_exception_group-py311.py`, and `python tests/regr_test.py --all` should pass.

Cc. @sobolevn

### DIFF
--- a/tests/regr_test.py
+++ b/tests/regr_test.py
@@ -136,8 +136,9 @@ def test_testcase_directory(package: PackageInfo, version: str, platform: str) -
         # only run the test if --python-version was set to 3.9 or higher (for example)
         for path in new_test_case_dir.rglob("*.py"):
             if match := re.fullmatch(r".*-py3(\d{1,2})", path.stem):
-                assert f"3.{match[1]}" in SUPPORTED_VERSIONS
-                if int(match[1]) <= int(version.split(".")[1]):
+                minor_version_required = int(match[1])
+                assert f"3.{minor_version_required}" in SUPPORTED_VERSIONS
+                if minor_version_required <= int(version.split(".")[1]):
                     flags.append(str(path))
             else:
                 flags.append(str(path))
--- a/tests/regr_test.py
+++ b/tests/regr_test.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 
 import argparse
 import os
+import re
 import shutil
 import subprocess
 import sys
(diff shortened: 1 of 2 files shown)


tests/regr_test.py
```diff
@@ -130,7 +131,18 @@ def test_testcase_directory(package: PackageInfo, version: str, platform: str) -
...
             env_vars["MYPYPATH"] = os.pathsep.join(map(str, new_typeshed.glob("stubs/*")))
             flags.extend(["--custom-typeshed-dir", str(td_path / "typeshed")])
-        result = subprocess.run([sys.executable, "-m", "mypy", new_test_case_dir, *flags], capture_output=True, env=env_vars)
+
+        # If the test-case filename ends with -py39,
+        # only run the test if --python-version was set to 3.9 or higher (for example)
+        for path in new_test_case_dir.rglob("*.py"):
+            if match := re.fullmatch(r".*-py3(\d{1,2})", path.stem):
+                assert f"3.{match[1]}" in SUPPORTED_VERSIONS
+                if int(match[1]) <= int(version.split(".")[1]):
```
AlexWaygood, line 140:
```suggestion
                minor_version_required = int(match[1])
                assert f"3.{minor_version_required}" in SUPPORTED_VERSIONS
                if minor_version_required <= int(version.split(".")[1]):
```

sobolevn (approved):
This looks like a good work-around!
Ideally we can teach mypy not to trigger `unused ignore` on unreachable code as well.

Thanks!

JelleZijlstra (approved):
I suppose this won't work if we want test cases for things that were *removed* in a later version, but we'll have to figure out an approach for that when we get to it.

AlexWaygood:
> I suppose this won't work if we want test cases for things that were _removed_ in a later version, but we'll have to figure out an approach for that when we get to it.

Here's hoping nobody proposes to add test cases for things that have already been deprecated and removed in released versions of Python :)

PR #9248, opened 2022-11-22
`__unicode__` always returns `str`

sobolevn:
I continue to improve basic magic methods across all source code.
`__unicode__` always returns `str`, despite the fact it is not used in python3.


stubs/html5lib/html5lib/treewalkers/etree_lxml.pyi
```diff
@@ -41,7 +41,7 @@ class FragmentWrapper:
     def __getitem__(self, key): ...
     def __bool__(self) -> bool: ...
     def getparent(self) -> None: ...
-    def __unicode__(self): ...
+    def __unicode__) -> str: ...
```
AlexWaygood, line 44:
```suggestion
    def __unicode__(self) -> str: ...
```

AlexWaygood (approved):

PR #9249, opened 2022-11-22
tqdm: Add missing tqdm.monitor class variable

yodaldevoid:
I'm not entirely sure why this was not caught before as the class variable has been there since 2016.

### DIFF
--- a/stubs/tqdm/tqdm/std.pyi
+++ b/stubs/tqdm/tqdm/std.pyi
@@ -4,6 +4,7 @@ from collections.abc import Callable, Iterable, Iterator, Mapping, MutableMappin
 from typing import Any, ClassVar, Generic, NoReturn, TypeVar, overload
 from typing_extensions import Literal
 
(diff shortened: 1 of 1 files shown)


AlexWaygood (approved):
> I'm not entirely sure why this was not caught before as the class variable has been there since 2016.

The stubs are a fair bit newer, however :)

yodaldevoid:
> > I'm not entirely sure why this was not caught before as the class variable has been there since 2016.
> 
> The stubs are a fair bit newer, however :)

My first thought was that the stubs were created after the class variable was added, but as you point out that is not the case. Oh well, the world has many mysteries and this is one of them.

AlexWaygood:
Oh, my point was simply that the `tqdm` stubs are a relatively new addition to typeshed, so it's sort-of to be expected that there might be still be a few omissions in the stubs :) It can take a little bit of time for issues like this to get spotted.

PR #9250, opened 2022-11-22
ctypes: Modified paramflags and iid of _FuncPointer and updated _PF.

### LINKED ISSUE #8968: The constructor of `_FuncPointer` can accept `None` to `paramflags` and `Structure` to `iid` in runtime.
https://github.com/python/typeshed/blob/dc7fdc6f5dc19935845eb407028ef86abe6c691a/stdlib/ctypes/__init__.pyi#L115-L117

https://github.com/python/typeshed/blob/dc7fdc6f5dc19935845eb407028ef86abe6c691a/stdlib/ctypes/__init__.pyi#L91-L103

This is simplified my project code.
And it can run in Py2.7 and Py3.3-3.10.
```py
            # argyypes: tuple[type[_CData], ...]
            # vtbl_offset: int
            # i: int
            # _iid_: subclass of `Structure` instance, not a pointer of that.
            # paramflags: tuple[...] | None
            prototype = WINFUNCTYPE(restype, *argtypes)
            if restype == HRESULT:
                ...
                raw_func = prototype(i + vtbl_offset, name, None, _iid_)   # <- pylance error
                func = prototype(i + vtbl_offset, name, paramflags, _iid_)   # <- pylance error
                ...
```

Probably this is a false-positive.

I think that line 102 of `ctypes/__init__.pyi` will be fixed as below.
```py
    def __init__(self, vtlb_index: int, name: str, paramflags: tuple[_PF, ...] | None = ..., iid: _CData = ...) -> None: ... 
```

AmberAnsari89:
fixes #8968


stubs/beautifulsoup4/bs4/element.pyi
```diff
@@ -247,7 +247,7 @@ class Tag(PageElement):
         name: str | None = ...,
         namespace: str | None = ...,
         prefix: str | None = ...,
-        attrs: dict[str, str] | None = ...,
+        attrs: dict[str, Any] | None = ...,
```
AlexWaygood, line 250:
```suggestion
        attrs: dict[str, str] | None = ...,
```

AmberAnsari89: is there action from my side on this?

AlexWaygood: No, I've already applied the suggestion here :)

PR #9251, opened 2022-11-22
ctypes: constructor of _FuncPointer can accept None to paramflags and Structure to iid in runtime

### LINKED ISSUE #8968: The constructor of `_FuncPointer` can accept `None` to `paramflags` and `Structure` to `iid` in runtime.

AmberAnsari89:
fixes #8968


stubs/beautifulsoup4/bs4/element.pyi
```diff
@@ -247,7 +247,7 @@ class Tag(PageElement):
         name: str | None = ...,
         namespace: str | None = ...,
         prefix: str | None = ...,
-        attrs: dict[str, str] | None = ...,
+        attrs: dict[str, Any] | None = ...,
```
JelleZijlstra, line 250: Let's leave this change for another PR.

JelleZijlstra:
The ctypes code can be hard to follow, but I think there's more we can do here to improve things. The implementation of the code being changed here is at https://github.com/python/cpython/blob/8f18ac04d32515eab841172c956a8cb14bcee9c3/Modules/_ctypes/_ctypes.c#L3627.

- Our stub has a `_FuncPointer` class that doesn't exist at runtime. But in fact, the code returns a subclass of `_ctypes.CFuncPtr`, so why don't we use that class in the stubs?
- All of the arguments are positional-only and should be marked as such in the stub.
- We're missing an overload for the case where no arguments are passed to the constructor.
- The overload with `vtbl_index` (not `vtlb_index` as the stub calls it) should exist only on Windows.

AmberAnsari89:
thank you @JelleZijlstra  for these information. However, i did not understand much from these.
Can you please guide me a bit more to understand with reference to actual requirement of issue under -- The constructor of _FuncPointer can accept None to paramflags and Structure to iid in runtime. #8968

PR #9255, opened 2022-11-23
Fix `ssl.pyi` and CI failure

sobolevn:


stdlib/ssl.pyi
```diff
@@ -180,6 +180,7 @@ class Options(enum.IntFlag):
     OP_NO_RENEGOTIATION: int
     if sys.version_info >= (3, 8):
         OP_ENABLE_MIDDLEBOX_COMPAT: int
+        OP_IGNORE_UNEXPECTED_EOF: int
```
AlexWaygood, line 183:
```suggestion
        if sys.platform == "linux"
            OP_IGNORE_UNEXPECTED_EOF: int
```

stdlib/ssl.pyi
```diff
@@ -196,6 +197,7 @@ OP_NO_TICKET: Options
 OP_NO_RENEGOTIATION: Options
 if sys.version_info >= (3, 8):
     OP_ENABLE_MIDDLEBOX_COMPAT: Options
+    OP_IGNORE_UNEXPECTED_EOF: Options
```
AlexWaygood, line 200:
```suggestion
    if sys.platform == "linux"
        OP_IGNORE_UNEXPECTED_EOF: Options
```

stdlib/ssl.pyi
```diff
@@ -180,6 +180,8 @@ class Options(enum.IntFlag):
     OP_NO_RENEGOTIATION: int
     if sys.version_info >= (3, 8):
         OP_ENABLE_MIDDLEBOX_COMPAT: int
+        if sys.platform == "linux"
```
AlexWaygood, line 183:
```suggestion
        if sys.platform == "linux":
```

sobolevn:
Oh, thanks for polishing it! I am working on pyright failures.

srittau:
> Thanks! No idea why this started failing randomly on main.

Possibly because GitHub Actions started to use a new Python minor version.

AlexWaygood:
> > Thanks! No idea why this started failing randomly on main.
> 
> Possibly because GitHub Actions started to use a new Python minor version.

That was my first thought, but it doesn't look like it. Since the new failures were all Ubuntu-related, my favoured hypothesis is now that something changed related to github's Ubuntu image.

And none of this explains why pyright started failing at exactly the same time. There's just been a new pyright release, but pyright should be pinned in CI.

srittau:
> Since the new failures were all Ubuntu-related, my favoured hypothesis is now that something changed related to github's Ubuntu image.

Quite possible. Maybe a new Ubuntu minor version with a newer SQLite library version.

sobolevn:
My findings:
- there are no new patch python releases
- `pyright` is always `1.1.128`
- `pyright` failures are related to `cryptography` modules only
- the output is quite strange: `/home/runner/work/typeshed/typeshed/stubs/cryptography/cryptography/x509/extensions.pyi:3:44 - error: "ObjectIdentifier" is not exported from module "cryptography.x509.oid" Import from "cryptography.hazmat._oid" instead (reportPrivateImportUsage)`. The interesting part is that there's no `cryptography.hazmat._oid` part
- `stubtest` failures were related to `ssl` module
- I think the faulires are related to some new `openssl` release

AlexWaygood:
> * `stubtest` failures were related to `ssl` module
> * I think the faulires are related to some new `openssl` release

_Some_ stubtest failures were related to openssl. There were also a bunch of "unused allowlist" errors relating to sqlite3 on Ubuntu on Python 3.11.

AlexWaygood:
> * The interesting part is that there's no `cryptography.hazmat._oid` module in `typeshed`

Is it possible that the runtime `cryptography` package could somehow be being installed in CI? That comes with inline types (which are much more up-to-date than typeshed's legacy stubs), so that could cause all sorts of havoc.

sobolevn:
sqllite also uses openssl 🤔

sobolevn:
> Is it possible that the runtime cryptography package could somehow be being installed in CI?

Yes, this is my theory right now.

PR #9258, opened 2022-11-23
Exclude `stubs/cryptoghraphy` from CI to fix pyright

sobolevn:
As far as I understand `stubs/cryptography` are outdated.
We ignore it in our `stubtest` runs, but not in our `pyright` runs.

I think these stubs should be ignored for now. And removed later on.

Refs
- https://github.com/python/typeshed/issues/5618
- https://github.com/python/typeshed/issues/9127

### DIFF
--- a/pyrightconfig.json
+++ b/pyrightconfig.json
@@ -5,6 +5,10 @@
         "stdlib",
         "stubs"
     ],
+    "exclude": [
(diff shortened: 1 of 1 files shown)


AlexWaygood (approved):
Good idea.

PR #9261, opened 2022-11-23
Use modern syntax in test-case files where possible

AlexWaygood:
By default, pyupgrade only rewrites `.py` files, so adding it as a pre-commit hook should have no effect on the `stubs/` or `stdlib/` directories.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,9 +1,4 @@
 repos:
-  - repo: https://github.com/asottile/pyupgrade
-    rev: v3.2.2 # must match requirements-tests.txt
-    hooks:
-      - id: pyupgrade
-        args: [--py37-plus]
   - repo: https://github.com/hadialqattan/pycln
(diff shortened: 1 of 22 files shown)


.pre-commit-config.yaml
```diff
@@ -1,4 +1,9 @@
 repos:
+  - repo: https://github.com/asottile/pyupgrade
```
AlexWaygood, line 2:
```suggestion
  # this only runs on the .py files
  - repo: https://github.com/asottile/pyupgrade
```

AlexWaygood:
The stubtest failures in `pycurl` are unrelated (presumably we'll have a daily test failure this evening).

hauntsaninja (commented):
Not sure we want to run this automatically,e.g. we're changing what we're type checking in the check_post example

AlexWaygood:
> Not sure we want to run this automatically,e.g. we're changing what we're type checking in the check_post example

Yeah, that's fair.

PR #9262, opened 2022-11-23
`pycurl`: Add a few missing constants

AlexWaygood:
Fixes the stubtest failure seen in #9261


stubs/pycurl/pycurl.pyi
```diff
@@ -98,8 +98,12 @@ CURL_HTTP_VERSION_LAST: int
 CURL_HTTP_VERSION_NONE: int
 CURL_VERSION_ALTSVC: int
 CURL_VERSION_BROTLI: int
+CURL_VERSION_GSAL: int
```
AlexWaygood, line 101:
```suggestion
CURL_VERSION_GASAL: int
```

AlexWaygood:
Going to skip review so we don't get a daily test failure in 30 minutes (and I think this is fairly unobjectionable :)

PR #9265, opened 2022-11-24
check_consistent.py: Add check ensuring packages are not installed for unspecified platforms

sobolevn:
I've missed this during review, good ideas always come late.

Right now we can install deps if they are specified for any platform.
Even if it is not tested. This can consume time of workers and lead to errors.

This PR allows:

```toml
platforms = ["linux", "win32"]
apt_dependencies = ["one"]
# brew_dependencies = ["some"]
```

```toml
platforms = ["linux", "win32"]
brew_dependencies = []
```

(I can remove ^ this)

But not:

```toml
platforms = ["linux", "win32"]
brew_dependencies = ["some"]
```

Refs https://github.com/python/typeshed/pull/9187

### DIFF
--- a/tests/check_consistent.py
+++ b/tests/check_consistent.py
@@ -198,9 +198,7 @@ def check_metadata() -> None:
         # Check that only specified platforms install packages:
         for supported_plat in supported_stubtest_platforms:
             if supported_plat not in specified_stubtest_platforms:
-                assert not tool_stubtest.get(
-                    metadata_mapping[supported_plat]
-                ), f"Installing system deps for unspecified platform {supported_plat} for {distribution}"
+                assert metadata_mapping[supported_plat] not in tool_stubtest, f"Installing system deps for unspecified platform {supported_plat} for {distribution}"
 
 
 def get_txt_requirements() -> dict[str, SpecifierSet]:
--- a/tests/check_consistent.py
+++ b/tests/check_consistent.py
@@ -16,10 +16,10 @@ from packaging.requirements import Requirement
 from packaging.specifiers import SpecifierSet
 from packaging.version import Version
 from utils import (
+    METADATA_MAPPING,
     VERSIONS_RE,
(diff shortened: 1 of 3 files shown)


tests/check_consistent.py
```diff
@@ -182,11 +189,19 @@ def check_metadata() -> None:
...
         assert (
             specified_stubtest_platforms <= supported_stubtest_platforms
         ), f"Unrecognised platforms specified: {supported_stubtest_platforms - specified_stubtest_platforms}"
 
+        # Check that only specified platforms install packages:
+        for supported_plat in supported_stubtest_platforms:
+            if supported_plat not in specified_stubtest_platforms:
+                assert not tool_stubtest.get(
+                    metadata_mapping[supported_plat]
+                ), f"Installing system deps for unspecified platform {supported_plat} for {distribution}"
```
AlexWaygood, line 203: (black will almost certainly reformat this)

```suggestion
                assert metadata_mapping[supported_plat] not in tool_stubtest, f"Installing system deps for unspecified platform {supported_plat} for {distribution}"
```

tests/utils.py
```diff
@@ -12,6 +12,9 @@
 import pathspec  # type: ignore[import]
 import tomli
 
+# Used to install system-wide packages for different OS types:
+metadata_mapping = {"linux": "apt_dependencies", "darwin": "brew_dependencies", "win32": "choco_dependencies"}
```
AlexWaygood, line 16: Nit: maybe use SCREAM_CASE for this, if it's a constant imported from multiple files?

```suggestion
METADATA_MAPPING = {"linux": "apt_dependencies", "darwin": "brew_dependencies", "win32": "choco_dependencies"}
```

AlexWaygood (commented):
Nice idea!

AlexWaygood (approved):
Thanks!

PR #9270, opened 2022-11-24
fpdf2:graphics-state.pyi changes

AmberAnsari89:
i added return types for first 3 @property method. Please verify the same.

i was wondering why variables inside the __init__ method are not annotated and also some initial methods are left out in graphics_state.pyi.  

Thanks & regards
Amber


AlexWaygood:
Because of the fact that you were directly editing your `main` branch in a previous PR, your `main` branch has got out of sync with typeshed's `main` branch. This means that your previous changes to `ctypes` and `bs4` are showing up in this PR.

The best way to get your `main` branch back in sync with typeshed's `main` is to do the following:

1. Close this PR
2. Check out your `main` branch locally  
3. Run the following commands:
    ```
    git checkout main
    git pull upstream main
    git reset --hard upstream/main
    git push origin main --force
    ```
4. Recreate this PR.

AmberAnsari89:
Thank you Alex... i will try to do using command. here i will get something
new and technical ..

PR #9276, opened 2022-11-25
Add color back to CI output

### LINKED ISSUE #9204: CI output is no longer colorised
See e.g. https://github.com/python/typeshed/actions/runs/3472143680/jobs/5802584010. I haven't done any investigation yet as to what the cause might be.

AlexWaygood:
Fixes #9204

### DIFF
--- a/.github/workflows/daily.yml
+++ b/.github/workflows/daily.yml
@@ -19,6 +19,7 @@ concurrency:
 
 env:
   PIP_DISABLE_PIP_VERSION_CHECK: 1
+  FORCE_COLOR: 1
 
 jobs:
   stubtest-stdlib:
--- a/.github/workflows/stubtest_third_party.yml
+++ b/.github/workflows/stubtest_third_party.yml
@@ -13,6 +13,7 @@ permissions:
 
 env:
   PIP_DISABLE_PIP_VERSION_CHECK: 1
+  FORCE_COLOR: 1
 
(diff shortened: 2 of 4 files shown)


AlexWaygood:
Still don't know exactly what changed here, but this seems to fix it!

srittau (approved):

JelleZijlstra (approved):

AlexWaygood:
Not sure what's taking the third-party stubtest on MacOS so long -- it should only be running on four stubs packages:

```pycon
>>> import tomli
>>> from pathlib import Path
>>> for path in Path("stubs").iterdir():
...     metadata = path / "METADATA.toml"
...     if "darwin" in tomli.loads(metadata.read_text(encoding="utf-8")).get("tool", {}).get("stubtest", {}).get("platforms", ["linux"]):
...         print(path.parts[-1])
...
JACK-Client
psutil
pyaudio
pynput
```

JelleZijlstra:
Yes, GitHub seems generally laggy at the moment. Maybe the runners ate too much Thanksgiving turkey. I'd recommend waiting a while longer and re-triggering CI if it doesn't eventually show up.

AlexWaygood:
Yes, looks like the `brew install` step failed in the macOS stubtest run. But even if that's not just a laggy-GitHub fluke, I'm confident it won't have been caused by this PR.

Issue #9277, opened 2022-11-25
Add method Catalog.is_identical for babel package


hyzyla:
Seems like `is_identacal` method of `Catalog` from `babel` package is missing

<img width="1792" alt="image" src="https://user-images.githubusercontent.com/7952949/204035583-a7868f28-9695-4dec-b39a-bc295979d4f9.png">

AlexWaygood:
Yes, looks like both `Catalog` and `Message` are missing the `is_identical()` method they have at runtime. PR welcome!

PR #9278, opened 2022-11-25
Add  methods for Catalog and Message from babel package

hyzyla:
Fixes https://github.com/python/typeshed/issues/9277


AlexWaygood (approved):
Thanks!

PR #9282, opened 2022-11-26
fpdf2:graphics-state.pyi changes v1.0

AmberAnsari89:
i added return types for first 3 https://github.com/Property method. Please verify the same.

i was wondering why variables inside the init method are not annotated and also some initial methods are left out in graphics_state.pyi.

Thanks & regards
Amber


AlexWaygood (commented):
Thanks @AmberAnsari89!

I'm not sure that any of these properties can ever return `None`, however. What's the evidence for that? It looks to me like they all always return `DeviceGray` rather than `DeviceGray | None`.

AmberAnsari89:
Thanks @AlexWaygood  for your feedback.
i think your are right.As it is returning the instance of DeviceGray and cannot be None. I was taking into consideration alpha component(a) of drawing.py. However it was for opacity and nothing related to returning of draw_color() and others.

Please confirm. i will do the other changes accordingly resubmit to verification.

AmberAnsari89:
please also confirm whether typing hint for the _statestack is not not necessary.
 def __init__(self, *args, **kwargs):
        self.__statestack = [
            dict(

AlexWaygood:
> please also confirm whether typing hint for the _statestack is not not necessary.
> def **init**(self, *args, **kwargs):
> self.__statestack = [
> dict(

You're correct — we shouldn't add an annotation for that attribute, since it's a private attribute :)

AmberAnsari89:
> 

thank you..i added PR .please verify and further i will speed up for other changes.

AlexWaygood (commented):
Now that I look at it again, it looks like `draw_color` can also sometimes be `DeviceRGB` as well: https://github.com/PyFPDF/fpdf2/blob/c0ec2d79e3475033137dc3cd6f9b5b6f48b1819e/fpdf/fpdf.py#L891-L894

So perhaps the annotations for these should all be `DeviceRGB | DeviceGray`. Could you change them to that, please? You'll have to add `from .drawing import DeviceRGB` at the top of the file, as well.

AlexWaygood (approved):
Thanks!

AlexWaygood:
> can i continue adding types hinting

Of course. No need to ask for approval. PRs like this are welcome.

AmberAnsari89:
> > can i continue adding types hinting
> 
> Of course. No need to ask for approval. PRs like this are welcome.

Thank you for your words.

PR #9284, opened 2022-11-27
Update redis to 4.3.5

sobolevn:
I've copied type annotations from the `redis` source code: 
- https://github.com/redis/redis-py/blob/master/redis/commands/timeseries/commands.py

Closes https://github.com/python/typeshed/pull/9252


hauntsaninja (approved):
Only spot checked, because copying from upstream type hints is likely to be right :-)

Issue #9216, opened 2022-11-16
cffi.FFI.dlopen() does not differ depending on sys.platform


jakkdl:
For some reason
https://github.com/python/typeshed/blob/7fc9b8e2c8fc7942847103c53418a90050f65791/stubs/cffi/cffi/api.pyi#L28

is not defined the same as
https://github.com/python/typeshed/blob/9137258cf43d07cf575c8d3fc15f9c979bd42a14/stubs/cffi/_cffi_backend.pyi#L165-L168

which I'm pretty sure is an error. https://cffi.readthedocs.io/en/latest/cdef.html?highlight=None#ffi-dlopen-loading-libraries-in-abi-mode

> `Help on method dlopen in module cffi.api:`
> `dlopen(name, flags=0) method of cffi.api.FFI instance`
    Load and return a dynamic library identified by 'name'.
    The standard C library can be loaded by passing **None**.
    Note that functions and types declared by 'ffi.cdef()' are not
    linked to a particular library, just like C headers; in the
    library we only look for the actual (untyped) symbols.

I haven't checked the other functions, but I suspect there may be more functions that for some strange reason differs in their typing between the files.

srittau:
Fixed by #10042.

PR #9285, opened 2022-11-27
Fix `CData` and `CType` types in `cffi`

sobolevn:
These attributes are classes, not instances:

```python
>>> import cffi.api
>>> f = cffi.api.FFI()

>>> f.CData
<class '_cffi_backend._CDataBase'>
>>> f.CType
<class '_cffi_backend.CType'>
>>> f.buffer
<class '_cffi_backend.buffer'>

>>> f.NULL
<cdata 'void *' NULL>
```

Source: https://foss.heptapod.net/pypy/cffi/-/blob/branch/default/c/_cffi_backend.c#L7378-7382
Refs https://github.com/python/typeshed/issues/9216


AlexWaygood:
@freundTech, could I trouble you for a review on this one, as the original author of the stubs? :)

JelleZijlstra:
The existing stub defines them as type aliases. Isn't that correct?

JelleZijlstra:
Closing as it's not clear what problem this solves, my question above wasn't answered, and there are merge conflicts.

PR #9290, opened 2022-11-28
Add several return types to known magic methods

sobolevn:
For some reason this method was marked to return `None`: https://github.com/sqlalchemy/sqlalchemy/blob/db2344b0a2a9ef164651d645a8da2d7a9d1bc250/lib/sqlalchemy/sql/elements.py#L698


hauntsaninja (approved):
Thanks!

PR #9291, opened 2022-11-28
Fix `__instancecheck__` definitions

sobolevn:
For `abc` case, it really returns `bool`:
1. https://github.com/python/cpython/blob/594de165bf2f21d6b28eb17003ea78fc20c0ffed/Lib/abc.py#L123
2. https://github.com/python/cpython/blob/594de165bf2f21d6b28eb17003ea78fc20c0ffed/Modules/_abc.c#L795


AlexWaygood (approved):
The [data model docs](https://docs.python.org/3.10/reference/datamodel.html#customizing-instance-and-subclass-checks) also seem to confirm that these should usually return `bool` in subclasses!

AlexWaygood:
It looks like these methods have been annotated as returning `Any` since the dawn of time: https://github.com/graingert/typeshed/commit/4a1674a148c50c5d02a38611deb5416e732912d7

PR #9295, opened 2022-11-28
Update jsonschema protocols.pyi

spagh-eddie:
These take in an instance, which need not be a JSON object. They could, for example, be arrays. See [this example](https://python-jsonschema.readthedocs.io/en/stable/api/jsonschema/protocols/#jsonschema.protocols.Validator.iter_errors) on the jsonschema docs.

```python
>>> schema = {"maxItems" : 2}
>>> Draft202012Validator(schema).is_valid([2, 3, 4])
False
```


AlexWaygood:
@sirosen, mind taking a look? :)

"spagh-eddie" is a great GitHub username btw.

sirosen:
> "spagh-eddie" is a great GitHub username btw.

Seconded. ❤️ 

---

I agree that a big union would be better than `Any`. I think it would be
```python
str | int | float | bool | None | Mapping[str, Any] | Sequence[Any]
```

I am in favor of this change though. I'll look at applying a near-equivalent change to `jsonschema` to reflect this.

AlexWaygood:
> I agree that a big union would be better than `Any`. I think it would be
> 
> ```python
> str | int | float | bool | None | Mapping[str, Any] | Sequence[Any]
> ```

Since `Mapping` is covariant in its argument type, and `Sequence` is fully covariant, we can actually maybe do even better here, by using a fancy recursive type alias here:

```python

_JsonParameter: TypeAlias = str | int | float | bool | None | Mapping[str, _JsonParameter] | Sequence[_JsonParameter]
```

And then use `_JsonParameter` as the annotation for all three methods.

@spagh-eddie, would you mind making the change?

sirosen:
> recursive type alias

I wasn't aware that we had this capability! Looks like recent `mypy` versions have this problem solved.

I agree; the recursive type would be best.

spagh-eddie:
> recursive type alias

I knew that of last year it was not supported. Glad to see support is starting!

PR #9299, opened 2022-11-29
Deduplicate pyright version

Avasam:
No one likes duplicated configs that have to be kept in sync.
Proposed solution: store it at a single location, then retrieve it when needed. This implementation also requires no parsing.
image


srittau:
+1 on the idea. But maybe we could add the pyright version to `requirements-tests.txt` as a comment?

AlexWaygood:
I agree it's a great idea to clean this up. Here's an alternative implementation to consider, where we store the version in a `tool.typeshed` section in our `pyproject.toml` file: https://github.com/AlexWaygood/typeshed/commit/593029b89b33d1b7f161700c302d66c1cc20897c. We could potentially use this `tool.typeshed` for other things in the future, such as listing the Python versions and Python platforms typeshed supports in our test suite.

Avasam:
I like the pyproject.toml location better. Thanks for the CI tom loader example.

AlexWaygood (approved):

srittau:
I'd prefer to have our dependencies in one place and not split over multiple files.

AlexWaygood:
> I'd prefer to have our dependencies in one place and not split over multiple files.

Fair, I get that. What would the `requirements.txt` solution look like?

Avasam:
> > I'd prefer to have our dependencies in one place and not split over multiple files.
> 
> Fair, I get that. What would the `requirements.txt` solution look like?

That's the thing. requiremetns.txt is for pip requirements. Highjacking it feels like a hack.
And adding something like `# pyright==1.1.278` could be confused with commenting out a https://pypi.org/project/pyright/ install

JelleZijlstra (approved):

srittau:
> > I'd prefer to have our dependencies in one place and not split over multiple files.
> 
> Fair, I get that. What would the `requirements.txt` solution look like?

Put it in a comment in the same format as other dependencies. Alternatively, put all dependencies into pyproject, but I don't think this would work yet. Still, this PR is an improvement over the status quo, so I won't reject it, but I still believe splitting dependencies over multiple files is quite confusing.

AlexWaygood:
> I still believe splitting dependencies over multiple files is quite confusing.

I do agree, but I also agree with @Avasam that putting it in `requirements-tests.txt` might be confusing for different reasons — it's a very different kind of dependency to the others, after all.

I'm going to merge this for now, as I think it's a strict improvement over the status quo, but I'm definitely open to further improvements in the future :) Agreed that putting all dependencies in `pyproject.toml` might be ideal, if it's possible.

PR #9304, opened 2022-11-30
Add stubs for click-default-group

pyhedgehog:
This PR adds stubs for click-default-group.


stubs/click-default-group/METADATA.toml
```diff
@@ -0,0 +1,5 @@
+version = "1.2.*"
+requires = ["click"]
```
AlexWaygood, line 2: Unfortunately "non-types dependencies" are not yet possible in typeshed -- see #5768. We've recently been working towards _making_ them possible, but this work is not yet complete. (It will hopefully be complete soon, but I don't have a timeline for you, I'm afraid.)

The best workaround we currently have is to do something like what `types-pycocotools` currently does for `numpy`:

https://github.com/python/typeshed/blob/5c3f10e1df3d6ad86d9d725aa42d0512d7b86ad9/stubs/pycocotools/pycocotools/cocoeval.pyi#L1-L10

_But_ that seems like quite a poor solution in this situation, given how fundamental `click` is to these stubs. I'm not really sure what to suggest :( an alternative would be to mark this as "deferred" for now, until we've fixed #5768.

AlexWaygood:
Hmm, looks like we might have a stub_uploader false-positive here @hauntsaninja -- the stub-uploader tests are failing because `click-default-group` supposedly doesn't declare `click` as a dependency at runtime. But it does! https://github.com/click-contrib/click-default-group/blob/b671ae5325d186fe5ea7abb584f15852a1e931aa/setup.py#L52

Any ideas what might be going on?

AlexWaygood:
(@pyhedgehog -- we're now able to have stubtest packages depend on third-party packages that have inline stubs! 🥳)

hauntsaninja:
The package doesn't ship wheels, so PyPI doesn't have requires_dist metadata; you'll see a comment in the source code about this.

AlexWaygood:
> The package doesn't ship wheels, so PyPI doesn't have requires_dist metadata; you'll see a comment in the source code about this (also visible in the test failure logs)

Got it, thanks. Unfortunate.

Avasam:
This would help: https://github.com/typeshed-internal/stub_uploader/pull/88

pyhedgehog:
After implementation of #5768, should I rebase my patch or do something else?

AlexWaygood:
> After implementation of #5768, should I rebase my patch or do something else?

Unfortunately this is stalled until either typeshed-internal/stub_uploader#88 is merged, or another solution to that problem is found. I'll try to take a look at typeshed-internal/stub_uploader#88 soon (but I'm not a stub-uploader maintainer, so I won't be able to merge it, unfortunately)

stubs/click-default-group/METADATA.toml
```diff
@@ -0,0 +1,5 @@
...
+
+[tool.stubtest]
+ignore_missing_stub = false
```
Avasam, line 5: This is now the default and is no longer needed

```suggestion
[tool.stubtest]
ignore_missing_stub = false
```

stubs/click-default-group/click_default_group.pyi
```diff
@@ -0,0 +1,41 @@
...
+    def resolve_command(self, ctx: click.Context, args: list[str]) -> tuple[str | None, click.Command | None, list[str]]: ...
+    def format_commands(self, ctx: click.Context, formatter: click.HelpFormatter) -> None: ...
+    def command(self, *args, **kwargs) -> click.Command: ...  # incomplete
+
+class DefaultCommandFormatter:
+    group: click.Group
+    formatter: click.HelpFormatter
+    mark: str
+    def __init__(self, group: click.Group, formatter: click.HelpFormatter, mark: str = ...) -> None: ...
+    def write_dl(self, rows: t.Sequence[tuple[str, str]], col_max: int = ..., col_spacing: int = ...) -> None: ...
```
AlexWaygood, line 25:
```suggestion
    def write_dl(self, rows: t.Sequence[tuple[str, str]], col_max: int = 30, col_spacing: int = -2) -> None: ...
```

stubs/click-default-group/METADATA.toml
```diff
@@ -0,0 +1,3 @@
+version = "1.2.*"
+requires = ["click"]
```
AlexWaygood, line 2:
```suggestion
# requires a version of click with a py.typed
requires = ["click >= 8.0.0"]
```

PR #9306, opened 2022-11-30
Complete `python-crontab`

sobolevn:
Source: https://gitlab.com/doctormo/python-crontab

Source are very messy in some parts.
I've left several `TODO` items for the future, because some parts are quite complex.
I've also intentionally simplified some corner cases: for example some code might call `a = int(b)`, so I annotated `b: int | str`, because this is what docs say for multiple cases.

This was not pleasant 😒


stubs/python-crontab/crontab.pyi
```diff
@@ -1,15 +1,19 @@
 import re
 import subprocess
 from _typeshed import Incomplete, Self
+from builtins import range as _range
 from collections import OrderedDict
-from collections.abc import Callable, Generator, Iterator
+from collections.abc import Callable, Iterable, Iterator
```
AlexWaygood, line 6:
```suggestion
from collections.abc import Callable, Generator, Iterable, Iterator
```

stubs/python-crontab/crontab.pyi
```diff
@@ -48,9 +53,9 @@ class CronTab:
     @property
     def log(self) -> CronLog: ...
     @property
-    def user(self) -> str | None: ...
+    def user(self) -> _User: ...
     @property
-    def user_opt(self) -> dict[str, str]: ...
+    def user_opt(self) -> dict[str, _User]: ...
```
JelleZijlstra, line 58:
```suggestion
    def user_opt(self) -> dict[str, str]: ...
```
It doesn't return `_user` if it's a boolean or None.

stubs/python-crontab/crontab.pyi
```diff
@@ -72,97 +80,104 @@ class CronTab:
...
+    user: _User
     valid: bool
     enabled: bool
     special: bool
-    comment: Incomplete
-    command: Incomplete
-    last_run: Incomplete
-    env: Incomplete
+    comment: str
+    command: str
```
JelleZijlstra, line 107:
```suggestion
    command: str | None
```

stubs/python-crontab/crontab.pyi
```diff
@@ -72,97 +80,104 @@ class CronTab:
...
-    def months(self) -> CronSlice: ...
+    def months(self) -> int | str: ...
     @property
-    def dow(self) -> CronSlice: ...
+    def dow(self) -> int | str: ...
     def __len__(self) -> int: ...
-    def __getitem__(self, key: SupportsIndex) -> CronSlice: ...
-    def __lt__(self, value) -> bool: ...
-    def __gt__(self, value) -> bool: ...
+    def __getitem__(self, key: str) -> int | str: ...
```
JelleZijlstra, line 162:
```suggestion
    def __getitem__(self, key: int | str) -> int | str: ...
```

PR #9308, opened 2022-12-01
More pywin32 stub completion

Avasam:
Completed based on usage of the following libraries in mypy_primer:
- apprise
- comtypes

Libraries that I use which also uses pywin32:
- PyWinCtl
- pywin32 itself (probably the biggest user of its own stubs)

As well as some of the most popular libraries that use both pywin32 and mypy (all over 1k stars on github):
- certbot
- anki
- flexget
- monkey
- twisted
- salt


stubs/pywin32/_win32typing.pyi
```diff
@@ -162,13 +164,16 @@ class PRINTER_DEFAULTS:
...
+    def AddAccessAllowedAce(self, __access: int, __sid: PySID) -> None: ...
+    @overload
+    def AddAccessAllowedAce(self, __revision: int, __access: int, __sid: PySID) -> None: ...
+    def AddAccessAllowedAceEx(self, __revision: int, __aceflags: int, __access: int, __sid: PySID) -> None: ...
     def AddAccessAllowedObjectAce(
         self, AceRevision, AceFlags, AccessMask, ObjectTypeGuid: PyIID, InheritedObjectTypeGuid: PyIID, sid: PySID
     ) -> None: ...
     def AddAccessDeniedAce(self, revision, access, sid: PySID, access1, sid1) -> None: ...
-    def AddAccessDeniedAceEx(self, revision, aceflags, access, sid: PySID) -> None: ...
+    def AddAccessDeniedAceEx(self, __revision: int, __aceflags: int, __access: int, _sid: PySID) -> None: ...
```
AlexWaygood, line 176:
```suggestion
    def AddAccessDeniedAceEx(self, __revision: int, __aceflags: int, __access: int, __sid: PySID) -> None: ...
```

stubs/pywin32/win32/win32evtlog.pyi
```diff
@@ -1,20 +1,31 @@
...
 
-def ReadEventLog(Handle, Flags, Offset, Size=...) -> list[Incomplete]: ...
-def ClearEventLog(handle, eventLogName: str) -> None: ...
+def ReadEventLog(
+    __Handle: _win32typing.PyEVTLOG_HANDLE, __Flags: int, __Offset: int, Size=...
+) -> list[_win32typing.PyEventLogRecord]: ...
+def ClearEventLog(handle: _win32typing.PyEVTLOG_HANDLE, eventLogName: str) -> None: ...
 def BackupEventLog(handle, eventLogName: str) -> None: ...
-def CloseEventLog(handle) -> None: ...
+def CloseEventLog(_handle: _win32typing.PyEVTLOG_HANDLE) -> None: ...
```
AlexWaygood, line 12:
```suggestion
def CloseEventLog(__handle: _win32typing.PyEVTLOG_HANDLE) -> None: ...
```

stubs/pywin32/win32/win32evtlog.pyi
```diff
@@ -1,20 +1,31 @@
...
 def DeregisterEventSource(handle) -> None: ...
 def NotifyChangeEventLog(handle, handle1) -> None: ...
-def GetNumberOfEventLogRecords(handle): ...
+def GetNumberOfEventLogRecords(handle: _win32typing.PyEVTLOG_HANDLE) -> int: ...
 def GetOldestEventLogRecord(): ...
-def OpenEventLog(serverName: str, sourceName: str) -> _win32typing.PyEVTLOG_HANDLE: ...
+def OpenEventLog(_serverName: str | None, _sourceName: str) -> _win32typing.PyEVTLOG_HANDLE: ...
```
AlexWaygood, line 17:
```suggestion
def OpenEventLog(__serverName: str | None, __sourceName: str) -> _win32typing.PyEVTLOG_HANDLE: ...
```

stubs/pywin32/win32/win32security.pyi
```diff
@@ -33,13 +33,15 @@ def ImpersonateLoggedOnUser(handle: int) -> None: ...
...
+def LogonUser(
+    __Username: str, __Domain: str | None, __Password: str, __LogonType: int, __LogonProvider: int
+) -> _win32typing.PyHANDLE: ...
 def LogonUserEx(
     Username: str, Domain: str, Password: str, LogonType, LogonProvider
 ) -> tuple[int, _win32typing.PySID, Incomplete, Incomplete]: ...
-def LookupAccountName(systemName: str, accountName: str) -> tuple[_win32typing.PySID, str, Incomplete]: ...
-def LookupAccountSid(systemName: str, sid: _win32typing.PySID) -> tuple[str, str, Incomplete]: ...
-def GetBinarySid(SID: str) -> _win32typing.PySID: ...
+def LookupAccountName(_systemName: str | None, __accountName: str) -> tuple[_win32typing.PySID, str, int]: ...
```
AlexWaygood, line 42:
```suggestion
def LookupAccountName(__systemName: str | None, __accountName: str) -> tuple[_win32typing.PySID, str, int]: ...
```

stubs/pywin32/win32/win32ts.pyi
```diff
@@ -5,10 +5,10 @@ def WTSCloseServer(Server: int) -> None: ...
...
 def WTSSetUserConfig(ServerName: str, UserName: str, ConfigClass) -> None: ...
 def WTSEnumerateServers(DomainName: str | None = ..., Version: int = ..., Reserved=...) -> tuple[str, ...]: ...
-def WTSEnumerateSessions(Server: int, Version: int = ..., Reserved=...) -> tuple[Incomplete, ...]: ...
-def WTSLogoffSession(Server: int, SessionId, Wait) -> None: ...
-def WTSDisconnectSession(Server: int, SessionId, Wait) -> None: ...
-def WTSQuerySessionInformation(Server: int, SessionId, WTSInfoClass) -> None: ...
+def WTSEnumerateSessions(__Server: int, __Version: int = ..., __Reserved=...) -> tuple[dict[str, str | int], ...]: ...
+def WTSLogoffSession(__Server: int, __SessionId: int, __Wait: bool) -> None: ...
+def WTSDisconnectSession(__Server: int, __SessionId: int, __Wait: bool) -> None: ...
+def WTSQuerySessionInformation(_Server: int, _SessionId: int, _WTSInfoClass: int) -> str: ...
```
AlexWaygood, line 11:
```suggestion
def WTSQuerySessionInformation(__Server: int, __SessionId: int, __WTSInfoClass: int) -> str: ...
```

AlexWaygood (commented):
I've highlighted a couple of places where it looks like you maybe wanted to make parameters positional-only, but you only added a single leading underscore instead of two :)

Avasam:
> I've highlighted a couple of places where it looks like you maybe wanted to make parameters positional-only, but you only added a single leading underscore instead of two :)

*sigh*, I keep doing this and not spotting it. Can't wait for `/` to be supported. Thanks!

Avasam:
I've updated a bit more by running the stub against pywin32's own code. Helped find a few issues with C modules.

AlexWaygood:
Not sure why mypy_primer is being so slow today — I filed https://github.com/hauntsaninja/mypy_primer/issues/62

AlexWaygood:
What's with the changes to `distutils`, `pickle` and `importlib` you made in 49d6722b625611ac32a87095c34784ab822126af? Shouldn't they be a separate PR(s)? 🧐

Avasam:
> What's with the changes to `distutils`, `pickle` and `importlib` you made in [49d6722](https://github.com/python/typeshed/commit/49d6722b625611ac32a87095c34784ab822126af)? Shouldn't they be a separate PR(s)? 🧐

They're all related to fixing type-checking errors in pywin32. But yes, I'll submit them separately.

stubs/pywin32/win32com/client/dynamic.pyi
```diff
@@ -1,14 +1,72 @@
...
+ALL_INVOKE_TYPES: Incomplete
+
+def debug_print(*args) -> None: ...
+def debug_attr_print(*args) -> None: ...
+def MakeMethod(func, inst, cls): ...
+
+PyIDispatchType = _win32typing.PyIDispatch
+PyIUnknownType = _win32typing.PyIUnknown
+
+_GoodDispatchTypes: TypeAlias = tuple[type[str], type[IIDType]]  # noqa: Y047
```
AlexWaygood, line 34: Not sure we need this `noqa` -- Y047 is "A private `TypeAlias` should be used at least once in the file in which it is defined." You use `_GoodDispatchTypes` just below on line 38!

```suggestion
_GoodDispatchTypes: TypeAlias = tuple[type[str], type[IIDType]]
```

stubs/pywin32/win32com/client/dynamic.pyi
```diff
@@ -1,14 +1,72 @@
...
+
+class _DispatchCreateClass(Protocol):
+    def __init__(
+        self,
+        IDispatch: str | PyIDispatchType | _GoodDispatchTypes | PyIUnknownType,
+        olerepr: build.DispatchItem | build.LazyDispatchItem,
+        userName: str | None = ...,
+        UnicodeToString: None = ...,
+        lazydata: Incomplete | None = ...,
+    ): ...
```
AlexWaygood, line 17: Unfortunately, I don't think this works the way you want it to :( `__init__` methods in protocols are an edge case that are kind of weird and somewhat confusing, both at runtime and in how they're treated by static type checkers. The runtime behaviour is a whole other discussion, so let's stick to the static aspects here ;)

Type checkers _don't_ regard `__init__` methods as part of the interface that a user declares when inheriting from `typing.Protocol`. That means that [they won't emit an error](https://mypy-play.net/?mypy=latest&python=3.11&gist=f4d508333db7b95d0c5ade55c32662e1) on code like the following, which you might expect them to flag:

```python
from typing import Protocol

class Foo(Protocol):
    def __init__(self, arg: str) -> None: ...

class Bar:
    def __init__(self, arg1: int, arg2: bytes) -> None: ...

def expects_foo(x: Foo) -> None:
    pass

expects_foo(Bar(42, b"bar"))
```

The reasons why they won't emit an error on code like that has to do with the Liskov Substitution Principle, which is generally considered only to apply to _objects that have already been constructed_ -- the _constructor methods_ of objects are considered outside of the purview of the LSP. Some further reading on this, if you're interested:
- https://stackoverflow.com/questions/69394820/mypy-accepts-an-incompatible-type-in-init-override
- https://softwareengineering.stackexchange.com/questions/431549/the-liskov-substitution-principle-and-python

AlexWaygood:
> They're all related to fixing type-checking errors in pywin32. But yes, I'll submit them separately.

Ah, I wasn't sure! But yeah, better to do them in a follow-up PR -- thanks!

stubs/pywin32/win32com/client/dynamic.pyi
AlexWaygood: Yeah, we already have it as a plugin, but we have that specific check disabled:

https://github.com/python/typeshed/blob/503eab31d7e11fbe7f458a3e96ba9fa1907ce3a3/.flake8#L25-L28

Basically, flake8-pyi's CI is very tightly coupled with `noqa` comments in typeshed. That often requires introducing `noqa` comments into typeshed for flake8-pyi checks that don't exist yet, in order to get flake8-pyi's CI to pass.

It's all slightly unfortunate, but I have yet to think of a better way of doing things :)

AlexWaygood: The problematic check in flake8-pyi's CI is this: https://github.com/PyCQA/flake8-pyi/blob/e8243d7cc648433cbbf3fd0621a13a50db063e80/.github/workflows/check.yml#L100-L117

What would be awesome would be if we could have a non-blocking mypy_primer-esque check at flake8-pyi, instead of a blocking check. Then flake8-pyi's CI wouldn't need to be so tightly coupled with typeshed's `noqa` comments, and we could enable `NQA102` in typeshed's CI. But that sounds like... a lot of work :(

AlexWaygood: Example: #7658 was required to make the CI on https://github.com/PyCQA/flake8-pyi/pull/213 go green. But the CI on #7658 would have failed if we'd had `NQA102` enabled.

Avasam: I see. This makes perfect sense. I think I have an idea, I'll play around. See if I can make an action similar to typeshed's mypy_primer_comment.

stubs/pywin32/win32com/client/dynamic.pyi
Avasam: image
image
image


But I also tried reimplementing it using Callback Protocols and it seems to be working just as well:
```py
_T = TypeVar("_T")

class _DispatchCreateClass(Protocol[_T_co]):
    @staticmethod
    def __call__(
        IDispatch: str | PyIDispatchType | _GoodDispatchTypes | PyIUnknownType,
        olerepr: build.DispatchItem | build.LazyDispatchItem,
        userName: str | None = ...,
        UnicodeToString: None = ...,
        lazydata: Incomplete | None = ...,
    ) -> _T_co: ...

@overload
def Dispatch(
    IDispatch: str | PyIDispatchType | _GoodDispatchTypes | PyIUnknownType,
    userName: str | None,
    createClass: _DispatchCreateClass[_T],
    typeinfo: _win32typing.PyITypeInfo | None = ...,
    UnicodeToString: None = ...,
    clsctx: int = ...,
) -> _T: ...
```

Sidenote, the parameter `createClass` of the first overload should not be ` | None` either way. Fixing that as well.

Avasam: If I change the *type* of a parameter, mypy won't catch it with `__init__`. So Callback protocols it is.

Fun fact, pyright did catch it however:
image

stubs/pywin32/win32com/client/dynamic.pyi
```diff
@@ -1,14 +1,73 @@
...
+
+class _DispatchCreateClass(Protocol[_T_co]):
+    @staticmethod
+    def __call__(
+        IDispatch: str | PyIDispatchType | _GoodDispatchTypes | PyIUnknownType,
+        olerepr: build.DispatchItem | build.LazyDispatchItem,
+        userName: str | None = ...,
+        UnicodeToString: None = ...,
+        lazydata: Incomplete | None = ...,
+    ) -> _T_co: ...
```
AlexWaygood, line 20: I'm still a _little_ uncertain that this is doing what you want it to, but I don't want to hold up this PR any longer if you're confident it works for your purposes :)

(And we can always fix it later if there is a problem.)

AlexWaygood (approved):

stubs/pywin32/win32com/client/dynamic.pyi
Avasam: Thanks! It definitely *seems* to do what I what it to. At the very least it acts the way I expect.

PR #9309, opened 2022-12-01
Mark `humanfriendly` as complete

sobolevn:
Source: https://github.com/xolox/python-humanfriendly/blob/master/humanfriendly/compat.py


AlexWaygood (approved):
These are all only flagged by stubtest because they're explicitly included in `__all__` by the library at runtime. But I think we can safely leave them out of the stub unless and until somebody complains :)

PR #9310, opened 2022-12-01
Remove `dockerfile-parse` from pyright's exclude

sobolevn:
I don't see why it should be there. Stubs are complete.


sobolevn:
Source: https://github.com/containerbuildsystem/dockerfile-parse/blob/1d4f9fa6a88e45434d37eafe99193c95a5a6b196/dockerfile_parse/util.py#L216-L261

AlexWaygood:
Could you possibly also remove this unnecessary dependency as part of this PR?

https://github.com/python/typeshed/blob/be4bfd660e8661b3d668a417ea622aaeab6305ae/stubs/dockerfile-parse/METADATA.toml#L2

I realised immediately after merging the PR that the final version of the PR no longer imported anything from `six`

AlexWaygood (approved):
Thank you!

PR #9312, opened 2022-12-01
Mark `dateparser` as complete

sobolevn:
Source: https://github.com/scrapinghub/dateparser

### DIFF
--- a/stubs/dateparser/@tests/stubtest_allowlist.txt
+++ b/stubs/dateparser/@tests/stubtest_allowlist.txt
@@ -12,5 +12,7 @@ dateparser.conf.date_order_chart
 dateparser.conf.language_order
 dateparser.languages.loader.language_locale_dict
 dateparser.languages.loader.language_order
+dateparser.languages.locale.ALWAYS_KEEP_TOKENS
 dateparser.custom_language_detection.language_mapping.language_map
 dateparser.custom_language_detection.fasttext.dateparser_model_home
+dateparser.timezone_parser.timezone_info_list


stubs/dateparser/@tests/requirements-stubtest.txt
```diff
@@ -0,0 +1,2 @@
+fasttext
```
AlexWaygood, line 1: It would be better to just specify the `extras` in `METADATA.toml`, similar to https://github.com/python/typeshed/blob/3c9d1bd942c9ae754156601ec67f160a192f271f/stubs/requests/METADATA.toml#L6

stubs/dateparser/dateparser/languages/locale.pyi
```diff
@@ -3,6 +3,7 @@ from re import Pattern
 from dateparser.conf import Settings
 
 NUMERAL_PATTERN: Pattern[str]
+ALWAYS_KEEP_TOKENS: list[str]
```
AlexWaygood, line 6: This is just a re-export of https://github.com/python/typeshed/blob/c6126b5f52c9e1ad294f7435d1a2242c2f14103a/stubs/dateparser/dateparser/languages/dictionary.pyi#L5

I think we can allowlist it

```suggestion
```

stubs/dateparser/dateparser/timezone_parser.pyi
```diff
@@ -16,3 +16,4 @@ def build_tz_offsets(search_regex_parts): ...
 def get_local_tz_offset(): ...
 
 local_tz_offset: Any
+timezone_info_list: list[dict[str, list[str | tuple[str, ...]]]]
```
AlexWaygood, line 19:
```suggestion
```

A re-export of https://github.com/python/typeshed/blob/c6126b5f52c9e1ad294f7435d1a2242c2f14103a/stubs/dateparser/dateparser/timezones.pyi#L3 (Though it would be great if you could put the new, improved type there!)

AlexWaygood (commented):
Mostly looks great, just two things!

AlexWaygood (approved):
Thanks!

PR #9313, opened 2022-12-01
Add stubs for show-in-file-manager

pyhedgehog:
This PR adds stubs for `show-in-file-manager`.
After #5768 should be patched to use `packaging.version.Version`.


AlexWaygood:
It looks like [the upstream library](https://github.com/damonlynch/showinfilemanager) is already fully annotated for this one. Have you asked them whether they'd be prepared to add a `py.typed` marker, so that users are able to make use of those annotations? If they'd consider it, that would probably be a better option than adding stubs to typeshed.

AlexWaygood:
Closing, since the upstream package has added a `py.typed` file (they now just need to cut a new release)

PR #9315, opened 2022-12-01
Use `__name__ == "__main__"` in `get_packages.py`

sobolevn:
I know that this package is not suitable for imports, but this was really catching my eye 👀

### DIFF
--- a/tests/get_packages.py
+++ b/tests/get_packages.py
@@ -5,6 +5,7 @@ import sys
 import tomli
 from utils import METADATA_MAPPING
 
+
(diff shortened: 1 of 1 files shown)


AlexWaygood:
Not sure I see the point in this; the sole purpose of this file is for it to be a utility script for use in CI 🤷‍♂️

sobolevn:
Yeah - no practical reason, it just grinds my gears.

AlexWaygood:
Sorry, this one just feels like unnecessary churn to me :/

PR #9316, opened 2022-12-01
Clarify how `*_exclude_list.txt` files work

sobolevn:
Pyright does not need this file at all. It is there just as a doc, I guess? There's no need for it.

Pytype has its path hardcoded into its codebase:

```python
def load_pytype_blocklist(self) -> List[str]:
    """List of modules that we maintain our own versions of."""
    return self._readlines("tests/pytype_exclude_list.txt")
```

### DIFF
--- a/tests/pyright_exclude_list.txt
+++ b/tests/pyright_exclude_list.txt
@@ -1,3 +1,0 @@
-Pyright is configured using a "pyrightconfig.json" file. You will find
-it at the root of the repo. It contains an "exclude" section that accepts
-globs for specifying which files and/or directories to exclude from analysis.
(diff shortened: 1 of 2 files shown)


AlexWaygood:
I don't see any reason to delete the pyright excludelist file. It seems like useful information to me. Our pyright tests work differently to all our other tests, and that might be confusing to people new to typeshed.

sobolevn:
We have the test process documented here: https://github.com/python/typeshed/tree/main/tests#pyright_testpy

And mimicing `pytype`'s test setup does not seem like a good idea for a completely different tool.

srittau:
Indeed. The idea behind this file was that people looking for the pyright excludes in the usual spot find it documented here. -1 on deleting it.

sobolevn:
> in the usual spot 

I might be missing something, but are any other tools using this convention? Or just `pytype`?

AlexWaygood:
Stubtest also has all of its configuration inside the `tests` directory. `mypy_test.py` and `regr_test.py` are both basically self-contained when it comes to configuration. It's only really pyright that looks outside of the `tests` directory for configuration settings.

PR #9317, opened 2022-12-01
Remove unused code from `pytype`

sobolevn:
It is not used since https://github.com/python/typeshed/pull/4975


AlexWaygood (approved):
Nice

PR #9323, opened 2022-12-02
`protobuf`: Annotate `well_known_types.pyi`

ikonst:
`google.protobuf.struct_pb2.Struct` and `.List` offer high-level methods which we are partially blind to, making things like `list(list_value)` fail.

The methods return either JSON primitive types (numbers are always floats!), `Struct` or `List`, and accept that along with plain sequences and mappings of the above. https://github.com/python/typing/issues/182 comes handy in that we can now rely on recursive type aliases.


stubs/protobuf/@tests/struct.py
```diff
@@ -0,0 +1,15 @@
+from google.protobuf.struct_pb2 import ListValue, Struct
```
AlexWaygood, line 1: This file is in the wrong place — it needs to be moved to `stubs/protobuf/@tests/test_cases/check_struct.py`, or it won't be tested with mypy, only with pyright

AlexWaygood:
You may need to invent a fictional `__iter__` method to get the type checkers to accept that `ListValue` is iterable 🙂 We do that in a few other places.

Type checkers aren't very good at understanding that classes with `__getitem__` are also iterable.

ikonst:
Thanks, I was just wondering why it "passes" in mypy...

AlexWaygood:
You'll need to add an allowlist to get stubtest to shut up about the fictional `__iter__` method — see https://github.com/python/typeshed/blob/c6b9b4ce85d4432639eb146e6f03806b08128a5a/stubs/Flask-SQLAlchemy/%40tests/stubtest_allowlist.txt#L2 for an example

AlexWaygood:
You'll need to add an entry to https://github.com/python/typeshed/blob/main/stubs/protobuf/%40tests/stubtest_allowlist.txt to get stubtest to shut up about the fictional `__iter__` method.

AlexWaygood:
No idea what's going on with pytype, that might need some investigation.

AlexWaygood:
Ahh, I'm guessing this breaks pytype because you're importing from one of the files that we exclude from the pytype check: https://github.com/python/typeshed/blob/main/tests/pytype_exclude_list.txt

I guess if need be, we could maybe just add more files to that excludelist. Hopefully that will fix it.

stubs/protobuf/google/protobuf/internal/well_known_types.pyi
```diff
@@ -63,29 +67,33 @@ class _FieldMaskTree:
...
+_StructValue: TypeAlias = struct_pb2.Struct | struct_pb2.ListValue | str | float | bool | None
+_StructValueArg: TypeAlias = _StructValue | Mapping[str, _StructValueArg] | Sequence[_StructValueArg]
+
 class Struct:
-    def __getitem__(self, key: tAny): ...
-    def __contains__(self, item: tAny): ...
-    def __setitem__(self, key: tAny, value: tAny) -> None: ...
-    def __delitem__(self, key: tAny) -> None: ...
+    def __getitem__(self, key: str) -> _StructValue: ...
+    def __contains__(self, item: _StructValueArg) -> bool: ...
```
AlexWaygood, line 75: Assuming `self.fields` is some kind of mapping, this should work with any object:

```suggestion
    def __contains__(self, item: object) -> bool: ...
```

Using `object` as the annotation here is in line with what we do with other mapping classes in typeshed: https://github.com/python/typeshed/blob/1c940e89141e0c09eeebbce2679760ec03173beb/stdlib/typing.pyi#L586

stubs/protobuf/google/protobuf/internal/well_known_types.pyi
```diff
@@ -63,29 +67,33 @@ class _FieldMaskTree:
...
     def __len__(self) -> int: ...
-    def __iter__(self): ...
-    def keys(self): ...
-    def values(self): ...
-    def items(self): ...
-    def get_or_create_list(self, key: tAny): ...
-    def get_or_create_struct(self, key: tAny): ...
-    def update(self, dictionary: tAny) -> None: ...
+    def __iter__(self) -> Iterator[str]: ...
+    def keys(self) -> Iterator[str]: ...
```
AlexWaygood, line 80: This looks like it probably returns a `collections.abc.KeysView`: https://github.com/protocolbuffers/protobuf/blob/ca561b0ef5ea5a9ca908b002e574e20e8b8fb1b8/python/google/protobuf/internal/well_known_types.py#L498

stubs/protobuf/google/protobuf/internal/well_known_types.pyi
```diff
@@ -63,29 +67,33 @@ class _FieldMaskTree:
...
+    def values(self) -> Iterator[_StructValue]: ...
+    def items(self) -> Iterator[tuple[str, _StructValue]]: ...
```
AlexWaygood, line 82: These both return lists: https://github.com/protocolbuffers/protobuf/blob/ca561b0ef5ea5a9ca908b002e574e20e8b8fb1b8/python/google/protobuf/internal/well_known_types.py#L500-L504

```suggestion
    def values(self) -> list[_StructValue]: ...
    def items(self) -> list[tuple[str, _StructValue]]: ...
```

stubs/protobuf/google/protobuf/internal/well_known_types.pyi
```diff
@@ -63,29 +67,33 @@ class _FieldMaskTree:
...
+    def get_or_create_list(self, key: str) -> struct_pb2.ListValue: ...
+    def get_or_create_struct(self, key: str) -> struct_pb2.Struct: ...
+    def update(self, dictionary: Mapping[str, _StructValueArg]) -> None: ...
```
AlexWaygood, line 85: Could potentially use our `SupportsItems` protocol for this one:

https://github.com/python/typeshed/blob/1c940e89141e0c09eeebbce2679760ec03173beb/stdlib/_typeshed/__init__.pyi#L112-L113

stubs/protobuf/google/protobuf/internal/well_known_types.pyi
```diff
@@ -63,29 +67,33 @@ class _FieldMaskTree:
...
-    def __delitem__(self, key: tAny) -> None: ...
-    def items(self) -> None: ...
-    def add_struct(self): ...
-    def add_list(self): ...
+    def append(self, value: _StructValue) -> None: ...
+    def extend(self, elem_seq: Iterable[_StructValue]) -> None: ...
+    def __getitem__(self, index: int) -> _StructValue: ...
+    def __setitem__(self, index: int, value: _StructValueArg) -> None: ...
+    def __delitem__(self, key: int) -> None: ...
+    def __iter__(self) -> Iterator[_StructValue]: ...
```
AlexWaygood, line 94: I know you added a comment in the allowlist, but I think it wouldn't hurt to add one here as well; it's a little surprising :)

```suggestion
    # Doesn't actually exist at runtime; needed so type checkers understand the class is iterable
    def __iter__(self) -> Iterator[_StructValue]: ...
```

AlexWaygood (commented):
Thanks! A few comments:

stubs/protobuf/google/protobuf/internal/well_known_types.pyi
ikonst: Are we expected to use those protocols for type annotation, or mostly for structural subtyping?

stubs/protobuf/google/protobuf/internal/well_known_types.pyi
ikonst: 83ffc90cbc94fd5b7e95b5a2a740aca2d17a73c2

stubs/protobuf/google/protobuf/internal/well_known_types.pyi
ikonst: Also, not sure how an attempt to derive from that protocol would work with the stubtest.

AlexWaygood: Not suggesting you should derive from the protocol. Just suggesting that the argument type doesn't necessarily need to be a `Mapping`, it only needs to have a `.items()` method, so you can just do:

```suggestion
    def update(self, dictionary: SupportsItems[str, _StructValueArg]) -> None: ...
```

stubs/protobuf/google/protobuf/internal/well_known_types.pyi
AlexWaygood: Hmm, well `Iterator` is incorrect either way, since a `list` is not an `Iterator` (lists don't have a `__next__` method). I won't object if you want to change it to `Iterable` instead of `list` 🙂

stubs/protobuf/google/protobuf/internal/well_known_types.pyi
ikonst: Ahh!

AlexWaygood (approved):
LGTM

PR #9324, opened 2022-12-02
openpyxl stub: add open as an alias for load_workbook

bpjatfb:
`open()` was added as alias for `load_workbook()` in April 2019: https://foss.heptapod.net/openpyxl/openpyxl/-/commit/d405d28aa2ae373e5484387487c30488038cab2a

I am making this commit to resolve type errors on code like the following:

```
openpyxl.open("test.xlsx")
         ^^^^
Undefined attribute [16]: Module openpyxl has no attribute open.
This module is shadowed by a stub file at openpyxl/__init__.pyi. Ensure open is defined in the stub file.
```

Please let me know if I can provide any more info or need to modify the PR!


bpjatfb:
Bad bot.

But seriously, the bot got rid of my change, and I'm not sure how to address it?  The CI says it's an unused import, but the whole point of the import is to export the alias..

bpjatfb:
I guess this section in CONTRIBUTING.md is relevant:

>Imports in stubs are considered private (not part of the exported API) unless:
>
> - they use the form from library import name as name (sic, using explicit as even if the name stays the same); or
> - they use the form from library import * which means all names from that library are exported.

Still working on figuring out how to make an alias public.

JelleZijlstra:
`open = load_workbook` I think?

bpjatfb:
Many thanks Jelle! It looks like that worked 😁

AlexWaygood:
Thanks!

PR #9326, opened 2022-12-04
Improve typing for ast nodes which represent assignment behaviour

### LINKED ISSUE #9287: Annotation for target attribute in ast.Assign and similar nodes is too broad
Currently, the `targets` attribute for `ast.Assign`, and other nodes where assignment happens is likely too broad - it is `ast.expr`. 
I have a suspicion that, this is likely too generic and should be narrowed down, to correct set of types that can truly serve the role of an assignment target. 

From a quick look at the [official docs for the assignment statement](https://docs.python.org/3/reference/simple_stmts.html#grammar-token-python-grammar-assignment_stmt), looks like the correct type should roughly be the union of the following types: ast.Name, ast.Attribute, ast.Subscript.

If I get a confirmation that indeed this could be improved, then I'd love to work on this change. In general in the PR I would take a look at all nodes that cause assignment of value to a target (Assign, AnnAsign, AugAssign, withitem, NamedExpr) and introduce a new type called `ast.assign_target`.

Domlenart:
Closes #9287.

I also decided to type ast.Assign. The new hint for it is still less broad then `expr`

### DIFF
--- a/stdlib/_ast.pyi
+++ b/stdlib/_ast.pyi
@@ -98,20 +98,20 @@ class Delete(stmt):
 class Assign(stmt):
     if sys.version_info >= (3, 10):
         __match_args__ = ("targets", "value", "type_comment")
-    targets: list[expr]
+    targets: list[Tuple | Name | Attribute | Subscript | Starred]
     value: expr
 
 class AugAssign(stmt):
     if sys.version_info >= (3, 10):
         __match_args__ = ("target", "op", "value")
-    target: expr
+    target: Name | Attribute | Subscript
     op: operator
     value: expr
 
 class AnnAssign(stmt):
     if sys.version_info >= (3, 10):
         __match_args__ = ("target", "annotation", "value", "simple")
-    target: expr
+    target: Name | Attribute | Subscript
     annotation: expr
     value: expr | None
     simple: int
@@ -355,7 +355,7 @@ if sys.version_info >= (3, 8):
     class NamedExpr(expr):
         if sys.version_info >= (3, 10):
             __match_args__ = ("target", "value")
-        target: expr
+        target: Name
         value: expr
 
 class Attribute(expr):
@@ -515,7 +515,7 @@ class alias(AST):
 class withitem(AST):
     if sys.version_info >= (3, 10):
         __match_args__ = ("context_expr", "optional_vars")
-    context_expr: expr
+    context_expr: Name | Subscript | Attribute
     optional_vars: expr | None
 
 if sys.version_info >= (3, 10):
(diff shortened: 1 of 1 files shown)


stdlib/_ast.pyi
```diff
@@ -98,20 +98,20 @@ class Delete(stmt):
 class Assign(stmt):
     if sys.version_info >= (3, 10):
         __match_args__ = ("targets", "value", "type_comment")
-    targets: list[expr]
+    targets: list[Tuple | Name | Attribute | Subscript | Starred]
```
JelleZijlstra, line 101: Missing List

JelleZijlstra:
Personally I lean towards not changing this at all: I think it's better to reflect the declared ASDL types directly instead of narrowing things down manually.

stdlib/_ast.pyi
AlexWaygood: I'd definitely prefer it if we left this one as it is. It feels problematic due to `list` invariance, as we can see from the `mypy_primer` output. The current annotation is more ergonomic, even if it's slightly less precise. The situation will get worse if we add better `__init__` methods for `ast` nodes, as proposed in #8378

We could maybe add a comment:

```suggestion
    # Strictly speaking a target can only be a Tuple, Name, Attribute, List, Subscript, or Starred,
    # But due to `list` invariance, that's a pretty annoying annotation to use here
    targets: list[expr]
```

AlexWaygood:
> Personally I lean towards not changing this at all: I think it's better to reflect the declared ASDL types directly instead of narrowing things down manually.

I feel like I'm much more in favour of the changes to `ast.NamedExpr` and `ast.AnnAssign` than the other changes. Those two nodes are highly restricted in the syntax that's allowed at runtime. But for the other nodes, you have to use very big unions, which are error-prone, and don't get us much more precise typing than what we have currently.

stdlib/_ast.pyi
Domlenart: OK I will revert this change. I agree with @AlexWaygood, the current state is better from ergonomic standpoint. 

Btw, as I am working a lot with `ast` module lately (writing something similar to https://www.archunit.org/ but for Python), I can take on the #8378 after this one is done and in general I'd like to contribute to improvement of typing for `ast` module.

stdlib/_ast.pyi
```diff
@@ -98,20 +98,20 @@ class Delete(stmt):
 class Assign(stmt):
     if sys.version_info >= (3, 10):
         __match_args__ = ("targets", "value", "type_comment")
-    targets: list[expr]
+    targets: expr
```
AlexWaygood, line 101:
```suggestion
    targets: list[expr]
```

AlexWaygood (approved):
I'm happy with this to go in, now that it's restricted to only `AugAssign`, `AnnAssign` and `NamedExpr`. I think it's a useful change.

PR #9327, opened 2022-12-04
fpdf:annotation of underline, font_style, font_streching

AmberAnsari89:


AlexWaygood:
Thanks @AmberAnsari89. No need to change anything for this PR, but for the future, you'd be wise to create a new `git` branch for each PR you file. You can see that the history for your previous `fpdf` PR is showing up in this PR, which is undesirable.

(Again, no need to change anything for this PR.)

AmberAnsari89:
Thank you for your help. I assumed that it is better to  keep all changes in a single PR for a file(graphics_state.pyi). Sure, here onward i would create new PR for every new submission.  Please close this PR.

AlexWaygood:
> Thank you for your help. I assumed that it is better to keep all changes in a single PR for a file(graphics_state.pyi).

This isn't about the size of your PR, it is about the `git` history that shows up in your PR :)

> Please close this PR.

As I said, it doesn't matter for this PR, there's no problem. But it _might_ cause issues in future PRs; this is just advice on general best practice :)

stubs/fpdf2/fpdf/graphics_state.pyi
```diff
@@ -20,17 +20,17 @@ class GraphicsStateMixin:
...
     def text_color(self, v: DeviceGray | DeviceRGB) -> None: ...
     @property
-    def underline(self): ...
+    def underline(self) -> bool: ...
     @underline.setter
-    def underline(self, v) -> None: ...
+    def underline(self, v: bool) -> None: ...
     @property
-    def font_style(self): ...
+    def font_style(self) -> str | None: ...
```
AlexWaygood, line 27: I don't think this can ever be `None` -- why do you think it might be? :)

```suggestion
    def font_style(self) -> str: ...
```

stubs/fpdf2/fpdf/graphics_state.pyi
```diff
@@ -20,17 +20,17 @@ class GraphicsStateMixin:
...
     @font_style.setter
-    def font_style(self, v) -> None: ...
+    def font_style(self, v: str | None) -> None: ...
```
AlexWaygood, line 29:
```suggestion
    def font_style(self, v: str) -> None: ...
```

AmberAnsari89:
understood.. i will take care for next PR.

AlexWaygood (commented):
Thanks! `underline` and `font_stretching` look good, but I'm not sure about `font_style`

stubs/fpdf2/fpdf/graphics_state.pyi
AmberAnsari89: Font_style  : Even i was not sure but i saw font_style: optional[str] =None  , and optional is either str or None and the default value of font_style is "". Therefore i thought ,None can be a return type or argument.

AlexWaygood: True, but that's for a different class, and it looks to me like there are quite a few tests in the test suite that assume that `FPDF.font_style` will always be a `str`. So I think I'd go with `str` :)

AmberAnsari89: > Font_style

Are you taking below test suit?
https://github.com/PyFPDF/fpdf2/blob/master/test/fonts/test_set_font.py#L121-L144

AlexWaygood (approved):

AmberAnsari89:
i have a question... what is the type hint for empty string ("") and font_style can have default .

AlexWaygood:
> what is the type hint for empty string ("")

Just the same as any other string! `str`

PR #9328, opened 2022-12-04
PIL: add optimize param to save

sbdchd:
This option is covered in https://github.com/python-pillow/Pillow/blob/6a2545f628b13bc4888f687c1028482fc93fafcf/docs/handbook/tutorial.rst#batch-processing

but it is pulled out of the `**params` by the `jpeg` plugin (and probably others as well)

example usage:

```python
im.save(path, optimize=True)
```

rel: https://pillow.readthedocs.io/en/stable/handbook/image-file-formats.html#jpeg-saving


AlexWaygood:
Thanks!

PR #9332, opened 2022-12-05
Don't run `mypy_primer` on stubsabot PRs

AlexWaygood:
We can be fairly confident that PRs updating only a `METADATA.toml` file will never produce an interesting `mypy_primer` diff.


hauntsaninja (approved):
Looks good, maybe if you added python2 = true or something...

PR #9333, opened 2022-12-05
types: fix signature of coroutine

gaohuazuo:
According to Python [doc](https://docs.python.org/3/library/typing.html#typing.Generator), type parameters for `Generator` should be `Generator[YieldType, SendType, ReturnType]`

An example of wrong diagnostics emitted by a downstream type checker: https://github.com/microsoft/pylance-release/issues/3716


JelleZijlstra:
Thanks, this looks right, though I'd advise you to stop using `@types.coroutine` :)

PR #9336, opened 2022-12-06
`mypy_primer`: Run on Python 3.11

AlexWaygood:
`mypy_primer` currently runs on Python 3.10 in CI, meaning we don't have any `mypy_primer` coverage for PRs fiddling with `ExceptionGroup`s or other new features added in Python 3.11


JelleZijlstra:
Seems like the problem is they pin `scipy==1.7.3`, and that doesn't support 3.11.

AlexWaygood:
Looks like everything's green now!

PR #9338, opened 2022-12-06
Add types for jmespath internals

mondeja:
Adds a bunch of types to jmespath. Some types like `Functions` and `Options` are part of the public API and some others are useful when parsing expressions.


stubs/jmespath/@tests/stubtest_allowlist.txt
```diff
@@ -0,0 +1,3 @@
+jmespath.lexer.LexerTokenizeResult
+jmespath.lexer.LexerTokenizeResults
+jmespath.visitor.TreeNode
```
AlexWaygood, line 3: We shouldn't add allowlist entries without adding a comment saying why we're ignoring the test failures. Why was stubtest failing for these three objects?

mondeja: Thanks for the review. I've added these objects here because them are useful when writing code that uses the library, so I want to be able to import them with `from jmespath.visitor import TreeNode`. What would be the recommended way to deal with them, must I keep them private?

AlexWaygood: Hmm, yes, I'd rather we gave them names with leading underscores in that case. Stubtest is pointing to a real issue here: since these names don't exist at runtime, attempting to import them at runtime will fail unless you're in an `if TYPE_CHECKING:` block. Using names with leading underscores will help warn end users of the danger.

mondeja: Makes sense, thanks for the review!

stubs/jmespath/jmespath/visitor.pyi
```diff
@@ -1,49 +1,57 @@
 from typing import Any
+from typing_extensions import TypedDict
+
+from jmespath.functions import Functions
 
 class Options:
     dict_cls: Any
```
AlexWaygood, line 7: Cf. this comment in the source code here: https://github.com/jmespath/jmespath.py/blob/bbe7300c60056f52413603cf3e2bcd0b6afeda3d/jmespath/visitor.py#L62-L69

It does have to be a _mutable_ mapping due to the `TreeInterpreter.visit_multi_select_dict()` method (`TreeInterpreter._dict_cls` is set to whatever `options.dict_cls` is in `TreeInterpreter.__init__`): https://github.com/jmespath/jmespath.py/blob/bbe7300c60056f52413603cf3e2bcd0b6afeda3d/jmespath/visitor.py#L233-L236

(You'll need to import `MutableMapping` from `collections.abc`)

```suggestion
    dict_cls: Callable[[], MutableMapping[Any, Any]]
```

stubs/jmespath/jmespath/visitor.pyi
```diff
@@ -1,49 +1,57 @@
...
-    custom_functions: Any
-    def __init__(self, dict_cls: Any | None = ..., custom_functions: Any | None = ...) -> None: ...
+    custom_functions: Functions
+    def __init__(self, dict_cls: Any | None = ..., custom_functions: Functions | None = ...) -> None: ...
```
AlexWaygood, line 9:
```suggestion
    def __init__(self, dict_cls: Callable[[], MutableMapping[Any, Any]] | None = ..., custom_functions: Functions | None = ...) -> None: ...
```

stubs/jmespath/jmespath/visitor.pyi
```diff
@@ -1,49 +1,57 @@
...
+    interpreter: Visitor
+    def __init__(self, expression: str, interpreter: Visitor) -> None: ...
+    def visit(self, node: _TreeNode, *args, **kwargs) -> _TreeNode: ...
 
 class Visitor:
     def __init__(self) -> None: ...
-    def visit(self, node, *args, **kwargs): ...
-    def default_visit(self, node, *args, **kwargs) -> None: ...
+    def visit(self, node: _TreeNode, *args, **kwargs) -> Any: ...
+    def default_visit(self, node: _TreeNode, *args, **kwargs) -> None: ...
```
AlexWaygood, line 20: (You'll need to import `NoReturn` from `typing`)

```suggestion
    def default_visit(self, node: _TreeNode, *args, **kwargs) -> NoReturn: ...
```

stubs/jmespath/jmespath/visitor.pyi
```diff
@@ -1,49 +1,57 @@
...
+
+class _TreeNode(TypedDict):
+    type: str
+    value: Any
+    children: list[_TreeNode]
 
 class TreeInterpreter(Visitor):
     COMPARATOR_FUNC: Any
```
AlexWaygood, line 28:
```suggestion
    COMPARATOR_FUNC: ClassVar[dict[str, Callable[[Any, Any], Any]]]
```

stubs/jmespath/jmespath/visitor.pyi
```diff
@@ -1,49 +1,57 @@
...
     MAP_TYPE: Any
```
AlexWaygood, line 29:
```suggestion
    MAP_TYPE: ClassVar[Callable[[], MutableMapping[Any, Any]]
```

stubs/jmespath/jmespath/visitor.pyi
```diff
@@ -1,49 +1,57 @@
...
     def __init__(self, options: Any | None = ...) -> None: ...
```
AlexWaygood, line 30:
```suggestion
    def __init__(self, options: Options | None = ...) -> None: ...
```

stubs/jmespath/jmespath/visitor.pyi
```diff
@@ -1,49 +1,57 @@
...
-    def visit_literal(self, node, value): ...
-    def visit_multi_select_dict(self, node, value): ...
-    def visit_multi_select_list(self, node, value): ...
-    def visit_or_expression(self, node, value): ...
-    def visit_and_expression(self, node, value): ...
-    def visit_not_expression(self, node, value): ...
-    def visit_pipe(self, node, value): ...
-    def visit_projection(self, node, value): ...
-    def visit_value_projection(self, node, value): ...
+    def default_visit(self, node: _TreeNode, *args, **kwargs) -> None: ...
```
AlexWaygood, line 31:
```suggestion
    def default_visit(self, node: _TreeNode, *args, **kwargs) -> NoReturn: ...
```

stubs/jmespath/jmespath/parser.pyi
```diff
@@ -1,15 +1,18 @@
 from typing import Any
 
+from jmespath.lexer import _LexerTokenizeResults
+from jmespath.visitor import _TreeNode
+
 class Parser:
-    BINDING_POWER: Any
-    tokenizer: Any
+    BINDING_POWER: dict[str, int]
```
AlexWaygood, line 7:
```suggestion
    BINDING_POWER: ClassVar[dict[str, int]]
```

stubs/jmespath/jmespath/parser.pyi
```diff
@@ -1,15 +1,18 @@
...
+    tokenizer: _LexerTokenizeResults
```
AlexWaygood, line 8: Looks like this can sometimes be `None`: https://github.com/jmespath/jmespath.py/blob/bbe7300c60056f52413603cf3e2bcd0b6afeda3d/jmespath/parser.py#L79

```suggestion
    tokenizer: _LexerTokenizeResults | None
```

stubs/jmespath/jmespath/parser.pyi
```diff
@@ -1,15 +1,18 @@
...
 
 class ParsedResult:
-    expression: Any
-    parsed: Any
-    def __init__(self, expression, parsed) -> None: ...
-    def search(self, value, options: Any | None = ...): ...
+    expression: str
+    parsed: _TreeNode
+    def __init__(self, expression: str, parsed: _TreeNode) -> None: ...
+    def search(self, value: Any, options: Any | None = ...) -> _TreeNode: ...
```
AlexWaygood, line 18:
```suggestion
    def search(self, value: Any, options: Options | None = ...) -> _TreeNode: ...
```

AlexWaygood (commented):
Thanks! This all looks correct, just some more suggestions here to make it even better:

stubs/jmespath/jmespath/visitor.pyi
```diff
@@ -1,49 +1,60 @@
...
+from typing_extensions import TypedDict
+
+from jmespath.functions import Functions
 
 class Options:
-    dict_cls: Any
-    custom_functions: Any
-    def __init__(self, dict_cls: Any | None = ..., custom_functions: Any | None = ...) -> None: ...
+    dict_cls: Callable[[], MutableMapping[Any, Any]]
+    custom_functions: Functions
```
AlexWaygood, line 9:
```suggestion
    dict_cls: Callable[[], MutableMapping[Any, Any]] | None
    custom_functions: Functions | None
```

AlexWaygood (approved):
Thanks! All looks good now, just noticed one more thing. I'll apply it and then merge once the CI has finished.

stubs/jmespath/jmespath/visitor.pyi
mondeja: Great 👍🏼

PR #9342, opened 2022-12-07
3rd-party stubtest: run on Python 3.10

AlexWaygood:


stubs/croniter/@tests/stubtest_allowlist.txt
```diff
@@ -2,9 +2,4 @@
 croniter.croniter.EXPANDERS
 croniter.croniter.HashExpander
 croniter.croniter.VALID_LEN_EXPRESSION
-croniter.croniter.hash_expression_re
-croniter.croniter.only_int_re
-croniter.croniter.special_weekday_re
-croniter.croniter.star_or_int_re
-croniter.croniter.step_search_re
```
AlexWaygood, line 9: I'm pretty confused about why stubtest errors on these being missing from the stub when run on Python 3.9, but not when run on Python 3.10. I can reproduce this behaviour locally (the same version of `croniter` is installed, `1.3.8`).

Here's the results on Python 3.9 if I remove these items from the allowlist:

```
error: croniter.croniter.hash_expression_re is not present in stub
Stub: in file C:\Users\alexw\coding\typeshed\stubs\croniter\croniter\croniter.pyi
MISSING
Runtime:
re.compile('^(?P<hash_type>h|r)(\\((?P<range_begin>\\d+)-(?P<range_end>\\d+)\\))?(\\/(?P<divisor>...

error: croniter.croniter.only_int_re is not present in stub
Stub: in file C:\Users\alexw\coding\typeshed\stubs\croniter\croniter\croniter.pyi
MISSING
Runtime:
re.compile('^\\d+$')

error: croniter.croniter.special_weekday_re is not present in stub
Stub: in file C:\Users\alexw\coding\typeshed\stubs\croniter\croniter\croniter.pyi
MISSING
Runtime:
re.compile('^(\\w+)#(\\d+)|l(\\d+)$')

error: croniter.croniter.star_or_int_re is not present in stub
Stub: in file C:\Users\alexw\coding\typeshed\stubs\croniter\croniter\croniter.pyi
MISSING
Runtime:
re.compile('^(\\d+|\\*)$')

error: croniter.croniter.step_search_re is not present in stub
Stub: in file C:\Users\alexw\coding\typeshed\stubs\croniter\croniter\croniter.pyi
MISSING
Runtime:
re.compile('^([^-]+)-([^-/]+)(/(\\d+))?$')
```

But I can't access these variables at runtime, not on Python 3.9 or Python 3.10:

```pycon
>>> import croniter.croniter
>>> croniter.croniter.hash_expression_re
AttributeError: type object 'croniter' has no attribute 'hash_expression_re'
```

So why does stubtest error on these being missing on the stub when run on Python 3.9? And what's different about Python 3.10?

AlexWaygood:
@Avasam, are the new `win32` errors just something that we need to allowlist?

srittau (approved):
Weird, but this mystery doesn't need to be solved in this PR.

stubs/croniter/@tests/stubtest_allowlist.txt
JelleZijlstra: I don't know what's different about 3.10 but you can't access them because `croniter.croniter` is a type that shadows the `croniter.croniter` module. I can access these regexes in 3.9:
```
In [19]: sys.modules["croniter.croniter"].step_search_re
Out[19]: re.compile(r'^([^-]+)-([^-/]+)(/(.*))?$', re.UNICODE)
```

AlexWaygood: > I don't know what's different about 3.10 but you can't access them because `croniter.croniter` is a type that shadows the `croniter.croniter` module.

Yeah, I get that -- the behaviour on 3.10 makes more sense to me! But I don't get what changed. I can access the regexes through the same `sys.modules` indirection on 3.10 as well:

```pycon
Python 3.10.7 (tags/v3.10.7:6cc6b13, Sep  5 2022, 14:08:36) [MSC v.1933 64 bit (AMD64)] on win32
>>> import croniter.croniter
>>> import sys
>>> sys.modules["croniter.croniter"].step_search_re
re.compile('^([^-]+)-([^-/]+)(/(\\d+))?$')
```

hauntsaninja:
Here's why:
```
(typeshed-bhth) ~/dev/typeshed main λ python3.10
Python 3.10.6 (main, Aug  7 2022, 02:34:30) [Clang 13.1.6 (clang-1316.0.21.2.5)] on darwin
>>> import sys
>>> import croniter.croniter
>>> sys.modules["croniter.croniter"].step_search_re.__module__
're'
>>> 
(typeshed-bhth) ~/dev/typeshed main λ python3.9                        
Python 3.9.9 (main, Aug  7 2022, 02:25:38) 
[Clang 13.1.6 (clang-1316.0.21.2.5)] on darwin
>>> import sys
>>> import croniter.croniter
>>> sys.modules["croniter.croniter"].step_search_re.__module__
AttributeError: 're.Pattern' object has no attribute '__module__'
```
We use `__module__` as a heuristic to detect whether a symbol is imported, in which case it's fine for it to be missing

hauntsaninja:
Relevant heuristic: https://github.com/hauntsaninja/mypy/blob/ceb976f1b9bc4c3ae7cd07a5222ec076f6c7ab72/mypy/stubtest.py#L313

AlexWaygood:
Aaahh, that explains it. Thanks!

Maybe we should use the `__module__` heuristic only for class objects -- WDYT? Instances of a class inherit the `__module__` attribute from the class, even if they're created in a different module to the class:

```pycon
>>> from collections import Counter
>>> Counter().__module__
'collections'
```

Avasam:
> @Avasam, are the new `win32` errors in CI just something that we need to allowlist? :)

Yeah, those can be added to the allowlist under `# failed to import, ModuleNotFoundError: No module named 'axdebug'`

AlexWaygood:
> > @Avasam, are the new `win32` errors in CI just something that we need to allowlist? :)
> 
> Yeah, those can be added to the allowlist under `# failed to import, ModuleNotFoundError: No module named 'axdebug'`

Thanks!

PR #9346, opened 2022-12-10
Use `Literal` for `compression` in `zipfile`

Viicos:
According to [docs](https://docs.python.org/3/library/zipfile.html?highlight=zipfile#zipfile.ZipFile):

> compression is the ZIP compression method to use when writing the archive, and should be [ZIP_STORED](https://docs.python.org/3/library/zipfile.html?highlight=zipfile#zipfile.ZIP_STORED), [ZIP_DEFLATED](https://docs.python.org/3/library/zipfile.html?highlight=zipfile#zipfile.ZIP_DEFLATED), [ZIP_BZIP2](https://docs.python.org/3/library/zipfile.html?highlight=zipfile#zipfile.ZIP_BZIP2) or [ZIP_LZMA](https://docs.python.org/3/library/zipfile.html?highlight=zipfile#zipfile.ZIP_LZMA); unrecognized values will cause [NotImplementedError](https://docs.python.org/3/library/exceptions.html#NotImplementedError) to be raised.

`Literal` could be used for the `compresslevel` parameter as well:

> The compresslevel parameter controls the compression level to use when writing files to the archive. When using [ZIP_STORED](https://docs.python.org/3/library/zipfile.html?highlight=zipfile#zipfile.ZIP_STORED) or [ZIP_LZMA](https://docs.python.org/3/library/zipfile.html?highlight=zipfile#zipfile.ZIP_LZMA) it has no effect. When using [ZIP_DEFLATED](https://docs.python.org/3/library/zipfile.html?highlight=zipfile#zipfile.ZIP_DEFLATED) integers 0 through 9 are accepted (see [zlib](https://docs.python.org/3/library/zlib.html#zlib.compressobj) for more information). When using [ZIP_BZIP2](https://docs.python.org/3/library/zipfile.html?highlight=zipfile#zipfile.ZIP_BZIP2) integers 1 through 9 are accepted (see [bz2](https://docs.python.org/3/library/bz2.html#bz2.BZ2File) for more information).

But this would require extra overloads for each `compression` value, so considering the fact that there's already a split between `3.8` and `3.11` this would make things more complicated.

Note: We could use `Literal` for the module constants, e.g. `ZIP_STORED`, `ZIP_DEFLATED`, etc. I can add another commit to this PR if required.

### DIFF
--- a/stdlib/zipfile.pyi
+++ b/stdlib/zipfile.pyi
@@ -275,10 +275,10 @@ if sys.version_info >= (3, 8):
 
 def is_zipfile(filename: StrOrBytesPath | _SupportsReadSeekTell) -> bool: ...
 
-ZIP_STORED: int
-ZIP_DEFLATED: int
-ZIP64_LIMIT: int
-ZIP_FILECOUNT_LIMIT: int
-ZIP_MAX_COMMENT: int
-ZIP_BZIP2: int
-ZIP_LZMA: int
+ZIP_STORED: Literal[0]
+ZIP_DEFLATED: Literal[8]
+ZIP64_LIMIT: Literal[2147483647]
+ZIP_FILECOUNT_LIMIT: Literal[65535]
+ZIP_MAX_COMMENT: Literal[65535]
+ZIP_BZIP2: Literal[12]
+ZIP_LZMA: Literal[14]
--- a/stdlib/zipfile.pyi
+++ b/stdlib/zipfile.pyi
@@ -277,8 +277,8 @@ def is_zipfile(filename: StrOrBytesPath | _SupportsReadSeekTell) -> bool: ...
 
 ZIP_STORED: Literal[0]
 ZIP_DEFLATED: Literal[8]
-ZIP64_LIMIT: Literal[2147483647]
-ZIP_FILECOUNT_LIMIT: Literal[65535]
-ZIP_MAX_COMMENT: Literal[65535]
+ZIP64_LIMIT: int
+ZIP_FILECOUNT_LIMIT: int
+ZIP_MAX_COMMENT: int
 ZIP_BZIP2: Literal[12]
 ZIP_LZMA: Literal[14]
--- a/stdlib/zipfile.pyi
+++ b/stdlib/zipfile.pyi
@@ -29,6 +29,7 @@ _DateTuple: TypeAlias = tuple[int, int, int, int, int, int]
 _ReadWriteMode: TypeAlias = Literal["r", "w"]
 _ReadWriteBinaryMode: TypeAlias = Literal["r", "w", "rb", "wb"]
 _ZipFileMode: TypeAlias = Literal["r", "w", "x", "a"]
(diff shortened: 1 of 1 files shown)


AlexWaygood (commented):
> Note: We could use `Literal` for the module constants, e.g. `ZIP_STORED`, `ZIP_DEFLATED`, etc. I can add another commit to this PR if required.

We need to either make both changes together, or not at all. Otherwise, type checkers will emit false positives on code like this, due to the fact that they will infer `ZIP_DEFLATED` as being an `int`, while the function requires a `Literal`:

```python
import zipfile

zip_output = zipfile.ZipFile("some/path.zip", "w", zipfile.ZIP_DEFLATED, allowZip64=True)
```

This is the cause of the `pip` error in the `mypy_primer` output: https://github.com/pypa/pip/blob/5f3f592c4581a059ff9de0fb8052ef5c6ef25fd4/src/pip/_internal/req/req_install.py#L714-L719

Note: I'm not yet sure I'm sold on the idea of making a change here at all, due to the risk of false-positive errors being emitted. But in order to see the impact of the change, we'll need to see what it looks like if we change the module-level constants to `Literal` values as well.

Viicos:
> We need to either make both changes together, or not at all. Otherwise, type checkers will emit false positives on code like this, due to the fact that they will infer ZIP_DEFLATED as being an int, while the function requires a Literal

Yes I've been checking the reported `mypy_primer` errors and they all seem related to this. I'll add another commit and see what is the updated report.

Viicos:
Related code failing:

```python
# vision (https://github.com/pytorch/vision)

_ZIP_COMPRESSION_MAP: Dict[str, int] = {
    ".bz2": zipfile.ZIP_BZIP2,
    ".xz": zipfile.ZIP_LZMA,
}


def _extract_zip(from_path: str, to_path: str, compression: Optional[str]) -> None:
    with zipfile.ZipFile(
        from_path, "r", compression=_ZIP_COMPRESSION_MAP[compression] if compression else zipfile.ZIP_STORED
    ) as zip:
        zip.extractall(to_path)
```

```python
# jinja (https://github.com/pallets/jinja)
zip_file = ZipFile(
    target, "w", dict(deflated=ZIP_DEFLATED, stored=ZIP_STORED)[zip]
)
```

This would require a manual fix on these repos. So this is still breaking things, but imo it would result in safer code, as `ZipFile` will throw a `NotImplementedError` if the compression method is not supported.

srittau (approved):
LGTM. I think we need to bite the bullet and need to make these kind of changes rather sooner than later.

AlexWaygood (commented):
Mostly LGTM too, though I'm not sure it makes as much sense for `ZIP64_LIMIT`, `ZIP_FILECOUNT_LIMIT` and `ZIP_MAX_COMMENT` to be made `Literal`s as it does for the other constants. Those are very big numbers, and the constants serve a different purpose to `ZIP_STORED`, `ZIP_DEFLATED`, `ZIP_BZIP2` and `ZIP_LZMA`.

AlexWaygood (approved):
Thanks @Viicos!

Issue #9296, opened 2022-11-29
calling .register on typing.KeysView subclass causes false positive error `Member "register" is unknown`


jab:
Type checkers that include the recent changes from ~~https://github.com/python/typeshed/pull/8977~~ https://github.com/python/typeshed/pull/9058 (e.g. pyright v1.1.281) now incorrectly emit an error like `Member "register" is unknown` for the following line of code:

https://github.com/jab/bidict/blob/522c9b69249d66c201fed608f95a5da8d3064ed8/bidict/_base.py#L51

Type checkers that do not include those changes (e.g. pyright v1.1.278) do not emit any error for that line, as expected.

Copy/pasted excerpt:
```python
import typing as t

from ._typing import KT  # a typevar


class BidictKeysView(t.KeysView[KT], t.ValuesView[KT]):
    """Since the keys of a bidict are the values of its inverse (and vice versa),
    the :class:`~collections.abc.ValuesView` result of calling *bi.values()*
    is also a :class:`~collections.abc.KeysView` of *bi.inverse*.
    """


dict_keys: t.Type[t.KeysView[t.Any]] = type({}.keys())
BidictKeysView.register(dict_keys)  # <-- false positive error emitted here
```

Please let me know if there is any other info I can provide. Thank you!

JelleZijlstra:
I don't think the typeshed change you link to is relevant. Probably #9058 is the culprit instead.

hauntsaninja:
It's relevant in that #9058 wouldn't have happened if not for #8977.

Also, #8977 was mostly done for aesthetic reasons, so if we need to fix in typeshed, I suggest we roll back #8977 (as per option 2 in https://github.com/python/typeshed/pull/8977#issuecomment-1297872394)

Of course, in the medium term we should also fix the mypy bug this exposed https://github.com/python/typeshed/pull/9058#issuecomment-1299972677 / https://github.com/python/mypy/issues/13986 (maybe it's already fixed?). The reason CI didn't complain is because this needs https://github.com/python/mypy/pull/13579 to break, which was not part of a released mypy at the time.

I'm assuming @jab ran into this in the context of bidict. We do have bidict in mypy_primer and didn't see any regressions, so maybe this is something mypy has some special casing for?

AlexWaygood:
[...]

PR #9348, opened 2022-12-10
Fix issue #9296 (calling register on KeysView subclass causes false positive error)

jab:
As per discussion in #9296, this PR reverts the commits in #8977 and #9058 that resulted in false positive errors when e.g. calling register() on a KeysView subclass, until a non-buggy implementation is landed. This also adds a regression test.

### DIFF
--- a/stdlib/builtins.pyi
+++ b/stdlib/builtins.pyi
@@ -804,7 +804,7 @@ class bytearray(MutableSequence[int], ByteString):
     def __alloc__(self) -> int: ...
 
 @final
-class memoryview(Sized, Sequence[int]):
+class memoryview(Sequence[int]):
     @property
     def format(self) -> str: ...
     @property
--- a/stdlib/builtins.pyi
+++ b/stdlib/builtins.pyi
@@ -804,7 +804,7 @@ class bytearray(MutableSequence[int], ByteString):
     def __alloc__(self) -> int: ...
 
 @final
-class memoryview(Sequence[int]):
+class memoryview(Sized, Sequence[int]):
     @property
     def format(self) -> str: ...
     @property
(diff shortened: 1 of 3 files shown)


stdlib/builtins.pyi
```diff
@@ -804,7 +804,7 @@ class bytearray(MutableSequence[int], ByteString):
     def __alloc__(self) -> int: ...
 
 @final
-class memoryview(Sequence[int]):
+class memoryview(Sized, Sequence[int]):
```
JelleZijlstra, line 807: Is this change necessary? I would prefer removing explicit protocol base classes if possible, since memoryview doesn't inherit from Sized at runtime.

jab: I'm not sure, I was following would I thought was the consensus after I saw https://github.com/python/typeshed/issues/9296#issuecomment-1332377034 (which suggested including this reversion), and then 10 days having gone by without a differing suggestion.

Can you tell me how if you want me to try undoing this and verifying that the regression is still fixed locally before I push again? I tried adding some garbage to the regression test here locally and then re-running `python tests/pyright_test.py` but it still passed, so I'm not sure how to check this.

JelleZijlstra: Let's just undo this change and see if CI succeeds.

```suggestion
class memoryview(Sequence[int]):
```

JelleZijlstra (approved):

stdlib/builtins.pyi
jab: Thanks, @JelleZijlstra. Btw, looks like the message of the commit that you merged to main still includes `Revert "typing: remove metaclass from Sized (#9058). This reverts commit a3ce51209544d64d03018b3927051375f1bebe82.`

More importantly, I would still like to know how to check if the regression test I wrote is actually doing its job: Did the tests that passed here in CI actually exercise the code path where this bug was observed? I'm concerned that when I changed "MyKeysView.register" to "MyKeysView.garbage" and tried to have pyright test that code, I didn't see any failures.

PR #9349, opened 2022-12-10
`regr_test.py`: improve interaction between `--all` and `--quiet` flags

JelleZijlstra:
cf. #9348 cc @jab

### DIFF
--- a/test_cases/stdlib/typing/check_regression_issue_9296.py
+++ b/test_cases/stdlib/typing/check_regression_issue_9296.py
@@ -13,4 +13,4 @@ d: dict[t.Any, t.Any] = {}
 dict_keys = type(d.keys())
 
 # This should not cause an error like `Member "register" is unknown`:
-MyKeysView.regixster(dict_keys)
+MyKeysView.register(dict_keys)
--- a/test_cases/stdlib/typing/check_regression_issue_9296.py
+++ b/test_cases/stdlib/typing/check_regression_issue_9296.py
@@ -13,4 +13,4 @@ d: dict[t.Any, t.Any] = {}
 dict_keys = type(d.keys())
 
 # This should not cause an error like `Member "register" is unknown`:
-MyKeysView.register(dict_keys)
+MyKeysView.regixster(dict_keys)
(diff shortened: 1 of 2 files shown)


JelleZijlstra:
```
pyright 1.1.278
/home/runner/work/typeshed/typeshed/test_cases/stdlib/typing/check_regression_issue_9296.py
  /home/runner/work/typeshed/typeshed/test_cases/stdlib/typing/check_regression_issue_9296.py:16:12 - error: Cannot access member "regixster" for type "Type[MyKeysView[KT@MyKeysView]]"
    Member "regixster" is unknown (reportGeneralTypeIssues)
  /home/runner/work/typeshed/typeshed/test_cases/stdlib/typing/check_regression_issue_9296.py:16:1 - error: Type of "regixster" is unknown (reportUnknownMemberType)
```
It works as expected.

AlexWaygood:
And we have a falling mypy run as well: https://github.com/python/typeshed/actions/runs/3665455929/jobs/6196543252

(Though it looks like I could probably sink a little more work into improving that traceback)

JelleZijlstra:
Is the reason the failure shows up ~16 times that we run mypy in a matrix (Python version, OS) and report the failure every time?

AlexWaygood:
> Is the reason the failure shows up ~16 times that we run mypy in a matrix (Python version, OS) and report the failure every time?

Yup! Just pushed a commit to this PR branch that hopefully clarifies things a bit?

The issue was that the new `--quiet` flag didn't really interact very well with the `--all` flag

JelleZijlstra:
Nice, feel free to hijack this PR for your change, or just open a new PR.

A more radical change could be to show only the first failure if there are multiple identical ones. After all, in most cases test cases are going to fail in the same way on multiple runs.

AlexWaygood:
> After all, in most cases test cases are going to fail in the same way on multiple runs.

This is true, but it won't _always_ be the case. I feel like this solution is ~good enough for now. I'm gonna commence hijacking ;)

test_cases/stdlib/typing/check_regression_issue_9296.py
```diff
@@ -13,4 +13,4 @@ class MyKeysView(t.KeysView[KT]):
 dict_keys = type(d.keys())
 
 # This should not cause an error like `Member "register" is unknown`:
-MyKeysView.register(dict_keys)
+MyKeysView.regixster(dict_keys)
```
AlexWaygood, line 16:
```suggestion
MyKeysView.register(dict_keys)
```

PR #9352, opened 2022-12-11
Annotate `unittest.TestCase.skipTest()` as no-return

Terrance:
This method unconditionally raises `unittest.SkipTest`, which ends a test method early.  `unittest.TestCase.fail()` works similarly, and is already annotated with `NoReturn` to indicate this behaviour.


AlexWaygood:
Thanks!

PR #9353, opened 2022-12-12
Mark asyncio.Runner as final

FasterSpeeding:
As per the comment at https://github.com/python/cpython/blob/71b032635d757348cb915c2e8d99392f5bbcb059/Lib/asyncio/runners.py#L47


hauntsaninja (approved):
Interesting, thanks!

Issue #9246, opened 2022-11-22
Automated stub uploader constantly failing because types-pika already exists on PyPI


ilevkivskyi:
See for example https://github.com/typeshed-internal/stub_uploader/actions/runs/3521987995/jobs/5904439698

PyPI: https://pypi.org/project/types-pika/
Ref https://github.com/python/typeshed/pull/9200

cc @srittau @JelleZijlstra @qubidt

srittau:
It's too bad there aren't protected namespaces on PyPI. I suggest we add a field `package` or similar to `METADATA.toml` where we can put an alternative package name if necessary.

AlexWaygood:
Is it worth temporarily removing the stubs from typeshed for now, as a stopgap measure?

ilevkivskyi:
Btw `mypy` (and I guess other tools too) will automatically propose to install `types-xxx` (if types for `xxx` are missing). I guess we can discuss with the owner of the original package @qubidt how to "merge efforts"?

I would propose to revert that PR for now, so that automatic uploader will work. But this is probably not urgent, I guess people can wait a day or two without stub updates (or we can use manual upload action if something is urgently needed).

(also cc @JukkaL for visibility)

srittau:
I've created a PR for the stub_uploader supporting a "stub_distribution" field if we decide to go that way: typeshed-internal/stub_uploader#76.

qubidt:
Happy to do whatever is needed to resolve this. Sorry for stepping on your namesake in the first place

AlexWaygood:
> Happy to do whatever is needed to resolve this. Sorry for stepping on your namesake in the first place

I definitely don't think this is your fault, @qubidt :) I don't think there's any reason for typeshed to have any intrinsic "right" to `types-*` names on PyPI, it's just become an assumption around these parts that those names were available!

AlexWaygood:
I do agree with @ilevkivskyi that we should work out how to collaborate, though, if that's something you'd be interested in @qubidt. It seems silly to have two different sets of stubs for `pika`.

It feels slightly unfair to ask, since you "got there first", but would you be interested in contributing your stubs to typeshed, @qubidt? It looks like you have a lot more "annotation coverage" than us at the moment over at https://github.com/qubidt/types-pika, but I think our CI checks are probably slightly more robust.

qubidt:
[...]

PR #9358, opened 2022-12-14
Distribute our pika stubs under an alternate name

srittau:
Cf. #9246


AlexWaygood (approved):

AlexWaygood:
Hmm, looks like we may need to adjust some more things on the stub-uploader side for this to be merged?

srittau:
typeshed-internal/stub_uploader#80

srittau:
And it worked: https://pypi.org/project/types-pika-ts/

PR #9360, opened 2022-12-14
pika: Make Method generic, annotate DeclareOk

srittau:


stubs/pika/pika/spec.pyi
```diff
@@ -338,9 +338,9 @@ class Exchange(Class):
         INDEX: ClassVar[int]
         def __init__(self) -> None: ...
         @property
-        def synchronous(self): ...
-        def decode(self, encoded, offset: int = ...): ...
-        def encode(self): ...
+        def synchronous(self) -> Literal[False]: ...
+        def decode(self: Self, encoded: bytes, offset: int = ...) -> Self: ...
```
srittau, line 342: `encoded` and `offset` are unused, so technically they take `object`, but there's no reason to provide wrong argument, so it's safer to demand to use the "correct" types here.

stubs/pika/pika/spec.pyi
```diff
@@ -338,9 +338,9 @@ class Exchange(Class):
...
+        def encode(self) -> list[bytes]: ...
```
srittau, line 343: Returns an empty list, but other methods return a `list[bytes]`.

stubs/pika/pika/spec.pyi
```diff
@@ -451,20 +451,18 @@ class Queue(Class):
...
-        queue: Incomplete
-        message_count: Incomplete
-        consumer_count: Incomplete
-        def __init__(
-            self, queue: Incomplete | None = ..., message_count: Incomplete | None = ..., consumer_count: Incomplete | None = ...
-        ) -> None: ...
+        queue: str
+        message_count: int
+        consumer_count: int
+        def __init__(self, queue: str, message_count: int, consumer_count: int) -> None: ...
```
srittau, line 461: See comment in stubtest_allowlist. The implementation has `None` defaults for the `__init__` arguments, but that would make instances not work. `encode()` would crash and having a `None` queue is not very useful.

stubs/pika/pika/channel.pyi
```diff
@@ -38,7 +40,7 @@ class Channel:
         exclusive: bool = ...,
         consumer_tag: str | None = ...,
         arguments: _ArgumentMapping | None = ...,
-        callback: Callable[[Method], object] | None = ...,
+        callback: Callable[[Method[Queue.DeclareOk]], object] | None = ...,
```
AlexWaygood, line 43: Should this be `ConsumeOk`? https://github.com/pika/pika/blob/22b381efa25b0b8671badcb7bb4485a7b2cefaca/pika/channel.py#L298-L299

stubs/pika/pika/spec.pyi
```diff
@@ -451,20 +451,18 @@ class Queue(Class):
         @property
         def synchronous(self): ...
```
AlexWaygood, line 452:
```suggestion
        def synchronous(self) -> Literal[True]: ...
```

stubs/pika/pika/spec.pyi
```diff
@@ -451,20 +451,18 @@ class Queue(Class):
...
         def decode(self, encoded, offset: int = ...): ...
```
AlexWaygood, line 453:
```suggestion
        def decode(self: Self, encoded: bytes, offset: int = ...) -> Self: ...
```

AlexWaygood (commented):
Mostly looks good!

AlexWaygood (approved):

PR #9362, opened 2022-12-14
Use a `TypeGuard` for `dataclasses.is_dataclass()`; refine `asdict()`, `astuple()`, `fields()`, `replace()`

### LINKED ISSUE #9345: `is_dataclass` should be a type guard and `asdict` should only work on objects with `__dataclass_fields__`
@AlexWaygood proposed the following stubs in this comment: https://github.com/python/mypy/pull/14263#issuecomment-1343337541

```python
from typing import Any, ClassVar, Protocol
from typing_extensions import TypeGuard

class _DataclassInstance(Protocol):
     __dataclass_fields__: ClassVar[dict[str, Any]]

def asdict(obj: _DataclassInstance) -> dict[str, Any]: ...
def is_dataclass(obj: object) -> TypeGuard[_DataclassInstance]: ...
```

and this works as expected with dataclasses on mypy and pyright, but it doesn't currently work on pyre because pyre doesn't know yet that dataclasses have the `__dataclass_fields__` classvar. (I haven't checked other type checkers.)

At runtime, `is_dataclass` really only checks for the `__dataclass_fields__` attribute:

https://github.com/python/cpython/blob/7a0f3c1d92ef0768e082ace19d970b0ef12e7346/Lib/dataclasses.py#L1263-L1267

However, it also works with types in addition to instances, so an overload would be needed.

See also this issue for context: https://github.com/python/mypy/issues/14215

AlexWaygood:
Fixes #9345. Refs python/mypy#14215


stdlib/dataclasses.pyi
```diff
@@ -213,7 +218,12 @@ else:
     ) -> Any: ...
 
 def fields(class_or_instance: Any) -> tuple[Field[Any], ...]: ...
-def is_dataclass(obj: Any) -> bool: ...
+@overload
+def is_dataclass(obj: _DataclassInstance | type[_DataclassInstance]) -> bool: ...
+@overload
+def is_dataclass(obj: type) -> TypeGuard[type[_DataclassInstance]]: ...
+@overload
+def is_dataclass(obj: object) -> TypeGuard[_DataclassInstance]: ...
```
AlexWaygood, line 226:
```suggestion
def is_dataclass(obj: object) -> TypeGuard[_DataclassInstance | type[_DataclassInstance]]: ...
```

stdlib/dataclasses.pyi
```diff
@@ -30,6 +30,11 @@ __all__ = [
 if sys.version_info >= (3, 10):
     __all__ += ["KW_ONLY"]
 
+class _DataclassInstance(Protocol):
+    __dataclass_fields__: ClassVar[dict[str, Any]]
```
AlexWaygood, line 34:
```suggestion
    __dataclass_fields__: ClassVar[dict[str, Field[Any]]]
```

test_cases/stdlib/check_dataclasses.py
```diff
@@ -0,0 +1,66 @@
...
+from typing_extensions import assert_type
+
+
+@dc.dataclass
+class Foo:
+    attr: str
+
+
+assert_type(dc.fields(Foo), Tuple[dc.Field[Any], ...])
+dc.asdict(Foo)  # type: ignore
```
AlexWaygood, line 12:
```suggestion
# These should fail due to the fact it's a dataclass class, not an instance
dc.asdict(Foo)  # type: ignore
```

test_cases/stdlib/check_dataclasses.py
```diff
@@ -0,0 +1,67 @@
+import dataclasses as dc
```
AlexWaygood, line 1:
```suggestion
from __future__ import annotations

import dataclasses as dc
```

test_cases/stdlib/check_dataclasses.py
```diff
@@ -0,0 +1,69 @@
...
+assert_type(dc.replace(f, attr="new"), Foo)
+
+if dc.is_dataclass(f):
+    # The inferred type doesn't change
+    # if it's already known to be a subtype of _DataclassInstance
+    assert_type(f, Foo)
+
+
+def test_other_isdataclass_overloads(x: type, y: object) -> None:
+    dc.fields(x)  # TODO: why does this pass mypy? It should fail, ideally...
```
AlexWaygood, line 38:
```suggestion
    # TODO: pyright correctly emits an error on this, but mypy does not -- why?
    # dc.fields(x)

```

AlexWaygood:
The pyright tests are failing due to https://github.com/microsoft/pyright/issues/4339.

Links to the `mypy_primer` hits:

- `porcupine` is using `dataclasses.asdict` on a class that isn't actually a dataclass, but is used as the base class for many dataclasses: https://github.com/Akuli/porcupine/blob/2845a5d537d85ddd7a7b7c8cbcb3e3a4d8d924ca/porcupine/utils.py#L369-L408
- `xarray-dataclasses`:
  - The code where the error is being emitted is https://github.com/astropenguin/xarray-dataclasses/blob/1f9816f19dd2da69b50432add4a256ac380a1134/xarray_dataclasses/v2/specs.py#L110-L119
  - The root cause, I think is that they're using a slightly more complex protocol than the one I've got here, which is generic around a `ParamSpec`: https://github.com/astropenguin/xarray-dataclasses/blob/1f9816f19dd2da69b50432add4a256ac380a1134/xarray_dataclasses/v2/typing.py#L49-L55.
- `homeassistant`:
  - The first error is a true positive:
    - The error is emitted here:  https://github.com/home-assistant/core/blob/3d7a3afca70c0e760ffb2dbf3c72ddec24fa0c0e/homeassistant/components/p1_monitor/diagnostics.py#L45
    - `coordinator.data[SERVICE_WATERMETER]` is defined as having an optional type here: https://github.com/home-assistant/core/blob/3d7a3afca70c0e760ffb2dbf3c72ddec24fa0c0e/homeassistant/components/p1_monitor/__init__.py#L67
  - Haven't looked at the second one yet

stdlib/dataclasses.pyi
```diff
@@ -212,8 +217,13 @@ else:
         metadata: Mapping[Any, Any] | None = ...,
     ) -> Any: ...
 
-def fields(class_or_instance: Any) -> tuple[Field[Any], ...]: ...
-def is_dataclass(obj: Any) -> bool: ...
+def fields(class_or_instance: _DataclassInstance | type[_DataclassInstance]) -> tuple[Field[Any], ...]: ...
+@overload
+def is_dataclass(obj: _DataclassInstance | type[_DataclassInstance]) -> bool: ...
```
KotlinIsland, line 222: Should this return `True`?

AlexWaygood: Sure

AlexWaygood:
```suggestion
def is_dataclass(obj: _DataclassInstance | type[_DataclassInstance]) -> Literal[True]: ...
```

AlexWaygood:
@stroxler / @pradeep90: from @thomkeh's experiments in #9345, it looks like pyre doesn't currently synthesize a `__dataclass_fields__` `ClassVar` for dataclass classes in the same way that mypy and pyright do. That means that if this PR were to be merged, it might cause false positives for pyre users. Would you be okay with it being merged, or would you rather we held off for now?

Same question for @rchen152 -- not sure if pytype synthesizes this attribute for dataclasses or not? Would merging this cause issues for pytype?

rchen152:
Yep, pytype synthesizes the `__dataclass_fields__` attribute! Somewhat amusingly, the entire reason we added it was to get this `pytype_extensions.Dataclass` protocol working: https://github.com/google/pytype/blob/8d0970083128ffc7f6042b08fad076f8c2ee01a6/pytype_extensions/__init__.py#L67

thomkeh:
Should `typing_extensions` also get a `Dataclass` protocol then? :thinking:

AlexWaygood:
> Yep, pytype synthesizes the __dataclass_fields__ attribute!

Brilliant, thanks @rchen152!

> Should `typing_extensions` also get a `Dataclass` protocol then? 🤔

_Maybe_. But it's probably better to add it to typeshed first and wait a while (a few months, at least) to see if it's stable before adding it to `typing_extensions`.

pradeep90:
> @stroxler / @pradeep90: from @thomkeh's experiments in #9345, it looks like pyre doesn't currently synthesize a `__dataclass_fields__` `ClassVar` for dataclass classes in the same way that mypy and pyright do. That means that if this PR were to be merged, it might cause false positives for pyre users. Would you be okay with it being merged, or would you rather we held off for now?

Yes, feel free to merge this change.

Pyre doesn't currently synthesize a `__dataclass_fields__` `ClassVar` for dataclasses. I'll fix it on the Pyre side. We vendor a typeshed version manually anyway, so we will ensure the fix has landed before we upgrade next.

cdce8p:
I've been working on updating Home Assistant for these changes. Some notes

There are a couple of places where I needed to use the dataclass protocol itself. One example are dataclass mixin classes. From that I think it would make sense to either include it in `typing_extensions` (like others suggested already) or in `_typeshed`.

At least for mypy, `pydantic` dataclasses don't seem to work, yet. Haven't had the time to look into it.
```py
import dataclasses
from pydantic.dataclasses import dataclass

@dataclass
class Info:
    var: str

def func(info: Info) -> None:
    dataclasses.asdict(info)
```

```
t:9: error: No overload variant of "asdict" matches argument type "Info"  [call-overload]
t:9: note: Possible overload variants:
t:9: note:     def asdict(obj: _DataclassInstance) -> Dict[str, Any]
t:9: note:     def [_T] asdict(obj: _DataclassInstance, *, dict_factory: Callable[[List[Tuple[str, Any]]], _T]) -> _T
```

Tested with `pydantic==1.10.4`. It's quite common that an upstream library uses `pydantic` dataclasses (-> `Info`), whereas the caller uses `dataclasses` (-> `func`).

tmke8:
As long as pydantic's dataclass has the `__dataclass_fields__` attribute, it should work, right?

AlexWaygood:
I think pydantic has a custom mypy plugin, so it's probably an issue with the plugin

AlexWaygood:
Anyway, thanks (as ever) for the very early testing here @cdce8p! Given how new/experimental the protocol is, and how it doesn't work particularly well with pyright due to https://github.com/microsoft/pyright/issues/4339, I think it definitely isn't a suitable candidate to be added to `typing_extensions` just yet. But adding it to `_typeshed` is something we can easily do now, so I'll try to get on that  (or alternatively, I'll happily accept a PR if you get there first ;)

AlexWaygood:
(To be clear, this PR shouldn't cause any regressions for pyright users, the pyright issue just means that this PR doesn't provide quite as much of a type-safety improvement for pyright users as it does for mypy users.)

cdce8p:
Small update regarding pydantic. Enabling it's mypy plugin does resolve the issue
```ini
[mypy]
plugins = pydantic.mypy
```
https://docs.pydantic.dev/mypy_plugin/

PR #9366, opened 2022-12-15
shutil.rmtree: add avoids_symlink_attacks attribute

### LINKED ISSUE #9365: Function attribute shutil.rmtree.avoids_symlink_attacks yields error
**Bug Report**

Mypy does non recognize function attribute `shutil.rmtree.avoids_symlink_attacks`.

**To Reproduce**

```python
from shutil import rmtree
print(rmtree.avoids_symlink_attacks)
```

**Expected Behavior**

No error should be recognized. Exit code should be 0.

**Actual Behavior**

```bash
$ mypy test.py 
test.py:2: error: "Callable[[Union[str, bytes, PathLike[str], PathLike[bytes]], bool, Optional[Callable[[Callable[..., Any], Any, Any], object]]], None]" has no attribute "avoids_symlink_attacks"  [attr-defined]
```

and exit code 1.

**Your Environment**

- Mypy version used: 0.991
- Mypy command-line flags: none
- Mypy configuration options from `mypy.ini` (and other config files): 
   ```
   [mypy]
   python_version = 3.10

   # Without this, mypy aggressively complains about missing imports referring to site packages.
   ignore_missing_imports = True

   ```
- Python version used: 3.10.8

Thanks for mypy! It is really very helpful for large projects!

AlexWaygood:
Fixes #9365

### DIFF
--- a/tests/stubtest_allowlists/py3_common.txt
+++ b/tests/stubtest_allowlists/py3_common.txt
@@ -165,6 +165,7 @@ poplib.POP3_SSL.stls  # bad declaration of inherited function. See poplib.pyi
 pyexpat.expat_CAPI
 select.poll  # Depends on configuration
 selectors.DevpollSelector  # Depends on configuration
+shutil.rmtree  # stubtest doesn't like that we have this as an instance of a callback protocol instead of a function
 socketserver.BaseServer.RequestHandlerClass  # is defined as a property, because we need `Self` type
 socketserver.BaseServer.fileno  # implemented in derived classes
 socketserver.BaseServer.get_request  # implemented in derived classes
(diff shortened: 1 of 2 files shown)


tests/stubtest_allowlists/py3_common.txt
```diff
@@ -165,6 +165,7 @@ poplib.POP3_SSL.stls  # bad declaration of inherited function. See poplib.pyi
 pyexpat.expat_CAPI
 select.poll  # Depends on configuration
 selectors.DevpollSelector  # Depends on configuration
+shutil.rmtree  # stubtest doesn't like that we have this as an instance of a callback protocol instead of a function
```
srittau, line 168: I don't like that either. :/ But it's probably the best solution if it works.

srittau (approved):
Probably the best solution for this if it works.

tests/stubtest_allowlists/py3_common.txt
AlexWaygood: Yup, it's horrible. But I know of no other way to represent this in the type system :(

PR #9378, opened 2022-12-17
Catch venv creation failure and print a helpful error message

### LINKED ISSUE #9375: stubtest_third_party.py fails on Debian systems with a bad error message due to no ensurepip module
I was testing a change locally and while running stubtest the test exits with a traceback like this one:

```console
$ python3.8 t.py
  File "t.py", line 1, in <module>
    import venv; venv.create("/tmp/venv_dir_108341Y4HR", with_pip=True, clear=True)
  File "/usr/lib/python3.8/venv/__init__.py", line 391, in create
    builder.create(env_dir)
  File "/usr/lib/python3.8/venv/__init__.py", line 68, in create
    self._setup_pip(context)
  File "/usr/lib/python3.8/venv/__init__.py", line 289, in _setup_pip
    subprocess.check_output(cmd, stderr=subprocess.STDOUT)
  File "/usr/lib/python3.8/subprocess.py", line 415, in check_output
    return run(*popenargs, stdout=PIPE, timeout=timeout, check=True,
  File "/usr/lib/python3.8/subprocess.py", line 516, in run
    raise CalledProcessError(retcode, process.args,
subprocess.CalledProcessError: Command '['/tmp/venv_dir_108341Y4HR/bin/python3.8', '-Im', 'ensurepip', '--upgrade', '--default-pip']' returned non-zero exit status 1.
```
I have the following `t.py`: `import venv; venv.create("/tmp/venv_dir_108341Y4HR", with_pip=True, clear=True)`
and python3.8 installed but not `python3.8-venv` on Ubunu 22.04.

In stubtest, the error is produced on this line:
https://github.com/python/typeshed/blob/04d4cbd92f2f6b2f845e1c3e0b257a96c3e170a5/tests/stubtest_third_party.py#L46

The traceback isn't very helpful and it doesn't clearly show the origin of the problem. I spent quiet some time to figure it out since I am usually a virtualvenv user and rarely interact with venv. Would you accept a PR that prints a more helpful message in case a subprocess.CalledProcessError is raised by this line?

hamdanal:
Closes #9375

### DIFF
--- a/tests/stubtest_third_party.py
+++ b/tests/stubtest_third_party.py
@@ -43,7 +43,17 @@ def run_stubtest(dist: Path, *, verbose: bool = False, specified_stubs_only: boo
 
     with tempfile.TemporaryDirectory() as tmp:
         venv_dir = Path(tmp)
-        venv.create(venv_dir, with_pip=True, clear=True)
+        try:
+            venv.create(venv_dir, with_pip=True, clear=True)
+        except subprocess.CalledProcessError as e:
+            print_command_failure(
+                (
+                    "Failed to create virtual environment. Note that on some linux systems, you need to install the `python3-venv`"
+                    " package via your system package manager."
+                ),
+                e,
+            )
+            return False
 
         if sys.platform == "win32":
             pip = venv_dir / "Scripts" / "pip.exe"
--- a/tests/stubtest_third_party.py
+++ b/tests/stubtest_third_party.py
@@ -47,7 +47,7 @@ def run_stubtest(dist: Path, *, verbose: bool = False, specified_stubs_only: boo
             venv.create(venv_dir, with_pip=True, clear=True)
         except subprocess.CalledProcessError as e:
             msg = (
-                "Failed to create virtual environment. Note that on some linux systems, you need to install the `python3-venv`"
+                "Failed to create virtual environment. Note that on some linux systems, you may need to install the `python3-venv`"
                 " package via your system package manager."
             )
             print_command_failure(msg, e)
--- a/tests/stubtest_third_party.py
+++ b/tests/stubtest_third_party.py
@@ -51,7 +51,7 @@ def run_stubtest(dist: Path, *, verbose: bool = False, specified_stubs_only: boo
                 " package via your system package manager."
             )
             print_command_failure(msg, e)
-            return False
+            sys.exit(2)
 
         if sys.platform == "win32":
             pip = venv_dir / "Scripts" / "pip.exe"
--- a/tests/stubtest_third_party.py
+++ b/tests/stubtest_third_party.py
@@ -46,13 +46,11 @@ def run_stubtest(dist: Path, *, verbose: bool = False, specified_stubs_only: boo
         try:
             venv.create(venv_dir, with_pip=True, clear=True)
         except subprocess.CalledProcessError as e:
-            print_command_failure(
-                (
-                    "Failed to create virtual environment. Note that on some linux systems, you need to install the `python3-venv`"
-                    " package via your system package manager."
-                ),
-                e,
+            msg = (
+                "Failed to create virtual environment. Note that on some linux systems, you need to install the `python3-venv`"
+                " package via your system package manager."
             )
+            print_command_failure(msg, e)
             return False
 
         if sys.platform == "win32":
--- a/tests/stubtest_third_party.py
+++ b/tests/stubtest_third_party.py
@@ -46,12 +46,13 @@ def run_stubtest(dist: Path, *, verbose: bool = False, specified_stubs_only: boo
         try:
             venv.create(venv_dir, with_pip=True, clear=True)
         except subprocess.CalledProcessError as e:
-            msg = (
-                "Failed to create virtual environment. Note that on some linux systems, you may need to install the `python3-venv`"
-                " package via your system package manager."
-            )
-            print_command_failure(msg, e)
-            sys.exit(2)
+            if "ensurepip" in e.cmd:
+                print_error("fail")
+                print_error(
+                    "stubtest requires a Python installation with ensurepip. "
+                    "If on Linux, you may need to install the python3-venv package."
+                )
+            raise
 
         if sys.platform == "win32":
             pip = venv_dir / "Scripts" / "pip.exe"


tests/stubtest_third_party.py
```diff
@@ -43,7 +43,17 @@ def run_stubtest(dist: Path, *, verbose: bool = False, specified_stubs_only: boo
...
+        try:
+            venv.create(venv_dir, with_pip=True, clear=True)
+        except subprocess.CalledProcessError as e:
+            print_command_failure(
+                (
+                    "Failed to create virtual environment. Note that on some linux systems, you need to install the `python3-venv`"
+                    " package via your system package manager."
+                ),
+                e,
+            )
```
AlexWaygood, line 55:
```suggestion
            msg = (
                "Failed to create virtual environment. Note that on some linux systems, you need to install the `python3-venv`"
                " package via your system package manager."
            )
            print_command_failure(msg, e)
```

AlexWaygood:
It's weird to me that `venv` imports fine, but then actually trying to use it crashes with a bad error message. But I'll trust that you know Debian better than I do.

hamdanal:
It is because the command that fails is calling `ensurepip` (see `subprocess.CalledProcessError: Command '['/tmp/venv_dir_108341Y4HR/bin/python3.8', '-Im', 'ensurepip', '--upgrade', '--default-pip']' returned non-zero exit status 1.` in the issue) and ensurepip is the unbundeled part

AlexWaygood:
> It is because the command that fails is calling `ensurepip` (see `subprocess.CalledProcessError: Command '['/tmp/venv_dir_108341Y4HR/bin/python3.8', '-Im', 'ensurepip', '--upgrade', '--default-pip']' returned non-zero exit status 1.` in the issue) and ensurepip is the unbundeled part

Oh I see, so the venv creation would succeed if we didn't pass `with_pip=True` to `venv.create()`? (That's not a solution, since we need `pip` in order to install the relevant dependencies into the venv, I'm just curious.)

hamdanal:
> Oh I see, so the venv creation would succeed if we didn't pass `with_pip=True` to `venv.create()`? (That's not a solution, since we need `pip` in order to install the relevant dependencies into the venv, I'm just curious.)

Yes exactly. It passes if you don't specify `with_pip=True`

tests/stubtest_third_party.py
```diff
@@ -43,7 +43,15 @@ def run_stubtest(dist: Path, *, verbose: bool = False, specified_stubs_only: boo
...
-        venv.create(venv_dir, with_pip=True, clear=True)
+        try:
+            venv.create(venv_dir, with_pip=True, clear=True)
+        except subprocess.CalledProcessError as e:
+            msg = (
+                "Failed to create virtual environment. Note that on some linux systems, you may need to install the `python3-venv`"
+                " package via your system package manager."
+            )
+            print_command_failure(msg, e)
+            return False
```
AlexWaygood, line 54: It might make mores sense to terminate early here -- if you're running stubtest on multiple packages, there's not much point continuing on to the next one if creating a `venv` failed for the first one

```suggestion
            sys.exit(2)
```

hamdanal: Good point, would `sys.exit(e.returncode)` be a better option in this case?

AlexWaygood: > Good point, would `sys.exit(e.returncode)` be a better option in this case?

I don't have a very strong opinion, but it looks like the returncode here is `1`. I think I'd rather go for `2` here, since `1` usually indicates for a test that "the test failed", and this isn't that -- this is "the test errored". But, again, no strong opinion. It shouldn't 0 though :p

AlexWaygood:
What does the output now look like on your machine, with the patch applied?

tests/stubtest_third_party.py
```diff
@@ -43,7 +43,15 @@ def run_stubtest(dist: Path, *, verbose: bool = False, specified_stubs_only: boo
 
     with tempfile.TemporaryDirectory() as tmp:
         venv_dir = Path(tmp)
-        venv.create(venv_dir, with_pip=True, clear=True)
+        try:
+            venv.create(venv_dir, with_pip=True, clear=True)
+        except subprocess.CalledProcessError as e:
```
AlexWaygood, line 48: Maybe we should only suppress the traceback if it's the known issue with `ensurepip`:

```suggestion
        except subprocess.CalledProcessError as e:
            if "ensurepip" not in e.cmd:
                raise
```

hamdanal:
> What does the output now look like on your machine, with the patch applied?

```
  File "/home/ali/workspace/dev/typeshed/tests/stubtest_third_party.py", line 205, in <module>
    main()
  File "/home/ali/workspace/dev/typeshed/tests/stubtest_third_party.py", line 198, in main
    if not run_stubtest(dist, verbose=args.verbose, specified_stubs_only=args.specified_stubs_only):
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ali/workspace/dev/typeshed/tests/stubtest_third_party.py", line 53, in run_stubtest
    print_command_failure(msg, e)
  File "/home/ali/workspace/dev/typeshed/tests/stubtest_third_party.py", line 166, in print_command_failure
    print_command_output(e)
  File "/home/ali/workspace/dev/typeshed/tests/stubtest_third_party.py", line 171, in print_command_output
    print(e.stderr.decode(), end="", file=sys.stderr)
          ^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'decode'
```
Good thing you asked :p.
Since the output of the subprocess call in `venv.create` is not captured, we cannot use `print_command_failure(msg, e)` to print the `subprocess.CalledProcessError`. How should we print the message in this case?

AlexWaygood:
(I'm AFK for a little bit, will look later)

hauntsaninja (commented):
Thanks for improving things here!

I have a preference to avoid using sys.exit inside run_stubtest if possible. Perhaps something like the following could work:
```
diff --git a/tests/stubtest_third_party.py b/tests/stubtest_third_party.py
index 61ba07e71..0eef14edc 100755
--- a/tests/stubtest_third_party.py
+++ b/tests/stubtest_third_party.py
@@ -184,11 +184,21 @@ def main() -> NoReturn:
         dists = [typeshed_dir / "stubs" / d for d in args.dists]
 
     result = 0
-    for i, dist in enumerate(dists):
-        if i % args.num_shards != args.shard_index:
-            continue
-        if not run_stubtest(dist, verbose=args.verbose, specified_stubs_only=args.specified_stubs_only):
-            result = 1
+    try:
+        for i, dist in enumerate(dists):
+            if i % args.num_shards != args.shard_index:
+                continue
+            if not run_stubtest(dist, verbose=args.verbose, specified_stubs_only=args.specified_stubs_only):
+                result = 1
+    except Exception as e:
+        try:
+            import ensurepip
+        except ImportError:
+            raise RuntimeError(
+                "stubtest requires a Python installation with ensurepip. "
+                "If on Linux, you may need to install the python3-venv package."
+            ) from e
+        raise
     sys.exit(result)
```

hamdanal:
Wouldn't this be too broad and may print the error message even for unrelated errors? If this happens it may become be confusing than before. Would something like this improve the situation:

```diff
index 61ba07e7..8b46ffe2 100755
--- a/tests/stubtest_third_party.py
+++ b/tests/stubtest_third_party.py
@@ -43,7 +43,16 @@ def run_stubtest(dist: Path, *, verbose: bool = False, specified_stubs_only: boo

     with tempfile.TemporaryDirectory() as tmp:
         venv_dir = Path(tmp)
-        venv.create(venv_dir, with_pip=True, clear=True)
+        try:
+            venv.create(venv_dir, with_pip=True, clear=True)
+        except subprocess.CalledProcessError as e:
+            if "ensurepip" in e.cmd:
+                msg = (
+                    "stubtest requires a Python installation with ensurepip. "
+                    "If on Linux, you may need to install the python3-venv package."
+                )
+                print(msg, file=sys.stderr)
+            raise

         if sys.platform == "win32":
             pip = venv_dir / "Scripts" / "pip.exe"
```
Here the message is printed in the correct place, only if the error was really caused by absence of `ensurepip`, no traceback is hidden and no `sys.exit` in `run_stubtest` (only the original error raised as before but with improved message).

TBH this would be the perfect use-case for python3.11's `Exception.add_note()`, but until then this seems reasonable.

That said if you don't like this solution, I'll happily use your patch.

AlexWaygood:
> Would something like this improve the situation:

I like the look of this! Happy to go with this if @hauntsaninja concurs.

(Would be great if you could double-check what the output on your machine is with this diff, as before, though.)

hamdanal:
> > Would something like this improve the situation:
> 
> I like the look of this! Happy to go with this if @hauntsaninja concurs.
> 
> (Would be great if you could double-check what the output on your machine is with this diff, as before, though.)

I get this now (sorry for the screenshot, but no colors otherwise)
Screenshot 2022-12-17 212921

with a slightly different diff
```diff
diff --git a/tests/stubtest_third_party.py b/tests/stubtest_third_party.py
index 61ba07e7..4479709c 100755
--- a/tests/stubtest_third_party.py
+++ b/tests/stubtest_third_party.py
@@ -43,7 +43,17 @@ def run_stubtest(dist: Path, *, verbose: bool = False, specified_stubs_only: boo

     with tempfile.TemporaryDirectory() as tmp:
         venv_dir = Path(tmp)
-        venv.create(venv_dir, with_pip=True, clear=True)
+        try:
+            venv.create(venv_dir, with_pip=True, clear=True)
+        except subprocess.CalledProcessError as e:
+            if "ensurepip" in e.cmd:
+                msg = (
+                    "stubtest requires a Python installation with ensurepip. "
+                    "If on Linux, you may need to install the python3-venv package."
+                )
+                print_error("fail")
+                print_error(msg)
+            raise

         if sys.platform == "win32":
             pip = venv_dir / "Scripts" / "pip.exe"
```
I added `print_error("fail")` because `six...` does not print immediately (it is buffered, need to flush stdout) and `print_error(msg)` because it makes the error stand out.

AlexWaygood:
> I get this now

That looks great to me!

> sorry for the screenshot

I think this is one of the rare "good uses of screenshot" — you're good!

hauntsaninja:
Diff looks good to me too! :-)

AlexWaygood (approved):
Thank you!

PR #9381, opened 2022-12-18
`python-dateutil`: Add `zoneinfo` submodule

### LINKED ISSUE #9377: python-dateutil: should import submodules relativedelta, parser...
quoted from a [pyright issue](https://github.com/microsoft/pyright/issues/3666) :  
"It looks like this might be a bug in the stub because the implementation actually does import (and re-export) relativedata via an `__all__` reference. (...) The `dateutil.__init__.pyi` stub would need to add statements like `from . import relativedelta` to indicate to type checkers that the submodule is automatically loaded by its parent module."

AlexWaygood:


AlexWaygood:
Sadly this doesn't fix #9377 (see comments in the issue), but the addition of the `zoneinfo` subpackage is probably nice to have anyway.

JelleZijlstra (approved):

PR #9382, opened 2022-12-18
`regr_test.py`: Allow non-types dependencies

AlexWaygood:
Work towards #5768

This PR reworks `tests/regr_test.py` so that the script will still succeed if test cases are added for a stubs package that has non-types dependencies. In the process, it undertakes a lot of the refactoring that will be needed to allow non-types dependencies in `mypy_test.py`.

The approach is very similar to the approach we take in `stubtest_third_party.py`: dynamically create virtual environments on the fly in order to test stubs in isolation. Unlike `stubtest_third_party.py`, however, we only create the venv if it's a package that has non-types dependencies listed. Creating a venv for each package with test cases, even if it's a package that didn't have non-types dependencies, would make the script unacceptably slow.

Cc. @Avasam

### DIFF
--- a/tests/regr_test.py
+++ b/tests/regr_test.py
@@ -119,7 +119,7 @@ def run_testcases(
             pip_exe, python_exe = make_venv(tmpdir_path / ".venv")
             pip_command = [pip_exe, "install", get_mypy_req(), *requirements.external_pkgs]
             try:
-                subprocess.run(pip_command, check=True, capture_output=True)
+                subprocess.run(pip_command, check=True, capture_output=True, text=True)
             except subprocess.CalledProcessError as e:
                 print(e.stderr)
                 raise
--- a/tests/regr_test.py
+++ b/tests/regr_test.py
@@ -204,7 +204,7 @@ def run_testcases(
 
     mypy_command = [python_exe, "-m", "mypy"] + flags
     if verbosity is Verbosity.VERBOSE:
-        verbose_log(f"\n{mypy_command=}")
+        verbose_log(f"{mypy_command=}")
         if "MYPYPATH" in env_vars:
             verbose_log(f"{env_vars['MYPYPATH']=}")
         else:
--- a/tests/regr_test.py
+++ b/tests/regr_test.py
@@ -112,13 +112,14 @@ def run_testcases(
         # mypy refuses to consider a directory a "valid typeshed directory"
         # unless there's a stubs/mypy-extensions path inside it,
         # so add that to the list of stubs to copy over to the new directory
-        for requirement in set(requirements.typeshed_pkgs) | {package.name, "mypy-extensions"}:
+        for requirement in {package.name, *requirements.typeshed_pkgs, "mypy-extensions"}:
             shutil.copytree(Path("stubs", requirement), new_typeshed / "stubs" / requirement)
 
         if requirements.external_pkgs:
             pip_exe, python_exe = make_venv(tmpdir_path / ".venv")
+            pip_command = [pip_exe, "install", get_mypy_req(), *requirements.external_pkgs]
             try:
-                subprocess.run([pip_exe, "install", get_mypy_req(), *requirements.external_pkgs], check=True, capture_output=True)
+                subprocess.run(pip_command, check=True, capture_output=True)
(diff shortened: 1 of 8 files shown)


AlexWaygood:
- The CI run on commit c60c379 shows what happens if we add non-types dependencies to stubs packages that have test cases
- The CI run on commit 9cb0b22 shows what happens if we _use_ a non-types dependency in a test case
- The CI run on commit 4800d60 shows that the test cases still fail, as expected, if there's something bad in them (i.e., this refactoring hasn't broken anything).

tests/utils.py
```diff
@@ -45,25 +49,80 @@ def print_success_msg() -> None:
...
-    for dependency in requires:
+    dependencies = data.get("requires", [])
+    assert isinstance(dependencies, list)
+    typeshed, external = [], []
+    for dependency in dependencies:
         assert isinstance(dependency, str)
-        assert dependency.startswith("types-"), f"unrecognized dependency {dependency!r}"
-        dependencies.append(dependency[6:].split("<")[0])
-    return tuple(dependencies)
+        if dependency.startswith("types-"):
```
hauntsaninja, line 66: I don't see a problem here, but there are two cases where this logic may do something slightly surprising / different from stub_uploader which will consider something a typeshed stub iff it's something it's uploaded. So maybe worth adding a comment:
> This function may consider things to be typeshed stubs even if they haven't been uploaded to PyPI. If a typeshed stub is removed, this function will consider it to be an external dependency.

AlexWaygood: Your comment made me realise that there is also an edge case that this function doesn't currently handle: stubs which are uploaded with nonstandard names to PyPI. It'll break, for example, if we ever get a stubs package that depends on `types-pika-ts`.

This is unlikely to come up in real life, but I may as well fix it.

AlexWaygood: How does 42f722f look?

AlexWaygood:
I realised that if `types-requests` started depending on `urllib3` instead of `types-urllib3` (for example), we'd end up creating 15 different venvs and `pip install`ing `urllib3` 15 times, just to run the test cases for `types-requests`. e948601 significantly optimizes for that. We now only create one tempdir for each test-case directory; the full python-version/platform matrix is now run inside the same directory.

9823065 adds much better logging configurability. You can now run `--verbosity VERBOSE` to see exactly when tempdirs and venvs are being created, when things are being `pip install`ed, exactly what subprocess commands are being run, etc.

tests/regr_test.py
```diff
@@ -103,67 +170,71 @@ def test_testcase_directory(package: PackageInfo, version: str, platform: str, q
         "--no-error-summary",
         "--platform",
         platform,
-        "--no-site-packages",
         "--strict",
         "--pretty",
+        "--no-incremental",
```
AlexWaygood, line 175: I'm not sure whether this is necessary or not. In theory it might make the various mypy runs more independent. But we haven't had an actual problem with the mypy cache so far, so this might just make mypy slower unnecessarily.

@hauntsaninja, do you have any opinions here?

hauntsaninja: Unless we have a concrete reason, I think no need for `--no-incremental`

AlexWaygood:
```suggestion
```

AlexWaygood:
I'm planning on merging this in a few days, if there are no objections from anybody, as I'd like to move on to working on `mypy_test.py` :)

tests/regr_test.py
```diff
@@ -103,67 +170,70 @@ def test_testcase_directory(package: PackageInfo, version: str, platform: str, q
...
+    if verbosity > Verbosity.QUIET:
+        print(msg, end=" ", flush=True)
+
+    result = run_testcases(package=package, version=version, platform=platform, tempdir=tempdir, verbosity=verbosity)
 
     if result.returncode:
-        if quiet:
+        if verbosity > Verbosity.QUIET:
             # We'll already have printed this if --quiet wasn't passed.
             # If--quiet was passed, only print this if there were errors.
```
JelleZijlstra, line 227:
```suggestion
            # If --quiet was passed, only print this if there were errors.
```

JelleZijlstra (approved):

PR #9385, opened 2022-12-19
Annotate more items in pika.channel

srittau:
Based mostly on the docstrings, with occasional peeks into the implementation and types-pika.


AlexWaygood (approved):
Looks good, just one thing I noticed

PR #9387, opened 2022-12-19
fpdf:annotation of dash_pattern, line_width, text_mode

AmberAnsari89:


stubs/fpdf2/fpdf/graphics_state.pyi
```diff
@@ -52,17 +52,17 @@ class GraphicsStateMixin:
...
+    def dash_pattern(self, v: dict[str, float]) -> None: ...
     @property
-    def line_width(self): ...
+    def line_width(self) -> float: ...
     @line_width.setter
-    def line_width(self, v) -> None: ...
+    def line_width(self, v: float) -> None: ...
     @property
-    def text_mode(self): ...
+    def text_mode(self) -> int: ...
```
AlexWaygood, line 63: (You'll need to add `from .enums import TextMode` at the top of the file):

```suggestion
    def text_mode(self) -> TextMode: ...
```

It's set to `TextMode.FILL` here and, because it's an enum, `TextMode.FILL` is an instance of `TextMode`: https://github.com/PyFPDF/fpdf2/blob/dcb87ea11709f40effbe26fe3746e48735b6f708/fpdf/fpdf.py#L330

The setter method in `graphics_state.py` also auto-converts things to `TextMode` objects when the attribute is set: https://github.com/PyFPDF/fpdf2/blob/dcb87ea11709f40effbe26fe3746e48735b6f708/fpdf/graphics_state.py#L169

AmberAnsari89: > from .enums import TextMode

i did not receive email if your feedback. Is something wrong here?

AlexWaygood (approved):

AmberAnsari89:
is int | str fine for TextMode setter?

AmberAnsari89:
Merry Christmas to you .

AlexWaygood:
> is int | str fine for TextMode setter?

Yes, because the setter calls `TextMode.coerce()`, which accepts `str | int`, but ensures that the attribute is transformed into an instance of `TextMode` before being stored.

Merry Christmas to you too!

AmberAnsari89:
> > is int | str fine for TextMode setter?
> 
> Yes, because the setter calls `TextMode.coerce()`, which accepts `str | int`, but ensures that the attribute is transformed into an instance of `TextMode` before being stored.
> 
> Merry Christmas to you


> > is int | str fine for TextMode setter?
> 
> Yes, because the setter calls `TextMode.coerce()`, which accepts `str | int`, but ensures that the attribute is transformed into an instance of `TextMode` before being stored.
> 
> Merry Christmas to you too!

@text_mode.setter
    def text_mode(self, v):
        self.__statestack[-1]["text_mode"] = TextMode.coerce(v)

here, coerce() ensures the transformation of that attribute to TextMode.  right?

AlexWaygood:
Precisely :)

PR #9389, opened 2022-12-20
Specific types for requests auth username and password

mauvilsa:
I noticed that `username` and `password` in requests auth are expected to be `str` or `bytes`, see [requests/auth.py#L35-L53](https://github.com/psf/requests/blob/ac3be98b19f4d09c6a970b271a3ae30f3d0858f7/requests/auth.py#L35-L53). So thought that in typeshed these types could be made more specific.


JelleZijlstra (approved):
Technically other types will also work; requests will just emit a DeprecationWarning and `str()` the value. But it makes sense to be stricter in the stubs, since the `str()` behavior is probably rarely useful.

PR #9394, opened 2022-12-21
Fix return type of `search` in jmespath stubs

### LINKED ISSUE #9393: Incorrect return type for jmespath.search()
After this PR last week, the return type of `jmespath.search()` is listed as an internal data structure named `_TreeNode`:

https://github.com/python/typeshed/pull/9338

However I don't think this is accurate. The return type is supposed to be `Any`, like it was before the changes in that PR. `jmespath.search()` returns a value based on the input, which could be anything at all.

mondeja:
Fixes #9393


AlexWaygood (approved):
Thanks!

mwgamble:
Thanks for looking at this so quickly :heart:

PR #9395, opened 2022-12-21
set session.headers to MutableMapping

### LINKED ISSUE #9390: requests.session().headers
According to [this](https://github.com/psf/requests/issues/5020#issuecomment-989082461) comment in the requests project: 

>We deliberately allow any mapping type to allow for a variety of use-cases. A change here won't be accepted

I think this means that the strict typing shown here is not correct:

https://github.com/python/typeshed/blob/main/stubs/requests/requests/sessions.pyi#L116

And instead something that satisfies a mutableMapping would be more correct?

The X to my Y problem is that I am very much used to working with `.headers` set to a built-in dict. And if it works and is supported by `requests` ,  could I submit a patch to the type stubs?  Satisfying this typing constraint as it is right now feels incorrect given the long history of the dict() being sufficient.

evilensky:
Resolves #9390.

The session.headers may be set to any mapping according to the project, and mutable mapping appears to be the most correct annotation.

### DIFF
--- a/stubs/requests/requests/sessions.pyi
+++ b/stubs/requests/requests/sessions.pyi
@@ -116,7 +116,7 @@ class Session(SessionRedirectMixin):
     # See https://github.com/psf/requests/issues/5020#issuecomment-989082461:
     # requests sets this as a CaseInsensitiveDict, but users may set it to any MutableMapping
     # or None.
-    headers: MutableMapping[str, str | bytes] | None
+    headers: MutableMapping[str, str | bytes]
     auth: _Auth | None
     proxies: _TextMapping
     # Don't complain if:
--- a/stubs/requests/requests/sessions.pyi
+++ b/stubs/requests/requests/sessions.pyi
@@ -115,7 +115,6 @@ class Session(SessionRedirectMixin):
     __attrs__: Any
     # See https://github.com/psf/requests/issues/5020#issuecomment-989082461:
     # requests sets this as a CaseInsensitiveDict, but users may set it to any MutableMapping
-    # or None.
     headers: MutableMapping[str, str | bytes]
     auth: _Auth | None
     proxies: _TextMapping
--- a/stubs/requests/requests/sessions.pyi
+++ b/stubs/requests/requests/sessions.pyi
@@ -113,7 +113,7 @@ class _Settings(TypedDict):
 
 class Session(SessionRedirectMixin):
     __attrs__: Any
-    headers: CaseInsensitiveDict[str | bytes]
+    headers: MutableMapping[str, str | bytes | None]
     auth: _Auth | None
     proxies: _TextMapping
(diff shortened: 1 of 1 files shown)


stubs/requests/requests/sessions.pyi
```diff
@@ -113,7 +113,7 @@ class _Settings(TypedDict):
 
 class Session(SessionRedirectMixin):
     __attrs__: Any
-    headers: CaseInsensitiveDict[str | bytes]
+    headers: MutableMapping[str, str | bytes | None]
```
AlexWaygood, line 116:
```suggestion
    # See https://github.com/psf/requests/issues/5020#issuecomment-989082461:
    # requests sets this as a CaseInsensitiveDict, but users may set it to any MutableMapping
    headers: MutableMapping[str, str | bytes | None]
```

stubs/requests/requests/sessions.pyi
```diff
@@ -113,7 +113,9 @@ class _Settings(TypedDict):
 
 class Session(SessionRedirectMixin):
     __attrs__: Any
-    headers: CaseInsensitiveDict[str | bytes]
+    # See https://github.com/psf/requests/issues/5020#issuecomment-989082461:
+    # requests sets this as a CaseInsensitiveDict, but users may set it to any MutableMapping
+    headers: MutableMapping[str, str | bytes | None]
```
AlexWaygood, line 118: Are you sure this should be `MutableMapping[str, str | bytes | None]` rather than `MutableMapping[str, str | bytes]`? This annotation was debated somewhat extensively in https://github.com/python/typeshed/pull/7773#discussion_r864929231. @janrito said at the time:

> We want to allow str or bytes on actual headers, but None only on updates

evilensky: No, I am not sure. But given that this code does not produce any runtime error:

```Python
session = requests.Session()
session.headers = None
resp = session.get('http://example.com')
assert resp.status_code == 200
```

It appears headers get merged-or-set-to-defaults with None?   https://github.com/psf/requests/blob/main/requests/sessions.py#L61-L71 

I'm new to both typeshed and requests so if it's typeshed's duty to help developers "above and beyond" the runtime code it self I can see how disallowing None could be useful at type checking time.

AlexWaygood: From the mypy_primer output, looks like changing it to `MutableMapping[str, str | bytes] | None` would be quite disruptive -- so I vote for keeping this as `MutableMapping[str, str | bytes]` for now, unless keeping it that way is causing a concrete problem for you.

AlexWaygood (approved):
Thanks @evilensky! Will merge once the CI has finished.

`requests` is tricky to write stubs for, there's loads of unexpected things that crop up...

PR #9396, opened 2022-12-21
Update `json.dump` to only require `fp: SupportsWrite[str]`

willfrey:
Instead of requiring `fp` be an object of type `IO[str]`, this weakens the annotation to only require `fp` be of type `SupportsWrite[str]`.

This is all [the documentation requires](https://docs.python.org/3/library/json.html#json.dump) and is all [the implementation requires](https://github.com/python/cpython/blob/3e715e0cc811ebccce4234c3e25ef30787151d00/Lib/json/__init__.py#L180).

### DIFF
--- a/stdlib/json/__init__.pyi
+++ b/stdlib/json/__init__.pyi
@@ -1,6 +1,6 @@
-from _typeshed import SupportsRead
+from _typeshed import SupportsRead, SupportsWrite
 from collections.abc import Callable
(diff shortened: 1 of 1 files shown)


JelleZijlstra (approved):

srittau (approved):
Thanks!

willfrey:
You're welcome! Thanks for approving this so quickly!

PR #9397, opened 2022-12-21
Check for unused `pyright: ignore` and differentiate from mypy ignores

Avasam:
This is possible since pyright 1.1.283. I had to bump from 1.1.284 because of a regression with comments.
Prevent `type: ignore[*]` comments from also blanket suppressing all pyright errors.
Added more comments to pyright config.
Fixed new pyright errors


stdlib/builtins.pyi
```diff
@@ -1188,7 +1188,7 @@ class property:
 class _NotImplementedType(Any):  # type: ignore[misc]
     # A little weird, but typing the __call__ as NotImplemented makes the error message
```
JelleZijlstra, line 1189: Unrelated to your PR but we should probably revisit this and see if it actually makes things better for all type checkers.

stdlib/collections/__init__.pyi
```diff
@@ -328,15 +328,15 @@ class _OrderedDictValuesView(ValuesView[_VT_co], Reversible[_VT_co]):
 # (At runtime, these are called `odict_keys`, `odict_items` and `odict_values`,
 # but they are not exposed anywhere)
 @final
-class _odict_keys(dict_keys[_KT_co, _VT_co], Reversible[_KT_co]):  # type: ignore[misc]
+class _odict_keys(dict_keys[_KT_co, _VT_co], Reversible[_KT_co]):  # type: ignore[misc]  # pyright: ignore
```
AlexWaygood, line 331: Are we able to add the pyright error codes for these `pyright: ignore`s?

stubs/redis/redis/client.pyi
```diff
@@ -568,7 +567,7 @@ class Pipeline(Redis[_StrType], Generic[_StrType]):
     def sscan_iter(self, name: _Key, match: _Key | None = ..., count: int | None = ...) -> Iterator[Any]: ...
     def hscan(self, name: _Key, cursor: int = ..., match: _Key | None = ..., count: int | None = ...) -> Pipeline[_StrType]: ...  # type: ignore[override]
     def hscan_iter(self, name, match: _Key | None = ..., count: int | None = ...) -> Iterator[Any]: ...
-    def zscan(self, name: _Key, cursor: int = ..., match: _Key | None = ..., count: int | None = ..., score_cast_func: Callable[[_StrType], _ScoreCastFuncReturn] = ...) -> Pipeline[_StrType]: ...  # type: ignore[override]
+    def zscan(self, name: _Key, cursor: int = ..., match: _Key | None = ..., count: int | None = ..., score_cast_func: Callable[[_StrType], Any] = ...) -> Pipeline[_StrType]: ...  # type: ignore[override]
```
AlexWaygood, line 570: Looks like the stubs are a little out of date here -- it looks like at runtime, this method is now just inherited from the superclass. I guess there was some time in the past when that wasn't true, but I think we can safely just delete this method now:

```suggestion
```

stubs/pywin32/win32comext/axdebug/documents.pyi
```diff
@@ -14,7 +14,7 @@ class DebugDocumentProvider(gateways.DebugDocumentProvider):
     def GetDocument(self): ...
 
 # error: Cannot determine consistent method resolution order (MRO) for "DebugDocumentText"
-class DebugDocumentText(gateways.DebugDocumentInfo, gateways.DebugDocumentText, gateways.DebugDocument):  # type: ignore[misc]
+class DebugDocumentText(gateways.DebugDocumentInfo, gateways.DebugDocumentText, gateways.DebugDocument):  # type: ignore[misc]  # pyright: ignore
```
AlexWaygood, line 17: Which pyright error are we ignoring here? Can we add the error code?

AlexWaygood (commented):
This looks great to me. You've uncovered several cases where it looks like we were accidentally ignoring several pyright errors that we didn't really want to, and I think that shows the value of the change.

I left a few comments below. As well as those: maybe we should add a note to `CONTRIBUTING.md` somewhere. It might be surprising for new contributors to find out that `type: ignore` comments don't silence errors emitted by pyright. It might go well in [the "Conventions" section](https://github.com/python/typeshed/blob/main/CONTRIBUTING.md#conventions), where we already talk about using mypy error codes for `type: ignore` comments wherever possible.

stdlib/collections/__init__.pyi
Avasam: There is no code!
> Base class "dict_keys[_KT_co@_odict_keys, _VT_co@_odict_keys]" is marked final and cannot be subclassed `Pylance`

> E:\Users\Avasam\Documents\Git\typeshed\stdlib\collections\__init__.pyi
    E:\Users\Avasam\Documents\Git\typeshed\stdlib\collections\__init__.pyi:331:19 - error: Base class "dict_keys[_KT_co@_odict_keys, _VT_co@_odict_keys]" is marked final and cannot be subclassed

stubs/pywin32/win32comext/axdebug/documents.pyi
Avasam: Same as above, no code, and if iirc, this does fail at runtime
> Cannot create consistent method ordering `Pylance`

I could create a feature request with pyright to add a code for "will fail at runtime". But this should rather be fixed upstream (and pywin32 already has accepted a couple fixes I found through type-checking).

And for the other case above, it's kindof a niche type-stub use-case, and is totally fixable by splitting off into another private/internal protocol (which I didn't do to keep changes minimal)

stubs/redis/redis/client.pyi
Avasam: I have not taken a look at source (just figured it's inferred as Any anyway and that's enough to fix the issue). I'll go with your suggestion.

stdlib/collections/__init__.pyi
AlexWaygood: Oh, strange! In that case, can we leave a comment to that effect? (Same below for the other one)

```suggestion
# pyright doesn't have a specific error code for this error!
class _odict_keys(dict_keys[_KT_co, _VT_co], Reversible[_KT_co]):  # type: ignore[misc]  # pyright: ignore
```

Avasam:
[`f1264a1` (#9397)](https://github.com/python/typeshed/pull/9397/commits/f1264a17bd71ad6c1d58d22c97ade4728ae7d135) / https://github.com/python/typeshed/actions/runs/3789522708 Weird stubtest edge-case when ran under Python 3.7 . At least it was an easy fix.

AlexWaygood (approved):
Great stuff, thank you!

PR #9398, opened 2022-12-21
initial apriltag stubs

johanvandegriff:
https://pypi.org/project/apriltag/

Stubs for the "apriltag" library. I'm new to the contribution process so let me know if I missed something.


johanvandegriff:
I'm confused about the failing runs. I removed the numpy import but [this job](https://github.com/python/typeshed/actions/runs/3753150257/jobs/6376121210) is still saying it can't find numpy (line 53 of the log). Also [this job](https://github.com/python/typeshed/actions/runs/3753150191/jobs/6376121124) is complaining about `Use "list[T]" instead of "typing.List[T]" (PEP 585 syntax)` but I don't use `typing.List[T]` anywhere.

hauntsaninja:
The issue is that the package you're trying to add stubs for is a little broken. It has a runtime dependency on numpy, but doesn't declare that dependency in e.g. its setup.py

I also can't find a canonical source for the PyPI package you posted, so not sure how you'd get that fixed.

https://github.com/swatbotics/apriltag has some version of the code, but says 
> Please note: I am not the maintainer of the pypi package listed at https://pypi.org/project/apriltag/ – GitHub issues filed here reporting problems with that package will be summarily closed. Sorry, I don't have time to support someone else's unofficial package.

johanvandegriff:
Thanks for your quick reply!

Since there was no repo, I just downloaded the source code from https://pypi.org/project/apriltag/#files

If there's no way to get this to work, should I close the pull request?

AlexWaygood:
I agree with @hauntsaninja -- it seems like the project you're adding stubs for here is abandoned and somewhat broken :(

The best course of action for you seems to be @hauntsaninja's suggestion:

> If I were you, I'd consider forking the package and adding types to it directly

Good luck!

PR #9399, opened 2022-12-22
Add constructor to CellType

zhykzhykzhyk:


JelleZijlstra (approved):
Confirmed that this works in 3.8 but not in 3.7, where it is impossible to create a Cell from Python.

PR #9402, opened 2022-12-23
Add initial stubs for untangle

richardnias:
This adds basic stubs for [untangle](https://pypi.org/project/untangle/). I started with autogenerated stubs and edited from there.


stubs/untangle/untangle.pyi
```diff
@@ -0,0 +1,37 @@
+from _typeshed import Incomplete
+from collections.abc import Iterator
+from typing import Any
+from xml.sax import handler
+
+def is_string(x: Any) -> bool: ...
```
AlexWaygood, line 6: When it's safe to pass literally any object into a function, we should use `object` rather than `Any`. `Any` should generally be reserved for situations where something is inexpressable in the type system.

```suggestion
def is_string(x: object) -> bool: ...
```

stubs/untangle/untangle.pyi
```diff
@@ -0,0 +1,37 @@
...
+
+class Element:
+    children: Incomplete
+    is_root: bool
+    cdata: str
+    def __init__(self, name: str | None, attributes: dict[str, Any] | None) -> None: ...
```
AlexWaygood, line 12: Are you sure `attributes` can be `None`? It looks like `get_attribute()` will fail if `None` is passed in to that parameter: https://github.com/stchris/untangle/blob/e988ee2ce3cb296308ef0d1442c74bd4b17b3b1b/untangle.py#L65

Also: does it _have_ to be a `dict`, or could we use `Mapping` here? We generally prefer to be as permissive as possible for parameter annotations.

stubs/untangle/untangle.pyi
```diff
@@ -0,0 +1,37 @@
+from _typeshed import Incomplete
+from collections.abc import Iterator
+from typing import Any
+from xml.sax import handler
+
+def is_string(x: Any) -> bool: ...
+
+class Element:
+    children: Incomplete
```
AlexWaygood, line 9:
```suggestion
    children: list[Element]
```

stubs/untangle/untangle.pyi
```diff
@@ -0,0 +1,37 @@
...
+    is_root: bool
+    cdata: str
+    def __init__(self, name: str | None, attributes: dict[str, Any] | None) -> None: ...
+    def add_child(self, element: Element) -> None: ...
+    def add_cdata(self, cdata: str) -> None: ...
+    def get_attribute(self, key: str) -> Any: ...
```
AlexWaygood, line 15: We can get a _little_ bit more type safety here by using `Any | None`

```suggestion
    def get_attribute(self, key: str) -> Any | None: ...
```

stubs/untangle/untangle.pyi
```diff
@@ -0,0 +1,37 @@
...
+    def get_elements(self, name: Incomplete | None = ...) -> list[Element]: ...
```
AlexWaygood, line 16:
```suggestion
    def get_elements(self, name: str | None = ...) -> list[Element]: ...
```

stubs/untangle/untangle.pyi
```diff
@@ -0,0 +1,37 @@
...
+    def __getitem__(self, key: str) -> Any: ...
```
AlexWaygood, line 17:
```suggestion
    def __getitem__(self, key: str) -> Any | None: ...
```

stubs/untangle/untangle.pyi
```diff
@@ -0,0 +1,37 @@
...
+    def __getattr__(self, key: str) -> Element: ...
+    def __hasattribute__(self, name: str) -> bool: ...
+    def __iter__(self) -> Iterator[Element]: ...
```
AlexWaygood, line 20: If this method is called on a subclass of Element, it will in fact yield an instance of the subclass rather than an instance of `Element`. We can model this behaviour using a `TypeVar`. In typeshed, we prefer to use the same `TypeVar` every time for places where this happens: a special TypeVar that's called `Self`:

```suggestion
    def __iter__(self: Self) -> Iterator[Self]: ...
```

(You'll need to add `from _typeshed import Self` to the top of the file)

stubs/untangle/untangle.pyi
```diff
@@ -0,0 +1,37 @@
...
+    def __bool__(self) -> bool: ...
+    __nonzero__: Incomplete
```
AlexWaygood, line 22:
```suggestion
    __nonzero__ = __bool__
```

stubs/untangle/untangle.pyi
```diff
@@ -0,0 +1,37 @@
...
+    def __eq__(self, val: object) -> bool: ...
+    def __dir__(self) -> list[str]: ...
+    def __len__(self) -> int: ...
+    def __contains__(self, key: str) -> bool: ...
+
+class Handler(handler.ContentHandler):
+    root: Incomplete
```
AlexWaygood, line 29:
```suggestion
    root: Element
```

stubs/untangle/untangle.pyi
```diff
@@ -0,0 +1,37 @@
...
+    elements: Incomplete
```
AlexWaygood, line 30:
```suggestion
    elements: list[Incomplete]
```

stubs/untangle/untangle.pyi
```diff
@@ -0,0 +1,37 @@
...
+    def __init__(self) -> None: ...
+    def startElement(self, name: str, attributes: dict[str, Any]) -> None: ...
```
AlexWaygood, line 32: I think we could maybe use `Mapping` here instead of `dict`

stubs/untangle/untangle.pyi
```diff
@@ -0,0 +1,37 @@
...
+    def endElement(self, name: str) -> None: ...
+    def characters(self, cdata: str) -> None: ...
+
+def parse(filename: str, **parser_features: dict[str, bool]) -> Element: ...
```
AlexWaygood, line 36: `dict[str, bool]` here means that every keyword argument needs to be a dictionary with string keys and bool values. I think that's probably not what you meant ;)

```suggestion
def parse(filename: str, **parser_features: bool) -> Element: ...
```

AlexWaygood (commented):
Thanks! Looks pretty good, I've left a few suggestions:

stubs/untangle/untangle.pyi
AlexWaygood: I guess `attributes` _can_ be `None`, since the library literally does this in `Handler.__init__`: https://github.com/stchris/untangle/blob/e988ee2ce3cb296308ef0d1442c74bd4b17b3b1b/untangle.py#L143

Still seems weird to me though 🤷‍♂️

stubs/untangle/untangle.pyi
richardnias: ah, you're right, that one always catches me out...

richardnias:
Thanks for the suggestions/comments!

pyrightconfig.stricter.json
```diff
@@ -69,6 +69,7 @@
         "stubs/stripe",
         "stubs/tqdm",
         "stubs/ttkthemes",
+        "stubs/untangle",
```
AlexWaygood, line 72: Doesn't look like we actually need to exclude this package from the stricter pyright settings — it's already fully annotated!

```suggestion
```

stubs/untangle/untangle.pyi
```diff
@@ -0,0 +1,37 @@
...
+    children: list[Element]
+    is_root: bool
+    cdata: str
+    def __init__(self, name: str | None, attributes: Mapping[str, Any] | None) -> None: ...
+    def add_child(self, element: Element) -> None: ...
+    def add_cdata(self, cdata: str) -> None: ...
+    def get_attribute(self, key: str) -> Any | None: ...
+    def get_elements(self, name: str | None = ...) -> list[Element]: ...
+    def __getitem__(self, key: str) -> Any | None: ...
+    def __getattr__(self, key: str) -> Element: ...
```
AlexWaygood, line 18: Just so you're aware: type checkers have special handling for `__getattr__` methods. If they see one defined, they'll assume that something funky is going on, and will generally assume any arbitrary attribute access is okay. E.g. because__getattr__ is included in the stub here, they won't emit an error for the following:

```python
e = Element("foo", {})
e.non_existing_attribute
```

That might be desirable in this case — but sometimes it's useful to omit `__getattr__` methods from stubs, even if they exist at runtime, to avoid this behaviour. Your call.

richardnias: Good to know. In this case I think it's ok, since the library returns an object with arbitrary attributes based on the structure of the XML file being parsed. So from a type checking perspective I think arbitrary access _is_ ok. If you try and access an attribute that isn't there in the XML, the library raises an `AttributeError`, but it's entirely dependent on the XML being parsed.

AlexWaygood (approved):
Looks great -- thanks!

PR #9404, opened 2022-12-23
Update compileall stripdir type

hauntsaninja:
See https://github.com/python/cpython/pull/19883


AlexWaygood (approved):
Strictly speaking, I guess we should wait for 3.10.10 and 3.11.2, but I don't feel too strongly about it if you wanna merge this now :)

PR #9407, opened 2022-12-24
`ast.get_docstring` does not work with arbitrary `AST` nodes

sobolevn:
Source: https://github.com/python/cpython/blob/00afa5066bd45348ed82a38d3442763b2ed1a068/Lib/ast.py#L291-L292


AlexWaygood (approved):

AlexWaygood:
Thanks!

PR #9411, opened 2022-12-26
Add type stubs for gdb.Value comparison operators

gsingh93:


gsingh93:
cc: @pmderodat

gsingh93:
Regarding the test failures, I wonder if this is a valid place to use `# type: ignore[override]`?

pmderodat:
Hello, thank you for the addition! I would have came across the same issue: although it indeed sems to be a good idea to make `__eq__` and `__ne__` accept any type of argument, this is not what gdb does (it raises a `TypeError`). I suppose disabling the error with an annotation could make sense, but I don’t know about the ramifications: let’s hear from typeshed maintainers.

pmderodat:
Ok, thank you! FYI, I checked the `TypeError` both with a live GDB session:

```
(gdb) pi gdb.Value(True) == object()
TypeError: Could not convert Python object: <object object at 0x7fd5f6988b00>.
```

and in the source code (https://sourceware.org/git/?p=binutils-gdb.git;a=blob;f=gdb/python/py-value.c;h=44b4ebcf4e4d38d6b89fa925b7f1c7ffadd3a5a8;hb=refs/heads/master#l1623): the  `valpy_richcompare` function in `py-value.c` implements all comparison operators, and ultimately calls `convert_value_from_python` on the “other” argument, which raises `TypeError` when that isn’t a `bool`, `int`, `float`, (lazy) `string` or another `gdb.Value`.

AlexWaygood (approved):
Thanks! Haven't verified, but I trust this is correct

PR #9425, opened 2022-12-29
Optional `__whence` for stream protocols in codecs.pyi

willfrey:


JelleZijlstra:
Could you clarify why you need this? It makes the types more restrictive (since the protocol is now narrower, fewer types will be accepted). On a quick look I couldn't find where the codecs module uses the `seek` method.

willfrey:
I suppose since `codecs.StreamReader` and `codecs.StreamWriter` both have a seek with an optional `whence` argument, this is not a good change, as you pointed out.

For the sake of completeness, my thinking was that since the underlying `.stream` attribute of the `StreamReader` and `StreamWriter` classes are public, I may want to access the binary stream directly and call seek on that. Often I'm just seeking back to the start, so `stream = writer.stream; stream.seek(0)` and `mypy` isn't happy with me because it wants that second `whence` argument explicitly when I want the default behavior of `whence=io.SEEK_SET`.

Unless that thinking is compelling somehow (and I suspect it isn't!) feel free to close this.

And thank you again for the incredibly quick response!

willfrey:
I'm going to go ahead and close this.

PR #9429, opened 2022-12-30
Add stubs for influxdb-client

srittau:
Except for `client.util.date_utils`, the stubs were auto-generated and then manually fixed/cleaned up. This is best reviewed commit by commit, where the first commit is just the output of create_baseline_stubs.

Upstream annotations were cleaned up, but not reviewed. I'm sure that the many `dict`/`list` annotations could be made more lenient, but I didn't check review that.


srittau:
The homeassistant primer warning is genuine. The `kwargs` dict needs a proper type annotation in the source code.

stubs/influxdb-client/influxdb_client/client/util/date_utils.pyi
```diff
@@ -0,0 +1,14 @@
+from datetime import datetime, timedelta, tzinfo
+from threading import Lock
+
+date_helper: DateHelper | None
+lock_: Lock
+
+class DateHelper:
+    timezone: tzinfo
+    def __init__(self, timezone: tzinfo = ...) -> None: ...
+    def parse_date(self, date_string: str) -> datetime: ...
```
JelleZijlstra, line 10: This is a little weird in that the implementation in the class is just `pass`, but `get_date_helper()` monkeypatches the method onto the instances it returns. Annotating it as `-> datetime` seems like the right call but a comment would be helpful.

stubs/influxdb-client/influxdb_client/client/write_api.pyi
```diff
@@ -0,0 +1,103 @@
...
+        record: str
+        | Iterable[str]
+        | Point
+        | Iterable[Point]
+        | dict[Incomplete, Incomplete]
+        | Iterable[dict[Incomplete, Incomplete]]
+        | bytes
+        | Iterable[bytes]
+        | _Observable
+        | NamedTuple
```
JelleZijlstra, line 92: Mypy accepts `NamedTuple` as a type annotation but I'm not sure that's a feature we can rely on in typeshed.

JelleZijlstra (commented):
Looks good, two nits

stubs/influxdb-client/influxdb_client/client/write_api.pyi
srittau: I've used a type alias to `tuple[Any, ...]` for now, similar to what I did to dataclass.

PR #9431, opened 2022-12-31
Add stubs for netaddr

hamdanal:
https://pypi.org/project/netaddr/


stubs/netaddr/netaddr/core.pyi
```diff
@@ -0,0 +1,39 @@
...
+
+def num_bits(int_val: int) -> int: ...
+
+class Subscriber:
+    def update(self, data: Incomplete) -> None: ...
+
+class PrettyPrinter(Subscriber):
+    fh: TextIO
+    write_eol: bool
+    def __init__(self, fh: TextIO = ..., write_eol: bool = ...) -> None: ...
```
AlexWaygood, line 26: For IO types, we generally prefer to be as permissive as possible. In this case, it looks like `fh` doesn't _have_ to be a `TextIO` object; it can be anything with a `write()` method. We can use the `SupportsWrite` protocol from the `stdlib/_typeshed` directory to express this (you'll have to add `from _typeshed import SupportsWrite` at the top of the file):

```suggestion
    fh: SupportsWrite[str]
    write_eol: bool
    def __init__(self, fh: SupportsWrite[str] = ..., write_eol: bool = ...) -> None: ...
```

stubs/netaddr/netaddr/eui/__init__.pyi
```diff
@@ -0,0 +1,81 @@
...
+
+class BaseIdentifier:
+    def __init__(self) -> None: ...
+    def __int__(self) -> int: ...
+    def __long__(self) -> int: ...
+    def __oct__(self) -> str: ...
+    def __hex__(self) -> str: ...
+    def __index__(self) -> int: ...
+    def __eq__(self, other: object) -> bool: ...
+    def __ne__(self, other: object) -> bool: ...
```
AlexWaygood, line 18: Where are these defined? I can't see them in the source code

stubs/netaddr/netaddr/eui/__init__.pyi
```diff
@@ -0,0 +1,81 @@
...
+
+class OUI(BaseIdentifier):
+    records: list[dict[str, object]]
+    def __init__(self, oui: str | int) -> None: ...
```
AlexWaygood, line 22: Missing `__hash__`

```suggestion
    def __init__(self, oui: str | int) -> None: ...
    def __hash__(self) -> int: ...
```

stubs/netaddr/netaddr/eui/__init__.pyi
```diff
@@ -0,0 +1,81 @@
...
+    @property
+    def reg_count(self) -> int: ...
+    def registration(self, index: int = ...) -> DictDotLookup: ...
+
+class IAB(BaseIdentifier):
+    IAB_EUI_VALUES: tuple[int, int]
```
AlexWaygood, line 28: (You'll need to add `from typing import ClassVar` at the top of the file)

```suggestion
    IAB_EUI_VALUES: ClassVar[tuple[int, int]]
```

AlexWaygood (commented):
A partial review. I haven't read past `netaddr/eui/__init__.pyi` yet.

stubs/netaddr/netaddr/ip/iana.pyi
```diff
@@ -0,0 +1,45 @@
...
+    xmlparser: XMLReader
+    fh: Incomplete
+    def __init__(self, fh: Incomplete, **kwargs: object) -> None: ...
+    def process_record(self, rec: Mapping[str, object]) -> dict[str, str] | None: ...
+    def consume_record(self, rec: object) -> None: ...
+    def parse(self) -> None: ...
+
+class IPv4Parser(XMLRecordParser): ...
+class IPv6Parser(XMLRecordParser): ...
+class IPv6UnicastParser(XMLRecordParser): ...
```
AlexWaygood, line 31: Unlike in the superclass, the `process_record()` method in these classes never returns `None`, so we should override it

```suggestion
class IPv4Parser(XMLRecordParser):
    def process_record(self, rec: Mapping[str, object]) -> dict[str, str]: ...

class IPv6Parser(XMLRecordParser):
    def process_record(self, rec: Mapping[str, object]) -> dict[str, str]: ...

class IPv6UnicastParser(XMLRecordParser):
    def process_record(self, rec: Mapping[str, object]) -> dict[str, str]: ...
```

stubs/netaddr/netaddr/ip/iana.pyi
```diff
@@ -0,0 +1,45 @@
...
+
+class MulticastParser(XMLRecordParser):
+    def normalise_addr(self, addr: str) -> str: ...
+
+class DictUpdater(Subscriber):
+    dct: dict[_IanaInfoKey, Incomplete]
```
AlexWaygood, line 37: The docstring says "lookup dict or dict like object". It needs to support item assignment, but I think any `MutableMapping` will do.

```suggestion
    dct: MutableMapping[_IanaInfoKey, Incomplete]
```

stubs/netaddr/netaddr/ip/iana.pyi
```diff
@@ -0,0 +1,45 @@
...
+    topic: str
+    unique_key: str
+    def __init__(self, dct: dict[_IanaInfoKey, Incomplete], topic: str, unique_key: str) -> None: ...
```
AlexWaygood, line 40:
```suggestion
    def __init__(self, dct: MutableMapping[_IanaInfoKey, Incomplete], topic: str, unique_key: str) -> None: ...
```

stubs/netaddr/netaddr/ip/iana.pyi
```diff
@@ -0,0 +1,45 @@
...
+    def update(self, data: Incomplete) -> None: ...
+
+def load_info() -> None: ...
+def pprint_info(fh: TextIO | None = ...) -> None: ...
```
AlexWaygood, line 44:
```suggestion
def pprint_info(fh: SupportsWrite[str] | None = ...) -> None: ...
```

(with `SupportsWrite` imported from `_typeshed`)

stubs/netaddr/netaddr/ip/nmap.pyi
```diff
@@ -0,0 +1,6 @@
+from collections.abc import Generator
+
+from netaddr.ip import IPAddress
+
+def valid_nmap_range(target_spec: str) -> bool: ...
+def iter_nmap_range(*nmap_target_spec: str) -> Generator[IPAddress, None, None]: ...
```
AlexWaygood, line 6:
```suggestion
def iter_nmap_range(*nmap_target_spec: str) -> Iterator[IPAddress]: ...
```

stubs/netaddr/netaddr/ip/sets.pyi
```diff
@@ -0,0 +1,47 @@
...
+from collections.abc import Generator, Iterable, Iterator
+from typing import NoReturn
+from typing_extensions import TypeAlias
+
+from netaddr.ip import IPAddress, IPNetwork, IPRange, _IPNetworkAddr
+
+_IPIterable: TypeAlias = IPNetwork | IPRange | IPSet | Iterable[_IPNetworkAddr | IPRange | int]
+
+class IPSet:
+    _cidrs: dict[IPNetwork, bool]
```
AlexWaygood, line 11: We _generally_ omit private attributes in typeshed, unless there's a specific reason for including them. Is there a specific reason to make an exception here?

```suggestion
```

stubs/netaddr/netaddr/ip/sets.pyi
```diff
@@ -0,0 +1,47 @@
...
+    def symmetric_difference(self, other: IPSet) -> IPSet: ...
+    __xor__ = symmetric_difference
+    def difference(self, other: IPSet) -> IPSet: ...
+    __sub__ = difference
+    def __len__(self) -> int: ...
+    @property
+    def size(self) -> int: ...
+    def iscontiguous(self) -> bool: ...
+    def iprange(self) -> IPRange | None: ...
+    def iter_ipranges(self) -> Generator[IPRange, None, None]: ...
```
AlexWaygood, line 47:
```suggestion
    def iter_ipranges(self) -> Iterator[IPRange]: ...
```

stubs/netaddr/netaddr/strategy/eui48.pyi
```diff
@@ -0,0 +1,42 @@
+from collections.abc import Iterable, Sequence
+from re import Pattern
+from typing_extensions import Literal
+
+AF_LINK: Literal[48]
+width: Literal[48]
+family: Literal[48]
+family_name: Literal["MAC"]
+version: Literal[48]
+max_int: Literal[281474976710655]
```
AlexWaygood, line 10: The precise value feels like something of an implementation detail here :)

```suggestion
max_int: int
```

stubs/netaddr/netaddr/strategy/eui48.pyi
```diff
@@ -0,0 +1,42 @@
...
+
+class mac_eui48:
+    word_size: int
+    num_words: int
+    max_word: int
+    word_sep: str
+    word_fmt: str
+    word_base: int
```
AlexWaygood, line 18: Should these all be `ClassVar`s?

stubs/netaddr/netaddr/strategy/eui64.pyi
```diff
@@ -0,0 +1,41 @@
+from collections.abc import Iterable, Sequence
+from re import Pattern
+from typing_extensions import Literal
+
+AF_EUI64: Literal[64]
+width: Literal[64]
+family: Literal[64]
+family_name: Literal["EUI-64"]
+version: Literal[64]
+max_int: Literal[18446744073709551615]
```
AlexWaygood, line 10:
```suggestion
max_int: int
```

stubs/netaddr/netaddr/strategy/eui64.pyi
```diff
@@ -0,0 +1,41 @@
...
+
+class eui64_base:
+    word_size: int
+    num_words: int
+    max_word: int
+    word_sep: str
+    word_fmt: str
+    word_base: int
```
AlexWaygood, line 18: Should these all be `ClassVar`s?

stubs/netaddr/netaddr/strategy/ipv6.pyi
```diff
@@ -0,0 +1,43 @@
...
+
+OPT_IMPORTS: bool
+width: Literal[128]
+word_size: Literal[16]
+word_sep: Literal[":"]
+family: Final = AF_INET6
+family_name: Literal["IPv6"]
+version: Literal[6]
+word_base: Literal[16]
+max_int: Literal[340282366920938463463374607431768211455]
```
AlexWaygood, line 14:
```suggestion
max_int: int
```

AlexWaygood (commented):
Here's the rest of my review -- a few more comments, but mostly fairly minor. This is a very high-quality PR -- thank you!

stubs/netaddr/netaddr/eui/__init__.pyi
hamdanal: These should be defined on the subclasses, I'll move them

stubs/netaddr/netaddr/eui/__init__.pyi
hamdanal: I'm afraid this hasn't made it into a release yet; only exists on github.

stubs/netaddr/netaddr/ip/glob.pyi
```diff
@@ -0,0 +1,15 @@
+from netaddr.ip import IPAddress, IPNetwork, IPRange, _IPAddressAddr, _IPNetworkAddr
+
+def valid_glob(ipglob: str) -> bool: ...
```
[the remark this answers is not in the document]
hamdanal: Would it be better types as a `TypeGuard`?

stubs/netaddr/netaddr/ip/sets.pyi
hamdanal: I type some core private attributes so that I get maximum help from vscode (pylance) to validate the types I set. I must've forgotten to remove this one. I'll remove it.

stubs/netaddr/netaddr/strategy/eui48.pyi
hamdanal: yes

stubs/netaddr/netaddr/eui/ieee.pyi
```diff
@@ -0,0 +1,33 @@
...
+class IABIndexParser(Publisher):
+    fh: BinaryIO
+    def __init__(self, ieee_file: BinaryIO | StrOrBytesPath | int) -> None: ...
+    def parse(self) -> None: ...
+
+def create_index_from_registry(
+    registry_fh: BinaryIO | StrOrBytesPath | int, index_path: StrOrBytesPath, parser: type[OUIIndexParser] | type[IABIndexParser]
+) -> None: ...
+def create_indices() -> None: ...
+def load_index(index: _INDEX, fp: BinaryIO) -> None: ...
```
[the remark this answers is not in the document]
hamdanal: I agree with the concept of permissive types. A note here though, the docstring says *"Load index from file into index data structure."*. doesn't this mean they expect a file and not any iterable of bytes? The implementation does accept any iterable of bytes `[x.decode('UTF-8') for x in fp]` but I supposed this has to do with some python2 compat.

stubs/netaddr/netaddr/ip/iana.pyi
```diff
@@ -0,0 +1,45 @@
...
+    def endElement(self, name: str) -> None: ...
+    def characters(self, content: str) -> None: ...
+
+class XMLRecordParser(Publisher):
+    xmlparser: XMLReader
+    fh: Incomplete
+    def __init__(self, fh: Incomplete, **kwargs: object) -> None: ...
+    def process_record(self, rec: Mapping[str, object]) -> dict[str, str] | None: ...
+    def consume_record(self, rec: object) -> None: ...
+    def parse(self) -> None: ...
```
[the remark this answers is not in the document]
hamdanal: I added `def __getattr__(self) -> Any: ...` similar to (all) other usages in typeshed. Let me know if you prefer `-> object` here

stubs/netaddr/netaddr/eui/__init__.pyi
AlexWaygood: Ah! That explains it :)

stubs/netaddr/netaddr/eui/ieee.pyi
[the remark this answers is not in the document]
AlexWaygood: Yep, I saw that comment in the source code too. The trouble is that the concept of a "file-like object" is really very poorly defined in Python. PEP 484 attempted to define "file-like" by adding the `IO` and `BinaryIO` interfaces to the `typing` module. But these interfaces are very hard to satisfy in real life, because they're `ABC`s with a large number of methods. In the wild, people often use custom file-like classes that don't inherit from any of the stdlib file-like classes, and so aren't recognised by type checkers as being subtypes of `BinaryIO`.

If we accept that there's no good way in the current type system of expressing "An object representing an opened file which, when iterated over, yields `bytes` objects" -- then we're left with having to decide these things on a case-by-case basis. We tend to prefer false negatives over false positives in typeshed, so here I'd go with `Iterable[bytes]`.

If it makes you feel better about things, you could define a type alias, and use that in the annotation instead?

```python
# Expected to be an object representing an opened file,
# but the only restrictions at runtime are that it has to be an Iterable of bytes
_BytesFile: TypeAlias = Iterable[bytes]
```

stubs/netaddr/netaddr/ip/glob.pyi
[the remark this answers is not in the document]
AlexWaygood: Oh, interesting idea. Sure, why not!

stubs/netaddr/netaddr/eui/ieee.pyi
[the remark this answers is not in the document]
hamdanal: That's a great explanation thank you. I'll commit the suggestion.

AlexWaygood (approved):
Thank you!

AlexWaygood:
Thank you for `rich-argparse` by the way -- I'm using it in one of my projects, and it's very nice :)

hamdanal:
Thanks Alex for all your contributions, helpful as always.

P.S: Today you get stability with rich-argparse, I just released 1.0. 🚀

PR #9434, opened 2022-12-31
Support external-dependencies in pyright

Avasam:
Extracted from #9374
Work towards #5768

Support external (non-type) dependencies in pyright in the CI.
This also adds a new reusable script that can be run from CLI to get a list of dependencies (similar to `get_packages.py`) and leverages the `read_dependencies` method added in #9382


.github/workflows/tests.yml
```diff
@@ -112,6 +112,17 @@ jobs:
       fail-fast: false
     steps:
       - uses: actions/checkout@v3
+      - uses: actions/setup-python@v4
+        with:
+          python-version: ${{ matrix.python-version }}
+          cache: pip
+          cache-dependency-path: requirements-tests.txt
+      - run: |
+          pip install tomli typing_extensions
```
JelleZijlstra:
```suggestion
          pip install pathspec tomli typing_extensions
```
Though maybe we should just rely on `requirements-tests.txt` instead?

.github/workflows/tests.yml
```diff
@@ -112,6 +112,17 @@ jobs:
       fail-fast: false
     steps:
       - uses: actions/checkout@v3
+      - uses: actions/setup-python@v4
+        with:
+          python-version: ${{ matrix.python-version }}
```
AlexWaygood, line 117: If we just use `"3.10"`, then we don't need the `cache`/`lru_cache` shim in `utils.py`. I don't think we need to use the Python version we're checking for this bit, do we?

```suggestion
          python-version: "3.10"

File: .github/workflows/tests.yml [hunk 3]
```diff
@@ -112,6 +112,17 @@ jobs:
...
+          cache: pip
+          cache-dependency-path: requirements-tests.txt
+      - run: |
+          pip install tomli pathspec packaging
```
AlexWaygood, line 121: As @JelleZijlstra says, it might be easier just to use `requirements-tests.txt` here. If the `pip install` step of our CI starts getting slow then we can rethink this approach, but at the moment I think it's fast enough, and it's less maintenance work to just rely on a single `requirements.txt` file. That way we don't need to update this file if we add a new third-party dependency to `tests/utils.py`; we'll only need to update `requirements-tests.txt`.

```suggestion
          pip install requirements-tests.txt
```

tests/get_external_dependencies.py
```diff
@@ -0,0 +1,16 @@
...
+from utils import read_dependencies
+
+distributions = sys.argv[1:]
+
+if not distributions:
+    distributions = os.listdir("stubs")
+
+for distribution in distributions:
+    for package in read_dependencies(distribution).external_pkgs:
+        print(f"{package} ")
```
AlexWaygood, line 16: Why the f-string here?

tests/get_packages.py
```diff
@@ -14,4 +14,4 @@
     for distribution in distributions:
         with open(f"stubs/{distribution}/METADATA.toml", "rb") as file:
             for package in tomli.load(file).get("tool", {}).get("stubtest", {}).get(METADATA_MAPPING[platform], []):
-                print(package)
+                print(f"{package} ")
```
AlexWaygood, line 17: Why is this change needed?

AlexWaygood (commented):
Thanks! A few nits:

.github/workflows/tests.yml
Avasam: I thought the same, but *maybe* it has an effect for the external-dependency version?

.github/workflows/tests.yml
Avasam: Installing a bunch of unrelated and unnecessary dependencies was indeed my concern. I was also curious what's the actual bare-minimum.

Come to worse we can also split it in another requirements file that is referenced by `requirements-tests.txt`.
Both of both world with minimal maintenance and installs.

.github/workflows/tests.yml
AlexWaygood: Oof. That makes me realise that if we have any packages that declare a dependency on `numpy` or `pandas`, that might cause issues for users of Python 3.7, since the latest versions of `numpy` and `pandas` only support Python 3.8+.

Still, I feel _inclined_ to say "Let's keep things simple for now, and just use 3.10 unconditionally". If some user comes to us and complains that the stubs package doesn't work properly on 3.7 or whatever, we can always rethink this :)

.github/workflows/tests.yml
AlexWaygood: > We could also split it in another requirements file that is referenced by `requirements-tests.txt`.
> Both of both world with minimal maintenance and installs.

Yup, I do this in `typeshed-stats`: https://github.com/AlexWaygood/typeshed-stats/blob/main/requirements/all.txt

It still feels like it's adding complexity in terms of cognitive load, though, and I'm still not _convinced_ it's necessary. The `pip install` step takes up quite a high percentage of the total CI time for typeshed-stats, so it was worth micro-optimising it! Here, I'm not so sure.

tests/get_external_dependencies.py
Avasam: For the space.
And the space was needed (at least locally) or it'd store something like `numpycomtypes` in the variable.
Might just be a weird windows bash issue though

AlexWaygood: Would this work?

```suggestion
        print(package, end=" ")
```

Avasam: yeah ok that's it, works fine in powershell, I'll change this to no space

tests/get_packages.py
Avasam: Windows-bash jank (same as above), reverted.

tests/get_external_dependencies.py
```diff
@@ -0,0 +1,15 @@
...
+
+from utils import read_dependencies
+
+distributions = sys.argv[1:]
+if not distributions:
+    distributions = os.listdir("stubs")
+
+for distribution in distributions:
+    for package in read_dependencies(distribution).external_pkgs:
+        print(package)
```
AlexWaygood, line 15: This will be problematic if we ever have packages that depend on, e.g., different versions of numpy. But that definitely feels like a hypothetical problem that we can tackle if and when it comes up -- this is good enough for now!

We _could_ consider doing some deduplication here, e.g.:

```suggestion
requirements: set[str] = set()

for distribution in distributions:
    requirements.update(read_dependencies(distribution).external_pkgs)

for requirement in sorted(requirements):
    print(requirement)
```

tests/utils.py
```diff
@@ -84,14 +86,15 @@ def read_dependencies(distribution: str) -> PackageDependencies:
...
+    external: set[str] = set()
     for dependency in dependencies:
         assert isinstance(dependency, str)
         maybe_typeshed_dependency = Requirement(dependency).name
         if maybe_typeshed_dependency in pypi_name_to_typeshed_name_mapping:
-            typeshed.append(pypi_name_to_typeshed_name_mapping[maybe_typeshed_dependency])
+            typeshed.add(pypi_name_to_typeshed_name_mapping[maybe_typeshed_dependency])
         else:
-            external.append(dependency)
+            external.add(dependency)
```
AlexWaygood, line 97: I'm not sure deduplication here is needed. We'll only have duplicate packages here if a package has something like `requires = ["numpy", "numpy"]` in its `METADATA.toml` file. And that seems pretty unlikely, no?

Avasam: Right, and `requires = ["numpy", "numpy"]` is exactly how I checked for duplication ^^" Makes sense.

I'll move the de-duplication to the right spot.

tests/get_external_dependencies.py
Avasam: >  e.g., different versions of numpy

Only if their version ranges are not compatible. And we're only affected by types, not implementation. So should be extremely rare imo, but yeah. I think the only alternative is a virtual environment. And I like this currently very simple solution 😄

tests/utils.py
```diff
@@ -7,16 +7,18 @@
 import subprocess
 import sys
 import venv
-from collections.abc import Mapping
+from collections.abc import Iterable, Mapping
 from functools import cache
 from pathlib import Path
-from typing import Iterable, NamedTuple
-from typing_extensions import Annotated
+from typing import TYPE_CHECKING, Iterable, NamedTuple
```
Avasam, line 13:
```suggestion
from typing import TYPE_CHECKING, NamedTuple
```

tests/utils.py
```diff
@@ -7,16 +7,18 @@
...
 from pathlib import Path
-from typing import Iterable, NamedTuple
-from typing_extensions import Annotated
+from typing import TYPE_CHECKING, NamedTuple
 
 import pathspec  # type: ignore[import]
 import tomli
 from packaging.requirements import Requirement
 
+if TYPE_CHECKING:
```
AlexWaygood, line 19: I feel like I prefer the original version here, where it wasn't guarded behind the `TYPE_CHECKING` flag. It's true that this avoids an import at runtime, but the performance benefit here is tiny, and I prefer the cleaner and simpler code :)

Having said that, I think `typing_extensions` is only available here because `mypy` depends on it, and we have `mypy` in our `requirements-tests.txt` file. We should probably add it to `requirements-tests.txt`, since implicit dependencies are generally a bad idea, and we also need the dependency for e.g. `stubtest_stdlib.py`. Want to make another PR adding it to `requirements-tests.txt`?

AlexWaygood:
Presumably `pyright_test.py` won't work unless contributors have already run `tests/get_external_dependencies.py` in their venv first. Maybe we should add a note to that effect in `tests/README.md` and/or the docstring for `pyright_test.py?

AlexWaygood:
Happy new year from the UK, by the way! 😀 🎆

tests/utils.py
Avasam: Oh yeah this should've been reverted too when installing from `requirements-tests.txt`.

Sure I'll do a PR to add the dep explicitely.

AlexWaygood (commented):
We're getting there! A few more comments below.

As well as these comments: I'm slightly concerned that the name of the new script (`get_external_dependencies.py`) is quite similar to the existing script `get_packages.py`. Can we disambiguate them somehow? Maybe we should rename `get_packages.py`, since that one has quite a vague name currently?

Avasam:
> [...] I'm slightly concerned that the name of the new script (`get_external_dependencies.py`) is quite similar to the existing script `get_packages.py`. Can we disambiguate them somehow? Maybe we should rename `get_packages.py`, since that one has quite a vague name currently?

How about something like `get_system_packages` and `get_external_[python]_requirements` ? Name is a bit long, but tab-completion is also a thing for CLI.

AlexWaygood:
> How about something like `get_system_packages` and `get_external_[python]_requirements` ? Name is a bit long, but tab-completion is also a thing for CLI.

Yeah, those sound great to me! 👍

AlexWaygood:
Actually, maybe something like `get_stubtest_system_requirements.py` and `get_external_stub_requirements.py`? That might clarify that the first script doesn't install requirements _of the stub package_, it just installs requirements _stubtest_ needs, in contrast to the second script.

tests/README.md
```diff
@@ -78,6 +78,15 @@ checks that would typically fail on incomplete stubs (such as `Unknown` checks).
 In typeshed's CI, pyright is run with these configuration settings on a subset of
 the stubs in typeshed (including the standard library).
 
+In order for `pyright_test` to work correctly, some third-party stubs may require dependencies external to typeshed to be installed in your virtual environment prior to running the test.
+You can list or install all of a stubs package's external dependencies using the following script:
+```bash
```
AlexWaygood, line 83: _Tiiiny_ nitpick: is `bash` a good syntax highlight to use here? It looks like it might be doing some funky things to the `<` and `>` characters in e.g. `<third_party_stub>`. Maybe no highlighting would be better here?

```suggestion
```
```

Comment (Avasam) [reply tests/README.md:83]: As far as https://github.com/github/linguist is concerned, `shell`, `sh`, `shell-script`, `bash` and `zsh` are all the same, and look fine.

bash
```bash
(.venv3)$ python tests/get_external_dependencies.py <third_party_stub>  # List external dependencies for <third_party_stub>
(.venv3)$ python tests/get_external_dependencies.py <third_party_stub1> <third_party_stub2>  # List external dependencies for <third_party_stub1> and <third_party_stub2>
(.venv3)$ python tests/get_external_dependencies.py  # List external dependencies for all non stubs packages in typeshed
(.venv3)$ pip install $(python tests/get_external_dependencies.py <third_party_stub>)  # Install non-types dependencies
```

shell
```shell
(.venv3)$ python tests/get_external_dependencies.py <third_party_stub>  # List external dependencies for <third_party_stub>
(.venv3)$ python tests/get_external_dependencies.py <third_party_stub1> <third_party_stub2>  # List external dependencies for <third_party_stub1> and <third_party_stub2>
(.venv3)$ python tests/get_external_dependencies.py  # List external dependencies for all non stubs packages in typeshed
(.venv3)$ pip install $(python tests/get_external_dependencies.py <third_party_stub>)  # Install non-types dependencies
```

powershell
```powershell
(.venv3)$ python tests/get_external_dependencies.py <third_party_stub>  # List external dependencies for <third_party_stub>
(.venv3)$ python tests/get_external_dependencies.py <third_party_stub1> <third_party_stub2>  # List external dependencies for <third_party_stub1> and <third_party_stub2>
(.venv3)$ python tests/get_external_dependencies.py  # List external dependencies for all non stubs packages in typeshed
(.venv3)$ pip install $(python tests/get_external_dependencies.py <third_party_stub>)  # Install non-types dependencies
```

empty
```
(.venv3)$ python tests/get_external_dependencies.py <third_party_stub>  # List external dependencies for <third_party_stub>
(.venv3)$ python tests/get_external_dependencies.py <third_party_stub1> <third_party_stub2>  # List external dependencies for <third_party_stub1> and <third_party_stub2>
(.venv3)$ python tests/get_external_dependencies.py  # List external dependencies for all non stubs packages in typeshed
(.venv3)$ pip install $(python tests/get_external_dependencies.py <third_party_stub>)  # Install non-types dependencies
```

It doesn't outright break the syntax, if anything it highlights that you need to replace this part. It also correctly handles comments and inline commands.

Editors may have different implementations. This is in VSCode with popular markdown syntax highlight (and github-like previewer) that I use:
image

This is with plugins disabled:
image

You can see here that "shell" isn't naturally supported in the preview. (code syntax highlight has no difference). But I wouldn't use that as an argument.
Of course the exact look will depend on the editor's implementation.
In any case, having **a** language specifier here only seems better than **no** language specifier.

Another reason to use bash specifically, from a purely semantic point of view, is that the doc seems to assume all commands are to be run on linux, where afaik bash (or bash-like) is most often the default. (and bash-like code usually works in powershell on Windows)

---

I would also like to add a language specifier for better syntax highlight in all relevant code snippets in the docs in a separate PR.

AlexWaygood: > I would also like to add a language specifier for better syntax highlight in all relevant code snippets in the docs in a separate PR.

That sounds like a great idea in general, by the way!

tests/README.md
```diff
@@ -78,6 +78,15 @@ checks that would typically fail on incomplete stubs (such as `Unknown` checks).
...
+(.venv3)$ python tests/get_external_dependencies.py <third_party_stub>  # List external dependencies for <third_party_stub>
+(.venv3)$ python tests/get_external_dependencies.py <third_party_stub1> <third_party_stub2>  # List external dependencies for <third_party_stub1> and <third_party_stub2>
+(.venv3)$ python tests/get_external_dependencies.py  # List external dependencies for all non stubs packages in typeshed
+(.venv3)$ pip install $(python tests/get_external_dependencies.py <third_party_stub>)  # Install non-types dependencies
```
Avasam, line 87:
```suggestion
# Install external dependencies for all non stubs packages in typeshed
(.venv3)$ DEPENDENCIES=$(python tests/get_external_dependencies.py)
(.venv3)$ if [ -n "$DEPENDENCIES" ]; then pip install $DEPENDENCIES; fi
```

tests/README.md
```diff
@@ -78,6 +78,17 @@ checks that would typically fail on incomplete stubs (such as `Unknown` checks).
 In typeshed's CI, pyright is run with these configuration settings on a subset of
 the stubs in typeshed (including the standard library).
 
+In order for `pyright_test` to work correctly, some third-party stubs may require dependencies external to typeshed to be installed in your virtual environment prior to running the test.
+You can list or install all of a stubs package's external dependencies using the following script:
+```bash
+(.venv3)$ python tests/get_external_dependencies.py <third_party_stub>  # List external dependencies for <third_party_stub>
+(.venv3)$ python tests/get_external_dependencies.py <third_party_stub1> <third_party_stub2>  # List external dependencies for <third_party_stub1> and <third_party_stub2>
+(.venv3)$ python tests/get_external_dependencies.py  # List external dependencies for all non stubs packages in typeshed
+# Install external dependencies for all non stubs packages in typeshed
```
AlexWaygood, line 87:
```suggestion
# Install external dependencies for all third-party stubs packages in typeshed
```

tests/README.md
```diff
@@ -78,6 +78,17 @@ checks that would typically fail on incomplete stubs (such as `Unknown` checks).
 In typeshed's CI, pyright is run with these configuration settings on a subset of
 the stubs in typeshed (including the standard library).
 
+In order for `pyright_test` to work correctly, some third-party stubs may require dependencies external to typeshed to be installed in your virtual environment prior to running the test.
+You can list or install all of a stubs package's external dependencies using the following script:
+```bash
+(.venv3)$ python tests/get_external_dependencies.py <third_party_stub>  # List external dependencies for <third_party_stub>
+(.venv3)$ python tests/get_external_dependencies.py <third_party_stub1> <third_party_stub2>  # List external dependencies for <third_party_stub1> and <third_party_stub2>
+(.venv3)$ python tests/get_external_dependencies.py  # List external dependencies for all non stubs packages in typeshed
```
AlexWaygood, line 86:
```suggestion
(.venv3)$ python tests/get_external_dependencies.py  # List external dependencies for all third-party packages in typeshed
```

AlexWaygood (approved):
LGTM, other than the issue about the names of the files themselves :)

Avasam:
> Actually, maybe something like `get_stubtest_system_requirements.py` and `get_external_stub_requirements.py`? That might clarify that the first script doesn't install requirements _of the stubs package_, it just installs requirements _stubtest_ needs, in contrast to the second script.
> 
> What do you think?

SGTM. I also fixed-up a comment.

AlexWaygood (approved):
Looks great! Will merge once CI passes. Thanks for working on this!

PR #9435, opened 2022-12-31
Accurate, deduplicated, `termcolor.colored` fallback

Avasam:
`utils.colored`'s signature now correctly reflects `termcolor.colored`
Attempt in `scripts/runtests.py` to deduplicate the same fallback (using utils module in an outside folder).
~~I think the solution is fine, but you tell me.~~ VSCode can pick-up the source just fine, ~~mypy might need an extra `mypy_path="tests"` to remove `# type: ignore[import]` at line 13.~~
Edit: Apparently it wasn't a great solution (see comments below)

### DIFF
--- a/scripts/runtests.py
+++ b/scripts/runtests.py
@@ -3,18 +3,14 @@ from __future__ import annotations
 
 import json
 import os
+import pathlib
 import re
 import subprocess
 import sys
 from pathlib import Path
 
-try:
-    from termcolor import colored
-except ImportError:
-
-    def colored(text: str, color: str = "") -> str:  # type: ignore[misc]
-        return text
-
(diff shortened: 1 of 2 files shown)


AlexWaygood:
Eeeeee, I have a strong aversion to mucking about with `sys.path` 😅 I've heard lots of horror stories about where that gets you...

Avasam:
> [...] I've heard lots of horror stories about where that gets you...

I'll trust your experience on that !
If there's no better solution, it's not that bad of a code duplication, so I can go back to using a fallback (but the one with proper signature)

AlexWaygood:
> If there's no better solution, it's not that bad of a code duplication, so I can go back to using a fallback (but the one with proper signature).

Yeah, that sounds fine to me for now :)

AlexWaygood (approved):

PR #9436, opened 2022-12-31
Offer sensible default VSCode settings

Avasam:
Offer sensible default VSCode settings.
These are the workspace settings, specific to typeshed, that I've been using and filling in over the past few months.
Workspace settings override user settings, so we can offer sensible, all-ready configurations for the typeshed workspace, reducing a lot of the pains a new contributor can have if their favorite editor is VSCode.

These settings obviously won't configure any extension not specified in `.vscode/extensions.json > recommendations`, and will assume the user may not have all recommended extensions installed or set as the formatter. Hence I've also configure some basic VSCode settings.

---

The recommended extensions include:
- The full python extensions suite needed for typeshed
- `.editorconfig` support and syntax highlighter
- A yaml formatter and validator
- A TOML syntax highliter and validator (no formatting)
- Mardown pack for syntax highlighting, and for the Mardown preview to match Github's (no formatter, though I'd recommend [davidanson.vscode-markdownlint](https://marketplace.visualstudio.com/items?itemName=DavidAnson.vscode-markdownlint))
For those not familiar with VSCode: Note that recommendations are not forced onto the user, a popup appears once to mention the workspace has recommended extensions, then they'll only show up under the "recommended" tab in the extensions menu.

---

It can still make sense for a user to want to modify `.vscode/settings.json` if they want to configure one of their extension to work properly with typeshed, without having to outright disable it for the workspace. It's also up to the dev to decide if they wanna follow these specific editor settings.

Unfortunately, VSCode doesn't (yet) offer any way to have "workspace defaults" or "user-worspace settings", so offering defaults to copy is the best we can do at the moment. And is what I've seen done in other open-source projects.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -3,3 +3,4 @@
 tests/stubtest_allowlists/*.txt linguist-language=ini
 tests/pytype_exclude_list.txt linguist-language=ini
 pyrightconfig*.json linguist-language=jsonc
+.vscode/*.json linguist-language=jsonc
--- a/.gitignore
+++ b/.gitignore
@@ -60,7 +60,9 @@ analyze.py
 # Editor backup files
 *~
 .*.sw?
-.vscode/
+.vscode/*
+!.vscode/settings.default.json
+!.vscode/extensions.json
 .idea/
 .venv*/
 
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,0 +1,33 @@
+// Keep in alphabetical order
+{
+  "recommendations": [
+    "bierner.github-markdown-preview",
+    "bungcip.better-toml",
+    "editorconfig.editorconfig",
+    "ms-python.black-formatter",
+    "ms-python.flake8",
+    "ms-python.isort",
+    "ms-python.python",
+    "ms-python.vscode-pylance",
+    "redhat.vscode-yaml",
+],
+"unwantedRecommendations": [
+    /*
+     * Don't recommend by default for this workspace
+    */
+    "christian-kohler.npm-intellisense",
+    /*
+     * Must disable in this workspace
+     * https://github.com/microsoft/vscode/issues/40239
+    */
+    // We use black
+    "ms-python.autopep8",
+    // Not using pylint
+    "ms-python.pylint",
+    // VSCode has implemented an optimized version
+    "coenraads.bracket-pair-colorizer",
+    "coenraads.bracket-pair-colorizer-2",
+    // Obsoleted by Pylance
+    "ms-pyright.pyright",
+  ],
+}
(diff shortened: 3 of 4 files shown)


hauntsaninja (commented):
Thanks for sharing, but I'm -0.5 on this. I don't want to get in the business of maintaining editor configurations, nor do I think typeshed has sufficiently unique needs to really justify shipping editor configurations. But maybe other maintainers feel differently!

AlexWaygood:
> But maybe other maintainers feel differently!

I don't use VSCode, so I'm pretty neutral here :)

In general I agree that I don't really want us into the business of maintaining editor configs (once we add one editor config, it gets hard to justify not adding all the other ones). So I guess I'm -0.5 as well.

Avasam:
I can understand the reticence

> it gets hard to justify not adding all the other ones

Fair, as I would be for adding the popular python ones (ie PyCharm, w/e most of you use, etc.) if it's useful.

The idea being that this can tremendously help new contributors, reduce friction with misconfigured editors, formatters and less reliance on waiting for pre-commit. As well as help them **see** errors and they're contributing (red/yellow squigglies).

As someone is who is more more experienced with VSCode settings, and already had to go through weeks of figuring out the right configurations, I figured I'd offer them so that other users of the editor don't have to redo all that figuring-out by themselves.

AlexWaygood:
FWIW: I used to use PyCharm, until I started contributing to typeshed and CPython. Hacking on CPython in particular just broke PyCharm's brain completely, so I abandoned it and went back to the editor I used before PyCharm, namely..... IDLE (don't laugh). It's as basic as they come, but the advantage of using an editor with 0 features is that none of them get broken when you try to hack on CPython itself 😛

I've tried using VSCode a few times, but it never seems to "click" for me. Don't know why, exactly.

AlexWaygood:
Maybe we could put these settings somewhere else online, and link to them from typeshed's contributing guide. That way future contributors won't have to wade through the weeks of figuring things out, like you did, but we also don't have the pressure of having to maintain a settings file that we don't necessarily understand the ins and outs of. Thoughts? :)

JelleZijlstra:
I'm actually supportive of merging this (and similar settings for other major editors, if someone contributes them). We can add a disclaimer comment that the maintainer team doesn't commit to maintaining the file or something, but in the meantime this can definitely help contributors.

AlexWaygood:
> I'm actually supportive of merging this (and similar settings for other major editors, if someone contributes them). We can add a disclaimer comment that the maintainer team doesn't commit to maintaining the file or something, but in the meantime this can definitely help contributors.

Then let's do it. @Avasam, can you add the disclaimer comment somewhere? :)

Avasam:
Feel free to wordsmith!

AlexWaygood (approved):
Thanks @Avasam!

PR #9408, opened 2022-12-25
`mypy_test.py`: Allow non-types dependencies

AlexWaygood:
Work towards #5768 (cc. @Avasam)

PR #9441, opened 2023-01-02
Improve mypy flags for regr_test.py and typecheck_typeshed.py

AlexWaygood:
- Remove redundant flags that are now enabled by default.
- Add `--pretty` to `typecheck_typeshed.py`.

`mypy_test.py` flags are being reworked in #9408.

### DIFF
--- a/tests/regr_test.py
+++ b/tests/regr_test.py
@@ -166,7 +166,6 @@ def run_testcases(
         "--python-version",
         version,
         "--show-traceback",
-        "--show-error-codes",
         "--no-error-summary",
         "--platform",
         platform,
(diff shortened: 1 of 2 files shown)


JelleZijlstra:
No strong feelings here, but there's some value in keeping these flags even if they are enabled by default: it's more explicit, and we are insulated against changes if mypy decides to flip the default again in the future.

AlexWaygood:
Fair points. I guess I'm still weakly in favour of this PR — it feels _unlikely_ that these changes to the mypy defaults will be reverted, and the added value from explicitness feels a _little_ slim to me here.

srittau (approved):
LGTM, I slightly prefer to remove the redundant options, but have no really strong feelings.

PR #9443, opened 2023-01-02
Shard `mypy_primer` more

AlexWaygood:
Might help reduce the performance impact of #9433 on `mypy_primer` CI runs.

### DIFF
--- a/.github/workflows/mypy_primer.yml
+++ b/.github/workflows/mypy_primer.yml
@@ -22,7 +22,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        shard-index: [0, 1, 2, 3]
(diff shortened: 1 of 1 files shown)


AlexWaygood:
Yeah, doesn't look like this helps

AlexWaygood:
It looks like we had:

0. A 6:01 shard
1. An 8:18 shard
2. A 5:45 shard
3. A 6:23 shard
4. A 17:02 shard(!!)
5. A 5:35 shard

For shard 4, it took **16 minutes** to check `manticore`, far exceeding any of the other projects. No other project took >1 minute.

Whether or not we decide to revert #9433, that's quite interesting! Cc. @hauntsaninja.

hauntsaninja:
That is interesting. I took a closer look, should be a simple fix to improve perf on manticore.

PR #9444, opened 2023-01-02
Deduplicate stubtest command in actions

Avasam:
Split off from https://github.com/python/typeshed/pull/9374/files#r1060179921

Makes the command to run have a single source of truth


AlexWaygood (approved):
LGTM, providing CI's happy!

PR #9449, opened 2023-01-03
Support third-party stub external dependencies in pytype

Avasam:
Work towards #5768
Based on, and mostly a copy of what's done in, #9434 (hence I'm keeping this as draft until the pyright PR is merged)

I'll revert the demo changes in `METADATA.toml` after the above is merged.

The pytype failure seen in #9374 seems to be caused by non-typed external dependencies, which we decided we won't support at the moment anyway.


tests/README.md
```diff
@@ -17,7 +17,18 @@ objects at runtime.
 in the `tests` and `scripts` directories.
 
 To run the tests, follow the [setup instructions](../CONTRIBUTING.md#preparing-the-environment)
-in the `CONTRIBUTING.md` document. In particular, we recommend running with Python 3.9+.
+in the `CONTRIBUTING.md` document. In particular, you have to run with Python 3.9+.
```
Avasam, line 20: I changed the wording here because since #9382, `test/utils.py` (which is used by all tests) uses `functools.cache`, which is unavailable in Python 3.8.

tests/README.md
```diff
@@ -17,7 +17,20 @@ objects at runtime.
...
+
+In order for `pytype_test` and `pyright_test` to work correctly, some third-party stubs
```
Avasam, line 22: Same wording, just added `pytype_test and` moved it up to avoid duplicating this entire paragraph.

Avasam:
Sucessful pytype run: https://github.com/python/typeshed/actions/runs/3833991938/jobs/6525955844

tests/README.md
AlexWaygood: In practice we've required 3.9+ for a lot longer, due to the use of `ast.unparse` in `tests/check_new_syntax.py` :)

.github/workflows/tests.yml
```diff
@@ -69,6 +69,13 @@ jobs:
           cache: pip
           cache-dependency-path: requirements-tests.txt
       - run: pip install -r requirements-tests.txt
+      - name: Install external dependencies for 3rd-party stubs
+        run: |
+          DEPENDENCIES=$(python tests/get_external_stub_requirements.py)
+          if [ -n "$DEPENDENCIES" ]; then
+            echo "Installing packages: $DEPENDENCIES"
+            pip install $DEPENDENCIES
+          fi
```
AlexWaygood, line 78: (Same rationale as for the pyright job)

```suggestion
          fi
     - run: pip freeze --all
```

CONTRIBUTING.md
```diff
@@ -46,9 +46,11 @@ virtual environment. If you're not familiar with what it is and how it works,
 please refer to this
 [documentation](https://packaging.python.org/guides/installing-using-pip-and-virtual-environments/).
 
+Some tests require extra setup steps to install the required dependencies.
```
AlexWaygood, line 49:
```suggestion
Note that some tests require extra setup steps to install the required dependencies.
```

tests/README.md
```diff
@@ -17,7 +17,20 @@ objects at runtime.
...
+may require dependencies external to typeshed to be installed in your virtual environment
```
AlexWaygood, line 23: Clarify that these dependencies are in *addition* to the ones specified in `requirements-tests.txt`

```suggestion
may require extra dependencies external to typeshed to be installed in your virtual environment
```

AlexWaygood (commented):
Looks good, thanks! Some very minor points below

AlexWaygood:
@rchen152, I'd love your opinion on this as a pytype maintainer, if you've the time.

We're trying to rework our test suite to allow our third-party stubs packages to have dependencies on non-stubs packages external to typeshed which ship with `py.typed` files. For `mypy_test.py`, my proposed approach in #9408 is to dynamically create virtual environments in order to test each stubs package in isolation. For pyright, however, we went with #9434, which simply installs all dependencies for all third-party stubs into one environment before testing them all together.

The pyright approach has the advantage that the test code is much simpler and faster, but could lead to false negatives. If you add `import numpy` in a stubs package, but forget to declare a dependency on `numpy` for that package, pyright might not emit an error due to `numpy` already being installed for some other stubs package that we're testing.

We could probably go either way when it comes to running pytype in CI — the "simple" approach we're using for pyright, or the "somewhat more principled" approach I'm proposing for mypy. The proposal in this PR is to use the pyright approach. Do you have any preference at all?

tests/README.md
Avasam: Good to know. I believe you could still run other tests, so "recommend" wasn't *too* inaccurate :P

I only noticed because I had 3.8 installed in WSL.

.github/workflows/tests.yml
Avasam: Thanks. Didn't notice that was missing.

AlexWaygood (approved):
LGTM, but I'll wait a little bit to see if the pytype maintainers have any thoughts before merging :)

rchen152:
Thanks for looping me in! This change looks good to me =) If mypy is already testing each package in isolation, it seems reasonable to use the "simple" approach for pytype.

AlexWaygood:
Brilliant, thanks for taking a look!

PR #9450, opened 2023-01-03
Autoformat YAML files

Avasam:
A quick formatting run of YAML files for things like consistent single vs double quotes, spacing before comments and indentation.
Adds a pre-commit hook to autoformat YAML files. I don't have a strong opinion on which formatter to use for YAML.

<details>
<summary>Old PR description</summary>

Just a quick formatting run of YAML files. Consistent single vs double quotes, spacing before comments and indentation. I don't think there's an established preference for YAML style so I let [redhat.vscode-yaml](https://marketplace.visualstudio.com/items?itemName=redhat.vscode-yaml) use its default settings.

I'm open to suggestions for a formatter to add to pre-commit. (vscode-yaml itself uses prettier, which I'm not a great fan of personally, at least in JS environments where I prefer ESLint+[dprint](https://dprint.dev/plugins/typescript/). But its up to you, seems fine for yaml, maybe markdown, etc.)

</details>

### DIFF
--- a/.github/workflows/daily.yml
+++ b/.github/workflows/daily.yml
@@ -6,8 +6,8 @@ on:
     - cron: "0 0 * * *"
   pull_request:
     paths:
-      - 'requirements-tests.txt'
-      - '.github/workflows/daily.yml'
+      - "requirements-tests.txt"
+      - ".github/workflows/daily.yml"
 
 # Please keep the permissions minimal, as stubtest runs arbitrary code from pypi.
 permissions:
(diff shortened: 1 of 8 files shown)


JelleZijlstra:
I don't think we should reformat these unless we also introduce autoformatting in CI; otherwise formatting will just regress again and this commit will clog the git history.

I would choose prettier out of familiarity but don't really care about what exactly we pick.

Avasam:
> [...] otherwise formatting will just regress again and this commit will clog the git history.

Yeah I agree. I added the prettier pre-commit hook. Seems to be working just-fine as-is.

Should be possible to support markdown as well just by adding "markdown" to `types_or`. Although when I tried it, it made a lot of changes that seem to differ from current styles. Markdownlint might be a better choice for that one.

> AssertionError: All pre-commit requirements must also be listed in `requirements-tests.txt` (missing 'prettier')

Uh oh. Humm, what'd be the best way to tell `Check file consistency` that prettier isn't a pypi package? Hardcode it in the test?

JelleZijlstra (approved):

srittau (approved):

PR #9452, opened 2023-01-03
update SimpleHTTPRequestHandler

ethanfurman:
The `index_pages` parameter has been removed from `__init__` as subclassing is the encouraged method for modifying `SimpleHttpRequestHandler`.

`index_pages` should be a tuple of 0 or more strings (not sure I have the type hint correct).

### DIFF
--- a/stdlib/http/server.pyi
+++ b/stdlib/http/server.pyi
@@ -52,7 +52,7 @@ class BaseHTTPRequestHandler(socketserver.StreamRequestHandler):
 class SimpleHTTPRequestHandler(BaseHTTPRequestHandler):
     extensions_map: dict[str, str]
     if sys.version_info >= (3, 12):
-        index_pages: tuple[str, ...]
+        index_pages: ClassVar[tuple[str, ...]]
     def __init__(
         self,
         request: socketserver._RequestType,
--- a/stdlib/http/server.pyi
+++ b/stdlib/http/server.pyi
@@ -52,25 +52,15 @@ class BaseHTTPRequestHandler(socketserver.StreamRequestHandler):
 class SimpleHTTPRequestHandler(BaseHTTPRequestHandler):
     extensions_map: dict[str, str]
     if sys.version_info >= (3, 12):
-        def __init__(
-            self,
-            request: socketserver._RequestType,
-            client_address: socketserver._AddressType,
-            server: socketserver.BaseServer,
-            *,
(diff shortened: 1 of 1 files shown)


AlexWaygood:
> The `index_pages` parameter has been removed from `__init__` as subclassing is the encouraged method for modifying `SimpleHttpRequestHandler`.

But it does nonetheless exist on the CPython `main` branch, no? :) https://github.com/python/cpython/blob/64ed609c532a12b27f67a1e12e9e02f136ee3a94/Lib/http/server.py#L664

We tend to be descriptive rather than prescriptive in typeshed. If a parameter exists at runtime, we tend to include it in the stub, even if it's not considered great style to use that parameter at runtime. If we do otherwise, people tend to complain loudly at us. It also means we have to switch off some of our tests that check that the stub is consistent with the runtime, and switching off those tests means our stubs become vulnerable to growing out of date.

(You don't see those tests failing on this PR because we haven't switched them on yet for 3.12, since 3.12 is still in development.)

ethanfurman:
[PR changing `__init__`](https://github.com/python/cpython/pull/100731) has now been merged.  :-)

AlexWaygood:
> [PR changing `__init__`](https://github.com/python/cpython/pull/100731) has now been merged. :-)

Ah! I didn't realise there was a pending PR :)

JelleZijlstra:
Maybe we're being a bit early with the 3.12 features here :) We can make this change though now that main has been changed.

JelleZijlstra (approved):

stdlib/http/server.pyi
```diff
@@ -52,25 +52,15 @@ class BaseHTTPRequestHandler(socketserver.StreamRequestHandler):
...
-        def __init__(
-            self,
-            request: socketserver._RequestType,
-            client_address: socketserver._AddressType,
-            server: socketserver.BaseServer,
-            *,
-            directory: str | None = ...,
-        ) -> None: ...
-
+        index_pages: tuple[str, ...]
```
AlexWaygood, line 55: I see you refer to `index_pages` as a "class variable" in the news entry for the CPython change. If the attribute is only meant to be set on the class itself, not on instances, we can represent it by using the special `ClassVar` annotation from the `typing` module. 

```suggestion
        index_pages: ClassVar[tuple[str, ...]]
```

AlexWaygood (approved):
Thanks!

PR #9453, opened 2023-01-03
Bump Pillow to 9.4

Avasam:
Closes #9448


stubs/Pillow/PIL/Image.pyi
```diff
@@ -31,53 +34,18 @@ _Color: TypeAlias = Union[int, tuple[int], tuple[int, int, int], tuple[int, int,
...
+NORMAL: Literal[0]  # deprecated
+SEQUENCE: Literal[1]  # deprecated
+CONTAINER: Literal[2]  # deprecated
 
 class DecompressionBombWarning(RuntimeWarning): ...
 class DecompressionBombError(Exception): ...
 
 MAX_IMAGE_PIXELS: int | None
 
-NONE: Literal[0]
```
Avasam, line 44: There's 3 sets of changes here:
1. Remove constants that also got removed
2. Add a comment for deprecated constants
3. Some global constants are now generated differently and extracted from enums. I've made that clearer by grouping them with their respective enums and adding a comment. To hopefully help maintainability.

stubs/Pillow/PIL/Image.pyi
```diff
@@ -153,9 +161,11 @@ class Image:
     palette: Any
     info: dict[Any, Any]
     readonly: int
-    pyaccess: Any
+    pyaccess: PyAccess | None
     is_animated: bool  # not present on all Image objects
     n_frames: int  # not present on all Image objects
+    # Can never be None, but can be missing
+    encoderconfig: tuple[Incomplete, ...] | _NotImplementedType
```
Avasam, line 168: Not sure how to properly deal with this case.

srittau, line 168: We usually just add the item with a comment:

```suggestion
    # Only defined after a call to save().
    encoderconfig: tuple[Incomplete, ...]
```

Alternatively, we could just leave it out, as this looks like an internal field.

stubs/Pillow/PIL/ImageFont.pyi
```diff
@@ -74,6 +74,7 @@ class FreeTypeFont:
         stroke_width: float = ...,
         anchor: str | None = ...,
         ink=...,
+        start: tuple[int, int] | None = ...,
```
srittau, line 77: Are you sure these are `int`s? Usually in PIL, coordinates are `float`s. In this case, the values are also passed to `self.font.render()`, though, and I'm not sure what that accepts.

stubs/Pillow/PIL/ImageFont.pyi
```diff
@@ -86,6 +87,7 @@ class FreeTypeFont:
...
```
srittau, line 90: Same here.

stubs/Pillow/PIL/WebPImagePlugin.pyi
```diff
@@ -1,13 +1,15 @@
 from typing import Any, ClassVar
-from typing_extensions import Literal
+from typing_extensions import Literal, TypeAlias
 
 from .ImageFile import ImageFile
 
 SUPPORTED: bool
+_XMP_Tags: TypeAlias = dict[str, _XMP_Tags]
```
srittau, line 7: This is a non-terminating recursive definition, i.e. impossible. I think this should be:

```suggestion
_XMP_Tags: TypeAlias = dict[str, str | _XMP_Tags]
```

srittau (changes requested):
Thanks! Remarks below.

stubs/Pillow/PIL/Image.pyi
```diff
@@ -1,10 +1,13 @@
-from _typeshed import Self, SupportsRead, SupportsWrite
+from _typeshed import Incomplete, Self, SupportsRead, SupportsWrite
+from builtins import _NotImplementedType
```
Avasam, line 2:
```suggestion
```

stubs/Pillow/PIL/WebPImagePlugin.pyi
Avasam: You are right! Thanks.

stubs/Pillow/PIL/ImageFont.pyi
Avasam: Checked the doc on the change: https://pillow.readthedocs.io/en/stable/releasenotes/9.4.0.html#added-start-position-for-getmask-and-getmask2
> fractional coordinates

Sounds like float to me.

It defaulted to an int. But that's probably just because `0` is simpler to write that `0.0`

srittau (approved):

PR #5847, opened 2021-08-05
`paramiko`: Change dependencies on `types-cryptography`, allow using upstream type hints

intgr:
Since version 3.4, the cryptography library itself ships with type hints: https://cryptography.io/en/latest/changelog/#v3-4

This is the first step for #5618. Currently, if I use `types-paramiko` and `cryptography`, it forces me to use `types-cryptography` as well. It should be up to the end-user to decide if they want upstream hints or from typeshed.

As discussed in #5618 and #5768, I have replaced it with a dependency on `types-paramiko` -> `cryptography` and `types-pyOpenSSL` -> `cryptography`.

Issue #8312, opened 2022-07-15
Maybe change METADATA.toml "requires"?


hauntsaninja:
The context is I'm looking into #5768.

There's a lot of logic we have that either assumes or checks that entries in "requires" are typeshed stub packages. It seems potentially useful to split "requires" into "requires_typeshed" (a list of required packages that originate from typeshed) and "requires_external" (a list of external dependencies). Doing so could e.g. help avoid issues where we have typos in some requires_typeshed deps or potentially implicitly trust packages that start with "types-". Could also make it easier to build tooling (e.g. knowing what we could put on MYPYPATH to test against latest typeshed).

srittau:
I'd prefer to keep just one `requires` field. We ship `METADATA.toml` with the types package in question and for external users the source of the dependency should not matter and seems rather artificial. For internal purposes it seems rather trivial to check whether a dependency is internal.

hauntsaninja:
I checked on grep.app and I think we're the only people who use `requires`, so not sure we should worry much about external users. I guess how much tooling and security benefit this would bring is debatable, but at the minimum this would be nice typo protection.

Anyway, your opinion should be the default here, so I won't push for this change. Are you on board with the rest of https://github.com/typeshed-internal/stub_uploader/pull/59, assuming `requires_typeshed` becomes a derived property computed from filtering the `requires` field of `METADATA.toml`? I still think adding something to the Metadata class is the sanest way to make sure all the old call sites that assume all deps come from typeshed only see typeshed deps.

srittau:
> I checked on grep.app and I think we're the only people who use requires, so not sure we should worry much about external users.

I agree that this is more a theoretical concern at the moment.

> Anyway, your opinion should be the default here, so I won't push for this change.

Well, your suggestion got a lot of upvotes, so it's quite possible I'm in the minority here, and it's also not an issue I'll fight tooth and nail over. So, I'd be interested what other people think.

> Are you on board with the rest of https://github.com/typeshed-internal/stub_uploader/pull/59

While I haven't reviewed the PR in detail yet, because of the issues under discussion, the general idea seems sound to me. And it definitely makes sense to distinguish between "internal" and "external" dependencies.

AlexWaygood:
> I'd be interested what other people think.

I like @hauntsaninja's idea best.

[...]

PR #9459, opened 2023-01-04
Delete stubs for `cryptography`

### LINKED ISSUE #5618: Remove cryptography (depends on #5952)

### LINKED ISSUE #5768: Allow non-types dependencies

AlexWaygood:
Closes #5768.
Closes #5618.
Closes #5847.


AlexWaygood:
Looks like the pytype test is still failing @Avasam :(

I'm taking a look at the `stub_uploader` error.

hauntsaninja:
Looks like a bug with https://github.com/typeshed-internal/stub_uploader/pull/80
`ordered` is not guaranteed to only contain stubs distributions (which is e.g. why we have `if dist in distributions`)

hauntsaninja:
I can make a PR!

AlexWaygood:
> I can make a PR!

Amazing, thank you! I got held up by the fact that a bunch of the tests don't pass on Windows :)

Avasam:
After further investigation, A lot of the pytype errors seem to reference types that mypy also flags as missing. I think we should fix mypy errors first and then take a new look at the pytype errors.

AlexWaygood:
> After further investigation, A lot of the pytype errors seem to reference types that mypy also flags as missing (ie: `cryptography.hazmat.primitives.ciphers`). I think we should fix mypy errors first and then take a new look at the pytype errors.

All the mypy tests pass locally with https://github.com/python/typeshed/pull/9408 applied ;)

Avasam:
> All the mypy tests pass locally with #9408 applied ;)

So mypy doesn't pass now, because cryptography is not installed in mypy_test. But will with #9408 .
pytype test on CI *does* install cryptography, and it's not using a virtual environment, so pytype should the types.

Unless I'm missing something here (and with the huge pytype errors, I might), seems like a pytype issue with py.typed modules.

AlexWaygood:
> > All the mypy tests pass locally with #9408 applied ;)
> 
> So mypy doesn't pass now, because cryptography is not installed in mypy_test. But will with #9408 . pytype test on CI _does_ install cryptography, and it's not using a virtual environment, so pytype should see the modules/types. I should double-check the same issue happens locally.
> 
> Unless I'm missing something here (and with the huge pytype errors, I might), seems like a pytype issue with py.typed modules.

@rchen152, any idea what's going on here with the pytype errors? As @Avasam says, it looks like `cryptography` is being successfully installed in CI (https://github.com/python/typeshed/actions/runs/3841901232/jobs/6542651921) :/

Avasam:
Oh, wanna bet `parse_pyi=True` expects only `.pyi` files?

Avasam:
Depending on how soon you can complete #5768 and get rid of cryptography, vs the complexity and time-to-implement a solution on pytype's side, supressing this exception is feasable, although not great:
```python
    stderr: str | None = None
    try:
        with pytype_config.verbosity_from(options):
            ast = loader.load_file(_get_module_name(filename), filename)
    except BadDependencyError as error:
        error_str = str(error)
        # pytype errors out with non-types dependencies.
        # mypy and pyright already take care of ensuring external types are installed and do exist,
        # but this issue is problematic as it hides other potential errors in the same file.
        if error_str.startswith("Can't find pyi for"):
            print(
                colored(
                    f"{error_str}. This is likely an issue with pytype in typeshed, and not an issue with the stub.", "yellow"
                )
            )
        else:
            stderr = traceback.format_exc()
    except Exception:
        stderr = traceback.format_exc()
```

AlexWaygood:
> Depending on how soon you want to complete #5768 and get rid of cryptography, vs the complexity and time-to-implement a solution on pytype's side, supressing this exception is feasable, although not great 😕

That's good to have as a fallback option, but definitely wouldn't be my preferred solution :)

rchen152:
Oh, I didn't realize that you wanted pytype_test to be able to process non-pyi dependencies. That seems a little challenging to set up. The simplest (well, definitely the quickest) thing to do is probably just to mark those dependencies as missing so that pytype treats them as having default stubs (`def __getattr__(name) -> Any`).

I'll try changing the `missing_file` argument to `pytype.imports.typeshed.Typeshed` to take a list of missing dependencies instead of a file, so that pytype_test can use tests/utils/read_dependencies to get a list of external dependencies (like what get_external_stub_requirements does) and pass the list to the `Typeshed` object.

Avasam:
Would closing #5768 also close #8312 since a decision seems to have been taken?

stubs/pyOpenSSL/METADATA.toml
```diff
@@ -1,2 +1,3 @@
 version = "23.0.*"
-requires = ["types-cryptography"]
+# Requires a version of cryptography with a `py.typed` file
+requires = ["cryptography>=35.0.0"]
```
Avasam, line 3: Interesting. I feel like it would make sense to enforce a minimal version for external dependencies of 3rd party stub.

AlexWaygood: Definitely something to explore in the future!

Avasam: Without having to enforce it through CI. Maybe a mention in documentation that external dependencies should be at least pinned to their first `py.typed` version? (and optimally to the oldest version that has the types referenced by the stub)

AlexWaygood:
> Would closing #5768 also close #8312 since a decision seems to have been taken?

Let's leave it open for now; see comments on the issue :)

stubs/pyOpenSSL/METADATA.toml
AlexWaygood: I'll try to work on docs as soon as this PR makes it over the line — this is a very good point, thanks!

rchen152:
> Oh, I didn't realize that you wanted pytype_test to be able to process non-pyi dependencies. That seems a little challenging to set up. The simplest (well, definitely the quickest) thing to do is probably just to mark those dependencies as missing so that pytype treats them as having default stubs (`def __getattr__(name) -> Any`).
> 
> I'll try changing the `missing_file` argument to `pytype.imports.typeshed.Typeshed` to take a list of missing dependencies instead of a file, so that pytype_test can use tests/utils/read_dependencies to get a list of external dependencies (like what get_external_stub_requirements does) and pass the list to the `Typeshed` object.

It took a little longer than I expected, but I believe I have this working. It requires both a change to pytype (which I'll export and release tomorrow or Wednesday) and a change to pytype_test, which I've prepared a draft pull request for: https://github.com/python/typeshed/pull/9486.

PR #9460, opened 2023-01-05
Fix Any subclassing in setuptools

Avasam:
Reference the docutils stub.

As far as I understand, there's no issue with this if docutils isn't installed. And gives proper type when it is.

Fixes 3 subclassing errors (the same error in setuptools, pyinstaller and Pygments stubs)

### DIFF
--- a/stubs/setuptools/setuptools/_distutils/command/check.pyi
+++ b/stubs/setuptools/setuptools/_distutils/command/check.pyi
@@ -1,12 +1,11 @@
 from typing import Any
-from typing_extensions import TypeAlias
+
+import docutils.utils
 
 from ..cmd import Command
 
-_Reporter: TypeAlias = Any  # really docutils.utils.Reporter
-
 # Only defined if docutils is installed.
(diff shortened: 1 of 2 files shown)


stubs/setuptools/setuptools/_distutils/command/check.pyi
```diff
@@ -1,12 +1,11 @@
 from typing import Any
-from typing_extensions import TypeAlias
 
-from ..cmd import Command
+import docutils.utils
 
-_Reporter: TypeAlias = Any  # really docutils.utils.Reporter
+from ..cmd import Command
 
 # Only defined if docutils is installed.
```
hauntsaninja, line 7: We can remove this comment

Avasam: The comment is still true. I think it has value if someone wonders why its in the stub but not in their runtime.

AlexWaygood (approved):

PR #9461, opened 2023-01-05
Explain why we subclass ImageQt with Any

Avasam:
These are way too complex, with 4 different possible sources (2 deprecated)
And we don't want to force the user to install PyQt or Pyside when they may not even use it.

Fixes 6 subclassing errors (the same error in Pillow, D3DShot, python-xlib, fpdf2, PyScreeze, and PyAutoGUI stubs)

Ref: #9491

### DIFF
--- a/stubs/Pillow/PIL/ImageQt.pyi
+++ b/stubs/Pillow/PIL/ImageQt.pyi
@@ -3,7 +3,10 @@ from typing_extensions import Literal, TypeAlias
 
 from .Image import Image
 
-QImage: TypeAlias = Any  # imported from either of {PyQt6,PySide6,PyQt5,PySide2}.QtGui
+# imported from either of {PyQt6,PySide6,PyQt5,PySide2}.QtGui
+# These are way too complex, with 4 different possible sources (2 deprecated)
+# And we don't want to force the user to install PyQt or Pyside when they may not even use it.
+QImage: TypeAlias = Any
 QPixmap: TypeAlias = Any
 
 qt_versions: Any
@@ -15,7 +18,7 @@ def fromqimage(im: Image | QImage) -> Image: ...
 def fromqpixmap(im: Image | QImage) -> Image: ...
 def align8to32(bytes: bytes, width: int, mode: Literal["1", "L", "P"]) -> bytes: ...
 
-class ImageQt(QImage):
+class ImageQt(QImage):  # type: ignore[misc]
     def __init__(self, im: Image) -> None: ...
 
 def toqimage(im: Image) -> ImageQt: ...


hauntsaninja (commented):
Maybe this should be a protocol?

Avasam:
> Maybe this should be a protocol?

I thought about it, but `ImageQt` is used as a return type, so it'd have to completely match QtImage from all 4 implementations. I did a quick try and it got messy real quick with thousands of lines copied from PyQt6.

Protocol could probably be simplified by using Any when it starts referencing other types (so at least we have all methods and members). But would a protocol still work as a return type?

AlexWaygood:
> Protocol could probably be simplified by using Any when it starts referencing other types (so at least we have all methods and members). But would a protocol still work as a return type?

Protocols can be used as a return type! I agree with @hauntsaninja that we can probably at least try to roughly describe the shape of the superclass here.

Avasam:
I meant that I'm concerned about the variance:

```
from typing import Protocol

class Abc:
    foo: str
    
class AbcProtocol(Abc, Protocol): ...  # type: ignore[misc]

reveal_type(AbcProtocol().foo). # str

class Abcd:
    foo: str
   
abcd: Abcd
abcd = AbcProtocol()  # Incompatible types in assignment (
abcproto: AbcProtocol = Abcd()
```

hauntsaninja:
I'm not sure I follow, wouldn't users just use the `QImage` type hint?
(Or is it a case where users do actually know what type they're getting back and e.g. do nominal checking or the cases where the types' APIs are non-overlapping are actually important to users?)

Avasam:
Here's a minimal example in PyQt6. This is what we need to support.
```python
from PyQt6 import QtGui
data: memoryview  # Some graphics data from elsewhere in the code

def show_qimage(image: QtGui.QImage):
    # Show the image someplace in the GUI
    pass

# With PyQt directly
data: memoryview  # Some graphics data from elsewhere in the code
qimage = QtGui.QImage(data, 640, 480, 640 * 4, QtGui.QImage.Format.Format_RGBA8888)
show_qimage(qimage)

# using the pillow wrapper instead
pil_image: Image.Image  # Some pillow image from elsewhere in the code
qimage = ImageQt.ImageQt(pil_image)
show_qimage(qimage) # Argument of type "ImageQt" cannot be assigned to parameter "image" of type "QImage" in function "show_qimage" "ImageQt" is incompatible with "QImage"
```

Avasam:
@AlexWaygood , I don't see an "easy" solution for this (or any *good* solution at all). Since type-checkers can't have a conditional type based on if a module exists or not. (without having to check for a full ImportError, which can't be done statically, I feel like checking for a module's existence could be a beneficial proposition)

Can I move this to #9491 ?

AlexWaygood:
> @AlexWaygood , I don't see an "easy" solution for this (or any _good_ solution at all). Since type-checkers can't have a conditional type based on if a module exists or not. (without having to check for a full ImportError, which can't be done statically, I feel like checking for a module's existence could be a beneficial proposition)
> 
> Can I move this to #9491 ?

I'm fine with that. It's not like we're introducing a regression here, after all -- we're already sbuclassing `Any` in the stub. If anybody wants to invest some time into working out a more principled solution, they're always welcome to do so at a later date!

Avasam:
Closing in favor of #9491

PR #9462, opened 2023-01-05
Fix Any subclassing in beautifulsoup4

Avasam:
Fixes 2 subclassing errors by referencing the html5lib stub


hauntsaninja (approved):
Thanks, looks great!

PR #9463, opened 2023-01-05
Merge `jmespath` from microsoft/python-type-stubs

Avasam:
Bring `types-jmespath` on par with https://github.com/microsoft/python-type-stubs/tree/main/jmespath

### DIFF
--- a/stubs/jmespath/jmespath/exceptions.pyi
+++ b/stubs/jmespath/jmespath/exceptions.pyi
@@ -12,6 +12,9 @@ class ParseError(JMESPathError):
     def __init__(self, lex_position: int, token_value: str, token_type: str, msg: str = ...) -> None: ...
 
 class IncompleteExpressionError(ParseError):
+    # When ParseError is used directly, the token always have a non-null value and type
+    token_value: str | None  # type: ignore[assignment]
+    token_type: str | None  # type: ignore[assignment]
     expression: str
     def set_expression(self, expression: str) -> None: ...
 
@@ -25,7 +28,6 @@ class ArityError(ParseError):
     expected_arity: int
     actual_arity: int
     function_name: str
-    expression: None
     def __init__(self, expected, actual, name) -> None: ...
 
 class VariadictArityError(ArityError): ...
--- a/stubs/jmespath/jmespath/exceptions.pyi
+++ b/stubs/jmespath/jmespath/exceptions.pyi
@@ -1,44 +1,41 @@
+from collections.abc import Sequence
 from typing import Any
 
 class JMESPathError(ValueError): ...
 
 class ParseError(JMESPathError):
-    lex_position: Any
-    token_value: Any
-    token_type: Any
-    msg: Any
-    expression: Any
-    def __init__(self, lex_position, token_value, token_type, msg=...) -> None: ...
+    lex_position: int
+    token_value: str
+    token_type: str
+    msg: str
+    expression: str | None
+    def __init__(self, lex_position: int, token_value: str, token_type: str, msg: str = ...) -> None: ...
 
 class IncompleteExpressionError(ParseError):
-    expression: Any
-    lex_position: Any
-    token_type: Any
-    token_value: Any
-    def set_expression(self, expression) -> None: ...
+    expression: str
+    def set_expression(self, expression: str) -> None: ...
 
 class LexerError(ParseError):
-    lexer_position: Any
-    lexer_value: Any
-    message: Any
-    expression: Any
-    def __init__(self, lexer_position, lexer_value, message, expression: Any | None = ...) -> None: ...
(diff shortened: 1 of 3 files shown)


stubs/jmespath/jmespath/exceptions.pyi
```diff
@@ -1,44 +1,41 @@
...
+    expression: str | None
+    def __init__(self, lex_position: int, token_value: str, token_type: str, msg: str = ...) -> None: ...
 
 class IncompleteExpressionError(ParseError):
-    expression: Any
-    lex_position: Any
-    token_type: Any
-    token_value: Any
-    def set_expression(self, expression) -> None: ...
+    expression: str
```
AlexWaygood, line 15: These are different to the superclass: https://github.com/jmespath/jmespath.py/blob/bbe7300c60056f52413603cf3e2bcd0b6afeda3d/jmespath/exceptions.py#L36-L37

```suggestion
    expression: str
    token_type: str | None
    token_value: str | None
```

stubs/jmespath/jmespath/exceptions.pyi
```diff
@@ -1,44 +1,41 @@
...
 
 class LexerError(ParseError):
-    lexer_position: Any
-    lexer_value: Any
-    message: Any
-    expression: Any
-    def __init__(self, lexer_position, lexer_value, message, expression: Any | None = ...) -> None: ...
+    lexer_position: int
+    lexer_value: str
+    message: str
```
AlexWaygood, line 21: These attributes are all the same as in the superclass; no need to override them in the stub

```suggestion
```

stubs/jmespath/jmespath/exceptions.pyi
```diff
@@ -1,44 +1,41 @@
...
 
 class ArityError(ParseError):
-    expected_arity: Any
-    actual_arity: Any
-    function_name: Any
-    expression: Any
+    expected_arity: int
+    actual_arity: int
+    function_name: str
+    expression: None
```
AlexWaygood, line 28: Feels hard to believe that this is _always_ `None`, otherwise there's not much point in having it as an attribute ;)

Maybe just don't override it here, and let it inherit the `str | None` annotation from the superclass?

```suggestion
```

stubs/jmespath/jmespath/functions.pyi
```diff
@@ -18,4 +18,4 @@ class FunctionRegistry(type):
 
 class Functions(metaclass=FunctionRegistry):
     FUNCTION_TABLE: Any
-    def call_function(self, function_name, resolved_args): ...
+    def call_function(self, function_name: str, resolved_args: Iterable[Any]) -> Any: ...
```
AlexWaygood, line 21: What's the reason for using `Any` here? Should these maybe be `Incomplete`?

stubs/jmespath/jmespath/parser.pyi
```diff
@@ -15,5 +18,5 @@ class Parser:
 class ParsedResult:
     expression: str
     parsed: _TreeNode
-    def __init__(self, expression: str, parsed: _TreeNode) -> None: ...
+    def __init__(self, expression: str, parsed: _TreeNodeDict) -> None: ...
```
AlexWaygood, line 21: What's the rationale for this change? `dict` invariance? We use `_TreeNode` for all the parameter annotations in `jmespath.visitor`, so I'd be inclined to do so here just for consistency's sake

stubs/jmespath/jmespath/functions.pyi
AlexWaygood: Nah that's fine, but a comment might be nice :)

stubs/jmespath/jmespath/parser.pyi
Avasam: `dict` Invariance yes.
If you think this is better, but want consistency, I'd rather change the other annotations in jmespath as well.

AlexWaygood: Can we maybe leave this one for now, and tackle it in a separate PR? Another contributor added all the `_TreeNode` annotations fairly recently, and I'd like to give them a heads-up before changing them all :)

stubs/jmespath/jmespath/exceptions.pyi
Avasam: I don't think it is, no, but I can see how it seemed like it, as the names are similar
lex_position -> lexer_position
token_value -> lexer_value
msg -> message

AlexWaygood: Oh wow, thanks, I completely misread that one. 🤦‍♂️

AlexWaygood (approved):
Thanks!

PR #9466, opened 2023-01-06
Return type for `PIL.Image.load`

### LINKED ISSUE #9126: Pillow load() wrong return value
Pillow load incorrectly typed to return none and not a pixelaccess object per pillow documentation. Https://pillow.readthedocs.io/en/stable/reference/PixelAccess.html

DMRobertson:
A first step towards #9126.

Points I wasn't sure about:
- is this well-typed enough to be useful?
- does the Union `PixelAccess | PyAccess` return value actually get us anything? E.g. perhaps a Protocol describing the common parts of these two would be more useful


Avasam:
(I am not a maintainer, I just contribute a lot here)
1. Incomplete types are always better than flat-out wrong ones. Typeshed also generally prefers false negative over false-positive. So even `def load(self): ...` or `def load(self) -> Incomplete: ...` would be better.

2. We tend to avoid unions in return types (see https://github.com/python/typeshed/blob/main/CONTRIBUTING.md#conventions and https://github.com/python/typeshed/blob/main/CONTRIBUTING.md#conventions).
In your case you can't use an overload because you can't type base on an `ImportError`. Imo a protocol would do the job (idk if someone else has a better option)

Avasam:
If a solution is found for the return type. Do you think you could do all the other `def load` as well? Since they seem to also return the result of `ImageFile.ImageFile.load` or `Image.Image.load`?

JelleZijlstra:
@DMRobertson could you respond to Avasam's feedback?

DMRobertson:
Apologies all for the delay in responding. I have some spare time now; let me see if I can fix this up.

DMRobertson:
Taking a look at this with fresher eyes, it think(?) the history here is:

- Image.load originally returned a `PixelAccess` object, the latter defined as a CPython extension only
- In order to support running this on PyPy another implementation called `PyAccess` was written using `cffi`. (https://github.com/python-pillow/Pillow/issues/248 and https://github.com/python-pillow/Pillow/pull/476). As the docs for PyAccess put it:
  > The [PyAccess](https://pillow.readthedocs.io/en/stable/reference/PyAccess.html#module-PIL.PyAccess) module provides a CFFI/Python implementation of the [PixelAccess Class](https://pillow.readthedocs.io/en/stable/reference/PixelAccess.html#pixelaccess). This implementation is far faster on PyPy than the PixelAccess version.

So I think these really are meant to be implementations of the same interface. For that reason, I think returning a rather than `ImplA | ImplB` best captures the intent of the library.

> 2\. I think a protocol may still have variance issue (trying to assign AccessProtocol to a PyAccess for instance).

Similarly, I think the PIL authors don't really intend for any downstream user to care about the distinction between PixelAccess and PyAccess, so I don't think what you describe is likely to be a problem in practice. (Though let's see what the primer says.)

> If a solution is found for the return type. Do you think you could do all the other `def load` as well? Since they seem to also return the result of `ImageFile.ImageFile.load` or `Image.Image.load`?

Sure, I can add that.

Akuli:
I think returning `ImplA | ImplB` is probably the best solution. If we add a protocol, it becomes difficult for users to use it in type annotations, but `PixelAccess | PyAccess` isn't too bad although it isn't great either.

A protocol can be closer to what PIL authors intended, but typeshed generally types things with a "practicality beats purity" mindset, and IMO adding a fictional protocol wouldn't be very practical.

Akuli (approved):
While I don't like the fictional protocol, this is a big improvement over `-> None` and can be improved in future PRs.

PR #9470, opened 2023-01-06
Use pyasn1 in ldap3

Avasam:
Gonna have to wait for stub-uploader to push `types-pyasn1` first.
~~Dependant on #9471~~
Closes #9471

Fixes *a lot* of Any subtyping issues.

Ref: #9491


stubs/ldap3/ldap3/protocol/rfc4511.pyi
```diff
@@ -1,32 +1,26 @@
...
-attributeSelectorConstraint: _Any
+from typing_extensions import Final as _Final
+
+from pyasn1.type.constraint import ConstraintsIntersection, SingleValueConstraint, ValueRangeConstraint, ValueSizeConstraint
+from pyasn1.type.namedtype import NamedTypes
+from pyasn1.type.namedval import NamedValues
+from pyasn1.type.tag import TagSet
+from pyasn1.type.univ import Boolean, Choice, Enumerated, Integer, Null, OctetString, Sequence, SequenceOf, SetOf
+
+LDAP_MAX_INT: _Final = 2147483647
```
Avasam, line 9: If I understood https://github.com/python/typeshed/issues/7258#issuecomment-1369821606 correctly

stubs/ldap3/ldap3/utils/asn1.pyi
```diff
@@ -1,37 +1,49 @@
...
 
-# Enable when pyasn1 gets stubs:
-# from pyasn1.codec.ber.encoder import AbstractItemEncoder
-_AbstractItemEncoder: TypeAlias = Any
+from pyasn1.codec.ber.encoder import AbstractItemEncoder, tagMap, typeMap
 
-CLASSES: Any
+# Use _typeshed.SupportsGetItemBuffer after PEP 688
+SupportsGetItemBuffer: TypeAlias = SliceableBuffer | IndexableBuffer
+_Unused: TypeAlias = object
```
Avasam, line 10: Proposal to make this a helper TypeAlias in _typeshed? (plus it's time to do it before mypy updates its vendored _typeshed with a new release)

stubs/ldap3/ldap3/protocol/rfc4511.pyi
AlexWaygood: This looks fine to me, but I don't see any reason to do the `from typing_extensions import Final as _Final` thing -- we can just do `from typing_extensions import Final` :)

stubs/ldap3/ldap3/utils/asn1.pyi
```diff
@@ -1,37 +1,49 @@
...
+class LDAPBooleanEncoder(AbstractItemEncoder):
     supportIndefLenMode: bool
     # Requires pyasn1 > 0.3.7
-    def encodeValue(self, value, asn1Spec, encodeFun, **options): ...
+    def encodeValue(self, value: bool | int, asn1Spec: _Unused, encodeFun: _Unused, **options: _Unused): ...
 
-customTagMap: Any
-customTypeMap: Any
+customTagMap = tagMap
+customTypeMap = typeMap
```
AlexWaygood, line 24: Hmm, maybe we should just remove these from the stub? They look like artefacts of stubgen-generated stubs.

Looking at the source code, they seem pretty implementation-detail-y: the block of code is introduced with [a comment saying](https://github.com/cannatag/ldap3/blob/040e09fe37cf06801960d494dfadf0a6ed94e38f/ldap3/utils/asn1.py#L39-L40):

```
# Monkeypatching of pyasn1 for encoding Boolean with the value 0xFF for TRUE
# THIS IS NOT PART OF THE FAST BER DECODER
```

I say we remove these symbols from the stub, and add an allowlist entry (with a comment saying why they're omitted). Then we don't have to add implementation details to the `pyasn1` stubs. (All of the `tagMap` and `typeMap` symbols are marked with comments in the source code saying that they're deprecated, and none of them are included in `__all__` at runtime).

stubs/ldap3/ldap3/utils/asn1.pyi
AlexWaygood: Sure :)

stubs/ldap3/ldap3/utils/asn1.pyi
Avasam: Yeah alright. Sounds good to me. I didn't realize they were marked as deprecated.

stubs/ldap3/ldap3/protocol/rfc4511.pyi
```diff
@@ -1,32 +1,26 @@
...
-attributeSelectorConstraint: _Any
+from typing_extensions import Final as _Final
+
+from pyasn1.type.constraint import ConstraintsIntersection, SingleValueConstraint, ValueRangeConstraint, ValueSizeConstraint
+from pyasn1.type.namedtype import NamedTypes
+from pyasn1.type.namedval import NamedValues
+from pyasn1.type.tag import TagSet
+from pyasn1.type.univ import Boolean, Choice, Enumerated, Integer, Null, OctetString, Sequence, SequenceOf, SetOf
+
+LDAP_MAX_INT: _Final = 2147483647  # noqa: Y015  # name conflict
```
Avasam, line 9: flake8-pyi not happy about the alias. (`Final` is fine)

Avasam:
I had the same 2 errors locally. This really looks like a stubtest bug: https://github.com/python/typeshed/actions/runs/3866310116/jobs/6590353764#step:5

stubs/ldap3/ldap3/protocol/rfc4511.pyi
AlexWaygood: Oh interesting. Yeah, flake8-pyi currently has no understanding of the fact that some imports might be aliased. I've been wondering about fixing it, but this kind of thing actually comes up surprisingly rarely in typeshed, and would involve adding some complexity to flake8-pyi.

Still, the fact that we have a legit false-positive here is a pretty good reason to fix it! 😀

stubs/ldap3/ldap3/extend/novell/endTransaction.pyi
```diff
@@ -1,14 +1,16 @@
...
+from ldap3.protocol.novell import EndGroupTypeResponseValue
 
 from ...extend.operation import ExtendedOperation
 
 class EndTransaction(ExtendedOperation):
     request_name: str
     response_name: str
-    request_value: Any
-    asn1_spec: Any
+    request_value: Incomplete
```
AlexWaygood, line 10: https://github.com/cannatag/ldap3/blob/040e09fe37cf06801960d494dfadf0a6ed94e38f/ldap3/extend/novell/endTransaction.py#L35

```suggestion
    request_value: EndGroupTypeRequestValue
```

stubs/ldap3/ldap3/extend/novell/endTransaction.pyi
```diff
@@ -1,14 +1,16 @@
...
+    asn1_spec: EndGroupTypeResponseValue | None
```
AlexWaygood, line 11: `config()` is always called at the end of `__init__` in the superclass, so I don't think this can ever actually be `None`: https://github.com/cannatag/ldap3/blob/040e09fe37cf06801960d494dfadf0a6ed94e38f/ldap3/extend/operation.py#L43.

(This is the "template method pattern", which I hate, ugh.)

```suggestion
    asn1_spec: EndGroupTypeResponseValue
```

AlexWaygood (commented):
Thanks! I've left 3 comments on the first file. It looks like those comments also apply to a lot of the other touched files, so I'll let you respond to those comments before doing a further review :)

stubs/ldap3/ldap3/protocol/rfc4511.pyi
AlexWaygood: I've opened https://github.com/PyCQA/flake8-pyi/issues/316.

AlexWaygood:
> I had the same 2 errors locally. This really looks like a stubtest bug: https://github.com/python/typeshed/actions/runs/3866310116/jobs/6590353764#step:5 (the two types mentionned are not at those lines!)

Yeah, I can't say I really understand what's going on here. The runtime source code is a bit of a maze, but even if it's doing something extremely dynamic that stubtest can't understand, the fact that stubtest is magicking line numbers out of thin air is a bit strange. Don't know if @hauntsaninja has any theories?

Anyway, for now I guess you should probably just allowlist them :)

stubs/ldap3/ldap3/protocol/rfc4511.pyi
Avasam: The class `Final` is declared lower down in this file. Should be more obvious now that there's a comment though

stubs/ldap3/ldap3/extend/novell/endTransaction.pyi
Avasam: Ah! Good catch. I've fixed and filled in all `request_value` and `asn1_spec`

hauntsaninja:
The bug is just the line number? If so, why does it need an allowlist entry? stubtest's logic for line numbers is a little hacky, looks like it's just the line number from the file that the parent class is in (pyasn1.type.univ)

Avasam:
> The bug is just the line number? If so, why does it need an allowlist entry? stubtest's logic for line numbers is a little hacky, looks like it's just the line number from the file that the parent class is in (pyasn1.type.univ)

I couldn't find the right type from source. But I got it now

stubs/ldap3/ldap3/protocol/rfc4511.pyi
```diff
@@ -1,32 +1,26 @@
...
-rangeInt1To127Constraint: _Any
-size1ToMaxConstraint: _Any
-responseValueConstraint: _Any
-numericOIDConstraint: _Any
-distinguishedNameConstraint: _Any
-nameComponentConstraint: _Any
-attributeDescriptionConstraint: _Any
-uriConstraint: _Any
-attributeSelectorConstraint: _Any
+from typing_extensions import Final as _Final
```
AlexWaygood, line 1:
```suggestion
# Alias the import to avoid name clash with a class called "Final"
from typing_extensions import Final as _Final
```

stubs/ldap3/ldap3/protocol/rfc4511.pyi
AlexWaygood, line 9: We just disabled Y015 entirely in #9490!

```suggestion
LDAP_MAX_INT: _Final = 2147483647
```

stubs/ldap3/ldap3/protocol/rfc4511.pyi
```diff
@@ -1,32 +1,26 @@
...
+rangeInt0ToMaxConstraint: ValueRangeConstraint
+rangeInt1To127Constraint: ValueRangeConstraint
+size1ToMaxConstraint: ValueSizeConstraint
+responseValueConstraint: SingleValueConstraint
+numericOIDConstraint: None
+distinguishedNameConstraint: None
+nameComponentConstraint: None
+attributeDescriptionConstraint: None
+uriConstraint: None
+attributeSelectorConstraint: None
```
AlexWaygood, line 20: Hmm, I struggle to believe that these are _always_ meant to be None, otherwise that's just a bit pointless having them. Maybe this?

```suggestion
numericOIDConstraint: Incomplete | None
distinguishedNameConstraint: Incomplete | None
nameComponentConstraint: Incomplete | None
attributeDescriptionConstraint: Incomplete | None
uriConstraint: Incomplete | None
attributeSelectorConstraint: Incomplete | None
```

stubs/ldap3/ldap3/utils/asn1.pyi
```diff
@@ -1,37 +1,46 @@
...
+def decode_sequence(message: _B, start, stop, context_decoders: Mapping[int, Callable[[_B, int, int], _R]]) -> _R: ...
+@overload
+def decode_sequence(message: _SupportsGetItemBuffer, start, stop, context_decoders: None = ...) -> _AllDecodersReturnType: ...
+def decode_integer(message, start, stop, context_decoders: _Unused = ...): ...
+def decode_octet_string(message, start, stop, context_decoders: _Unused = ...): ...
+def decode_boolean(message, start, stop, context_decoders: _Unused = ...): ...
+def decode_bind_response(message, start, stop, context_decoders: _Unused = ...): ...
+def decode_extended_response(message, start, stop, context_decoders: _Unused = ...): ...
+def decode_intermediate_response(message, start, stop, context_decoders: _Unused = ...): ...
+def decode_controls(message, start, stop, context_decoders: _Unused = ...): ...
```
AlexWaygood, line 35: `start` and `stop` look like they're always `int`s

```suggestion
def decode_sequence(message: _B, start: int, stop: int, context_decoders: Mapping[int, Callable[[_B, int, int], _R]]) -> _R: ...
@overload
def decode_sequence(message: _SupportsGetItemBuffer, start: int, stop: int, context_decoders: None = ...) -> _AllDecodersReturnType: ...
def decode_integer(message, start: int, stop: int, context_decoders: _Unused = ...): ...
def decode_octet_string(message, start: int, stop: int, context_decoders: _Unused = ...): ...
def decode_boolean(message, start: int, stop: int, context_decoders: _Unused = ...): ...
def decode_bind_response(message, start: int, stop: int, context_decoders: _Unused = ...): ...
def decode_extended_response(message, start: int, stop: int, context_decoders: _Unused = ...): ...
def decode_intermediate_response(message, start: int, stop: int, context_decoders: _Unused = ...): ...
def decode_controls(message, start: int, stop: int, context_decoders: _Unused = ...): ...
```

AlexWaygood (commented):
Thanks! A few more things, but nearly there!

stubs/ldap3/ldap3/extend/novell/getBindDn.pyi
```diff
@@ -1,11 +1,10 @@
...
-
 from ...extend.operation import ExtendedOperation
+from ...protocol.novell import Identity
 
 class GetBindDn(ExtendedOperation):
     request_name: str
     response_name: str
     response_attribute: str
-    asn1_spec: Any
+    asn1_spec: Identity | None
```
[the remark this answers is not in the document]
Avasam: Yep I missed that one

stubs/ldap3/ldap3/protocol/rfc4511.pyi
Avasam: I can remove the comment (explained above), but gotta keep the noqa https://github.com/PyCQA/flake8-pyi/issues/316

AlexWaygood: Yeah, but we now ignore all Y015 errors in typeshed following #9490: https://github.com/python/typeshed/blob/ab1921a737456688b93a43fc7c38021927d0f13c/.flake8#L33

AlexWaygood: Confirmed locally that flake8 still passes with the `# noqa` removed, due to the global ignore.

I guess flake8-noqa isn't complaining because there _is_ a genuine Y015 violation on this line, so the `# noqa` isn't actually _unused_ -- it's just "redundant", since it's already covered by the per-file ignore in the config file.

stubs/ldap3/ldap3/protocol/rfc4511.pyi
Avasam: Source:
https://github.com/cannatag/ldap3/blob/dev/ldap3/protocol/rfc4511.py#L60
```py
# custom constraints
numericOIDConstraint = None  # TODO
distinguishedNameConstraint = None  # TODO
nameComponentConstraint = None  # TODO
attributeDescriptionConstraint = None  # TODO
uriConstraint = None  # TODO
attributeSelectorConstraint = None  # TODO
```

I'm pretty sure they're None because they have yet to add the correct constraint. They are only "used" below in the same module in commented code.

Typing it as None seems better than omitting and adding an allowlist entry: if/when they are implemented, stubsabot can pick up the change.

AlexWaygood: Hmm, okay. It's pretty unusual, though -- can we add a comment to clarify why we're typing it this way? :)

stubs/ldap3/ldap3/protocol/rfc4511.pyi
Avasam: Indeed I pulled main and it works now.

Interesting, I wonder if that's something they could improve or if it's a hard limitation.

AlexWaygood: I haven't looked into flake8-noqa does its magic -- no idea really! 😀

AlexWaygood (approved):
Thanks!

stubs/ldap3/ldap3/protocol/rfc4511.pyi
Avasam: Looks like a known issue: https://github.com/plinss/flake8-noqa/issues/12
I had already upvoted it and didn't even remember ^^"

PR #9471, opened 2023-01-06
Add missing tagMap & typeMap in pyasn1 encoder/decoder

Avasam:
Required for #9470
These are meant to be exported (and are actually used) as a helper mapping of types for specific encoder/decoder

To avoid very large unions and complex overloads, I've kept only the base types.


AlexWaygood:
Oh, I guess stubest didn't flag these as missing because of the fact that the runtime defines `__all__`, and these are omitted from `__all__` at runtime. But if our `ldap3` stubs need them, that's a good enough reason to include them here!

AlexWaygood:
Actually, on second thought, I'm really not sure we should add these symbols to the stub. They're all marked with comments in the source code saying that they're deprecated aliases, and none of them are included in `__all__`. See my comments here: https://github.com/python/typeshed/pull/9470#discussion_r1064014070

PR #2053, opened 2018-04-14
add __fspath__ support to os.path

JelleZijlstra:
Fixes #1997, #2068.

This is tricky because we need to get the return values right (see #1960 for
prior attempts) and we often run into python/mypy#3644. I found that I
could express most signatures correctly using a series of overloads.

A few other changes in here:
- Added splitunc, which according to https://docs.python.org/3/library/os.path.html
  should exist in both Unix and Windows.
- Made the second argument to os.path.curdir Optional to match the implementation.
- Fixed os.path.split, whose previous Path-aware signature triggered python/mypy#3644.

I couldn't come up with a correct signature for `os.path.commonpath`, so I am falling back to using Any as its return type.

I used the following test program to make sure mypy accepted the signatures used here:

```python
import os
from os.path import abspath, commonpath, relpath, split, join
from typing import Generic, AnyStr, Any

class PL(os.PathLike[AnyStr]):
    def __init__(self, path: AnyStr) -> None:
        self.path: AnyStr = path
    def __fspath__(self) -> AnyStr:
        return self.path

bad_path: PL[Any] = PL('')

reveal_type(abspath('hello'))  # str
reveal_type(abspath(b'hello'))  # bytes
reveal_type(abspath(PL('hello')))  # str
reveal_type(abspath(PL(b'hello')))  # bytes

reveal_type(commonpath(['hello', PL('hello')]))  # Any (couldn't get it work correctly)
reveal_type(commonpath([b'hello', PL(b'hello')]))  # Any

reveal_type(relpath('hello'))  # str
reveal_type(relpath('hello', PL('hello')))  # str
reveal_type(relpath(b'hello', b'hello'))  # bytes
reveal_type(relpath(PL('hello')))  # str
reveal_type(relpath(PL(b'hello'), b'hello'))  # bytes
reveal_type(relpath(bad_path, 'hello'))  # str
reveal_type(relpath(bad_path, bad_path))  # Any

reveal_type(split('hello'))  # (str, str)
reveal_type(split(b'hello'))  # (bytes, bytes)
reveal_type(split(PL('hello')))  # (str, str)
reveal_type(split(PL(b'hello')))  # (bytes, bytes)

reveal_type(join('hello', PL('hello')))  # str
reveal_type(join(PL('hello'), 'hello'))  # str
reveal_type(join(b'hello', PL(b'hello')))  # bytes
reveal_type(join(b'hello', PL('hello')))  # error
```

I made the same changes to the Python 2 stub in order to keep the two consistent.

PR #9474, opened 2023-01-08
Replace overloads in `os.path` stubs with Unions of TypeVars

FichteFoll:
Support for this has been added in a recently merged mypy PR (https://github.com/python/mypy/pull/14396) and this commit makes full use of it, getting rid of all overloads in the `os.path` stubs.

Continues the efforts of https://github.com/python/typeshed/pull/2053 (and the previous attempts referenced there).

Introduces a new TypeAlias `GenericOrLiteralPath` and replaces  functions of the following styles:

```py
@overload
def a(path: PathLike[AnyStr]) -> AnyStr: ...
@overload
def a(path: AnyStr) -> AnyStr: ...

@overload
def b(path: PathLike[AnyStr]) -> AnyStr: ...
@overload
def b(path: AnyOrLiteralStr) -> AnyOrLiteralStr: ...
```

with:

```py
def a(path: GenericPath[AnyStr]) -> AnyStr: ...
def b(path: GenericPath[AnyOrLiteralStr]) -> AnyOrLiteralStr: ...
```

Note that quite unexpectedly running `mypy_test.py` does not result in a failure on the latest stable locally and I don't know why (at least not at this time of day), so running against CI to find out more.

Related: https://github.com/python/mypy/issues/12554


hauntsaninja (commented):
Note that mypy doesn't support LiteralString (it currently mostly just treats it like str), so PRs like this will need careful testing with pyright

FichteFoll:
Fails in pyright. I'll take a look tomorrow on how to setup a pyright environment for me to test locally.
https://github.com/python/typeshed/actions/runs/3864921546/jobs/6588076084#step:7:30

FichteFoll:
Problems have been addressed. 

What remains to be tested is how/whether a `LiteralString` is properly transferred into the return type because the `AnyStr` (or `AnyStr_co`) type var(s) used in `GenericPath` are restricted to `str` and `bytes` respectively, but from shallow thinking a `LiteralString` should satisfy this.

FichteFoll:
None of the things noted by mypy_primer seem to be relevant to the changes in this PR.

JelleZijlstra:
The streamlit and pytorch changes are definitely not unrelated: they're the direct result of changes made by this PR. What seems to be going on is that mypy now infers Any for cases where it previously inferred a non-Any type, which means that this PR would cause a regression.

FichteFoll:
Right, it appears to be the same issue as with werkzeug where the typevar is infered to `Any`.

Unfortunately I have no idea to time to figure out how to proceed with this, for the moment at least.

AlexWaygood:
This has a lot of merge conflicts now, and it doesn't look like maintainers have been given permission to push to the PR branch, so I can't resolve them. As such, I'm closing this as stale for now -- but feel free to open a new PR if you'd like to fix the conflicts and see what `mypy_primer` looks like now that mypy 1.1.1 is out.

PR #9475, opened 2023-01-08
Add `FileDescriptorOrPath` and `Unused` type aliases

### LINKED ISSUE #9297: `StrOrBytesPath | FileDescriptor` alias
Opening an issue so we don't forget.

[Original comment](https://github.com/python/typeshed/pull/9279#discussion_r1033813535) from @AlexWaygood :
> We could actually consider adding this alias to `_typeshed`; there's a bunch of functions in `os/__init__.pyi` that could use it. `os/__init__.pyi` currently defines an `_FdOrAnyPath` alias. We don't need to add the alias to `_typeshed` in this PR, though.

There's a handful of `_OpenFile` aliases also defined in multiple stubs

There's also a lot of `int | StrOrBytesPath`, `StrOrBytesPath | int` and `StrOrBytesPath | FileDescriptor` which could be replaced by the alias.

Avasam:
Closes #9297 and see https://github.com/python/typeshed/pull/9470#discussion_r1063745593

I found myself using a `_Unused` TypeAlias a lot for semantic reasons.

Of course they can't be used everywhere until the next mypy update.

### DIFF
--- a/stdlib/_typeshed/__init__.pyi
+++ b/stdlib/_typeshed/__init__.pyi
@@ -36,6 +36,9 @@ AnyStr_co = TypeVar("AnyStr_co", str, bytes, covariant=True)  # noqa: Y001
 # "Incomplete | None" instead of "Any | None".
 Incomplete: TypeAlias = Any
 
+# To describe a method parameter that is unused and will work with anything.
+Unused: TypeAlias = object
+
 # stable
 class IdentityFunction(Protocol):
     def __call__(self, __x: _T) -> _T: ...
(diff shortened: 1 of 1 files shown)


stdlib/_typeshed/__init__.pyi
```diff
@@ -36,6 +36,9 @@ AnyStr_co = TypeVar("AnyStr_co", str, bytes, covariant=True)  # noqa: Y001
 # "Incomplete | None" instead of "Any | None".
 Incomplete: TypeAlias = Any
 
+# To describe a method parameter that is unused and will work with anything.
```
AlexWaygood, line 39:
```suggestion
# To describe a function parameter that is unused and will work with anything.
```

AlexWaygood (approved):
Thanks!

AlexWaygood:
> Of course they can't be used everywhere until the next mypy update.

(stdlib usage is fine now, but, as you say, we'll have to hold off on third-party usage until mypy 1.0 comes along :)

Issue #9465, opened 2023-01-06
Reason for `re.Match.group` and `re.Match.__getitem__` returning `AnyStr | Any`?


willfrey:
Is there a reason why the return annotation is `AnyStr | Any` instead of `AnyStr | None` in the snippets below?

https://github.com/python/typeshed/blob/b1cb9c8a8fde1552b75dc17d4d9612905f6766d4/stdlib/re.pyi#L72-L77

https://github.com/python/typeshed/blob/b1cb9c8a8fde1552b75dc17d4d9612905f6766d4/stdlib/re.pyi#L96-L99

If not, I'd like to submit a PR to change the return type hints accordingly.

Thank you!

AlexWaygood:
There's been a lot of discussion of the signature of `Match.group()` over the years:

- #3171
- #3172
- #3190
- #3902
- #5557

Trying to piece together a precise history at this point is quite honestly a little difficult!

AlexWaygood:
Well, yup, looks like `AnyStr | Any` is still the best we can do here for now! https://github.com/python/typeshed/pull/9482#issuecomment-1375989845

AlexWaygood:
(Thanks for the PR, by the way! It's always good to check whether these decisions made years ago still apply. Type checkers are always getting better with these things, and experimenting is definitely encouraged in this repo :)

willfrey:
Thanks! I get annoyed with `Any` creeping its way into my code, so I try to stamp that out when I can. Problem is, I _still_ don't have a good intuition of when I should use `cast()` vs. an explicit type annotation for a variable vs. some sort of assertion to narrow the inferred type, especially as it pertains when something has type `Any` or some union with `Any`.

Taking an example right from the `re` docs,

```python3
import re

m = re.match(r"(\w+) (\w+)", "Isaac Newton, physicist")
assert m is not None
first_name = m[1]  # Has type `str | Any`

# Do this?
g1: str = m[1]

# Or this?
g1 = cast(str, m[1])

# Or this?
g1 = m[1]
assert isinstance(g1, str)
```

The first two make `disallow-any-expr` directives happy whereas the third one does not.

@AlexWaygood Do you have any personal philosophy of when to use what? If so, I'd love to hear it!

hauntsaninja:
The first is more type safe, `cast` will let you lie

willfrey:
@hauntsaninja Yeah, agreed about `cast`. It's the BFG of shutting up the type checker.

[...]

PR #9482, opened 2023-01-09
Change Any -> None in `re.Match` group-related methods

willfrey:
This is to see how `mypy_primer` feels about these changes in 2023, as per @AlexWaygood's suggestion in #9465.


AlexWaygood:
That looks suboptimal.

willfrey:
@AlexWaygood I concur 😆

PR #9484, opened 2023-01-09
Add stubs for `zstd`

### LINKED ISSUE #9483: Stubs for zstd
[zstd](https://github.com/sergey-dryabzhinsky/python-zstd) is an extension library written in C for Python. I'd like to add stub files for it. However, I don't know whether I should make a pull request on their repo or add them to typeshed.

I'd like to make a PR on their repo, the issue is I don't know how to provide stub files along with the module itself. I tried to add the stub files in the `src/` directory, both as `zstd.pyi` and `__init__.pyi`, then I tried to add these files to the `package_data` section in the `setup.py` file. Nonetheless, the stubs seem to get ignored in all cases; testing on a project I have which depends on zstd, I keep getting:

```
error: Cannot find implementation or library stub for module named "zstd"  [import]
```

Is it even possible to ship those .pyi directly with the package? If so, how can I do so? Or should I just add them to typeshed anyway? Thanks!

victorbnl:
Based on [`zstd`'s documentation](https://github.com/sergey-dryabzhinsky/python-zstd#api)

Closes #9483


stubs/zstd/zstd.pyi
```diff
@@ -0,0 +1,13 @@
+class Error(Exception): ...
+
+def ZSTD_compress(data: str | bytes, level: int | None, threads: int | None) -> str | bytes: ...
```
JelleZijlstra, line 3: If I remember my C argument codes correctly, the first arg should be `_typeshed.ReadableBuffer` (https://github.com/sergey-dryabzhinsky/python-zstd/blob/master/src/python-zstd.c#L68 - I think y# is buffers). Also, the return type appears to be always `bytes`.

(We don't care about Python 2 in typeshed.)

victorbnl: I just tested and you seem to be right, I change it

JelleZijlstra: Thanks! I haven't looked at the other functions but I suspect the same things apply elsewhere. A few functions appear to be aliases for this one.

victorbnl: Yep there are a lot of aliases, namely

- **ZSTD_compress**
  - compress
  - dumps
- **ZSTD_uncompress**
  - decompress
  - uncompress
  - loads
- **ZSTD_version**
  - version

pyrightconfig.stricter.json
```diff
@@ -73,6 +73,7 @@
         "stubs/ttkthemes",
         "stubs/urllib3",
         "stubs/vobject",
+        "stubs/zstd",
```
JelleZijlstra, line 76: Stubs should be added here if they *fail* the stricter checks. I think the current stubs will pass the stricter checks, so we should remove this line.
```suggestion
```

victorbnl: It does indeed. If it is so, maybe the `create_baseline_stubs.py` script should be adapted to make it clearer. By default, it adds this line. Maybe it would be relevant at least to point out in the output this line should be deleted manually when the user finished editing the stubs.

stubs/zstd/zstd.pyi
```diff
@@ -0,0 +1,15 @@
+from _typeshed import ReadableBuffer
+
+class Error(Exception): ...
+
+def ZSTD_compress(data: ReadableBuffer, level: int | None, threads: int | None) -> bytes: ...
```
JelleZijlstra, line 5:
```suggestion
def ZSTD_compress(__data: ReadableBuffer, __level: int = ..., __threads: int = ...) -> bytes: ...
```
All arguments are positional-only since the code uses PyArg_ParseTuple, which works only with positional args. (This also applies to the other functions.)

Also, the last two arguments are optional but do not allow None.

stubs/zstd/zstd.pyi
```diff
@@ -0,0 +1,15 @@
...
+def ZSTD_external() -> int: ...
+def ZSTD_uncompress(data: ReadableBuffer) -> bytes: ...
```
JelleZijlstra, line 7: This one is also positional-only.

PR #9486, opened 2023-01-10
pytype_test: Mark typeshed-external dependencies as missing.

rchen152:


srittau (approved):
LGTM (once the new pytype release is out).

srittau:
I think adding `types-setuptools` to `requirements-tests.txt` should fix the build error.

tests/pytype_test.py
```diff
@@ -131,14 +135,43 @@ def find_stubs_in_paths(paths: Sequence[str]) -> list[str]:
...
+    for fi in files_to_test:
+        parts = fi.split(os.sep)
+        try:
+            idx = parts.index("stubs")
+        except ValueError:
+            continue
+        stub_distributions.add(parts[idx + 1])
+    missing_modules = set()
+    for distribution in stub_distributions:
+        for pkg in utils.read_dependencies(distribution).external_pkgs:
```
AlexWaygood, line 155: I think you need to do a little bit of parsing here, since the strings returned by `read_dependencies(distribution).external_pkgs` might include version constraints (e.g. https://github.com/python/typeshed/pull/9459/files#diff-a8f442db0372b4c61c228311fea276c35a7350c083cfcc08a32d2b22164ecee7R3).

(You'll need to add `from packaging.requirements import Retirement` to the top of the file.)

After this PR is merged, I'll try to file a PR to clarify exactly what's returned by `read_dependencies(distribution)`, since this does seem like a bit of a "gotcha" at the moment :)

```suggestion
        for req in utils.read_dependencies(distribution).external_pkgs:
            pkg = Requirement(req).name
```

AlexWaygood (approved):
Looks great to me! (Ignore the bogus comment I deleted)

tests/pytype_test.py
```diff
@@ -131,14 +135,43 @@ def find_stubs_in_paths(paths: Sequence[str]) -> list[str]:
...
+            # See https://stackoverflow.com/a/54853084.
```
Avasam, line 156: Tiny nit to fix the link on github
```suggestion
            # See https://stackoverflow.com/a/54853084
```

Avasam:
https://github.com/python/typeshed/pull/9459#issuecomment-1371551013
> That seems a little challenging to set up. The simplest (well, definitely the quickest) thing to do is probably just to mark those dependencies as missing so that pytype treats them as having default stubs

I think this is great to unblock us right now. I'll just note that this means that, at the moment, stubs-only packages (or top-level c-modules described by `.pyi` files) that would otherwise work will be considered "missing" by pytype (for example: `lxml-stubs`).

AlexWaygood:
Thanks so much for the quick response, as ever, @rchen152!

PR #9492, opened 2023-01-11
Fix Any subclassing in `SQLAlchemy`

Avasam:
Fixes 10 Any subclassing issues:
- The same 3 in SQLAlchemy, Flask-Migrate and Flask-SQLAlchemy
- Removed `_AsyncIoGreenlet` wich, looking at source code and from the name of the class and module, seems to be implementation details

Ref: #9491, #9493


Avasam:
> The `sqlalchemy-stubs` package is an alternative to this package and also includes a mypy plugin for more precise types.

How much of an "alternative" is `types-AQLAlchemy` to `sqlalchemy-stubs`? Is the latter too reliant on mypy-plugin? Or is the former only there so we could reference it in `Flask-SQLAlchemy` ?

AlexWaygood:
I'd prefer us to find another solution here. It feels really weird for us to declare a dependency on mypy so that we can provide accurate stubs for a directory that only exists to provide a mypy plugin for a completely separate set of SQLAlchemy stubs.

Avasam:
Come to think of it, weird that mypy_test can't find mypy. We shouldn't need to depend on it at all. I guess it's because of the virtual environment being created.

This should be better now :)

Avasam:
Just as a discussion point. Since SQLAlchemy 2.0 is typed. And sqlalchemy2-stubs is actively being developed for 1.4 by the same organization (replacing dropbox's sqlalchemy-stub). And we can now reference external stubs.
Does typeshed plan on continuing to maintain types-SQLAlchemy once those come out of pre-release?

srittau:
sqlalchemy2-stubs has some shortcomings, mostly around the fact that it requires a mypy plugin. I had some more plans with our stubs, but we will most likely drop the stubs some point after SQLAlchemy 2 comes out, whenever that is. That said, I'm opposed to depend on sqlalchemy2-stubs.

AlexWaygood:
@srittau, do you think we actually need the `ext/mypy` directory? It feels weird having stubs for a mypy plugin that's only there to support another set of stubs. Could we just delete the directory?

srittau:
@AlexWaygood Sure. I think we have too many internal modules in our stubs anyway.

AlexWaygood (approved):
Thank you!

Avasam:
> @AlexWaygood Sure. I think we have too many internal modules in our stubs anyway.

Gotta love fixing things by just removing code !

AlexWaygood:
> > @AlexWaygood Sure. I think we have too many internal modules in our stubs anyway.
> 
> Gotta love fixing things by just removing code !

Fixed a bug by removing the feature ;)

srittau (approved):

PR #9494, opened 2023-01-11
Use of twisted in pika

Avasam:
Fixes 2 Any subclassing issues

~~This would require the stub_uploader to allow twisted~~
The twisted types have been replicated.

I've completed as many things referencing twisted as I could find.

Ref: #9491


AlexWaygood:
I don't think there's enough of a win here for this to be worth it. `twisted` is just an optional backend for `pika`; if somebody decides to use `tornado` as a backend instead, then the extra dependency is pointless. And even if they _do_ use `twisted`, the `twisted` types still don't seem fundamental to the package, really.

Just because we _can_ have non-types dependencies now doesn't mean we should add them everywhere. We should still strive for stubs packages to have minimal dependencies, and present minimal security risks, wherever possible.

Avasam:
So it's an optional dependency just like Qt is to Pillow.

> I don't think there's enough of a win here for this to be worth it.
> […] the twisted types still don't seem fundamental to the package, really

So would `Any` subclassing be acceptable here?
Or if its not too complex, I can look into replicating the base class(es) is the stub.

AlexWaygood:
> So would `Any` subclassing be acceptable here?
> Or if its not too complex, I can look into replicating the base class(es) is the stub.

`Any` subclassing is never _good_, so if you'd like to try mimicking the shape using protocols or whatever, sure 😀

stubs/pika/@tests/stubtest_allowlist.txt
```diff
@@ -3,9 +3,8 @@ pika.compat.StringIO.seek
 pika.compat.StringIO.truncate
 
 # Requires external libraries to be installed.
-pika\.adapters\.gevent_connection.*
-pika\.adapters\.tornado_connection.*
-pika\.adapters\.twisted_connection.*
+pika.adapters.gevent_connection
```
AlexWaygood, line 6: Any reason not to add `gevent` to `requirements-stubtest.txt` too, and get some stubtest coverage for `gevent_connection.pyi`?

Avasam: afaik gevent is not type. But maybe that doesn't matter for stubtest?
I know we don't use untyped `requires`. But if it's fine for stubtest so it can scan a file then I'll add it.

AlexWaygood: Stubtest attempts to dynamically import every module it can find at runtime. But it currently can't import `pika.adapters.gevent_connection`, because `gevent_connection.py` has `import gevent` at the top of the file. The requirements in `requirements-stubtest.txt` don't need to be typed; they're just there to help stubtest out in situations like this, where doing `pip install pika` isn't sufficient for stubtest to be able to do its dynamic runtime inspection.

(Arguably `pika` should probably specify a `gevent` "extra" at runtime, so that we could tell stubtest to do `pip install pika[gevent]`, rather than hacking about with extra `requirements-stubtes.txt` files. But it is what it is.)

stubs/pika/@tests/requirements-stubtest.txt
```diff
@@ -0,0 +1,2 @@
+twisted
```
Avasam, line 1:
```suggestion
twisted
gevent
```

stubs/pika/@tests/stubtest_allowlist.txt
Avasam, line 6:
```suggestion
```

stubs/pika/@tests/stubtest_allowlist.txt
```diff
@@ -3,9 +3,7 @@ pika.compat.StringIO.seek
 pika.compat.StringIO.truncate
 
 # Requires external libraries to be installed.
```
AlexWaygood, line 5:
```suggestion
# Fake module so we don't have to declare a dependency on `twisted` just for an optional backend
```

JelleZijlstra:
I appreciate the work here but I'm wary of adding so much code that we can't validate with stubtest or other automatic tools.

It might make sense to import the types from Twisted without declaring an explicit dependency on Twisted for the stubs package. Users who use the Twisted backend presumably will also have Twisted installed in their venv, so they will get precise types. If they don't, type checkers will just fall back to Any.

Avasam:
> It might make sense to import the types from Twisted without declaring an explicit dependency on Twisted for the stubs package.

That's something I've suggested a few times for different stubs with similar issues. (the idea of "optional" or "test only" dependencies). The issue with the current setup is that we can do that for stubtest, but not for every other tests like mypy, pyright, pytype, flake8, ... Especially with Pyright you can end up with unknowns that bubble up and you have to `# pyright: ignore` a bunch of stuff.

I'll see what I can do for Twisted in Pika, it's been a while I haven't touched this PR.

Avasam:
Alright, this is what it looks like in pika. Not too bad given that usage of `twisted` is restricted to the `twisted_connection` module.
The same could be done for `tornado_connection` and `gevent_connection`.

AlexWaygood (approved):
It's not exactly _pretty_, but I'm on board with this new approach!

JelleZijlstra (approved):

PR #9495, opened 2023-01-11
Fix Any subclassing in pyinstaller

Avasam:
This one was easy to fix without having to resort to external requires. Plus it's Python 3.7 only

Ref: #9491


AlexWaygood (approved):
Nice!

PR #9496, opened 2023-01-11
Add comment explaining NonCallableMock's Any subclassing

Avasam:
Brought back and reworded a 5yo comment explaining why `NonCallableMock` is subclassed with `Any`.

Feel free to wordsmith.

Ref: #9491


srittau:
The problem here are not so much the false positives, but that mocks are explicitly designed to stand in for other types, something that can't be expressed with our static type system.

AlexWaygood:
I'd also rather avoid introducing unused `type: ignore`s -- I'd rather introduce `type: ignore`s for the false-positives all together as part of #9491.

Avasam:
Closing in favor of #9491

PR #9497, opened 2023-01-11
Delete stubs/SQLAlchemy/sqlalchemy/ext/mypy directory

AlexWaygood:
This directory is a mypy plugin at runtime, and the plugin is only useful if you're using the separate stubs-sqlalchemy project instead of typeshed's stubs for sqlalchemy. Moreover, without declaring a dependency on mypy for the stubs, this directory is hard to type with any degree of precision


AlexWaygood:
Closing in favour of #9492

PR #9498, opened 2023-01-11
SQLAlchemy: Annotate text()

srittau:


srittau:
Depends on #9490.

srittau:
Concerning the primer output:

* pandas does `stmt = text(stmt)`, which reassigns a different type, so the new warning is genuine.
* The other new warnings are genuine, although technically unrelated to this PR. But I will fix them.

AlexWaygood (approved):
Wow, the `_generative` decorator seems almost like it's deliberately designed to confuse stubgen

PR #9499, opened 2023-01-11
Allow stubs to depend on `numpy`

AlexWaygood:
Currently, if a stubs package declares a dependency on `numpy`, the latest version of `numpy` will be installed in CI when testing that package. However, the CI will then fail, because mypy will detect that `numpy` is using PEP-570 syntax, and accordingly will emit errors when run with the `--python-version=3.7` flag, since PEP 570 is new in Python 3.8.

The solution is to actually run mypy _on_ 3.7 in CI, rather than simply running mypy with the `--python-version=3.7` flag. If we run mypy _on_ 3.7, pip's dependency resolver will understand that the latest version of `numpy` isn't compatible with the version of Python we're using, and will give us an older version of `numpy` that's compatible with Python 3.7.

A few tweaks to `tests/mypy_test.py` and `tests/utils.py` are required in order to make them compatible with Python 3.7. We also can't install the whole of `requirements-tests.txt` on Python 3.7, since `flake8==6.0.0` is not compatible with Python 3.7.

Helps unblock #8974

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -91,10 +91,10 @@ jobs:
       - uses: actions/checkout@v3
       - uses: actions/setup-python@v4
         with:
-          python-version: "3.10"
+          python-version: ${{ matrix.python-version }}
           cache: pip
           cache-dependency-path: requirements-tests.txt
-      - run: pip install -r requirements-tests.txt
+      - run: pip install $(grep mypy== requirements-tests.txt) packaging pathspec termcolor tomli typing-extensions
       - run: python ./tests/mypy_test.py --platform=${{ matrix.platform }} --python-version=${{ matrix.python-version }}
 
   regression-tests:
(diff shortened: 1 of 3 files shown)


hauntsaninja:
Hmm, not sure we should do this. numpy types have evolved a lot in recent versions

AlexWaygood:
> Hmm, not sure we should do this. numpy types have evolved a lot in recent versions

An alternative solution might be to add a `requires_python` field to the `METADATA.toml` files. We could use that to declare that e.g. `types-tensorflow` requires Python 3.8+, and also skip testing those stubs when running mypy with `--python-version=3.7`.

hauntsaninja:
Actually, this is fine, since we'll test with new numpy on new Python. (assuming the (lack of) improvements in a numpy don't cause trouble for us)

hauntsaninja (approved):
Looks good!

PR #9501, opened 2023-01-11
stdlib: add defaults

JelleZijlstra:
Ran https://github.com/JelleZijlstra/stubdefaulter on the stdlib


JelleZijlstra:
To keep this PR manageable, I'm going to fix any issues in CI by just reverting the affected functions.

Notes on CI:

- stubtest failure because stubtest doesn't see a str as compatible with a TypeVar(bound=str) in urllib.parse
- urllib.parse.urlencode also has a default of None where our type is str, needs investigation; same for sysconfig.is_python_build, platform.libc_ver, ssl._create_unverified_context
- and shutil.make_archive but with 0 vs. False; same for platform.platform, distutils.util.execute, distutils.util.byte_compile, distutils.spawn, and a bunch of others
- flake8-pyi seems to think string defaults are annotations, so it will need a fix

JelleZijlstra:
We also got a pytype crash

```
  File "/opt/hostedtoolcache/Python/3.10.9/x64/lib/python3.10/site-packages/pytype/pyi/visitor.py", line 36, in _call_visitor
    return super()._call_visitor(node)
  File "/opt/hostedtoolcache/Python/3.10.9/x64/lib/python3.10/site-packages/pytype/ast/visitor.py", line 55, in _call_visitor
    return visitor(node)
  File "/opt/hostedtoolcache/Python/3.10.9/x64/lib/python3.10/site-packages/pytype/pyi/parser.py", line [16](https://github.com/python/typeshed/actions/runs/3896137144/jobs/6652302058#step:7:17)2, in visit_Pyval
    return self.convert_late_annotation(node.value)
  File "/opt/hostedtoolcache/Python/3.10.9/x64/lib/python3.10/site-packages/pytype/pyi/parser.py", line 142, in convert_late_annotation
    typ = a.body[0].value  # pytype: disable=attribute-error
IndexError: list index out of range
```

I suspect like flake8-pyi it's something to do with treating string defaults as annotations but I'll have to look at the code.

hauntsaninja:
Just checking, you ran latest stubtest locally with the default checking you added?

JelleZijlstra:
Not yet, I forgot it hadn't been released yet.

hauntsaninja:
We've often used pre-release stubtest in CI. Might be a good idea to do that now too!

JelleZijlstra:
For reference fdc5863337c9b5988c33ddb7a27e4a2e066a10ad is where we undid stubtest-from-master most recently, but some of the infra has changed since then so it's not just a matter of reverting and updating the commit.

JelleZijlstra:
New version with the latest stubdefaulter:

- Now only adds defaults for 3.11 branches
- Now also adds defaults to methods
- Undid a bunch of things stubtest complained about

The PR currently doesn't include a change to the version of stubtest. If we want to run stubtest from mypy master, I'd rather do that in a separate PR.

srittau (approved):
A quick glance shows no obvious problems. But I mostly rely on our tests here.

PR #9505, opened 2023-01-12
Fix some Any subclassing in tqdm

Avasam:
Fixes ~~4~~ 2 Any subclassing issues by replicating the types used in external modules.
~~The `rich` one isn't complete as it was starting to get quite complex (I don't wanna re-implement the entire library's types either!)~~

Ref: #9491


stubs/tqdm/tqdm/keras.pyi
```diff
@@ -1,10 +1,31 @@
 from _typeshed import Incomplete
-from typing import Any
-from typing_extensions import TypeAlias
 
 __all__ = ["TqdmCallback"]
 
-_Callback: TypeAlias = Any  # Actually tensorflow.keras.callbacks.Callback
+# keras.callbacks.Callback
```
AlexWaygood, line 5:
```suggestion
# tensorflow.python.keras.callbacks.Callback
```

AlexWaygood (commented):
The `keras` and `dask` changes look great. Mind splitting the `rich` changes off into a separate PR? They're a little more complex and my brain's going a little fuzzy thinking about nominal vs structural subtyping 😅

stubs/tqdm/tqdm/keras.pyi
Avasam: Weirdly enough, it's actually using keras, and if that fails, it fallsback to tensorflow... which uses keras (probably does some extra in that case idk)

AlexWaygood: Oh good point, ignore me!

Avasam:
> The `keras` and `dask` changes look great. Mind splitting the `rich` changes off into a separate PR? They're a little more complex and my brain's going a little fuzzy thinking about nominal vs structural subtyping 😅

If it feels too complex for our needs, I can always make more things `Incomplete`. We mostly need the structure, detailed types are bonuses.

AlexWaygood (approved):
Thanks!

AlexWaygood:
> > The `keras` and `dask` changes look great. Mind splitting the `rich` changes off into a separate PR? They're a little more complex and my brain's going a little fuzzy thinking about nominal vs structural subtyping 😅
> 
> If it feels too complex for our needs, I can always make more things `Incomplete`. We mostly need the structure just to fix the Any subclassing, detailed types are bonuses.

Maybe we could start off with something simpler for now? We can always refine it in the future :)

PR #9506, opened 2023-01-12
urllib.parse.urlencode: encoding and errors can be None

JelleZijlstra:
The runtime defaults are None. Found in #9501.


hauntsaninja (approved):
This lgtm, but why didn't stubtest catch this

JelleZijlstra:
It does on master (see the latest run in my defaults PRs).

hauntsaninja:
Weird, bisects to https://github.com/python/mypy/pull/14135

hauntsaninja:
Not clear to me why, but I guess I'll take it

PR #9507, opened 2023-01-12
ssl, sysconfig: fix issues with defaults

JelleZijlstra:
- ssl._create_unverified_context allows None since 3.10:
  https://github.com/python/cpython/commit/2875c603b2a7691b55c2046aca54831c91efda8e#diff-f6439be9c66350dde4c35dbeea0352c96cc970ba12b0478f6ae36f10725bd8c5
- sysconfig.is_python_build ignores its argument since 3.11:
  https://github.com/python/cpython/commit/067597522a9002f3b8aff7f46033f10acb2381e4
  (was backported into 3.11)


hauntsaninja (approved):
Looks good, although I'd maybe type it as `check_home: None = None`

Avasam:
> Looks good, although I'd maybe type it as `check_home: None = None`

Does it still work if you give it anything? If it's just unused (but warns because it's deprecated if you pass anything), we recently added an Alias to represent that semantic:
`Unused: TypeAlias = object`

hauntsaninja:
No, it will complain if you pass something (non-None), which I find desirable in this case.

AlexWaygood (approved):
LGTM too, though I agree with @hauntsaninja's point about `is_python_build` -- having a type checker warn if you pass anything not-`None` on Python >=3.10 would be useful.

JelleZijlstra:
I'd prefer to stick with the current annotation until we have PEP 702 available. With PEP 702, I would type it as
```
@overload
@deprecated("check_home argument is deprecated")
def is_python_build(check_home: object) -> bool: ...
@overload
def is_python_build() -> bool: ...
```

Setting the type as None feels wrong because it's basically an accident that None doesn't raise the deprecation warning; the point of the warning is that the parameter is going to be removed, not that None will be the only valid type in the future.

AlexWaygood:
> Setting the type as None feels wrong because it's basically an accident that None doesn't raise the deprecation warning; the point of the warning is that the parameter is going to be removed, not that None will be the only valid type in the future.

I don't have a strong opinion, but "getting a warning if you pass anything not-`None`" feels slightly better than "never getting a warning".

But I agree that it's not a particularly principled way of doing things.

PR #9509, opened 2023-01-12
subprocess: universal_newlines may be None

JelleZijlstra:
All the way back to 3.7:
https://github.com/python/cpython/blob/3.7/Lib/subprocess.py#L684

It's treated like False


hauntsaninja (approved):
Thanks, looking at that PR I bisected, I think it's to do with a bug that it fixes regarding value constrained type variables

PR #9510, opened 2023-01-12
Use environment marker to restrict flake8 install

Avasam:
From https://github.com/python/typeshed/pull/8974#discussion_r1067873831

Use of environment marker to indicate that a dependency isn't usable with Python 3.7

- More obvious than grepping
- Reduces duplication
- Keeps the CI the same as others
- Works locally
- pip will event log why it's not installing flake8

### DIFF
--- a/.github/workflows/stubtest_stdlib.yml
+++ b/.github/workflows/stubtest_stdlib.yml
@@ -44,6 +44,6 @@ jobs:
           cache: pip
           cache-dependency-path: requirements-tests.txt
       - name: Install dependencies
-        run: pip install $(grep mypy== requirements-tests.txt)
+        run: pip install -r requirements-tests.txt
       - name: Run stubtest
         run: python tests/stubtest_stdlib.py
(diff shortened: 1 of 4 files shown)


.github/workflows/stubtest_stdlib.yml
```diff
@@ -44,6 +44,6 @@ jobs:
           cache: pip
           cache-dependency-path: requirements-tests.txt
       - name: Install dependencies
-        run: pip install $(grep mypy== requirements-tests.txt)
+        run: pip install -r requirements-tests.txt
```
Avasam, line 47: Arguably only installing the 1-2 dependency we need for a single test doesn't sound like a bad idea either.

AlexWaygood: Ah, it would be good to keep stubtest stdlib ones as they are. Unlike stubtest third-party, it doesn't create venvs on the fly, and it's useful to be 100% sure that it's running in as isolated an environment as possible

Avasam: But keep `-r requirements-tests.txt` fir third party right?

AlexWaygood: Yup, the third-party one depends on `utils.py`, and we can be sure it's testing packages in isolation because of the way it creates venvs on the fly 👍

AlexWaygood (approved):
LGTM. Thanks!

PR #9487, opened 2023-01-10
Additional type hints added to stub `openpyxl`

ArnabRollin:
### Reason for change
The VS Code Python static type-checker does not recognise

Here is an example:
```python
from openpyxl import load_workbook

wb = load_workbook("myxl.xlsx") # Error: Function 'load_workbook' is partially unknown
```
Mypy doesn't seem to find the `types-openpyxl` package

```console
(Working-with-Excel) -> pipenv install types-openpyxl
```
```
Installing types-openpyxl...
Pipfile.lock (43417e) out of date, updating to (83c742)...
Locking [packages] dependencies...
Locking [dev-packages] dependencies...
Updated Pipfile.lock (c868c9d2bef5b5f5f3c9b17280f5f7a15a9a9522fd50d8cb558d1eb0e083c742)!
Installing dependencies from Pipfile.lock (83c742)...
To activate this project's virtualenv, run pipenv shell.
Alternatively, run a command inside the virtualenv with pipenv run.
```
```console
(Working-with-Excel) -> mypy app.py
```
```
app.py:1: error: Library stubs not installed for "openpyxl"  [import]
app.py:1: note: Hint: "python3 -m pip install types-openpyxl"
app.py:1: note: (or run "mypy --install-types" to install all missing stub packages)
app.py:1: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
```
Please point out if I am wrong anywhere.

PR #9511, opened 2023-01-12
Merge openpyxl stubs from microsoft/python-type-stubs

Avasam:
Bring openpyxl stubs at least up to par with https://github.com/microsoft/python-type-stubs/tree/main/openpyxl

Add class variable types and descriptors for affected classes.
For a bit of context: "descriptor" types are assigned to class variables. Those become property-like with getters that type-check on assignment.
There's also "Convertible" descriptors that do type coercion.

This is not meant to complete the stub, far from there. Only to make it good enough to upstream. And a few other things here and there.

There's overlap with ~~#9487~~ #9764, I'd suggest finishing that PR first, then I'll double check any conflict that arise.
I'll look at primer results after that as well.


stubs/openpyxl/@tests/stubtest_allowlist.txt
```diff
@@ -1,2 +1,429 @@
...
+openpyxl.chart.axis.TextAxis.__init__
+openpyxl.styles.colors.RgbColor.__init__
+openpyxl.styles.named_styles._NamedCellStyle.__init__
+openpyxl.styles.numbers.NumberFormat.__init__
+
+# Leaked into the global space
+openpyxl.utils.cell.col
+openpyxl.utils.cell.i
+
+# Unintended re-exports
```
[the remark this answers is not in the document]
AlexWaygood: Stubtest will have many fewer false-positives regarding re-exports after the next mypy release, following https://github.com/python/mypy/pull/14270 :)

stubs/openpyxl/openpyxl/drawing/image.pyi
```diff
@@ -1,9 +1,14 @@
-from _typeshed import Incomplete
+from typing_extensions import Literal
+
+from PIL import Image as ImageModule
+
+# Using "object" because modules can't by used as type
+PILImage: object | Literal[False]
```
Avasam, line 6: Technically this should be `PILImage: ImageModule | Literal[False]`
But I believe the intent is to use it as a boolean.

stubs/openpyxl/openpyxl/styles/colors.pyi
```diff
@@ -1,60 +1,150 @@
...
 
-COLOR_INDEX: Incomplete
-BLACK: Incomplete
-WHITE: Incomplete
-BLUE: Incomplete
-aRGB_REGEX: Incomplete
+_Unused: TypeAlias = object
+_S = TypeVar("_S", bound=Serialisable)
+
+COLOR_INDEX: tuple[
```
Avasam, line 16: Tempted to disable formatting for this line XD

AlexWaygood:
(Adding DO-NOT-MERGE for now, as I agree we should resolve #9487 first before tackling this :)

stubs/openpyxl/openpyxl/xml/functions.pyi
```diff
@@ -1,6 +1,74 @@
+import sys
```
Avasam, line 1: This file... is a mess...

The lack of import conditionals in type checkers, and/or `AnyOf` really hurts here.

Pyright handles it a bit better than mypy (as per the comment below)

I could extract all functions and manually overload them.

But I still have one big issue with `Element`, which is either a Class (note: a class, not an instance) or a function. Depending on where it's imported from.

stubs/openpyxl/openpyxl/xml/_functions.pyi
```diff
@@ -0,0 +1,60 @@
+# This file does not exist at runtime. It is a helper file to overload imports in openpyxl.xml.functions
+# TODO: Create missing stubs and add missing types upstream
```
Avasam, line 2: eh, maybe not. lxml is fine (as it's what this library was originally built on), but the other two sounds overkill.

Avasam:
In the meantime, @AlexWaygood , can we consider allowing [lxml-stubs](https://pypi.org/project/lxml-stubs/) in the stubuploader?

Also, is there a list somewhere of those "stubs taken out of typeshed because incomplete"? I feel it shouldn't be too hard to add 'em back with modern stubgen.

AlexWaygood:
> In the meantime, @AlexWaygood , can we consider allowing [lxml-stubs](https://pypi.org/project/lxml-stubs/) in the stubuploader?

SGTM but I'm not a stub-uploader maintainer :)

AlexWaygood:
> Also, is there a list somewhere of those "stubs taken out of typeshed because incomplete"? I feel it shouldn't be too hard to add 'em back with modern stubgen.

What's this a reference to, sorry?

Avasam:
> Not sure I understand, sorry -- what's this a reference to?

https://pypi.org/project/lxml-stubs/
> These type annotations were initially included in [typeshed](https://www.github.com/python/typeshed), but lxml's annotations are still incomplete and have therefore been extracted from typeshed to avoid unintentional false positive results.

I feel it's not the first time I see that. Or maybe it is and I just read it twice on the same page.

AlexWaygood:
You'll have to ask one of the other maintainers -- sounds like that all happened before I arrived on the typeshed scene!

AlexWaygood:
> > In the meantime, @AlexWaygood , can we consider allowing [lxml-stubs](https://pypi.org/project/lxml-stubs/) in the stubuploader?
> 
> SGTM but I'm not a stub-uploader maintainer :)

On second thought -- with our current setup, I think you'll be unable to use `lxml-stubs` even if wwe add them to the stub-uploader allowlist :( The current rules on the stub-uploader are that non-types dependencies must be explicitly included in the "non-types allowlist", _and_ also be specified as a requirement of the runtime package. But the runtime package doesn't declare `lxml-stubs` as a dependency, so the stub-uploader tests will fail _even if_ we add `lxml-stubs` in the stub-uploader :/

This is an annoying situation but I'm not sure what the solution might be.

stubs/openpyxl/openpyxl/reader/excel.pyi
```diff
@@ -19,8 +28,10 @@ class ExcelReader:
     def read_workbook(self) -> None: ...
     def read_properties(self) -> None: ...
     def read_theme(self) -> None: ...
-    def read_chartsheet(self, sheet, rel) -> None: ...
+    def read_chartsheet(self, sheet: ChildSheet, rel: Relationship) -> None: ...
     def read_worksheets(self) -> None: ...
     def read(self) -> None: ...
 
-def load_workbook(filename, read_only: bool = ..., keep_vba=..., data_only: bool = ..., keep_links: bool = ...): ...
+def load_workbook(
```
Avasam, line 35: Conflict:
This, and `ExcelReader.__init__` is ultimately passed to `ZipFile` in `_validate_archive`.

Even in read mode it still needs a `name` variable, `seek` and `tell` methods.

stubs/openpyxl/openpyxl/reader/excel.pyi
```diff
@@ -1,16 +1,25 @@
...
-    def __init__(self, fn, read_only: bool = ..., keep_vba=..., data_only: bool = ..., keep_links: bool = ...) -> None: ...
+    archive: ZipFile
+    valid_files: list[str]
+    read_only: bool
+    keep_vba: bool
+    data_only: bool
+    keep_links: bool
+    shared_strings: list[str]
+    def __init__(
+        self, fn: StrPath | IO[bytes], read_only: bool = ..., keep_vba: bool = ..., data_only: bool = ..., keep_links: bool = ...
```
Avasam, line 21: Conflict:
This, and `load_workbook` is ultimately passed to `ZipFile` in `_validate_archive`.

Even in read mode it still needs a `name` variable, `seek` and `tell` methods.

stubs/openpyxl/openpyxl/reader/excel.pyi
AlexWaygood: I'd be okay with switching it to IO[bytes]`. We could probably come up with a more precise protocol, but for now consistency with the stdlib `zipfile` stub is probably the most important thing

stubs/openpyxl/openpyxl/workbook/workbook.pyi
```diff
@@ -20,54 +31,58 @@ class Workbook:
...
+        scope: int | None = ...,
     ) -> None: ...
-    def add_named_style(self, style) -> None: ...
+    def add_named_style(self, style: NamedStyle) -> None: ...
     @property
-    def named_styles(self): ...
-    def get_named_ranges(self): ...
-    def add_named_range(self, named_range) -> None: ...
+    def named_styles(self) -> list[str]: ...
+    def get_named_ranges(self) -> list[DefinedName]: ...
```
Avasam, line 78: Conflict: See https://github.com/python/typeshed/pull/9764#discussion_r1113412107

AlexWaygood: Your rationale in https://github.com/python/typeshed/pull/9764#discussion_r1113412107 is convincing; let's go with a `list`

stubs/openpyxl/openpyxl/worksheet/worksheet.pyi
```diff
@@ -24,95 +37,101 @@ class Worksheet(_WorkbookChild):
...
+    def __init__(self, parent: Workbook | None, title: str | _Decodable | None = ...) -> None: ...
     @property
-    def sheet_view(self): ...
+    def sheet_view(self) -> SheetView: ...
     @property
-    def selected_cell(self): ...
+    def selected_cell(self) -> str: ...
     @property
-    def active_cell(self): ...
+    def active_cell(self) -> str: ...
```
Avasam, line 46: Conflict: As seen by `Selection.__init__`, these are the string representation of cells. Descriptors also force these for be strings.

AlexWaygood: Good point, I agree -- these should both be strings.

stubs/openpyxl/openpyxl/worksheet/worksheet.pyi
```diff
@@ -82,7 +89,7 @@ class Worksheet(_WorkbookChild):
         max_col: int | None = None,
         *,
         values_only: Literal[True],
-    ) -> Generator[tuple[str | float | datetime | None, ...], None, None]: ...
+    ) -> Generator[tuple[_KnownTypes, ...], None, None]: ...
     @overload
     def iter_rows(
```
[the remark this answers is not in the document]
AlexWaygood: > Adding this case would basically double the overloads. I don't necessarily mind it, but is there a way to simplify the existing overloads first?

Not that I'm aware of, sadly...

Avasam:
I've added comments for the less obvious conflicts. Everything else should be self-explanatory (hopefully).
Ready for review!

stubs/openpyxl/openpyxl/xml/_functions.pyi
```diff
@@ -0,0 +1,59 @@
+# This file does not exist at runtime. It is a helper file to overload imports in openpyxl.xml.functions
+# TODO: Add missing types in lxml-stubs and et_xmlfile and figure out how to import lxml-stubs
```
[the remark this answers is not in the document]
Avasam: Got my answer, definitely not in favor of type hints or the type system at all:
https://foss.heptapod.net/openpyxl/openpyxl/-/issues/1553#note_145934
https://foss.heptapod.net/openpyxl/openpyxl/-/issues/1570#note_150282
https://foss.heptapod.net/openpyxl/openpyxl/-/issues/1812#note_183907
https://foss.heptapod.net/openpyxl/openpyxl/-/issues/1902#note_190424
https://foss.heptapod.net/openpyxl/openpyxl/-/issues/1922#note_192543

So that'd be a hard sell. If they don't want to even link to `types-openpyxl`, I doubt they'd add an extra dep to `lxml` if the only "users" asking for it are typeshed contributors.

At least I'll stop entertaining that idea 😅

AlexWaygood (commented):
Thanks for this. Here's a few initial thoughts just from reading through the PR. I haven't really compared the PR with the implementation yet.

In general, I think you're going a _little_ overboard with trying to represent the runtime exactly with `Literal` types. Using very large `Literal`s can often result in unreadable/ugly stubs, but it can also cause performance issues with type checkers. Moreover, if we try to reflect details in the stub which are ultimately implementation details, it can make it harder to maintain for us; but it can also do a disservice to end users of the stub. Users of IDEs might find that hovering over a function generates an unreadable "wall-of-text" function signature in the tooltip that's hard to read. Users of a type checker might come to depend on an implementation detail that we've reproduced exactly in the stub, but that they really shouldn't be depending on at the end of the day.

Sometimes, simpler is better :)

stubs/openpyxl/openpyxl/workbook/workbook.pyi
```diff
@@ -28,49 +30,56 @@ class Workbook:
...
     @property
     def data_only(self) -> bool: ...
     @property
     def write_only(self) -> bool: ...
     @property
-    def excel_base_date(self) -> datetime: ...
+    def excel_base_date(self) -> datetime.datetime: ...
     @property
-    def active(self) -> _WorkbookChild | None: ...
+    def active(self) -> _WorkbookChild: ...
```
mynhardtburger, line 49: The returned object will either be a Worksheet (likely case), WriteOnlyWorksheet or Chartsheet. These are all children of _WorkbookChild, but has many more methods associated.

Is is possible to be more specific in the return type, so that method/property code completion can be more accurate?

Or is it expected that the coder would need to specify the type they expect?

Avasam: The thing about union return types, is that you're stuck doing `isinstance` checks anyway. (see https://github.com/python/mypy/issues/1693 and https://github.com/python/typing/issues/566 ).

But if the known types are limited, I can narrow it down further. If you're only interested in the shared properties/methods. It won't change much.

If it wasn't for the fact that you can use an index to set the active sheet, you could restrict it to `Worksheet` or `Chartsheet` (not `WriteOnlyWorksheet` because it doesn't seem to have a `sheet_state).` But with an index you can set any worksheet as the active one.

Avasam:
Big batch of changes. I've commented out lxml stuff for now. Will see about completely moving it out later (gotta go sleep!).
Btw I haven't validated mypy_primer errors yet either.

JukkaL:
I agree with @AlexWaygood above about literal types. I think that when in doubt, it's usually better to not use a literal type. In my experience users rarely use literal types when annotating code, so literal types in stubs can cause errors that are perceived by users as false positives, even if they are technically valid.

stubs/openpyxl/@tests/stubtest_allowlist.txt
```diff
@@ -1,2 +1,262 @@
...
+openpyxl.compat.numbers.NUMERIC_TYPES
+openpyxl.cell.cell.NUMERIC_TYPES
+openpyxl.xml._functions
+openpyxl.xml.functions.Element.__iter__
+openpyxl.xml.functions.tostring
+openpyxl.xml.functions.xmlfile
+
+# Inconsistent __init__ because
+# - using the default value results in an error because of the runtime type-guards
+# - or, keyword arguments are explicitely specified
```
AlexWaygood, line 13: Typo :)

```suggestion
# - or, keyword arguments are explicitly specified
```

stubs/openpyxl/openpyxl/cell/cell.pyi
```diff
@@ -1,42 +1,52 @@
...
-TYPE_FORMULA: str
-TYPE_NUMERIC: str
-TYPE_BOOL: str
-TYPE_NULL: str
-TYPE_INLINE: str
-TYPE_ERROR: str
-TYPE_FORMULA_CACHE_STRING: str
-VALID_TYPES: Incomplete
+TIME_TYPES: tuple[type[_TimeTypes], ...]
+_TimeTypes: TypeAlias = datetime.datetime | datetime.date | datetime.time | datetime.timedelta
```
AlexWaygood, line 15: Nit: it's less confusing for human readers if we avoid the forward reference here

```suggestion
_TimeTypes: TypeAlias = datetime.datetime | datetime.date | datetime.time | datetime.timedelta
TIME_TYPES: tuple[type[_TimeTypes], ...]
```

stubs/openpyxl/openpyxl/cell/cell.pyi
```diff
@@ -1,42 +1,52 @@
...
+TIME_FORMATS: dict[_TimeTypes, str]
```
AlexWaygood, line 16:
```suggestion
TIME_FORMATS: dict[type[_TimeTypes], str]
```

stubs/openpyxl/openpyxl/cell/cell.pyi
```diff
@@ -1,42 +1,52 @@
...
+STRING_TYPES: tuple[type[str], type[bytes]]
+_StringTypes: TypeAlias = str | bytes
+# We could get the exact tuple length and order with the following line, ignoring Y015
+# But pytype doesn't support it and mypy oversimplifies the type. Pyright fully expands it.
+# KNOWN_TYPES = NUMERIC_TYPES + TIME_TYPES + STRING_TYPES + (bool, type(None))
```
AlexWaygood, line 21: I feel like we don't really need this comment -- stubs are meant to be static declarations of types, after all. It's kind of understood that we want to keep dynamic inference of types to a minimum, because that's sort of the whole point of stubs :D

```suggestion
```

stubs/openpyxl/openpyxl/cell/cell.pyi
```diff
@@ -1,42 +1,52 @@
...
+KNOWN_TYPES: tuple[_KnownTypes, ...]
+_KnownTypes: TypeAlias = _NumericTypes | _TimeTypes | _StringTypes | bool | None
```
AlexWaygood, line 23: - Avoid the forward reference
- It's a tuple of _types_, not a tuple of _instances_

```suggestion
_KnownTypes: TypeAlias = _NumericTypes | _TimeTypes | _StringTypes | bool | None
KNOWN_TYPES: tuple[type[_KnownTypes], ...]
```

stubs/openpyxl/openpyxl/compat/numbers.pyi
```diff
@@ -1,4 +1,21 @@
...
+import numpy
+
+if sys.version_info >= (3, 8):
+    import numpy._typing
+
+    _NBitBase: TypeAlias = numpy._typing.NBitBase
+else:
+    _NBitBase: TypeAlias = Any
+
+NUMERIC_TYPES: tuple[Decimal | type[int | float | numpy.bool_ | numpy.floating[_NBitBase] | numpy.integer[_NBitBase]], ...]
```
AlexWaygood, line 15:
```suggestion
NUMERIC_TYPES: tuple[type[Decimal | int | float | numpy.bool_ | numpy.floating[_NBitBase] | numpy.integer[_NBitBase]], ...]
```

stubs/openpyxl/openpyxl/compat/numbers.pyi
```diff
@@ -1,4 +1,21 @@
...
+
+# Referenced outside this module
+_NumericTypes: TypeAlias = (  # noqa: Y047
+    int | float | Decimal | type[numpy.bool_] | type[numpy.floating[_NBitBase]] | type[numpy.integer[_NBitBase]]
```
AlexWaygood, line 19:
```suggestion
    int | float | Decimal | type[numpy.bool_ | numpy.floating[_NBitBase] | numpy.integer[_NBitBase]]
```

stubs/openpyxl/openpyxl/drawing/image.pyi
```diff
@@ -1,9 +1,15 @@
-from _typeshed import Incomplete
+from types import ModuleType
+from typing_extensions import Literal
+
+from PIL import Image as ImageModule
+
+# Using "object" because modules can't by used as type
```
AlexWaygood, line 6: There's a typo here, but I'm also not sure I really understand this comment

```suggestion
# Using "object" because modules can't be used as types
```

stubs/openpyxl/openpyxl/styles/alignment.pyi
```diff
@@ -1,39 +1,64 @@
...
 from openpyxl.descriptors.serialisable import Serialisable
 
-horizontal_alignments: Incomplete
-vertical_aligments: Incomplete
+horizontal_alignments: Final[tuple[_HorizontalAlignementsType, ...]]
+_HorizontalAlignementsType: TypeAlias = Literal[
+    "general", "left", "center", "right", "fill", "justify", "centerContinuous", "distributed"
+]
+vertical_aligments: Final[tuple[_VerticalAlignementsType, ...]]
+_VerticalAlignementsType: TypeAlias = Literal["top", "center", "bottom", "justify", "distributed"]
```
AlexWaygood, line 13: - Fix spelling
- Avoid unnecessary forward references

```suggestion
_HorizontalAlignmentsType: TypeAlias = Literal[
    "general", "left", "center", "right", "fill", "justify", "centerContinuous", "distributed"
]
horizontal_alignments: Final[tuple[_HorizontalAlignmentsType, ...]]
_VerticalAlignmentsType: TypeAlias = Literal["top", "center", "bottom", "justify", "distributed"]
vertical_aligments: Final[tuple[_VerticalAlignmentsType, ...]]
```

stubs/openpyxl/openpyxl/styles/fills.pyi
```diff
@@ -1,79 +1,134 @@
...
+    "gray125",
+    "lightDown",
+    "lightGray",
+    "lightGrid",
+    "lightHorizontal",
+    "lightTrellis",
+    "lightUp",
+    "lightVertical",
+    "mediumGray",
+]
```
AlexWaygood, line 53:
```suggestion
_FillsType: TypeAlias = Literal[
    "solid",
    "darkDown",
    "darkGray",
    "darkGrid",
    "darkHorizontal",
    "darkTrellis",
    "darkUp",
    "darkVertical",
    "gray0625",
    "gray125",
    "lightDown",
    "lightGray",
    "lightGrid",
    "lightHorizontal",
    "lightTrellis",
    "lightUp",
    "lightVertical",
    "mediumGray",
]
fills: Final[tuple[_FillsType, ...]]
```

stubs/openpyxl/openpyxl/styles/fills.pyi
```diff
@@ -1,79 +1,134 @@
...
 
 class Fill(Serialisable):
     tagname: str
     @classmethod
-    def from_tree(cls, el): ...
+    def from_tree(cls: Self, el: _Element) -> Self: ...
```
AlexWaygood, line 58: Hmm, that doesn't look right. If you're using `_typeshed.Self`, the `cls` parameter should be annotated with `type[Self]`. If you're using `typing_extensions.Self`, it doesn't need to be annotated at all.

stubs/openpyxl/openpyxl/styles/fonts.pyi
```diff
@@ -1,58 +1,95 @@
...
-        outline: Incomplete | None = ...,
-        shadow: Incomplete | None = ...,
-        condense: Incomplete | None = ...,
-        extend: Incomplete | None = ...,
+        name: str | None = ...,
+        sz: _FloatSetter | None = ...,
+        b: _BoolSetter = ...,
+        i: _BoolSetter = ...,
+        charset: _IntegerSetter | None = ...,
+        u: Literal["single", "double", "singleAccounting", "doubleAccounting", None] = ...,
```
AlexWaygood, line 76: Looks like this type appears several times in this file; maybe it'd be another good candidate for a type alias

AlexWaygood (commented):
Some more comments from reading through. (Still haven't properly compared with the runtime source in detail.)

stubs/openpyxl/openpyxl/drawing/image.pyi
Avasam: outdated comment. I have since then learned of `ModuleType` which is enough for the needs here.
```suggestion
```

stubs/openpyxl/openpyxl/cell/cell.pyi
Avasam: This comment should've gonna away with your last batch of suggestions anyway. Definitely no longer relevant!

stubs/openpyxl/openpyxl/compat/numbers.pyi
Avasam:
```suggestion
NUMERIC_TYPES: tuple[type[int | float | Decimal | numpy.bool_ | numpy.floating[_NBitBase] | numpy.integer[_NBitBase]], ...]
```

stubs/openpyxl/openpyxl/styles/fills.pyi
Avasam: I need to update usages of `Self` anyway, so I'll double check this at the same time.

stubs/openpyxl/@tests/stubtest_allowlist.txt
```diff
@@ -1,2 +1,429 @@
...
+openpyxl.xml.functions.LXML
+openpyxl.xml.functions.REL_NS
+openpyxl.xml.functions.SHEET_DRAWING_NS
+openpyxl.xml.functions.SHEET_MAIN_NS
+openpyxl.xml.functions.VTYPES_NS
+openpyxl.xml.functions.XML_NS
+
+# Variable differs from runtime type because attributes are actually
+# non-Convertible descriptors and the value is obtained through a dynamic getter
+openpyxl.cell.text.InlineFont.color
```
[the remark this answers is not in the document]
Avasam: I might wanna look into if it'd work to type the descriptors as *actual* descriptors with __set__ and __get__, even though the runtime doesn't actually implement both (but relies on a superclass' __getattr__)

AlexWaygood:
> (I've been quite occupied last week, and I got sick this week. I'll look into using proper descriptors in stubs and cleaning up this PR further soon once I feel a bit better)

The ball's in my court on this one anyway, I owe you the next round of review comments :)

AlexWaygood:
#9921 will fix the stubtest failures relating to the new modules added in recent releases

srittau:
I've merged in the existing changes, but this might have introduces new problems, since some changes were incompatible.

Avasam:
I do think I have a viable solution for the fake desrciptors, by making them look like real generic descriptors! I can even deal with `allow_none` properly.
image

There use is internal only, so as long as the getter/setter works properly and stubtest is happy, should be good. Added bonus that I can correctly type properties using the descriptors instead of the more verbose `@property` workaround that stubtest also doesn't like.
I'm making a completely different PR for that.

stubs/openpyxl/openpyxl/worksheet/dimensions.pyi
```diff
@@ -30,7 +30,7 @@ class Dimension(Strict, StyleableObject):
     outlineLevel: Integer[Literal[True]]
     outline_level: Alias
     collapsed: Bool[Literal[False]]
-    style: Alias
+    style: Alias  # type: ignore[assignment]
```
Avasam, line 33:
```suggestion
    style: Alias
```
This change doesn't feel necessary. Like, it should already be failing if that was an issue?

stubs/openpyxl/openpyxl/worksheet/_reader.pyi
```diff
@@ -70,11 +72,10 @@ class WorkSheetParser:
     col_breaks: ColBreak
     rich_text: bool
     protection: SheetProtection  # initialized after call to parse_sheet_protection()
-
     def __init__(
```
Avasam, line 75:
```suggestion

    def __init__(
```

JelleZijlstra:
@Avasam this is now the oldest open PR and CI is green, but it's still in draft. Is it ready now? If not, what is the plan for this PR?

Avasam:
I've been extracting changes by scope and concepts they relate to, which allowed me to validate the changes actually made sense and fix entire sections of openpyxl stubs at once.

The only thing left here would be a PR that covers adding annotations for openpyxl Cell Values and that should close this PR.

Avasam:
Last PR opened: https://github.com/python/typeshed/pull/13929

srittau:
Thank you for this gargantuan task!

PR #9512, opened 2023-01-12
3rd-party stubtest: use MYPYPATH for internal dependencies

AlexWaygood:
Currently, if we're running stubtest on our stubs for `caldav`, we `pip install` all the requirements for `types-caldav` into the virtual environment before running the test. However, for typeshed-internal requirements, this is unnecessary: we can add the typeshed-internal requirements for `caldav` to MYPYPATH instead. This has two advantages:

- `pip install`ing stuff is slow; this should be faster. Some of our stubs have a lot of internal dependencies: when testing `caldav`, we currently have to `pip install` `types-requests`, `types-urllib3` and `types-vobject`.
- Currently it's impossible to e.g. add two stubs packages in one pull request, with one depending on the other. Instead, you have to add stubs package A, wait for the stub-uploader to push the new stubs to PyPI, and then add stubs package B, which depends on stubs package A. This has tripped up contributors in the past. With this update, we won't need to wait around for the stub-uploader.

### DIFF
--- a/tests/stubtest_third_party.py
+++ b/tests/stubtest_third_party.py
@@ -13,7 +13,7 @@ from typing import NoReturn
 
 import tomli
 
-from utils import colored, get_mypy_req, make_venv, print_error, print_success_msg
+from utils import colored, get_mypy_req, get_recursive_requirements, make_venv, print_error, print_success_msg
 
 
 def run_stubtest(dist: Path, *, verbose: bool = False, specified_stubs_only: bool = False) -> bool:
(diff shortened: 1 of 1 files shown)


hauntsaninja (approved):

tests/stubtest_third_party.py
```diff
@@ -126,15 +131,15 @@ def run_stubtest(dist: Path, *, verbose: bool = False, specified_stubs_only: boo
...
 
     if verbose:
-        print_commands(dist, pip_cmd, stubtest_cmd)
+        print_commands(dist, pip_cmd, stubtest_cmd, mypypath)
 
     return True
 
 
-def print_commands(dist: Path, pip_cmd: list[str], stubtest_cmd: list[str]) -> None:
+def print_commands(dist: Path, pip_cmd: list[str], stubtest_cmd: list[str], mypypath: str) -> None:
```
Avasam, line 139: @AlexWaygood The `dist` argument after this change has been left unused. Was that intentional ?

AlexWaygood: Nope! It makes sense to get rid of it :D

PR #9514, opened 2023-01-12
Update FilterType to retuning bool

### LINKED ISSUE #5700: logging.Logger.addFilter
```
porcupine/plugins/langserver.py:789: error: Argument 1 to "addFilter" of "Filterer" has incompatible type "HackFilter"; expected "Union[Filter, Callable[[LogRecord], int]]"
```

`HackFilter` is a class that has a `filter()` method but doesn't inherit from anything. This should work, according to docs: "You don’t actually need to subclass Filter: you can pass any instance which has a filter method with the same semantics." https://docs.python.org/3/library/logging.html

`Callable[[LogRecord], int]` seems wrong. The imlpementation basically does `self.filters.append(filter)`. Does it really handle callables and objects with a `.filter()` method everywhere where it uses `self.filters`?

KalleDK:
The default Filter returns a bool and not an int

Closes #5700

### DIFF
--- a/stdlib/logging/__init__.pyi
+++ b/stdlib/logging/__init__.pyi
@@ -64,7 +64,7 @@ if sys.version_info >= (3, 11):
 _SysExcInfoType: TypeAlias = Union[tuple[type[BaseException], BaseException, TracebackType | None], tuple[None, None, None]]
 _ExcInfoType: TypeAlias = None | bool | _SysExcInfoType | BaseException
 _ArgsType: TypeAlias = tuple[object, ...] | Mapping[str, object]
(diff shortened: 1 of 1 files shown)


AlexWaygood (approved):
It looks like `homeassistant` just copied their type alias from the typeshed stub: https://github.com/home-assistant/core/blob/2ca73be9d0f02a00d65c9b05e7df5e06318adaff/homeassistant/components/zha/core/gateway.py#L94. So I think they'll just need to update thier type alias; I don't think they're actually passing callables that return `int` to this function.

Having said that, I'll leave this open in case any other maintainers have any thoughts, since the benefits also seem a little slim.

srittau (approved):

PR #9516, opened 2023-01-12
`Flask-Cors`: Add `Flask` as a dependency

AlexWaygood:


stubs/Flask-Cors/METADATA.toml
```diff
@@ -1 +1,3 @@
 version = "3.0.*"
+# Requires a versions of flask/werkzeug with `py.typed` files
```
AlexWaygood, line 2:
```suggestion
# Requires versions of flask/werkzeug with `py.typed` files
```

JelleZijlstra (approved):

PR #9518, opened 2023-01-13
`pycocotools`: Add `numpy` as a dependency

AlexWaygood:
Source code here: https://github.com/cocodataset/cocoapi/tree/master/PythonAPI/pycocotools.

I'm in large part trusting that the commented-out annotations here are correct. From what I can tell from the source code, they look pretty good, but I haven't verified.


AlexWaygood:
Oh, pycocotools doesn't declare numpy as a dependency. There goes that 🤷‍♂️

Avasam:
@AlexWaygood Could still be worth updating the TODOs since they're no longer blocked by #5768

Avasam:
Weird, I just checked, this is their setup.py:
[setup.py.txt](https://github.com/python/typeshed/files/10417178/setup.py.txt)

And this is their .eggs-info/requires:
[requires.txt](https://github.com/python/typeshed/files/10417181/requires.txt)

AlexWaygood:
> Weird, I just checked, this is their setup.py: [setup.py.txt](https://github.com/python/typeshed/files/10417178/setup.py.txt)
> 
> And this is their .eggs-info/requires: [requires.txt](https://github.com/python/typeshed/files/10417181/requires.txt)

It's because of this:

https://github.com/typeshed-internal/stub_uploader/blob/022168078ff4f6ef99fa79401fc1a98c8ad0f326/stub_uploader/metadata.py#L205-L207

See also: the same problem with `click-default-group` here: https://github.com/python/typeshed/pull/9304#issuecomment-1378076625

Avasam:
This would help: https://github.com/typeshed-internal/stub_uploader/pull/88

PR #7625, opened 2022-04-15
sqlite3: add 3.11 additions

JelleZijlstra:
- Blob from python/cpython#30680 (and anticipating that python/cpython#91550 will be merged)
- Aggregate window functions from python/cpython#20903
- Serialize/deserialize from python/cpython#26728
- Limit setting from python/cpython#28463

PR #9519, opened 2023-01-13
Update Unused and __exit__ parameters in stdlib

Avasam:
Follow up to #9475

- Use of `Unused` type alias for unused params.
- Removed an allowlist entry `uuid.getnode` by correctly typing an undocumented/unused param
- Properly(?) typed `asyncio.runners.Runner.__exit__` and `sqlite3.dbapi2.Blob.__exit__` using `type[BaseException] | None, BaseException | None, TracebackType | None` instead of objects. I couldn't check the source for these two, so I'm not sure if the params are meant to be unsued.

There's also 4 `object | None` in stdlib, should I change those to `object` as well?


Avasam:
Random CI failure, you can re-run just that job

AlexWaygood:
Hmm, while support for _arbitrary_ aliases in flake8-pyi is unrealistic, we could certainly add some special-casing for the `_typeshed.Unused` alias specifically. I think I'd prefer to do that rather than add a bajillion `# noqa` comments :)

AlexWaygood:
> Hmm, while support for _arbitrary_ aliases in flake8-pyi is unrealistic, we could certainly add some special-casing for the `_typeshed.Unused` alias specifically. I think I'd prefer to do that rather than add a bajillion `# noqa` comments :)

Are you interested in working on a flake8-pyi PR for this? If not, I can take a look at adding special-casing

Avasam:
> Are you interested in working on a flake8-pyi PR for this? If not, I can take a look at adding special-casing

Sure. If I"m stuck I"ll let you know

stdlib/sqlite3/dbapi2.pyi
```diff
@@ -227,7 +227,7 @@ else:
 
 if sys.version_info < (3, 8):
     class Cache:
-        def __init__(self, *args: Incomplete, **kwargs: object) -> None: ...
+        def __init__(self, *args: Incomplete, **kwargs: Unused) -> None: ...
```
AlexWaygood, line 230: I'm not sure the existing annotation for this is correct, but also: it's 3.7-only, so I'm not sure I care too much 😛

stdlib/sqlite3/dbapi2.pyi
```diff
@@ -453,6 +453,8 @@ if sys.version_info >= (3, 11):
         def seek(self, __offset: int, __origin: int = ...) -> None: ...
         def __len__(self) -> int: ...
         def __enter__(self: Self) -> Self: ...
-        def __exit__(self, __typ: object, __val: object, __tb: object) -> Literal[False]: ...
+        def __exit__(
+            self, __typ: type[BaseException] | None, __val: BaseException | None, __tb: TracebackType | None
+        ) -> Literal[False]: ...
```
AlexWaygood, line 458: This signature was previously discussed in https://github.com/python/typeshed/pull/7625#discussion_r851630434: the parameters here really do take any object

AlexWaygood (commented):
Mostly looks good, but a few points below! Also, you have a fair few unused `noqa` comments following the new flake8-pyi release ;)

stdlib/sqlite3/dbapi2.pyi
Avasam: Nor can I find "Cache" in the docs 🤷 
https://docs.python.org/3.7/library/sqlite3.html

Avasam:
Thanks for looking into the methods I was uncertain about.

I still have one unanswered question left:
> There's also 4 `object | None` in stdlib, should I change those to object as well?

AlexWaygood:
> I still have one unanswered question left:
> 
> > There's also 4 `object | None` in stdlib, should I change those to `object` as well?

Sure, why not!

AlexWaygood (approved):
Thanks!

PR #9521, opened 2023-01-13
Fix setuptools stubtest on windows

Avasam:
I'm fairly certain that's just a shim for implementation details.
Fixes stubtest on windows. Stub on windows and linux should be equivalent now.


AlexWaygood (approved):
LGTM

PR #9522, opened 2023-01-13
Allowlist-only fixes for stubtest windows

Avasam:
Batch of changes in allowlist entries (no stub change) to fix some third party stubtest runs on windows.

### DIFF
--- a/stubs/Pillow/@tests/stubtest_allowlist.txt
+++ b/stubs/Pillow/@tests/stubtest_allowlist.txt
@@ -1,5 +1,7 @@
 PIL.FpxImagePlugin
 PIL.ImageQt.ImageQt
-PIL.ImageQt.QImage  # Does exist at runtime usually, but stubtest can't see it
-PIL.ImageQt.QPixmap  # Does exist at runtime usually, but stubtest can't see it
 PIL.MicImagePlugin
+# Requires Pyside or PyQt to be installed.
+# We don't install either in stubtest so it can't see it
+PIL.ImageQt.QImage
+PIL.ImageQt.QPixmap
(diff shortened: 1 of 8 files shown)


stubs/Pillow/@tests/stubtest_allowlist.txt
```diff
@@ -1,5 +1,7 @@
 PIL.FpxImagePlugin
 PIL.ImageQt.ImageQt
-PIL.ImageQt.QImage  # Does exist at runtime usually, but stubtest can't see it
-PIL.ImageQt.QPixmap  # Does exist at runtime usually, but stubtest can't see it
 PIL.MicImagePlugin
+# Requires Pyside or PyQt to be installed.
```
AlexWaygood, line 4: Could we add Pyside/PyQt to a `stubtest-requirements.txt` file?

Avasam: We'd have to decide which. It's either/or. Pillow will only use one of the 4.
PySide6 or PyQt6. (PySide2 and PyQt5 being for Qt5).
I haven't looked into details if the signature of the bindings is equivalent. Or if it would matter for stubtest.

PySide is official bindings by the Qt team (aka Qt for Python). Has a public issue tracker.
PyQt is made by riverbank computing, closed source and no issue tracker. More popular because historically (not so much nowadays) it's been more complete than PySide (which didn't always exist). Also the name helps.

AlexWaygood: My instinct would be to go for `PyQt`, purely because I've heard of it (and I'm guessing it's probably more widely used by users of `type s-Pillow` as well) 🙂

AlexWaygood: But I guess we can leave this for now, postpone for another PR

AlexWaygood (approved):

PR #9525, opened 2023-01-14
Complete stubtest and fix any subclassing in `tqdm`

Avasam:
Split off from #9505

Fixes the last two Any subclassing issues for #9491


AlexWaygood (commented):
I'm concerned about how maintainable this is going to be :(

Stubtest will warn us if significant changes are made to `_rich_shims.ProgressColumn`, because it's a superclass of `tqdm.rich.FractionColumn` and `tqdm.rich.RateColumn` -- if changes to `_rich_shims.ProgressColumn` are made at runtime, stubtest will complain that `tqdm.rich.FractionColumn` and `tqdm.rich.RateColumn` are inconsistent.

But for the other classes in `_rich_shims.pyi`, I don't think we'll get any coverage from stubtest at all, and that worries me.

Avasam:
> I'm concerned about how maintainable this is going to be :(
> [...]
> But for the other classes in `_rich_shims.pyi`, I don't think we'll get any coverage from stubtest at all, and that worries me.

I don't mind simplifying and removing parts of the shim that can't be validated by stubtest (I'll try a few things so I'm confident in what is or isn't validated) as long as we have the structure of `ProgressColumn` (even if attributes and methods end up untyped)

AlexWaygood:
> I don't mind simplifying and removing parts of the shim that can't be validated by stubtest (I'll try a few things so I'm confident in what is or isn't validated) as long as we have the structure of `ProgressColumn` (even if attributes and methods end up untyped)

That sounds like a great plan to me 👍

Avasam:
stubtest doesn't seem to care what I do with `ProgressColumn` :/ I added random parameters and methods. Removed exisitng ones (with `ignore_missing_stub = false`) and it always passes.

AlexWaygood:
> stubtest doesn't seem to care what I do with `ProgressColumn` 😕 I added random parameters and methods. Removed exisitng ones (with `ignore_missing_stub = false`) and it always passes. It only cares about the ABCMeta metaclass if I remove the ABC baseclass.

That's a bummer :/ I guess something like the patch described in https://github.com/python/typeshed/issues/3968#issue-613086106 would help with that somewhat.

I still feel like I'm okay with `ProgressColumn` being typed in detail, but a bit wary of some of the others being typed in so much detail. Even if it's not something that stubtest tests _now_, it feels like the kind of thing that _should_ be stubtestable. With the other classes, it feels more dangerous

Avasam:
Yeah I decided to only keep `rich.progress.ProgressColumn` as well. Changed to focus to "completing stubtest" since fixing the metaclass issue incidentaly fixes the Any subclassing. And even if stubtest "should" catch missing methods, `ignore_missing_stub = false` would need to be set anyway.

AlexWaygood (approved):
Looks great! Thanks for your patience and for being willing to revise this :)

PR #9526, opened 2023-01-14
Use numpy in pyscreeze

Avasam:


Avasam:
humm, another case of the dependency not being marked as an actual dependency... Makes sense if it's optional anyway

Avasam:
Might be a bit hacky for your taste, but I think that it shows the idea behind a "typeshed testing only requires" / "optional requires". Which could be done cleaner than this.

PR #9314, opened 2022-12-01
Complete stubs for `bleach`

sobolevn:
`bleach` source: https://github.com/mozilla/bleach
`html5lib` source: https://github.com/html5lib/html5lib-python

PR #9527, opened 2023-01-14
Update comments post #5768 

Avasam:
Now that #5768 is closed and completed. Comments mentioning the inability to use certain imports due to the types dependency restriction are no longer true. However, some still can't be installed for a different reason: The dependency isn't declared by the original package. So the stub_uploader refuses it.

I looked for mentions of `numpy`, `comtypes` and 5768. If you know more that I missed, please let me know.

### DIFF
--- a/stubs/PyScreeze/pyscreeze/__init__.pyi
+++ b/stubs/PyScreeze/pyscreeze/__init__.pyi
@@ -9,8 +9,12 @@ from PIL import Image
 _Unused: TypeAlias = object
 _P = ParamSpec("_P")
 _R = TypeVar("_R")
-# TODO: cv2.Mat is not available as a type yet: https://github.com/microsoft/python-type-stubs/issues/211
+# TODO: cv2.Mat is not available as a type yet:
+# https://github.com/microsoft/python-type-stubs/issues/211
+# https://github.com/microsoft/python-type-stubs/tree/main/cv2
+# https://github.com/opencv/opencv/pull/20370
 # cv2.Mat is just an alias for a numpy NDArray, but can't import that either.
+# Because pyscreeze does not declare it as a dependency, stubsuploader won't let it.
 _Mat: TypeAlias = Incomplete
 
 useOpenCV: bool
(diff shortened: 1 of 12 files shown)


stubs/PyScreeze/pyscreeze/__init__.pyi
```diff
@@ -9,8 +9,12 @@ from PIL import Image
 _Unused: TypeAlias = object
 _P = ParamSpec("_P")
 _R = TypeVar("_R")
-# TODO: cv2.Mat is not available as a type yet: https://github.com/microsoft/python-type-stubs/issues/211
+# TODO: cv2.Mat is not available as a type yet:
+# https://github.com/microsoft/python-type-stubs/issues/211
+# https://github.com/microsoft/python-type-stubs/tree/main/cv2
+# https://github.com/opencv/opencv/pull/20370
 # cv2.Mat is just an alias for a numpy NDArray, but can't import that either.
+# Because pyscreeze does not declare it as a dependency, stubsuploader won't let it.
```
AlexWaygood, line 17:
```suggestion
# Because pyscreeze does not declare it as a dependency, stub_uploader won't let it.
```

AlexWaygood (commented):
Let's leave the `pycocotools` comments as they are for now. It's a shortcoming of the logic in the `stub_uploader` (https://github.com/python/typeshed/pull/9518#issuecomment-1382725664), and I'd like to try fixing those shortcomings :)

The other changes look great!

Avasam:
> Let's leave the `pycocotools` comments as they are for now. It's a shortcoming of the logic in the `stub_uploader` ([#9518 (comment)](https://github.com/python/typeshed/pull/9518#issuecomment-1382725664)), and I'd like to try fixing that shortcoming :)

Ahh! (Read the comment you left on the pycocotools pr too). Alright I'll undo that comment since it may be fixed separately (and doesn't accurately describe the issue anyway)

Avasam:
The only ones I didn't make a PR for are Bleach because it's already tackled in #9314.

tests/mypy_test.py
```diff
@@ -264,7 +264,7 @@ def run_mypy(
             # Stub completion is checked by pyright (--allow-*-defs)
             "--allow-untyped-defs",
             "--allow-incomplete-defs",
-            "--allow-subclassing-any",  # TODO: Do we still need this now that non-types dependencies are allowed? (#5768)
+            "--allow-subclassing-any",  # See #9491
```
Avasam, line 267: If #9491 is accepted, then this line will be removed. If not then this links to the discussion/decision.

AlexWaygood (approved):

PR #9528, opened 2023-01-14
Complete and validate `redis.ocsp` (use requests, cryptography and pyopenssl)

Avasam:
I based the `bytes` parameters on the typing in `requests` and `cryptography`. I don't know if they can take a `ReadableBuffer`


stubs/redis/METADATA.toml
```diff
@@ -1 +1,5 @@
 version = "4.4.0"
+requires = ["types-pyOpenSSL"]
```
AlexWaygood, line 2: What does the stub uploader say if we add `cryptography` here? It would be nice to add it if possible, since we're doing imports from `cryptography` directly in the stub, and it's good to avoid implicit dependencies where possible.

Avasam: I agree it should be there and not implied from types-pyOpenSSL. I didn't realize I forgot it.

stubs/redis/redis/ocsp.pyi
```diff
@@ -1,13 +1,21 @@
...
-    def components_from_socket(self): ...
-    def components_from_direct_connection(self): ...
-    def build_certificate_url(self, server, cert, issuer_cert): ...
-    def check_certificate(self, server, cert, issuer_url): ...
-    def is_valid(self): ...
+    SOCK: SSLObject
+    HOST: str
+    PORT: int
+    CA_CERTS: str | None
+    def __init__(self, sock: SSLObject | SSLSocket, host: str, port: int, ca_certs: str | None = ...) -> None: ...
```
AlexWaygood, line 15: Why is the type for `sock` in the `__init__` method different to the attribute type?

AlexWaygood (commented):
Just one question

stubs/redis/redis/ocsp.pyi
Avasam: I updated one but forgot the other.

AlexWaygood (approved):

PR #9530, opened 2023-01-14
Complete jack-client stub

Avasam:
Used numpy, completed a TODO and FIXME, used some runtime inspection to complete missing class members.
I can't use `_cffi_backend.__CDataOwn` as a base type because it's missing from `types-cffi` (the name suggests it's private)


Avasam:
JACK-Client has the `extra_requires` `NumPy`, I feel like it's messing with the stubsuploader test

stubs/JACK-Client/jack/__init__.pyi
```diff
@@ -1,12 +1,38 @@
...
+    frame_rate: int
+    frame_time: float
+    next_time: float
+    padding: _CDataBase  # <cdata 'int32_t[7]'>
+    tick: int
+    ticks_per_beat: float
+    unique_1: int
+    unique_2: int
+    usecs: int
+    valid: Literal[0, 1]
```
JelleZijlstra, line 34: Are you sure about this? `position2dict` does bit fiddling on this field that suggests there are more possible values.

stubs/JACK-Client/jack/__init__.pyi
```diff
@@ -196,12 +223,12 @@ class OwnPort(Port):
...
     def get_buffer(self) -> _CBufferType: ...
-    def get_array(self) -> _NDArray: ...
+    def get_array(self) -> NDArray[numpy.float32]: ...
 
 class OwnMidiPort(MidiPort, OwnPort):
-    def __init__(self, *args: Any, **kwargs: Any) -> None: ...
-    def get_buffer(self) -> _CBufferType: ...
-    def get_array(self) -> _NDArray: ...
+    def __init__(self, port_ptr: _CDataBase, client: Client) -> None: ...
+    def get_buffer(self) -> NoReturn: ...
```
JelleZijlstra, line 230: These raise NotImplementedError. Returning NoReturn isn't helpful for users, since that will just make code calling these unreachable. Maybe mark them with `@abstractmethod` instead.

JelleZijlstra (commented):
Thanks for taking this on. I didn't get this thing to import but here's a few comments from looking at the source for the package.

stubs/JACK-Client/jack/__init__.pyi
Avasam: Not 100% certain. I have no issues taking it safe and using `int`

stubs/JACK-Client/jack/__init__.pyi
Avasam: But then you get `Cannot instantiate abstract class "OwnMidiPort" with abstract attributes "get_array" and "get_buffer"` when `OwnMidiPort` is *not* abstract.

```
    def get_buffer(self):
        """Not available for MIDI ports."""
        raise NotImplementedError('get_buffer() not available on MIDI ports')

    def get_array(self):
        """Not available for MIDI ports."""
        raise NotImplementedError('get_array() not available on MIDI ports')
```

it sounds more like "These don't exist on MIDI ports" than "Not yet implemented / subclass it and implement it". Basically 
```py
get_buffer = None
get_buffer = None
```
would be equivalent. `NotImplementedError` may not have been the best choice of error to raise from the library.

PR #9531, opened 2023-01-14
Allow stubtest to validate `PIL.FpxImagePlugin` and `PIL.MicImagePlugin`

Avasam:
Allow stubtest to validate `PIL.FpxImagePlugin` and `PIL.MicImagePlugin`
Fixed most allowlist entries
Completed those two files as much as I can (`def load` will depend on #9466)
Fixed a parameter type in `PIL.ImageQt`


JelleZijlstra:
There is a merge conflict now.

stubs/Pillow/PIL/ImageQt.pyi
```diff
@@ -3,20 +3,20 @@ from typing_extensions import Literal, TypeAlias
 
 from .Image import Image
 
-QImage: TypeAlias = Any  # imported from either of {PyQt6,PySide6,PyQt5,PySide2}.QtGui
-QPixmap: TypeAlias = Any
+_QImage: TypeAlias = Any  # imported from either of {PyQt6,PySide6,PyQt5,PySide2}.QtGui
```
AlexWaygood, line 6: Won't this change cause type checkers to emit false-positive errors if users do `from PIL.ImageQT import QImage`?

Avasam: Doesn't feel like it's meant to be re-exported. You'd do `from PyQt6.QtGui import QImage`. It also only exists (because it's only imported) if the corresponding Qt bindings library is installed.
Also not referenced in the doc/API: https://pillow.readthedocs.io/en/stable/reference/ImageQt.html
(even most other public methods are wrapped in `PIL.Image`)

AlexWaygood: That makes sense, thank you!

stubs/Pillow/PIL/MicImagePlugin.pyi
```diff
@@ -1,11 +1,15 @@
...
+from typing_extensions import Literal, TypeAlias
 
 from .TiffImagePlugin import TiffImageFile
 
+_OleFileIO: TypeAlias = Any  # olefile.OleFileIO
+_OleStream: TypeAlias = Any  # olefile.OleStream
+
 class MicImageFile(TiffImageFile):
+    ole: _OleFileIO
     format: ClassVar[Literal["MIC"]]
```
AlexWaygood, line 11:
```suggestion
    format: ClassVar[Literal["MIC"]]
    format_description: ClassVar[str]
```

stubs/Pillow/PIL/MicImagePlugin.pyi
```diff
@@ -1,11 +1,15 @@
...
-    fp: Any
-    frame: Any
-    def seek(self, frame) -> None: ...
-    def tell(self): ...
+    fp: _OleStream
+    frame: int
```
AlexWaygood, line 13: Can be `None`: https://github.com/python-pillow/Pillow/blob/77b316c8f01bac2a7bf8b37a8010732d172ad6dc/src/PIL/MicImagePlugin.py#L67

stubs/Pillow/PIL/MicImagePlugin.pyi
```diff
@@ -1,11 +1,15 @@
...
+    def seek(self, frame: int) -> None: ...
+    def tell(self) -> int: ...
```
AlexWaygood, line 15:
```suggestion
    def tell(self) -> int | None: ...
```

stubs/Pillow/PIL/FpxImagePlugin.pyi
```diff
@@ -13,4 +14,9 @@ class FpxImageFile(ImageFile):
     format: ClassVar[Literal["FPX"]]
     format_description: ClassVar[str]
     fp: _OleStream | None
+    maxid: int
+    rawmode: str
+    jpeg: dict[int, Incomplete]
+    tile_prefix: Incomplete
```
AlexWaygood, line 20: Where's this one defined?

stubs/Pillow/PIL/MicImagePlugin.pyi
```diff
@@ -9,7 +9,10 @@ _OleStream: TypeAlias = Any  # olefile.OleStream
 class MicImageFile(TiffImageFile):
     ole: _OleFileIO
     format: ClassVar[Literal["MIC"]]
+    format_description: ClassVar[str]
     fp: _OleStream
-    frame: int
+    frame: int | None
+    images: list[str]
```
AlexWaygood, line 15: It looks to me like this is actually a list of sequences of strings, rather than a list of strings: the `path` variable here looks to be a sequence of strings, and the code does `self.images.append(path)`: https://github.com/python-pillow/Pillow/blob/77b316c8f01bac2a7bf8b37a8010732d172ad6dc/src/PIL/MicImagePlugin.py#L56-L59

stubs/Pillow/PIL/FpxImagePlugin.pyi
Avasam: https://github.com/python-pillow/Pillow/blob/main/src/PIL/FpxImagePlugin.py#L221

stubs/Pillow/PIL/MicImagePlugin.pyi
Avasam: Indeed. A `list` of str to be more exact: https://github.com/decalage2/olefile/blob/master/olefile/olefile.py
```suggestion
    images: list[list[str]]
```

stubs/Pillow/PIL/FpxImagePlugin.pyi
AlexWaygood: Weird, wonder how I missed that. Thanks!

stubs/Pillow/PIL/MicImagePlugin.pyi
AlexWaygood: Yup, that looks correct!

AlexWaygood (approved):

PR #9532, opened 2023-01-14
Resolve stubtest allowlist entries for `ldap3`

Avasam:


AlexWaygood:
Looks like there's a bug in `check_consistent.py`. This line:

https://github.com/python/typeshed/blob/85792bca716f5fc24272c5f154937ddd20678144/tests/check_consistent.py#L194

Should be changed to:

```python
specified_stubtest_platforms = set(tool_stubtest.get("platforms", ["linux"])) 
```

Feel free to fix it as part of this PR!

AlexWaygood:
On second thoughts, I'll send a PR to centralize metadata parsing in our tests. It'll reduce the risk of future bugs like this, and also be more type-safe.

Avasam:
> On second thoughts, I'll send a PR to centralize metadata parsing in our tests. It'll reduce the risk of future bugs like this, and also be more type-safe.

That's actually something I thought about. And even commented with hauntsaninja: https://github.com/python/typeshed/pull/9298#issuecomment-1330145345
Reduce code duplication. Increased type safety for typeshed's own testing code. Consistent defaults. Easier for contributors. etc.
I think it could be a good idea to centralize it even across projects: https://github.com/typeshed-internal/stub_uploader/blob/main/stub_uploader/metadata.py

AlexWaygood:
Feel free to do the easy fix for `check_consistent.py` as part of this PR! I don't want to hold this PR up until another maintainer's had a chance to review #9534.

AlexWaygood (approved):

PR #9534, opened 2023-01-14
Centralize METADATA.toml parsing in the test suite

AlexWaygood:
We have various bits of logic scattered across the test suite to parse METADATA.toml files. This is a bug magnet: see e.g. https://github.com/python/typeshed/pull/9532#issuecomment-1382752690. (Failing CI run here: https://github.com/python/typeshed/actions/runs/3918898414/jobs/6699564673.) By centralizing the logic all in one place, we reduce the risk of bugs in our tests, and also improve type safety.

This PR centralizes METADATA.toml-parsing logic into a new file, `tests/parse_metadata.py`.


tests/check_consistent.py
```diff
@@ -162,46 +130,8 @@ def _find_stdlib_modules() -> set[str]:
 
 def check_metadata() -> None:
```
Avasam, line 131: At this point, this feels redundant since other tests will fail anyway.
Unless you wanna use it as a smoke test.

AlexWaygood: I get what you mean, but the other tests will assert the correctness of the METADATA.toml files "by accident". I feel like it's useful to have a test that asserts the correctness of the METADATA.toml files "on purpose". That way we still have this guaranteed to be tested in CI, even if some other tests no longer need to get stuff from METADATA.toml files in the future, for whatever reason.

Avasam: Yeah that's a smoke test (an easy test that says, if this fails, everything else will fail anyway, so it's easy to find the source of the error). Just wanted to check if it was on purpose :)

tests/parse_metadata.py
```diff
@@ -0,0 +1,264 @@
...
+        "extras",
+        "ignore_missing_stub",
+        "platforms",
+    }
+}
+_DIST_NAME_RE: Final = re.compile(r"^[a-z0-9]([a-z0-9._-]*[a-z0-9])?$", re.IGNORECASE)
+
+
+@cache
+def read_metadata(distribution: str) -> StubMetadata:
```
Avasam, line 139: Idk if you wanna add them here or in a future improvement PR. But last time I played around with metadata validation, I figured there was a bunch of validation missing. Here's all of 'em here if you want inspiration: https://github.com/Avasam/typeshed/pull/13/files#diff-dcd1108b72e2746cfbed1553d3af6dec2f0623b2c1014dadb646c8577d64e046R77

AlexWaygood: Oh, I didn't realise you'd already put so much work into this -- you should have said!

Though I prefer some of the choices made in my PR ;) (`dataclasses` > `TypedDict`s, in my opinion)

Avasam: I agree with the choices you made. I was just loosely playing around with some ideas. Nothing too serious.

AlexWaygood: Still, y'know, feel free to say "Hey, I've already been poking around at this problem!" in cases like this :)

In terms of METADATA.toml validation, I can't see _much_ in your PR that isn't covered in this PR -- but if I'm missing something, feel free to add it in a followup PR!

tests/parse_metadata.py
```diff
@@ -0,0 +1,264 @@
...
+    for external dependencies, leave them as they are in the METADATA.toml file.
+
+    Note that this function may consider things to be typeshed stubs
+    even if they haven't yet been uploaded to PyPI.
+    If a typeshed stub is removed, this function will consider it to be an external dependency.
+    """
+    pypi_name_to_typeshed_name_mapping = get_pypi_name_to_typeshed_name_mapping()
+    dependencies = read_metadata(distribution).requires
+    typeshed, external = [], []
+    for dependency in dependencies:
```
AlexWaygood, line 233:
```suggestion
    typeshed, external = [], []
    for dependency in read_metadata(distribution).requires:
```

tests/parse_metadata.py
```diff
@@ -0,0 +1,263 @@
+"""Tools to help parse and validate information stored in METADATA.toml files."""
+from __future__ import annotations
+
+import os
+import re
+from collections.abc import Mapping
+from dataclasses import dataclass
+from pathlib import Path
+from typing_extensions import Annotated, Final, NamedTuple, TypeGuard, final
```
AlexWaygood, line 9:
```suggestion
from typing import NamedTuple
from typing_extensions import Annotated, Final, TypeGuard, final
```

Avasam:
What about the `[mypy-tests]` sections? https://github.com/python/typeshed/pull/9534/files#diff-b42eb36bcaa2d1a163ed1c202e81cffd058915ca695b660d49c636b3a5b56857R199-R216
(which doesn't seem to be currently used anywhere?)

AlexWaygood:
> What about the `[mypy-tests]` sections? https://github.com/python/typeshed/pull/9534/files#diff-b42eb36bcaa2d1a163ed1c202e81cffd058915ca695b660d49c636b3a5b56857R199-R216 (which doesn't seem to be currently used anywhere?)

I believe that if somebody _did_ try to add a per-stub config section for mypy, `check_consistent.py` would already complain about it. And, as you say, it's not currently used anywhere. I'd prefer not to touch it for this PR.

PR #9536, opened 2023-01-14
Fix Any subclassing in `fpdf2`

Avasam:
Can't reference `fontools` directly because it does not have a `py.typed` marker (it's only partially typed).
It would be a bit overkill anyway.

Fixes an Any subclassing error in `fpdf2` by shimming the needed `fontools` types.

Ref: #9491


Avasam:
Unlike in #9525, all type definitions here are subclasses of each-other. And parameters use standard types. So the whole shim file makes more sense

srittau (approved):

PR #9543, opened 2023-01-15
Update tensorflow to 2.11

hmc-cs-mdrissi:
Fixes stubtest failure in [stubsabot](https://github.com/python/typeshed/pull/9539).


AlexWaygood:
Thank you!

Issue #8955, opened 2022-10-21
Remove partial classes/modules from CONTRIBUTING?


srittau:
We still have a section about "Incomplete stubs" in CONTRIBUTING. I believe we should remove this as nowadays stubs are generally generated using stubgen, so this shouldn't be an issue anymore. I don't want to encourage new incomplete stubs, and it makes the CONTRIBUTING file a longer read.

We only should leave the sentence about unannotated arguments and return values.

ramvikrams:
Done with the changes

Avasam:
With this, should `ignore_missing_stub` default to false? If we don't want to encourage new incomplete stubs, it's mainly really needed for stubs that are already present and incomplete, or would result in an allowlist that's too complex.

PR #9548, opened 2023-01-16
Rework the "Incomplete stubs" section

srittau:
Also move it to the "Conventions" section due its scope change.

Closes: #8955

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -490,39 +490,27 @@ into any of those categories, use your best judgement.
 * Use `HasX` for protocols that have readable and/or writable attributes
   or getter/setter methods (e.g. `HasItems`, `HasFileno`).
 
-### Incomplete stubs
+### Incomplete annotations
 
-We accept partially annotated stubs, especially for larger packages.
-These should follow the following guidelines:
-
-* A package must include all public submodules, all public classes,
-  functions and methods, and all other public fields.
-* Included functions and methods must list all arguments, but the argument
-  types and the return type can be left unannotated or only be partially
-  annotated.
-* A type can be partially annotated by using the `_typeshed.Incomplete`
-  marker. Do not use `Incomplete` as the only type for argument or return
-  types. See below for an example.
-
-Example:
+When submitting new stubs, it is not necessary to annotate all arguments, return types, and fields. Such items should either be left unannotated or use `_typeshed.Incomplete` if this is not possible:
 
 ```python
 from _typeshed import Incomplete
 
-pub_field1: Incomplete  # unannotated field with required annotation
-pub_field2: list[Incomplete]  # partially annotated
+field: Incomplete  # unannotated 
 
-def pub_func(x: str, y, z: Incomplete | None = None) -> Any | None: ...
+def foo(x): ...  # unannotated argument and return type
 ```
 
-In this example, argument `x` is fully annotated and `y` is left unannotated.
-`Incomplete` is not used for `y`.
-`z` is partially annotated; `None` is one of the allowed types, but other
-types are allowed as well, but this needs further investigation. The return
-type is fully annotated. In this case, the return type was investigated and
-it was found that it can't be fully represented using the current type system,
-which is why `Any` was used. A return type that could potentially be improved
-would be annotated as `Incomplete | None`.
+`Incomplete` can also be used for partially known types:
+
+```python
+def foo(x: Incomplete | None = None) -> list[Incomplete]: ...
+```
+
+### `Any` vs. `Incomplete`
+
+While `Incomplete` is a type alias of `Any`, they serve difference purposes: `Incomplete` is a placeholder where a proper type might be substituted. It's a "to do" item and should be replaced if possible. `Any` is used when it's not possible to accurately type an item using the current type system. It should be used sparingly.
 
 ## Submitting Changes
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -492,7 +492,9 @@ into any of those categories, use your best judgement.
 
 ### Incomplete annotations
 
-When submitting new stubs, it is not necessary to annotate all arguments, return types, and fields. Such items should either be left unannotated or use `_typeshed.Incomplete` if this is not possible:
+When submitting new stubs, it is not necessary to annotate all arguments,
+return types, and fields. Such items should either be left unannotated or
+use `_typeshed.Incomplete` if this is not possible:
 
 ```python
 from _typeshed import Incomplete
@@ -510,7 +512,11 @@ def foo(x: Incomplete | None = None) -> list[Incomplete]: ...
 
 ### `Any` vs. `Incomplete`
 
-While `Incomplete` is a type alias of `Any`, they serve difference purposes: `Incomplete` is a placeholder where a proper type might be substituted. It's a "to do" item and should be replaced if possible. `Any` is used when it's not possible to accurately type an item using the current type system. It should be used sparingly.
+While `Incomplete` is a type alias of `Any`, they serve difference purposes:
+`Incomplete` is a placeholder where a proper type might be substituted.
+It's a "to do" item and should be replaced if possible. `Any` is used when
+it's not possible to accurately type an item using the current type system.
+It should be used sparingly.
 
 ## Submitting Changes
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -501,23 +501,22 @@ These need to follow the following guidelines:
   types and the return type can be left unannotated or only be partially
   annotated.
 * A type can be partially annotated by using the `_typeshed.Incomplete`
-  marker. A field can be left unannotated by using `Incomplete` as well.
-  Do not use `Incomplete` unless it's syntactically required. See below for
-  an example.
+  marker. Do not use `Incomplete` as the only type for argument or return
+  types. See below for an example.
 
-Example of an unannotated field and a partially annotated function:
+Example:
 
 ```python
 from _typeshed import Incomplete
 
-pub_field1: Incomplete  # unannotated
+pub_field1: Incomplete  # unannotated field with required annotation
 pub_field2: list[Incomplete]  # partially annotated
 
 def pub_func(x: str, y, z: Incomplete | None = None) -> Any | None: ...
 ```
 
 In this example, argument `x` is fully annotated and `y` is left unannotated.
-`Incomplete` is not used for `y` as it's not syntactically required.
+`Incomplete` is not used for `y`.
 `z` is partially annotated; `None` is one of the allowed types, but other
 types are allowed as well, but this needs further investigation. The return
 type is fully annotated. In this case, the return type was investigated and
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -493,7 +493,7 @@ into any of those categories, use your best judgement.
 ### Incomplete stubs
 
 We accept partially annotated stubs, especially for larger packages.
-These need to follow the following guidelines:
+These should follow the following guidelines:
 
 * A package must include all public submodules, all public classes,
   functions and methods, and all other public fields.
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -524,6 +490,40 @@ into any of those categories, use your best judgement.
...
+  Do not use `Incomplete` unless it's syntactically required. See below for
+  an example.
+
+Example of an unannotated field and a partially annotated function:
+
+```python
+from _typeshed import Incomplete
+
+pub_field1: Incomplete  # unannotated
+pub_field2: list[Incomplete]  # partially annotated
```
Avasam, line 514: Something I've noticed for a short while now. And may require further discussion. Is that typeshed is "lying" and hiding "partially unknown types" to strict pyright users.

srittau: Possible, maybe @erictraut can shed some light on that.

erictraut: I think it's a good idea to use `Incomplete` as an alias for `Any` (in cases where it is syntactically required) to indicate that someone hasn't yet made the investment to provide a correct type annotation. That makes it clear that it's a "work in progress". Contributors who want to improve typeshed can then focus their attention on eliminating `Incomplete`. 

The result for strict pyright users isn't any worse than if `Any` were used in these locations.

Avasam: Wouldn't `pub_field2: list` allow to warn users about using unknown types through `reportUnknownMemberType` and `reportUnknownVariableType` ?

erictraut: Yes, that's correct. Since `[Incomplete]` isn't syntactically required, the guidance would seem to indicate that it should be omitted in this particular case.

srittau: Good point, I will update the wording to require `Incomplete` in these cases.

erictraut: I think @Avasam is arguing that the wording is fine but the code sample should be changed — that `pub_field2` should use `list` rather than `list[Incomplete]`. A type annotation of `list` is more useful for users of pyright who are using strict-mode checks because it will be evaluated as `list[Unknown]` which will result in an error if it is used. By contrast, `list[Incomplete]` will be interpreted as `list[Any]` which will not result in any strict-mode errors.

srittau: We have handled this differently in typeshed so far. Using bare `list` would trigger "missing generic argument" warnings if I remember correctly. And I think this is a valid warning that we should keep.

Avasam: Exactly what Eric Traut said.
It would require removing `"reportMissingTypeArgument": "error"` from the non-stricter configuration though. Which can be done separately from this PR (and then update the doc). I just wanted to bring attention to this.

erictraut: Or you could use `# pyright: ignore[reportMissingTypeArgument]` to suppress errors on individual lines.

srittau: I don't like either approach. I think we should continue using `list[Incomplete]`.

erictraut: I don't have a strong opinion. It sounds like there's an established precedent in typeshed already, so it makes sense to stick with it.

@Avasam, if you're concerned about specific "holes" created by the use of `Incomplete`, you can plug those holes by contributing improvements that replace those instances of `Incomplete` with the proper type.

srittau: One thing that would really help is a proper `typing.Unknown` type in Python. But that gets too off-topic for this PR.

srittau: Also, since I didn't give a proper justification for my opinion:

* `reportMissingTypeArgument` is a useful warning, not only in completed stubs. Disabling it would be a disservice.
* Using per-instance warning suppression would only work for pyright, but I think mypy warns about it as well. Also, `list[Incomplete]` is more concise and it's immediately obvious that the annotation is incomplete. It's also greppable.

JelleZijlstra: I had a vague recollection that pyright was meant to treat the `_typeshed.Incomplete` alias like `Unknown`. Sounds like that didn't happen, but it would solve the issue here.

AlexWaygood: If pyright did treat `Incomplete` like `Unknown`, presumably we'd get a warning from pyright about an unknown (or partially unknown) type every time we did `x: Incomplete` or `y: list[Incomplete]`, so we'd have to `pyright: ignore` it every time we did something like that in a stub with our stricter pyright settings enabled.

CONTRIBUTING.md
```diff
@@ -524,6 +490,39 @@ into any of those categories, use your best judgement.
...
+
+We accept partially annotated stubs, especially for larger packages.
+These need to follow the following guidelines:
+
+* A package must include all public submodules, all public classes,
+  functions and methods, and all other public fields.
+* Included functions and methods must list all arguments, but the argument
+  types and the return type can be left unannotated or only be partially
+  annotated.
+* A type can be partially annotated by using the `_typeshed.Incomplete`
```
hmc-cs-mdrissi, line 503: Why is,

```python
def foo(x) -> int:
  ...
```

preferred over,

```python
def foo(x: Incomplete) -> int:
  ...
```

The latter I can grep for easily to eventually fix it. The former feels easier to forget.

AlexWaygood: See #8472 for where we discussed this point in depth

Avasam: > The latter I can grep for easily to eventually fix it. 

Alternatively you can check the stub under pyright stricter config and they'll all show up.

AlexWaygood (commented):
I feel sort-of neutral on this.

Most of our stubs these days are contributed using the `create_baseline_stubs` script. For stubs created using that script, there's indeed no reason to produce partial or incomplete stubs. The idea is to produce "low-quality but complete" stubs, and iterate on them in further PRs to slowly improve the quality.

But not all stubs contributions follow this process: our recently-contributed TensorFlow stubs, for example, are quite incomplete, but what has been contributed has been of a much higher quality than the stubs that `create_baseline_stubs` would autogenerate. The idea here is different: to produce "high-quality but incomplete" stubs, and iterate on them in further PRs to slowly work towards completion.

Is there a strong reason to forbid the second way of doing things?

Avasam:
> Is there a strong reason to forbid the second way of doing things?

In theory, anything missing can be filled in by stubgen. I've never tried it, but maybe pytype's `merge-pyi` could help. Although that's more something maintainers would care about, and could run afterward.

Idk if *having* to provide "stubtest complete" stubs could discourage a potential new contributor to add a new stub gets the feeling this will mean having to put in more work that what they thought they needed. Or if it increases the complexity of their initial stub due to stubtest quirks.

CONTRIBUTING.md
Avasam: @AlexWaygood But then you'd disable the rule entirely in typeshed because it wouldn't mean anything for typeshed stubs (`list` would be equivalent to `list[Incomplete]`, so no point in raising an error).

Also, relevant question I had asked on pyright discussion forum and Eric's answer: https://github.com/microsoft/pyright/discussions/4471

> Incomplete isn't an official symbol in the type system or in the Python runtime, and it's not documented in any PEPs. I'd therefore be reluctant to give it any special-case meaning in pyright. It's defined in typeshed.pyi as a type alias for Any, so that's how pyright will treat it.

Speaking of which, is there a proposition or something somewhere that I could upvote for official `Unknown`/`Incomplete` support? I heard that idea being thrown around a handful of times.

AlexWaygood: > Speaking of which, is there a proposition or something somewhere that I could upvote for official `Unknown`/`Incomplete` support? I heard that idea being thrown around a handful of times.

I don't know of such a proposal, but I'd also upvote it if there were one. I think it's a very useful concept, and I've often thought it was a shame mypy didn't have something similar.

hmc-cs-mdrissi:
I was assuming some libraries would be treated as exceptional cases. In tensorflow’s case using mypy’s stubgen on it last I tried it crashed. Also problematically tensorflow’s relies on heavy dynamic import patching so most classes/functions where they are defined is in an internal implementation detail and the public documented location is not what stubgen can generate. The last problem is tensorflow also uses a good amount of dynamic patching including for some core methods so my previous experience using pyright generated stubs, they were prone to false positives that my team had bad experience with them.

But at same time I think tensorflow is pretty weird case and most libraries generated stubs shouldn’t hit these issues. I wouldn’t design rules based on tensorflow and as long as special cases can still be allowed as needed this recommendation seems fine.

AlexWaygood:
> I was assuming some libraries would be treated as exceptional cases.

To be clear: I don't think anybody is proposing to lay down rules that can never be broken, so we almost certainly would still have accepted the TensorFlow stubs PR even if this PR had already been merged.

But that's very useful feedback about why you went about TensorFlow stubs the way you did, and makes me lean in favour of merging this PR :)

srittau:
Maybe it would be useful to just explain `Incomplete` and leave out all comments on incomplete stubs, like originally proposed in #8955.

----

### Incomplete annotations

When submitting new stubs, it is not necessary to annotate all arguments, return types, and fields. Such items should either be left unannotated or use `_typeshed.Incomplete` if this is not possible:

```python

field: Incomplete  # unannotated 

def foo(x): ...  # unannotated argument and return type
```

`Incomplete` can also be used for partially known types:

```python
def foo(x: Incomplete | None = None) -> list[Incomplete]: ...
```

### `Any` vs. `Incomplete`

While `Incomplete` is a type alias of `Any`, they serve difference purposes: `Incomplete` is a placeholder where a proper type might be substituted. It's a "to do" item and should be replaced if possible. `Any` is used when it's not possible to accurately type an item using the current type system. It should be used sparingly.

srittau:
So, any input by one of the co-maintainers? @JelleZijlstra @AlexWaygood @hauntsaninja @Akuli 

I personally like the short version in my last comment.

AlexWaygood (approved):
Happy with either. @hmc-cs-mdrissi's comments about why he's doing TensorFlow stubs the way he is have resolved my prior concerns about the current wording. Your alternative also looks fine :)

CONTRIBUTING.md
```diff
@@ -524,6 +490,39 @@ into any of those categories, use your best judgement.
...
   or getter/setter methods (e.g. `HasItems`, `HasFileno`).
 
+### Incomplete stubs
+
+We accept partially annotated stubs, especially for larger packages.
+These need to follow the following guidelines:
+
+* A package must include all public submodules, all public classes,
+  functions and methods, and all other public fields.
+* Included functions and methods must list all arguments, but the argument
```
Akuli, line 500: I imagine these rules could seem arbitrary and complicated to someone new to typeshed. We should mention that the `create_baseline_stubs` script generates stubs that follow these rules.

Akuli (commented):
Good to see our conventions being documented :) A couple nits.

CONTRIBUTING.md
srittau: @Akuli What do you think about the proposal here: https://github.com/python/typeshed/pull/9548#issuecomment-1386880372 This is much simpler.

Akuli: I like it, it is less confusing :)

srittau:
I have now substituted and concise version.

CONTRIBUTING.md
```diff
@@ -524,6 +490,27 @@ into any of those categories, use your best judgement.
...
+
+`Incomplete` can also be used for partially known types:
+
+```python
+def foo(x: Incomplete | None = None) -> list[Incomplete]: ...
+```
+
+### `Any` vs. `Incomplete`
+
+While `Incomplete` is a type alias of `Any`, they serve difference purposes: `Incomplete` is a placeholder where a proper type might be substituted. It's a "to do" item and should be replaced if possible. `Any` is used when it's not possible to accurately type an item using the current type system. It should be used sparingly.
```
AlexWaygood, line 513: Nit: can we wrap to <90 characters where possible?

AlexWaygood (approved):

PR #9551, opened 2023-01-16
Reenable flake8-pyi's Y011 and Y015

AlexWaygood:
Cc. @Avasam for `PyScreeze` and `pywin32`; cc. @hmc-cs-mdrissi for `TensorFlow`.

Following the new release of flake8-pyi, default values for parameters are allowed in many situations. The new rules are that default values are allowed if the default value is an `int`, a `float`, a complex number, a string, a `bytes` object, a `bool`, `None`, or `...`.

For other default values, we'll continue to use `...`. This has two advantages:

- It keeps stubs concise and readable.
- It means that stubtest will be easily able to verify any default values that we do include in the stubs, something which it might struggle with if we were to include more complex default values.


.flake8
```diff
@@ -23,22 +23,19 @@
 #            Currently can't be enabled due to a few lingering bugs in mypy regarding
 #            PEP 604 type aliases (see #4819).
 
-# Outdated rules in flake8-pyi that should possibly be deprecated altogether:
-#     Y011   All default values for typed function arguments must be `...`
-#     Y015   Attribute must not have a default value other than `...`
 #
```
AlexWaygood, line 26:
```suggestion
```

hmc-cs-mdrissi (approved):
Tensorflow side looks good to me. I'll miss enums a bit, but I can see understanding enums would likely be too complex for a static checker like flake8-pyi

srittau (approved):

Avasam:
Awesome. I was waiting for this to be fully supported to update the stubs I contributed and use.

PR #9557, opened 2023-01-17
Use ParamSpec for `singledispatch` and `singledispatchmethod`

iyume:
Since ParamSpec is introduced in PEP, the type hints for `functools.singledispatch` and `functools.singledispatchmethod` can be more precious.

The change is like:

```python
from functools import singledispatch, singledispatchmethod
from typing import Union
from typing_extensions import reveal_type


@singledispatch
def foo(tp: Union[int, str], verbose: bool) -> int:
    ...


@foo.register
def foo_int(tp: int) -> int:
    # pylance: Argument of type "(tp: int) -> int" cannot be assigned to parameter "cls" of type "(_T1@register, verbose: bool) -> int" in function "register"
    # Type "(tp: int) -> int" cannot be assigned to type "(_T1@register, verbose: bool) -> int"
    # Function accepts too many positional parameters; expected 1 but received 2
    ...


@foo.register
def foo_str(tp: str, verbose: bool) -> int:
    ...


@foo.register
def foo_error(tp: complex, verbose: bool) -> int:
    # type checker should check the subtype of first parameter
    ...


reveal_type(foo)  # Type of "foo" is "_SingleDispatchCallable[int | str, (verbose: bool), int]"
reveal_type(foo_str)  # Type of "foo_str" is "(str, verbose: bool) -> int"
reveal_type(foo_error)  # Type of "foo_error" is "(complex, verbose: bool) -> int"


foo(1)  # pylance: Argument missing for parameter "verbose"


reveal_type(foo.dispatch(int))  # Type of "foo.dispatch(int)" is "(int, verbose: bool) -> int"


class A:
    @singledispatchmethod
    def foo(self, tp: Union[int, str], verbose: bool) -> int:
        ...

    @foo.register
    def foo_int(self, tp: int, verbose: bool) -> int:
        ...


reveal_type(A.foo)  # Type of "A.foo" is "(A, int | str, verbose: bool) -> int"
reveal_type(A().foo)  # Type of "A().foo" is "(int | str, verbose: bool) -> int"
reveal_type(A.foo_int)  # Type of "A.foo_int" is "(A, int, verbose: bool) -> int"

```


AlexWaygood:
Thanks for the PR, but unfortunately, it looks like this causes a lot of false positives to be emitted on user code :/

The semantics of using `Concatenate` on methods (rather than functions) are somewhat poorly defined in the typing docs and PEPs, unfortunately, which leads to quite a few situations like this. `singledispatch` and `singledispatchmethod` will also ultimately need to be special-cased by type checkers quite a lot anyway, so we should think very carefully before complicating the stubs for these decorators.

Issue #9550, opened 2023-01-16
Older stubs still explicitly using `Any` rather than `Incomplete`


Avasam:
Thanks to typeshed-stats (#9386), I was able to make the following list of stubs majorly using `Any`, indicating that they've likely not been switched over to `Incomplete` yet.
Ordered by ratio of "explicit `Any` / explicit `Incomplete` + 1" parameters and variables.
Excludes stubs marked as obsolete and ratio <= 1 rounded down.

|Package Name|Explicit `Any`|Explicit `Incomplete`|Ratio
|-|-|-|-
|boto|633|0|633
|oauthlib|392|0|392
|passlib|278|0|278
|braintree|506|1|253
|PyYAML|244|0|244
|psycopg2|224|0|224
|Pygments|217|0|217
|html5lib|207|0|207
|stripe|203|0|203
|mock|157|0|157
|protobuf|153|0|153
|commonmark|141|0|141
|babel|134|0|134
|vobject|114|0|114
|beautifulsoup4|103|0|103
|dateparser|101|0|101
|PyMySQL|100|0|100
|Markdown|100|0|100
|httplib2|94|0|94
|aws-xray-sdk|79|0|79
|humanfriendly|76|0|76
|Pillow|397|5|66
|waitress|64|0|64
|invoke|98|1|49
|python-jose|48|0|48
|docutils|47|0|47
|jsonschema|45|0|45
|google-cloud-ndb|259|5|43
|python-gflags|41|0|41
|pyflakes|37|0|37
|setuptools|470|12|36
|mysqlclient|35|0|35
|jmespath|33|0|33
|parsimonious|28|0|28
|opentracing|26|0|26
|regex|25|0|25
|toml|22|0|22
|decorator|22|0|22
|python-xlib|20|0|20
|polib|18|0|18
|requests|64|3|16
|paramiko|46|2|15
|pyserial|14|0|14
|simplejson|14|0|14
|stdlib|2598|186|14
|pynput|12|0|12
|slumber|11|0|11
|console-menu|11|0|11
|SQLAlchemy|2962|301|10
|Flask-Cors|9|0|9
|six|9|0|9
|singledispatch|8|0|8
|caldav|31|3|8
|xmltodict|7|0|7
|paho-mqtt|6|0|6
|redis|549|100|5
|ldap3|557|104|5
|JACK-Client|5|0|5
|typed-ast|5|0|5
|Send2Trash|4|0|4
|pep8-naming|4|0|4
[...]

PR #9558, opened 2023-01-17
`stubs/` directory: replace `Any | None = ...` with `Incomplete | None = ...`

AlexWaygood:
See #9550 for context.

The first commit was produced just by doing this in the Python REPL:

```pycon
>>> from pathlib import Path
>>> for path in Path("stubs").rglob("*.pyi"):
...     source = path.read_text()
...     if ": Any | None = ..." in source:
...         source = "\n".join(["from _typeshed import Incomplete"] + source.splitlines())
...         source = source.replace(": Any | None = ...", ": Incomplete | None = ...")
...         _ = path.write_text(source)
...
>>> exit()
```

The second commit applies pycln, then isort, then black.


srittau (approved):
Seems fine.

PR #9560, opened 2023-01-18
Tensorflow Stubs 2

hmc-cs-mdrissi:
# Summary
This fills in more of [tf.math](https://www.tensorflow.org/api_docs/python/tf/math). This roughly covers half of tf.math functions focusing on more commonly used ones. Most of the signatures are similar with overloads for three types of tensors, normal, ragged, sparse. Most functions support ragged, few support sparse. I tested each in repl by doing,

```python
>>> import tensorflow as tf
>>> x = tf.constant([[1., 0.], [1., 2.]])
>>> r = tf.RaggedTensor.from_tensor(x)
>>> sp = tf.sparse.from_dense(x)
>>> tf.math.add(x, x)
>>> tf.math.add(r, r)
>>> tf.math.add(sp, sp)
```

and repeating for various operations/tensor types.


stubs/tensorflow/tensorflow/__init__.pyi
```diff
@@ -11,7 +11,42 @@ import numpy
 from tensorflow.dtypes import *
 
 # Most tf.math functions are exported as tf, but sadly not all are.
-from tensorflow.math import abs as abs
+from tensorflow.math import (
```
hmc-cs-mdrissi, line 14: Commenting out this import fixes pytype crash. I'm unsure what actual fix for it though is.

AlexWaygood: What happens if you comment out just the `pow as pow` line?

hmc-cs-mdrissi: Tested and same error. Not specific to pow. `from tensorflow.math import add as add` by itself is enough to trigger crash.

Playing with it I found a workaround. I think the error is caused by import cycles and using wildcard import to define DType. If I explicitly import DType from tensorflow.dtypes the crash is fixed. @rchen152

AlexWaygood: Thanks! For context: I wondered if `pow` might be the problem because of the fact that pytype uses a custom `builtins` stub rather than the one in typeshed -- this can often cause surprising errors in typeshed's CI. But great that you found the actual problem here, and a workaround :)

rchen152: Odd, typeshed definitely uses `import *` elsewhere, so I wonder why pytype was unhappy here. I'll look into it.

JelleZijlstra:
(Adding myself to remind myself to take a look, but any other maintainer should feel free to proceed without my review if they get here first.)

JelleZijlstra (approved):
Didn't review closely but from looking at a few functions this looks good.

PR #9561, opened 2023-01-18
asyncio: fix default for subprocess_exec

JelleZijlstra:
It must be False, not True.

https://github.com/python/cpython/blob/3.11/Lib/asyncio/base_events.py#L1675


stdlib/asyncio/base_events.pyi
```diff
@@ -399,7 +399,7 @@ class BaseEventLoop(AbstractEventLoop):
         stdout: int | IO[Any] | None = -1,
         stderr: int | IO[Any] | None = -1,
         universal_newlines: Literal[False] = False,
-        shell: Literal[True] = ...,
+        shell: Literal[False] = ...,
```
srittau, line 402:
```suggestion
        shell: Literal[False] = False,
```

srittau (approved):

PR #9546, opened 2023-01-16
Update fpdf2 stubs to 2.6.1

srittau:
Closes #9542

PR #9562, opened 2023-01-18
[fpdf2] Restore string default values

srittau:
Cf. #9546


AlexWaygood (approved):
Haven't double-checked the source, but I assume these are correct :)

PR #9563, opened 2023-01-18
Improve pre-commit config

AlexWaygood:
- Add a few more hooks. These are all very fast, and I've found them useful in other projects:
  - Autofixes:
    - `trailing-whitespace`: fixes trailing whitespace
    - `requirements-txt-fixer`: alphabetises items in `requirements.txt` files
    - `end-of-file-fixer`: makes sure every file ends with a single newline character
    - `mixed-line-ending`: Makes sure Windows users don't accidentally introduce CRLF line endings into a file that uses LF line endings
  - None-autofixes:
    - `check-yaml`: loads YAML files to validate syntax
    - `check-toml`: loads TOML files to validate syntax
    - `check-merge-conflict`: detects merge-conflict strings in files and blocks them from accidentally being committed
    - `check-case-conflict`: checks for files with names that would conflict on a case-insensitive filesystem like MacOS HFS+ or Windows FAT; blocks them from being committed.
  - Change the bot schedule to quarterly, to reduce noisy PRs
  - Change the `black` language target-version to Python 3.10, synching the setting here with the changes that were made to our `pyproject.toml` file in #7538


tests/check_consistent.py
```diff
@@ -220,8 +221,9 @@ def get_precommit_requirements() -> dict[str, SpecifierSet]:
         if not repo.get("python_requirement", True):
             continue
         hook = repo["hooks"][0]
-        package_name, package_rev = hook["id"], repo["rev"]
-        package_specifier = SpecifierSet(f"=={package_rev.removeprefix('v')}")
+        package_name = Path(urllib.parse.urlparse(repo["repo"]).path).name
```
JelleZijlstra, line 224: Is this meant to be part of this PR?

AlexWaygood: Yes. The test as it currently stands is poorly designed. Without this change, `check_consistent.py` fails if we add `pre-commit-hooks` to our pre-commit config, because of the fact that the hooks in the `pre-commit-hooks` repo have different names to the repo as a whole.

AlexWaygood: Example failing run: https://github.com/python/typeshed/actions/runs/3949452817/jobs/6760681897

JelleZijlstra: I see, thanks!

PR #9564, opened 2023-01-18
`parsimonious`: Update return types of `NodeVisitor`'s methods

Four-Velocity:
According to the [parsimonious/nodes.pyi](https://github.com/python/typeshed/blob/65c4ddfc57367a46c66e4f4d2a8f76e159c0a3a2/stubs/parsimonious/parsimonious/nodes.pyi) `NodeVisitor.visit` should return `Any`, while `NodeVisitor.parse` and `NodeVisitor.match` should return `Node`.

According to the [code and docstrings of the parsimonious](https://github.com/erikrose/parsimonious/blob/0.10.0/parsimonious/nodes.py#L244) `parse` and `match` methods are just a shortcuts for the `visit` method:

>``SomeVisitor().parse('some_string')`` is a shortcut for
   ``SomeVisitor().visit(some_grammar.parse('some_string'))``.

> ``SomeVisitor().match('some_string')`` is a shortcut for
    ``SomeVisitor().visit(some_grammar.match('some_string'))``.

So the return type of the `parse` and `match` should be the same as for the `visit`


Four-Velocity:
Ok, perhaps I need some time or help to figure out how to specify it correctly 🙃

stubs/parsimonious/parsimonious/nodes.pyi
```diff
@@ -26,15 +26,15 @@ class RegexNode(Node):
 
 class RuleDecoratorMeta(type): ...
 
-_VisitorResultT = TypeVar("_VisitorResultT")
+_VisitRT = Any
 
 class NodeVisitor(metaclass=RuleDecoratorMeta):
```
JelleZijlstra, line 32: Haven't looked at the code but I suspect this class should be made generic.

Four-Velocity: Sorry, crazy week
You mean that `_VisitRT` should be `Generic` instead of `TypeAlias`?

Four-Velocity: Oh, ok, think i figured it out
Like that?

```python
_VisitorResultT = TypeVar("_VisitorResultT")

class NodeVisitor(Generic[_VisitorResultT], metaclass=RuleDecoratorMeta):
    ...
    def visit(self, node: Node) -> _VisitorResultT: ...
    ...
```

Four-Velocity:
Hey @JelleZijlstra 👋 
Looks like it's done

JelleZijlstra (approved):

PR #9565, opened 2023-01-18
Stubs replace `: Any | None` with `: Incomplete | None`

Avasam:
First commit does  `: Any | None\n` to `: Incomplete | None\n`
Second commit does  `: Any | None` to `: Incomplete | None`

Split them up for ease of review, just in case. With #9558 I think we can effectively say that all  `: Any | None` were to be replaced.

The only instances of `Any | None` left are ClassVar type parameter, return types and 2 Aliases because of name clash:
- `tAny | None` in stubs\protobuf\google\protobuf\internal\well_known_types.pyi
- `_Any | None` in stubs\SQLAlchemy\sqlalchemy\dialects\postgresql\array.pyi

Ref #9550


stubs/SQLAlchemy/sqlalchemy/sql/selectable.pyi
```diff
@@ -191,7 +191,7 @@ class Values(Generative, FromClause):
     name: Any
     literal_binds: Any
     def __init__(self, *columns, **kw) -> None: ...
-    def alias(self: Self, name: Any | None, **kw) -> Self: ...  # type: ignore[override]
+    def alias(self: Self, name: Incomplete | None, **kw) -> Self: ...  # type: ignore[override]
```
AlexWaygood, line 194: None of the ones in this file are stubgen artefacts; they were all added in https://github.com/python/typeshed/pull/7603/files. _But_, there's no source-code comment indicating why they used `Any` here (and the comments on the PR thread give no insight), so I think it's fine to change the `Any` uses in this file to `Incomplete`.

stubs/pyinstaller/PyInstaller/lib/modulegraph/modulegraph.pyi
```diff
@@ -2,15 +2,16 @@
...
+from typing import Protocol
 
 class _SupportsGraphident(Protocol):
     graphident: str
 
 # code, filename and packagepath are always initialized to None. But they can be given a value later.
 class Node:
     # Compiled code. See stdlib.builtins.compile
-    code: Any | None
+    code: Incomplete | None
```
AlexWaygood, line 14: You only added this one recently and, from the comment, it looks like it's meant to be `Any` ;)

stubs/setuptools/pkg_resources/__init__.pyi
```diff
@@ -79,7 +79,7 @@ class Requirement:
     specs: list[tuple[str, str]]
     # TODO: change this to packaging.markers.Marker | None once we can import
     #       packaging.markers
-    marker: Any | None
+    marker: Incomplete | None
```
AlexWaygood, line 82: Oh, we should, you know, take action on this comment now (in a separate PR)

AlexWaygood:
(The only actionable thing in my review is the `pyinstaller` comment.)

stubs/pyinstaller/PyInstaller/lib/modulegraph/modulegraph.pyi
Avasam: Definitely, I don't see anything else obvious after manually reviewing as well.

JelleZijlstra: I haven't checked the code but it looks like this should just be `types.CodeType`.

AlexWaygood: This PR no longer touches this line, so I think that can be deferred to another PR

Avasam: It's the return of `stdlib/builtins.pyi > compile`, which is currently typed as `Any`, but also has this comment:
> # TODO: `compile` has a more precise return type in reality; work on a way of expressing that?

AlexWaygood: - #9567

JelleZijlstra: Right, I'd be surprised if PyInstaller is using PyCF_ONLY_AST here.

Avasam: (technically it's modulegraph, pyinstaller vendors it) It is using PyCF_ONLY_AST  as well, but it's the non-ast call that is stored.

PR #9567, opened 2023-01-18
builtins: do the TODO on compile()

JelleZijlstra:


stdlib/builtins.pyi
```diff
@@ -1220,27 +1220,90 @@ if sys.version_info >= (3, 10):
...
+    def compile(
+        source: str | ReadableBuffer | _ast.Module | _ast.Expression | _ast.Interactive,
+        filename: str | ReadableBuffer | _PathLike[Any],
+        mode: str,
+        flags: Literal[1024],
+        dont_inherit: int = False,
+        optimize: int = -1,
+        *,
+        _feature_version: int = -1,
+    ) -> _ast.AST: ...
```
JelleZijlstra, line 1258: We can potentially be more precise here: if mode is "eval", the node must be an `_ast.expr`.

JelleZijlstra:
> 
> ```diff
> sympy (https://github.com/sympy/sympy)
> + sympy/parsing/sympy_parser.py:1075: error: Incompatible types in assignment (expression has type "CodeType", variable has type "str")  [assignment]

https://github.com/sympy/sympy/blob/3450b7573138badc277b238118d088b7d842594a/sympy/parsing/sympy_parser.py#L1075

This is a side-effect of more precise types; previously mypy allowed the redefinition because compile() returned Any.

> 
> jinja (https://github.com/pallets/jinja)
> + src/jinja2/environment.py:704: error: Unused "type: ignore" comment
> ```

https://github.com/pallets/jinja/blob/main/src/jinja2/environment.py#L704

This fixes a false positive from mypy.

stdlib/builtins.pyi
AlexWaygood: Would we have to duplicate all 8 overloads from `ast.parse`, or does `compile` only take "exec", "eval" or "single" for the `mode` parameter?

JelleZijlstra: Yes, we would. `ast.parse` just calls `compile()`. I think it's fine to stick with `AST` for simplicity.

JelleZijlstra: Though I think we'd only need four overloads because the second argument to `compile()` is required, unlike `ast.parse`.

PR #9570, opened 2023-01-20
stdlib: Add bound Hashable for keys of mappings

kruvasyan:


AlexWaygood:
Thanks for the PR, but as you can see from the failing CI and the large output from mypy_primer, this would currently cause far too many false-positive errors in user code if we were to merge this :)

A5rocks:
Hello! Quite a bit of the results *would* be solved by https://github.com/python/mypy/pull/11478 except that's marked as closed. I see that there's a single blocking review comment -- is it possible for someone to pick it back up?

hauntsaninja:
I can't re-open that PR because of some issues with sobolevn's repo. Want to open a new PR?

A5rocks:
I could! I'm afraid I'm not familiar with this part of mypy so I wouldn't be to resolve issues (quickly, at least). Should I anyways?

kruvasyan:
@A5rocks I turned it into a [tracking issue](https://github.com/python/typeshed/issues/9571) and pointed out all the packages with problems. We can gradually fix the bugs and mark the completed

hauntsaninja:
@A5rocks yes please! At least it'll help prevent it being totally lost to time (no one is going to really dig up closed PRs, but open ones still stand a chance)

Issue #8988, opened 2022-10-26
Should stubs contain default values?


hauntsaninja:
This has come up recently, for similar reasons as #4881 and discussed a little there

From @gramster
> an advantage of including default values is that tools like pyright or pylance will use stubs (if available) for the signatures they show to users when hovering over a function call, and users want to see the default values. I'd like to understand your point of view though as to why you would prefer them removed?

I'm overall fine with this, though I'd like to see the following:
- we restrict it to simple literal values (e.g. anything that can go into a Literal)
- we can still use `...` at author's discretion (useful in overloads or cases where value doesn't actually matter or to avoid awful `x: Literal["x"] = "x"` things)
- we update e.g. stubtest to verify stuff

The case for:
- lets IDEs provide more information to users

The case against:
- most things are longer than `...` and terseness may be more valuable than the additional information
- it's a thing that needs doing and maintaining
- if we allow complicated default values there are more costs

A similar thing could apply to constants.

JelleZijlstra:
I agree, a couple of small comments:

> * we restrict it to simple literal values (e.g. anything that can go into a Literal)

Other values that we could accept would include floats and types (e.g., for params like `dict_factory: Callable[..., ...] = dict`).

> * we can still use `...` at author's discretion (useful in overloads or cases where value doesn't actually matter or to avoid awful `x: Literal["x"] = "x"` things)
> * we update e.g. stubtest to verify stuff

It should be easy for stubtest to check defaults for correctness on Python functions (it's in the signature). For C functions, it's not necessarily possible to infer the default though, and indeed there may not be a default in a useful sense.

What other tools would need updating? I think we'd have to change flake8-pyi and hopefully nothing else.

srittau:
It would be useful - but not necessary - to update stubgen as well.

AlexWaygood:
I like the idea of adding simple default values to stubs. I agree that it's very useful in IDEs, and I think it'll be easier to maintain than adding docstrings to stubs.

The `= ...` convention that we currently use is nice because it keeps stubs as concise as possible. But... it looks _really weird_ to people who aren't used to it. I think it's nice for stubs to use the same syntax and conventions as normal Python wherever possible, and I remember finding the `= ...`s everywhere really strange and alien when I first started reading through typeshed's stubs.

> What other tools would need updating? I think we'd have to change flake8-pyi and hopefully nothing else.

[...]

PR #9572, opened 2023-01-20
Fix a few issues with parameter defaults

AlexWaygood:
Some of these vary depending on the machine you're running on (the ones in `sysconfig.pyi`, `platform.pyi`, `ctypes/__init__.pyi`, `multiprocessing/heap.pyi`, `pydoc.pyi`). We might be able to give slightly better defaults for some of these if https://github.com/PyCQA/flake8-pyi/pull/336 (or something like it) is merged.

Some others currently have integer defaults in the stub, but enum defaults at runtime, causing stubtest to complain if you use mypy `master` (the ones in `ssl.pyi`, `socket.pyi`, `asyncore.pyi`).

The `email.parser` ones just need some method overrides in the stub that weren't necessary before we added default values.

Refs https://github.com/python/typeshed/issues/8988#issuecomment-1398366699, and following comments.


AlexWaygood:
I haven't changed any of the more opinionated ones disallowed by https://github.com/PyCQA/flake8-pyi/pull/336, such as a few numbers which are simply very long -- I figured we should let the discussion on that PR finish first :)

stdlib/platform.pyi
```diff
@@ -37,9 +37,7 @@ def java_ver(
     release: str = "", vendor: str = "", vminfo: tuple[str, str, str] = ..., osinfo: tuple[str, str, str] = ...
 ) -> tuple[str, str, tuple[str, str, str], tuple[str, str, str]]: ...
 def system_alias(system: str, release: str, version: str) -> tuple[str, str, str]: ...
-def architecture(
```
JelleZijlstra, line 40: part of me wants to keep this as an easter egg

JelleZijlstra (approved):

stdlib/platform.pyi
AlexWaygood: It's iconic

PR #9574, opened 2023-01-20
create_datagram_endpoint expects a str when family=AF_UNIX

gandalf3:
Since python 3.7, [`asyncio.loop.create_datagram_endpoint`](https://docs.python.org/3.8/library/asyncio-eventloop.html#asyncio.loop.create_datagram_endpoint) expects the arguments `local_addr` and `remote_addr` to be `str` or `None` when the argument `family` is `socket.AF_UNIX` (and if when `sock` is not `None`, it is also a unix socket). Here is a link to the relevant line in the standard library: https://github.com/python/cpython/blob/main/Lib/asyncio/base_events.py#L1319

Here is a link to the pull request on the cpython repository which effected this change: https://github.com/python/cpython/pull/3164

i'm not at all sure if I've made the changes correctly. is there a way to represent the dependency on the `family` argument that i'm missing?

### DIFF
--- a/stdlib/asyncio/events.pyi
+++ b/stdlib/asyncio/events.pyi
@@ -512,8 +512,8 @@ class AbstractEventLoop:
         async def create_datagram_endpoint(
             self,
             protocol_factory: Callable[[], _ProtocolT],
-            local_addr: tuple[str, int] | None = None,
-            remote_addr: tuple[str, int] | None = None,
+            local_addr: tuple[str, int] | str | None = None,
+            remote_addr: tuple[str, int] | str | None = None,
             *,
             family: int = 0,
             proto: int = 0,
--- a/stdlib/asyncio/events.pyi
+++ b/stdlib/asyncio/events.pyi
@@ -491,38 +491,21 @@ class AbstractEventLoop:
     async def sendfile(
         self, transport: WriteTransport, file: IO[bytes], offset: int = 0, count: int | None = None, *, fallback: bool = True
     ) -> int: ...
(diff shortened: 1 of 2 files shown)


srittau:
Please note that Python 3.7 is the minimum version we support in typeshed, so there's no need for an extra branch for Python >= 3.7.

gandalf3:
thanks, i think i've fixed it now. i was going off the versions file which [lists 3.4 as the minimum version](https://github.com/python/typeshed/blob/main/stdlib/VERSIONS#L65) but i am probably reading it wrong.

stdlib/asyncio/events.pyi
```diff
@@ -491,21 +491,38 @@ class AbstractEventLoop:
...
+            remote_addr: tuple[str, int] | str | None = None,
+            *,
+            family: int = 0,
+            proto: int = 0,
+            flags: int = 0,
+            reuse_address: bool | None = None,
+            reuse_port: bool | None = None,
+            allow_broadcast: bool | None = None,
+            sock: socket | None = None,
+        ) -> tuple[DatagramTransport, _ProtocolT]: ...
```
AlexWaygood, line 525: This branch also isn't needed

gandalf3: omg my bad just woke up thanks for catching!

AlexWaygood: No worries!

srittau (approved):

srittau:
Thanks!

PR #9577, opened 2023-01-22
Add SSLContext.security_level property for Python 3.10

sethmlarson:
Python 3.10 added the property `security_level` which returns an `int` and is read-only.


srittau (approved):
Thanks!

PR #9583, opened 2023-01-24
Upgrade bleach stubs for 6.0

srittau:
Add/improve some annotations


srittau:
Closes: #9582

stubs/bleach/bleach/linkifier.pyi
```diff
@@ -28,21 +21,34 @@ class Linker:
...
-    url_re: Any
-    email_re: Any
+    url_re: Pattern[str]
+    email_re: Pattern[str]
     def __init__(
-        self, source, callbacks=..., skip_tags: Container[str] | None = ..., parse_email: bool = ..., url_re=..., email_re=...
+        self,
+        source,
+        callbacks: Iterable[_Callback] | None = ...,
+        skip_tags: Container[str] | None = None,
```
AlexWaygood, line 42: `callbacks` and `skip_tags` need to be sized in order for [these lines](https://github.com/mozilla/bleach/blob/13d6c0c91d2b1bef52ad2b5a2812f76dda18298f/bleach/linkifier.py#L238-L239) to work (otherwise there's a chance that the iterable will be immediately consumed). So I think `Collection` would be better for both of these:

```suggestion
        callbacks: Collection[_Callback] | None = ...,
        skip_tags: Collection[str] | None = None,
```

(The same applies for the attribute annotations immediately above)

stubs/bleach/bleach/linkifier.pyi
```diff
@@ -28,21 +21,34 @@ class Linker:
...
     ) -> None: ...
-    def __getattr__(self, item: str) -> Incomplete: ...
+    def apply_callbacks(self, attrs, is_new): ...
+    def extract_character_data(self, token_list): ...
+    def handle_email_addresses(self, src_iter): ...
+    def strip_non_url_bits(self, fragment): ...
+    def handle_links(self, src_iter): ...
+    def handle_a_tag(self, token_buffer): ...
+    def extract_entities(self, token): ...
+    def __iter__(self): ...
```
AlexWaygood, line 54:
```suggestion
    def __iter__(self) -> Iterator[Incomplete]: ...
```

stubs/bleach/bleach/sanitizer.pyi
```diff
@@ -1,65 +1,77 @@
...
         attributes: _Attributes = ...,
-        allowed_protocols: Container[str] = ...,
-        strip_disallowed_elements: bool = ...,
-        strip_html_comments: bool = ...,
-        css_sanitizer: CSSSanitizer | None = ...,
-        **kwargs,
+        allowed_protocols: Iterable[str] = ...,
+        attr_val_is_uri: frozenset[tuple[str | None, str]] = ...,
+        svg_attr_val_allows_ref: frozenset[tuple[str | None, str]] = ...,
+        svg_allow_local_href: frozenset[tuple[str | None, str]] = ...,
```
AlexWaygood, line 71: These are the same type as the `_Attr` alias that you added in `bleach/callbacks.pyi`, and I don't think that's a coincidence. Maybe the `_Attr` alias you added there should be moved to `bleach/__init__.pyi`, renamed as `_HTMLAttr`, and imported here as well?

AlexWaygood (commented):
Mostly looks good!

stubs/bleach/bleach/linkifier.pyi
srittau: A simple `if iter:` shouldn't consume an iterable. In the general case it should be equivalent with `if iter is not None:`.

AlexWaygood: Good point, my reasoning about the iterable possibly being consumed was bogus. I still think a sized iterable is expected here, though. Unsized iterables are always truthy, even if they're empty:

```pycon
>>> x = iter([])
>>> bool(x)
True
```

AlexWaygood: Actually maybe you're right and they're just guarding against `None` being passed in. Feel free to ignore me :)

AlexWaygood (approved):

PR #9584, opened 2023-01-24
Complete `croniter` stubs

AlexWaygood:
Fix a few complaints that stubtest will have when the next release of mypy comes out. Delete the stubtest allowlist while we're about it.


hauntsaninja (approved):
Thanks!

PR #9585, opened 2023-01-24
Fix some stubtest complaints before they happen

AlexWaygood:
This adds some missing objects to various stdlib and third-party stubs. Stubtest doesn't complain about these now, but will do when mypy 1.0 is released; this should make upgrading to mypy 1.0 easier when that happens.


hauntsaninja (approved):
Nice, thanks!

PR #9587, opened 2023-01-25
codecs: _buffer_encode returns (bytes, int)

adamjorr:
_buffer_encode should return a tuple[bytes, int], not just bytes. See https://github.com/python/cpython/blob/main/Lib/codecs.py#L231-L234 for the interface definition.

I think even this typing is not fully correct since the interface anticipates bytes -> bytes and str -> str encoders, but that's a separate issue.


srittau (approved):
Thanks!

srittau:
Maybe it would be best to return `tuple[Any, int]`, but I'm fine with the PR as is as it's strictly better than the current annotation.

adamjorr:
> Maybe it would be best to return `tuple[Any, int]`, ...

I agree, but then I think we should also consider changing `encode` to accept and return Any and it becomes a bigger change than I want to tackle, and possibly makes it less useful for the majority use case. Happy for someone else to take it on and run with it, though!

PR #9592, opened 2023-01-26
Add more overloads to the `re` stubs to help out pyright

### LINKED ISSUE #9591: re.match(), re.search() and re.fullmatch() cannot be used with AnyStr
Given this test file:

```python
from typing import AnyStr

from re import Match, Pattern

def check_re_search(pattern: Pattern[AnyStr], string: AnyStr) -> Match[AnyStr]:
    match = pattern.search(string)
    if match is None:
        raise ValueError(f"'{string!r}' does not match {pattern!r}")
    return match
```

pyright reports the following error:

```
No configuration file found.
No pyproject.toml file found.
stubPath /home/agateau/tmp/typings is not a valid directory.
Assuming Python platform Linux
Searching for source files
Found 1 source file
pyright 1.1.291
/home/agateau/tmp/crs.py
  /home/agateau/tmp/crs.py:6:21 - error: Could not bind method "search" because "Pattern[AnyStr@check_re_search]" is not assignable to parameter "self"
    "Pattern[AnyStr@check_re_search]" is incompatible with "Pattern[str]"
      TypeVar "AnyStr@Pattern" is invariant
        Type "AnyStr@check_re_search" cannot be assigned to type "str" (reportGeneralTypeIssues)
  /home/agateau/tmp/crs.py:6:21 - error: Could not bind method "search" because "Pattern[AnyStr@check_re_search]" is not assignable to parameter "self"
    "Pattern[AnyStr@check_re_search]" is incompatible with "Pattern[bytes]"
      TypeVar "AnyStr@Pattern" is invariant
        Type "AnyStr@check_re_search" cannot be assigned to type "bytes" (reportGeneralTypeIssues)
```

The same code works fine if one replaces all `AnyStr` with either `str` or `bytes`.

I originally reported this on pyright issue tracker (microsoft/pyright#4534) but it seems the issue is actually here (see https://github.com/microsoft/pyright/issues/4534#issuecomment-1405208232)

Quoting erictraut comment:

> The fix for this would require an additional overload within the typeshed stubs. Something like this:
>
> ```python
>    @overload
>    def search(self, string: AnyStr | ReadableBuffer, pos: int = ..., endpos: int = ...) -> Match[AnyStr] | None: ...
> ```

AlexWaygood:
Fixes #9591


AlexWaygood:
#9593 is a simpler approach, but one which mypy might not like so much (let's see what mypy_primer says).

AlexWaygood:
Mypy hates #9593, so I think this is the best we can do.

test_cases/stdlib/check_re.py
```diff
@@ -0,0 +1,23 @@
...
+
+import re
+import typing as t
+from typing_extensions import assert_type
+
+
+def check_search(str_pat: re.Pattern[str], bytes_pat: re.Pattern[bytes]) -> None:
+    assert_type(str_pat.search("x"), t.Optional[t.Match[str]])
+    assert_type(bytes_pat.search(b"x"), t.Optional[t.Match[bytes]])
+    assert_type(bytes_pat.search(bytearray(b"x")), t.Optional[t.Match[bytes]])
```
JelleZijlstra, line 11: This is an unusual case because of the implicit promotion, maybe also try a type that isn't implicitly promoted (e.g., `pickle.PickleBuffer`).

AlexWaygood: _You_ added this test in #7679, I'm just moving it to a different file in this PR ;)

But I'll make the change!

JelleZijlstra (approved):

PR #9593, opened 2023-01-26
[alt] Improve `re` overloads to help out pyright

### LINKED ISSUE #9591: re.match(), re.search() and re.fullmatch() cannot be used with AnyStr

AlexWaygood:
A simpler approach than #9592, but let's see what mypy_primer says. Fixes #9591


AlexWaygood:
Doesn't look good.

PR #9596, opened 2023-01-27
Fix all fixable stubtest_allowlist entries in SQLAlchemy

Avasam:
This PR focuses solely on fixing as many stubtest_allowlist entries as possible (unless they're caused by an issue with subtest itself).
Also regrouped and reorganized some entries while updating comments.


stubs/SQLAlchemy/sqlalchemy/orm/strategy_options.pyi
```diff
@@ -23,44 +24,116 @@ class Load(Generative, LoaderOption):
...
+class loader_option(Generic[_F]):
+    name: str
+    _dynamic: _F
+    fn: _F
+    __call__: _F  # Cheesy "__call__" definition to use the dynamic methods instead
+
+class _contains_eager(Protocol):
+    def __call__(self, loadopt, attr, alias: Incomplete | None = ...) -> loader_option[_contains_eager]: ...
+
+contains_eager: loader_option[_contains_eager]
```
Avasam, line 74: This is the workaround I'm mentioning in the comment above:
```suggestion
class _contains_eager(Protocol):
    def __call__(self, loadopt, attr, alias: Incomplete | None = ...) -> loader_option[_contains_eager]: ...

def contains_eager(loadopt, attr, alias: Incomplete | None = ...) -> loader_option[_contains_eager]: ...

contains_eager = loader_option[_contains_eager]()  # pyright: ignore[reportGeneralTypeIssues]  # noqa: F811
```

As you can see it's quite hacky and relies on ordering quirks from pyright and mypy. And it would only be to show more information in a specific IDE (Pylance in VSCode).

stubs/SQLAlchemy/sqlalchemy/orm/collections.pyi
```diff
@@ -81,12 +90,38 @@ class CollectionAdapter:
...
+    def __setitem__(self, key: _KT, value: _VT, _sa_initiator: Event | Literal[False] | None = None) -> None: ...
+    @overload
+    def pop(self, key: _KT) -> _VT: ...
+    @overload
+    def pop(self, key: _KT, default: _VT | _T | _symbol | symbol = ...) -> _VT | _T: ...
+    @overload  # type: ignore[override]
+    def setdefault(self, key: _KT, default: _T) -> _VT | _T: ...
+    @overload
+    def setdefault(self, key: _KT, default: None = None) -> _VT | None: ...
+    def update(self, __other: Mapping[_KT, _VT] = ..., **kw: _VT) -> None: ...  # type: ignore[override]
```
AlexWaygood, line 127: should this maybe be overloaded? https://github.com/python/typeshed/blob/7f986bdf859a5c009474356addb7420b0a79844b/stdlib/typing.pyi#L610-L635

stubs/SQLAlchemy/sqlalchemy/orm/strategy_options.pyi
```diff
@@ -23,44 +24,116 @@ class Load(Generative, LoaderOption):
...
+# wich is callable with their original parameters.
+#
+# While both mypy and pyright's validation work, Pylance is unable to
+# show the parameters and return types.
+#
+# There is a workaround (define the method for Pylance, then reassign
+# an instance of loader_option to it for mypy, and add pyright+Flake8
+# suppressions), but it is too hacky and relies on some unsupported quirks.
+#
+# Asking Pylance to add support for these generic callables might be preferable.
```
AlexWaygood, line 60: Don't know about pylance, but pyright is (reasonably) pretty resistant to adding any special-casing for Python libraries outside of the stdlib, so I don't know how likely this is to happen, sadly.

stubs/SQLAlchemy/sqlalchemy/orm/strategy_options.pyi
```diff
@@ -23,44 +24,116 @@ class Load(Generative, LoaderOption):
...
+###
+
+_F = TypeVar("_F", bound=Callable[..., loader_option[Any]])
+
+class loader_option(Generic[_F]):
+    name: str
+    _dynamic: _F
+    fn: _F
+    __call__: _F  # Cheesy "__call__" definition to use the dynamic methods instead
```
AlexWaygood, line 69: Would you be able to add a couple of test cases for this, so we can be sure the type hints are working correctly?

stubs/SQLAlchemy/sqlalchemy/testing/provision.pyi
```diff
@@ -1,33 +1,117 @@
...
-    @classmethod
-    def init(cls, fn): ...
-    def for_db(self, *dbnames): ...
-    def __call__(self, cfg, *arg): ...
+log: Logger
+FOLLOWER_IDENT: Incomplete | None
 
 def create_follower_db(follower_ident) -> None: ...
-def setup_config(db_url, options, file_config, follower_ident): ...
+def setup_config(db_url, options, file_config, follower_ident) -> Config: ...
```
AlexWaygood, line 13: Worth annotating https://github.com/python/typeshed/blob/7f986bdf859a5c009474356addb7420b0a79844b/stubs/SQLAlchemy/sqlalchemy/testing/config.pyi#L29 as well, since we know the return type now?

stubs/SQLAlchemy/sqlalchemy/testing/provision.pyi
```diff
@@ -1,33 +1,117 @@
...
+
+_F = TypeVar("_F", bound=Callable[..., str | URL | None])
+_S = TypeVar("_S", bound=str)
+
+class register(Generic[_F]):
+    fns: dict[str, _F]
+    @classmethod
+    def init(cls: type[Self], fn: _F) -> Self: ...
+    def for_db(self: Self, *dbnames: str) -> Callable[[_F], Self]: ...
+    __call__: _F
```
AlexWaygood, line 42: Same here -- maybe add a test case or two?

stubs/SQLAlchemy/sqlalchemy/util/compat.pyi
```diff
@@ -24,21 +25,32 @@ from urllib.parse import (
...
+is64bit: bool
 has_refcount_gc: bool
-contextmanager = contextlib.contextmanager
 dottedgetter = operator.attrgetter
-namedtuple = collections.namedtuple  # noqa: Y024
-next = builtins.next
+
+class StringIO(_StringIO):
+    def seek(self, __pos: int, __whence: int = 0) -> int: ...
+    def truncate(self, __pos: int | None = None) -> int: ...
```
AlexWaygood, line 53: I don't see these methods being overridden at runtime on the `rel_1_4` branch?

AlexWaygood (commented):
Thank you! Overall this looks great, just a few comments below:

stubs/SQLAlchemy/sqlalchemy/util/compat.pyi
AlexWaygood: Stubtest probably emits errors here for the same reason as it emits errors on the stdlib:

https://github.com/python/typeshed/blob/7f986bdf859a5c009474356addb7420b0a79844b/tests/stubtest_allowlists/py3_common.txt#L111-L112

I'm planning on working on reducing stubtest errors for re-exported objects after mypy 1.0 is released. (There are enough stubtest changes on mypy `master` already; working on it now would make upgrading too disruptive for us imo.)

stubs/SQLAlchemy/@tests/stubtest_allowlist.txt
```diff
@@ -1,114 +1,30 @@
...
-sqlalchemy.orm.strategy_options.Load.load_only
-sqlalchemy.orm.strategy_options.Load.noload
-sqlalchemy.orm.strategy_options.Load.raiseload
-sqlalchemy.orm.strategy_options.Load.selectin_polymorphic
-sqlalchemy.orm.strategy_options.Load.selectinload
-sqlalchemy.orm.strategy_options.Load.subqueryload
-sqlalchemy.orm.strategy_options.Load.undefer
-sqlalchemy.orm.strategy_options.Load.undefer_group
-sqlalchemy.orm.strategy_options.Load.with_expression
+# wrong argument name in implementation ("self" instead of "cls")
```
AlexWaygood, line 12: Wait, aren't these ones also fixable?

stubs/SQLAlchemy/@tests/stubtest_allowlist.txt
```diff
@@ -126,1067 +42,23 @@ sqlalchemy.sql.annotation.Annotated.__new__
 # as having conflicting __new__/__init__ methods
 sqlalchemy.orm.unitofwork.PostSortRec.__new__
 
-# unclear problems
+# KeyError/AttributeError on import due to dynamic initialization from a different module
+sqlalchemy.testing.fixtures
+sqlalchemy.testing.pickleable
+sqlalchemy.testing.plugin.bootstrap
+
+# method arguments starting with double underscores in the implementation trips up stubtest
```
AlexWaygood, line 50: Interesting, this could be a stubtest bug. Maybe file a mypy issue if it is?

JelleZijlstra: Probably causes confusion because in typing this means pos-only args?

AlexWaygood: My suspicion is that stubtest's logic to do with name-mangling might be going wrong somehow. But I don't know, haven't looked into this at all.

hauntsaninja:
(not a review, but those line diff counts look great!!)

stubs/SQLAlchemy/sqlalchemy/orm/strategy_options.pyi
Avasam: pyright is accurate in its errors for parameters type, name, count etc. here. So it looks like something that might've to be handled on Pylance's side to present that information directly to the user.
I'll open a feature request on pylance's side to support this use-case once I can boil it down to a minimal repro. Whether they accept or not this is now accurate and better than it was before (the __call__ signature was misleading anyway given its dynamic nature).

I'll update the inline comment to not mention said workaround, that's more something that belongs as a PR comment like this one.
For posterity, here's the original comment description with some images:

```python
# While both mypy and pyright's validation work, Pylance is unable to
# show the parameters and return types.
#
# There is a workaround (define the method for Pylance, then reassign
# an instance of loader_option to it for mypy, and add pyright+Flake8
# suppressions), but it is too hacky and relies on some unsupported quirks.
#
# Asking Pylance to add support for these generic callables might be preferable.
```

The workaround: https://github.com/python/typeshed/pull/9596#discussion_r1088543062

Pyright spots errors and Pylance picks them up:
image

Pylance does not see parameters and docstring:
image

With workaround:
image

stubs/SQLAlchemy/sqlalchemy/orm/collections.pyi
Avasam: Ah yes, that seems more accurate and matches the source
https://github.com/sqlalchemy/sqlalchemy/blob/rel_1_4_46/lib/sqlalchemy/orm/collections.py#L1403-L1415

stubs/SQLAlchemy/sqlalchemy/util/compat.pyi
Avasam: As far as I can tell it should be using StringIO directly. But I kindof assumed sqlalchemy did some magic I hadn't spotted. And I think I manually checked at runtime? (not sure if I did for this one).

Makes sense if the same issue happens in IO. I'll allowlist them and this can be fixed with an eventual mypy/stubtest update.

Avasam:
> (not a review, but those line diff counts look great!!)

One of the few times we like seeing a lot of red ^^

---

I've yet to add some tests, and we'll see if the last changes affect `mypy_primer`. But that's all the time I had right now. Will tackle those later.

stubs/SQLAlchemy/sqlalchemy/orm/strategy_options.pyi
Avasam: I think the recent changes are much simpler and probably won't require test cases. I'll let you decide.

stubs/SQLAlchemy/@tests/stubtest_allowlist.txt
Avasam: https://github.com/python/mypy/issues/14560

stubs/SQLAlchemy/sqlalchemy/testing/provision.pyi
Avasam: It looks like the tests I can do for `register` are really limited. For example, because it's generic, I get `[assert-type]` errors like:
`Expression is of type "register[Callable[[object, object, object], None]]", not
"register[Any]"`, not "register[Any]"` or `Expression is of type "register[Callable[[object, object, object], None]]", not
"register[Callable[[object, object, object], None]]"`

At least I can test that the first parameter is correctly replaced.

https://github.com/python/typeshed/actions/runs/4045446150/jobs/6956947952

Avasam:
#9622 concerning "Check file consistency"

AlexWaygood:
(I'm on vacation this week and trying to have something of an OSS break while I'm away — will take a look when I get back!)

JelleZijlstra:
@AlexWaygood do you want to take another look?

stubs/SQLAlchemy/@tests/test_cases/check_register.py
```diff
@@ -0,0 +1,67 @@
...
+    pass
+
+
+# The decorator changes the first parameter to "cfg: str | URL | _ConfigProtocol"
+@register.init
+def no_args(__foo: Foo) -> None:
+    pass
+
+
+no_args(cfg="")
```
JelleZijlstra, line 47: Is it useful to assert_type the return type of these and the next functions?

Avasam:
(I've been quite occupied last week, and I got sick this week, will go through updated reviews soon once I feel a bit better)

stubs/SQLAlchemy/sqlalchemy/orm/strategy_options.pyi
Avasam: (tests added in a previous commit)

stubs/SQLAlchemy/sqlalchemy/testing/provision.pyi
Avasam: (tests added in a previous commit)

Avasam:
@AlexWaygood @JelleZijlstra Can I get an updated review on this? I answered the comments and I don't think there was anything new to add (just merged main to resolve conflicts)

stubs/SQLAlchemy/sqlalchemy/util/langhelpers.pyi
```diff
@@ -133,12 +133,16 @@ class hybridmethod:
     def classlevel(self, func): ...
 
 class _symbol(int):
-    def __new__(cls, name, doc: Incomplete | None = None, canonical: Incomplete | None = None): ...
+    def __new__(  # noqa: Y034  # Excplicitely instanciates _symbol
```
JelleZijlstra, line 136:
```suggestion
    def __new__(  # noqa: Y034  # Excplicitly instantiates _symbol
```

stubs/SQLAlchemy/sqlalchemy/util/langhelpers.pyi
```diff
@@ -133,12 +133,16 @@ class hybridmethod:
...
+        self, name, doc: Incomplete | None = None, canonical: Incomplete | None = None
+    ) -> _symbol: ...
     def __reduce__(self): ...
 
 class symbol:
     symbols: Any
-    def __new__(cls, name, doc: Incomplete | None = None, canonical: Incomplete | None = None): ...
+    def __new__(  # type: ignore[misc]  # Excplicitely instanciates _symbol
```
JelleZijlstra, line 143:
```suggestion
    def __new__(  # type: ignore[misc]  # Explicitly instantiates _symbol
```

JelleZijlstra (approved):

JelleZijlstra:
@AlexWaygood let's merge this?

stubs/SQLAlchemy/sqlalchemy/util/langhelpers.pyi
```diff
@@ -133,12 +133,16 @@ class hybridmethod:
     def classlevel(self, func): ...
 
 class _symbol(int):
-    def __new__(cls, name, doc: Incomplete | None = None, canonical: Incomplete | None = None): ...
+    def __new__(  # noqa: Y034  # Excplicitly instantiates _symbol
```
AlexWaygood, line 136:
```suggestion
    def __new__(  # noqa: Y034  # Explicitly instantiates _symbol
```

AlexWaygood:
> @AlexWaygood let's merge this?

I remember having some qualms a while back about a few minor things,  but I can't remember what they are anymore, and I don't think they really matter in the scheme of things :)

Let's go for it.

AlexWaygood:
@Avasam sorry for the delay on this one 😔

Avasam:
> I remember having some qualms a while back about a few minor things that I wanted to investigate further, but I can't remember what they are anymore, and I don't think they really matter in the scheme of things :)

There's a follow-up PR anyway (which may or may not have conflicts now, I haven't checked). If you remember anything you can mention it there too.

Avasam:
Just leaving a comment here to link to https://github.com/python/typeshed/issues/9127#issuecomment-1528827341

PR #9568, opened 2023-01-18
[SQLAlchemy] Annotate row classes

srittau:

PR #9597, opened 2023-01-28
Combine microsoft/python-type-stubs/sqlalchemy

Avasam:
~~Includes #9596 and #9568, so merge those first. (keeping this as draft until then, I've also yet to self-review my own PR with a fresh mind)~~
If you'd like me to split up this PR (maybe you'd like to review a specific section in isolation) and you can think of a decent way to do so, let me know.
In the same vein, I'd rather not add more type-hints, even if they seem obvious (unless it's to fix an issue). I would consider it scope-creeping given the motivation behind this PR: https://github.com/python/typeshed/pull/9595#issuecomment-1405962252 

Brings `types-sqlalchemy` on par with https://github.com/microsoft/python-type-stubs/tree/main/sqlalchemy and https://github.com/dropbox/sqlalchemy-stubs. Except for making most classes (mainly `Column` and `Row` related ones) generic. As that would cause hundreds, if not thousands of additional (and unhelpful) `mypy_primer` differences. It also helps keep this PR more manageable and focus on the non-generic type issues.

Merged stubs, avoided new generics, manually reviewed possible types and fixed existing type issues I found.
Standardized imports (relative seems to be preferred for this stub) and fixed a handful of pytype references issues with imports. (https://github.com/google/pytype/issues/1087)
I tried my best to fully comment the confusing parts.


stubs/SQLAlchemy/sqlalchemy/sql/schema.pyi
```diff
@@ -1,105 +1,205 @@
...
+    @memoized_property
+    def info(self) -> Mapping[str, Any]: ...
+    # FIXME: Only specifying TraversibleType in __init__ to work around a mypy bug!
+    # example from pandas.tests.io.tests_sql:
+    # ```python
+    # date_type = TEXT if dialect == "sqlite" else DateTime
+    # bool_type = Integer if dialect == "sqlite" else Boolean
+    # Column("DateCol", date_type)
+    # ```
+    # mypy thinks date_type is TraversibleType instead of Union[type[TEXT], type[DateTime]]
```
Avasam, line 89: @hauntsaninja This looks like a mypy bug (pyright has no issues with that snippet of code). Are you aware of this? (I haven't tried in 1.0 either) All those types should be valid `type[TypeEngine]`, but instead mypy narrows it down to their metaclass `TraversibleType`.

mypy_primer comment before I added this workaround: https://github.com/Avasam/typeshed/pull/24#issuecomment-1407323271
https://github.com/Avasam/typeshed/actions/runs/4030841677/jobs/6929764991

Avasam:
@hauntsaninja I have a mypy_primer question for you. What's up with these diff lines -->
image

And does it mean anything to us or is it related to: https://github.com/hauntsaninja/mypy_primer/blob/master/mypy_primer.py#L521

Avasam:
Done with the big merge!

As usual, a bunch of nonsensical pytype errors that I'll have to try and figure out.

I'll go through the updated mypy_primer entried as well.

Avasam:
I'm not getting this pytype error locally with WSL.

PR #9598, opened 2023-01-28
Increase `mypy_primer` comment length

Avasam:
To be exact, we only need to reserve 170 characters to truncate over a thousand lines. But I figured that `65300` was a nice round number with a good buffer.

Sources:
image
and https://github.com/orgs/community/discussions/27190#discussioncomment-3254953

### DIFF
--- a/.github/workflows/mypy_primer_comment.yml
+++ b/.github/workflows/mypy_primer_comment.yml
@@ -50,9 +50,11 @@ jobs:
           script: |
             const fs = require('fs')
             let data = fs.readFileSync('fulldiff.txt', { encoding: 'utf8' })
-            // posting comment fails if too long, so truncate
-            if (data.length > 30000) {
-              let truncated_data = data.substring(0, 30000)
+
+            // Maximum comment length is 65536 characters. We need much less than 236 for extra text.
+            const MAX_LENGTH = 65300
+            if (data.length > MAX_LENGTH) {
+              let truncated_data = data.substring(0, MAX_LENGTH)
               let lines_truncated = data.split('\n').length - truncated_data.split('\n').length
               data = truncated_data + `\n\n... (truncated ${lines_truncated} lines) ...\n`
             }


AlexWaygood (approved):

JelleZijlstra (approved):
Might be worth doing the same on mypy.

JelleZijlstra:
Also, over at Black we have a similar system called diff-shades, and there we get around the comment length limit by printing the full diff to the action's log instead and linking to it. It's a little less convenient, but we could consider it as an option for very long diffs if we think it's important to be able to see the full comment. I'd personally consider it overkill since a very long mypy-primer diff is usually enough indication that a change is a bad idea.

AlexWaygood:
> Also, over at Black we have a similar system called diff-shades, and there we get around the comment length limit by printing the full diff to the action's log instead and linking to it. It's a little less convenient, but we could consider it as an option for very long diffs if we think it's important to be able to see the full comment. I'd personally consider it overkill since a very long mypy-primer diff is usually enough indication that a change is a bad idea.

Interesting, that sounds like overkill for typeshed, but might be pretty useful for mypy. It would have been very useful to easily see the full diffs for the `typing.Self` PRs, where the diff was often huge because of the number of projects that were already using `typing.Self`, despite it being unsupported by mypy. And there's value in keeping the workflow relatively in sync across the two projects; it's less cognitive load for maintainers.

Avasam:
Note that the action prints the full diff! This is purely for convenience with the comment since I hit the old limit recently with a legit PR.
Splitting the comment might be overkill for stubtest and flake8-pyi's primers. But if it can link to the action run when truncating that could be a nice qol improvement.

hauntsaninja:
The full diffs are available for download if you click through.

To Alex's suggestion for mypy: we now have per project truncation on mypy. 

Avasam, do you have a PR in mind where 30k chars more of comment would have been useful?

AlexWaygood:
> Avasam, do you have a PR in mind where 30k chars more of comment would have been useful?

I think probably #9597

Avasam:
> > Avasam, do you have a PR in mind where 30k chars more of comment would have been useful?
> 
> I think probably #9597

I had up to 230+ truncated lines at some point because of generics-related changes (which would result in 35-40k+ additional chars), but decided to revert that and keep for later.

The idea is more that if we limit the length for technical reasons, may as well get closer to the limit, not halfway there (and it doesn't negatively affect readability). In other words, making the limit *less* arbitrary.

Truncating in multiple comments is something that can be done if someone wants to tackle it, but this hard-limit would still be relevant, and is likely still gonna be hit from the homeassistant project alone once I tackle SQLAlchemy generics.

hauntsaninja (approved):

PR #9600, opened 2023-01-28
Improve stubsabot logic for finding the first release with a `py.typed` file

AlexWaygood:
Addresses some of the confusion seen in #9595

### DIFF
--- a/scripts/stubsabot.py
+++ b/scripts/stubsabot.py
@@ -174,7 +174,22 @@ async def find_first_release_with_py_typed(pypi_info: PypiInfo, *, session: aioh
     release_iter = pypi_info.releases_in_descending_order()
     while await release_contains_py_typed(release := next(release_iter), session=session):
         first_release_with_py_typed = release
-    return first_release_with_py_typed
+    first_version_with_py_typed = first_release_with_py_typed.version
+    if not first_version_with_py_typed.is_prerelease:
+        return first_release_with_py_typed
+    normalised_versions_to_version_strings = {
+        packaging.version.Version(version_string): version_string for version_string in pypi_info.releases
+    }
+    # We should be able to safely count on this being a non-empty iterable,
+    # since this function hopefully won't be called
+    # unless a non-prerelease version has been released
+    # that has a py.typed file
+    first_nonprerelease_version_with_py_typed = min(
+        version
+        for version in normalised_versions_to_version_strings
+        if version > first_version_with_py_typed and not version.is_prerelease
+    )
+    return pypi_info.get_release(version=normalised_versions_to_version_strings[first_nonprerelease_version_with_py_typed])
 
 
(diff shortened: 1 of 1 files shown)


scripts/stubsabot.py
```diff
@@ -174,7 +174,22 @@ async def find_first_release_with_py_typed(pypi_info: PypiInfo, *, session: aioh
...
     while await release_contains_py_typed(release := next(release_iter), session=session):
         first_release_with_py_typed = release
-    return first_release_with_py_typed
+    first_version_with_py_typed = first_release_with_py_typed.version
+    if not first_version_with_py_typed.is_prerelease:
+        return first_release_with_py_typed
+    normalised_versions_to_version_strings = {
+        packaging.version.Version(version_string): version_string for version_string in pypi_info.releases
+    }
+    # We should be able to safely count on this being a non-empty iterable,
```
JelleZijlstra, line 183: I don't understand why so much new code is needed, and the comment gives me pause. What if instead we put `if release.version.is_prerelease: continue` in the while loop above?

AlexWaygood: That won't work because in the `while` loop we're iterating over the releases in reverse order, so if we `continue`, that means that we've already missed the first non-prerelease with a `py.typed` file.

AlexWaygood: We can simplify by iterating over the releases forwards instead of backwards, but that will entail making many more network requests than is strictly necessary. But it would be simpler, and maybe we don't care about spamming PyPI with loads of network requests

AlexWaygood: actually I think you're right and I'm not talking any sense

JelleZijlstra: We probably do need some additional changes because this function seems to assume there is at least one py.typed release, which will no longer be the case. We may have to ignore prereleases somewhere else too.

AlexWaygood: I don't _think_ there are any changes needed elsewhere:
- `SQLAlchemy` has had lots of prereleases that included `py.typed` files, but stubsabot only tried to mark our stubs as obsolete when they released a non-prerelease that had a `py.typed` file. It just got the body and title of the PR incorrect.
- Everything seems to work fine when I run the script locally.
- This function is only called if the "latest" release according to PyPI has a `py.typed` file in it:
  https://github.com/python/typeshed/blob/0024dc4f335feca71cfc9f1027a7c1ebd3f453fb/scripts/stubsabot.py#L420-L425

  I think, if PyPI is working correctly, it won't claim that a prerelease is the "latest" release, which is _why_ we didn't get stubsabot trying to mark our stubs for SQLAlchemy as obsolete until a non-prelease with a `py.typed` file is released.

JelleZijlstra: Ah, I missed the fact that SQLAlchemy 2.0 has actually been released. That makes sense now.

JelleZijlstra (approved):

PR #9601, opened 2023-01-28
Add some missing Self return annotations in redis

wookie184:
Add some missing Self return annotations in redis

Source for `Monitor.__aenter__`:
https://github.com/redis/redis-py/blob/e39c7babdd80ecdd930bed9d201e8fb9187309ac/redis/asyncio/client.py#L571

Source for `PubSub.__aenter__`:
https://github.com/redis/redis-py/blob/e39c7babdd80ecdd930bed9d201e8fb9187309ac/redis/asyncio/client.py#L671

The non-async equivalents already seem to be correct (e.g. https://github.com/python/typeshed/blob/b316dd947e96d0e0d17626fe11f0e5bfa2aa3782/stubs/redis/redis/client.pyi#L343)


stubs/redis/redis/asyncio/client.pyi
```diff
@@ -116,7 +116,7 @@ class Monitor:
     connection: Any
     def __init__(self, connection_pool: ConnectionPool) -> None: ...
     async def connect(self) -> None: ...
-    async def __aenter__(self): ...
+    async def __aenter__(self) -> Self: ...
```
AlexWaygood, line 119:
```suggestion
    async def __aenter__(self: Self) -> Self: ...
```

stubs/redis/redis/asyncio/client.pyi
```diff
@@ -142,7 +142,7 @@ class PubSub:
         ignore_subscribe_messages: bool = ...,
         encoder: Incomplete | None = ...,
     ) -> None: ...
-    async def __aenter__(self): ...
+    async def __aenter__(self) -> Self: ...
```
AlexWaygood, line 145:
```suggestion
    async def __aenter__(self: Self) -> Self: ...
```

AlexWaygood (commented):
Thanks! We're not yet able to use `typing(_extensions).Self` due to lack of mypy support, so we currently use `_typeshed.Self` instead. `_typeshed.Self is just a normal `TypeVar`, so you have to annotate the parameter as well as the return type:

AlexWaygood:
(Mypy already has support for `typing(_extensions).Self` on the master branch; we're just waiting for version 1.0 to be released.)

AlexWaygood (approved):
Thank you!

Issue #9608, opened 2023-01-29
Several errors in overloaded functions in the stdlib


AlexWaygood:
Attempting to add defaults to overloaded functions in #9606 highlighted many instances where we have incorrect overloads in the stdlib. An example is this overload in `_bisect.pyi`:

https://github.com/python/typeshed/blob/808a431a57e7efb2ec8afa97840c368e760d2e11/stdlib/_bisect.pyi#L13-L21

This overload incorrectly marks the `key` parameter as having a default value. But this has several issues:
- If no value is provided for this parameter by the user, we actually want the type checker to pick the _first_ overload (not this overload, which is the second overload). Specifying a default value for this parameter in this overload is confusing.
- If the type checker _did_ select this overload, the `_T` TypeVar would go unsolved if no value was specified for the `key` parameter.
- If we try to add the default value for this parameter (`None`), mypy will (correctly) complain that we're using an implicit optional type for the type annotation.

There are several issues like this in our stdlib stubs that should be fixed (see https://github.com/python/typeshed/pull/9606/commits/b39305ee691e2789e5e6ba4e525b2e619f0d33b4 for a summary of the problematic functions — this is the commit in #9606 in which I reverted all the defaults `stubdefaulter` had added that caused mypy to emit errors).

AlexWaygood:
There are several functions in `asyncio.events` and `asyncio.base_events` where `stubdefaulter` added default values that I had to revert (see https://github.com/python/typeshed/commit/b39305ee691e2789e5e6ba4e525b2e619f0d33b4). I can see that the overloads are incorrect, but I'm not sure what the correct fix is. I'd appreciate it if somebody who knows their `asyncio` better than I do could take a look.

srittau:
I'm not sure anyone understands asyncio ...

AlexWaygood:
Other than the asyncio ones, all the fixable ones that I know about have now been fixed (though we'll probably discover a few more once https://github.com/JelleZijlstra/stubdefaulter/issues/35 is implemented...).

I'll open a followup issue for the asyncio ones. Many thanks, everybody, for the speedy reviews on my PRs!

PR #6941, opened 2022-01-17
Improve type for setdefault()

JelleZijlstra:
- With one argument, it may return None
- With two arguments, it returns the default's type or the dict's value type.
- Also remove incorrect `= ...` from `pop()`. The one-argument case has its own overload.

Context: https://github.com/python/typing/discussions/1033#discussioncomment-1986359

PR #9612, opened 2023-01-29
Improve several stdlib `setdefault` methods

AlexWaygood:
1. Improve the first overload of `MutableMapping.setdefault`. It's okay to specify a value for the `default` parameter and still have this overload match, as long as the value passed for this parameter is `None`. Adjust `collections.OrderedDict.setdefault` and `weakref.WeakKeyDictionary.setdefault` to match. Add @JelleZijlstra's test cases from https://github.com/python/typeshed/pull/6941#issuecomment-1016074301, to make sure that the method is still working as expected.
2. Bring `ChainMap.setdefault` in sync with `MutableMapping.setdefault`. At runtime, it is directly inherited from `MutableMapping.setdefault`. (We don't do that in the stub because we pretend `MutableMapping.setdefault` has positional-only parameters so that `dict` can inherit from `MutableMapping` without having to `type: ignore` the `setdefault` method.
3. Remove the `= ...` from `weakref.WeakValueDictionary.setdefault`. This _has_ got a default value of `None` at runtime, but it will always error out if you don't specify a value for the `default` parameter. (This is because you can't create a weak reference to `None`.) It will serve users better if we specify in the stub that you have to provide a value for the `default` parameter.

Refs #9608.


AlexWaygood:
This is not a comprehensive review of all `setdefault` methods across the stdlib, but I haven't got the energy for that right now.

test_cases/stdlib/typing/check_MutableMapping.py
```diff
@@ -0,0 +1,16 @@
+from typing import Any, Union
```
AlexWaygood, line 1:
```suggestion
from __future__ import annotations

from typing import Any, Union
```

JelleZijlstra (approved):

PR #9620, opened 2023-01-30
`email.message`: Fix invalid `TypeVar` usage; add some default values

AlexWaygood:
Refs #9608. For all of the functions touched here, the `_T` TypeVar would go unsolved if no value was passed by the user for the `failobj` parameter. We have to use overloads here to ensure that `_T` will never silently be resolved as `Any`


AlexWaygood:
Links to the primer hits:

https://github.com/pypa/pip/blob/e3e7bc34eb486622ebbb6412afc98ee57fcbff4a/src/pip/_internal/metadata/_json.py#L67-L70

https://github.com/aio-libs/aiohttp/blob/4635161ee8e7ad321cca46e01ce5bfeb1ad8bf26/aiohttp/helpers.py#L748-L749

These are both true positives: these are cases where `_T` was previously being silently resolved to `Any` by mypy, meaning that it didn't notice the typing errors in the user code.

srittau (approved):

srittau:
This would a good case for type var defaults (python/typing#307) as well.

PR #9622, opened 2023-01-30
Remove redundant `reportUnnecessaryTypeIgnoreComment=true`

Avasam:
Seems pretty redundant since #9397 `reportUnnecessaryTypeIgnoreComment` is set to `error` in all configs

### DIFF
--- a/stubs/invoke/@tests/test_cases/check_task.py
+++ b/stubs/invoke/@tests/test_cases/check_task.py
@@ -1,4 +1,3 @@
-# pyright: reportUnnecessaryTypeIgnoreComment=true
 from __future__ import annotations
 
 from invoke import Context, task
--- a/stubs/protobuf/@tests/test_cases/check_struct.py
+++ b/stubs/protobuf/@tests/test_cases/check_struct.py
@@ -1,4 +1,3 @@
-# pyright: reportUnnecessaryTypeIgnoreComment=true
 from __future__ import annotations
 
 from google.protobuf.struct_pb2 import ListValue, Struct
--- a/stubs/regex/@tests/test_cases/check_finditer.py
+++ b/stubs/regex/@tests/test_cases/check_finditer.py
@@ -1,5 +1,3 @@
-# pyright: reportUnnecessaryTypeIgnoreComment=true
-
 from __future__ import annotations
 
 from typing import List
--- a/stubs/requests/@tests/test_cases/check_post.py
+++ b/stubs/requests/@tests/test_cases/check_post.py
@@ -1,5 +1,3 @@
-# pyright: reportUnnecessaryTypeIgnoreComment=true
-
 from __future__ import annotations
 
 from collections.abc import Iterable
--- a/tests/check_consistent.py
+++ b/tests/check_consistent.py
@@ -14,8 +14,8 @@ from pathlib import Path
 import yaml
 from packaging.requirements import Requirement
 from packaging.specifiers import SpecifierSet
-from parse_metadata import read_metadata
 
+from parse_metadata import read_metadata
 from utils import VERSIONS_RE, get_all_testcase_directories, get_gitignore_spec, spec_matches_path, strip_comments
 
 extension_descriptions = {".pyi": "stub", ".py": ".py"}
--- a/tests/check_consistent.py
+++ b/tests/check_consistent.py
@@ -75,14 +75,6 @@ def check_test_cases() -> None:
             with open(file, encoding="UTF-8") as f:
                 lines = {line.strip() for line in f}
             assert "from __future__ import annotations" in lines, "Test-case files should use modern typing syntax where possible"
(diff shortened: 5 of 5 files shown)


srittau:
Are the pyright configs used when checking scripts? As far as I understand, our scripts are currently checked with mypy only, at least in CI. That said, I'd be glad if we could apply this PR.

srittau:
Cc @AlexWaygood

AlexWaygood:
The pyright configs are not used when checking our scripts, but they are used when checking our test-case files. The stdlib test cases have a dedicated pyrightconfig file all to themselves, but the third-party test cases use the same config files as our third-party stubs (they have to, because the third-party test cases are in the same top-level folder as the third-party stubs).

The reason why I initially introduced this check into check_consistent.py was because we didn't at the time have `reportUnnecessaryTypeIgnoreComment=true` enabled for the `stubs/` directory. But it's crucial to have this setting enabled for our test cases, or pyright won't verify our test cases that check for false negatives.

Since, as @Avasam says, `reportUnnecessaryTypeIgnoreComment=true`  is nowadays enabled for the whole of the stubs directory, we can remove this check from check_consistent.py. But this makes me realise that our pyright settings for our test cases are probably now broken in a new way. This line in our pyright config files means that pyright now straight-up ignores `type: ignore` comments altogether in the `stubs/` directory:

https://github.com/python/typeshed/blob/b5a26d1ca258a86a6f133dfabf58790553dc21cf/pyrightconfig.json#L36

Presumably pyright won't emit any errors for unused `type: ignore` comments if it's just ignoring all of them anyway. So maybe we should replace the current check that enforces `reportUnnecessaryTypeIgnoreComment=true` to be added at the top of each third-party test-case file with a new check that enforces `enableTypeIgnoreComments=true` to be added at the top of each test-case file.

Avasam:
> Are the pyright configs used when checking scripts? As far as I understand, our scripts are currently checked with mypy only, at least in CI. That said, I'd be glad if we could apply this PR.

Pyright is run on the `test_cases` folders when run with the `pyrightconfig.testcases.json` configuration.

---

@AlexWaygood `"enableTypeIgnoreComments": true` specifically for test_cases:
https://github.com/python/typeshed/blob/main/pyrightconfig.testcases.json#L8-L10

`python .\tests\pyright_test.py .\stubs\SQLAlchemy\@tests -p .\pyrightconfig.testcases.json`
```
8.19.2
Running: C:\Program Files\nodejs\npx.CMD pyright@1.1.285 .\stubs\SQLAlchemy\@tests -p .\pyrightconfig.testcases.json
Loading configuration file at E:\Users\Avasam\Documents\Git\typeshed\pyrightconfig.testcases.json
Assuming Python version 3.9
Assuming Python platform Windows
Auto-excluding **/node_modules
Auto-excluding **/__pycache__
Auto-excluding **/.*
stubPath E:\Users\Avasam\Documents\Git\typeshed\typings is not a valid directory.
Searching for source files
Found 2 source files
pyright 1.1.285
E:\Users\Avasam\Documents\Git\typeshed\stubs\SQLAlchemy\@tests\test_cases\check_register.py
  E:\Users\Avasam\Documents\Git\typeshed\stubs\SQLAlchemy\@tests\test_cases\check_register.py:45:21 - error: Unnecessary "# type: ignore" comment
```
image

AlexWaygood:
My suspicions are confirmed — pyright should be emitting an error on this PR about the unused `type: ignore` comment, but isn't:

- #9623

> python .\tests\pyright_test.py .\stubs\SQLAlchemy\@tests -p .\pyrightconfig.testcases.json

Remember that we don't use this script in CI; we use pyright-action, which works differently

> @AlexWaygood `"enableTypeIgnoreComments": true` specifically for test_cases. So it doesn't ignore it, with a comment to warn anyone who'd want to deactivate it:

As I explained above, the `pyrightconfig.testcases.json` file is only for the stdlib test cases. It isn't used for the third-party test cases, which was the whole reason why I added this check to `check_consistent.py` in the first place :)

AlexWaygood:
Okay, so we can see in #9623 that pyright isn't verifying our third-party test cases guarding against false negatives in our CI. I also can't work out how to fix that. Adding `# pyright: enableTypeIgnoreComments=true` to the top of the file didn't make pyright pick up on the unused `type: ignore` comment. Neither have my various attempts to use glob patterns in `pyrightconfig.testcases.json`, to try to get pyright to use _that_ config file for testing our third-party test cases. (The [pyright docs](https://github.com/microsoft/pyright/blob/main/docs/configuration.md#main-pyright-config-options) say it should be possible; I don't know why it isn't working here.)

The good news is that pyright _is_ verifying our _stdlib_ test cases guarding against false negatives: see #9624.

AlexWaygood:
Thank you for raising this issue! I'm closing this in favour of #9650, but I've listed you as a co-author over there :)

PR #9624, opened 2023-01-30
[DO-NOT-MERGE] Checking to see if pyright verifies stdlib test cases against false negatives correctly

AlexWaygood:
If our test-cases setup is working correctly, pyright should hopefully complain about this `type: ignore` being unused.


AlexWaygood:
Well _that's_ a relief

PR #9625, opened 2023-01-30
Add a default value to `typing.Sequence.index`

AlexWaygood:
I'm guessing this will cause a huge mypy_primer diff, but let's see


JelleZijlstra:
I believe we don't have the `| None` because `list` and `tuple` don't accept it. So while the Sequence ABC technically requires None to be accepted, the most common practical Sequences don't.

AlexWaygood:
> I believe we don't have the `| None` because `list` and `tuple` don't accept it. So while the Sequence ABC technically requires None to be accepted, the most common practical Sequences don't.

That makes sense. Sounds like a good candidate to be added to the proposed blacklist in https://github.com/JelleZijlstra/stubdefaulter/issues/36

AlexWaygood:

Interesting, it looks like this would actually only impact typeshed. But I guess reflecting the "true" runtime signature here would probably just lead to an annoying number of `type: ignore`s across typeshed, and have minimal benefits to users. So I guess practicality beats purity here — probably still not worth it.

PR #9626, opened 2023-01-30
`TensorFlow`: add missing v2 imports like 'compat'

bschnurr:


stubs/tensorflow/tensorflow/__init__.pyi
```diff
@@ -8,6 +8,53 @@ from typing import Any, NoReturn, overload
 from typing_extensions import TypeAlias
 
 import numpy
+from tensorflow._api.v2 import (
```
hmc-cs-mdrissi, line 11: These stubs avoid referencing undocumented/private modules like _api. They instead mimic public documentation/expected imports. So I'd recommend doing,

```python
from tensorflow import (
audio as audio,
autodiff as autodiff,
...
)
```

without _api.v2

Other thing is you can't import things like audio without first defining stub folder/files for them. The stubs don't know about actual library modules. Minimal file would be doing,

audio/__init__.pyi

and having getattr defined in that file.

bschnurr: My main goes was to expose 'tf.compat'  seen here
https://www.tensorflow.org/api_docs/python/tf/compat

and in this tensornet code base
https://github.com/taehoonlee/tensornets/blob/c9b1d78f806892193efdebee2789a47fd148b984/tensornets/utils.py#L25

I'll se if your form finds 'tf.compat'

if not i'll try your directory suggestion.

PR #9629, opened 2023-01-30
`urllib.parse`: Fix issues with TypeVar usage and default values

AlexWaygood:
The TypeVar usage in `urlencode` is invalid: the `_Q` TypeVar will go unsolved if no value is passed in for the `safe` parameter. To fix this, it must become an overloaded function.

For `urlparse` and `urlsplit`, the `= ...`s are unnecessary and incorrect for the second overload; the first overload takes care of the cases where a user doesn't specify a value for the `scheme` parameter. However, it seems like there are a lot of cases of people passing in `str | None` for the first parameter in the wild, which makes simply removing the `= ...`s here quite disruptive. Add a third overload to make the change less disruptive.

Refs #9608


stdlib/urllib/parse.pyi
```diff
@@ -1,6 +1,7 @@
 import sys
 from collections.abc import Callable, Iterable, Mapping, Sequence
 from typing import Any, AnyStr, Generic, NamedTuple, TypeVar, overload
+from typing_extensionsn import TypeAlias
```
AlexWaygood, line 4:
```suggestion
from typing_extensions import TypeAlias
```

stdlib/urllib/parse.pyi
```diff
@@ -169,20 +193,32 @@ def urljoin(base: AnyStr, url: AnyStr | None, allow_fragments: bool = True) -> A
 def urlparse(url: str, scheme: str | None = "", allow_fragments: bool = True) -> ParseResult: ...
 @overload
 def urlparse(
-    url: bytes | bytearray | None, scheme: bytes | bytearray | None = ..., allow_fragments: bool = True
+    url: bytes | bytearray | None, scheme: bytes | bytearray | None, allow_fragments: bool = True
 ) -> ParseResultBytes: ...
 @overload
+def urlparse(
```
JelleZijlstra, line 199: What does this overload cover? We should reject mixed calls, since those throw an error. We should also remove `| None` from the first overload; passing None always fails.

stdlib/urllib/parse.pyi
```diff
@@ -169,20 +193,32 @@ def urljoin(base: AnyStr, url: AnyStr | None, allow_fragments: bool = True) -> A
...
+    url: str | bytes | bytearray | None, scheme: str | bytes | bytearray | None = "", allow_fragments: bool = True
+) -> ParseResult | ParseResultBytes: ...
+@overload
 def urlsplit(url: str, scheme: str | None = "", allow_fragments: bool = True) -> SplitResult: ...
```
JelleZijlstra, line 203:
```suggestion
def urlsplit(url: str, scheme: str = "", allow_fragments: bool = True) -> SplitResult: ...
```

```
>>> urllib.parse.urlsplit("http://quora.com", None)
  File "/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/urllib/parse.py", line 459, in urlsplit
    scheme = scheme.replace(b, "")
             ^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'replace'
```

stdlib/urllib/parse.pyi
```diff
@@ -169,20 +193,32 @@ def urljoin(base: AnyStr, url: AnyStr | None, allow_fragments: bool = True) -> A
...
 
 if sys.version_info >= (3, 11):
     @overload
-    def urlsplit(url: bytes | None, scheme: bytes | None = ..., allow_fragments: bool = True) -> SplitResultBytes: ...
+    def urlsplit(url: bytes | None, scheme: bytes | None, allow_fragments: bool = True) -> SplitResultBytes: ...
+    @overload
```
JelleZijlstra, line 208: Again, not sure what this overload is for.

stdlib/urllib/parse.pyi
AlexWaygood: The original version of this PR didn't have this overload, but I was pretty scared by the size of the mypy_primer diff: https://github.com/python/typeshed/pull/9629#issuecomment-1409519042.

So I figured I'd try adding a fallback overload of some kind. But I agree it doesn't make much sense.

JelleZijlstra: Hm, the first one I checked there was doing essentially `urlparse(str | None)`. And indeed that works:

```
>>> urllib.parse.urlparse(None)
ParseResultBytes(scheme=b'', netloc=b'', path=b'', params=b'', query=b'', fragment=b'')
```

Maybe we need an overload where the first arg is specifically None?

stdlib/urllib/parse.pyi
```diff
@@ -169,21 +193,27 @@ def urljoin(base: AnyStr, url: AnyStr | None, allow_fragments: bool = True) -> A
 def urlparse(url: str, scheme: str | None = "", allow_fragments: bool = True) -> ParseResult: ...
 @overload
 def urlparse(
-    url: bytes | bytearray | None, scheme: bytes | bytearray | None = ..., allow_fragments: bool = True
+    url: bytes | bytearray | None, scheme: bytes | bytearray | None, allow_fragments: bool = True
 ) -> ParseResultBytes: ...
 @overload
-def urlsplit(url: str, scheme: str | None = "", allow_fragments: bool = True) -> SplitResult: ...
+def urlparse(url: str | None, scheme: Literal[""] = "", allow_fragments: bool = True) -> ParseResult | ParseResultBytes: ...
```
JelleZijlstra, line 199:
```suggestion
def urlparse(url: None, scheme: Literal[""] = "", allow_fragments: bool = True) -> ParseResultBytes: ...
```
Anything involving str is covered by the first overload, right?

stdlib/urllib/parse.pyi
```diff
@@ -169,21 +193,27 @@ def urljoin(base: AnyStr, url: AnyStr | None, allow_fragments: bool = True) -> A
 def urlparse(url: str, scheme: str | None = "", allow_fragments: bool = True) -> ParseResult: ...
```
JelleZijlstra, line 193:
```suggestion
def urlparse(url: str, scheme: str = "", allow_fragments: bool = True) -> ParseResult: ...
```
None doesn't work here.

stdlib/urllib/parse.pyi
AlexWaygood: No, the first overload will only be selected if the value passed in is inferred to be of type `str` (or a subclass of `str`). If the value passed in is inferred to be of type `str | None`, the type checker will reject the first overload.

Similarly, the second overload covers passing in `None`, but if you have a value that's inferred to be of type `str | None`, the type checker will reject that overload when you pass in that value.

JelleZijlstra: Type checkers should be able to split unions. https://mypy-play.net/?mypy=latest&python=3.10&gist=3d5319a9d1060a7d95b1927bb8c67e9d

AlexWaygood: Okay -- let's wait for mypy_primer to give its verdict on the current version, then I'll try this version :)

AlexWaygood: You were right, I was wrong!

JelleZijlstra (approved):

PR #9630, opened 2023-01-31
Various `asyncio` classes: remove the `loop` argument on 3.10

AlexWaygood:
For all of these classes, the `loop` argument exists in their `__init__` methods on 3.10, but `TypeError` is raised if you pass any value in for this argument. It will serve users better if we simply omit the argument from the stub.

On 3.11+, the argument no longer exists at runtime. For <=3.9, you can pass values into this parameter without `TypeError` being raised, and we can add `None` as a default for the `loop` argument.


tests/stubtest_allowlists/py3_common.txt
```diff
@@ -220,6 +220,24 @@ webbrowser.UnixBrowser.remote_action_newtab  # always overridden in inheriting c
 webbrowser.UnixBrowser.remote_action_newwin  # always overridden in inheriting class
 xml.parsers.expat.expat_CAPI
 
+# ==========
+# The "lock" argument exists at runtime,
```
AlexWaygood, line 224:
```suggestion
# The "loop" argument exists at runtime,
```

JelleZijlstra (approved):

PR #9631, opened 2023-01-31
Add several version-dependent default values to parameters in the stdlib

AlexWaygood:
Luckily there aren't too many of these


stdlib/builtins.pyi
```diff
@@ -849,8 +849,10 @@ class memoryview(Sequence[int]):
     def __setitem__(self, __s: slice, __o: ReadableBuffer) -> None: ...
     @overload
     def __setitem__(self, __i: SupportsIndex, __o: SupportsIndex) -> None: ...
-    if sys.version_info >= (3, 8):
-        def tobytes(self, order: Literal["C", "F", "A"] | None = ...) -> bytes: ...
+    if sys.version_info >= (3, 10):
+        def tobytes(self, order: Literal["C", "F", "A"] | None = "C") -> bytes: ...
+    elif sys.version_info >= (3, 8):
+        def tobytes(self, order: Literal["C", "F", "A"] | None = None) -> bytes: ...
```
JelleZijlstra, line 855: Looking at the code the default is also effectively `C` in 3.9, there just was no Argument Clinic yet to put the default: https://github.com/python/cpython/blob/c33aaa9d559398bbf2b80e891bf3ae6a716e4b8c/Objects/memoryobject.c#L2139.

Seems fine to keep as is though so we don't have to deal with stubtest allowlists.

stdlib/distutils/log.pyi
```diff
@@ -7,7 +8,11 @@ ERROR: int
 FATAL: int
 
 class Log:
-    def __init__(self, threshold: int = ...) -> None: ...
+    if sys.version_info >= (3, 9):
+        def __init__(self, threshold: int = 3) -> None: ...
+    else:
+        def __init__(self, threshold: int = 30) -> None: ...
```
JelleZijlstra, line 14: Not sure what's going on but it looks to me like the default is also 3 (=`WARN`) in 3.7: https://github.com/python/cpython/blob/d729c5c9d49df41d5986be7507b98cbec1b7349e/Lib/distutils/log.py#L16

JelleZijlstra (approved):
Looks good apart from one weird case in distutils.

stdlib/distutils/log.pyi
AlexWaygood: Really weird. This is what I get locally on 3.7:

```pycon
>>> import distutils.log
>>> help(distutils.log.Log.__init__)
Help on function __init__ in module distutils.log:

__init__(self, threshold=30)
    Initialize the logger with a name and an optional level.
```

My local 3.7 environment is using conda -- I can only assume something's wrong with my local environment somehow? Anyhow, I'll revert this one.

JelleZijlstra: Maybe stuptools is working its magic.

JelleZijlstra: Interestingly I get this too:

```
Python 3.7.16 (default, Dec 26 2022, 20:28:00) 
[Clang 14.0.0 (clang-1400.0.29.202)] on darwin
>>> import distutils.log
>>> distutils.log.Log
<class 'distutils.log.Log'>
>>> distutils.log.Log.__init__.__defaults__
(30,)
```

And it's indeed setuptools' fault:

```
>>> distutils.log
<module 'distutils.log' from '/Users/jelle/.pyenv/versions/3.7.16/lib/python3.7/site-packages/setuptools/_distutils/log.py'>
```

AlexWaygood: > And it's indeed setuptools' fault:

classic

PR #9636, opened 2023-01-31
Improve `statistics.median_grouped` on <=3.10

AlexWaygood:
Providing an `int` or a `float` for the `interval` parameter appears to work, regardless of what weird and whacky inputs you supply for the `data` parameter:

```pycon
>>> import statistics
>>> statistics.median_grouped([1, 5, 6, 8, 10], interval=2.0)
6.0
>>> statistics.median_grouped([1, 5, 6, 8, 10], interval=2)
6.0
>>> statistics.median_grouped([1.5, 5, 6, 8, 10], interval=2.0)
6.0
>>> from fractions import Fraction
>>> statistics.median_grouped([1.5, Fraction(5, 5), 6, 8, 10], interval=2.0)
6.0
>>> statistics.median_grouped([Fraction(1, 2), Fraction(4, 5), Fraction(2, 3)], interval=Fraction(1, 2))
0.6666666666666667
>>> from decimal import Decimal
>>> statistics.median_grouped([Decimal("0.5"), Decimal("0.7")], interval=4)
-1.3
```

(The union return is still required: if you supply a sequence with length exactly 1 to the `data` parameter, the sole member of the sequence is returned as-is, without any coercion to `float`.)


JelleZijlstra:
Seems like we messed something up in mypy-primer.

AlexWaygood:
> Seems like we messed something up in mypy-primer.

The diffs were well formatted in #9629, so I don't _think_ this is related to any changes we've made over at typeshed. Something else must have changed.

JelleZijlstra:
https://github.com/hauntsaninja/mypy_primer/commit/bb842b55e22c29891b1c74158489ec1225b9451f should fix.

PR #9637, opened 2023-01-31
Improve a few version-dependent tkinter functions

AlexWaygood:
On 3.8+, the second argument of `_tkinter.create` defaults to `""` rather than `None`, and `TypeError` is raised if a non-`str` is passed in:

```pycon
>>> import _tkinter
>>> _tkinter.create("foo", None)
TypeError: create() argument 2 must be str, not None
```

Meanwhile, on 3.9+, `tkinter.Tk.__init__` and `tkinter.Tcl` have `bool` defaults, which can be added to the stub. But on <=3.8, these functions have `int` defaults, which can't be added to the stub (since the parameters are annotated with `bool`, and I don't think we should change that).

### DIFF
--- a/stdlib/tkinter/__init__.pyi
+++ b/stdlib/tkinter/__init__.pyi
@@ -728,29 +728,18 @@ class _ExceptionReportingCallback(Protocol):
 
 class Tk(Misc, Wm):
     master: None
-    # Make sure to keep __init__ in sync with other functions that use the same args.
-    # Use `git grep screenName` to find them
-    if sys.version_info >= (3, 9):
-        def __init__(
-            self,
-            screenName: str | None = None,
-            baseName: str | None = None,
-            className: str = "Tk",
-            useTk: bool = True,
-            sync: bool = False,
-            use: str | None = None,
-        ) -> None: ...
-    else:
-        def __init__(
-            self,
-            screenName: str | None = None,
-            baseName: str | None = None,
-            className: str = "Tk",
-            useTk: bool = ...,
-            sync: bool = ...,
-            use: str | None = None,
-        ) -> None: ...
-
+    def __init__(
+        # Make sure to keep in sync with other functions that use the same
+        # args.
+        # use `git grep screenName` to find them
+        self,
+        screenName: str | None = None,
+        baseName: str | None = None,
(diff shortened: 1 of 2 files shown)


Avasam:
> on <=3.8, these functions have int defaults

Does it default to 0 or 1? That's something I've seen a lot in older libraries (especially those originally written in Python 2). If they're used as boolean anyway internally (and not stored, ie as attributes), I think it makes sense to just expose them as bools in the stubs. `bool | Literal[0,1]` if your reaaaaaly want to 😅

stdlib/tkinter/__init__.pyi
```diff
@@ -728,18 +728,29 @@ class _ExceptionReportingCallback(Protocol):
...
+            sync: bool = False,
+            use: str | None = None,
+        ) -> None: ...
+    else:
+        def __init__(
+            self,
+            screenName: str | None = None,
+            baseName: str | None = None,
+            className: str = "Tk",
+            useTk: bool = ...,
```
Akuli, line 749: Can't we just allowlist this? It would be useful to see `True` and `False` in IDEs, even if you don't happen to be using a new Python version.

AlexWaygood: Sure!

AlexWaygood: No allowlist additions needed for now, but we'll have to add some when mypy 1.0 comes out and stubtest starts verifying default values.

stdlib/tkinter/__init__.pyi
```diff
@@ -800,7 +800,7 @@ class Tk(Misc, Wm):
     def wantobjects(self, *args, **kwargs): ...
     def willdispatch(self): ...
 
-def Tcl(screenName: str | None = None, baseName: str | None = None, className: str = "Tk", useTk: bool = ...) -> Tk: ...
+def Tcl(screenName: str | None = None, baseName: str | None = None, className: str = "Tk", useTk: bool = False) -> Tk: ...
```
Akuli, line 803: The `useTk` argument here is a bit weird. It never makes sense to set it to `True`, because you are supposed to instantiate `Tk()` directly.

Akuli (approved):

PR #9639, opened 2023-01-31
Add defaults for params and constants in PyAutoGUI

Avasam:
1. ran https://github.com/JelleZijlstra/stubdefaulter
2. manually added default kwarg parameters and constants


AlexWaygood (approved):
Looks good other than the pytype crash!

PR #9641, opened 2023-01-31
Add defaults for params and constants in pyscreeze

Avasam:
1. Ran https://github.com/JelleZijlstra/stubdefaulter
2. Manually added default kwarg parameters
3. Manually added defaults and Final for constants

### DIFF
--- a/stubs/PyScreeze/pyscreeze/__init__.pyi
+++ b/stubs/PyScreeze/pyscreeze/__init__.pyi
@@ -47,11 +47,11 @@ def locate(
     needleImage: str | Image.Image | _Mat,
     haystackImage: str | Image.Image | _Mat,
     *,
-    grayscale: bool | None = ...,
-    limit: _Unused = ...,
-    region: tuple[int, int, int, int] | None = ...,
-    step: int = ...,
-    confidence: SupportsFloat | SupportsIndex | str = ...,
+    grayscale: bool | None = None,
+    limit: _Unused = 1,
+    region: tuple[int, int, int, int] | None = None,
+    step: int = 1,
+    confidence: SupportsFloat | SupportsIndex | str = 0.999,
 ) -> Box | None: ...
 
 # _locateAll_python / _locateAll_pillow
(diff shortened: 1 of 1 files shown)


JelleZijlstra:
> Manually added default kwarg parameters

Hm is there something missing in stubdefaulter around this? If so please let me or Alex know so we can fix it.

Avasam:
> > Manually added default kwarg parameters
> 
> Hm is there something missing in stubdefaulter around this? If so please let me or Alex know so we can fix it.

Nah, it's **kwargs passed to another function. We expand the kwargs in the stubs. Not anything stubdefaulter could know about (nor does stubtest).

I got 1 false positive from stubdefaulter though, the signature of `locateAll` depends on if openCV is installed. (it becomes `_locateAll_opencv` or `_locateAll_python`). In the stubs, we represent this through an overload, and stubdefaulter  gave the same defaults to both overloads.

JelleZijlstra:
Thanks for the explanation! For `@overload` we currently just fill all defaults from the implementation (I reasoned about why at JelleZijlstra/stubdefaulter#22). That's been helping @AlexWaygood find a bunch of incorrect stubs in the stdlib, so I think on balance it's the right call. Your case here sounds very unusual.

Avasam:
> [...] so I think on balance it's the right call. Your case here sounds very unusual.

I agree. It's a special case that required taking a special look at it. It was also obvious to spot and other tools caught it.

stubs/PyScreeze/pyscreeze/__init__.pyi
```diff
@@ -16,11 +16,12 @@ _R = TypeVar("_R")
...
 _Mat: TypeAlias = Incomplete
 
-useOpenCV: bool
-RUNNING_PYTHON_2: Literal[False]
-GRAYSCALE_DEFAULT: Literal[False]
+useOpenCV: Final[bool]
+RUNNING_PYTHON_2: Final = False
+GRAYSCALE_DEFAULT: Final = False
+scrotExists: Final[bool]
+# Meant to be overrisable for backward-compatibility
```
JelleZijlstra, line 23:
```suggestion
# Meant to be overridable for backward-compatibility
```

JelleZijlstra (approved):

PR #9642, opened 2023-01-31
Add defaults for params and constants in python-xlib

Avasam:
1. Ran https://github.com/JelleZijlstra/stubdefaulter
2. Manually added default kwarg parameters
3. Manually added defaults and Final for constants


Avasam:
Weird how locally (Windows and WSL) I have to ignore allowlist this, but not on the CI. Anyway this should be better in the next mypy version from what I understand. https://github.com/python/typeshed/actions/runs/4058762277/jobs/6986074581#step:5:57

Avasam:
I'm guessing Hex represented ints aren't considered acceptable "simple ints" in Flake8-PYI. If that's the case and not just an oversight, I'll update the values.

Avasam:
String defaults with spaces or `-` in them fails pytype 🤔 🙃

AlexWaygood:
> I'm guessing Hex represented ints aren't considered acceptable "simple ints" in Flake8-PYI. If that's the case and not just an oversight, I'll update the values.

Any number where the string representation is >7 characters long is currently considered unacceptable according to flake8-pyi. It's very arbitrary, but we had to draw the line somewhere, and the exact value probably isn't very useful for very big numbers anyway :)

JelleZijlstra:
I don't think it's about hex, the limitation is on how big the value is. The theory is that large defaults are probably not useful for users. My initial stdlib defaults PR added a bunch of defaults based on random settings on my machine. We could revisit that decision, though; the defaults flake8-pyi is complaining about here seems reasonable.

The pytype bug indicates that it's trying to parse the defaults as type annotations. I fixed one place where it did that, but maybe you've found another.

Avasam:
Ah I see! Indeed hex representation of 32-bit integers tends to be 10 chars ("0x" + 8*[0-F]). `0xFFFFFFFF` --> `4294967295` which is also 10 chars. So maybe 10 could be a slightly less arbitrary limit?

AlexWaygood:
> Ah I see! Indeed hex representation of 32-bit integers tends to be 10 chars ("0x" + 8*[0-F]). `0xFFFFFFFF` --> `4294967295` which is also 10 chars. So maybe 10 could be a slightly less arbitrary limit? I'll update this PR after a decision has been taken on that.

Sure, sounds reasonable! `sys.maxsize` is 19 digits on my machine, so flake8-pyi will still complain when stubdefaulter tries to add defaults which are `=sys.maxsize` at runtime. Feel free to make a flake8-pyi PR if @JelleZijlstra also agrees, and I can do a hotfix release for you :)

stubs/python-xlib/Xlib/X.pyi
```diff
@@ -1,345 +1,347 @@
...
+ExposureMask: Final = 32768
+VisibilityChangeMask: Final = 65536
+StructureNotifyMask: Final = 131072
+ResizeRedirectMask: Final = 262144
+SubstructureNotifyMask: Final = 524288
+SubstructureRedirectMask: Final = 1048576
+FocusChangeMask: Final = 2097152
+PropertyChangeMask: Final = 4194304
+ColormapChangeMask: Final = 8388608
+OwnerGrabButtonMask: Final = 16777216
```
AlexWaygood, line 40: If we used octal or hexadecimal for these, it might be more readable -- it would be more obvious that these are all powers of 2.

(Similar for a bunch of other constants in this file.)

AlexWaygood (commented):
One nit, otherwise this looks good.

Note that our CI doesn't verify defaults for global constants at the moment (only defaults for function and method parameters). I filed https://github.com/python/mypy/pull/14568 to address that, though the PR could be improved a bit; I need to get back to it. But I'm happy to trust you that the defaults for the global constants are all correct 😄

stubs/python-xlib/Xlib/X.pyi
JelleZijlstra: Ping on this, I do agree that hex would probably work better here.

AlexWaygood:
> (I've been quite occupied last week, and I got sick this week, will go through reviews soon once I feel a bit better)

No worries at all -- I hope you get well soon!

AlexWaygood (commented):
A few more places I spotted where I think hexadecimal might work better

stubs/python-xlib/Xlib/ext/xinput.pyi
```diff
@@ -11,110 +11,110 @@ from Xlib.xobject import drawable, resource
...
+NotifyNonlinearVirtual: Final = 4
+NotifyPointer: Final = 5
+NotifyPointerRoot: Final = 6
+NotifyDetailNone: Final = 7
+GrabtypeButton: Final = 0
+GrabtypeKeycode: Final = 1
+GrabtypeEnter: Final = 2
+GrabtypeFocusIn: Final = 3
+GrabtypeTouchBegin: Final = 4
+AnyModifier: Final = 2147483648
```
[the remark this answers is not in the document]
Avasam: Ah, my search was too simple, I only looked for `1<<` not `1 <<` >.<

AlexWaygood (approved):
Thanks!

PR #9643, opened 2023-01-31
Bump flake8-pyi to 23.1.2

AlexWaygood:


AlexWaygood:
The third-party stubtest failures are not caused by this PR.

PR #9623, opened 2023-01-30
[DO-NOT-MERGE] Checking to see if pyright verifies third-party test cases against false negatives correctly

AlexWaygood:
If our test-case setup is working correctly, pyright should error out with a complaint about an unused `type: ignore` comment.

PR #9650, opened 2023-02-01
Improve pyright verification of third-party test cases in CI

AlexWaygood:
As seen in #9623, following some recent changes to our pyright config, pyright no longer verifies our checks against false-negative errors that we have in our third-party test cases. (Our standard-library test cases are fine.) Luckily, the latest release of pyright means that we can use the same pyright config for our third-party test cases as we do for our standard-library test cases, by using wildcards in `include` and `exclude` patterns. This means that we can remove a no-longer-necessary test from `check_consistent.py`, and it means that we no longer have to have `# pyright: reportUnnecessaryTypeIgnoreComment=true` comments at the top of every third-party test-case file.

Also in this PR:

- Add a new test to `check_consistent.py`, ensuring that `.py` files are only found in the `@tests/test_cases/` directory, and not at the top level of the `@tests/` directory. This isn't necessarily obvious, and caused confusion in #9323.
- Change the workflow file so that the pyright configs are run in a more logical order. The old order was: "strict config first on a subset of the stubs; then, if that succeeds, do the test-cases config on the stdlib test cases; then, if _that_ succeeds, run the basic config on all the stubs". The new order is: "run the basic config first on all the stubs; then, if that succeeds, run the strict config on a subset of the stubs; then, if _that_ succeeds, run the test-case config on the stdlib and third-party test cases".

Co-authored-by: Avasam <samuel.06@hotmail.com>

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -146,6 +146,12 @@ jobs:
           python-platform: ${{ matrix.python-platform }}
           python-version: ${{ matrix.python-version }}
           no-comments: ${{ matrix.python-version != '3.10' || matrix.python-platform != 'Linux' }} # Having each job create the same comment is too noisy.
+      - uses: jakebailey/pyright-action@v1
+        with:
+          version: ${{ steps.pyright_version.outputs.value }}
+          python-platform: ${{ matrix.python-platform }}
+          python-version: ${{ matrix.python-version }}
+          no-comments: ${{ matrix.python-version != '3.10' || matrix.python-platform != 'Linux' }} # Having each job create the same comment is too noisy.
           project: ./pyrightconfig.stricter.json
       - uses: jakebailey/pyright-action@v1
         with:
@@ -154,12 +160,6 @@ jobs:
           python-version: ${{ matrix.python-version }}
           no-comments: ${{ matrix.python-version != '3.10' || matrix.python-platform != 'Linux' }} # Having each job create the same comment is too noisy.
           project: ./pyrightconfig.testcases.json
-      - uses: jakebailey/pyright-action@v1
-        with:
-          version: ${{ steps.pyright_version.outputs.value }}
-          python-platform: ${{ matrix.python-platform }}
-          python-version: ${{ matrix.python-version }}
-          no-comments: ${{ matrix.python-version != '3.10' || matrix.python-platform != 'Linux' }} # Having each job create the same comment is too noisy.
 
   stub-uploader:
     name: Run the stub_uploader tests
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -140,20 +140,23 @@ jobs:
         with:
           file: "pyproject.toml"
           field: "tool.typeshed.pyright_version"
-      - uses: jakebailey/pyright-action@v1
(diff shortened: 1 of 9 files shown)


AlexWaygood:
Here's a failing run showing that, with these changes, pyright will once again error for unused `type: ignore` comments in third-party test-case files: https://github.com/python/typeshed/actions/runs/4065870433/jobs/7001171534

.github/workflows/tests.yml
```diff
@@ -140,26 +140,29 @@ jobs:
...
           field: "tool.typeshed.pyright_version"
-      - uses: jakebailey/pyright-action@v1
+      - name: Run pyright with basic settings on all the stubs
+        uses: jakebailey/pyright-action@v1
         with:
           version: ${{ steps.pyright_version.outputs.value }}
           python-platform: ${{ matrix.python-platform }}
           python-version: ${{ matrix.python-version }}
           no-comments: ${{ matrix.python-version != '3.10' || matrix.python-platform != 'Linux' }} # Having each job create the same comment is too noisy.
-          project: ./pyrightconfig.stricter.json
```
Avasam, line 149: I think running stricter first would fail faster in most cases. Since they both run on all files not excluded by the stricter settings. Stricter runs on less files and a failure with basic config would also result in a failure with stricter ones.

But this job is also pretty fast anyway.

AlexWaygood: Also I've found it pretty annoying in the past when we're working to enable new pyright lints, the way you have to play "pyright whackamole" with our current setup. You think you've got all the pyright errors sorted, but then you find out, oh no, it was only the strict-config errors you've got sorted! Most of typeshed is still to go, it's just we never even got to the second step 😶

This order feels more intuitive to me :)

tests/check_consistent.py
```diff
@@ -65,24 +65,22 @@ def check_stubs() -> None:
         allowed = {"METADATA.toml", "README", "README.md", "README.rst", "@tests"}
         assert_consistent_filetypes(dist, kind=".pyi", allowed=allowed)
 
+        tests_dir = dist / "@tests"
+        if tests_dir.exists() and tests_dir.is_dir():
+            py_files_present = bool([file for file in tests_dir.iterdir() if file.suffix == ".py"])
```
JelleZijlstra, line 70:
```suggestion
            py_files_present = any(file for file in tests_dir.iterdir() if file.suffix == ".py")
```
This reads more naturally but has the somewhat undesirable effect of relying on `bool(file)`. Feel free to ignore if you like your version better.

JelleZijlstra (approved):

tests/check_consistent.py
AlexWaygood: I considered doing this -- would you like that more?

```suggestion
            py_files_present = len([file for file in tests_dir.iterdir() if file.suffix == ".py"]) > 0
```

JelleZijlstra: Either is fine. I guess the part that bothers me is constructing a list just to check that it's not empty, but it's not worth doing something more complicated.

AlexWaygood: Oh wait, we can just do this, right?

```suggestion
            py_files_present = any(file.suffix == ".py" for file in tests_dir.iterdir())
```

PR #9651, opened 2023-02-01
Stdlib: add some very large integer defaults

AlexWaygood:
Previously numbers this big were disallowed by flake8-pyi, but we relaxed that rule slightly in the latest release


stdlib/asyncio/windows_events.pyi
```diff
@@ -42,7 +42,7 @@ if sys.platform == "win32":
         ) -> list[PipeServer]: ...
 
     class IocpProactor:
-        def __init__(self, concurrency: int = ...) -> None: ...
+        def __init__(self, concurrency: int = 4294967295) -> None: ...
```
JelleZijlstra, line 45:
```suggestion
        def __init__(self, concurrency: int = 0xffffffff) -> None: ...
```
This is what the source code says and it looks a lot less random.

PR #9653, opened 2023-02-01
Mark SQLAlchemy stubs with `ignore_missing_stub = false`

Avasam:
Follow-up to https://github.com/python/typeshed/pull/9596 (keeping as draft until that one is merged)
Mostly used stubgen for the missing types. Except for `stubs/SQLAlchemy/sqlalchemy/sql/annotation.pyi` which had to be done manually.

Changes to imports:
- Standardize relative vs absolute imports (this stub seems to prefer relative imports)
- Prefer importing directly from the source module, rather than using re-exports (which was often needed for the next point)
- Update imports to work around https://github.com/google/pytype/issues/1087
- Allowlist various compat module re-exports

There's still a pytype issue I'm not certain about.


Avasam:
I don't understand what this Pytype error is trying to tell me. Other than that, this PR is ready to be looked at.

JelleZijlstra (approved):

PR #9654, opened 2023-02-01
Fix optional USB attributes in pyserial's `ListPortInfo`

hamdanal:
Some USB specific attributes in pyserial's [`serial.tools.list_port_common.ListPortInfo`](https://github.com/python/typeshed/blob/main/stubs/pyserial/serial/tools/list_ports_common.pyi#L11-L24) class are not always available and should be marked as optional. They depend on the USB device and its driver correctly reporting these attributes. I discovered this recently with a new device that does not report its serial number. Only the Vendor ID `vid` and Product ID `pid` are guaranteed (This can be seen [here](https://github.com/pyserial/pyserial/blob/master/serial/tools/list_ports_linux.py#L52-L62) where `vid` and `pid` are always cast as `int` while other attributes are left as `str | None` for USB devices).

This is a follow up to #9347 and the discussion at https://github.com/python/typeshed/pull/9347#issuecomment-1358245865 (CC. @hauntsaninja)

### DIFF
--- a/stubs/pyserial/serial/tools/list_ports_common.pyi
+++ b/stubs/pyserial/serial/tools/list_ports_common.pyi
@@ -8,20 +8,21 @@ class ListPortInfo:
     name: str
     description: str
     hwid: str
-    # USB specific data: the attributes below are specific to USB devices only and should be marked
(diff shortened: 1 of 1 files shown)


hauntsaninja (approved):
Thanks for strictifying this!

PR #9655, opened 2023-02-01
stdlib: Add defaults for positional-only parameters

AlexWaygood:
The first four commits were created using `stubdefaulter` on various versions of Python; the rest were done manually.

Locally, I've run mypy-master stubtest on this PR using Python 3.7, 3.8, 3.9, 3.10 and 3.11.


JelleZijlstra (approved):

AlexWaygood:
@JelleZijlstra, it might be worth you running `stubdefaulter` again yourself, since you're not on Windows -- I think there's lots of posix-only modules I'm missing when I run `stubdefaulter` on my machine :)

JelleZijlstra:
Done, I'll also do a Linux run when I get a chance.

PR #9656, opened 2023-02-01
`unicodedata`: overload many functions and methods

AlexWaygood:
The stub currently says that these functions/methods have default values, but that isn't really true -- if you don't supply a value for the second parameter, `ValueError` is raised when you try to pass in a value for the first parameter that doesn't exist in the unicode database


AlexWaygood:
```pycon
>>> import unicodedata as u
>>> u.name("\n")
ValueError: no such name
>>> u.name("\n", 6)
6
>>> u.numeric("x")
ValueError: not a numeric character
>>> u.numeric("x", 6)
6
>>> u.digit("x")
ValueError: not a digit
>>> u.digit("x", 6)
6
>>> u.decimal("x")
ValueError: not a decimal
>>> u.decimal("x", 6)
6

PR #9658, opened 2023-02-01
Remove default for `size` parameter to `multiprocessing.heap.Heap`

JelleZijlstra:
It's system-dependent. stubdefaulter told me `multiprocessing.heap.Heap: parameter size: stub default 4096 != runtime default 16384`, which presumably means it's 4096 on @AlexWaygood's system and 16384 on mine. I looked in the code and the default is set to `mmap.PAGESIZE`, which in turn is set from some system call at import time.


AlexWaygood (approved):
Oops, I think I took this out in #9572 and then accidentally added it back in, in a later PR

PR #9659, opened 2023-02-01
stdlib: add defaults for modules accessible on MacOS

JelleZijlstra:
Ran stubdefaulter with standard options on a Mac.


AlexWaygood (approved):
All looks highly plausible!

PR #9662, opened 2023-02-02
Fix python-jose unverified jws and jwt functions

hasier:
The documentation for these functions specify `str`, but the underlying logic accepts `bytes`. Similarly, the response comes in raw `bytes` rather than `str`s, contrary to what the docs say.


stubs/python-jose/jose/jws.pyi
```diff
@@ -19,6 +19,6 @@ def verify(
     algorithms: str | Container[str] | None,
     verify: bool = ...,
 ) -> bytes: ...
-def get_unverified_header(token: str) -> dict[str, Any]: ...
-def get_unverified_headers(token: str) -> dict[str, Any]: ...
-def get_unverified_claims(token: str) -> str: ...
+def get_unverified_header(token: str | bytes) -> dict[str, Any]: ...
+def get_unverified_headers(token: str | bytes) -> dict[str, Any]: ...
+def get_unverified_claims(token: str | bytes) -> bytes: ...
```
hasier, line 24: All 3 use `_load` underneath.
- `_load` checking for `str` input and converting to `bytes` https://github.com/mpdavis/python-jose/blob/96474ecfb6ad3ce16f41b0814ab5126d58725e2a/jose/jws.py#L173-L174
- `_load` uses `base64url_decode` to load the claims https://github.com/mpdavis/python-jose/blob/96474ecfb6ad3ce16f41b0814ab5126d58725e2a/jose/jws.py#L193
  - `base64url_decode` expects and returns `bytes` https://github.com/mpdavis/python-jose/blob/96474ecfb6ad3ce16f41b0814ab5126d58725e2a/jose/utils.py#L66-L78

Typing for `base64url_decode` is already correct https://github.com/python/typeshed/blob/f1aede7162a3ee497e618a71dc4b42447b74abb1/stubs/python-jose/jose/utils.pyi#L11

AlexWaygood (approved):
Thank you for the detailed comments explaining the motivation for the changes -- it made it very easy to review!

PR #9666, opened 2023-02-03
Update `cachetools` stubs to 5.3.*

AlexWaygood:
Closes #9579

### DIFF
--- a/stubs/cachetools/METADATA.toml
+++ b/stubs/cachetools/METADATA.toml
@@ -1,4 +1,4 @@
-version = "5.2.*"
+version = "5.3.*"
 
 [tool.stubtest]
 ignore_missing_stub = false
--- a/stubs/cachetools/cachetools/__init__.pyi
+++ b/stubs/cachetools/cachetools/__init__.pyi
@@ -97,7 +97,10 @@ class TLRUCache(_TimedCache[_KT, _VT]):
     def expire(self, time: float | None = ...) -> None: ...
 
 def cached(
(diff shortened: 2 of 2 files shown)


srittau (approved):

bfontaine:
If I understand it correctly, this is automatically generated. How to fix an issue with the generated typing, then?

This is the generated type hints for `TLRUCache`:
```python
class TLRUCache(_TimedCache[_KT, _VT]):
    def __init__(
        self,
        maxsize: float,
        ttu: Callable[[_KT, _VT, float], float],
        timer: Callable[[], float] = ...,
        getsizeof: Callable[[_VT], float] | None = ...,
    ) -> None: ...
    @property
    def ttu(self) -> Callable[[_KT, _VT, float], float]: ...
    def expire(self, time: float | None = ...) -> None: ...
```

However in the [upstream documentation](https://cachetools.readthedocs.io/en/latest/#cachetools.TLRUCache), an example code uses a `timer` that return a `datetime` object, and so a `ttu` callable that takes and return a `datetime`. That example is not valid against these type hints.

To support this example, type hints should use a generic type `T` instead of a `float`, so the code would look like this:

```python
T = TypeVar("T")

class TLRUCache(_TimedCache[_KT, _VT]):
    def __init__(
        self,
        maxsize: float,
        ttu: Callable[[_KT, _VT, T], T],
        timer: Callable[[], T] = ...,
        getsizeof: Callable[[_VT], float] | None = ...,
    ) -> None: ...
    @property
    def ttu(self) -> Callable[[_KT, _VT, T], T]: ...
    def expire(self, time: T | None = ...) -> None: ...
```

How could one fix this in a way that’s compatible with future updates to this file?

PR #9671, opened 2023-02-04
Add markdown.treeprocessors.UnescapeTreeprocessor

andersk:
This replaced the deprecated `markdown.postprocessors.UnescapePostprocessor` in https://github.com/Python-Markdown/markdown/pull/1272.

### DIFF
--- a/stubs/Markdown/markdown/treeprocessors.pyi
+++ b/stubs/Markdown/markdown/treeprocessors.pyi
@@ -1,5 +1,7 @@
 from _typeshed import Incomplete
-from typing import Any
+from re import Pattern
+from typing import Any, ClassVar
+from xml.etree.ElementTree import Element
 
 from . import util
 
@@ -7,7 +9,7 @@ def build_treeprocessors(md, **kwargs): ...
 def isString(s): ...
 
 class Treeprocessor(util.Processor):
-    def run(self, root) -> Any | None: ...
+    def run(self, root: Element) -> Element | None: ...
 
 class InlineProcessor(Treeprocessor):
     inlinePatterns: Any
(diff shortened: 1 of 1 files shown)


stubs/Markdown/markdown/treeprocessors.pyi
```diff
@@ -1,20 +1,26 @@
...
+from xml.etree.ElementTree import Element
 
 from . import util
 
 def build_treeprocessors(md, **kwargs): ...
 def isString(s): ...
 
 class Treeprocessor(util.Processor):
-    def run(self, root) -> Any | None: ...
+    def run(self, root: Element) -> Element | None: ...
```
AlexWaygood, line 12: The docstring refers to this method as taking an `ElementTree` object, and `xml.etree.ElementTree.ElementTree` is a different class to `xml.etree.ElementTree.Element`. Is the docstring incorrect?

stubs/Markdown/markdown/treeprocessors.pyi
```diff
@@ -1,20 +1,26 @@
...
     def __init__(self, md) -> None: ...
     stashed_nodes: Any
     parent_map: Any
-    def run(self, tree, ancestors: Incomplete | None = ...): ...
+    def run(self, tree: Element, ancestors: Incomplete | None = ...) -> Element: ...
 
 class PrettifyTreeprocessor(Treeprocessor): ...
+
+class UnescapeTreeprocessor(Treeprocessor):
+    RE: Pattern[str]
```
AlexWaygood, line 25: Looks like this should be a `ClassVar` (you'll have to import `ClassVar` from `typing`):

```suggestion
    RE: ClassVar[Pattern[str]]
```

stubs/Markdown/markdown/treeprocessors.pyi
andersk: Good question, but yes, it’s really an `Element`, not an `ElementTree`: note [this `.getroot()` call](https://github.com/Python-Markdown/markdown/blob/3.4.1/markdown/core.py#L264).

AlexWaygood: Thanks! I hate typing XML 🙃

AlexWaygood (approved):

PR #9672, opened 2023-02-04
stdlib: use `bool` for annotations where the default is `False`

AlexWaygood:
Found by grepping for `: int = False`

(I also grepped for `: int = True`, but there weren't any hits.)


AlexWaygood:
It looks like a big fallout because of the verbose error messages, but in actual fact that's only 3 new hits being reported in user code.

I could also just revert the changes to `builtins.compile()`, however. Curious for other people's thoughts.

Akuli (approved):
I like the simple defaults in stubs :)

Akuli:
I think it's fine to get a new mypy error for using an integer as the `compile()` argument. The documentation of `compile()` talks about zero and nonzero integers, but was written before Python had a `bool` type (Python 2.4 docs have the same text).

AlexWaygood:
Then let's do it.

PR #9675, opened 2023-02-04
Update `mypy_extensions` stubs to 1.0.x

JukkaL:
Some context about how mypy and mypyc use the added types:
https://github.com/python/mypy_extensions/pull/31


JukkaL:
The types are magical and stubtest can't really do anything useful with them. We may need to disable stubtest here.

AlexWaygood:
> The types are magical and stubtest can't really do anything useful with them. We may need to disable stubtest here.

You can add regular expressions to the [stubtest allowlist](https://github.com/python/typeshed/blob/main/stubs/mypy-extensions/%40tests/stubtest_allowlist.txt). These additions should fix all stubtest errors:

```
i64\..*
i32\..*
i16\..*
u8\..*
```

AlexWaygood (approved):

PR #9676, opened 2023-02-04
Export `DataclassInstance` protocol from `_typeshed`

cdce8p:
Add `DataclassInstance` protocol to `_typeshed`. Especially when dealing with mixin classes, it's quite common to have to use it. Moving it to `_typeshed` means users can import it from there without having to reimplement it themselves.

Followup to https://github.com/python/typeshed/pull/9362#issuecomment-1416758036
/CC  @AlexWaygood


AlexWaygood (commented):
Any reason not to just put it in `_typeshed/__init__.pyi`? It feels a bit small to go in its own file. We can always move it to a dataclasses-specific file later if we add more dataclasses-specific protocols or aliases.

Also: I'd like to add a comment noting that the protocol is quasi-experimental, and linking to the pyright issue. The type won't work as intended for pyright users unless and until the pyright issue is reconsidered (or we think of another way of doing this in the stubs).

AlexWaygood (approved):

AlexWaygood:
Thanks!

PR #9677, opened 2023-02-04
colorama: Preserve input type for `wrap_stream`

JelleZijlstra:
Implementation looks like this:

```python
def wrap_stream(stream, convert, strip, autoreset, wrap):
    if wrap:
        wrapper = AnsiToWin32(stream,
            convert=convert, strip=strip, autoreset=autoreset)
        if wrapper.should_wrap():
            stream = wrapper.stream
    return stream
```

### DIFF
--- a/stubs/colorama/colorama/initialise.pyi
+++ b/stubs/colorama/colorama/initialise.pyi
@@ -1,7 +1,9 @@
 from contextlib import AbstractContextManager
-from typing import Any, TextIO
+from typing import Any, TextIO, TypeVar
 
 from .ansitowin32 import StreamWrapper
+
+_TextIOT = TypeVar("_TextIOT", bound=TextIO)
 
 orig_stdout: TextIO | None
 orig_stderr: TextIO | None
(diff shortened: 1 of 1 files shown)


JelleZijlstra:
My motivation here is to make that error in Black go away :)

AlexWaygood (approved):
If I understand correctly, in some circumstances this actually returns a `colorama.ansitowin32.StreamWrapper` instance rather than the object that was passed in. But `colorama.ansitowin32.StreamWrapper` objects just [delegate nearly all attribute access to the underlying stream anyway](https://github.com/tartley/colorama/blob/21c4b94fe21ce29c85c896ace828da24b7527641/colorama/ansitowin32.py#L18-L20). So it basically comes to the same thing, and it's probably not worth worrying about.

JelleZijlstra:
@AlexWaygood the stub reflects that, the function returns a union. Thanks for taking a look!

AlexWaygood:
> @AlexWaygood the stub reflects that, the function returns a union. Thanks for taking a look!

Hah, so it does :)

PR #9678, opened 2023-02-04
Add better types for asyncio.gather

Gobot1234:
Before:
image

After: 
image


Gobot1234:
Looking at #9105 I don't think the case presented where you have vargs of different types actually happens that much whereas unpacking as shown in the screenshots above is much more common and helps catch errors that would have gone amiss.

AlexWaygood:
This looks promising.

The `homeassistant` hit is here: https://github.com/home-assistant/core/blob/04e05af44cbea2156d904d0d34e4e0d41c9725c1/homeassistant/components/amcrest/camera.py#L404-L418. They _just_ miss out on the precise overloads, since they're passing in 6 coroutines. Previously the return type of the call to `asyncio.gather` was inferred as `list[Any]`; with this change, it is inferred as `list[object]`. That's probably not too terrible; and we could probably get rid of the hit by just adding an overload for if you pass in six coroutines.

Some of the test cases in https://github.com/python/typeshed/blob/main/test_cases/stdlib/asyncio/check_gather.py are failing, and will need to be updated.

Lastly, as well as adding the new overload, we should probably also update the existing fallback overloads here to be more precise as well:

https://github.com/python/typeshed/blob/cd5f976968c64b52a968f87fc5b81e8fbd7afac6/stdlib/asyncio/tasks.pyi#L142-L143

https://github.com/python/typeshed/blob/cd5f976968c64b52a968f87fc5b81e8fbd7afac6/stdlib/asyncio/tasks.pyi#L232-L235

They should probably be changed to:

```python
@overload
def gather(*coros_or_futures: _FutureLike[_T], return_exceptions: bool) -> Future[list[_T | BaseException]]: ...  # type: ignore[misc]

Gobot1234:
[home-assistant/core@04e05af/homeassistant/components/amcrest/camera.py#L404-L418](https://github.com/home-assistant/core/blob/04e05af44cbea2156d904d0d34e4e0d41c9725c1/homeassistant/components/amcrest/camera.py#L404-L418) would be inferred as list[str | bool] but I guess that can be solved as you said by adding another overload.

Gobot1234:
I don't think this is gonna work (with pyright at least) cause adding a special case for the 0 arg call (which is required to avoid the return of `await asyncio.gather()` being `list[BaseException]`, which thinking about it isn't technically wrong because it will always be empty) is preferred by `asyncio.gather(*iterable)` making the return for that `list[BaseException]`.

Gobot1234:
All of the failures look like mypy bugs and the failure in pyright is IMO acceptable I just don't know what the test needs to be changed to to pass

Gobot1234:
Hmm the paasta things seem like they do now break which is annoying. I'm not entirely sure what's wrong there though.

Gobot1234:
This seems to me like it should now work, I'm slightly confused by the behaviour I'm seeing by mypy and pyright here, unpacking a list seems to only give one element?

srittau (approved):
LGTM

AlexWaygood (approved):
This looks great, thanks for persevering with this! Pushed one more test just to be safe.

AlexWaygood:
> Mypy users are still going to get some things inferred as `list[object]` rather than `list[Any]`, which will likely be annoying for them:
> 
> ```python
> async def test_gather(awaitable1: Awaitable[int], awaitable2: Awaitable[str]) -> None:
>     x = await asyncio.gather(awaitable1, awaitable2, awaitable1, awaitable1, awaitable1, awaitable1, awaitable1)
>     assert_type(x, List[int])  # E: expression is of type "list[object]", not "list[int]"
> ```
> 
> But honestly, that's just a mypy join-vs-union problem (and I doubt there are many people manually passing in 7 or more items manually anyway). The pyright improvement here is definitely worth it, I think.
> 
> Thanks for persevering with this!!

Scratch that all, I made a mistake in the test case I wrote! Have pushed another test case which I think passes, just to be safe.

PR #9680, opened 2023-02-04
`keyword`: Improve types of `kwlist` and `softkwlist`

### LINKED ISSUE #9679: Incorrect types for `keyword.kwlist` and `keyword.softkwlist`
For some reason, `kwlist` and `softkwlist` typed as `Sequence[str]` instead of `list[str]`
Corresponding to `Lib/keyword.py`, it's pure lists.

Eclips4:
Resolves #9679

### DIFF
--- a/stdlib/keyword.pyi
+++ b/stdlib/keyword.pyi
@@ -1,5 +1,6 @@
 import sys
-from typing import Final, Sequence
+from collections.abc import Sequence
+from typing_extensions import Final
 
 if sys.version_info >= (3, 9):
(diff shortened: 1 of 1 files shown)


stdlib/keyword.pyi
```diff
@@ -1,5 +1,5 @@
 import sys
-from collections.abc import Sequence
+from typing import Final, Sequence
```
AlexWaygood, line 2:
```suggestion
from collections.abc import Sequence
from typing_extensions import Final
```

AlexWaygood (approved):
Thanks!

PR #9681, opened 2023-02-06
openpyxl: add a few type annotations

parched:
I've just added a few of the type annotations I specifically needed, but will add some more at a later date.


parched:
Just spotted #9487 and #9511, closing in favour of them.

PR #9682, opened 2023-02-06
`urllib.HTTPError.headers` is a settable property

sobolevn:
Code: https://github.com/python/cpython/blame/main/Lib/urllib/error.py#L61-L67

### DIFF
--- a/stdlib/urllib/error.pyi
+++ b/stdlib/urllib/error.pyi
@@ -11,6 +11,8 @@ class URLError(IOError):
 class HTTPError(URLError, addinfourl):
     @property
     def headers(self) -> Message: ...  # type: ignore[override]
+    @headers.setter
+    def headers(self, headers: Message) -> None: ...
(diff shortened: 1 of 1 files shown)


stdlib/urllib/error.pyi
```diff
@@ -11,6 +11,8 @@ class URLError(IOError):
 class HTTPError(URLError, addinfourl):
     @property
     def headers(self) -> Message: ...  # type: ignore[override]
```
AlexWaygood, line 13: From my local experimentation, looks like this `type: ignore` is no longer needed (not sure if that's new in this PR, or if it's been true for a while)

```suggestion
    def headers(self) -> Message: ...
```

AlexWaygood (approved):

sobolevn:
Let's try it!

hauntsaninja (approved):
So it is! :-)

PR #9683, opened 2023-02-06
Use `OSError` instead of `IOError`

sobolevn:
It is documented:

> The following exceptions are kept for compatibility with previous versions; starting from Python 3.3, they are aliases of [OSError](https://docs.python.org/3/library/exceptions.html#OSError).

Link: https://docs.python.org/3/library/exceptions.html#IOError

I think, that we should use `OSError` instead:
1. Because the source code uses `OSError`. Example: https://github.com/python/cpython/blame/main/Lib/urllib/error.py#L19
2. Because it does not affect semantics
3. It is a more modern way of doing things, probably `IOError` is a Python2-era artifact in typeshed


AlexWaygood (approved):
Thanks!

PR #9685, opened 2023-02-07
Remove a no-longer-needed `# noqa`

AlexWaygood:
Some PEP 604 bugs were fixed in mypy 1.0, unblocking these changes


AlexWaygood:
...or not

PR #9688, opened 2023-02-07
Stdlib: add 'obvious' default values

AlexWaygood:
All parameters annotated with `Literal[False]` must, logically speaking, have a default of `False` (if they have a default value at all). The same goes for parameters annotated with `None`, `Literal[True]`, `Literal[0]`, `Literal["foo"]`, etc. etc.

I ran this code in the interactive REPL to create this PR:

```pycon
>>> from pathlib import Path
>>> import re
>>> for path in Path("stdlib").rglob("*.pyi"):
...     source = path.read_text()
...     new_source = re.sub(r": Literal\[([a-zA-Z0-9_\"]+)\] = \.\.\.", r": Literal[\1] = \1", source)
...     new_source = re.sub(r": None = \.\.\.", ": None = None", new_source)
...     if new_source != source:
...         path.write_text(new_source)
```


srittau (approved):
Didn't check against the implementation, but the reasoning (and spot checking) is sound.

AlexWaygood:
> I ran this code in the interactive REPL to create this PR:
> 
> ```
> >>> from pathlib import Path
> >>> import re
> >>> for path in Path("stdlib").rglob("*.pyi"):
> ...     source = path.read_text()
> ...     new_source = re.sub(r": Literal\[([a-zA-Z0-9_\"]+)\] = \.\.\.", r": Literal[\1] = \1", source)
> ...     new_source = re.sub(r": None = \.\.\.", ": None = None", new_source)
> ...     if new_source != source:
> ...         path.write_text(new_source)
> ```

@JelleZijlstra we could possibly add this kind of logic to stubdefaulter, but I'm not sure if it would be a "good fit" with the rest of the code. WDYT?

JelleZijlstra:
I'd be open to adding it.

AlexWaygood:
> I'd be open to adding it.

https://github.com/JelleZijlstra/stubdefaulter/pull/54 :)

Issue #9690, opened 2023-02-07
Mypy errors after flake8-pyi Y037 changes: Type application has too many types (1 expected)


intgr:
types-requests version 2.28.11.11 and types-urllib3 version 1.26.25.5 somehow cause issues with lots of djangotestframework-stubs CI tests. (https://github.com/typeddjango/djangorestframework-stubs/actions/runs/4114959794/jobs/7103023727)

According to [changelog](https://github.com/typeshed-internal/stub_uploader/blob/main/data/changelogs/requests.md) #9686 was the only change in that version.

Errors like this appear in every test:

> requests-stubs/sessions.pyi:80: error: Type application has too many types (1 expected) (diff)
> requests-stubs/sessions.pyi:81: error: Type application has too many types (1 expected) (diff)
> requests-stubs/sessions.pyi:99: error: Type application has too many types (1 expected) (diff)
> requests-stubs/sessions.pyi:106: error: Type application has too many types (1 expected) (diff)
> urllib3-stubs/fields.pyi:6: error: Type application has too many types (1 expected) (diff)

AlexWaygood:
Same question as https://github.com/python/typeshed/pull/9686#issuecomment-1420906330: do your tests run with mypy 1.0? If not, then you may need to upgrade to mypy 1.0 (or temporarily pin to a lower version of `types-requests`).

That error message comes from https://github.com/python/mypy/issues/11098, but that bug should have been fixed in mypy 1.0.

AlexWaygood:
We should probably document better that our stubs can generally only be guaranteed to be compatible with the latest version of `$TYPECHECKER`. We've previously discussed that in #6843.

intgr:
> do your tests run with mypy 1.0?

[What does it look like I'm doing](https://www.youtube.com/watch?v=_UZFI-8D5uA) 😆 

No, there's a PR https://github.com/typeddjango/djangorestframework-stubs/pull/339 that at first I thought had issues due to changes in another upstream project. Whilst fixing those in https://github.com/typeddjango/djangorestframework-stubs/pull/345 I ran into this issue.

But seriously, mypy 1.0 hasn't even been out for 24 **_HOURS_** yet! Is it really considered reasonable in this project to break users so quickly following a release?

AlexWaygood:
> But seriously, mypy 1.0 hasn't even been out for 24 **_HOURS_** yet! Is it really considered reasonable in this project to break users so quickly following a release?

[...]

PR #9694, opened 2023-02-08
Use `typing_extensions.Self` in the stdlib

AlexWaygood:
Only doing the stdlib for now, so that we don't have a repeat of #9690.

The first commit in this PR was prepared using the following script, which uses `libcst`:

<details>

```python
from copy import deepcopy
from pathlib import Path

import libcst


class ModuleFixer(libcst.CSTTransformer):
    def leave_ImportFrom(self, original_node: libcst.ImportFrom, updated_node: libcst.ImportFrom) -> libcst.ImportFrom:
        if not (isinstance(original_node.module, libcst.Name) and original_node.module.value == "_typeshed"):
            return updated_node
        if isinstance(original_node.names, libcst.ImportStar):
            return updated_node
        new_imports = [
            item
            for item in original_node.names
            if not (isinstance(item, libcst.ImportAlias) and isinstance(item.name, libcst.Name) and item.name.value == "Self")
        ]
        if len(original_node.names) == 2 and len(new_imports) == 1:
            new_imports[0] = new_imports[0].with_changes(comma=libcst.MaybeSentinel.DEFAULT)
        return updated_node.with_changes(names=new_imports)

    def leave_Parameters(self, original_node: libcst.Parameters, updated_node: libcst.Parameters) -> libcst.Parameters:
        if not original_node.params:
            return updated_node

        first_param = original_node.params[0]
        if not (
            isinstance(first_param, libcst.Param)
            and isinstance(first_param.annotation, libcst.Annotation)
        ):
            return updated_node

        first_annotation = first_param.annotation.annotation
        if isinstance(first_annotation, libcst.Name):
            if first_annotation.value != "Self":
                return updated_node
        elif isinstance(first_annotation, libcst.Subscript):
            if not (
                isinstance(first_annotation.value, libcst.Name)
                and first_annotation.value.value == "type"
                and len(first_annotation.slice) == 1
                and isinstance(first_annotation.slice[0], libcst.SubscriptElement)
                and isinstance(first_annotation.slice[0].slice, libcst.Index)
                and isinstance(first_annotation.slice[0].slice.value, libcst.Name)
                and first_annotation.slice[0].slice.value.value == "Self"
            ):
                return updated_node
        else:
            return updated_node

        new_params = list(updated_node.params)
        new_params[0] = first_param.with_changes(annotation=None)
        return updated_node.with_changes(params=new_params)


def change_self_annotations(path: Path) -> None:
    source = path.read_text()
    if not source:
        return
    if "from _typeshed import" not in source:
        return
    if not (": Self" in source or ": type[Self]" in source):
        return
    sourcelines = source.splitlines()
    try:
        sourcelines.remove("from _typeshed import Self")
    except ValueError:
        pass
    first_non_comment_lineno = next(i for i, line in enumerate(sourcelines) if line.strip() and not line.strip().startswith("#"))
    sourcelines.insert(first_non_comment_lineno, "from typing_extensions import Self")
    cst = libcst.parse_module("\n".join(sourcelines) + "\n")
    new_source = cst.visit(ModuleFixer()).code
    path.write_text(new_source)


def main() -> None:
    for path in Path("stdlib").rglob("*.pyi"):
        print(f"Fixing '{path}'...")
        change_self_annotations(path)


if __name__ == "__main__":
    main()
```

</details>


AlexWaygood:
> ```diff
> mongo-python-driver (https://github.com/mongodb/mongo-python-driver)
> + bson/son.py: note: In member "__new__" of class "SON":
> + bson/son.py:69: error: Value of type variable "Self" of "__new__" of "dict" cannot be "SON[_Key, _Value]"  [type-var]
> ```

Not sure what's going on here, but it looks like a minor mypy bug in its handling of `typing_extensions.Self`. The other two mypy_primer hits are just noise.

srittau (approved):
The primer hit is probably due to `dict.__new__` now being incompatible with the bson type annotation. I'm fine with that.

AlexWaygood:
> The primer hit is probably due to `dict.__new__` now being incompatible with the bson type annotation. I'm fine with that.

It's weird that we`dict.__new__` has a different signature to `dict.__init__` in our stubs. Maybe we should look into whether we actually need that.

stdlib/builtins.pyi
```diff
@@ -196,9 +198,9 @@ _LiteralInteger = _PositiveInteger | _NegativeInteger | Literal[0]  # noqa: Y026
 
 class int:
     @overload
-    def __new__(cls: type[Self], __x: str | ReadableBuffer | SupportsInt | SupportsIndex | SupportsTrunc = ...) -> Self: ...
+    def __new__(cls, __x: str | ReadableBuffer | SupportsInt | SupportsIndex | SupportsTrunc = ...) -> Self: ...
```
Avasam, line 201: I noticed this change has introduced a new pyright issue at  https://github.com/python/typeshed/blob/main/scripts/stubsabot.py#L37

AlexWaygood: My guess is that's probably a pyright issue in how it handles `Self`. What's the error message pyright produces?

Avasam: It thinks `member` is a `int` when previously it saw it as a `ActionLevelSelf`
image
image

vs before

image

I guess pyright doesn't take into account that the first classmethod parameter (`cls`) could be something else than it's own class.
mypy does it correctly:
image

AlexWaygood: Hmm, yeah that seems like it might be a pyright bug (though Eric will be the judge of that). If you file an issue over there, he'll probably either fix it or triage it as "wontfix" within 24 hours :D

We also don't need the `ActionLevelSelf` `TypeVar` in stubsabot anymore, though, we can just use `typing_extensions.Self` there.

AlexWaygood: > We also don't need the `ActionLevelSelf` `TypeVar` in stubsabot anymore, though, we can just use `typing_extensions.Self` there.

- #9785

Avasam: Alright, created an issue here: https://github.com/microsoft/pyright/issues/4668

ShaneHarvey:
Coming here because I'm attempting to fix or # type: ignore the SON error (in https://github.com/mongodb/mongo-python-driver/pull/1194).

> It's weird that dict.__new__ has a different signature to dict.__init__ in our stubs. Maybe we should look into whether we actually need that.

Has an issue been created for this yet? It's not clear to me whether this is a bug in pymongo or a bug in mypy.

AlexWaygood:
> Coming here because I'm attempting to fix or # type: ignore the SON error (in [mongodb/mongo-python-driver#1194](https://github.com/mongodb/mongo-python-driver/pull/1194)).
> 
> > It's weird that dict.**new** has a different signature to dict.**init** in our stubs. Maybe we should look into whether we actually need that.
> 
> Has an issue been created for this yet? It's not clear to me whether this is a bug in pymongo or a bug in mypy.

It looks like a mypy bug to me. The `dict.__new__`/`dict.__init__` discrepancy in our stubs is _strange_, but I don't _think_ it's the cause of this error. I'd suggest opening an issue over at mypy.

ShaneHarvey:
Thanks I opened: https://github.com/python/mypy/issues/15135

PR #9696, opened 2023-02-09
Update `__exit__` parameters in stubs

Avasam:
Follow-up to #9519 for third-party stubs now that a new version of mypy has been released
Ref #9297

I could not validate `psycopg2`, `hdbcli` and `_cffi_backend` so I left those as-is


srittau:
I'm not sure that replacing proper `__exit__` types with (essentially) `object` is the right course of action. I can't imagine a situation where you would want to pass some arbitrary into `__exit__`. In fact I'd say that's a sign of potential bug (even if `__exit__` itself accepts it.) It also makes sub-classing more difficult, as an overridden `__exit__` would also need to accept `object`, instead of the proper types.

srittau:
I'd say *ideally* all `__exit__` signatures would use the following signature (with the exception of the return type), even those that just use `*args` in the implementation:

```python
_E = TypeVar("_E", bound=BaseException)

class X:
    @overload
    def __exit__(self, __type: None, __value: None, __exc: None) -> None: ...
    @overload
    def __exit__(self, __type: type[_E], __value: type[_E], __exc: TracebackType) -> None: ...
```

But since that is unwieldy, we can use the best approximation in a current situation. So I'm actually fine to use `Unused` for new annotations if that's the fastest way to do it, but we shouldn't regress back from the ideal situation. But I'd be interested in what the other maintainers think.

Avasam:
> [...] So I'm actually fine to use `Unused` for new annotations if that's the fastest way to do it, but we shouldn't regress back from the ideal situation.

Ok, I think my last commit satisfies your comment. (now the PR only adds new `Unused` when it's unused, new `object` for `*args` params and changes `*args: object` to `*args: Unused` when unused.

> I'd say ideally all `__exit__` signatures would use the following signature (with the exception of the return type), even those that just use `*args` in the implementation: [...]
> But since that is unwieldy, we can use the best approximation in a current situation.

This isn't a "stopgap", "urgent" or "in the meantime" PR. It's just splitting off a would-be bigger PR about semantically using `Unused` instead of `object` or `_Unused` for unused parameters.
Since it's focusing entirely on exit dunder methods. I don't mind changing it to do it "the proper way" everywhere (and maybe documenting in CONTRIBUTING.md any decision about typing exit dunder methods)

srittau:
Side note: Changing `__exit__` to overloads could be disruptive or could even now work. I'd definitely would test it with a small explorative PR first.

JelleZijlstra:
This PR has a lot of conflicts now.

I'd be hesitant to use the overloaded signature @srittau suggests, though it is probably technically correct:
- There may be rare instances where only some of the arguments are None, though recent changes to CPython make this less and less likely.
- Using this signature will force users writing subclasses to also write the verbose signature with overloads, which feels like unnecessary busywork.

AlexWaygood:
My preference would be to _generally_ type `__(a)exit__` methods like this, if the method uses `*args` at runtime:

```python
def __exit__(self, *args: Unused) -> None: ...
```

And to _generally_ type `__(a)exit__` methods like this, if the method includes all three parameters explicitly at runtime:

```python
def __exit__(self, exc_typ: type[BaseException] | None, exc_val: BaseException | None, exc_tb: TracebackType | None
```

(There will obviously be exceptions to these rules where doing something else makes sense.)

---

I agree with @srittau that using `object` or `Unused` for annotations with the three-argument form could make subclassing unnecessarily difficult, as subclasses will be obliged to use `object` for annotating those parameters in the subclass.

I also agree with @srittau that a truly precise signature for `__(a)exit__` methods would require overloads, but that this would be pretty tedious to do everywhere, and probably wouldn't _actually_ improve type safety significantly. And I agree with @JelleZijlstra that using overloads could also make it harder for people to subclass these classes, so has downsides as well.

Avasam:
> My preference would be to _generally_ type `__(a)exit__` methods like this, if the method uses `*args` at runtime:
>
> And to _generally_ type `__(a)exit__` methods like this, if the method includes all three parameters explicitly at runtime:
> 
> ```python
> def __exit__(self, exc_typ: type[BaseException] | None, exc_val: BaseException | None, exc_tb: TracebackType | None
> ```

I know you said there could be exceptions. But that sounds like something Flake8-PYI could check for and/or add in its existing `__(a)exit__` checks.

AlexWaygood:
> I know you said there could be exceptions. But that sounds like something Flake8-PYI could check for and/or add in its existing `__(a)exit__` checks, if the exceptions are rare enough.

Y036 actually used to be stricter, but Jelle persuaded me to relax it slightly in https://github.com/PyCQA/flake8-pyi/pull/209 after we came across a situation where there was a good case for doing something different. Maybe the exceptions that break the rule are rare enough that we could consider reverting that change — idk, I don't really have a strong opinion on exactly how strict flake8-pyi should be on this `¯\_(ツ)_/¯`

stubs/caldav/caldav/davclient.pyi
```diff
@@ -50,7 +51,9 @@ class DAVClient:
         ssl_cert: str | tuple[str, str] | None = ...,
     ) -> None: ...
     def __enter__(self) -> Self: ...
-    def __exit__(self, exc_type: object, exc_value: object, traceback: object) -> None: ...
+    def __exit__(
+        self, exc_type: type[BaseException] | None, exc_value: BaseException | None, traceback: TracebackType | None
```
Avasam, line 55: This used to be typed with objects. Please validate it still makes sense.

stubs/influxdb-client/influxdb_client/client/influxdb_client.pyi
```diff
@@ -43,7 +44,9 @@ class InfluxDBClient(_BaseClient):
         profilers: Incomplete | None = ...,
     ) -> None: ...
     def __enter__(self) -> Self: ...
-    def __exit__(self, exc_type: object, exc_value: object, traceback: object) -> None: ...
+    def __exit__(
+        self, exc_type: type[BaseException] | None, exc_value: BaseException | None, traceback: TracebackType | None
```
Avasam, line 48: This used to be typed with objects. Please validate it still makes sense.

stubs/influxdb-client/influxdb_client/client/influxdb_client_async.pyi
```diff
@@ -37,7 +38,9 @@ class InfluxDBClientAsync(_BaseClient):
         profilers: Incomplete | None = ...,
     ) -> None: ...
     async def __aenter__(self) -> Self: ...
-    async def __aexit__(self, exc_type: object, exc: object, tb: object) -> None: ...
+    async def __aexit__(
+        self, exc_type: type[BaseException] | None, exc: BaseException | None, tb: TracebackType | None
```
Avasam, line 42: This used to be typed with objects. Please validate it still makes sense.

stubs/pika/pika/adapters/blocking_connection.pyi
```diff
@@ -57,7 +58,9 @@ class BlockingConnection:
         self, parameters: Parameters | Sequence[Parameters] | None = ..., _impl_class: Incomplete | None = ...
     ) -> None: ...
     def __enter__(self) -> Self: ...
-    def __exit__(self, exc_type: object, value: object, traceback: object) -> None: ...
+    def __exit__(
+        self, exc_type: type[BaseException] | None, value: BaseException | None, traceback: TracebackType | None
```
Avasam, line 62: This used to be typed with objects. Please validate it still makes sense.

stubs/psycopg2/psycopg2/_psycopg.pyi
```diff
@@ -462,7 +462,7 @@ class connection:
     def tpc_rollback(self, __xid: str | bytes | Xid = ...) -> None: ...
     def xid(self, format_id, gtrid, bqual) -> Xid: ...
     def __enter__(self) -> Self: ...
-    def __exit__(self, __type: object, __name: object, __tb: object) -> None: ...
+    def __exit__(self, __type: type[BaseException] | None, __name: BaseException | None, __tb: TracebackType | None) -> None: ...
```
Avasam, line 465: This used to be typed with objects. Please validate it still makes sense.

stubs/pyserial/serial/threaded/__init__.pyi
```diff
@@ -42,4 +43,6 @@ class ReaderThread(threading.Thread):
     def close(self) -> None: ...
     def connect(self) -> tuple[Self, Protocol]: ...
     def __enter__(self) -> Protocol: ...
-    def __exit__(self, __exc_type: object, __exc_val: object, __exc_tb: object) -> None: ...
+    def __exit__(
+        self, __exc_type: type[BaseException] | None, __exc_val: BaseException | None, __exc_tb: TracebackType | None
```
Avasam, line 47: This used to be typed with objects. Please validate it still makes sense.

stubs/redis/redis/asyncio/cluster.pyi
```diff
@@ -76,7 +77,9 @@ class RedisCluster(AbstractRedis, AbstractRedisCluster, Generic[_StrType]):  # T
     async def initialize(self) -> Self: ...
     async def close(self) -> None: ...
     async def __aenter__(self) -> Self: ...
-    async def __aexit__(self, exc_type: object, exc_value: object, traceback: object) -> None: ...
+    async def __aexit__(
+        self, exc_type: type[BaseException] | None, exc_value: BaseException | None, traceback: TracebackType | None
```
Avasam, line 81: This used to be typed with objects. Please validate it still makes sense.

stubs/tqdm/tqdm/std.pyi
```diff
@@ -200,7 +201,9 @@ class tqdm(Generic[_T], Iterable[_T], Comparable):
     def __reversed__(self) -> Iterator[_T]: ...
     def __contains__(self, item: object) -> bool: ...
     def __enter__(self) -> Self: ...
-    def __exit__(self, exc_type: object, exc_value: object, traceback: object) -> None: ...
+    def __exit__(
+        self, exc_type: type[BaseException] | None, exc_value: BaseException | None, traceback: TracebackType | None
```
Avasam, line 205: This used to be typed with objects. Please validate it still makes sense.

stubs/redis/redis/asyncio/utils.pyi
```diff
@@ -9,4 +10,6 @@ class pipeline(Generic[_StrType]):
     p: Pipeline[_StrType]
     def __init__(self, redis_obj: Redis[_StrType]) -> None: ...
     async def __aenter__(self) -> Pipeline[_StrType]: ...
-    async def __aexit__(self, exc_type: object, exc_value: object, traceback: object) -> None: ...
+    async def __aexit__(
+        self, exc_type: type[BaseException] | None, exc_value: BaseException | None, traceback: TracebackType | None
```
Avasam, line 14: This used to be typed with objects. Please validate it still makes sense.

stubs/redis/redis/asyncio/cluster.pyi
```diff
@@ -146,10 +149,14 @@ class ClusterPipeline(AbstractRedis, AbstractRedisCluster, Generic[_StrType]):
     def __init__(self, client: RedisCluster[_StrType]) -> None: ...
     async def initialize(self) -> Self: ...
     async def __aenter__(self) -> Self: ...
-    async def __aexit__(self, exc_type: object, exc_value: object, traceback: object) -> None: ...
+    async def __aexit__(
+        self, exc_type: type[BaseException] | None, exc_value: BaseException | None, traceback: TracebackType | None
```
Avasam, line 153: This used to be typed with objects. Please validate it still makes sense.

stubs/redis/redis/asyncio/cluster.pyi
```diff
@@ -146,10 +149,14 @@ class ClusterPipeline(AbstractRedis, AbstractRedisCluster, Generic[_StrType]):
...
+    ) -> None: ...
     def __await__(self) -> Awaitable[Self]: ...
     def __enter__(self) -> Self: ...
-    def __exit__(self, exc_type: object, exc_value: object, traceback: object) -> None: ...
+    def __exit__(
+        self, exc_type: type[BaseException] | None, exc_value: BaseException | None, traceback: TracebackType | None
```
Avasam, line 158: This used to be typed with objects. Please validate it still makes sense.

Avasam:
> [...] I don't really have a strong opinion on exactly how strict flake8-pyi should be on this `¯\_(ツ)_/¯`

Me neither, just throwing suggestions :)

Just in case, I've flagged all the 3 parameters definitions that used to all be objects.

stubs/humanfriendly/humanfriendly/testing.pyi
```diff
@@ -65,7 +65,7 @@ class MockedProgram(CustomSearchPath):
     program_signal_file: Any
     def __init__(self, name, returncode: int = ..., script: Incomplete | None = ...) -> None: ...
     def __enter__(self): ...
-    def __exit__(self, *args, **kw): ...
+    def __exit__(self, *args: object, **kw): ...
```
AlexWaygood, line 68:
```suggestion
    def __exit__(self, *args: object, **kw: object): ...
```

AlexWaygood (commented):
All the places you flagged look fine -- just one suggestion

AlexWaygood (approved):
Thanks!

PR #9699, opened 2023-02-09
[fpdf2] FPDF.add_font() accepts PurePaths

srittau:


stubs/fpdf2/fpdf/fpdf.pyi
```diff
@@ -295,7 +296,11 @@ class FPDF:
         style: RenderStyle | str | None = ...,
     ) -> None: ...
     def add_font(
-        self, family: str | None = ..., style: _FontStyle = ..., fname: str | None = ..., uni: bool | Literal["DEPRECATED"] = ...
+        self,
+        family: str | None = None,
+        style: _FontStyle = "",
+        fname: str | PurePath | None = None,
```
JelleZijlstra, line 302: Looks like it technically accepts "PathLike objects with a reasonable str()", but `str | PurePath` is a good enough approximation.

PR #9702, opened 2023-02-09
Use `typing_extensions.Self` in the `stubs/` directory

AlexWaygood:
Note that this will mean that the changed third-party packages are no longer compatible with mypy <1.0. However, we've already broken compatibility with mypy <1.0 in several third-party stubs in #9689 and #9695 `¯\_(ツ)_/¯`.

I used the same script as in #9694 to create the first commit in this PR.


srittau:
> Note that this will mean that the changed third-party packages are no longer compatible with mypy <1.0.

We probably need to have a conversation how we can better signal type checker compatibility to our users, but for now I'm fine with breaking compatibility when a new type checker version comes out. Automatically upgrading type stub packages is dangerous in any case.

erictraut:
@AlexWaygood, this looks like a bug in pyright. I'll try to get it fixed by the end of the day. If this is blocking you, I could do a hot fix release, or we could wait until next Tuesday for the normal weekly release.

AlexWaygood:
> If this is blocking you, I could do a hot fix release, or we could wait until next Tuesday for the normal weekly release.

Don't worry, I can easily revert the changes on the affected files for now. No need for a hotfix release, I don't think.

erictraut:
I've fixed the pyright bug. It will be included in the next weekly release (1.1.294).

AlexWaygood:
> I've fixed the pyright bug. It will be included in the next weekly release (1.1.294).

Thank you very much!

AlexWaygood:
7fc50c6 of this PR can be reverted now that we've updated to the latest pyright in CI, but I don't have access to a computer right now and it doesn't seem possible to revert a commit on your phone

srittau:
I've reverted https://github.com/python/typeshed/commit/7fc50c6cf2fd2226f0d8c6b957d6617195eb76ba

srittau (approved):
Spot checking looks good.

AlexWaygood:
> I've reverted [7fc50c6](https://github.com/python/typeshed/commit/7fc50c6cf2fd2226f0d8c6b957d6617195eb76ba)

Thank you very much!

tsibley:
@srittau [wrote](https://github.com/python/typeshed/pull/9702#issuecomment-1424536113):

> We probably need to have a conversation how we can better signal type checker compatibility to our users, but for now I'm fine with breaking compatibility when a new type checker version comes out. Automatically upgrading type stub packages is dangerous in any case.

End user here. Better signaling would be nice. It's too bad there's no "Conflicts" field for Python package metadata, since it would be ~perfect for the third-party typeshed stub packages to declare which versions of type checkers they aren't compatible with.

PR #9704, opened 2023-02-10
Update `Unused` parameters in `stubs/`

Avasam:
Follow-up to #9519 for third-party stubs now that a new version of mypy has been released
Ref #9297
Avoids touching `__a?exit__`. That's in #9696

Three steps:
1. Use `Unused` instead of `_Unused`
2. Update some stray `object | None` to `object` (unless there's a helpful reason for it, in which case a comment was added)
3. Change unused parameters typed as `object` to be typed with `Unused`


stubs/pyasn1/pyasn1/type/base.pyi
```diff
@@ -41,7 +41,7 @@ class NoValue:
     def __getattr__(self, attr) -> None: ...
     # def __new__.<locals>.getPlug.<locals>.plug
     @type_check_only
-    def plug(self, *args: object, **kw: object): ...
+    def plug(self, *args: Unused, **kw: Unused) -> NoReturn: ...
```
JelleZijlstra, line 44: I'm not sure I understand why this method is in the stub at all.

JelleZijlstra (approved):

stubs/SQLAlchemy/sqlalchemy/util/_collections.pyi
```diff
@@ -16,9 +16,9 @@ collections_abc = collections.abc
 EMPTY_SET: frozenset[Any]
 
 class ImmutableContainer:
-    def __delitem__(self, *arg: object, **kw: object) -> NoReturn: ...
-    def __setitem__(self, *arg: object, **kw: object) -> NoReturn: ...
-    def __setattr__(self, *arg: object, **kw: object) -> NoReturn: ...
+    def __delitem__(self, *arg: Unused, **kw: Unused) -> NoReturn: ...
+    def __setitem__(self, *arg: Unused, **kw: Unused) -> NoReturn: ...
+    def __setattr__(self, *arg: Unused, **kw: Unused) -> NoReturn: ...
```
AlexWaygood, line 21: (I haven't looked at the source code here, just speculating.)

If instances of `ImmutableContainer` are meant to be, well, immutable, it might produce better type checking if we just omitted these methods from the stub altogether.

stubs/pyasn1/pyasn1/type/base.pyi
Avasam: Pretty sure it's just to use as an alias below. Source: https://github.com/pyasn1/pyasn1/blob/main/pyasn1/type/base.py#L212

stubs/SQLAlchemy/sqlalchemy/util/_collections.pyi
Avasam: Maybe relevant to your question:
https://github.com/microsoft/pyright/issues/4653
https://github.com/python/mypy/issues/14726

PR #9706, opened 2023-02-10
Make datetime classmethods return correct type

### LINKED ISSUE #9578: Python's datetime.datetime.strptime has wrong return type
According to the stubs, `datetime.datetime.strptime` returns a `datetime`. However, it actually returns the type of the class. So if I subclass `datetime` and call the subclass's `strptime` method, I get back an instance of the subclass and not a `datetime` object, violating the return type constraint.

PythonCoderAS:
Fix #9578

Confirmed the methods changed do not have the weird behavior as seen in versions prior to Python 3.8 via testing.


AlexWaygood (approved):
> Confirmed the methods changed do not have the weird behavior as seen in versions prior to Python 3.8 via testing.

The change is correct on Python 3.7 also (typeshed still supports 3.7):

```pycon
(py37) C:\Users\alexw\coding>python
Python 3.7.16 (default, Jan 17 2023, 16:06:28) [MSC v.1916 64 bit (AMD64)] :: Anaconda, Inc. on win32
>>> import datetime
>>> class Foo(datetime.datetime): ...
...
>>> Foo.combine(datetime.date(1999, 1, 1), datetime.time())
Foo(1999, 1, 1, 0, 0)
>>> type(_)
<class '__main__.Foo'>
>>> Foo.strptime("1999-1-1", "%Y-%m-%d")
Foo(1999, 1, 1, 0, 0)
>>> type(_)
<class '__main__.Foo'>
```

PR #9707, opened 2023-02-10
Tensorflow keras layer

hmc-cs-mdrissi:
# Summary
Next tensorflow stub pr. Primary goal of this one was to define a reasonable initial stub for [Layers](https://www.tensorflow.org/api_docs/python/tf/keras/layers/Layer), one of the most common types user interacts with. The other files are mainly types needed to cover Layer's arguments. I've included a couple layer's too. Adding a stub for 1 layer is fairly easy as most of the time only thing needed is constructor and all public methods come from parent Layer class.

I ended up adding a fair amount of stubtest allowlist and tried to give an explanation for each. `call/__call__` strangeness is probably biggest one and I can explain more if it's unclear. At it's core the magic is as a user, you define

`def call(self, inputs): ...`

but you are never supposed to call that method and instead use `__call__` which does a good amount of signature introspection/argument adjustment for you.


stubs/tensorflow/@tests/stubtest_allowlist.txt
```diff
@@ -15,8 +15,49 @@ tensorflow.DType.__getattr__
...
+# import tensorflow as tf
+# tf.initializers
+# Usually these two ways are same module, but for tensorflow the first way
+# often does not work and the second way does. The documentation describes
+# tf.initializers as module and has that type if accessed the second way,
+# but the real module file is completely different name (even package) and dynamically handled.
+# tf.initializers at runtime is <module 'keras.api._v2.keras.initializers' from '...'>
+tensorflow.initializers
+
+# Layer constructor's always have **kwargs, but only allow a few specific values. PEP 692
```
hmc-cs-mdrissi, line 38: I don't think we have PEP 692 tracker ticket yet. Will make one later if no one covers it first.

stubs/tensorflow/@tests/stubtest_allowlist.txt
```diff
@@ -15,8 +15,49 @@ tensorflow.DType.__getattr__
...
+# Tensorflow imports are cursed.
+# import tensorflow.initializers
+# import tensorflow as tf
+# tf.initializers
+# Usually these two ways are same module, but for tensorflow the first way
+# often does not work and the second way does. The documentation describes
+# tf.initializers as module and has that type if accessed the second way,
+# but the real module file is completely different name (even package) and dynamically handled.
+# tf.initializers at runtime is <module 'keras.api._v2.keras.initializers' from '...'>
+tensorflow.initializers
```
hmc-cs-mdrissi, line 36: This is probably most cursed weirdness of imports (pylint is unhappy too with this). A lot of other tf modules do similar thing and this part of allowlist will grow over time.

stubs/tensorflow/tensorflow/keras/layers.pyi
```diff
@@ -0,0 +1,198 @@
...
+# In actual implementation there's 12 allowed keyword arguments, but only
+# 4 are documented and other 8 are mainly internal. The other 8 can be found
+# https://github.com/keras-team/keras/blob/e6784e4302c7b8cd116b74a784f4b78d60e83c26/keras/engine/base_layer.py#L329
+# PEP 692 support would be very helpful here and allow removing stubtest allowlist for
+# all layer constructors.
+
+# TODO: Replace last Any after adding tf.keras.mixed_precision.Policy.
+_LayerDtype = tf._DTypeLike | dict[str, Any] | Any
+
+# Layer's compute_output_shape commonly have instance as first argument name instead of self.
```
hmc-cs-mdrissi, line 110: I'm unsure it's worthwhile to actually have `def compute_output_shape(instance, input_shape: tf.TensorShape) -> tf.TensorShape: ...` for layers where stubtest complains about instance vs self. Feels very implementation detail and maybe it be better to just allowlist all build/compute_output_shape for layer's.

stubs/tensorflow/@tests/stubtest_allowlist.txt
JelleZijlstra: I think PEP 692 might already have sufficient support to use it; at least mypy and pyright support it. Haven't checked pytype though.

hmc-cs-mdrissi: Testing basic 692 support:

```python
class Foo(TypedDict, total=False):
    a: int
    b: str


def f(**kwargs: Unpack[Foo]) -> None:
    ...


f(a=1, b="2")  # OK
f(a=1, b=2)  # Error: b has type str
f(a=1, b="2", c=3)  # Error: unexpected keyword argument "c"
```

- pyright works fine
- mypy works if we add --enable-incomplete-feature=Unpack (otherwise it gives `error: "Unpack" support is experimental, use --enable-incomplete-feature=Unpack to enable`)
- pytype produces `typing_extensions.Unpack not supported yet [not-supported-yet]`. No errors otherwise.
- pyre is same as pytype and just produces `pep_692.py:11:16 Undefined or invalid type [11]: Annotation `Unpack` is not defined as a type.`

I'll open ticket to track.

hmc-cs-mdrissi:
Hmm optuna hit is valid, but input_dim argument is undocumented at least on layer [documentation](https://www.tensorflow.org/api_docs/python/tf/keras/layers/Layer#args_1). input_dim is one of the internal arguments [here](https://github.com/keras-team/keras/blob/e6784e4302c7b8cd116b74a784f4b78d60e83c26/keras/engine/base_layer.py#L329). 

Same issue for streamlit error. input_shape argument is also not documented anywhere I can find. There's 6 more internal layer arguments that could also be added.

I weakly prefer not to include arguments that do not appear in documentation, but I'm fine adding them in if desired.

hmc-cs-mdrissi:
General review poke. There are a few small questions I’ve left but otherwise been ready for over week.

stubs/tensorflow/tensorflow/_aliases.pyi
```diff
@@ -0,0 +1,14 @@
...
+# equivalent.
+
+from collections.abc import Mapping, Sequence
+from typing import Any, TypeVar
+from typing_extensions import TypeAlias
+
+import numpy
+
+_T1 = TypeVar("_T1")
+_ContainerGeneric: TypeAlias = Mapping[str, _ContainerGeneric[_T1]] | Sequence[_ContainerGeneric[_T1]] | _T1
```
JelleZijlstra, line 12: These don't need to be underscore-prefixed since the module itself is private.

stubs/tensorflow/tensorflow/keras/constraints.pyi
```diff
@@ -0,0 +1,22 @@
...
+_Constraint: TypeAlias = str | dict[str, Any] | Constraint | None
+
+@overload
+def get(identifier: None) -> None: ...
+@overload
+def get(identifier: Callable[[Tensor], Tensor]) -> Callable[[Tensor], Tensor]: ...
+@overload
+def get(identifier: str | dict[str, Any]) -> Constraint: ...
+@overload
+def get(identifier: _Constraint) -> Constraint | Callable[[Tensor], Tensor] | None: ...
```
JelleZijlstra, line 21: Using the type alias seems unnecessary here as all types but `Constraint` are already covered by previous overloads.

stubs/tensorflow/tensorflow/keras/initializers.pyi
```diff
@@ -0,0 +1,48 @@
...
+zeros = Zeros
+
+_Initializer: TypeAlias = str | Initializer | Callable[[_ShapeLike], Tensor] | dict[str, Any] | None
+
+@overload
+def get(identifier: None) -> None: ...
+@overload
+def get(identifier: str | Initializer | dict[str, Any]) -> Initializer: ...
+@overload
+def get(identifier: _Initializer) -> Initializer | None: ...
```
JelleZijlstra, line 47: Same as before, seems like this is only the `Callable`.

stubs/tensorflow/tensorflow/keras/regularizers.pyi
```diff
@@ -0,0 +1,23 @@
...
+_Regularizer: TypeAlias = str | dict[str, Any] | Regularizer | None
+
+@overload
+def get(identifier: None) -> None: ...
+@overload
+def get(identifier: Callable[[Tensor], Tensor]) -> Callable[[Tensor], Tensor]: ...
+@overload
+def get(identifier: str | dict[str, Any]) -> Regularizer: ...
+@overload
+def get(identifier: _Regularizer) -> Regularizer | Callable[[Tensor], Tensor] | None: ...
```
JelleZijlstra, line 22: Another similar issue.

stubs/tensorflow/tensorflow/keras/constraints.pyi
hmc-cs-mdrissi: Fixed and made overloads non overlapping.

One question for this and similar comments is it recommended to make these generic? As,

```python
from tensorflow.initializers import Zeros

get(Zeros()) # returns same type Zeros().
```

I feel neutral as most common overload I think is str/dict[str, Any] case but otherwise constraint case always returns same type.

stubs/tensorflow/tensorflow/keras/initializers.pyi
hmc-cs-mdrissi: Fixed. For this one re-reading [source](https://github.com/keras-team/keras/blob/v2.11.0/keras/initializers/__init__.py#L153-L201) I missed one case (type[Initializer] is also allowed).

stubs/tensorflow/tensorflow/keras/constraints.pyi
```diff
@@ -0,0 +1,20 @@
...
+from typing import Any, overload
+from typing_extensions import TypeAlias
+
+from tensorflow import Tensor
+
+class Constraint:
+    def get_config(self) -> dict[str, Any]: ...
+    def __call__(self, __w: Tensor) -> Tensor: ...
+
+_Constraint: TypeAlias = str | dict[str, Any] | Constraint | None  # noqa: Y047
```
JelleZijlstra, line 12: Since this is only used in `initializers.pyi`, maybe just define it there. Otherwise, add a comment explaining why we're ignoring the linter.

stubs/tensorflow/tensorflow/keras/layers.pyi
JelleZijlstra: I would go with the simplest solution that passes our CI while maintaining safety. Consider marking the parameter as positional-only, for example.

JelleZijlstra (commented):
Sorry for the wait! I left a few comments.

stubs/tensorflow/tensorflow/_aliases.pyi
```diff
@@ -0,0 +1,14 @@
...
+# equivalent.
+
+from collections.abc import Mapping, Sequence
+from typing import Any, TypeVar
+from typing_extensions import TypeAlias
+
+import numpy
+
+_T1 = TypeVar("_T1")
+ContainerGeneric: TypeAlias = Mapping[str, ContainerGeneric[_T1]] | Sequence[ContainerGeneric[_T1]] | _T1
```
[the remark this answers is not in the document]
hmc-cs-mdrissi: It's mainly use as an input type in other stub files. An example is model can have a loss function passed to it of type ContainerGeneric[Loss] (Loss, dict of loss, list of loss, any nested flavor).

It's not used yet so I'm neutral to dropping it and adding it back when I add other stubs files that do need it.

stubs/tensorflow/tensorflow/keras/layers.pyi
hmc-cs-mdrissi: Positional only looks to make ci happy so I’ve done that.

PR #9708, opened 2023-02-10
Update `xml.sax.parse` to accept `StrPath`

willfrey:
It appears to support `os.PathLike[str]` in addition to just `str`, based on trying it with `pathlib.Path` and a custom class implementing `__fspath__`.


willfrey:
I also found this, which supports making this change to the annotation

https://github.com/python/cpython/blob/61f2be08661949e2f6dfc94143436297e60d47de/Lib/xml/sax/saxutils.py#L342-L343

srittau:
Thanks! I just tested it on Python 3.7 and 3.8 and it seems not to work with 3.7 (but with 3.8). Could you add a `version_info` check for that?

stdlib/xml/sax/__init__.pyi
```diff
@@ -34,7 +34,9 @@ else:
     def make_parser(parser_list: list[str] = ...) -> XMLReader: ...
 
 def parse(
-    source: str | _SupportsReadClose[bytes] | _SupportsReadClose[str], handler: ContentHandler, errorHandler: ErrorHandler = ...
+    source: StrPath | _SupportsReadClose[bytes] | _SupportsReadClose[str],
```
Avasam, line 37: Shouldn't `source` also take `xmlreader.InputSource` ? `source` is coerced into an `InputSource` and the docstring also seems to imply one can be passed directly.
https://github.com/python/cpython/blob/3.11/Lib/xml/sax/saxutils.py#L338
https://github.com/python/cpython/blob/3.11/Lib/xml/sax/expatreader.py#L105
https://github.com/python/cpython/blob/3.11/Lib/xml/sax/xmlreader.py#L117

@srittau Dunno if you wanna consider this outside the scope of this PR, because the type for parameter `source` is missing in a few locations. But I wanted to at least raise this.

Avasam:
> Thanks! I just tested it on Python 3.7 and 3.8 and it seems not to work with 3.7 (but with 3.8). Could you add a `version_info` check for that?

Source to confirm your suspicions: https://github.com/python/cpython/blob/3.7/Lib/xml/sax/saxutils.py#L339

willfrey:
@srittau Can do!

srittau (approved):

PR #9711, opened 2023-02-11
Fix stubtest errors for updating aiofiles.

hmc-cs-mdrissi:
Resolves [stubsabot](https://github.com/python/typeshed/pull/9703). Not a library I've used so generally read over implementation.


stubs/aiofiles/@tests/stubtest_allowlist.txt
```diff
@@ -9,6 +9,15 @@ aiofiles.threadpool.binary.AsyncBufferedIOBase.readable
 aiofiles.threadpool.binary.AsyncBufferedIOBase.seekable
 aiofiles.threadpool.binary.AsyncBufferedIOBase.tell
 aiofiles.threadpool.binary.AsyncBufferedIOBase.writable
+aiofiles.threadpool.binary.AsyncIndirectBufferedIOBase.close
```
hmc-cs-mdrissi, line 12: All of the allowlist are copy pasting the non-indirect version above it. This version introduces indirect variants of most of IO classes that need exception for same reason (at runtime they delegate with *args/**kwargs).

stubs/aiofiles/aiofiles/base.pyi
```diff
@@ -13,6 +13,9 @@ class AsyncBase(Generic[_T]):
     def __aiter__(self: Self) -> Self: ...
     async def __anext__(self) -> _T: ...
 
+class AsyncIndirectBase(AsyncBase[_T]):
+    def __init__(self, name: str, loop: Any, executor: Any, indirect: Callable[[], TextIO | BinaryIO]) -> None: ...
```
hmc-cs-mdrissi, line 17: I'm least confident about the type for indirect. In implementation indirect is only passed sys.stdin/sys.stdin.buffer/etc. It looks like intent is if _T is str then TextIO (or maybe IO[str]) while if _T is bytes then BinaryIO. 

Could add overloads to describe _T relationship although confused why _T is unconstrained type variable (seems like it should be str/bytes always).

stubs/aiofiles/aiofiles/threadpool/binary.pyi
```diff
@@ -34,8 +36,15 @@ class AsyncBufferedIOBase(_UnknownAsyncBinaryIO):
...
 class AsyncBufferedReader(AsyncBufferedIOBase):
     async def peek(self, __size: int = ...) -> bytes: ...
 
+class AsyncIndirectBufferedReader(AsyncIndirectBase[bytes], AsyncBufferedIOBase):
+    async def peek(self, __size: int = ...) -> bytes: ...
+
 class AsyncFileIO(_UnknownAsyncBinaryIO):
     async def readall(self) -> bytes: ...
+
+class AsyncIndirectFileIO(AsyncIndirectBase[bytes], AsyncFileIO): ...
```
AlexWaygood, line 50: Similar comment to the one I made about `AsyncIndirectBufferedIOBase` -- I'm wary of inheriting from `AsyncFileIO` here.

stubs/aiofiles/aiofiles/threadpool/text.pyi
```diff
@@ -37,3 +37,5 @@ class AsyncTextIOWrapper(AsyncBase[str]):
     def name(self) -> FileDescriptorOrPath: ...
     @property
     def mode(self) -> str: ...
+
+class AsyncTextIndirectIOWrapper(AsyncIndirectBase[str], AsyncTextIOWrapper): ...
```
AlexWaygood, line 41: Similar comment to the one I made about AsyncIndirectBufferedIOBase -- I'm wary of inheriting from `AsyncTextIOWrapper` here.

AlexWaygood (commented):
Thank you, and sorry for the delay in anybody getting to this! A few thoughts:

JelleZijlstra:
@hmc-cs-mdrissi could you respond to Alex's feedback?

stubs/aiofiles/aiofiles/threadpool/binary.pyi
```diff
@@ -34,8 +36,15 @@ class AsyncBufferedIOBase(_UnknownAsyncBinaryIO):
     @property
     def raw(self) -> FileIO: ...
 
+class AsyncIndirectBufferedIOBase(AsyncIndirectBase[bytes], AsyncBufferedIOBase): ...
+
 class AsyncBufferedReader(AsyncBufferedIOBase):
     async def peek(self, __size: int = ...) -> bytes: ...
 
+class AsyncIndirectBufferedReader(AsyncIndirectBase[bytes], AsyncBufferedIOBase):
```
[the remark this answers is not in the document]
hmc-cs-mdrissi: Guessing you meant AsyncIndirectBufferedIOBase double checking the runtime?

stubs/aiofiles/aiofiles/threadpool/binary.pyi
```diff
@@ -34,8 +36,15 @@ class AsyncBufferedIOBase(_UnknownAsyncBinaryIO):
     @property
     def raw(self) -> FileIO: ...
 
+class AsyncIndirectBufferedIOBase(AsyncIndirectBase[bytes], AsyncBufferedIOBase): ...
```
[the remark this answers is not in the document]
hmc-cs-mdrissi: Sure I've duplicated methods needed from AsyncBufferedIOBase

stubs/aiofiles/aiofiles/threadpool/text.pyi
hmc-cs-mdrissi: This one inherits a lot of methods. I added _UnknownAsyncTextIO to have methods defined once and have both public text io's inherit from that. The mro now for public classes matches runtime.

If you'd prefer to not use that solution I'll do duplication way similar to others.

stubs/aiofiles/aiofiles/base.pyi
hmc-cs-mdrissi: Hmm, I'd prefer to leave that for a follow up/future pr to be able to see effect by itself.

hmc-cs-mdrissi:
Sorry on delay. I think I've covered all comments now.

stubs/aiofiles/aiofiles/base.pyi
AlexWaygood: sure

stubs/aiofiles/aiofiles/threadpool/text.pyi
AlexWaygood: The solution you have looks good

AlexWaygood (approved):
Thanks for the fixes!

PR #9713, opened 2023-02-11
Set HOMEBREW_NO_INSTALL_CLEANUP to speedup brew installs

Avasam:
As the inline comment says: Speedup brew install. Environments are isolated, no need to cleanup old versions.
While working on a PR that touched pyaudio, I noticed this from runs like https://github.com/python/typeshed/actions/runs/4148676771/jobs/7176901691#step:5:54

I also moved the `NONINTERACTIVE` under `env`

### DIFF
--- a/.github/workflows/daily.yml
+++ b/.github/workflows/daily.yml
@@ -18,6 +18,8 @@ concurrency:
   cancel-in-progress: true
 
 env:
+  HOMEBREW_NO_INSTALL_CLEANUP: 1 # Speedup brew install. Environments are isolated, no need to cleanup old versions
+  NONINTERACTIVE: 1 # Required for brew install on CI
   PIP_DISABLE_PIP_VERSION_CHECK: 1
   FORCE_COLOR: 1
   TERM: xterm-256color # needed for FORCE_COLOR to work on mypy on Ubuntu, see https://github.com/python/mypy/issues/13817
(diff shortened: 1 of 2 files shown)


JelleZijlstra:
Thanks!

AlexWaygood:
I don't know much about homebrew -- would it be worth adding the other two environment variables that were just added to CPython in https://github.com/python/cpython/pull/102074/files as well?

Avasam:
> I don't know much about homebrew -- would it be worth adding the other two environment variables that were just added to CPython's workflow in [python/cpython#102074 (files)](https://github.com/python/cpython/pull/102074/files) as well?

Me neither, but here's the doc: https://docs.brew.sh/Manpage#environment
`HOMEBREW_NO_AUTO_UPDATE` definitely. For the same reasons we don't need to update pip in the CI
`HOMEBREW_NO_ANALYTICS` is up to you I guess. Analytics is good for the homebrew devs. I think it's more a matter of opting out of data collection than speeding up CI: https://docs.brew.sh/Analytics

AlexWaygood:
> `HOMEBREW_NO_ANALYTICS` is up to you I guess. Analytics is good for the homebrew devs. I think it's more a matter of opting out of data collection than speeding up CI (although anything that speeds up CI isn't bad either): https://docs.brew.sh/Analytics#opting-out

Running stubtest on macOS in CI is consistently slowesr/more backed-up than the other two operating systems, so I'd be up for it tbh :p

AlexWaygood:
Also, same as CPython (https://github.com/python/cpython/pull/102074#discussion_r1112007421), we're probably _somewhat_ atypical users of homebrew 😄

PR #9714, opened 2023-02-11
Simplify `pyrightconfig.json` , add missing comments and extra strict settings in all pyright configs

Avasam:
First commit simply simplifies `pyrightconfig.json`, adds missing comments to all 3 files and moves a TODO in `pyrightconfig.stricter.json` so that both base and stricter configs are identical below the `// Extra strict settings` comment.

Second commit turns on settings that are not on by default even with `"typeCheckingMode": "strict"`. Or excplicitely disables them if should be.
`"reportShadowedImports": "none"` has an added advantage of stopping Pylance from highlighting entire stdlib stub files:
image

Fixed a small `reportUninitializedInstanceVariable` issue in tests.


test_cases/stdlib/asyncio/check_task.py
```diff
@@ -4,7 +4,7 @@
 
 
 class Waiter:
-    def __init__(self) -> None:
+    def __init__(self) -> None:  # pyright: ignore [reportMissingSuperCall] # Is this a false-positive?
```
hmc-cs-mdrissi, line 7: It is not a false positive. Every class can potentially be a subclass. For waiter,

class Bar:
  …
class Foo(Waiter, Bar):
  …

in Foo’s mro Waiter has parent class of Bar. This rule is very strict (key reason why it is not part of strict mode) as python multiple inheritance means you almost always need super call if class needs to worry about multiple inheritance usage.

Avasam: Oh I see! Thanks for the explanation. I realize I also had disabled it in my own projects a while ago but didn't understand why (I also wasn't quite familiar with python typing at the time).

So this rule just doesn't make sense in the context of type testing.

test_cases/stdlib/builtins/check_dict.py
```diff
@@ -18,7 +18,7 @@
 
 
 class KeysAndGetItem(Generic[_KT, _VT]):
-    data: dict[_KT, _VT]
+    data: dict[_KT, _VT] = {}
```
AlexWaygood, line 21: I'd prefer for this to be initialised in an `__init__` method, so it's 100% unambiguous that it's an instance attribute and not a class attribute

pyrightconfig.json
```diff
@@ -9,35 +9,24 @@
...
+    // Allowed in base settings for incomplete stubs, checked in stricter settings
+    "reportIncompleteStub": "none",
+    "reportMissingParameterType": "none",
+    "reportUnknownMemberType": "none",
+    "reportUnknownParameterType": "none",
+    "reportUnknownVariableType": "none",
+    // Extra strict settings
+    "reportCallInDefaultInitializer": "error",
+    "reportImplicitStringConcatenation": "error",
+    "reportUninitializedInstanceVariable": "error",
```
AlexWaygood, line 22: This one won't have any effect for stubs, right? (Same comment for the same line in the stricter config file)

pyrightconfig.testcases.json
```diff
@@ -6,12 +6,21 @@
         "stubs/**/@tests/test_cases"
     ],
     "typeCheckingMode": "strict",
+    // Extra strict settings
+    "reportShadowedImports": "error", // Don't accidentally name a file something that shadows stdlib
+    "reportCallInDefaultInitializer": "error",
```
AlexWaygood, line 11: Not sure it makes sense to have this enabled for our test-case files. It's often not great practice in Python code, but we probably wouldn't include this (anti-)pattern in a test case unless we had a reason to do so, in which case this lint would be annoying. flake8-bugbear also already checks for this in our `.py` files anyway.

pyrightconfig.testcases.json
```diff
@@ -6,12 +6,21 @@
...
+    "reportImplicitStringConcatenation": "error",
+    "reportPropertyTypeMismatch": "error",
```
AlexWaygood, line 13: Same comment here — the (anti-)pattern this lint guards against is often a bad idea, but we wouldn't use it in a test case unless we had a specific reason to do so, in which case this lint would just be annoying

test_cases/stdlib/builtins/check_dict.py
```diff
@@ -20,18 +20,21 @@
 class KeysAndGetItem(Generic[_KT, _VT]):
     data: dict[_KT, _VT]
 
+    def __init__(self, data: dict[_KT, _VT]):
```
AlexWaygood, line 23: I'm surprised pyright didn't complain about the return type being unknown!

```suggestion
    def __init__(self, data: dict[_KT, _VT]) -> None:
```

pyrightconfig.testcases.json
Avasam: I think this makes sense given `"enableTypeIgnoreComments": true` and the associated comment.

pyrightconfig.json
Avasam: Indeed, just like `reportMissingSuperCall`. I've updated it.

pyrightconfig.testcases.json
Avasam: Updated and added a comment with similar wording as the one for the other two configs.

test_cases/stdlib/builtins/check_dict.py
AlexWaygood: You marked this as resolved but the suggestion hasn't been applied — any reason why not? :)

Avasam: Must be a Github android app bug (been trying it over the last week), or I miskicked.

AlexWaygood (approved):
Thank you!

PR #9715, opened 2023-02-11
Ignore flake8 extensions stubbed by typeshed

Avasam:
The in-code comment explains the rationale:
Some flake8 plugins are stubbed by typeshed. When working on a stub, it is useful to install it in your venv for intellisense and reference reasons. Because of this, it can be expected to find said plugins in the local python environment, which are automatically picked-up by flake8. Resulting in a bunch of extra flake8 errors that only appear locally.

Also simplified configs a bit by de-duplicating violations that are ignored in all files.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,3 +1,13 @@
+# The following rules come from plugins that are not used by typeshed.
+# Since typeshed stubs them, they can still be expected to be found in a
+# developer's venv for intellisense and reference reasons
+#     A      flake8-builtins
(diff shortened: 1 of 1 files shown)


AlexWaygood (approved):
If I'm locally installing a package that's stubbed by typeshed, I usually do it in a "throwaway" virtual environment that's entirely separate from my usual typeshed virtual environment, for this very reason. But this seems reasonable!

PR #9716, opened 2023-02-11
Fix misnamed parameter in stubtest_third_party

Avasam:
I spotted this a while ago, but then forgot about it again.
Seems like when I introduced this in #9173 we totally glanced over that the name of the parameter is completely wrong, it's not specified "stubs", but specified "platforms" (in METADATA.toml)

### DIFF
--- a/.github/workflows/daily.yml
+++ b/.github/workflows/daily.yml
@@ -91,7 +91,7 @@ jobs:
             PYTHON_EXECUTABLE="python"
           fi
 
-          $PYTHON_EXECUTABLE tests/stubtest_third_party.py --specified-stubs-only --num-shards 4 --shard-index ${{ matrix.shard-index }}
+          $PYTHON_EXECUTABLE tests/stubtest_third_party.py --specified-platforms-only --num-shards 4 --shard-index ${{ matrix.shard-index }}
   stub-uploader:
     name: Run the stub_uploader tests
     runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


AlexWaygood (commented):
It's hard to find a great name for this flag. I interpreted `--specified-stubs-only` as being shorthand for "only run stubtest on the stubs that specify they're tested on this platform in their `METADATA.toml` files"

Avasam:
> It's hard to find a great name for this flag. I interpreted `--specified-stubs-only` as being shorthand for "only run stubtest on the stubs that specify they're tested on this platform in their `METADATA.toml` files"

I totally meant it as "only run if the current platform is specified" with a focus on the platform.

The problem with the name `--specified-stubs-only` is that it implies only specified stubs are ran. Which stubtest already does by virtue of its `dists` arguments.
Your interpretation makes sense, but the problem I see is that it's not clearly implied what's specified: *"only run stubtest on the stubs that **specify** they're tested **on this platform** in their METADATA.toml files"*, vs. "only run stubtest on the stubs that **specify** they're tested (let's say)**without isgnoreing missing stubs their METADATA.toml files" for example.

A more realistic example of specifying something that reuses the same flag would be adding a python version specification. But I'd rather renamed it `specified-system-only` then.

At the end of the day it's a flag meant only to be used by the CI, not directly by devs. And there's a description that comes with it. So it's not major. But it was a typo from what I originally meant (the kind of typo where you think about something and end up subconsciously typing that word instead). So it can either be fixed, or retroactively change the meaning/semantic I meant and it's no longer a typo 😉

JelleZijlstra (approved):
The new name does seem better from a cursory reading of the code.

PR #9725, opened 2023-02-13
Update PyYAML dump_all functions

bartvm:
The dump_all functions actually accept any iterable ([the docs](https://pyyaml.org/wiki/PyYAMLDocumentation) say: "`yaml.dump_all` accepts a list or a generator producing Python objects to be serialized into a YAML document.").


srittau (approved):
Thanks!

PR #9732, opened 2023-02-14
Add comments when subclassing Any

Avasam:
Since maintainers seems to be in favor of keeping `--allow-subclassing-any`, this adds a few comments on the few cases left and closes #9491
Everything else have been completed in different PRs (except one, but I'm fine closing #9491 with 1 left)


srittau (approved):
Thanks!

PR #9733, opened 2023-02-14
Add types for PIL.ImageTk

TeamSpen210:
This fills in the types in the `PIL.ImageTk` module, as well as adding a test to verify that these classes are accepted in `tkinter` (which appeared to break at some point). Due to how the `tkinter._Image` protocol is currently specified, this did mean I needed to add a `tk` attribute to `BitmapImage`, though it doesn't actually have this attribute. Unfortunately the actual criteria for being accepted as a Tkinter attribute is simply returning a valid image name from `__str__()`, which isn't really possible to describe in the type system. Perhaps a typing-only base class with no members that the images inherit from would work better?


stdlib/tkinter/__init__.pyi
```diff
@@ -666,7 +666,7 @@ class Wm:
     iconmask = wm_iconmask
     def wm_iconname(self, newName: Incomplete | None = None) -> str: ...
     iconname = wm_iconname
-    def wm_iconphoto(self, default: bool, __image1: Image, *args: Image) -> None: ...
+    def wm_iconphoto(self, default: bool, __image1: _ImageSpec, *args: _ImageSpec) -> None: ...
```
Akuli, line 669: Ideally this wouldn't take `_ImageSpec` because it does not work with bitmap images. But I don't have any better suggestions. Tkinter images have a `.type()` method that returns `'photo'` or `'bitmap'`, but it doesn't help because PIL images don't have it.

stubs/Pillow/PIL/ImageTk.pyi
```diff
@@ -1,18 +1,53 @@
...
-    def __init__(self, image: Incomplete | None = ..., size: Incomplete | None = ..., **kw) -> None: ...
+    tk: _tkinter.TkappType
+    def __init__(
+        self,
+        image: Image | _Mode | None = None,
+        size: _Size | None = None,
+        *,
+        file: StrOrBytesPath | SupportsRead[bytes] = ...,
+        data: ReadableBuffer = ...,
+        # These are the same as tkinter.PhotoImage:
```
Akuli, line 17: Maybe we should add a comment to `tkinter/__init__.pyi` that reminds about keeping these in sync.

stubs/Pillow/PIL/ImageTk.pyi
```diff
@@ -1,18 +1,53 @@
...
-    def height(self): ...
-    def paste(self, im, box: Incomplete | None = ...) -> None: ...
+    def width(self) -> int: ...
+    def height(self) -> int: ...
+    # box is deprecated and unused
+    def paste(self, im: Image, box: _Box | None = ...) -> None: ...
 
 class BitmapImage:
-    def __init__(self, image: Incomplete | None = ..., **kw) -> None: ...
+    tk: _tkinter.TkappType  # Not actually present, but required for tkinter._Image protocol
```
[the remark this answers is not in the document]
TeamSpen210: We could, but that would be really generic, it'd easily produce false positives.

stdlib/tkinter/__init__.pyi
TeamSpen210: Perhaps instead of a protocol, we could replace `_Image` with a pair of typing-only marker abstract base classes, that the tkinter and PIL images inherit from.

stubs/Pillow/PIL/ImageTk.pyi
```diff
@@ -1,18 +1,52 @@
...
+import tkinter
+from _typeshed import ReadableBuffer, StrOrBytesPath, SupportsRead
 from typing import Any
 
-class PhotoImage:
-    tk: Any
-    def __init__(self, image: Incomplete | None = ..., size: Incomplete | None = ..., **kw) -> None: ...
+from PIL.Image import Image, _Box, _Mode, _Size
+
+class PhotoImage(tkinter._PhotoImageLike):
```
Akuli, line 8: This won't work very well with how third-party stubs are released. When running a type checker, you typically have a new version of third-party stubs (released every 3 hours if pull requests have been merged) and an old version of stdlib (comes with the type checker, mypy is released less than once in a month I think). So any users who upgrade their `pillow-stubs` package wouldn't actually get the protocol in  `tkinter/__init__.pyi`.

My experience is that mypy tends to hide errors coming from stub packages you have installed, and treat anything unknown as `Any`, so it would think that `ImageTk.PhotoImage` inherits from `Any` and therefore report no errors for trying to do `some_pil_photoimage.attribute_that_doesnt_exist`.

But this is really a temporary problem that will go away as new versions of type checkers are released.

stdlib/tkinter/__init__.pyi
```diff
@@ -3206,12 +3206,19 @@ class OptionMenu(Menubutton):
...
 
-class _Image(Protocol):
-    tk: _tkinter.TkappType
-    def height(self) -> int: ...
-    def width(self) -> int: ...
+# Marker to indicate that it is a valid bitmap/photo image. PIL implements compatible versions
+# which don't share a class hierachy. The actual API is a __str__() which returns a valid name,
+# not something that type checkers can detect.
+@type_check_only
+class _Image: ...
```
Akuli, line 3213: I'm not sure if defining a "fake" base class has some surprising consequences that I'm not aware of. Stubbing something differently than what the runtime does is often a bad idea, but here we need to do it anyway. But this seems like a good idea and I think we should try it.

That said, I don't think a protocol with `width(self) -> int` and `height(self) -> int` would ever match the wrong kind of object in practice. Most libraries would define `width` and `height` as properties or attributes rather than as methods that take no arguments.

Akuli (commented):
Tkinter is tricky :)

stubs/Pillow/PIL/ImageTk.pyi
TeamSpen210: That is very true, I didn't think about the different release schedules. Not sure what to do there.

Akuli: Me neither. Both options have their own advantages and disadvantages :)

I'll probably merge this in a few days, unless someone comes up with a better idea or some reasoning to prefer the `width()`,`height()` protocol over fake classes.

Issue #9520, opened 2023-01-13
Should we delete the `_distutils` directory from our `setuptools` stubs ?


Avasam:
From @AlexWaygood :
https://github.com/python/typeshed/pull/9460#issuecomment-1372047833
> I sort of wonder if we shouldn't just delete the `_distutils` directory from our `setuptools` stubs. The frequency of the changes to that directory (and the leading underscore in the name) indicates to me that the maintainers see the whole directory as an implementation detail, and it just causes work for us to keep the directory up to date. I'm also not sure it's very useful keeping the directory around for type-checking purposes, since `setuptools` does some terrible `importlib` hack so that you can import the directory just by doing `import distutils` at runtime -- so there's little reason to ever import anything from `setuptools._distutils` directly.

(also distutils being deprecated)

AlexWaygood:
The distutils situation on 3.10+ at the moment is a complete mess.

There's currently two versions of `distutils`: the dead, decaying version in the stdlib, and the version maintained by setuptools that lives in the `setuptools/_distutils` directory. Because of the importlib hack `setuptools` does, if you do `import distutils` on 3.10+, you're actually getting `setuptools._distutils` rather than the stdlib `distutils`. But there's more! For bootstrapping reasons, setuptools comes pre-installed in most builds of Python. That means that if you have a fresh Python install, _without having downloaded anything from pip_, you get the third-party version of distutils rather than the stdlib distutils.

Needless to say, stubtest doesn't understand any of this, and is very confused by the whole thing. Most patch releases of 3.10 these days bump the bundled version of `setuptools` they come with. That then leads to our stdlib stubtest exploding with errors, because stubtest thinks it's the stdlib stubtest that's changed, when in actual fact it's stuff in `setuptools/_distutils` that's changed.

Practically speaking, it's now pretty difficult to import the stdlib `distutils` on 3.10 (you get the setuptools version whether you like it or not!).

srittau:
If I remember correctly, some setuptools classes are also subclasses of distutils, which could it make difficult to remove it.

Avasam:
If I understand correctly. Unless https://github.com/python/cpython/pull/101039 is backported, typeshed will have to wait until at least Python 3.10 (the first version where it's deprecated) is the oldest supported python version by type-checkers to even consider removing `stdlib/distutils`. Or at least have `stdlib/distutils` reflect `stubs/setuptools/setuptools/_distutils`

[...]

PR #9734, opened 2023-02-14
Fix `stdlib/disutils` testing

Avasam:
In hindsight the solution was quite simple: Just don't have setuptools installed in a test environment.
Turned on stubtest for distutils for python 3.10+
Fixed enough stubtest to get the CI green.

Missing variables and functions were added with stubgen. Then filled in necessary types for pyright to pass.

Solves one of the two concerns raised in #9520 (namely distutils being a mess to test, and untestable on 3.10+)
As for the second one: `distutils` shouldn't be imported with setuptools>=60 installed. But even if we wanted to support that use-case, we could.


Avasam:
In its current state, this PR shows that the tests are run correctly on all version.
But I could split off stubtest allowlist entries fixes into a different PR, and add a few entries for now just to get the CI green.

.github/workflows/daily.yml
```diff
@@ -45,7 +45,7 @@ jobs:
           cache: pip
           cache-dependency-path: requirements-tests.txt
       - name: Install dependencies
-        run: pip install $(grep mypy== requirements-tests.txt)
+        run: pip install typing_extensions $(grep ^pathspec requirements-tests.txt)
```
AlexWaygood, line 48: Since we're switching to creating a "clean" virtual environment dynamically in this PR in which to run the test, we may as well simplify this step a little bit (and bring it in line with all our other CI jobs that `pip install` stuff). There's no longer any need to worry about the script running in an isolated environment:

```suggestion
        run: pip install -r requirements-tests.txt
```

.github/workflows/stubtest_stdlib.yml
```diff
@@ -44,6 +44,6 @@ jobs:
           cache: pip
           cache-dependency-path: requirements-tests.txt
       - name: Install dependencies
-        run: pip install $(grep mypy== requirements-tests.txt)
+        run: pip install typing_extensions $(grep ^pathspec requirements-tests.txt)
```
AlexWaygood, line 47: Same rationale as for the `daily.yml` workflow

```suggestion
        run: pip install -r requirements-tests.txt
```

tests/stubtest_stdlib.py
```diff
@@ -6,55 +6,98 @@
...
+    with tempfile.TemporaryDirectory() as tmp:
+        venv_dir = Path(tmp)
+        try:
+            pip_exe, python_exe = make_venv(venv_dir)
+        except Exception:
+            print_error("fail")
+            raise
+
+        with Path("requirements-tests.txt").open() as requirements_txt:
+            requirements = [str(requirement) for requirement in parse_requirements(requirements_txt) if requirement.key == "mypy"]
```
AlexWaygood, line 35: I'd prefer it if we used the [`packaging.requirements.Requirement` class](https://packaging.pypa.io/en/stable/requirements.html#requirements) for this parsing logic, since the `pkg_resources` library is deprecated. (I looked into whether we could replace our usage of `pkg_resources` in `pytype_test.py`, but I concluded that it wasn't worth it -- we'd have to add `importlib_metadata` as a dependency in order for our tests to run on <3.10.)

stdlib/distutils/core.pyi
```diff
@@ -1,8 +1,17 @@
...
 from distutils.cmd import Command as Command
 from distutils.dist import Distribution as Distribution
 from distutils.extension import Extension as Extension
-from typing import Any
+from os import PathLike
+from typing import Any, AnyStr
+
+USAGE: str
+
+def gen_usage(script_name: PathLike[AnyStr]) -> str: ...
```
AlexWaygood, line 11: Using a `TypeVar` just once in the signature here doesn't make much sense; `_typeshed.StrOrBytesPath` would be a better fit

```suggestion
def gen_usage(script_name: StrOrBytesPath) -> str: ...
```

stdlib/distutils/core.pyi
```diff
@@ -1,8 +1,17 @@
...
+
+setup_keywords: Incomplete
+extension_keywords: Incomplete
```
AlexWaygood, line 14:
```suggestion
setup_keywords: tuple[str, ...]
extension_keywords: tuple[str, ...]
```

stdlib/distutils/cygwinccompiler.pyi
```diff
@@ -1,4 +1,20 @@
...
 class CygwinCCompiler(UnixCCompiler): ...
 class Mingw32CCompiler(CygwinCCompiler): ...
+
+CONFIG_H_OK: str
+CONFIG_H_NOTOK: str
+CONFIG_H_UNCERTAIN: str
+
+def check_config_h() -> tuple[Literal["ok", "not ok", "uncertain"], str]: ...
+
+RE_VERSION: Incomplete
```
AlexWaygood, line 17:
```suggestion
RE_VERSION: re.Pattern[bytes]
```

stdlib/distutils/dist.pyi
```diff
@@ -1,8 +1,10 @@
-from _typeshed import FileDescriptorOrPath, SupportsWrite
+from _typeshed import FileDescriptorOrPath, Incomplete, SupportsWrite
 from collections.abc import Iterable, Mapping
 from distutils.cmd import Command
 from typing import IO, Any
 
+command_re: Incomplete
```
AlexWaygood, line 6:
```suggestion
command_re: re.Pattern[str]
```

stdlib/distutils/fancy_getopt.pyi
```diff
@@ -1,14 +1,15 @@
...
 _GR: TypeAlias = tuple[list[str], OptionDummy]
 
-def fancy_getopt(
-    options: list[_Option], negative_opt: Mapping[_Option, _Option], object: Any, args: list[str] | None
-) -> list[str] | _GR: ...
-def wrap_text(text: str, width: int) -> list[str]: ...
+longopt_pat: str
+longopt_re: Incomplete
+neg_alias_re: Incomplete
+longopt_xlate: Incomplete
```
AlexWaygood, line 12:
```suggestion
longopt_re: re.Pattern[str]
neg_alias_re: re.Pattern[str]
longopt_xlate: dict[int, int]
```

stdlib/distutils/fancy_getopt.pyi
```diff
@@ -20,5 +21,14 @@ class FancyGetopt:
     def get_option_order(self) -> list[tuple[str, str]]: ...
     def generate_help(self, header: str | None = None) -> list[str]: ...
 
+def fancy_getopt(
+    options: list[_Option], negative_opt: Mapping[_Option, _Option], object: Any, args: list[str] | None
+) -> list[str] | _GR: ...
+
+WS_TRANS: Incomplete
```
AlexWaygood, line 28:
```suggestion
WS_TRANS: dict[int, str]
```

tests/stubtest_stdlib.py
Avasam: I can't find how to parse a `requirements.txt` file using `packaging.requirements.Requirement` without also having to manually remove comments. Sure the `mypy` line doesn't have comments right now. But if it does, I would like if it didn't break.

AlexWaygood: `Requirement(line.split("#")[0].strip())`? :)

Avasam: That's what I ended up going for too

tests/stubtest_stdlib.py
```diff
@@ -6,55 +6,105 @@
...
+
+        # Install the same mypy version as in "requirements-tests.txt"
+        with Path("requirements-tests.txt").open() as requirements_txt:
+            mypy_requirement = next(
+                (
+                    Requirement(requirement.split("#")[0])
+                    for requirement in requirements_txt.readlines()
+                    if requirement.startswith("mypy")
+                )
+            )
```
AlexWaygood, line 42: On second thought, we should probably use this preexisting function for this logic, no?

https://github.com/python/typeshed/blob/d92332d120ed0a042d470d326809f9c548be7967/tests/utils.py#L82-L85

(If you want to improve that function in this PR by stripping comments, that's fine!)

Avasam: Didn't see or remember that. Absolutely!

AlexWaygood (approved):
LGTM. @hauntsaninja, does this seem like a reasonable approach to you? It involves adding a little bit of complexity to `stubtest_stdlib.py`, but I think it's worth it.

hauntsaninja (changes requested):
(only looked at stubtest_stdlib)

Approach looks fine, but let's remove the return code logic from the pip commands. We only care about preserving the stubtest return code, we don't care about pip's return codes. I'd also not capture output from those installs.

Avasam:
I see pip's output is verbose enough printing our own error message is redundant
image

hauntsaninja (commented):
Thanks, changes to stubtest_stdlib.py lgtm!

AlexWaygood (approved):
Thank you!

AlexWaygood:
> I'd also not capture output from those installs.

Hmm, are you sure about that @hauntsaninja? The output from pip is now pretty noisy when I run `stubtest_stdlib.py` locally (which I do quite a lot):

```
(.venv) C:\Users\alexw\coding\typeshed>python tests/stubtest_stdlib.py
Collecting mypy==1.0
  Using cached mypy-1.0.0-cp311-cp311-win_amd64.whl (8.8 MB)
Collecting typing-extensions>=3.10
  Using cached typing_extensions-4.5.0-py3-none-any.whl (27 kB)
Collecting mypy-extensions>=0.4.3
  Using cached mypy_extensions-1.0.0-py3-none-any.whl (4.7 kB)
Installing collected packages: typing-extensions, mypy-extensions, mypy
Successfully installed mypy-1.0.0 mypy-extensions-1.0.0 typing-extensions-4.5.0

[notice] A new release of pip available: 22.3.1 -> 23.0.1
[notice] To update, run: C:\Users\alexw\AppData\Local\Temp\tmppzy1r9wi\Scripts\python.exe -m pip install --upgrade pip
Found existing installation: setuptools 65.5.0
Uninstalling setuptools-65.5.0:
  Successfully uninstalled setuptools-65.5.0
C:\Users\alexw\AppData\Local\Temp\tmppzy1r9wi\Scripts\python.exe -m mypy.stubtest --check-typeshed --custom-typeshed-dir . --allowlist tests\stubtest_allowlists\py3_common.txt --allowlist tests\stubtest_allowlists\py311.txt --allowlist tests\stubtest_allowlists\win32.txt --allowlist tests\stubtest_allowlists\win32-py311.txt
Success: no issues found in 484 modules
stubtest succeeded
```

I'd find the output easier to parse if we suppressed `pip`'s prints to the terminal (other than when pip errors out, obviously).

PR #9739, opened 2023-02-15
Bump pyright to 1.1.294; improve `psutil` stubs

AlexWaygood:


srittau:
I only checked the Darwin failure: It fails on the return type `-> psfullmem`, which is typed as `pfullmem: Any`. I guess this isn't intended to be a type alias, just an untyped field (should be `Incomplete`) and therefore seems fine to me. This looks like a pyright bug to me.

AlexWaygood:
> I only checked the Darwin failure: It fails on the return type `-> psfullmem`, which is typed as `pfullmem: Any`. I guess this isn't intended to be a type alias, just an untyped field (should be `Incomplete`) and therefore seems fine to me. This looks like a pyright bug to me.

Possibly, but we can also just fix the error by improving the stubs here 😉

srittau (approved):
I didn't confirm the pfullmem fields, but LGTM.

PR #9740, opened 2023-02-16
stubs for xtea library

rajohnson:
Adds stubs for the xtea encyption library. 

I think that this is ready to review, unless CI catches something I've missed.


pyrightconfig.stricter.json
```diff
@@ -72,7 +72,8 @@
         "stubs/tqdm",
         "stubs/ttkthemes",
         "stubs/urllib3",
-        "stubs/vobject",
+        "stubs/vobject",,
```
JelleZijlstra, line 75:
```suggestion
        "stubs/vobject",
```

stubs/xtea/xtea/counter.pyi
```diff
@@ -0,0 +1,12 @@
...
+PY_3: int
+
+def to_bytes(integer: int, length: int, byteorder: str) -> bytes: ...
+def from_bytes(bytesarray: bytes | bytearray, byteorder: str) -> int: ...
+
+class Counter:
+    byteorder: str
+    def __init__(self, nonce: bytes | bytearray, byteorder: Literal["big", "little"] = "big") -> None: ...
+    def __call__(self): ...
+    def reset(self) -> None: ...
```
JelleZijlstra, line 12:
```suggestion
from typing_extensions import Literal
from typing import SupportsBytes
from _typeshed import ReadableBuffer

PY_3: bool

def to_bytes(integer: int, length: int, byteorder: Literal["big", "little"] ) -> bytes: ...
def from_bytes(bytesarray: SupportsBytes | ReadableBuffer, byteorder: Literal["big", "little"] ) -> int: ...

class Counter:
    byteorder: str
    def __init__(self, nonce: SupportsBytes | ReadableBuffer, byteorder: Literal["big", "little"] = "big") -> None: ...
    def __call__(self) -> bytes: ...
    def reset(self) -> None: ...
```
Mostly syncing the types with `int.to_bytes` and `int.from_bytes`, which these delegate to.

stubs/xtea/xtea/__init__.pyi
```diff
@@ -0,0 +1,25 @@
...
+
+def new(key, **kwargs): ...
+
+class XTEACipher(PEP272Cipher):
+    block_size: int
+    IV: bytes | None
+    counter: Callable[[], bytes] | None
+    rounds: float | None
+    cycles: float | None
+    endian: str | None
```
JelleZijlstra, line 22:
```suggestion
    rounds: int
    cycles: int
    endian: str
```
From the `__init__` code, looks like these can't be None and the first two are always ints.

hauntsaninja:
Note that stub_uploader is different from stubtest.

stub_uploader will not be happy with `pep272_encryption` as a requirement because of:
https://github.com/typeshed-internal/stub_uploader/blob/32dca63e0cdb13393454bde4ed76fefa72d99a4f/stub_uploader/metadata.py#L151-L163
I should write up an issue that's not on that PR, but see the link in code

stubs/xtea/xtea/__init__.pyi
```diff
@@ -0,0 +1,25 @@
...
+    def __init__(self, key: bytes, mode: int | None = ..., **kwargs) -> None: ...
+    def encrypt_block(self, key: bytes, block, **kwargs): ...
+    def decrypt_block(self, key: bytes, block, **kwargs): ...
```
JelleZijlstra, line 25:
```suggestion
    def __init__(self, key: ReadableBuffer, mode: int | None = ..., **kwargs) -> None: ...
    def encrypt_block(self, key: ReadableBuffer, block: ReadableBuffer, **kwargs): ...
    def decrypt_block(self, key: ReadableBuffer, block: ReadableBuffer, **kwargs): ...
```
These are passed to `struct.unpack` which accepts any `ReadableBuffer`. (Though in fact the two methods appear to ignore the `key` parameter.)

rajohnson:
@hauntsaninja My understanding is that `pep272_encryption` would have to be reviewed by the typeshed team and manually added to `EXTERNAL_REQ_ALLOWLIST` in `stub_uploader/metadata.py`  in order for this to pass. Is this correct?  

`pep272_encryption` is pretty obscure and was created by the author of the library that I'm trying to make stubs for. I'm not sure that it will be worth your time if my understanding is correct, is it worth continuing this pull request?

hauntsaninja:
Yeah, I think it would be a little challenging to add `pep272_encryption` to that allowlist. I wrote up https://github.com/typeshed-internal/stub_uploader/issues/90 which has details of the exact situation.

Three alternative options are:
- Given that the author of `xtea` has type annotations on `pep272_encryption`, maybe they'd be willing to accept a PR adding annotations directly to `xtea`? In general, if types can be upstreamed that's preferable to having types in typeshed
- I'm not sure that this would work well here, but we could just include the API of `PEP272Cipher` in `XTEACipher` and have users of these stubs rely on substructural typing (Protocols)
- If you're willing to wait, maybe we can resolve some of the barriers that make us hesitant to add things to the allowlist in stub_uploader, as described in that issue

In all cases, we're very happy to review the stubs themselves and are grateful for your contribution of them! :-)

rajohnson:
Ok, contributing the stubs to xtea makes a lot of sense, I'll pursue that option first. Thanks for your help!

PR #9742, opened 2023-02-16
Fix default value of input, remove now unused allowlist entry for epoll

hauntsaninja:
Fixes #9741

PR #9743, opened 2023-02-16
Switch order of dataclass decorator overloads

koogoro:
When type checking in a non-strict Optional context where `None` inhabits all types, the `type[_T]`-type overload will always resolve over the `None`-type overload. Switching the order of these better represents the priority of overloads when running in this context. mypy ensures that listing an overload with a `None` argument second in non-strict-optional order results in a type error: https://github.com/python/mypy/blob/83660d0ad72ab5a61cea5fe5955e66c33ef54111/test-data/unit/check-overloading.test#L4619


hauntsaninja (approved):
Ooo, --no-strict-optional in 2023 is bold ;-) But this looks good, thank you!

hauntsaninja:
You can ignore CI failures, they're related to new Python versions being released and will be fixed by https://github.com/python/typeshed/pull/9742

PR #9746, opened 2023-02-17
Allow urlparse and urlsplit of bytes URLs

koogoro:
Right now, `urlparse(url)` and `urlsplit(url)` doesn't work if `url` is a `bytes`. This is because the only overload that can match this case is the one where `url: bytes | bytearray`, but even this one doesn't match since the default value for `scheme` is `""`, which isn't `bytes | bytearray | None`. Adding `Literal[""]` to the types for `scheme` won't cause the overloads to overlap, since they're differentiated by the type of `url`, and will allow `urlparse` applied to a single `bytes` argument to work again.


hauntsaninja (approved):
Thanks, this matches the runtime better!

AlexWaygood (commented):
Thanks! Looks like we could actually combine the first two overloads of `urlparse` and `urlsplit` in that case — want to do that?

koogoro:
Is there an easy way to do that? It seems hard to me since they still have different return types, but if there's a good way to combine them then I'll do it.

AlexWaygood:
> Is there an easy way to do that? It seems hard to me since they still have different return types, but if there's a good way to combine them then I'll do it.

Good point — we can't combine the overloads for `urlparse`, but we can for `urlsplit`

AlexWaygood:
Sorry, I meant to combine the second and third overloads, not the first two overloads! Apologies for the garbled review

AlexWaygood (approved):
Thanks!

PR #9747, opened 2023-02-18
Update pkg_resources-stubs for use in pytype_test

Avasam:
Fixes a type issue in pytype_test. https://github.com/python/typeshed/blob/main/tests/pytype_test.py#L158
Unfortunately I can't immediately remove the `type: ignore` comment because typeshed's tests depend on `types-setuptools`.


stubs/setuptools/pkg_resources/__init__.pyi
```diff
@@ -119,10 +125,15 @@ class EntryPoint:
...
-def get_distribution(dist: Requirement | str | Distribution) -> Distribution: ...
+@overload
+def get_distribution(dist: _D) -> _D: ...
+@overload
+def get_distribution(dist: _PkgReqType) -> DistInfoDistribution: ...
 
 class Distribution(IResourceProvider, IMetadataProvider):
-    PKG_INFO: str
+    PKG_INFO: ClassVar[str]
+    # Initialized to None, but is not meant to be instanciated directly
```
AlexWaygood, line 135: comment typo :)

```suggestion
    # Initialized to None, but is not meant to be instantiated directly
```

Avasam: That's my french coming through :P

stubs/setuptools/pkg_resources/__init__.pyi
```diff
@@ -119,10 +125,15 @@ class EntryPoint:
     def resolve(self) -> Any: ...
 
 def find_distributions(path_item: str, only: bool = ...) -> Generator[Distribution, None, None]: ...
-def get_distribution(dist: Requirement | str | Distribution) -> Distribution: ...
+@overload
+def get_distribution(dist: _D) -> _D: ...
+@overload
+def get_distribution(dist: _PkgReqType) -> DistInfoDistribution: ...
```
JelleZijlstra, line 131: This calls `get_provider()`, which according to our stub below returns a `Distribution`, not a `DistInfoDistribution`. I haven't traced the code to verify which is right, but we should be consistent.

stubs/setuptools/pkg_resources/__init__.pyi
```diff
@@ -119,10 +125,15 @@ class EntryPoint:
...
 
 class Distribution(IResourceProvider, IMetadataProvider):
-    PKG_INFO: str
+    PKG_INFO: ClassVar[str]
+    # Initialized to None, but is not meant to be instantiated directly
+    egg_info: str
```
JelleZijlstra, line 136: I don't see this attribute at all (on setuptools 67.3.1).

stubs/setuptools/pkg_resources/__init__.pyi
Avasam: image

Avasam: I think you're right. Didn't look too far into it but might depend on if the distribution was `.egg`, `.egg-info` or `.dist-info`

stubs/setuptools/pkg_resources/__init__.pyi
Avasam: Found it, it's in `NullProvider`. If you provide no metadata to Distribution, it uses `EmptyProvider` which is a `NullProvider`

```
>>> type(pkg_resources.get_distribution("numpy")._provider)
<class 'pkg_resources.PathMetadata'>
>>> type(Distribution()._provider)
<class 'pkg_resources.EmptyProvider'>
>>> type(get_provider("numpy"))
<class 'pkg_resources.DefaultProvider'>
```
`PathMetadata`, `EmptyProvider` and `DefaultProvider` all have `NullProvider` in common

stubs/setuptools/pkg_resources/__init__.pyi
```diff
@@ -243,19 +253,49 @@ def register_loader_type(loader_type: type, provider_factory: Callable[[types.Mo
...
+    module_path: str
     def __init__(self, path: str, egg_info: str) -> None: ...
 
-class ZipProvider(EggProvider): ...
+class ZipProvider(EggProvider):
+    eagers: list[str] | None
+    zip_pre: str
 
 class EggMetadata(ZipProvider, IResourceProvider):
+    loader: types._LoaderProtocol
```
JelleZijlstra, line 292: This comes from the `zipimporter` argument to `__init__` so the types should match.

stubs/setuptools/pkg_resources/__init__.pyi
```diff
@@ -243,19 +253,49 @@ def register_loader_type(loader_type: type, provider_factory: Callable[[types.Mo
...
 
 class IResourceProvider(IMetadataProvider): ...
-class NullProvider: ...
-class EggProvider(NullProvider): ...
+
+class NullProvider:
+    egg_name: str | None
+    egg_info: str | None
+    loader: types._LoaderProtocol | None
+    module_path: str | None
```
JelleZijlstra, line 261:
```suggestion
    module_path: str
```
The initializer unconditionally sets it to a string.

Avasam: EmptyProvider sets it explicitely to None

Should I keep `module_path: str | None` for sublcassing. Or add `# type: error[assignment]` to `EmptyProvider.module_path` ?

JelleZijlstra (approved):

PR #9750, opened 2023-02-18
Add error codes to type ignores and removed unused ones

JonathanPlasse:


AlexWaygood (changes requested):
Thanks for the PR! However, not using mypy error codes for `type: ignore`s in the `test_cases` directory is a deliberate policy. Please see #8083 for the motivation here :)

PR #9751, opened 2023-02-18
Remove unnecessary per-file-ignores in .flake8

JonathanPlasse:
The remove ignores were for `.pyi` files not `.py`.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -20,7 +20,7 @@
 
 [flake8]
 per-file-ignores =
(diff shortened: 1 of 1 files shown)


AlexWaygood:
Thanks, but I think we'd prefer to leave these as they are. These rules are all enforced by `black` in CI (see the comments on lines 1-6 of the file you're editing here), so it's annoying to have flake8 point out things that will be flagged and autofixed by `black` anyway before being committed to `main`.

PR #9752, opened 2023-02-18
Type `Loader` parameter in `PyYAML/yaml`

Avasam:
Fixes an unknown parameter issue at https://github.com/python/typeshed/blob/main/tests/check_consistent.py#L146

I didn't feel it necessary, but I could make the parameters more flexible by making protocols of `Reader & Parser & Scanner`, `Reader & Parser`, `Reader & Parser & Composer`, `Reader & Parser & BaseConstructor`, ...
Plus you'd need to add `type: ignore[misc]  # pyright: ignore` to all of 'em because of mixing Protocol and non-Protocol superclasses.


stubs/PyYAML/yaml/__init__.pyi
```diff
@@ -21,20 +23,20 @@ from .tokens import *
...
 _Yaml: TypeAlias = Any
 
-__with_libyaml__: Any
-__version__: str
-
 _T = TypeVar("_T")
 _Constructor = TypeVar("_Constructor", bound=BaseConstructor)
 _Representer = TypeVar("_Representer", bound=BaseRepresenter)
 
+__with_libyaml__: Any
```
AlexWaygood, line 30: https://github.com/yaml/pyyaml/blob/957ae4d495cf8fcb5475c6c2f1bce801096b68a5/lib/yaml/__init__.py#L12-L16

```suggestion
__with_libyaml__: bool
```

AlexWaygood (approved):
Looks good, just one nit :)

stubs/PyYAML/yaml/__init__.pyi
Avasam: Sure, may as well while we're here.

PR #9753, opened 2023-02-18
Mark `threading.Lock` as `final`

sobolevn:
It is actually a factory function: https://github.com/python/cpython/blob/128379b8cdb88a6d3d7fed24df082c9a654b3fb8/Lib/threading.py#L108

Related: https://discuss.python.org/t/make-threading-lock-rlock-factory-functions-subclassable/24008/2

I am leaving `RLock` aside, because it is more complex: python version is subclassable, while C version is not. But, we always use C version when calling `RLock`!


AlexWaygood:
The feedback from the discuss thread seems to be "you probably shouldn't be subclassing these, even if you can", so maybe we should do the same to `RLock` as well, even if the Python version is subclassable? :)

sobolevn:
I would prefer to wait a bit. Because I hope to properly deprecate pure-python `RLock` in 3.12!

And after that we can surely mark it as `@final` :)

AlexWaygood:
> I would prefer to wait a bit. Because I hope to properly deprecate pure-python `RLock` in 3.12!

I don't understand why this is relevant to whether or not we mark `RLock` with `@final` in the stub -- we mark several classes in the stdlib with `@final` even if they're, strictly speaking, subclassable, because you're not _meant_ to subclass them even if you _can_. E.g. see #9353 for a recent example :)

Having said that, this change is obviously correct in and of itself, so feel free to merge! :)

sobolevn:
Ok, I was thinking only about "you are able to", not "you are meant to".
Now this makes sense! 👍

AlexWaygood (approved):

PR #9754, opened 2023-02-18
Fix potentially unbound issues in stubsabot

Avasam:
Fixes "potentially unbound" type issues in stubsabot.
Simplify a bit the section about finding the first `py.typed` version and whether the stub is obsolete.

### DIFF
--- a/scripts/stubsabot.py
+++ b/scripts/stubsabot.py
@@ -170,8 +170,9 @@ async def release_contains_py_typed(release_to_download: PypiReleaseDownload, *,
         raise AssertionError(f"Unknown package type: {packagetype!r}")
 
 
-async def find_first_release_with_py_typed(pypi_info: PypiInfo, *, session: aiohttp.ClientSession) -> PypiReleaseDownload:
+async def find_first_release_with_py_typed(pypi_info: PypiInfo, *, session: aiohttp.ClientSession) -> PypiReleaseDownload | None:
     release_iter = pypi_info.releases_in_descending_order()
+    first_release_with_py_typed: PypiReleaseDownload | None = None
     while await release_contains_py_typed(release := next(release_iter), session=session):
         if not release.version.is_prerelease:
             first_release_with_py_typed = release
@@ -418,10 +419,12 @@ async def determine_action(stub_path: Path, session: aiohttp.ClientSession) -> U
     if latest_version in spec:
         return NoUpdate(stub_info.distribution, "up to date")
 
+    obsolete_since: PypiReleaseDownload | None = None
     is_obsolete = await release_contains_py_typed(latest_release, session=session)
     if is_obsolete:
-        first_release_with_py_typed = await find_first_release_with_py_typed(pypi_info, session=session)
-        relevant_version = version_obsolete_since = first_release_with_py_typed.version
+        obsolete_since = await find_first_release_with_py_typed(pypi_info, session=session)
+        assert obsolete_since is not None
+        relevant_version = obsolete_since.version
     else:
         relevant_version = latest_version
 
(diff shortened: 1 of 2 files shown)


hauntsaninja (changes requested):
Let's keep the diff_analysis where it was so we don't have to do work for obsolete stubs

Avasam:
> Let's keep the diff_analysis where it was so we don't have to do work for obsolete stubs

Oh yeah, I see. New commit should do that and still avoid "potentially unbound".

scripts/stubsabot.py
```diff
@@ -435,21 +438,22 @@ async def determine_action(stub_path: Path, session: aiohttp.ClientSession) -> U
 
     diff_info = await get_diff_info(session, stub_info, pypi_info, relevant_version)
     if diff_info is not None:
-        github_repo_path, old_tag, new_tag, diff_url = diff_info
+        *_, diff_url = diff_info
         links["Diff"] = diff_url
```
Avasam, line 442: I assume this is more readable than:
```suggestion
        links["Diff"] = diff_info[3]
```

scripts/stubsabot.py
```diff
@@ -435,21 +438,22 @@ async def determine_action(stub_path: Path, session: aiohttp.ClientSession) -> U
 
     diff_info = await get_diff_info(session, stub_info, pypi_info, relevant_version)
     if diff_info is not None:
-        github_repo_path, old_tag, new_tag, diff_url = diff_info
+        *_, diff_url = diff_info
```
hauntsaninja, line 441: diff_info is already a namedtuple, so let's use attribute access in both places instead of futzing with unpacking logic.

scripts/stubsabot.py
Avasam: I just realized you can do `*_, links["Diff"] = diff_info`. The worst of both worlds 😃 (looks neat though)

scripts/stubsabot.py
```diff
@@ -170,8 +170,9 @@ async def release_contains_py_typed(release_to_download: PypiReleaseDownload, *,
         raise AssertionError(f"Unknown package type: {packagetype!r}")
 
 
-async def find_first_release_with_py_typed(pypi_info: PypiInfo, *, session: aiohttp.ClientSession) -> PypiReleaseDownload:
+async def find_first_release_with_py_typed(pypi_info: PypiInfo, *, session: aiohttp.ClientSession) -> PypiReleaseDownload | None:
     release_iter = pypi_info.releases_in_descending_order()
+    first_release_with_py_typed: PypiReleaseDownload | None = None
```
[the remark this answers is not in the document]
hauntsaninja: Alternatively I'd also be fine with moving release_contains_py_typed into here

scripts/stubsabot.py
Avasam: Didn't see it was a named tuple. Yeah that'll be much better, even than teh original code.

scripts/stubsabot.py
```diff
@@ -418,10 +419,12 @@ async def determine_action(stub_path: Path, session: aiohttp.ClientSession) -> U
     if latest_version in spec:
         return NoUpdate(stub_info.distribution, "up to date")
 
+    obsolete_since: PypiReleaseDownload | None = None
     is_obsolete = await release_contains_py_typed(latest_release, session=session)
     if is_obsolete:
```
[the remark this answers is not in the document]
Avasam: Alright, wasn't sure if you'd like to keep the variable name just for readability.

scripts/stubsabot.py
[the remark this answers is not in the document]
Avasam: Something like [`85a2d26` (#9754)](https://github.com/python/typeshed/pull/9754/commits/85a2d266590177dd29b4a87176a71af4fcae7b47) ?

scripts/stubsabot.py
```diff
@@ -170,8 +170,14 @@ async def release_contains_py_typed(release_to_download: PypiReleaseDownload, *,
         raise AssertionError(f"Unknown package type: {packagetype!r}")
 
 
-async def find_first_release_with_py_typed(pypi_info: PypiInfo, *, session: aiohttp.ClientSession) -> PypiReleaseDownload:
+async def find_first_release_with_py_typed(
+    pypi_info: PypiInfo, latest_release: PypiReleaseDownload, *, session: aiohttp.ClientSession
```
Avasam, line 174: Technically `latest_release` can be obtained with `pypi_info.get_latest_release()`. But that would result in calling `get_latest_release` twice because it's also used for an early check.

I leave leave this as-is, change the first parameter to expect the iterator of releases instead, or "cache" `get_latest_release` (for example by making it private and exposing it through a property that calls `_get_latest_release` only once)

Avasam: Maybe it could return a bit less early, and use `release_iter = pypi_info.releases_in_descending_order()`. With the `releases_in_descending_order` moving in `find_first_release_with_py_typed`, building an iterator early is probably much better than a redundant call to pypi

hauntsaninja (approved):
Thanks, this looks good!

scripts/stubsabot.py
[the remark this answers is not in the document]
hauntsaninja: Wait, sorry I read a thing wrong, your change was fine as is!

PR #9755, opened 2023-02-18
Add `object.__subclasshook__`

NeilGirdhar:
Consider:
```python
from dataclasses import dataclass, is_dataclass
from abc import ABC
from typing import Any

class AwesomeDataclassInstance(ABC):
    def __init_subclass__(cls) -> None:
        super().__init_subclass__()
        dataclass(cls)

    @classmethod
    def __subclasshook__(cls, sub: Any, /) -> bool:
        return isinstance(sub, type) and is_dataclass(sub) or super().__subclasshook__(sub)
```
It will fail because `__subclasshook__` is not found.


AlexWaygood (commented):
One the one hand, this is incorrect: neither `abc.ABC` nor `abc.ABCMeta` define `__subclasshook__`. The reason why the `super().__subclasshook__` call works is that it's actually `builtins.object` that defines a `__subclasshook__` method:

```pycon
>>> object.__dict__["__subclasshook__"]
<method '__subclasshook__' of 'object' objects>
>>> object.__subclasshook__(int)
NotImplemented
```

On the other hand, calling `__subclasshook__` makes no sense unless you're doing it on a class that has `ABCMeta` as the metaclass. So I'm not sure I'd want to add it to the stub for `builtins.object`. So maybe this is an acceptable "white lie".

I'm curious what other maintainers think on this. In any event, if we are going to do this, I think a comment in the stub would be warranted, since it isn't _actually_ defined in the `abc.ABC` class at runtime.

NeilGirdhar:
> One the one hand, this is incorrect:

I'm also curious what the experts think.  In my experience, it's better to be accurate (rather than the "white lie" that's here) since you don't know what someone might do.  So, if you agree, I'll move the method to `object`.

On the other hand, why is this on `object` instead of on `ABC`?  Isn't that just polluting `object`?  Should suggest that Python move this method to `ABC`?

srittau:
In my opinion, we should stay as close to the implementation as possible, except if there's a really good reason. I believe we should add `object.__subclasshook__`.

JelleZijlstra:
Agree, we should add `object.__subclasshook__` in typeshed since it exists at runtime.

Possibly it shouldn't exist, but that's a decision to make on the CPython project. I don't understand the implementation of ABCs well enough to understand why the decision was made to add this method to `object`, but maybe there's a good reason.

AlexWaygood:
Sounds good, let's move it to `builtins.object` then.

NeilGirdhar:
Makes sense, done!

stdlib/builtins.pyi
```diff
@@ -110,6 +110,8 @@ class object:
 
     def __dir__(self) -> Iterable[str]: ...
     def __init_subclass__(cls) -> None: ...
+    @classmethod
+    def __subclasshook__(cls, __subclass: Any) -> bool: ...
```
AlexWaygood, line 114: Surely it doesn't make sense to call this method on things that aren't known to be classes?

```suggestion
    def __subclasshook__(cls, __subclass: type) -> bool: ...
```

NeilGirdhar: Done

AlexWaygood (approved):

PR #9758, opened 2023-02-19
Improve the first overload of `is_dataclass`

### LINKED ISSUE #9723: Repair is_dataclass definition?
Consider:
```python
from dataclasses import is_dataclass

def f(X: type) -> None:
    if is_dataclass(X):
        reveal_type(X)  # information: Type of "X" is "type"
```
Ideally, `X` would have type `type[DataclassInstance]`.  It doesn't because `is_dataclass` is defined:
```python
@overload
def is_dataclass(obj: DataclassInstance | type[DataclassInstance]) -> Literal[True]: ...
@overload
def is_dataclass(obj: type) -> TypeGuard[type[DataclassInstance]]: ...
@overload
def is_dataclass(obj: object) -> TypeGuard[DataclassInstance | type[DataclassInstance]]: ...
```
And the argument of type `type` matches the first overload, which does not incorporate a `TypeGuard`.  Eric Traut recommends [erasing the first overload  of `is_dataclass`](https://github.com/microsoft/pyright/issues/4626).

AlexWaygood:
Fixes #9723. I think this is the best compromise for now.


test_cases/stdlib/check_dataclasses.py
```diff
@@ -55,17 +57,23 @@ def test_other_isdataclass_overloads(x: type, y: object) -> None:
...
-        dc.replace(x)  # type: ignore
+
+        # Mypy correctly emits an error on these due to the fact
+        # that it's a dataclass class, not a dataclass instance.
+        # Pyright, however, handles ClassVar members in protocols differently.
+        # See https://github.com/microsoft/pyright/issues/4339
+        #
+        # dc.asdict(x)
+        # dc.astuple(x)
+        # dc.replace(x)
```
AlexWaygood, line 70: We could possibly put these tests inside an `if MYPY` block, and add `"defineConstant": { "MYPY": false } }` to our pyrightconfig file for our test cases, to ensure that pyright doesn't check any code inside `if MYPY` blocks. I'm sort-of wary of doing that, however, as I'd really like to keep typechecker-specific test cases to a minimum in typeshed.

JelleZijlstra (approved):

JelleZijlstra (approved):

PR #9760, opened 2023-02-19
Fix broken stubsabot logic in `find_first_release_with_py_typed`

### LINKED ISSUE #9759: stubsabot: find_first_release_with_py_typed() doesn't work when only the latest release has py.typed
https://github.com/python/typeshed/blob/da9366d0163bb65817216bba13ec4538f0a6e6c0/scripts/stubsabot.py#L173-L180

This function returns `None` if the latest release has `py.typed` and other releases don't. This is because the first `next(release_iter)` consumes the first release and essentially hides it from the rest of this function.

Broken in #9754

AlexWaygood:
Fixes #9759

I tested this locally: with this patch, `DateTimeRange` is correctly marked as obsolete if I run it locally. Without it, stubsabot just tries to bump `DateTimeRange` to the latest version, without noticing that `DateTimeRange` added a `py.typed` file in the latest version, released in the last 24 hours.

### DIFF
--- a/scripts/stubsabot.py
+++ b/scripts/stubsabot.py
@@ -171,7 +171,10 @@ async def release_contains_py_typed(release_to_download: PypiReleaseDownload, *,
 
 
 async def find_first_release_with_py_typed(pypi_info: PypiInfo, *, session: aiohttp.ClientSession) -> PypiReleaseDownload | None:
-    release_iter = pypi_info.releases_in_descending_order()
+    release_iter = (
+        release for release in pypi_info.releases_in_descending_order()
+        if not release.version.is_prerelease
+    )
     latest_release = next(release_iter)
     # If the latest release is not py.typed, assume none are.
     if not (await release_contains_py_typed(latest_release, session=session)):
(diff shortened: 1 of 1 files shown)


scripts/stubsabot.py
```diff
@@ -172,11 +172,12 @@ async def release_contains_py_typed(release_to_download: PypiReleaseDownload, *,
...
 async def find_first_release_with_py_typed(pypi_info: PypiInfo, *, session: aiohttp.ClientSession) -> PypiReleaseDownload | None:
     release_iter = pypi_info.releases_in_descending_order()
+    latest_release = next(release_iter)
     # If the latest release is not py.typed, assume none are.
-    if not (await release_contains_py_typed(release := next(release_iter), session=session)):
+    if not (await release_contains_py_typed(latest_release, session=session)):
         return None
 
-    first_release_with_py_typed: PypiReleaseDownload | None = None
+    first_release_with_py_typed = latest_release
```
Akuli, line 180: if I understand correctly, we should never return pre-releases, even if the latest pre-release version happens to be the only release containing a `py.typed`.

We could filter away pre-releases very early:

```python3
release_iter = (
    release for release in pypi_info.releases_in_descending_order()
    if not release.version.is_prerelease
)
```

AlexWaygood: > if I understand correctly, we should never return pre-releases, even if the latest pre-release version happens to be the only release containing a `py.typed`.

Yes, that's correct. Good catch.

Akuli (approved):

scripts/stubsabot.py
Avasam: Oh I see where I went wrong. Sorry about that.
And nice to see even further improvements.

PR #9761, opened 2023-02-19
Various stdlib dunders: correct parameter names; improve types; add defaults

AlexWaygood:


stdlib/_decimal.pyi
```diff
@@ -89,7 +89,7 @@ class Decimal:
     def __mul__(self, __other: _Decimal) -> Decimal: ...
     def __neg__(self) -> Decimal: ...
     def __pos__(self) -> Decimal: ...
-    def __pow__(self, __other: _Decimal, __modulo: _Decimal | None = ...) -> Decimal: ...
+    def __pow__(self, __other: _Decimal, __mod: _Decimal | None = None) -> Decimal: ...
```
JelleZijlstra, line 92:
```suggestion
    def __pow__(self, __value: _Decimal, __mod: _Decimal | None = None) -> Decimal: ...
```

```
In [7]: decimal.Decimal.__pow__.__text_signature__
Out[7]: '($self, value, mod=None, /)'
```
Note that the signature (and docstring) is the same for *all* C-defined slot functions of the same slot due to how the slot machinery works.

stdlib/_decimal.pyi
```diff
@@ -116,7 +116,7 @@ class Decimal:
     def __floor__(self) -> int: ...
     def __ceil__(self) -> int: ...
     def fma(self, other: _Decimal, third: _Decimal, context: Context | None = None) -> Decimal: ...
-    def __rpow__(self, __other: _Decimal, __context: Context | None = ...) -> Decimal: ...
+    def __rpow__(self, __other: _Decimal, __mod: Context | None = None) -> Decimal: ...
```
JelleZijlstra, line 119:
```suggestion
    def __rpow__(self, __value: _Decimal, __mod: Context | None = None) -> Decimal: ...
```

JelleZijlstra (approved):

PR #9762, opened 2023-02-19
Add `__set__` to functools.cached_property

### LINKED ISSUE #9726: Make `@cached_property` settable?
[The documentation](https://docs.python.org/3/library/functools.html#functools.cached_property) notes:
>The mechanics of [cached_property()](https://docs.python.org/3/library/functools.html#functools.cached_property) are somewhat different from [property()](https://docs.python.org/3/library/functions.html#property). A regular property blocks attribute writes unless a setter is defined. In contrast, a cached_property allows writes.

And mypy has `@cached_property` special-cased to be settable: https://github.com/python/mypy/blob/ad82257170636871a0062ebe6fb82fb45523f580/mypy/semanal.py#L1508-L1509 So, mypy allows assignments to the property even though `__set__` isn't defined in typeshed.

But in pyright it isn't settable, because pyright just follows typeshed. I raised an issue about it: https://github.com/microsoft/pyright/issues/4438 but pyright (somewhat understandably) doesn't want to special-case `@cached_property`.

So, my proposal is now to add `__set__` method to the typestub of `@cached_property`. Something like this:
```python
class cached_property(Generic[_T]):
    # [...]
    def __set__(self, instance: object, value: _T) -> None: ...
```

I have confirmed that at runtime, overwriting a cached property seems to work at any time, even if it hasn't been called yet:
```python
>>> from functools import cached_property
>>> class A:
...   @cached_property
...   def c(self):
...     return 0
... 
>>> a = A()
>>> a.c = 3
>>> a.c
3
>>> a = A()
>>> a.c
0
>>> a.c = 2
>>> a.c
2
```

tmke8:
closes #9726

See discussion in the linked issue.

There shouldn't be any change in mypy_primer, because mypy already special-cases `@cached_property` to be settable.


test_cases/stdlib/check_functools.py
```diff
@@ -0,0 +1,28 @@
...
+    assert_type(A(x=1).x, int)
+
+    class B:
+        @cached_property
+        def x(self) -> int:
+            return 0
+
+    b = B()
+    assert_type(b.x, int)
+    b.x = cast(int, 4)
```
AlexWaygood, line 27: Heh, is the `cast()` because pyright infers the type as being `Literal[4]` otherwise? If so, I prefer this little workaround:

```python
def check_cached_property_settable(x: int) -> None:
    b = B()
    assert_type(b.x, int)
    b.x = x
    assert_type(b.x, int)
```

AlexWaygood: Actually, you might even be able to get away with just this:

```python
b = B()
assert_type(b.x, int)
x: int = 4
b.x = x
assert_type(b.x, int)
```

tmke8: The second one didn't work -- pyright was still narrowing the type. I went with the first one.

AlexWaygood (approved):
LGTM, thanks! I'll leave this open for a few days in case any other maintainers have objections, will merge otherwise :)

srittau (approved):

PR #9763, opened 2023-02-19
lru_cache.cache_info().maxsize can be None for infinite cache size

asvetlov:
Docs are here: https://docs.python.org/3/library/functools.html#functools.lru_cache


AlexWaygood (approved):
Thanks!

asvetlov:
Welcome!

PR #9764, opened 2023-02-19
Additional type hints added to stub openpyxl

mynhardtburger:
This is a continuation of the #9487 stale PR, with the recommended changes implemented.


mynhardtburger:
@Avasam FYI

stubs/openpyxl/openpyxl/worksheet/worksheet.pyi
```diff
@@ -24,106 +33,106 @@ class Worksheet(_WorkbookChild):
...
         self,
-        min_row: Incomplete | None = ...,
-        max_row: Incomplete | None = ...,
-        min_col: Incomplete | None = ...,
-        max_col: Incomplete | None = ...,
+        min_row: int | None = ...,
+        max_row: int | None = ...,
+        min_col: int | None = ...,
+        max_col: int | None = ...,
         values_only: bool = ...,
```
mynhardtburger, line 77: @AlexWaygood I struggled to implement the @overloads. Kept on getting `Overloaded function signatures 1 and 2 overlap with incompatible return types` errors when using `Literal[True]` and `Literal[False]` for the two possible return types.

stubs/openpyxl/openpyxl/worksheet/worksheet.pyi
```diff
@@ -24,106 +33,106 @@ class Worksheet(_WorkbookChild):
...
+    ) -> Generator[Cell | str | float | datetime | None, None, None]: ...
     @property
-    def columns(self): ...
-    def set_printer_settings(self, paper_size, orientation) -> None: ...
-    def add_data_validation(self, data_validation) -> None: ...
-    def add_chart(self, chart, anchor: Incomplete | None = ...) -> None: ...
-    def add_image(self, img, anchor: Incomplete | None = ...) -> None: ...
-    def add_table(self, table) -> None: ...
+    def columns(self) -> Generator[Cell, None, None]: ...
+    def set_printer_settings(self, paper_size: int, orientation: None | Literal["default", "portrait", "landscape"]) -> None: ...
```
AlexWaygood, line 93: I think `paper_size` can also be `None`, if I understand the `openpyxl` descriptor mechanisms correctly: https://foss.heptapod.net/openpyxl/openpyxl/-/blob/branch/3.0/openpyxl/worksheet/page.py#L19

```suggestion
    def set_printer_settings(self, paper_size: int | None, orientation: None | Literal["default", "portrait", "landscape"]) -> None: ...
```

stubs/openpyxl/openpyxl/worksheet/worksheet.pyi
```diff
@@ -24,106 +33,106 @@ class Worksheet(_WorkbookChild):
...
+    def add_data_validation(self, data_validation: DataValidation) -> None: ...
+    def add_chart(self, chart: Incomplete, anchor: Incomplete | None = ...) -> None: ...
+    def add_image(self, img: Incomplete, anchor: Incomplete | None = ...) -> None: ...
```
AlexWaygood, line 96: We prefer to leave the parameters unannotated, if the types are entirely unknown. See  #8472 for some reasoning behind this.

```suggestion
    def add_chart(self, chart, anchor: Incomplete | None = ...) -> None: ...
    def add_image(self, img, anchor: Incomplete | None = ...) -> None: ...
```

stubs/openpyxl/openpyxl/worksheet/worksheet.pyi
```diff
@@ -24,106 +33,106 @@ class Worksheet(_WorkbookChild):
...
+    def add_table(self, table: Table) -> None: ...
     @property
-    def tables(self): ...
-    def add_pivot(self, pivot) -> None: ...
+    def tables(self) -> TableList: ...
+    def add_pivot(self, pivot: Incomplete) -> None: ...
```
AlexWaygood, line 100:
```suggestion
    def add_pivot(self, pivot) -> None: ...
```

stubs/openpyxl/openpyxl/worksheet/worksheet.pyi
```diff
@@ -24,106 +33,106 @@ class Worksheet(_WorkbookChild):
...
-    @property
-    def print_title_rows(self): ...
+    def append(self, iterable: Iterable[Incomplete]) -> None: ...
+    def insert_rows(self, idx: int, amount: int = ...) -> None: ...
+    def insert_cols(self, idx: int, amount: int = ...) -> None: ...
+    def delete_rows(self, idx: int, amount: int = ...) -> None: ...
+    def delete_cols(self, idx: int, amount: int = ...) -> None: ...
+    def move_range(self, cell_range: CellRange | str, rows: int = ..., cols: int = ..., translate: bool = ...) -> None: ...
+    @property
+    def print_title_rows(self) -> str: ...
```
AlexWaygood, line 126: This can also return `None` if this `if` condition isn't met: https://foss.heptapod.net/openpyxl/openpyxl/-/blob/branch/3.0/openpyxl/worksheet/worksheet.py#L831

```suggestion
    def print_title_rows(self) -> str | None: ...
```

stubs/openpyxl/openpyxl/worksheet/worksheet.pyi
```diff
@@ -24,106 +33,106 @@ class Worksheet(_WorkbookChild):
...
     @print_title_rows.setter
-    def print_title_rows(self, rows) -> None: ...
+    def print_title_rows(self, rows: str | None) -> None: ...
     @property
-    def print_title_cols(self): ...
+    def print_title_cols(self) -> str: ...
```
AlexWaygood, line 130: This can return `None` if this `if` condition isn't met: https://foss.heptapod.net/openpyxl/openpyxl/-/blob/branch/3.0/openpyxl/worksheet/worksheet.py#L850

```suggestion
    def print_title_cols(self) -> str | None: ...
```

stubs/openpyxl/openpyxl/worksheet/worksheet.pyi
```diff
@@ -24,106 +33,106 @@ class Worksheet(_WorkbookChild):
...
     @print_title_cols.setter
-    def print_title_cols(self, cols) -> None: ...
+    def print_title_cols(self, cols: str | None) -> None: ...
     @property
-    def print_titles(self): ...
+    def print_titles(self) -> str: ...
```
AlexWaygood, line 134: This can return `None` if `self.print_title_rows` and `self.print_title_cols` are both `None`: https://foss.heptapod.net/openpyxl/openpyxl/-/blob/branch/3.0/openpyxl/worksheet/worksheet.py#L866-871

```suggestion
    def print_titles(self) -> str | None: ...
```

AlexWaygood (commented):
Thanks! This looks pretty good now -- here's a few comments from comparing this to the source code

stubs/openpyxl/openpyxl/worksheet/worksheet.pyi
AlexWaygood: Yes, the overloads are a little tricky to get right here. I took the liberty of pushing to your branch to add them in 8ed83b4

stubs/openpyxl/openpyxl/reader/excel.pyi
```diff
@@ -19,8 +25,14 @@ class ExcelReader:
...
     def read_properties(self) -> None: ...
     def read_theme(self) -> None: ...
-    def read_chartsheet(self, sheet, rel) -> None: ...
+    def read_chartsheet(self, sheet: Chartsheet, rel: Relationship) -> None: ...
     def read_worksheets(self) -> None: ...
     def read(self) -> None: ...
 
-def load_workbook(filename, read_only: bool = ..., keep_vba=..., data_only: bool = ..., keep_links: bool = ...): ...
+def load_workbook(
+    filename: SupportsRead[bytes] | str,
```
AlexWaygood, line 33: As pointed out by the `mypy_primer` diff, it looks like it's also okay to pass in `os.PathLike[str]` objects for the `filename` parameter, since `os.PathLike[str]` objects are accepted by `os.path.splitext`:

```pycon
>>> from pathlib import Path
>>> import os.path
>>> os.path.splitext(Path('foo/bar/baz.pyi'))[-1].lower()
'.pyi'
```

(Argument validation and type coercion for objects passed into `load_workbook` is done by the `_validate_archive` function here, which passes the `filename` argument to `os.path.splitext()` if it doesn't have a `read()` method: https://foss.heptapod.net/openpyxl/openpyxl/-/blob/branch/3.0/openpyxl/reader/excel.py#L67-97.)

We have a `StrPath` alias in the `_typeshed` directory that we use to describe the `str | PathLike[str]` union -- I recommend that you use it here. (You'll have to add `from _typeshed import StrPath` at the top of the file.)

```suggestion
    filename: SupportsRead[bytes] | StrPath,
```

stubs/openpyxl/openpyxl/workbook/workbook.pyi
```diff
@@ -20,54 +29,54 @@ class Workbook:
...
-    def add_named_style(self, style) -> None: ...
+    def add_named_style(self, style: NamedStyle) -> None: ...
     @property
-    def named_styles(self): ...
-    def get_named_ranges(self): ...
-    def add_named_range(self, named_range) -> None: ...
-    def get_named_range(self, name): ...
-    def remove_named_range(self, named_range) -> None: ...
+    def named_styles(self) -> list[str]: ...
+    def get_named_ranges(self) -> DefinedNameList: ...
```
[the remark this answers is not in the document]
mynhardtburger: You are right, it returns a `openpyxl.descriptors.Sequence`. Initially `self.defined_names.definedName` is a `tuple`, but if `DefinedNameList._cleanup()` or `DefinedNameList.append()` methods is called it is turned into a `list`.

Do you suggest a return type of `openpyxl.descriptors.Sequence` or `tuple[DefinedName] | list[DefinedName]`? The latter is more descriptive, but I'm not sure whether it is appropriate specify the underlying types compared the the type which `type()` will return.

stubs/openpyxl/openpyxl/worksheet/worksheet.pyi
mynhardtburger: Is forcing the argument to be a keyword argument the only way around those errors?

stubs/openpyxl/openpyxl/workbook/workbook.pyi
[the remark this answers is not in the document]
AlexWaygood: I think you made the right call, let's stick with the more descriptive types.

stubs/openpyxl/openpyxl/worksheet/worksheet.pyi
AlexWaygood: So, the issue here is that the type signature we'd "like" to write would be something like this. Note that each overload has a different return type:

```python
@overload
def iter_rows(
    self,
    min_row: int | None = None,
    max_row: int | None = None,
    min_col: int | None = None,
    max_col: int | None = None,
    values_only: Literal[True],
) -> Generator[tuple[str | float | datetime | None, ...], None, None]: ...
@overload
def iter_rows(
    self,
    min_row: int | None = ...,
    max_row: int | None = ...,
    min_col: int | None = ...,
    max_col: int | None = ...,
    values_only: Literal[False] = False,
) -> Generator[tuple[Cell, ...], None, None]: ...
@overload
def iter_rows(
    self,
    min_row: int | None = None,
    max_row: int | None = None,
    min_col: int | None = None,
    max_col: int | None = None,
    values_only: bool,
) -> Generator[tuple[Cell | str | float | datetime | None, ...], None, None]: ...
```

That would clearly spell out to the type checker that there are three distinct things that could happen here:
- The user could pass in a value that's known to be `Literal[True]` in, in which case the first overload is selected
- The user could either not pass in a value for that parameter, or pass a value that's known to be `Literal[False]` in, in which case the second overload is selected
- The user could pass in a value that's known to be a `boolean` of some kind, but isn't known to be either `Literal[True]` or `Literal[False]`, in which case the third parameter is selected.

Importantly, the second overload is the only one here that has a default value given for the `values_only` parameter, because we want to make sure that the type checker always selects this overload if the user doesn't pass in a value for this parameter.

_Unfortunately_, two of those^ overloads result in `SyntaxError`s when parsed by Python, because you can't have a parameter with a default value preceding a parameter without a default value in a function signature. That means that we have to expand out the first overload into two overloads, in order to make sure that we're still allowing all valid function calls:

```python
@overload
def iter_rows(
    self,
    min_row: int | None,
    max_row: int | None,
    min_col: int | None,
    max_col: int | None,
    values_only: Literal[True],
) -> Generator[tuple[str | float | datetime | None, ...], None, None]: ...
@overload
def iter_rows(
    self,
    min_row: int | None = None,
    max_row: int | None = None,
    min_col: int | None = None,
    max_col: int | None = None,
    *,
    values_only: Literal[True],
) -> Generator[tuple[str | float | datetime | None, ...], None, None]: ...
```

And we have to do the same with the third overload, which has the same issue. That means that we end up with the five overloads that I added in https://github.com/python/typeshed/commit/8ed83b411545cafe16c5d2c9d0089dd6f12e1f8a

stubs/openpyxl/openpyxl/workbook/workbook.pyi
```diff
@@ -20,54 +30,54 @@ class Workbook:
...
-    def add_named_style(self, style) -> None: ...
+    def add_named_style(self, style: NamedStyle) -> None: ...
     @property
-    def named_styles(self): ...
-    def get_named_ranges(self): ...
-    def add_named_range(self, named_range) -> None: ...
-    def get_named_range(self, name): ...
-    def remove_named_range(self, named_range) -> None: ...
+    def named_styles(self) -> list[str]: ...
+    def get_named_ranges(self) -> list[DefinedName] | tuple[DefinedName]: ...
```
AlexWaygood, line 73:
```suggestion
    def get_named_ranges(self) -> list[DefinedName] | tuple[DefinedName, ...]: ...
```

Remember that `tuple[T]` means "a single-element tuple where the sole element is of type `T`". You want `tuple[T, ...]` if you want to signify "a tuple of arbitrary length, where all elements are of type `T`".

We could also possibly simplify this by just doing `Sequence[DefinedName]` (with `Sequence` imported from `collections.abc`)?

```suggestion
    def get_named_ranges(self) -> Sequence[DefinedName]: ...
```

AlexWaygood (commented):
Just one last point, then we're good to go!

stubs/openpyxl/openpyxl/workbook/workbook.pyi
mynhardtburger: `collections.abc.Sequence` is broader than just `list` and `tuple`. It also includes `str` and `bytes`: [https://docs.python.org/3/glossary.html#term-sequence](https://docs.python.org/3/glossary.html#term-sequence)

I don't think `collections.abc.Sequence` would be appropriate.

AlexWaygood: > `collections.abc.Sequence` is broader than just `list` and `tuple`. It also includes `str` and `bytes`: https://docs.python.org/3/glossary.html#term-sequence

`str` and `bytes` are indeed both sequences, correct — but neither is a subtype of `Sequence[DefinedName]`. `str` is a subtype of `Sequence[str]`, and `bytes` is a subtype of `Sequence[int]`. So if I were writing this PR, I would favour the simpler approach here :)

But your approach isn't wrong by any means, and I don't want to be too picky :)

AlexWaygood (approved):
Thanks @ArnabRollin for the initial PR, and thanks @mynhardtburger for finishing it! 🥳

stubs/openpyxl/openpyxl/reader/excel.pyi
Avasam: This, and `ExcelReader.__init__` is ultimately passed to `ZipFile` in `_validate_archive`.

After double-checking, ZipFile could use some overloads. It only needs a name in "read" mode.

stubs/openpyxl/openpyxl/workbook/workbook.pyi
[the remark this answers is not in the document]
Avasam: 2 things here I just noticed (sorry it's been a while that I've looked at openpyxl)
1: descriptors are not meant to be used directly for type hints. Descriptors are assigned to ClassVars of classes subtyping Serializable. Serializable members act as properties where the setter uses the descriptor for runtime vlaidation of the setter
2. Whilst the `Sequence` descriptor allows both list and tuple (https://foss.heptapod.net/openpyxl/openpyxl/-/blob/branch/3.1/openpyxl/descriptors/sequence.py#L18), `DefinedNameList.definedName` specifically needs to be sliceable and appendable (at least in 3.0, this doesn't seem to be the case anymore for 3.1) https://foss.heptapod.net/openpyxl/openpyxl/-/blob/branch/3.0/openpyxl/workbook/defined_name.py, the latter meaning it can't be a tuple.

stubs/openpyxl/openpyxl/workbook/workbook.pyi
Avasam: 2 things here I just noticed (sorry it's been a while that I've looked at openpyxl)
1: descriptors are not meant to be used directly for type hints. Descriptors are assigned to ClassVars of classes subtyping Serializable. Serializable members act as properties where the setter uses the descriptor for runtime vlaidation of the setter
2. Whilst the `Sequence` descriptor allows both list and tuple (https://foss.heptapod.net/openpyxl/openpyxl/-/blob/branch/3.1/openpyxl/descriptors/sequence.py#L18), `DefinedNameList.definedName` specifically needs to be sliceable and appendable (at least in 3.0, this doesn't seem to be the case anymore for 3.1) https://foss.heptapod.net/openpyxl/openpyxl/-/blob/branch/3.0/openpyxl/workbook/defined_name.py#L207-209, the latter meaning it can't be a tuple.

stubs/openpyxl/openpyxl/worksheet/worksheet.pyi
```diff
@@ -24,106 +34,166 @@ class Worksheet(_WorkbookChild):
...
+        max_col: int | None = None,
+        min_row: int | None = None,
+        max_row: int | None = None,
+        *,
+        values_only: bool,
+    ) -> Generator[tuple[Cell | str | float | datetime | None, ...], None, None]: ...
+    @property
+    def columns(self) -> Generator[Cell, None, None]: ...
+    def set_printer_settings(
+        self, paper_size: int | None, orientation: None | Literal["default", "portrait", "landscape"]
```
Avasam, line 153: Another case for https://github.com/PyCQA/flake8-pyi/issues/175 and https://github.com/PyCQA/flake8-pyi/issues/283 -like checks in Flake8-PYI.

PR #9774, opened 2023-02-20
Add defaults to csv stubs

AlexWaygood:
I prepared this PR manually using the docs at https://docs.python.org/3/library/csv.html#dialects-and-formatting-parameters and REPL experimentation


srittau (approved):
LGTM, though I didn't compare against implementation.

PR #9779, opened 2023-02-20
Change `ignore_missing_stub` default to `false`

### LINKED ISSUE #9538: Change `ignore_missing_stub` default to `false`?
I've been wondering about this.

Possible reasons:
1. More and more stubs are "stubtest" complete (not counting pyright and mypy)
	1.1. It should be relatively easy to "minimally complete" missing ones using stugben. Especially after the reduction of stubtest false-positives by https://github.com/python/mypy/pull/14270
	1.2. Eventually most if not all stubs will have `ignore_missing_stub = false`, which could be the only entry under `[tool.stubtest]`
2. It's easier to search for `ignore_missing_stub = true` than to search for a missing line.
3. `None` would mean `False` in the metadata parser. (feels more intuitive?)
4. #8955
    > I don't want to encourage new incomplete stubs
    
Cons:
- Have to update the metadata parser
- A commit that flips (adds/removes) the value in all metadata.toml
- Maybe wait until there's even less (if not all) "stubtest incomplete" stubs left.

Avasam:
Closes #9538 which got lots of thumbs up, and hopefully I'm not misinterpreting @hauntsaninja 's comment in https://github.com/python/mypy/issues/14692#issuecomment-1428479005 saying he advocated against `--ignore-missing-stub` by default.

I wanted to wait until after mypy 1.0 (with its many stubtest fixes) for this. Should help promote complete stubs.

CC @AlexWaygood because this affects typeshed-stats (unless it reuses `parse_metadata.py`, idk how you implemented it)

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -170,16 +170,17 @@ By default, stubtest emits an error if a public object is present at runtime
 but missing from the stub. However, this behaviour can be disabled using the
 `--ignore-missing-stub` option.
 
-Many third-party stubs packages in typeshed are currently incomplete, and so by
-default, `stubtest_third_party.py` runs stubtest with the
+Most third-party stubs packages in typeshed are currently complete (as far as stubtest is
+concerned), and so by default, `stubtest_third_party.py` does not run stubtest with the
 `--ignore-missing-stub` option to test our third-party stubs. However, this
-option is not used if the distribution has `ignore_missing_stub = false` in the
+option is used if the distribution has `ignore_missing_stub = true` in the
 `tool.stubtest` section of its `tests/METADATA.toml` file. This setting
-indicates that the package is considered "complete", for example:
-https://github.com/python/typeshed/blob/6950c3237065e6e2a9b64810765fec716252d52a/stubs/emoji/METADATA.toml#L3-L4
+indicates that the package is considered "incomplete", for example:
+<!-- TODO: Change to permalink once this is merged -->
+<https://github.com/python/typeshed/blob/main/stubs/setuptools/METADATA.toml#L5>
 
-You can help make typeshed's stubs more complete by adding
-`ignore_missing_stub = false` to the `tests/METADATA.toml` file for a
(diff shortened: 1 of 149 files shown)


tests/README.md
```diff
@@ -170,16 +170,17 @@ By default, stubtest emits an error if a public object is present at runtime
 but missing from the stub. However, this behaviour can be disabled using the
 `--ignore-missing-stub` option.
 
-Many third-party stubs packages in typeshed are currently incomplete, and so by
-default, `stubtest_third_party.py` runs stubtest with the
+Most third-party stubs packages in typeshed are currently complete (as far as stubtest is
```
hauntsaninja, line 173: Github won't let me leave a suggestion, but I'd replace this paragraph with something like:

> If a distribution has `ignore_missing_stub = true` in the
`tool.stubtest` section of its `tests/METADATA.toml` file, `stubtest_third_party.py` will test that distribution with the `--ignore-missing-stub` option. This indicates that the stubs for this distribution are considered "incomplete".

hauntsaninja (commented):
Thanks!

AlexWaygood:
> CC @AlexWaygood because this affects typeshed-stats (unless it reuses `parse_metadata.py`, idk how you implemented it)

Yeah it'll break typeshed-stats, but it's not hard to unbreak :) the implementation's here, if you're curious: https://github.com/AlexWaygood/typeshed-stats/blob/main/src/typeshed_stats/gather.py#L406

tests/README.md
Avasam: A lot less verbose, I like it. Would you keep the example?

hauntsaninja: Nah

AlexWaygood (approved):
Woohoo!

AlexWaygood:
Fixed the `typeshed-stats` logic in https://github.com/AlexWaygood/typeshed-stats/commit/5689d242ce7cdc67b1211d4a3e25f0dae32c2a4a :)

Issue #5725, opened 2021-07-04
aiofiles: looping over file doesn't work


Akuli:
```python
import asyncio
from pathlib import Path

import aiofiles

FILENAME = Path.home() / ".zshrc"

async def main() -> None:
    async with aiofiles.open(FILENAME, "r") as input_fd:
        async for line in input_fd:
            print(line)

if __name__ == '__main__':
    asyncio.run(main())
```

```
scratch.py:10: error: "Coroutine[Any, Any, Iterator[str]]" has no attribute "__anext__"
Found 1 error in 1 file (checked 1 source file)
```

Originally reported at https://github.com/Tinche/aiofiles/issues/105

JelleZijlstra:
I think #5724 fixed this, can you verify?

Akuli:
Still happens:

```
(e39) akuli@akuli-desktop:/tmp$ mypy a.py
a.py:10: error: "Coroutine[Any, Any, Any]" has no attribute "__anext__"
Found 1 error in 1 file (checked 1 source file)
(e39) akuli@akuli-desktop:/tmp$ pip freeze
mypy==0.910
mypy-extensions==0.4.3
toml==0.10.2
types-aiofiles==0.1.6
typing-extensions==3.10.0.0
(e39) akuli@akuli-desktop:/tmp$ python3 --version
Python 3.9.1
```

PR #9780, opened 2023-02-20
Reverts aiofiles back to _typeshed Self

erwinkinn:
Hello,all.
This PR reverts those changes @AlexWaygood made about five days ago. After that I'd started getting this mypy error:
```
error: Self? has no attribute "__anext__"  [attr-defined]
```
It's quite similar to [#5725](https://github.com/python/typeshed/issues/5725), but the problem is typing_extensions.Self was not made to support an async iterator protocol.
So I've tried to rollback to an abstract type `Self` from `_typeshed` making a hot-hot fix just for now.

I'm not sure, maybe it's better to create an issue at typing_extensions for the occasion...

### DIFF
--- a/stubs/aiofiles/aiofiles/base.pyi
+++ b/stubs/aiofiles/aiofiles/base.pyi
@@ -1,7 +1,7 @@
+from _typeshed import Self
 from collections.abc import Coroutine, Generator, Iterator
 from types import CodeType, FrameType, TracebackType, coroutine
 from typing import Any, Generic, TypeVar
-from typing_extensions import Self
 
 _T = TypeVar("_T")
 _T_co = TypeVar("_T_co", covariant=True)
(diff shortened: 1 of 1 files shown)


Avasam:
FWIW, the changes in #9694 added type errors in typeshed's own code, see comment: https://github.com/python/typeshed/pull/9694/files#r1109192407 about usage of `int.__new__(...)`

erwinkinn:
Yes, may be u r right. I'm under my way making an issue at typing_extensions at the moment. 
I have planned to reference this merge request.

erwinkinn:
image
Oh, thanks to bot )) However, it's still on my screen in VS Code (`mypy` works as a plugin)

erwinkinn:
@Avasam, sorry, I'm not sure I did understand it right. Did you mean a larger discussion about `typing_extensions.Self` support within the `typeshed` project? I thought it was about an async iterator protocol support for `typing_extensions.Self` :) Anyway... Discussion is needed obviously )

PR #9783, opened 2023-02-20
Attempt to clarify the ByteString situation

hauntsaninja:
See https://github.com/python/cpython/issues/102092


srittau:
+1 at least for removing the subclass dependency to match runtime. (Even if the abc is registered at runtime.)

tests/stubtest_allowlists/py3_common.txt
```diff
@@ -608,6 +608,11 @@ xml.etree.ElementTree.Element.__iter__
 xml.etree.cElementTree.Element.__iter__
 typing.IO.__iter__  # See https://github.com/python/typeshed/commit/97bc450acd60c1bcdafef3ce8fbe3b95a9c0cac3
 
+# Pretend typing.ByteString is a Union, to better match its documented semantics.
+# As a side effect, this changes the definition of collections.abc.ByteString, which is okay,
+# because it's not an ABC that makes any sense.
```
hauntsaninja, line 613:
```suggestion
# Pretend typing.ByteString is a Union, to better match its documented semantics.
# As a side effect, this changes the definition of collections.abc.ByteString, which is okay,
# because it's not an ABC that makes any sense and was deprecated in 3.12
```

AlexWaygood (approved):

AlexWaygood (approved):

AlexWaygood (approved):

AlexWaygood:
For the record, I only clicked "approve" once; not sure how that happened!

JelleZijlstra (approved):

JelleZijlstra:
This still marked as draft, @hauntsaninja shall we land it?

PR #9787, opened 2023-02-21
Enable some more mypy lints when checking our own code

AlexWaygood:


tests/typecheck_typeshed.py
```diff
@@ -57,15 +57,19 @@ def run_mypy_as_subprocess(directory: str, platform: str, version: str) -> Retur
...
+        "--enable-error-code",
+        "possibly-undefined",
+        "--enable-error-code",
+        "redundant-expr",
     ]
     if directory == "tests" and platform == "win32":
         command.extend(["--exclude", "tests/pytype_test.py"])
-    result = subprocess.run(command, capture_output=True)
+    result = subprocess.run(command, capture_output=True, text=True)
     stdout, stderr = result.stdout, result.stderr
```
JelleZijlstra, line 68: Nittiest of nits (so feel free to just merge when CI is green) but putting these in separate variables feels unnecessarily complicated.

JelleZijlstra (approved):

tests/typecheck_typeshed.py
AlexWaygood: The nits are out in full force tonight 😄

PR #9793, opened 2023-02-21
Typecheck typeshed's code with pyright

Avasam:
When it comes to typeshed's own code, it is tested with mypy but not with pyright. A bit odd given how strictly the stubs are tested against pyright. Making typeshed, imo, a bit of a "poorly shod shoemaker" on that aspect.
Pyright specifically complements mypy's missing features well.

But this is not just about typeshed itself. As it is a pillar in Python's typing ecosystem, problems caught by it, and improvements made thanks to it (especially those pushing the limitations of the current type system) have bigger repercussions on the ecosystem as a whole.

Tangible improvements:
All of the PRs and issues mentioned here are directly related, and opened due, to me trying to run pyright on typeshed's own code and working on this PR:
- Some new type issues or can be caught sooner (see comment https://github.com/python/typeshed/pull/9694/files#r1109192407)
- More chance of a pyright bug or limitation being reported by a typeshed maintainer
  - https://github.com/microsoft/pyright/issues/4483
  - https://github.com/microsoft/pyright/issues/4259
  - https://github.com/microsoft/pyright/issues/4243
  - https://github.com/microsoft/pyright/issues/4244
  - https://github.com/microsoft/pyright/issues/4668
- Improve typeshed's own type-safety
  - See `.py` changes in this PR
  - The need and use for a centralized matadata file parser is something I had mentioned not too long before AlexWaygood started on it https://github.com/python/typeshed/pull/9532#issuecomment-1382754615
  - https://github.com/python/typeshed/pull/9299
  - https://github.com/python/typeshed/pull/9293
  - https://github.com/python/typeshed/pull/9714
  - https://github.com/python/typeshed/pull/9397
  - https://github.com/python/typeshed/issues/9155
  - https://github.com/python/typeshed/pull/9549
  - https://github.com/python/typeshed/pull/9435
  - https://github.com/python/typeshed/pull/9754
- Spot missing and problematic types in libraries used by typeshed. Which are mostly quite common libraries that could really serve being fully typed
  - https://github.com/python/typeshed/pull/9747
  - https://github.com/python/typeshed/pull/9752
  - https://github.com/hadialqattan/pycln/pull/186
  - https://github.com/cpburnz/python-pathspec/pull/67
  - https://github.com/cpburnz/python-pathspec/pull/71
  - https://github.com/cpburnz/python-pathspec/pull/75
  - https://github.com/google/pytype/issues/1325
  - https://github.com/psf/black/issues/3399
  - https://github.com/sdispater/tomlkit/pull/272
  - https://github.com/python/typeshed/issues/2886#issuecomment-1437577504
- And possibly more

(I waited until most of these got resolved before bringing this PR)


stubs/setuptools/pkg_resources/__init__.pyi
```diff
@@ -4,12 +4,14 @@ import zipimport
 from _typeshed import Incomplete
 from abc import ABCMeta
 from collections.abc import Callable, Generator, Iterable, Sequence
-from typing import IO, Any, TypeVar, overload
-from typing_extensions import Self, TypeAlias
+from re import Pattern
```
JelleZijlstra, line 7: Do these changes really belong in this PR?

tests/check_consistent.py
```diff
@@ -140,11 +141,21 @@ def get_txt_requirements() -> dict[str, SpecifierSet]:
...
+    repos: list[PreCommitConfigRepos]
+
+
 def get_precommit_requirements() -> dict[str, SpecifierSet]:
     with open(".pre-commit-config.yaml", encoding="UTF-8") as precommit_file:
         precommit = precommit_file.read()
-    yam = yaml.load(precommit, Loader=yaml.Loader)
-    precommit_requirements = {}
+    yam: PreCommitConfig = yaml.load(precommit, Loader=yaml.Loader)
+    precommit_requirements = dict[str, SpecifierSet]()
```
JelleZijlstra, line 158:
```suggestion
    precommit_requirements: dict[str, SpecifierSet] = {}
```
Nit, but would prefer this as it separates the types more explicitly from the runtime.

tests/mypy_test.py
```diff
@@ -57,6 +57,7 @@
 Platform: TypeAlias = Annotated[str, "Must be one of the entries in SUPPORTED_PLATFORMS"]
 
 
+@dataclass
```
JelleZijlstra, line 60: Does this really need to be a dataclass? Making something both an argparse Namespace and a dataclass seems confusing.

tests/utils.py
```diff
@@ -72,7 +72,7 @@ def make_venv(venv_dir: Path) -> VenvInfo:
         if "ensurepip" in e.cmd and b"KeyboardInterrupt" not in e.stdout.splitlines():
             print_error(
                 "stubtest requires a Python installation with ensurepip. "
-                "If on Linux, you may need to install the python3-venv package."
+                + "If on Linux, you may need to install the python3-venv package."
```
JelleZijlstra, line 75: Maybe we should just turn off the pyright rule against implicit string concatenation?

scripts/stubsabot.py
```diff
@@ -32,7 +32,9 @@
 
 class ActionLevel(enum.IntEnum):
     def __new__(cls, value: int, doc: str) -> Self:
-        member = int.__new__(cls, value)
+        # Fails type-checking since updating "Self" types in #9694
+        # See: https://github.com/microsoft/pyright/issues/4668
+        member: Self = int.__new__(cls, value)  # pyright: ignore[reportGeneralTypeIssues]
```
AlexWaygood, line 37: No longer necessary now the pyright bug's been fixed

```suggestion
        member = int.__new__(cls, value)
```

scripts/stubsabot.py
```diff
@@ -581,7 +583,12 @@ def get_update_pr_body(update: Update, metadata: dict[str, Any]) -> str:
     if update.diff_analysis is not None:
         body += f"\n\n{update.diff_analysis}"
 
-    stubtest_will_run = not metadata.get("tool", {}).get("stubtest", {}).get("skip", False)
+    stubtest_will_run = (
+        not metadata.get("tool", {}).get("stubtest", {})
+        # Loss of type due to infered [dict[Unknown, Unknown]]
+        # scripts/stubsabot.py can't import tests/parse_metadata
```
AlexWaygood, line 589: Formatting nit

```suggestion
    # Loss of type due to infered [dict[Unknown, Unknown]]
    # scripts/stubsabot.py can't import tests/parse_metadata
    stubtest_will_run = (
        not metadata.get("tool", {}).get("stubtest", {})
```

tests/check_new_syntax.py
```diff
@@ -22,7 +22,7 @@ def visit_If(self, node: ast.If) -> None:
                 new_syntax = "if " + ast.unparse(node.test).replace("<", ">=", 1)
                 errors.append(
                     f"{path}:{node.lineno}: When using if/else with sys.version_info, "
-                    f"put the code for new Python versions first, e.g. `{new_syntax}`"
+                    + f"put the code for new Python versions first, e.g. `{new_syntax}`"
```
AlexWaygood, line 25: I agree with @JelleZijlstra: I really don't like this pyright lint against implicit string concatenation. Implicit string concatenation can lead to bugs in some situations, but it's often pretty idiomatic. The pyright lint seems like it has lots of false positives that forbid idiomatic Python.

tests/mypy_test.py
```diff
@@ -209,7 +210,8 @@ def add_configuration(configurations: list[MypyDistConf], distribution: str) ->
     with Path("stubs", distribution, "METADATA.toml").open("rb") as f:
         data = tomli.load(f)
 
-    mypy_tests_conf = data.get("mypy-tests")
+    # TODO: This could be added to parse_metadata.py, but is currently unused
+    mypy_tests_conf: dict[str, dict[str, dict[str, dict[str, Any]]]] = data.get("mypy-tests", {})
```
AlexWaygood, line 214: I'd just go with a simpler, less verbose type hint for now, tbh, since this is currently unused

```suggestion
    mypy_tests_conf: dict[str, Any] = data.get("mypy-tests", {})
```

tests/parse_metadata.py
```diff
@@ -151,7 +154,7 @@ def read_metadata(distribution: str) -> StubMetadata:
     given in the `requires` field, for example.
     """
     with Path("stubs", distribution, "METADATA.toml").open("rb") as f:
-        data: dict[str, object] = tomli.load(f)
+        data: dict[str, Any] = tomli.load(f)
```
AlexWaygood, line 157: Why is this change required? I deliberately used `dict[str, object]` when I wrote this module so that we'd have very strict type checking in this function. This makes the code much less type-safe.

tests/parse_metadata.py
```diff
@@ -188,7 +191,8 @@ def read_metadata(distribution: str) -> StubMetadata:
     uploaded_to_pypi = data.get("upload", True)
     assert type(uploaded_to_pypi) is bool
 
-    tools_settings = data.get("tool", {})
+    empty_tools: dict[str, set[str]] = {}
```
AlexWaygood, line 194:
```suggestion
    empty_tools: dict[str, dict[str, object]] = {}
```

tests/pytype_test.py
```diff
@@ -19,11 +19,16 @@
...
 import pkg_resources
-from pytype import config as pytype_config, load_pytd  # type: ignore[import]
-from pytype.imports import typeshed  # type: ignore[import]
 
 from parse_metadata import read_dependencies
 
+assert sys.platform != "win32"
+# Lack of pytype typing
+# pyright: reportUnknownVariableType=false, reportUnknownMemberType=false, reportUnknownArgumentType=false, reportMissingTypeStubs=false
+# pytype is not py.typed https://github.com/google/pytype/issues/1325
```
AlexWaygood, line 28: Could these pyright directives go at the top of the file?

pyrightconfig.typeshed.json
```diff
@@ -0,0 +1,23 @@
+{
```
AlexWaygood, line 1: I'd prefer it if this config file were called something like `pyrightconfig.scripts_and_tests.json` — `pyrightconfig.typeshed.json` doesn't feel very descriptive. The whole project's typeshed! 😀

tests/parse_metadata.py
AlexWaygood: If necessary, I'd prefer a `cast()` here if it lets us avoid `Any` for the value type:

```suggestion
        data = cast(dict[str, object], tomli.load(f))
```

stubs/setuptools/pkg_resources/__init__.pyi
Avasam: Nope, it's meant to be split off in #9747

pyrightconfig.typeshed.json
Avasam: Fair point! Will rename

tests/check_consistent.py
Avasam: Same, I thought I had changed them all (except `set`) to separate typing and assignment. Missed that one thanks!

tests/check_new_syntax.py
Avasam: I can turn it off. I preferred turning everything on and getting your comments :)

tests/mypy_test.py
Avasam: I don't think so? I may have tried both approaches and accidentally left both in >.<

AlexWaygood: Those are all false positives in this case, since `argparse` uses `setattr()` to initialise all of the instance variables on the namespace object passed to it. I suppose pyright could theoretically add special casing to `argparse.Namespace` objects to prevent this error being emitted in this kind of situation, but I'll wager Eric probably won't be enthusiastic about that idea...

tests/parse_metadata.py
Avasam:
```
E:\Users\Avasam\Documents\Git\typeshed\tests\parse_metadata.py
  E:\Users\Avasam\Documents\Git\typeshed\tests\parse_metadata.py:199:29 - error: Type of "get" is partially unknown
    Type of "get" is "Overload[(__key: Unknown, /) -> (Unknown | None), (__key: Unknown, __default: Unknown | _T@get, /) -> (Unknown | _T@get)]" (reportUnknownMemberType)
  E:\Users\Avasam\Documents\Git\typeshed\tests\parse_metadata.py:199:29 - error: Type of "get" is partially unknown
    Type of "get" is "Overload[(__key: Unknown, /) -> (Unknown | None), (__key: Unknown, __default: Unknown | _T@get, /) -> (Unknown | _T@get)] | Overload[(__key: str, /) -> (set[str] | None), (__key: str, 
__default: set[str] | _T@get, /) -> (set[str] | _T@get)]" (reportUnknownMemberType)
  E:\Users\Avasam\Documents\Git\typeshed\tests\parse_metadata.py:12:20 - error: Import "Any" is not accessed (reportUnusedImport)
4 errors, 0 warnings, 0 informations
```

Basically at line 195 `tools_settings` is `object | dict[str, set[str]]` and `isinstance(tools_settings, dict)` narrows `object` to `dict[Unknown, Unknown]`.
So the type becomes `dict[Unknown, Unknown] | dict[str, set[str]]`

In this case a cast doesn't help.

I could fix it with a typeddict like I did in check_consistent. Or a custom type-guard. Or add `reportUnknownMemberType` witht eh suppressions up to, but I don't like that I had to add those in the first place.

AlexWaygood: I'd _really_ rather not introduce `Any` here; it'll mean mypy will verify much less of this function and make the code much less type-safe imo. So I guess I'd advocate adding `reportUnknownMembeType=false` to the suppressions at the top of the file.

tests/mypy_test.py
Avasam: In that case, I'd argue "meaning it's used like a dataclass". (though I don't doubt I could be wrong). Could do `@dataclass(init=False, repr=False, eq=False)` if we don't need/want extra methods.

tests/parse_metadata.py
Avasam: I'll do that. I'm fine with it because, as the comment says. The whole point of this file is to abstract away those errors. So we expect a bunch of `Unknown` and partially `Unknown` because we're checking the content of files that have not been validated yet.

tests/mypy_test.py
AlexWaygood: It's probably simpler just to decorate it with `dataclass(init=False)` and remove the inheritance from `argparse.Namespace` (which isn't necessary for argparse, it just gives us nicer `__repr__` and `__eq__` methods -- if we're using dataclasses, there's no need for the inheritance here).

pyrightconfig.testcases.json
```diff
@@ -30,4 +29,6 @@
     "reportUnnecessaryIsInstance": "none",
     // The name of the self/cls parameter is out of typeshed's control.
     "reportSelfClsParameterName": "none",
+    // More of a lint. Unwanted.
+    "reportImplicitStringConcatenation": "none",
```
AlexWaygood, line 33: I'd actually be okay with keeping this enabled for the `stubs/`, `stdlib/` and `test_cases/` directories. I can't see a valid use case for implicit string concatenation in a `.pyi` file or a test case, so the concerns about annoying false positives don't apply; and it could catch some bugs.

tests/mypy_test.py
```diff
@@ -221,7 +223,7 @@ def add_configuration(configurations: list[MypyDistConf], distribution: str) ->
         assert module_name is not None, f"{section_name} should have a module_name key"
         assert isinstance(module_name, str), f"{section_name} should be a key-value pair"
 
-        values = mypy_section.get("values")
+        values = mypy_section.get("values", {})
         assert values is not None, f"{section_name} should have a values section"
         assert isinstance(values, dict), "values should be a section"
```
AlexWaygood, line 228: This change makes the following two assertions redundant. I'm not sure I understand why pyright doesn't like this code, but would it prefer this instead?

```suggestion
        assert "values" in mypy_section, f"{section_name} should have a values section"
        values = mypy_section["values"]
        assert isinstance(values, dict), "values should be a section"
```

Avasam: With a tweak, but yes! Now it's keeping the explicit `dict[str, dict[str, Any]]` on values, rather than narrowing it down to `Any` then `dict[Unknown, Unknown]`

requirements-tests.txt
```diff
@@ -15,6 +15,6 @@ pyyaml==6.0
 termcolor>=2
 tomli==2.0.1
 tomlkit==0.11.6
-types-pyyaml
+types-pyyaml>=6.0.12.7
```
Avasam, line 18: I can do a single PR for this if you prefer. As a followup to https://github.com/python/typeshed/pull/9752

AlexWaygood (approved):

requirements-tests.txt
AlexWaygood: Fine to do it as part of this PR imo.

tests/mypy_test.py
```diff
@@ -531,7 +533,7 @@ def test_typeshed(code: int, args: TestConfig, tempdir: Path) -> TestResults:
 
 
 def main() -> None:
-    args = parser.parse_args(namespace=CommandLineArgs())
+    args = parser.parse_args(namespace=CommandLineArgs)
```
JelleZijlstra, line 536: What is the motivation for this change? Makes more sense to pass an instance here.

Avasam: Indeed this is no longer necessary with `@dataclass(init=False)`

Avasam:
Thanks for the reviews and patience! This started absolutely huge that I split off in its own chunks over time. And there was some artefacts of that leftover.
I self-reviewed initially, but I guess I should've done it a few more times along the way. 😅

JelleZijlstra (approved):

tests/mypy_test.py
AlexWaygood: I filed the bug report with pyright as I figured Eric might be interested -- his response indicates that the `reportUninitializedInstanceVariable` diagnostic is basically unmaintained and may be removed altogether at some point `¯\_(ツ)_/¯`

PR #9795, opened 2023-02-22
Remove most of `setuptools._distutils`

### LINKED ISSUE #9520: Should we delete the `_distutils` directory from our `setuptools` stubs ?

Avasam:
Just like regular implementation details, private or vendored modules, we only keep what's used in our stubs. Because it's easier to reference and find the source if the structure is the same.
I *could* have imported from `distutils` to remove `setuptools._distutils` completely. But that's not a good idea for two reasons imo:
1. Eventually `stdlib/distutils` won't exist anymore and that'll cause issues.
2. May as well use the version `setuptools` vendors, with the correct types, that we can test with stubtest.

Closes #9520
Also resolves overuse of `Any` in the leftover `setuptools._distutils` classes for #9550

I added an extra_description which I believe can be very valuable to users of `types-setuptools` who may not be aware of `distutils`'s situation. Feel free to wordsmith.
Preview:
image


stubs/setuptools/METADATA.toml
```diff
@@ -1,5 +1,14 @@
 version = "67.3.*"
 requires = ["types-docutils"]
```
AlexWaygood, line 2: We can get rid of this dependency, I think, if we go ahead with this :)

```suggestion
```

Avasam:
apprise is a types-babel issue. babel and mypy are the same issue: They could be importing from either distutils or setuptools.
It used to be fine because types-setuptools classes used to (incorrectly) subclass distutils classes. But now it's correctly showing that the types have no direct relation.

mypy is fixable, it's voluntarily importing `setuptools` for its monkey patching. I think it can just change the return type of `get_extension` https://github.com/python/mypy/blob/master/mypyc/build.py#L56

As for apprise, well, babel doesn't directly depend on setuptools, so we can't just update the type there. https://github.com/python-babel/babel/blob/master/babel/messages/frontend.py#L41
Unless you consider that having installed babel means you've had to have setuptools installed because of https://github.com/python-babel/babel/blob/master/setup.py#L4

stubs/setuptools/METADATA.toml
```diff
@@ -1,5 +1,13 @@
...
-requires = ["types-docutils"]
+extra_description = """\
+Having [setuptools](https://pypi.org/project/setuptools/) installed results in incorrect \
+[distutils](https://docs.python.org/3/distutils/introduction.html) typing. \
+`types-setuptools` makes no attempt at correcting this issue. As of \
+`setuptools>=60.0.0`, you [should not be using `distutils` directly](\
+https://setuptools.pypa.io/en/latest/deprecated/distutils-legacy.html). \
+`distutils` is deprecated as of Python 3.10 and will be fully removed in Python 3.12. \
+(see [PEP 632](https://peps.python.org/pep-0632/))
+"""
```
AlexWaygood, line 10: I'm not sure that this is necessary -- it sort of feels like it's `setuptools`'s problem rather than ours. If we _do_ want some kind of a note here, I'd prefer it if it were slightly more concise.

AlexWaygood (approved):
LGTM (other than my one comment). Will leave open for a bit in case any other maintainers object, though.

Avasam:
Nice timing https://github.com/python/typeshed/actions/runs/4245809415/jobs/7381814113

stubs/setuptools/METADATA.toml
Avasam: Does this look better to you?
> Having [setuptools](https://pypi.org/project/setuptools/) installed results in incorrect [distutils](https://docs.python.org/3/distutils/introduction.html) typing. `types-setuptools` makes no attempt at correcting this issue. See [Porting from Distutils](https://setuptools.pypa.io/en/latest/deprecated/distutils-legacy.html) and [PEP 632 – Deprecate distutils module](https://peps.python.org/pep-0632/).

AlexWaygood: Hmm, the more I think about it, the more I lean towards not including this note at all. It feels like unnecessary finger-pointing at `setuptools` from us. The `distutils` situation is very frustrating for me, a typeshed maintainer, but as far as I know we haven't received any complaints from users of our stubs about incorrect `distutils` typing, so it sort of feels like this note is attempting to solve something that may not _actually_ be much of an issue for users. If we do get complaints or queries from users, we can maybe rethink this, but for now I think we can _probably_ do without this :)

AlexWaygood:
> ```diff
> mypy (https://github.com/python/mypy)
> + mypyc/build.py:64: error: Incompatible import of "Extension" (imported name has type "Type[setuptools.extension.Extension]", local name has type "Type[distutils.extension.Extension]")  [assignment]
> ```

I think you're correct that this one is easy to fix, but it _is_ going to cause mypy's CI to suddenly and unexpectedly start failing once this PR is merged, since they [don't pin `types-setuptools` in CI](https://github.com/python/mypy/blob/284142d6e29d4588107f5c68f95c43986a2e26e5/build-requirements.txt#L4). Either you or I should probably go fix it as soon as this PR is merged :)

stubs/setuptools/METADATA.toml
Avasam: I can see what you mean. I also expect the amount of users who'll actually land on the PyPI page to be taught about the issue of mixing setuptools and distutils together to be really small. (pyright ships stubs, and the average mypy user probably installs stubs from mypy's CLI suggestion).

stubs/setuptools/setuptools/_distutils/dist.pyi
```diff
@@ -53,7 +54,7 @@ class DistributionMetadata:
 class Distribution:
     cmdclass: dict[str, type[Command]]
     metadata: DistributionMetadata
-    def __init__(self, attrs: Mapping[str, Any] | None = ...) -> None: ...
+    def __init__(self, attrs: Mapping[str, object] | None = ...) -> None: ...
```
AlexWaygood, line 57: `Incomplete` might be better here -- I don't really know what's going on in the source code here (and don't really wanna spend time finding out! 😉), but it's probably not actually true that literally any object is okay for the mapping values: https://github.com/pypa/setuptools/blob/9022a215459137d8bd9446d995c0dd159079a2b2/setuptools/_distutils/dist.py#L232-L265.

AlexWaygood (commented):
Two more nits (sorry!)

stubs/setuptools/setuptools/command/upload_docs.pyi
```diff
@@ -8,7 +10,8 @@ class upload_docs(upload):
     user_options: ClassVar[list[tuple[str, str | None, str]]]
     boolean_options: ClassVar[list[str]]
     def has_sphinx(self): ...
-    sub_commands: Any
+    # The callable parameter is self: Self, but using Self still trips up mypy
+    sub_commands: ClassVar[list[tuple[str, Callable[[Self], bool] | None]]]  # type: ignore[misc, assignment]
```
Avasam, line 14: I didn't need to modify this file, but I wanted to show what I think may be a mypy issue with `Self` I spotted. pyright looks correct to me.

image

Example with `sub_commands` removed and inferred from superclass:
image

AlexWaygood: Interesting. I think using `Self` in `ClassVar`s might be underspecified/I'm not sure exactly what the semantics should be really. But this looks fine for now imo.

AlexWaygood (approved):
Thank you!

PR #9798, opened 2023-02-22
`argparse`: simplify `parse_args` overloads

AlexWaygood:
I think we can reduce the number of overloads from 5 to 3 while still allowing all valid calls.

### DIFF
--- a/stdlib/argparse.pyi
+++ b/stdlib/argparse.pyi
@@ -161,16 +161,11 @@ class ArgumentParser(_AttributeHolder, _ActionsContainer):
             add_help: bool = True,
             allow_abbrev: bool = True,
         ) -> None: ...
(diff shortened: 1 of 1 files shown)


stdlib/argparse.pyi
```diff
@@ -161,17 +161,12 @@ class ArgumentParser(_AttributeHolder, _ActionsContainer):
             add_help: bool = True,
             allow_abbrev: bool = True,
         ) -> None: ...
-    # The type-ignores in these overloads should be temporary.  See:
-    # https://github.com/python/typeshed/pull/2643#issuecomment-442280277
```
AlexWaygood, line 165: The `type: ignore`s have been here for 5 years now, time to stop claiming that they're "temporary" 😄

Akuli (approved):

PR #9800, opened 2023-02-23
Remove a default value in `distutils`

### LINKED ISSUE #9799: Daily tests failed on Thu Feb 23 2023

AlexWaygood:
Fixes #9799.

Stubtest started failing on our `setuptools` stubs because of a default value added to our stdlib `distutils` stubs. The reason for this is because our `setuptools` stubs erroneously claim here that the `setuptools._distutils.core.Distribution` class is the same as the stdlib `distutils.dist.Distribution` class:

https://github.com/python/typeshed/blob/06755e10ba0d39e7e4c18fcc2663d9da564a71ad/stubs/setuptools/setuptools/_distutils/core.pyi#L3

In actual fact, they're not, and they have different default values for this parameter. But reverting the addition of the default value is the simplest short-term fix for now.

(An alternative fix would be to just merge #9795.)


AlexWaygood:
Merging so the "stubtest failed" bot doesn't create a duplicate issue in a few hours

PR #9806, opened 2023-02-24
Improve typechecking of the test suite

AlexWaygood:
- Remove an unnecessary `pyright: ignore` in stubsabot. It's fine to have slightly less strict typing here, since the types in `METADATA.toml` files are verified in `check_consistent.py`.
- Enable pyright's `reportUnknownMemberType` setting in `tests/parse_metadata.py`.
- Use more explicit `object` annotations in `parse_metadata.py`, to enforce stricter mypy typechecking.
- Use the lowest supported Python version as the default for mypy's `--python-version` argument in `tests/typecheck_typeshed.py`, rather than the version of Python you're running locally.

Cc. @Avasam, for interest :)

### DIFF
--- a/.github/workflows/typecheck_typeshed_code.yml
+++ b/.github/workflows/typecheck_typeshed_code.yml
@@ -51,7 +51,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: actions/setup-python@v4
         with:
-          python-version: "3.10"
+          python-version: "3.9"
           cache: pip
           cache-dependency-path: requirements-tests.txt
       - run: pip install -r requirements-tests.txt
--- a/.github/workflows/typecheck_typeshed_code.yml
+++ b/.github/workflows/typecheck_typeshed_code.yml
@@ -35,7 +35,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: actions/setup-python@v4
         with:
-          python-version: "3.9"
+          python-version: "3.11"
           cache: pip
           cache-dependency-path: requirements-tests.txt
       - run: pip install -r requirements-tests.txt
@@ -51,7 +51,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: actions/setup-python@v4
         with:
-          python-version: "3.9"
+          python-version: "3.10"
           cache: pip
           cache-dependency-path: requirements-tests.txt
       - run: pip install -r requirements-tests.txt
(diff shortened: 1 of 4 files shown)


.github/workflows/typecheck_typeshed_code.yml
```diff
@@ -51,7 +51,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: actions/setup-python@v4
         with:
-          python-version: "3.9"
+          python-version: "3.10"
```
AlexWaygood, line 54: Can't bump this to 3.11 or pyright issues these warnings when typechecking the `tests` folder on linux:

```
/home/runner/work/typeshed/typeshed/tests/pytype_test.py:29:6 - error: Import "pytype" could not be resolved (reportMissingImports)
Error: Import "pytype" could not be resolved (reportMissingImports)
/home/runner/work/typeshed/typeshed/tests/pytype_test.py:[30](https://github.com/AlexWaygood/typeshed/actions/runs/4262796149/jobs/7418727501#step:6:31):6 - error: Import "pytype.imports" could not be resolved (reportMissingImports)
Error: Import "pytype.imports" could not be resolved (reportMissingImports)
```

AlexWaygood: > I think you meant to bump python-version: "3.9" below as well.

No -- I'd like it to perform type checking so that it flags if features we're using don't exist on Python 3.9. But it's harmless for the `pip install` step here to run on Python 3.10, which might theoretically be faster.

Now that you bring it up, I realise that this probably just causes confusion having the two versions being different, so I'll just revert the change on this line :)

AlexWaygood:
```suggestion
          python-version: "3.9"
```

Avasam: What about the one in the mypy job?
Afaik tests are run on python 3.10 (except mypy_test, iirc because there's some weird edge cases with 3.7, something about AST parsing).
Using 3.11 features (let's say in tests/utils.py) would pass `typecheck_typeshed_code.yml` but fail in 3.10 in CI tests with shared code (and fail on 3.9 locally)".

Or are you fine with that?

I know 3.11 *should* definitely be faster (that was one of the main selling point of the release :P ).

AlexWaygood: We can run the mypy job _on_ Python 3.11 (to get the added speed benefits that you point out), but _with_ `--python-version=3.9` passed to mypy. One of the other changes I've made in this PR is to have `--python-version=3.9` be the default for the `typecheck_typeshed.py` script that we use for running mypy on the `scripts/` and `tests/` directory.

For the pyright job, we'd only be using Python 3.11 to `pip install` the requirements, so it really doesn't matter as much :P

PR #9811, opened 2023-02-24
Add get_origin annotations

Ovsyanka83:
## Rationale
When you know for a fact that you are using get_origin on a GenericAlias, your type checker will show you errors about the result being `Any | None`. These type hints attempt to fix this issue at least for the most common use cases. I tried handling every overload that was possible to handle but if you have any suggestions -- please, share them :)

CI seems to be broken because not every import is available in earlier versions. Am thinking on how to handle it right now.


stdlib/typing_extensions.pyi
```diff
@@ -155,6 +156,13 @@ def get_type_hints(
...
 ) -> dict[str, Any]: ...
 def get_args(tp: Any) -> tuple[Any, ...]: ...
+@overload
+def get_origin(tp: GenericAlias) -> type: ...
+@overload
+def get_origin(tp: ParamSpecArgs | ParamSpecKwargs) -> ParamSpec: ...
+@overload
+def get_origin(tp: UnionType) -> type[UnionType]: ...
+@overload
 def get_origin(tp: Any) -> Any | None: ...
```
AlexWaygood, line 166: You should be able to fix the CI errors by doing version-dependent overloads like this:

```suggestion
if sys.version_info >= (3, 10):
    @overload
    def get_origin(tp: ParamSpecArgs | ParamSpecKwargs) -> ParamSpec: ...
    @overload
    def get_origin(tp: UnionType) -> type[UnionType]: ...
if sys.version_info >= (3, 9):
    @overload
    def get_origin(tp: GenericAlias) -> type: ...
    @overload
    def get_origin(tp: Any) -> Any | None: ...
else:
    def get_origin(tp: Any) -> Any | None: ...
```

AlexWaygood (commented):
Thanks! Could you make the same changes in `typing.pyi`?

stdlib/typing_extensions.pyi
Ovsyanka83: Will fix CI right now

stdlib/typing_extensions.pyi
```diff
@@ -5,6 +5,7 @@ import typing
 from _collections_abc import dict_items, dict_keys, dict_values
 from _typeshed import IdentityFunction, Incomplete
 from collections.abc import Iterable
+from types import GenericAlias, UnionType
```
AlexWaygood, line 8: You need to put these imports behind version guards as well (and move them after the `from typing` imports)

stdlib/typing.pyi
```diff
@@ -744,9 +749,21 @@ else:
         obj: _get_type_hints_obj_allowed_types, globalns: dict[str, Any] | None = None, localns: dict[str, Any] | None = None
     ) -> dict[str, Any]: ...
 
+def get_args(tp: Any) -> tuple[Any, ...]: ...
+
 if sys.version_info >= (3, 8):
```
AlexWaygood, line 754: Looks like you accidentally moved the `get_args` function outside of the `>= (3, 8)` block here (it only exists in `typing` on 3.8+)

```suggestion
if sys.version_info >= (3, 8):
    def get_args(tp: Any) -> tuple[Any, ...]: ...
```

Ovsyanka83: Yep, no worries. Still in progress :)

Ovsyanka83:
I should've used pre-commit here all along but I guess lesson learned.

AlexWaygood (approved):
Thanks!

AlexWaygood:
> I should've used pre-commit here all along but I guess it's lesson learned.

No worries at all! The reason we have autofixes turned on in CI is so that people _don't_ have use pre-commit locally :)

stdlib/typing_extensions.pyi
```diff
@@ -155,7 +160,21 @@ def get_type_hints(
     include_extras: bool = False,
 ) -> dict[str, Any]: ...
 def get_args(tp: Any) -> tuple[Any, ...]: ...
-def get_origin(tp: Any) -> Any | None: ...
+
+if sys.version_info >= (3, 10):
+    @overload
+    def get_origin(tp: ParamSpecArgs | ParamSpecKwargs) -> ParamSpec: ...
```
AlexWaygood, line 166: Oh, actually, this overload can work on 3.7+ for `typing_extensions` (but _not_ for `typing`, where `ParamSpec`s only exist on 3.10+):

```pycon
(py37) C:\Users\alexw\coding\typeshed>python
Python 3.7.16 (default, Jan 17 2023, 16:06:28) [MSC v.1916 64 bit (AMD64)] :: Anaconda, Inc. on win32
>>> from typing_extensions import ParamSpec, get_origin
>>> type(get_origin(ParamSpec("P").args))
<class 'typing_extensions.ParamSpec'>
```

stdlib/typing_extensions.pyi
```diff
@@ -155,7 +160,21 @@ def get_type_hints(
...
+    def get_origin(tp: UnionType) -> type[UnionType]: ...
+
+if sys.version_info >= (3, 9):
+    @overload
+    def get_origin(tp: GenericAlias) -> type: ...
+    @overload
+    def get_origin(tp: Any) -> Any | None: ...
+
+else:
+    def get_origin(tp: Any) -> Any | None: ...
```
AlexWaygood, line 177:
```suggestion
if sys.version_info >= (3, 10):
    @overload
    def get_origin(tp: UnionType) -> type[UnionType]: ...

if sys.version_info >= (3, 9):
    @overload
    def get_origin(tp: GenericAlias) -> type: ...

@overload
def get_origin(tp: ParamSpecArgs | ParamSpecKwargs) -> ParamSpec: ...
@overload
def get_origin(tp: Any) -> Any | None: ...
```

stdlib/typing_extensions.pyi
Ovsyanka83: Will take care of it later today. Thanks!

AlexWaygood (approved):

PR #9813, opened 2023-02-25
bisect: don't require Sequences

Rogdham:
This PR changes the type of the first argument of `bisect_left` and `bisect_right` (and by design `bisect` as well) from the `bisect` module.

Indeed, a full `Sequence` is not needed, only the following is needed:
 - If the `hi` parameter is `None`, only `__getitem__` and `__len__`
 - Else, only `__getitem__`

I'm asking for `SupportsLenAndGetItem` instead of only `SupportsGetItem` in the second case to avoid an exponential number of `@overload`.

---

You can check that only these attributes are needed with the following code:
```py
from bisect import bisect, bisect_left, bisect_right


class Minimal:
    content = 'abbbbcd'

    def __len__(self):
        return len(self.content)
    
    def __getitem__(self, pos):
        return self.content[pos]

obj = Minimal()

assert bisect(obj, 'b') == 5
assert bisect_left(obj, 'b') == 1
assert bisect_right(obj, 'b') == 5
```

### DIFF
--- a/stdlib/_bisect.pyi
+++ b/stdlib/_bisect.pyi
@@ -1,6 +1,6 @@
 import sys
-from _typeshed import SupportsRichComparisonT
-from collections.abc import Callable, MutableSequence, Sequence
+from _typeshed import SupportsLenAndGetItem, SupportsRichComparisonT
+from collections.abc import Callable, MutableSequence
 from typing import TypeVar, overload
 
 _T = TypeVar("_T")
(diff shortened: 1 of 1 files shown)


JelleZijlstra:
It's a bit more complicated because it uses the C sequence protocol and therefore doesn't allow e.g. dicts, but I don't think we need to try to express that in the stub.

Rogdham:
Thank you!

PR #8498, opened 2022-08-06
Add classmethod to stripe.PaymentIntent.confirm.

PIG208:
Similar to #7230, `stripe.PaymentIntent.confirm` can be called as an
instance method or a `classmethod`.

PR #9814, opened 2023-02-25
stripe checkout session stub

### LINKED ISSUE #9693: Add missing stripe.checkout.Session.expire stub
Stripe [2.62.0](https://github.com/stripe/stripe-python/blob/master/CHANGELOG.md#2620---2021-11-11) introduced `stripe.checkout.Session.expire`.

This function is missing in the typeshed stubs.

denyszhak:
Closes #9693 
Session was missing stubs


AlexWaygood:
Thanks! A lot of the methods in `stripe` are pretty hard to type -- the decorator [you can see here in the source code](https://github.com/stripe/stripe-python/blob/3f92d16e9f11ea9d3b109daebe0a5d3b1771f90f/stripe/api_resources/checkout/session.py#L50) means that the `expire()` method can be called both as an instance method and as a classmethod at runtime. It's not really possible to express that in the stub without a few hacks, unfortunately, so I've pushed an update to your PR along the lines of how we've previously approached this problem in #8498.

PR #9815, opened 2023-02-25
Stdlib: correct many parameter names

AlexWaygood:
I prepared this PR with the help of [this script](https://github.com/AlexWaygood/argcorrector) (which you'll notice has a very similar structure to `stubdefaulter`).


stdlib/multiprocessing/synchronize.pyi
```diff
@@ -30,7 +30,7 @@ class Condition(AbstractContextManager[bool]):
     ) -> None: ...
 
 class Event:
-    def __init__(self, lock: _LockLike | None = ..., *, ctx: BaseContext) -> None: ...
+    def __init__(self, *, ctx: BaseContext) -> None: ...
```
AlexWaygood, line 33: It doesn't look like this parameter has ever existed at runtime (from what I can see); I'm not sure how it made its way into the stub

stdlib/statistics.pyi
```diff
@@ -123,9 +123,7 @@ if sys.version_info >= (3, 10):
         intercept: float
 
 if sys.version_info >= (3, 11):
-    def linear_regression(
-        __regressor: Sequence[_Number], __dependent_variable: Sequence[_Number], *, proportional: bool = False
-    ) -> LinearRegression: ...
+    def linear_regression(__x: Sequence[_Number], __y: Sequence[_Number], *, proportional: bool = False) -> LinearRegression: ...
```
hauntsaninja, line 126: This one seems like a regression ;-)

stdlib/types.pyi
```diff
@@ -589,7 +589,7 @@ if sys.version_info >= (3, 9):
         @property
         def __parameters__(self) -> tuple[Any, ...]: ...
         def __init__(self, origin: type, args: Any) -> None: ...
-        def __getitem__(self, __typeargs: Any) -> GenericAlias: ...
+        def __getitem__(self, __key: Any) -> GenericAlias: ...
```
hauntsaninja, line 592: typeargs is more descriptive

hauntsaninja (commented):
Looks good! Two cases where the old names are better than the new names

stdlib/statistics.pyi
AlexWaygood:
```suggestion
    def linear_regression(__regressor: Sequence[_Number], __dependent_variable: Sequence[_Number], *, proportional: bool = False) -> LinearRegression: ...
```

stdlib/statistics.pyi
```diff
@@ -123,9 +123,7 @@ if sys.version_info >= (3, 10):
...
 
 elif sys.version_info >= (3, 10):
-    def linear_regression(__regressor: Sequence[_Number], __dependent_variable: Sequence[_Number]) -> LinearRegression: ...
+    def linear_regression(__x: Sequence[_Number], __y: Sequence[_Number]) -> LinearRegression: ...
```
AlexWaygood, line 129:
```suggestion
    def linear_regression(__regressor: Sequence[_Number], __dependent_variable: Sequence[_Number]) -> LinearRegression: ...
```

stdlib/types.pyi
AlexWaygood:
```suggestion
        def __getitem__(self, __typeargs: Any) -> GenericAlias: ...
```

hauntsaninja (approved):

AlexWaygood:
Thanks for the review!

PR #9818, opened 2023-02-27
Pyserial: remove apparently invalid method signature

alexrudd2:
Apologies in advance if I'm off-base, this is my first attempt at Python typing.

running `mypy` on an application built on top of `pyserial` complains:
> Unexpected keyword argument "size" for "read" of "SerialBase"  [call-arg]

This is because the stub file shows `__size` as a keyword argument.
` def read(self, __size: int = ...) -> bytes: ...  # same as io.RawIOBase.read but always returns bytes`

When I look at the actual [`pyserial` code](https://github.com/pyserial/pyserial/blob/31fa4807d73ed4eb9891a88a15817b439c4eea2d/serial/serialutil.py#L253), there is no `read()` method defined.  Instead, `SerialBase` inherits it from `io.RawIOBase` in the standard library.  According to the [documentation](https://github.com/pyserial/pyserial/blob/31fa4807d73ed4eb9891a88a15817b439c4eea2d/serial/serialutil.py#L253), `read()` accepts a `size` parameter and returns `bytes` (or `None`).

Presumably this line was added in the stub file for the `None` case?  But even then I see no reason to use `__size` instead of `size`.   I think the line should be either deleted altogether, or changed to 
` def read(self, size: int = ...) -> bytes: ...  # same as io.RawIOBase.read but always returns bytes`

### DIFF
--- a/stubs/pyserial/serial/serialutil.pyi
+++ b/stubs/pyserial/serial/serialutil.pyi
@@ -63,7 +63,6 @@ class SerialBase(io.RawIOBase):
         inter_byte_timeout: float | None = ...,
         exclusive: float | None = ...,
     ) -> None: ...
-    def read(self, __size: int = ...) -> bytes: ...  # same as io.RawIOBase.read but always returns bytes
     @property
     def port(self) -> str | None: ...
     @port.setter


srittau:
Thanks for your contribution! In typeshed, we strive to match the stubs to the implementation as closely as possible. In that sense, removing `read()` from `SerialBase` seems correct to me. That said, that won't solve your problem, as `RawIOBase.read()` is also marked as a positional-only argument (at least in HEAD).

Also, I'm sure that the tighter return type was added for a reason, so I'd like @hamdanal's input here, as the original submitter of the stubs.

alexrudd2:
> Thanks for your contribution! In typeshed, we strive to match the stubs to the implementation as closely as possible. In that sense, removing `read()` from `SerialBase` seems correct to me. That said, that won't solve your problem, as `RawIOBase.read()` is also marked as a positional-only argument (at least in HEAD).
> 
> Also, I'm sure that the tighter return type was added for a reason, so I'd like @hamdanal's input here, as the original submitter of the stubs.

Ok, I did not know about `/` marking positional -only arguments.  The documentation indicates this changed in Python3.9, so the calling code (which is 5+ years old) must in fact be already broken.

hamdanal:
This is a tricky one as pyserial breaks inheritance consistency and its docs at the same time.

Regarding the return type, although `SerialBase` does not implement `read`, it requires its subclasses to implement a stricter version that always returns `bytes`. You can see this in the code [here](https://github.com/pyserial/pyserial/blob/31fa4807d73ed4eb9891a88a15817b439c4eea2d/serial/serialutil.py#L549-L550).

As for the positional only argument vs pyserial documentation, pyserial itself is not consistent in naming these argument in all the subclasses of `SerialBase` so I followed `io.RawIOBase` signature of making them positional only.

Now that I look at it, `read()` MUST be implemented by `SerialBase` subclasses unlike `io.RawIOBase`. I'll send a PR marking it as `abstractmethod`. Thanks for bringing this up.

PR #9823, opened 2023-02-27
invoke: Update `Context.cd()` to accept a `pathlib.Path` or a string

rouge8:
This matches the (unreleased) type hints for invoke: https://github.com/pyinvoke/invoke/blob/e95631215fac9ef454f064e2a701f8fce351752d/invoke/context.py#L346


stubs/invoke/invoke/context.pyi
```diff
@@ -13,7 +14,7 @@ class Context(DataProxy):
     def prefix(self, command: str) -> AbstractContextManager[None]: ...
     @property
     def cwd(self) -> str: ...
-    def cd(self, path: str) -> AbstractContextManager[None]: ...
+    def cd(self, path: PathLike | str) -> AbstractContextManager[None]: ...
```
AlexWaygood, line 17: We have an alias in `_typeshed for this union, it would be good to use it here:

https://github.com/python/typeshed/blob/6b98f7307ba12b86a44a8499bab58ae87e768529/stdlib/_typeshed/__init__.pyi#L134

```suggestion
    def cd(self, path: StrPath) -> AbstractContextManager[None]: ...
```

rouge8: Nice! Fixed.

AlexWaygood (commented):
It looks to me like the runtime implementation is incorrect. They're calling `str()` on the `path` argument here, rather than [`os.fspath`](https://docs.python.org/3/library/os.html#os.fspath). That means that the function will work with `str` and `pathlib.Path` objects, but it won't _actually_ work with arbitrary `os.PathLike` objects, even if that's what the type hints in the upstream source code _say_. So maybe the type hint should actually be `str | pathlib.Path` rather than `StrPath`.

(Upstream should probably update their implementation so that it calls `os.fspath()` rather than `str()` on the `path` parameter. Then the implementation would match the type hint that they've provided.)

rouge8:
> It looks to me like the runtime implementation is incorrect. They're calling `str()` on the `path` argument [here](https://github.com/pyinvoke/invoke/blob/e95631215fac9ef454f064e2a701f8fce351752d/invoke/context.py#L392), rather than [`os.fspath`](https://docs.python.org/3/library/os.html#os.fspath). That means that the function will work with `str` and `pathlib.Path` objects, but it won't _actually_ work with arbitrary `os.PathLike` objects, even if that's what the type hints in the upstream source code _say_. So maybe the type hint should actually be `str | pathlib.Path` rather than `StrPath`.
> 
> (Upstream should probably update their implementation so that it calls `os.fspath()` rather than `str()` on the `path` parameter. Then the implementation would match the type hint that they've provided.)

Ooh, good catch. I'll open a PR upstream to fix that.

PR #9826, opened 2023-02-27
Add entries to `pika`'s stubtest_requirements

Avasam:
Split off from #9494 , these changes are beneficial on their own, and necessitate less review.


AlexWaygood:
Oh, looks like we need to delete some fictitious things :)

Issue #8625, opened 2022-08-26
`types-python-dateutil` Add return type for `rrule.__iter__`?


michaeloliverx:
```python
from datetime import date

from dateutil.rrule import MONTHLY, rrule

datetimes = [dt for dt in rrule(MONTHLY, dtstart=date(2001, 1, 1), until=date(2005, 1, 1))]

reveal_type(datetimes)  # Revealed type is "builtins.list[Any]"
```

https://github.com/python/typeshed/issues/8625#tasklist-block-882f1d08-3570-4c11-af49-38329f2aad04

AlexWaygood:
PR welcome! :)

AmberAnsari89:
i want to work on it..please suggest exact changes to be made.

srittau:
#9828

PR #9828, opened 2023-02-28
rrule.__iter__ return type

denyszhak:
PR to https://github.com/python/typeshed/issues/8625
Seems like both rrule and rruleset iterate over datetime.datetime


srittau (approved):
Thanks!

PR #9835, opened 2023-03-03
Add `__name__` to `_Wrapped` in functools

tmke8:
There can be scenarios where `__name__` isn't set, as pointed out here: https://github.com/python/mypy/pull/14815#issuecomment-1453509611 but the same is true for `Callable`, so this has precedent.

### DIFF
--- a/test_cases/stdlib/check_functools.py
+++ b/test_cases/stdlib/check_functools.py
@@ -1,10 +1,28 @@
 from __future__ import annotations
 
 import sys
+from functools import wraps
+from typing import Callable, ParamSpec, TypeVar
+from typing_extensions import ParamSpec, assert_type
+
+P = ParamSpec("P")
+T_co = TypeVar("T_co", covariant=True)
+
+
+def my_decorator(func: Callable[P, T_co]) -> Callable[P, T_co]:
+    func_name = func.__name__
+
+    @wraps(func)
+    def wrapper(*args: P.args, **kwargs: P.kwargs):
+        print(args)
+        return func(*args, **kwargs)
+
+    wrapper.__name__ = func_name
+    return wrapper
+
 
 if sys.version_info >= (3, 8):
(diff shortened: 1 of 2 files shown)


test_cases/stdlib/check_functools.py
```diff
@@ -1,10 +1,28 @@
 from __future__ import annotations
 
 import sys
+from functools import wraps
+from typing import Callable, ParamSpec, TypeVar
```
AlexWaygood, line 5:
```suggestion
from typing import Callable, TypeVar
```

test_cases/stdlib/check_functools.py
```diff
@@ -1,10 +1,28 @@
...
+
+P = ParamSpec("P")
+T_co = TypeVar("T_co", covariant=True)
+
+
+def my_decorator(func: Callable[P, T_co]) -> Callable[P, T_co]:
+    func_name = func.__name__
+
+    @wraps(func)
+    def wrapper(*args: P.args, **kwargs: P.kwargs):
```
AlexWaygood, line 16:
```suggestion
    def wrapper(*args: P.args, **kwargs: P.kwargs) -> T_co:
```

test_cases/stdlib/check_functools.py
```diff
@@ -1,10 +1,28 @@
...
+        print(args)
+        return func(*args, **kwargs)
+
+    wrapper.__name__ = func_name
```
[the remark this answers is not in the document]
tmke8: done! `__module__`, `__doc__` and `__annotations__` were already provided by `object`.

tmke8:
Mypy primer looks as expected :rocket:

AlexWaygood (approved):
LGTM. @erictraut, does this make sense to you?

erictraut:
Yeah, makes sense to me.

PR #9844, opened 2023-03-06
[oauthlib] Add types in oauth1/rfc5849/request_validator.pyi

malthejorgensen:
This adds type annotations to `oauthlib.oauth1.rfc5849.RequestValidator` according to the doctype strings in [the original file itself](https://github.com/oauthlib/oauthlib/blob/master/oauthlib/oauth1/rfc5849/request_validator.py) (as of v3.2.2 - the latest version).

It should be noted that `RequestValidator` is intended to be an abstract class, so many methods return an exception akin to the `NotImplementedError`-exception.

The types added in this PR are the intended types as specified in the docstrings of `oauthlib` rather than `NoReturn` which would be more "technically correct" for the abstract methods that always raise an exception. I believe the types in this PR provide more value than simply having `NoReturn` for many of the methods.


stubs/oauthlib/oauthlib/oauth1/rfc5849/request_validator.pyi
```diff
@@ -1,61 +1,70 @@
...
-    ) -> None: ...
-    def validate_redirect_uri(self, client_key, redirect_uri, request) -> None: ...
-    def validate_requested_realms(self, client_key, realms, request) -> None: ...
+        self,
+        client_key: str,
+        timestamp,
+        nonce: str,
+        request: Request,
+        request_token: Incomplete | None = ...,
+        access_token: Incomplete | None = ...,
```
AlexWaygood, line 58:
```suggestion
        request_token: str | None = None,
        access_token: str | None = None,
```

stubs/oauthlib/oauthlib/oauth1/rfc5849/request_validator.pyi
```diff
@@ -1,61 +1,70 @@
...
     def validate_realms(
-        self, client_key, token, request, uri: Incomplete | None = ..., realms: Incomplete | None = ...
-    ) -> None: ...
-    def validate_verifier(self, client_key, token, verifier, request) -> None: ...
-    def verify_request_token(self, token, request) -> None: ...
-    def verify_realms(self, token, realms, request) -> None: ...
-    def save_access_token(self, token, request) -> None: ...
-    def save_request_token(self, token, request) -> None: ...
-    def save_verifier(self, token, verifier, request) -> None: ...
+        self, client_key: str, token: str, request: Request, uri: str | None = ..., realms: list[str] | None = ...
```
AlexWaygood, line 63:
```suggestion
        self, client_key: str, token: str, request: Request, uri: str | None = None, realms: list[str] | None = None
```

AlexWaygood (commented):
Thanks! Mostly looks good; a few small comments:

stubs/oauthlib/oauthlib/oauth1/rfc5849/request_validator.pyi
```diff
@@ -1,61 +1,70 @@
...
+    def check_access_token(self, request_token: str) -> bool: ...
+    def check_nonce(self, nonce: str) -> bool: ...
+    def check_verifier(self, verifier) -> bool: ...
+    def check_realms(self, realms: list[str]) -> bool: ...
     @property
     def dummy_client(self) -> None: ...
     @property
     def dummy_request_token(self) -> None: ...
     @property
     def dummy_access_token(self) -> None: ...
```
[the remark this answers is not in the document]
malthejorgensen: According to the original file, these all return `str` so I've added that change in https://github.com/python/typeshed/pull/9844/commits/d44b18893d2e53ce29973004b4aa964f02c31034

malthejorgensen:
Thank you @AlexWaygood!

I've commited your suggestions via the Github UI. This makes the commits look a bit messy, but I wanted to make sure I got all of your suggested changes right and preserved authorship.

AlexWaygood (approved):
Thanks, this is a great improvement!

> I've commited your suggestions via the Github UI. This makes the commits look a bit messy, but I wanted to make sure I got all of your suggested changes right and preserved authorship.

No worries about the messy commits -- everything is squash-merged at typeshed, so we care very little about git history on the PR itself ^_^

PR #9847, opened 2023-03-06
`pytype_test.py`: Fix typechecking errors

AlexWaygood:
The type checkers have changed their minds about why they don't like this snippet of code, following the new release of `types-setuptools`, following #9747

(Failing CI run here: https://github.com/python/typeshed/actions/runs/4346178794)

Cc. @Avasam

PR #9849, opened 2023-03-06
pytype_test.py: Fix typechecking errors following #9747

Avasam:
Follow up to #9747

Superseeds https://github.com/python/typeshed/pull/9847#discussion_r1126830446 ( Closes #9847 ) by fixing the potentially `None` type error in a more principled way. The code already implicitly made this assert assumption.

CC. @AlexWaygood


AlexWaygood (approved):
Ah nice, thank you!

AlexWaygood (approved):
Ah nice, thank you!

PR #9856, opened 2023-03-08
[pydoc] `safeimport` can return `None`

smallSwed:
The `locate` (see: [pydoc.py#L1730](https://github.com/python/cpython/blob/ffb41eaaf4b04f7f52ee095253fcc1c5ba62ca28/Lib/pydoc.py#L1730)) and `safeimport`  (see: [pydoc.py#L454](https://github.com/python/cpython/blob/ffb41eaaf4b04f7f52ee095253fcc1c5ba62ca28/Lib/pydoc.py#L454) or [pydoc.py#L460](https://github.com/python/cpython/blob/ffb41eaaf4b04f7f52ee095253fcc1c5ba62ca28/Lib/pydoc.py#L460)) functions could return a `None`, while in the `pydoc.pyi` only object type was hinted.


stdlib/pydoc.pyi
```diff
@@ -186,7 +186,7 @@ def tempfilepager(text: str, cmd: str) -> None: ...
 def ttypager(text: str) -> None: ...
 def plainpager(text: str) -> None: ...
 def describe(thing: Any) -> str: ...
-def locate(path: str, forceload: bool = ...) -> object: ...
+def locate(path: str, forceload: bool = ...) -> object | None: ...
```
AlexWaygood, line 189: `None` is a subtype of `object`

AlexWaygood (commented):
The `safeimport` change looks great! I think `None` is redundant in the union `object | None`, though, since `object` is the top type, so not so sure about the change to `locate`

stdlib/pydoc.pyi
smallSwed: That's true. 
Including the `None` would provide better type hinting for users/type checkers, but most of them will not complain, I noticed the SonarLint does on the `is None` check if I use any of these functions. I should probably open an issue on their page.
If we do not accept this we should probably remove these also [stubs/six/six/__init__.pyi#L53](https://github.com/python/typeshed/blob/c661ac3ad6dc32d789e7458654dda49d095b2ba9/stubs/six/six/__init__.pyi#L53) and[stubs/console-menu/consolemenu/console_menu.pyi](https://github.com/python/typeshed/blob/c661ac3ad6dc32d789e7458654dda49d095b2ba9/stubs/console-menu/consolemenu/console_menu.pyi#L23)

AlexWaygood: > Including the `None` would provide better type hinting for users/type checkers

I don't think so -- I think type checkers should just immediately simplify the `object | None` union to `object`, if they're working correctly. Do you have evidence of a situation where a type checker treats `object | None` differently to `object`? I'd be curious to see it.

> If we do not accept this we should probably remove these also [stubs/six/six/**init**.pyi#L53](https://github.com/python/typeshed/blob/c661ac3ad6dc32d789e7458654dda49d095b2ba9/stubs/six/six/__init__.pyi#L53) and[stubs/console-menu/consolemenu/console_menu.pyi](https://github.com/python/typeshed/blob/c661ac3ad6dc32d789e7458654dda49d095b2ba9/stubs/console-menu/consolemenu/console_menu.pyi#L23)

Sure, I'll happily accept a separate PR cleaning those up :)

AlexWaygood (approved):
Thanks!

stdlib/pydoc.pyi
AlexWaygood: Hmm, yeah, I think SonarLint's behaviour is incorrect there. Might want to file a bug report with them :)

PR #9859, opened 2023-03-09
[pywin32] win32gui.FindWindowEx parameters __ClassName and __WindowName are optional

bparzella:
Passing an empty string doesn't work in this case, it raises an error:

```
>>> win32gui.FindWindowEx(None, None, None, "c:\\windows\\system32\\cmd.exe") 
3277888
>>> win32gui.FindWindowEx(None, None, "", "c:\\windows\\system32\\cmd.exe")   
pywintypes.error: (123, 'FindWindowEx', 'Die Syntax für den Dateinamen, Verzeichnisnamen oder die Datenträgerbezeichnung ist falsch.')
```

Windows API reference: https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-findwindowexa


srittau (approved):
I have no experience with pywin32, but this looks correct to me. Thanks!

PR #9861, opened 2023-03-10
[alt] typing: accept buffers in `IO.write`

### LINKED ISSUE #9082: Support buffers in `IO[bytes].write`

AlexWaygood:
An alternative PR to #9084 (closes #9084)

Fixes #9082

Co-authored-by: JelleZijlstra <jelle.zijlstra@gmail.com>


stdlib/tempfile.pyi
```diff
@@ -215,7 +215,17 @@ class _TemporaryFileWrapper(Generic[AnyStr], IO[AnyStr]):
...
+    @overload
+    def write(self: _TemporaryFileWrapper[bytes], s: ReadableBuffer) -> int: ...
+    @overload
     def write(self, s: AnyStr) -> int: ...
+    @overload
+    def writelines(self: _TemporaryFileWrapper[str], lines: Iterable[str]) -> None: ...
+    @overload
+    def writelines(self: _TemporaryFileWrapper[bytes], lines: Iterable[ReadableBuffer]) -> None: ...
+    @overload
     def writelines(self, lines: Iterable[AnyStr]) -> None: ...
```
srittau, line 229: What's the purpose of the overloads here? Why not use a simple union?

stdlib/tempfile.pyi
```diff
@@ -215,7 +215,17 @@ class _TemporaryFileWrapper(Generic[AnyStr], IO[AnyStr]):
     def tell(self) -> int: ...
     def truncate(self, size: int | None = ...) -> int: ...
     def writable(self) -> bool: ...
+    @overload
+    def write(self: _TemporaryFileWrapper[str], s: str) -> int: ...
+    @overload
+    def write(self: _TemporaryFileWrapper[bytes], s: ReadableBuffer) -> int: ...
+    @overload
     def write(self, s: AnyStr) -> int: ...
```
srittau, line 223: Is the third overload needed? This should be covered by the preceding two overloads. (Same for further overloads below.)

AlexWaygood: Unfortunately the test cases I've added don't pass for pyright unless we include this third overload. This is due to a design decision by pyright in the way it handles contrained `TypeVar`s that leads to different behaviour from mypy: https://github.com/microsoft/pyright/blob/main/docs/mypy-comparison.md#constrained-type-variables

It's the same issue that led to https://github.com/microsoft/pyright/issues/4534 being filed, which we then fixed over at typeshed in #9592.

AlexWaygood:

Two errors go away (`psycopg`, `cwltool`), one new error appears (`rich`), and two existing errors have their error code changed (`urllib3`). Looks promising to me. @JelleZijlstra, what do you think?

JelleZijlstra (approved):

PR #9862, opened 2023-03-10
Better reflect non-datetime uses with tzinfo

pganssle:
The actual tzinfo interface allows for passing a `time` as well as `None` to the three main abstract methods, and it expects that when you pass it a `datetime` you will *always* get a valid answer (whereas for `time` and `None` it depends on whether or not the zone represents a fixed offset).

I am not sure if this is the right way to stack `overload` and `abstractmethod` decorators.


pganssle:
Hmm... 

> + sphinx/builders/gettext.py: note: In member "dst" of class "LocalTimeZone":
> + sphinx/builders/gettext.py:192:5: error: Signature of "dst" incompatible with supertype "tzinfo"  [override]
> + sphinx/builders/gettext.py:192:5: note:      Superclass:
> + sphinx/builders/gettext.py:192:5: note:          @overload
> + sphinx/builders/gettext.py:192:5: note:          def dst(self, datetime, /) -> timedelta
> + sphinx/builders/gettext.py:192:5: note:          @overload
> + sphinx/builders/gettext.py:192:5: note:          def dst(self, Optional[time], /) -> Optional[timedelta]
> + sphinx/builders/gettext.py:192:5: note:      Subclass:
> + sphinx/builders/gettext.py:192:5: note:          def dst(self, dt: Optional[datetime]) -> timedelta

Not sure I fully understand the situation here. It seems like `dst(self, dt: Optional[datetime]) -> timedelta` is a valid choice for an implementation, except that it says it doesn't support `datetime.time`. Is that the reason this is failing, or is the issue that it doesn't understand that the cases in the superclass are:

```
1. datetime -> timedelta
2. None -> timedelta
3. None -> None
```

And the cases in the subclass are:

```
1. datetime -> timedelta
3. None -> timedelta
```

Seems compatible to me.

If it's the `time` thing, that feels like it could be tricky, since right now valid uses of `tzinfo` are forbidden by the typeshed stubs (e.g. how do you even use the `tzinfo` parameter to `datetime.time`?), but adding that to the signature will break any subclasses that haven't enumerated it.

pganssle:
> If a superclass says that it will accept a `datetime.time` for a parameter in some circumstances, but a subclass says that it doesn't accept a `datetime.time` under any circumstances for the same parameter, then that violates the Liskov Substitution Principle.

OK, so in this case I think that this will just be a breaking change, since the actual `tzinfo` interface does allow you to pass it a `datetime.time`. Should we just ignore the breakages in third party types, then, and focus on the typeshed breakage, which appears to be this:

```
stdlib/zoneinfo/__init__.pyi:31: error: Signature of "dst" incompatible with supertype "tzinfo"  [override]
stdlib/zoneinfo/__init__.pyi:31: note:      Superclass:
stdlib/zoneinfo/__init__.pyi:31: note:          def dst(self, Union[datetime, time, None], /) -> Optional[timedelta]
stdlib/zoneinfo/__init__.pyi:31: note:      Subclass:
stdlib/zoneinfo/__init__.pyi:31: note:          @overload
stdlib/zoneinfo/__init__.pyi:31: note:          def dst(self, datetime, /) -> timedelta
stdlib/zoneinfo/__init__.pyi:31: note:          @overload
stdlib/zoneinfo/__init__.pyi:31: note:          def dst(self, Optional[time], /) -> Optional[timedelta]
```

Not sure why that's not considered compatible, since the subclass does accept `datetime`, `None` and `time`, and returns either `timedelta` or `None`, its just that the subclass is being more specific about some of the situations where you would never get a `None`.

hauntsaninja:
In general, typeshed is okay with forcing people who subclass types to update when we improve the types, so the two new primer hits from that in sphinx and mongodb aren't the worst.

That said, there's definitely a mypy bug here. If we fix that, it'll be the happiest way forward. I filed https://github.com/python/mypy/issues/15121 for this

stdlib/zoneinfo/__init__.pyi
```diff
@@ -19,10 +19,19 @@ class ZoneInfo(tzinfo):
...
     @classmethod
-    def clear_cache(cls, *, only_keys: Iterable[str] | None = None) -> None: ...
-    def tzname(self, __dt: datetime | None) -> str | None: ...
-    def utcoffset(self, __dt: datetime | None) -> timedelta | None: ...
-    def dst(self, __dt: datetime | None) -> timedelta | None: ...
+    def clear_cache(cls, *, only_keys: Iterable[str] | None = ...) -> None: ...
+    @overload
+    def tzname(self, __dt: datetime) -> str: ...
+    @overload
+    def tzname(self, __dt: time | None) -> str: ...
```
JelleZijlstra, line 26:
```suggestion
    def tzname(self, __dt: time | None) -> str | None: ...
```
Currently the two overloads return the same type

JelleZijlstra:
Thanks for contributing! I'm closing this PR for now, because it still
fails some tests and has unresolved review feedback
after three months of inactivity. If you are still interested, please feel free to open

PR #9867, opened 2023-03-11
add missing properties for redis.lock.Lock (#9851)

### LINKED ISSUE #9851: Missing Redis properties
I'm totally not sure if this is the right place to ask or not, so feel free to point me elsewhere if this isn't the forum.

I'm using the redis package and notice that whlle types-redis has most of the methods, none of the properties seem to be there.

Example:
```python
import redis.client
import resis.lock

lock = redis.lock.Lock(name='foo', redis=redis.client.Redis('localhost'))

# The below isn't known by mypy, or any other properties like timeout or redis (the client)
print(lock.name) # type: ignore[attr-defined]
```

smallSwed:
I added missing properties for `redis.lock.Lock`. If anything other needs to be done let me know.

Fixes #9851

### DIFF
--- a/stubs/redis/redis/lock.pyi
+++ b/stubs/redis/redis/lock.pyi
@@ -15,6 +15,13 @@ class Lock:
     lua_extend: ClassVar[Incomplete | None]
     lua_reacquire: ClassVar[Incomplete | None]
     lua_release: ClassVar[Incomplete | None]
+    redis: Redis[Any]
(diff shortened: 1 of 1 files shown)


smallSwed:
PR for https://github.com/python/typeshed/issues/9851

AlexWaygood (approved):
Thanks!

PR #9870, opened 2023-03-12
add signal.SIGSTKFLT

gruebel:
I hope I added it in the right spot. It was added with Python 3.11 is only available on Linux, not MacOS.


gruebel:
@AlexWaygood thanks, didn't see it there 🙈

AlexWaygood (approved):

AlexWaygood:
Thanks!

PR #9873, opened 2023-03-13
Tensorflow protobuf stubs

hmc-cs-mdrissi:
# Summary
Add generated stubs for commonly used protobufs in tensorflow. All stubs were generated with mypy-protobuf making a similar script as generate_proto_stubs.sh

# Discussion
For review each commit mainly touches sync_protobuf_stubs.sh or it just generates tensorflow stubs. Odd commits define script/update Metadata.toml, even commits generate protobuf stubs.

I included most of the tensorflow folders that define protobufs I've seen/think appear in public apis. There are a couple more folders that have .proto files I could add if anyone requests or if we want to just have all of them.

edit: My commit rule sorta broke on later commits as I went through CI errors. The only non generated files are adding bash script for generating pb2.pyi and updating configs (metadata.toml, allowlist, etc).


hmc-cs-mdrissi:
Hmm, on error message of 

`AssertionError: Only 'stub' files allowed in the "stubs/tensorflow" directory; got: stubs/tensorflow/sync_protobuf_stubs.sh`

where should I place this script? Should I move it to top level scripts directory?

AlexWaygood:
> Hmm, on error message of
> 
> `AssertionError: Only 'stub' files allowed in the "stubs/tensorflow" directory; got: stubs/tensorflow/sync_protobuf_stubs.sh`
> 
> where should I place this script? Should I move it to top level scripts directory?

I guess that seems like the best place!

hmc-cs-mdrissi:
Sure moved script up there.

On mypy primer, jax2tf errors are caused by [mypy-protobuf](https://github.com/nipunn1313/mypy-protobuf#types-enum-int-values-more-strongly) being stricter then runtime when it comes to enums vs int. 

The last error `"source_line" to "OpMetadataGraph"` looks correct to me and they just got one type hint off. The first error, `tensorflow.python.profiler` looks like I need to add a couple `__init__.pyi` files and leave them as incomplete with getattr for now.

AlexWaygood:
The stubs generated with mypy-protobuf tend to break pytype FYI, so you'll probably have to add some files to https://github.com/python/typeshed/blob/main/tests/pytype_exclude_list.txt unfortunately :/

hmc-cs-mdrissi:
Yeah looks I'll need to update pytype exclusion list.

For flake8 errors should I update [flake exclusion](https://github.com/python/typeshed/blob/978d885d0b1afe919f66c063f27e3eb0b1c68a9d/.flake8#L42) for pb2.py to have F401? All of the errors are roughly,

`F401 'tensorflow.tsl.protobuf.histogram_pb2.HistogramProto' imported but unused`

for imports that look like,

```
from tensorflow.tsl.protobuf.histogram_pb2 import (
    HistogramProto as HistogramProto,
)
```

I think these should be treated as re-exports and unsure why flake8 is unhappy with them.

AlexWaygood:
> I think these should be treated as re-exports and unsure why flake8 is unhappy with them.

Yeah, this is confusing me also. I'd like to take a closer look and see if I can figure out what's going on here. Pyright and pycln both seem happy with these imports.

AlexWaygood:
> For flake8 errors should I update [flake exclusion](https://github.com/python/typeshed/blob/978d885d0b1afe919f66c063f27e3eb0b1c68a9d/.flake8#L42) for pb2.py to have F401? All of the errors are roughly,
> 
> `F401 'tensorflow.tsl.protobuf.histogram_pb2.HistogramProto' imported but unused`
> 
> for imports that look like,
> 
> ```
> from tensorflow.tsl.protobuf.histogram_pb2 import (
>     HistogramProto as HistogramProto,
> )
> ```
> 
> I think these should be treated as re-exports and unsure why flake8 is unhappy with them.

It seems like this is actually just pyflakes's standard behaviour; I can reproduce it by running flake8 on just this:

```python
from collections import deque as deque
```

I'm surprised we haven't run into this behaviour by pyflakes before now at typeshed. The discussion on https://github.com/PyCQA/pyflakes/issues/474 indicates that they are very unlikely to change that behaviour.

AlexWaygood:
I'll file a PR to silence F401 for all of typeshed -- it seems pretty unneeded in our CI, given that pyright and pycln already both check for unused imports, very effectively.

AlexWaygood:
- #9878

AlexWaygood:
Just pytype that's still failing now!

stubs/tensorflow/tensorflow/core/example/example_parser_configuration_pb2.pyi
```diff
@@ -0,0 +1,127 @@
+"""
+@generated by mypy-protobuf.  Do not edit manually!
+isort:skip_file
```
AlexWaygood, line 3: Huh, I hadn't noticed before that mypy-protobuf set these isort directives in the pyi files it generated. (No change requested!)

hmc-cs-mdrissi:
All green now and I think last commit should make mypy primer match what I expect.

scripts/sync_tensorflow_protobuf_stubs.sh
AlexWaygood, line 1: Should this script have steps to run `isort` and `black`, like our existing `generate_proto_stubs.sh` script?

https://github.com/python/typeshed/blob/28dd6c1a2e9ae91fd239a25afdf6166105aa81dc/scripts/generate_proto_stubs.sh#L76-L77

hmc-cs-mdrissi: I'll add black. isort seems confusing to add because generated files contain `isort:skip_file`

AlexWaygood: The `*_pb2.pyi` files will be excluded from being blackened anyway due to this line: https://github.com/python/typeshed/blob/c9fa41683a9d38aba1774394f18ed69b5bae6e57/pyproject.toml#L8

See the comment immediately above for the motivation. So maybe we should just remove the isort and black steps from our existing script.

@nipunn1313, thoughts?

hmc-cs-mdrissi: Hmm black also doesn't actually do anything for pb2.pyi currently. Our black configuration is set to exclude pb2.pyi files entirely [here](https://github.com/python/typeshed/blob/c9fa41683a9d38aba1774394f18ed69b5bae6e57/pyproject.toml#L8).

So I think both of those lines in other script do nothing and are just ignored.

AlexWaygood: > edit: Sorry didn't refresh page and see your middle comment til just now. Yes I think we should just drop isort/black in script.

We can wait for Nipunn's thoughts and do that in a separate PR, if he agrees; no need to do that here

nipunn1313: yep that sounds good!

scripts/sync_tensorflow_protobuf_stubs.sh
```diff
@@ -0,0 +1,65 @@
+#!/bin/bash
+set -euxo pipefail
+
+# Partly based on scripts/generate_proto_stubs.sh.
+
+# Generates the protobuf stubs for the given tensorflow version using
+# mypy-protobuf. Should be run like ./sync_protobuf_stubs.sh
```
JelleZijlstra, line 7:
```suggestion
# mypy-protobuf. Should be run like ./sync_tensorflow_protobuf_stubs.sh
```

scripts/sync_tensorflow_protobuf_stubs.sh
```diff
@@ -0,0 +1,65 @@
...
+pip install mypy-protobuf=="$MYPY_PROTOBUF_VERSION"
+
+mkdir repository
+pushd repository &> /dev/null
+    git clone https://github.com/tensorflow/tensorflow.git
+    pushd tensorflow &> /dev/null
+        git checkout v"$TENSORFLOW_VERSION"
+
+        # Folders here cover the more commonly used protobufs externally and
+        # there dependencies. Tensorflow has more protobufs and can be added if requested.
```
JelleZijlstra, line 31:
```suggestion
        # their dependencies. Tensorflow has more protobufs and can be added if requested.
```

stubs/tensorflow/@tests/stubtest_allowlist.txt
```diff
@@ -64,3 +64,7 @@ tensorflow.keras.layers.Layer.__new__
 # disagreements.
 tensorflow.keras.layers.*.build
 tensorflow.keras.layers.*.compute_output_shape
+
+# pb2.pyi generated by mypy-protobuf diverge with runtime in many ways. There stubs
```
JelleZijlstra, line 68:
```suggestion
# pb2.pyi generated by mypy-protobuf diverge with runtime in many ways. These stubs
```

PR #9874, opened 2023-03-13
Add stubs for WebOb

Daverball:
I've already created some more simple private stubs for WebOb, so I thought I'd try to clean them up and contribute them to the community. WebOb makes very heavy use of descriptors, which I've tried to fully leverage, but I will have to test if all the overloads actually work correctly. It might be worth going back to something more simplistic again which replaces some of the descriptors which just plain attributes, even if WebOb internally would be more generous what it accepts as an input for the assignment.


pyrightconfig.stricter.json
```diff
@@ -73,7 +73,8 @@
         "stubs/tqdm",
         "stubs/ttkthemes",
         "stubs/urllib3",
-        "stubs/vobject",
+        "stubs/vobject",,
```
AlexWaygood, line 76:
```suggestion
        "stubs/vobject",
```

stubs/WebOb/METADATA.toml
```diff
@@ -0,0 +1 @@
+version = "1.8.*"
```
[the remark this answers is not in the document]
Daverball: I've added an allowlist for the things I specifically chose to omit and/or change

stubs/WebOb/@tests/stubtest_allowlist.txt
```diff
@@ -0,0 +1,89 @@
...
+webob.multidict.NoVars.popitem
+webob.multidict.NoVars.setdefault
+webob.multidict.NoVars.update
+webob.request.AdhocAttrMixin.__delattr__
+webob.request.AdhocAttrMixin.__getattr__
+webob.request.AdhocAttrMixin.__setattr__
+webob.request.BaseRequest.__init__
+webob.request.NoDefault
+webob.response.Response.set_cookie
+webob.response.ResponseBodyFile.write
```
AlexWaygood, line 89: Are all of these being allowlisted because they're missing from the stub? We prefer not to allowlist anything without a comment stating why we're ignoring stubtest for this specific situation. With an allowlist this long and no comments, it's difficult to tell which objects are simply missing from the stub, and which objects stubtest is complaining about for some other reason, such as them having a different type or signature to what they have at runtime

Daverball: There is a reason for everything. The vast majority of missing stubs (all the ones ending with `_re`) are either a plain string that will be used as part of a regex that is assembled later, a marked regex string pattern or a compiled regex pattern that is used internally for parsing some part of a request/http header.
I didn't really like how inconsistently they were used, so it didn't really make sense to me to expose them to the public API, as far as I'm concerned these should all be prefixed with an underscore, but that's not up to me.

Then there is a whole slew of Python 2.7 compatibility stuff in the compat module as well as some stuff that was previously deprecated and is already removed upstream (methods such as `iteritems`, `iterkeys`, `itervalues`) as well as the `expires` argument on `set_cookie` which has been replaced by `max_age`)

I've moved some of the methods in the acceptparse from the derived class to the intermediary internal superclass, since the method signatures were the same, to reduce copy pasta, but I could move them back if that is not a good enough reason to have them on the allowlist.

I've removed `MimeAccept` since it is a deprecated API and wouldn't really interact well with the accept properties.

I've removed the `DEFAULT` parameter on the `AdhhocAttrMixin` methods purely since it seemed like internal API, I don't really see the point of ever manually setting that parameter either, which is why I removed it to make the signature less confusing.

I think that covers everything, except `webob.request.NoDefault` which I overlooked. I will fix that one.

AlexWaygood: Okay great -- could you add all of that rationale as comments in the allowlist? I really like the way hamdanal did it when they contributed stubs for `pyserial`:

https://github.com/python/typeshed/blob/ad9c7c1d0c842ac0cb3b010ecdd9402aac5f1bd3/stubs/pyserial/%40tests/stubtest_allowlist.txt#L1-L10

stubs/WebOb/webob/multidict.pyi
```diff
@@ -0,0 +1,113 @@
...
+    dict_of_lists = mixed
+    def __contains__(self, key) -> bool: ...
+    has_key = __contains__
+    def copy(self): ...
+    def __len__(self) -> int: ...
+    def iterkeys(self): ...
+    keys = iterkeys
+    items = iterkeys
+    values = iterkeys
+    __iter__ = iterkeys
```
Daverball, line 113: I forgot to annotate these. This is mostly a note to myself. It would be nice if there was a way to ensure `MultiDict | NoVars` could narrow to just `MultiDict` with a simple truthyness check, since `NoVars` is never truthy. Unfortunately I don't think mypy does any type narrowing on static truthyness of types yet. It would be annoying to make everywhere accept either of these since they don't share a common superclass.

stubs/WebOb/webob/request.pyi
```diff
@@ -0,0 +1,244 @@
...
+    def is_xhr(self) -> bool: ...
+    host: str
+    @property
+    def domain(self) -> str: ...
+    body: bytes
+    json: Any
+    json_body: Any
+    text: str
+    @property
+    def POST(self) -> MultiDict[str, str] | NoVars: ...
```
Daverball, line 129: Technically this should be `MultiDict[str, str | cgi.FieldStorage] | NoVars`. In practice having to deal with `isinstance` checks on every value you get out of `POST` is fairly annoying, but it is also the only way to be fully type safe, so I'm not sure what the best answer is here. Having `NoVars` in the type union is annoying as well, since usually you don't really care whether it's `NoVars` or `MultiDict` in a request, since `NoVars` will behave just like an empty `MultiDict`. Sometimes it's useful to distinguish between an empty POST and this request can't even have POST parameters to begin with, but `Request.method` should be sufficient for that...

TLDR: Should I prioritize type safety or ease of use? This is also kind of a use-case where `AnyOf` would be nice to have. It's a nice band-aid solution for APIs that weren't really thought out well enough in terms of type safety, without having to throw type safety completely out of the window by returning `Any`.

stubs/WebOb/webob/request.pyi
```diff
@@ -0,0 +1,244 @@
...
+    @property
+    def GET(self) -> GetDict: ...
+    @property
+    def params(self) -> NestedMultiDict[str, str]: ...
```
Daverball, line 133: This has the same issue as `POST`, since it combines `POST` and `GET` it technically should be `NestedMultiDict[str, str | cgi.FieldStorage]`, but at least you would not have to deal with `NoVars`, which is nice.

stubs/WebOb/webob/static.pyi
Daverball, line 1: I should probably add type annotations for this file, even though it's not the part of WebOb anybody likely cares about, it's very little work.

stubs/WebOb/webob/cookies.pyi
```diff
@@ -0,0 +1,129 @@
...
+        secure: bool = ...,
+        max_age: Incomplete | None = ...,
+        httponly: bool = ...,
+        samesite: Incomplete | None = ...,
+        path: str = ...,
+        domains: Incomplete | None = ...,
+        hashalg: str = ...,
+        serializer: Incomplete | None = ...,
+    ) -> None: ...
+    def bind(self, request): ...
```
Daverball, line 129: I should add type annotations for the Cookie API. I personally have never used it, but it does appear to contain some useful things.

stubs/WebOb/webob/static.pyi
AlexWaygood: You can do that in a followup PR. We encourage incremental PRs here, it makes things easier for us to review :)

stubs/WebOb/webob/multidict.pyi
AlexWaygood: Mypy can certainly use truthiness for type narrowing! https://mypy-play.net/?mypy=latest&python=3.11&gist=5322fe41b970986d96a392ab5e31f021

AlexWaygood:
Is this PR ready for proper review? If so, please take it out of draft :)

Daverball:
I'd like to play around with the annotations in a couple of different code bases first to see if there are any particularly rough edges or stuff that doesn't resolve as intended. I won't really have time to do that until later this week though.

stubs/WebOb/webob/multidict.pyi
Daverball: Nice, in my short experiment it didn't seem to narrow down the type union by doing exactly what you did in your simple example. I will definitely follow up on this, when I play around with the stubs in some code bases, since `if request.POST:` is a common pattern when using WebOb and deciding whether or not to pass the `POST` data forward to something like `WTForms` which will behave differently depending on whether it was passed something empty or nothing at all.

stubs/WebOb/webob/request.pyi
Daverball: I ended up switching to the type safe version that does take `cgi.FieldStorage` into account, but it uses a dummy subclass to let static analyzers know that `FieldStorage.file` and `FieldStorage.filename` are always set, to make it slightly less annoying to use. webob interally subclasses FieldStorage to slightly adjust the behavior, but since it does not change the way people interface with the object, it should not matter, as long as what we return passes `isinstance(fs, cgi.FieldStorage)`.

stubs/WebOb/webob/request.pyi
Daverball: Same resolution as with POST

stubs/WebOb/webob/cookies.pyi
Daverball: I did add some of the cookie annotations. The `CookieProfile` part of the API can be added in a follow-up PR.

Daverball:
I did some testing in various codebases and in a sandbox with `reveal_type` mostly with the Request/Response classes, since that is really the part that matters most. There were some issues with asymmetric properties that I had to solve by adding custom descriptors.

All in all I'm satisfied from a library users' perspective now, so this should be ready for a more general review.

stubs/WebOb/@tests/stubtest_allowlist.txt
```diff
@@ -0,0 +1,136 @@
...
+webob.acceptparse._AcceptInvalidOrNoHeader.copy
+webob.acceptparse._AcceptInvalidOrNoHeader.parsed
+webob.acceptparse._AcceptLanguageInvalidOrNoHeader.__add__
+webob.acceptparse._AcceptLanguageInvalidOrNoHeader.__radd__
+webob.acceptparse._AcceptLanguageInvalidOrNoHeader.copy
+webob.acceptparse._AcceptLanguageInvalidOrNoHeader.lookup
+webob.acceptparse._AcceptLanguageInvalidOrNoHeader.parsed
+
+# These are here due to the slightly more strict nature of the type annotation
+# of these desriptors for type checking, it does not really have any runtime
```
JelleZijlstra, line 87:
```suggestion
# of these descriptors for type checking, it does not really have any runtime
```

stubs/WebOb/@tests/stubtest_allowlist.txt
```diff
@@ -0,0 +1,136 @@
...
+webob.cachecontrol.CacheControl.max_stale
+webob.cachecontrol.CacheControl.min_fresh
+webob.cachecontrol.CacheControl.s_max_age
+webob.cachecontrol.CacheControl.s_maxage
+webob.cachecontrol.CacheControl.stale_if_error
+webob.cachecontrol.CacheControl.stale_while_revalidate
+webob.cachecontrol.CacheControl.update_dict
+webob.cachecontrol.UpdateDict.setdefault
+
+# These need to be ignored due to how WebOb decided to let people know that
```
JelleZijlstra, line 100:
```suggestion
# These need to be ignored due to how WebOb decided to let people know
```

stubs/WebOb/webob/acceptparse.pyi
```diff
@@ -0,0 +1,503 @@
...
+    @overload
+    def __get__(self, __obj: Any, __type: type | None = ...) -> AcceptNoHeader | AcceptValidHeader | AcceptInvalidHeader: ...
+    @overload
+    def __set__(self, __obj: Any, __value: str | None) -> None: ...
+    @overload
+    def __set__(self, __obj: Any, __value: AcceptNoHeader | AcceptValidHeader | AcceptInvalidHeader) -> None: ...
+    @overload
+    def __set__(self, __obj: Any, __value: SupportsItems[str, float | tuple[float, str]]) -> None: ...
+    @overload
+    def __set__(self, __obj: Any, __value: _ListOrTuple[str | tuple[str, float, str] | list[Any]]) -> None: ...
```
JelleZijlstra, line 123: Doesn't seem like this needs overloads, you can just use a union as the argument type for `__value`.

Daverball: You are completely correct, however in this case I think a bunch of overloads reads nicer than a very large type union. I've seen overloads used like this in other places in typeshed and liked how it looks compared to a very large type union where after maybe the fourth or fifth type in the union you just stop reading, especially once the types get as complicated as the last two overloads.

So here it's separated neatly into the plain string version of the header, then the specialized objects for representing the header from WebOb, then the Mapping like way to represent this data, and then the sequence like way to represent in.

You could make a case for combining the first two overloads, but I think the other two should stay distinct. I've also debated whether or not to add the final overload at all, because it isn't completely type safe and kind of confusing to read, it just happens to be one of the supported and documented ways to pass this data in.

stubs/WebOb/webob/static.pyi
```diff
@@ -0,0 +1,34 @@
...
+class DirectoryApp:
+    path: str | bytes
+    index_page: str | None
+    hide_index_with_redirect: bool
+    fileapp_kw: dict[str, Any]
+    def __init__(
+        self, path: StrOrBytesPath, index_page: str = "index.html", hide_index_with_redirect: bool = False, **kw: Any
+    ) -> None: ...
+    def make_fileapp(self, path: StrOrBytesPath) -> FileApp: ...
+    def __call__(self, req: Request) -> Response | FileApp: ...
```
Daverball, line 33: Same as above, this should use @wsgify

stubs/WebOb/webob/multidict.pyi
```diff
@@ -0,0 +1,114 @@
...
+    @overload
+    def __init__(self, data: SupportsItems[str, str], env: WSGIEnvironment) -> None: ...
+    @overload
+    def __init__(self, data: Iterable[tuple[str, str]], env: WSGIEnvironment) -> None: ...
+    def on_change(self) -> None: ...
+
+class NestedMultiDict(MultiDict[_KT, _VT]):
+    dicts: tuple[MultiDict[_KT, _VT]]
+    def __init__(self, *dicts: MultiDict[_KT, _VT]) -> None: ...
+    def __setitem__(self, key: _KT, value: _VT) -> NoReturn: ...
```
JelleZijlstra, line 77: I feel we should *not* use NoReturn for cases where a function always throws an exception to indicate it should not be called. `NoReturn` is meant only for functions that a user would expect to not return, like `sys.exit`, and its primary effect is meant to be to make code after it unreachable. I don't think that's what we want for these functions. See python/typing#1043.

In the meantime, I would suggest annotating these functions with their natural return types (e.g., `None`).

Daverball: That's fair, I would have to agree that unreachable errors are not a good way to let people know that they are not allowed to call a certain function/overload.

On `NoVars` I think we can do a little bit better though, by just removing the method altogether and adding an entry in `stubtest_allowlist.txt`. I would prefer to at least get a static error in functions that return `MultiDict | NoVars`. I am a little bit annoyed that WebOb opted for implementing the whole `MutableMultiDict` interface, just so they can give a slightly better error message at runtime if you try to mutate `NoVars` or retrieve a specific key from it (and to allow it to live inside a `NestedMultiDict`).

Neither `NestedMultiDict` nor `NoVars` are a `MutableMapping` (`NoVars` isn't even a `Mapping`, it just supports most of its methods, save for one) but currently both of them would pass as one. I wish they'd clean this mess up, although I understand why they wouldn't at this point, since it would break people's code everywhere they use `isinstance(x, MultiDict)` currently. It would need to be changed to a common abstract base class `MultiMapping` which both `MultiDict` and `NestedMultiDict` inherit from.

stubs/WebOb/webob/response.pyi
```diff
@@ -0,0 +1,180 @@
...
+class ResponseBodyFile:
+    mode: Literal["wb"]
+    closed: Literal[False]
+    response: Response
+    def __init__(self, response: Response): ...
+    @property
+    def encoding(self) -> str | None: ...
+    def write(self, text: str | bytes) -> int: ...
+    def writelines(self, seq: Sequence[str | bytes]) -> int: ...
+    def close(self) -> NoReturn: ...
```
JelleZijlstra, line 161: Let's remove NoReturn here too

stubs/WebOb/webob/response.pyi
```diff
@@ -0,0 +1,180 @@
...
+    def __iter__(self) -> Iterator[bytes]: ...
+    def next(self) -> bytes: ...
+    __next__ = next
+    def close(self) -> None: ...
+
+class EmptyResponse:
+    def __init__(self, app_iter: Iterator[bytes] | None = None) -> None: ...
+    def __iter__(self) -> Iterator[bytes]: ...
+    def __len__(self) -> int: ...
+    def next(self) -> NoReturn: ...
```
JelleZijlstra, line 179: And here

PR #9876, opened 2023-03-13
appdirs: improve annotation for appauthor argument

Jakob-Stadler:
According to the appdirs doc string and code [1], the appauthor argument in all functions supports False as argument value to skip insertion of author directories.

On a side note, appdirs has been recently deprecated in favor of platformdirs [2], not sure if and what steps have to be taken on the typeshed side.

I had some legacy code flag `user_data_dir(APP_NAME, appauthor=False)` (a false positive), so it should still be fixed in typeshed regardless.

[1] https://github.com/ActiveState/appdirs/blob/master/appdirs.py#L52

[2] https://github.com/ActiveState/appdirs/commit/8734277956c1df3b85385e6b308e954910533884


AlexWaygood:
> On a side note, appdirs has been recently deprecated in favor of platformdirs [2], not sure if and what steps have to be taken on the typeshed side.

Ah, thanks for flagging. I see `platformdirs` ships with a `py.typed` file, as well. In that case (while I'm happy to review/merge this PR first), we can probably mark these stubs as `no_longer_updated`, and remove them from typeshed pretty quickly. It would fall under the "If third-party stubs should be removed for other reasons" section in this policy here: https://github.com/python/typeshed/blob/main/CONTRIBUTING.md#third-party-library-removal-policy.

AlexWaygood (approved):
Thanks!

PR #9877, opened 2023-03-13
Mark `types-appdirs` as no longer updated

AlexWaygood:
`appdirs` has officially been deprecated in favour of `platformdirs`, a fork of `appdirs` that is maintained and ships with a `py.typed` file.

See https://github.com/ActiveState/appdirs/commit/8734277956c1df3b85385e6b308e954910533884

### DIFF
--- a/stubs/appdirs/METADATA.toml
+++ b/stubs/appdirs/METADATA.toml
@@ -1,1 +1,8 @@
 version = "1.4.*"
+no_longer_updated = true
+extra_description = """\
+The `appdirs` package has been deprecated in favour of the `platformdirs`\
+package, which ships with a `py.typed` file. Users should migrate their\
+`appdirs` code to use `platformdirs` instead, which will remove the need for\
+the `types-appdirs` package.\
+"""


srittau (approved):

stubs/appdirs/METADATA.toml
```diff
@@ -1 +1,8 @@
 version = "1.4.*"
+no_longer_updated = true
+extra_description = """\
+The `appdirs` package has been deprecated in favour of the `platformdirs`\
+package, which ships with a `py.typed` file. Users should migrate their\
+`appdirs` code to use `platformdirs` instead, which will remove the need for\
```
Akuli, line 6: Nit: This looks a little funny on pypi, because there's no spaces before the backslashes. For example, `for\` followed by `the` at the start of the next line shows up as `forthe` on pypi. https://pypi.org/project/types-appdirs/

AlexWaygood: Yeah, I noticed — my bad... We can push an update to PyPI adding the spaces before we remove the stubs entirely... Or we can just live with the typos for the dead stubs package...

Akuli: I'm fine with it either way :)

AlexWaygood: After I noticed my mistake, I also spotted a typo in the `extra_description` field for `types-gdb`: "is is" should be "it is": https://pypi.org/project/types-gdb/

AlexWaygood: Here you go :)

- https://github.com/python/typeshed/pull/9883

PR #9890, opened 2023-03-15
Add remaining types to setuptools.build_meta

henryiii:
This is an attempt to finish the incomplete types in `setuptools.build_meta`. This is really useful if you are "wrapping" the PEP 715 interface to add something (like a dynamic dependency.

Also adds the missing editable hooks (PEP 660), added in setuptools 62 IIRC.

Original source: https://github.com/pypa/setuptools/blob/main/setuptools/build_meta.py

Example of usage (currently full of type ignores): https://github.com/scikit-build/scikit-build-core/blob/main/src/scikit_build_core/setuptools/build_meta.py


stubs/setuptools/setuptools/build_meta.pyi
```diff
@@ -1,6 +1,6 @@
-from _typeshed import Incomplete
 from typing import Any
 
+from colletions.abc import Mapping
```
AlexWaygood, line 3:
```suggestion
from collections.abc import Mapping
```

henryiii: Oops, thanks! I'm trying to test this locally against scikit-build-core with mypy ATM. "--custom-typeshed-dir" doesn't work for stubs, I think, so I'm adding via stubs dir. I'll undraft once it works! :)

AlexWaygood: Yeah, for testing third-party stubs you have to use the `MYPYPATH` environment variable to point to the directory

henryiii:
Any idea how I can test this with a local copy of mypy? Mypy seems to refuse to use the typeshed stubs even if I add `--custom-typeshed-dir`, it it seems to produce `Skipping analyzing "setuptools.build_meta": module is installed, but missing library stubs or py.typed marker` (also the other subdirs) if I add it to the MYPYPATH.

stubs/setuptools/setuptools/build_meta.pyi
henryiii: Ahh, just saw this. Okay, can keep trying with MYPYPATH...

henryiii: Ahh, just realized these dirs are nested! Sorry for the noise.

henryiii:
Not a complete test of the API, but I was able to remove my type ignores and it's passing!

stubs/setuptools/setuptools/build_meta.pyi
```diff
@@ -14,20 +14,37 @@ class Distribution(dist.Distribution):
...
+def prepare_metadata_for_build_wheel(metadata_directory: str, config_settings: Mapping[str, Any] | None = ...) -> str: ...
+def build_wheel(
+    wheel_directory: str, config_settings: Mapping[str, Any] | None = ..., metadata_directory: str | None = ...
+) -> str: ...
+def build_sdist(sdist_directory: str, config_settings: Mapping[str, Any] | None = ...) -> str: ...
+def build_editable(
+    wheel_directory: str, config_settings: Mapping[str, Any] | None = ..., metadata_directory: str | None = ...
+) -> str: ...
+def get_requires_for_build_editable(config_settings: Mapping[str, Any] | None = ...) -> list[str]: ...
+def prepare_metadata_for_build_editable(metadata_directory: str, config_settings: Mapping[str, Any] | None = ...) -> str: ...
```
AlexWaygood, line 48: We can actually just mimic the runtime more closely here, to reduce code duplication:

```suggestion
_BACKEND: _BuildMetaBackend = ...
get_requires_for_build_wheel = _BACKEND.get_requires_for_build_wheel
get_requires_for_build_sdist = _BACKEND.get_requires_for_build_sdist
prepare_metadata_for_build_wheel = _BACKEND.prepare_metadata_for_build_wheel
build_wheel = _BACKEND.build_wheel
build_sdist = _BACKEND.build_sdist
```

AlexWaygood (commented):
Looks good! The stubs for the global functions can be simplified a bit, see my comment below.

For some of the parameters currently annotated with `str`, they will _strictly_ speaking also accept `os.PathLike` objects, since the arguments are immediately passed to `os.path.abspath`, which accepts path-like objects. But the fact that these parameters accept path-like objects now seems like an accident of the current implementation, rather than something that's designed to be supported. So let's stick with your current annotation of `str` for those objects.

AlexWaygood (approved):
Thanks!

stubs/setuptools/setuptools/build_meta.pyi
henryiii: Nice, wasn't sure that would work.

henryiii: Also, that's private, which is why I didn't add it originally.

AlexWaygood: > Why `_BACKEND: _BuildMetaBackend = ...` instead of `_BACKEND: _BuildMetaBackend`?

It's a little hack to appease flake8 -- pyflakes doesn't fully support `.pyi` files, and will erroneously complain about `_BuildMetaBackend` being undefined in the assignments that immediately follow the definition of `_BACKEND` if we just do `_BACKEND: _BuildMetaBackend` 🙂

AlexWaygood: > Also, that's private, which is why I didn't add it originally.

Yeah, our policy is basically "Don't include private stuff for the sake of including private stuff, but if it's necessary to improve the typing of public stuff, it's fine" (other exceptions also exist).

PR #9891, opened 2023-03-16
Add better types for predicate parameter to inspect.getx functions

Gobot1234:


stdlib/inspect.pyi
```diff
@@ -161,12 +161,20 @@ TPFLAGS_IS_ABSTRACT: Literal[1048576]
 
 modulesbyfile: dict[str, Any]
 
+_GetMembersPredicateTypeGuard: TypeAlias = Callable[[_T], TypeGuard[_T]]
```
AlexWaygood, line 164: Shouldn't this be:

```suggestion
_GetMembersPredicateTypeGuard: TypeAlias = Callable[[Any], TypeGuard[_T]]
```

I don't know of any `TypeGuard` functions that only accept objects that are already known to be of the type the `TypeGuard` function is verifying. If a type checker complains, you could try writing it as a generic callback protocol instead:

```python
class _GetMembersPredicateTypeGuard(Protocol[_T]):
    def __call__(self, __obj: Any) -> TypeGuard[_T]: ...
```

It might actually be more readable as a callback protocol, in this case.

AlexWaygood (commented):
I checked out this PR branch locally, and tried adding this test case to a new `test_cases/stdlib/check_inspect.py` file:

```python

import inspect
from typing import Any, List, Tuple
from typing_extensions import TypeGuard, assert_type

def foo(obj: Any) -> TypeGuard[list[str]]:
    return isinstance(obj, list) and all(isinstance(item, str) for item in obj)

class Bar: ...
x = Bar()

assert_type(inspect.getmembers(x), List[Tuple[str, Any]])
assert_type(inspect.getmembers(x, None), List[Tuple[str, Any]])
assert_type(inspect.getmembers(x, foo), List[Tuple[str, List[str]]])
```

It seems to pass pyright, but the last assertion fails on mypy (mypy infers `list[tuple[str, Any]]` for all three). I guess this is just because pyright has better handling of generic `TypeGuard`s in this situation?

AlexWaygood (approved):
I verified that this is a clear improvement for pyright users, and it doesn't lead to any _regressions_ for mypy users (mypy users just won't get the same benefits right now). So I'm happy for this to go in.

PR #9892, opened 2023-03-16
[openpyxl] Annotate Worksheet and other items

srittau:


srittau:
I'll update #9511 if this gets merged first.

stubs/openpyxl/openpyxl/worksheet/worksheet.pyi
```diff
@@ -49,7 +79,10 @@ class Worksheet(_WorkbookChild):
     @freeze_panes.setter
     def freeze_panes(self, topLeftCell: Incomplete | None = ...) -> None: ...
     def cell(self, row: int, column: int, value: str | None = ...) -> Cell: ...
-    def __getitem__(self, key: str | int | slice) -> Cell | tuple[Cell, ...]: ...
+    @overload
+    def __getitem__(self, key: int | slice) -> tuple[Cell, ...]: ...
+    @overload
+    def __getitem__(self, key: str) -> Any: ...  # Cell | tuple[Cell, ...]
```
srittau, line 85: Another instance where python/typing#566 would come in handy.

stubs/openpyxl/openpyxl/cell/cell.pyi
```diff
@@ -52,11 +52,11 @@ class Cell(StyleableObject):
...
     def check_error(self, value: object) -> str: ...
     @property
-    def value(self) -> str | float | datetime | None: ...
+    def value(self) -> str | float | date | time | timedelta | None: ...
     @value.setter
-    def value(self, value: str | float | datetime | None) -> None: ...
+    def value(self, value: str | float | date | time | timedelta | None) -> None: ...
     @property
-    def internal_value(self) -> str | float | datetime | None: ...
+    def internal_value(self) -> str | float | date | time | timedelta | None: ...
```
AlexWaygood, line 59: According to @Avasam's annotations in #9511, these could also be `bytes` or `decimal.Decimal`

stubs/openpyxl/openpyxl/utils/bound_dictionary.pyi
```diff
@@ -1,7 +1,9 @@
...
+from typing import Generic, TypeVar
 
-class BoundDictionary(defaultdict[Incomplete, Incomplete]):
-    reference: Incomplete
-    def __init__(self, reference: Incomplete | None = ..., *args, **kw) -> None: ...
-    def __getitem__(self, key): ...
+_KT = TypeVar("_KT")
+_VT = TypeVar("_VT")
+
+class BoundDictionary(defaultdict[_KT, _VT], Generic[_KT, _VT]):
```
AlexWaygood, line 7: (Non-blocking style nit): the multiple inheritance with `Generic` here isn't necessary (the class is generic by virtue of inheriting from `defaultdict`, which is generic), and I'd prefer to not have it. But I don't feel too strongly about this; you could argue that mixing in `Generic` here makes it more explicit.

```suggestion
class BoundDictionary(defaultdict[_KT, _VT]):
```

stubs/openpyxl/openpyxl/utils/bound_dictionary.pyi
```diff
@@ -1,7 +1,9 @@
...
+    reference: str | None
+    def __init__(self, reference: str | None = None, *args, **kw) -> None: ...
```
AlexWaygood, line 9: The `*args, **kwargs` here just forward on the arguments to `defaultdict.__init__`, but we have [8 overloads for that](https://github.com/python/typeshed/blob/f6f1fa3df4d41d5267a4440204b3d0d3c198e9ed/stdlib/collections/__init__.pyi#L363-L388). So I'm fine with keeping this as it is for now.

stubs/openpyxl/openpyxl/worksheet/dimensions.pyi
```diff
@@ -1,95 +1,118 @@
...
 
 class Dimension(Strict, StyleableObject):
-    __fields__: Incomplete
-    index: Incomplete
-    hidden: Incomplete
-    outlineLevel: Incomplete
-    outline_level: Incomplete
-    collapsed: Incomplete
-    style: Incomplete
+    __fields__: tuple[str, ...]
```
AlexWaygood, line 17:
```suggestion
    __fields__: ClassVar[tuple[str, ...]]
```

stubs/openpyxl/openpyxl/worksheet/dimensions.pyi
```diff
@@ -1,95 +1,118 @@
...
+        hidden: bool,
+        outlineLevel: int | None,
+        collapsed: bool,
+        worksheet: Worksheet,
+        visible: bool = True,
+        style: Incomplete | None = None,
     ) -> None: ...
-    def __iter__(self): ...
-    def __copy__(self): ...
+    def __iter__(self) -> Generator[tuple[str, str], None, None]: ...
```
AlexWaygood, line 36: I'd prefer the simpler annotation here, as it seems to me that it's an implementation detail that this returns a generator as opposed to some other kind of iterator. But I won't insist :)

```suggestion
    def __iter__(self) -> Iterator[tuple[str, str]]: ...
```

stubs/openpyxl/openpyxl/worksheet/dimensions.pyi
```diff
@@ -1,95 +1,118 @@
...
 
 class RowDimension(Dimension):
-    __fields__: Incomplete
-    r: Incomplete
-    s: Incomplete
-    ht: Incomplete
-    height: Incomplete
-    thickBot: Incomplete
-    thickTop: Incomplete
+    __fields__: tuple[str, ...]
```
AlexWaygood, line 40: This has the same type as in the superclass; no need to duplicate it here

```suggestion
```

stubs/openpyxl/openpyxl/worksheet/dimensions.pyi
```diff
@@ -1,95 +1,118 @@
...
+
+    r: Alias
+    s: Alias
+    ht: Float
+    height: Alias
+    thickBot: Bool
+    thickTop: Bool
```
AlexWaygood, line 47: How useful are these `Alias` annotations? We don't currently have a return-type annotation for `openpyxl.descriptors.base.Alias.__get__`. But even if we did, I think the only accurate return type we could give it would be `Any`, since there's no way of expressing statically "this is a descriptor for aliasing an attribute to another arbitrary attribute" (unless we made `Alias` generic in the stubs, I suppose).

https://github.com/python/typeshed/blob/f6f1fa3df4d41d5267a4440204b3d0d3c198e9ed/stubs/openpyxl/openpyxl/descriptors/base.pyi#L71-L75

We might be better off using simpler annotations for now, which will accurately reflect the types users will get at runtime if they access these attributes on instances of `RowDimension`:

```suggestion
    r: int
    s: int
    ht: Float
    height: float
    thickBot: Bool
    thickTop: Bool
```

stubs/openpyxl/openpyxl/worksheet/dimensions.pyi
```diff
@@ -1,95 +1,118 @@
...
-    bestFit: Incomplete
-    auto_size: Incomplete
-    index: Incomplete
-    min: Incomplete
-    max: Incomplete
-    collapsed: Incomplete
-    __fields__: Incomplete
+    width: Float
+    bestFit: Bool
+    auto_size: Alias
```
AlexWaygood, line 77:
```suggestion
    auto_size: bool
```

stubs/openpyxl/openpyxl/worksheet/dimensions.pyi
```diff
@@ -1,95 +1,118 @@
...
+    index: String  # type: ignore[assignment]
+    min: Integer
+    max: Integer
+    collapsed: Bool
+    __fields__: tuple[str, ...]
```
AlexWaygood, line 82: Same as in the superclass

```suggestion
```

stubs/openpyxl/openpyxl/worksheet/dimensions.pyi
```diff
@@ -1,95 +1,118 @@
...
-class DimensionHolder(BoundDictionary):
-    worksheet: Incomplete
-    max_outline: Incomplete
-    default_factory: Incomplete
-    def __init__(self, worksheet, reference: str = ..., default_factory: Incomplete | None = ...) -> None: ...
-    def group(self, start, end: Incomplete | None = ..., outline_level: int = ..., hidden: bool = ...) -> None: ...
-    def to_tree(self): ...
+class DimensionHolder(BoundDictionary[str, _DimT], Generic[_DimT]):
+    worksheet: Worksheet
+    max_outline: Incomplete | None
```
AlexWaygood, line 108: https://foss.heptapod.net/openpyxl/openpyxl/-/blob/branch/3.2/openpyxl/worksheet/dimensions.py#L240

```suggestion
    max_outline: int | None
```

stubs/openpyxl/openpyxl/worksheet/worksheet.pyi
```diff
@@ -102,9 +135,9 @@ class Worksheet(_WorkbookChild):
         values_only: bool,
     ) -> Generator[tuple[Cell | str | float | datetime | None, ...], None, None]: ...
     @property
-    def rows(self) -> Generator[Cell, None, None]: ...
+    def rows(self) -> Generator[tuple[Cell, ...], None, None]: ...
     @property
-    def values(self) -> Generator[str | float | datetime | None, None, None]: ...
+    def values(self) -> Generator[tuple[str | float | datetime | None, ...], None, None]: ...
```
AlexWaygood, line 140: @Avasam's annotations for these are slightly different:

https://github.com/python/typeshed/pull/9511/files#diff-dabc24e406450b802c51dd6c9d1062ac0fed975dc4daef0e9cbe468aedcb6978L103-R114

AlexWaygood (commented):
Thanks, a few comments:

stubs/openpyxl/openpyxl/cell/cell.pyi
Avasam: Unless you can show this can be restricted further, it does seem like anything valid for the function `get_type` (line ~64) can be a valid cell value type.

srittau: I've extracted a type alias and added all types (except numpy types).

stubs/openpyxl/openpyxl/worksheet/dimensions.pyi
srittau: Making `Alias` generic is probably the best course of action, although I'd consider this out of scope for this PR.

stubs/openpyxl/openpyxl/cell/rich_text.pyi
```diff
@@ -0,0 +1,24 @@
+from typing import Protocol, overload
+from typing_extensions import Self
+
+from openpyxl.descriptors import Strict, String, Typed
+
+class _HasStr(Protocol):
```
AlexWaygood, line 6: As `object` defines `__str__`, I'm not sure it's possible to come up with a class that _doesn't_ implement this protocol from a type checker's point of view

srittau: We could just use `object` here instead. But I'd at least alias `_HasStr` to `object` to make the intention clear. (Even if the implementation checking for `__str__` is nonsense.)

AlexWaygood: > But I'd at least alias `_HasStr` to `object` to make the intention clear.

I'd be much happier with that — it would be more obvious that we knew what we were doing, rather than it "looking like an obvious mistake" :)

Though `list[object]` in a parameter annotation will be pretty problematic due to invariance, so `Any` might be better

stubs/openpyxl/openpyxl/worksheet/dimensions.pyi
AlexWaygood: Okay, we can leave this for now

stubs/openpyxl/openpyxl/cell/rich_text.pyi
```diff
@@ -1,19 +1,20 @@
...
-    def __str__(self) -> str: ...  # noqa: Y029
-
 class TextBlock(Strict):
     font: Typed
     text: String
 
     def __init__(self, font: Typed, text: String) -> None: ...
     def __eq__(self, other: TextBlock) -> bool: ...  # type: ignore[override]
 
+# CellRichText accepts any object, which can be converted to a string,
```
AlexWaygood, line 14: Nit:

```suggestion
# CellRichText accepts any object that can be converted to a string,
```

stubs/openpyxl/openpyxl/cell/rich_text.pyi
```diff
@@ -0,0 +1,28 @@
...
+    @overload
+    def __init__(self, __args: list[_HasStr] | tuple[_HasStr]) -> None: ...
+    @overload
+    def __init__(self, *args: _HasStr) -> None: ...
+    @classmethod
+    def from_tree(cls, node) -> Self: ...
+    def __add__(self, arg: _HasStr) -> CellRichText: ...
+    def append(self, arg: _HasStr) -> None: ...
+    def extend(self, arg: Iterable[_HasStr]) -> None: ...
+    def as_list(self) -> list[str]: ...
```
AlexWaygood, line 28: Sorry for going back and forth on this. I just took another look at the source code and, while the implementation is indeed pretty permissive, the docstring and comments are fairly clear that the members of the list should be either "simple strings or TextBlock() instances". So maybe we could be stricter in our annotations here:

```suggestion
class CellRichText(list[str | TextBlock]):
    @overload
    def __init__(self, __args: list[str] | list[TextBlock] | list[str | TextBlock] | tuple[str | TextBlock, ...]) -> None: ...
    @overload
    def __init__(self, *args: str | TextBlock) -> None: ...
    @classmethod
    def from_tree(cls, node) -> Self: ...
    def __add__(self, arg: Iterable[str | TextBlock]) -> CellRichText: ...
    def append(self, arg: str | TextBlock) -> None: ...
    def extend(self, arg: Iterable[str | TextBlock]) -> None: ...
    def as_list(self) -> list[str]: ...
```

srittau: Sounds like a good idea.

AlexWaygood (approved):
LGTM!

PR #9893, opened 2023-03-16
fix: add missing distribution to distutils cmd

henryiii:
This is the first thing `__init__` does with distribution, and it was missing.


henryiii:
The missing attribute stuff should be fixed by #9895.

henryiii:
FYI, these PRs caught a mistake already: https://github.com/scikit-build/scikit-build-core/pull/231/files#r1140613496 :)

PR #9895, opened 2023-03-16
fix: fill out more of distutils & setuptools dist

henryiii:
This was an attempt to fill out more of the typing for dist. I've re-synced `setuptools/_distutils/dist.py` with `distutils/dist.py`, as the vendored version was missing a lot of info from the other one.


stubs/setuptools/setuptools/dist.pyi
```diff
@@ -1,36 +1,39 @@
...
-    def has_contents_for(self, package): ...
+    def exclude_package(self, package: str) -> None: ...
+    def has_contents_for(self, package: str) -> bool | None: ...
     def exclude(self, **attrs) -> None: ...
-    def get_cmdline_options(self): ...
-    def iter_distribution_names(self) -> None: ...
+    def get_cmdline_options(self) -> dict[str, dict[str, str | None]]: ...
+    def iter_distribution_names(self) -> Generator[str, None, None]: ...
     def handle_display_options(self, option_order): ...
+    def run_command(self, command: str) -> None: ...
```
henryiii, line 37: This last one was in setuptools but missing here.

stubs/setuptools/setuptools/_distutils/dist.pyi
```diff
@@ -54,7 +57,84 @@ class DistributionMetadata:
 class Distribution:
     cmdclass: dict[str, type[Command]]
     metadata: DistributionMetadata
-    def __init__(self, attrs: Mapping[str, Incomplete] | None = ...) -> None: ...
+    def __init__(self, attrs: Mapping[str, Any] | None = ...) -> None: ...
```
henryiii, line 60: I don't think it's possible to know all attrs, so I _think_ this is Any, not Incomplete.

stdlib/distutils/dist.pyi
```diff
@@ -101,16 +101,39 @@ class Distribution:
...
+    def has_pure_modules(self) -> bool: ...
+    def has_ext_modules(self) -> bool: ...
+    def has_c_libraries(self) -> bool: ...
+    def has_modules(self) -> bool: ...
+    def has_headers(self) -> bool: ...
+    def has_scripts(self) -> bool: ...
+    def has_data_files(self) -> bool: ...
+    def is_pure(self) -> bool: ...
+
+    # Autogenerated getters
```
henryiii, line 118: These are all generated by distutils but were missing here.

henryiii:
Is it a good idea to include something in a stub if it's identical type-wise to the base class? That is:

```python
class A:
    def f() -> None:
        return None

class(B):
    def f() -> None:
       return super().f()
```

Should that be:

```python
class A:
    def f() -> None: ...

class(B):
    def f() -> None: ...
```

Or

```python
class A:
    def f() -> None: ...

class(B): ...
```

? This happens quite a bit in setuptool's Distribution.

AlexWaygood:
> Is it a good idea to include something in a stub if it's identical type-wise to the base class? That is:

We generally omit methods on the subclass if they have identical signatures to the methods on the superclass. But there are a few exceptions (mostly for dunders) — e.g. `__eq__` and `__format__` should always be included on the subclass, for Reasons

AlexWaygood:
But if the signature differs on the subclass, you should definitely include it. If it's incompatible with the superclass, you should add a `type: ignore`: it's not our job to ensure that the runtime adheres to LSP :)

henryiii:
Is the stubtest taking into account side effects of running `__init__`? It seems like it has to, but it's claiming the generated methods (https://github.com/pypa/setuptools/blob/be6c0218bcba78dbd4ea0b5a8bb9acd5d5306240/setuptools/_distutils/dist.py#L151-L153) don't exist at runtime.

AlexWaygood:
Stubtest works by importing the module at runtime and dynamically inspecting the objects it finds in it, comparing the results to what it sees in the stub. So it often struggles with methods generated inside `__init__` methods, since it never actually instantiates any of the classes at runtime :)

If need be, we will just have to add allowlist entries to tell it to be quiet about these methods.

(Note: I haven't studied your proposed changes here yet; this is just a direct response to the question you posed above :)

henryiii:
Okay, that would explain it. I'd have expected something simple like assigning inside an init method to also have a problem (thought that's not usually making a method, perhaps it's only checking methods?) Anyway, I've pulled that out separately, as I think it's easier to review on it's own.

AlexWaygood:
> perhaps it's only checking methods?

Yup, 99% of methods are statically defined, so if it can't see a method in the stub at runtime, stubtest assumes you've made a mistake and emits an error. But 99% of instance attributes are assigned in `__init__` methods, so if there's an instance attribute in the stub that it can't see at runtime, it generally _won't_ emit an error.

Stubtest is basically just a series of hardcoded consistency checks; it doesn't have a generalised way of comparing the runtime to the stub. But it's very effective at what it does (it just has a few edge cases, and you've stumbled on one of them!).

stdlib/distutils/dist.pyi
```diff
@@ -7,7 +7,7 @@ from typing import IO, Any
 command_re: Pattern[str]
 
 class DistributionMetadata:
-    def __init__(self, path: FileDescriptorOrPath | None = None) -> None: ...
+    def __init__(self, path: FileDescriptorOrPath | None = ...) -> None: ...
```
AlexWaygood, line 10: Why have you removed the default value here, and from several other functions in this PR? We've been trying to add default values where possible over the last few months, as they're very helpful for IDE users. It's also helped us expose a number of latent bugs in typeshed that previously went unnoticed (e.g. see #9608).

henryiii: Oops, I thought it was preferable to avoid coping the default, and this was due to copying. I'll revert that.

AlexWaygood: We changed our policy fairly recently on this, no worries :)

henryiii: In my defense, these were inconsistent between the two files, so I was just making this one match the other one.

AlexWaygood: Yes, well, we haven't finished deciding exactly how many defaults we want added to the stdlib yet (https://github.com/PyCQA/flake8-pyi/issues/357) — give us time to move on to the third-party stubs!!

stdlib/distutils/dist.pyi
```diff
@@ -58,21 +58,21 @@ class Distribution:
     metadata: DistributionMetadata
     def __init__(self, attrs: Mapping[str, Any] | None = None) -> None: ...
     def get_option_dict(self, command: str) -> dict[str, tuple[str, str]]: ...
-    def parse_config_files(self, filenames: Iterable[str] | None = None) -> None: ...
-    def get_command_obj(self, command: str, create: bool = ...) -> Command | None: ...
-    global_options: Incomplete
+    def parse_config_files(self, filenames: Iterable[str] | None = ...) -> None: ...
```
AlexWaygood, line 61:
```suggestion
    def parse_config_files(self, filenames: Iterable[str] | None = None) -> None: ...
```

stdlib/distutils/dist.pyi
```diff
@@ -58,21 +58,21 @@ class Distribution:
...
+    def get_command_obj(self, command: str, create: bool = None) -> Command | None: ...
```
henryiii, line 62:
```suggestion
    def get_command_obj(self, command: str, create: bool = True) -> Command | None: ...
```

stubs/setuptools/setuptools/_distutils/dist.pyi
```diff
@@ -54,7 +57,61 @@ class DistributionMetadata:
 class Distribution:
     cmdclass: dict[str, type[Command]]
     metadata: DistributionMetadata
-    def __init__(self, attrs: Mapping[str, Incomplete] | None = ...) -> None: ...
+    def __init__(self, attrs: Mapping[str, Any] | None = None) -> None: ...
     def get_option_dict(self, command: str) -> dict[str, tuple[str, str]]: ...
     def parse_config_files(self, filenames: Iterable[str] | None = ...) -> None: ...
-    def get_command_obj(self, command: str, create: bool = ...) -> Command | None: ...
+    def get_command_obj(self, command: str, create: bool = None) -> Command | None: ...
```
henryiii, line 63:
```suggestion
    def get_command_obj(self, command: str, create: bool = True) -> Command | None: ...
```

stubs/setuptools/setuptools/_distutils/dist.pyi
```diff
@@ -54,7 +57,61 @@ class DistributionMetadata:
 class Distribution:
     cmdclass: dict[str, type[Command]]
     metadata: DistributionMetadata
-    def __init__(self, attrs: Mapping[str, Incomplete] | None = ...) -> None: ...
+    def __init__(self, attrs: Mapping[str, Any] | None = None) -> None: ...
     def get_option_dict(self, command: str) -> dict[str, tuple[str, str]]: ...
     def parse_config_files(self, filenames: Iterable[str] | None = ...) -> None: ...
```
henryiii, line 62:
```suggestion
    def parse_config_files(self, filenames: Iterable[str] | None = None) -> None: ...
```

henryiii:
```
distutils.core.Distribution.get_command_obj is inconsistent, runtime argument "create" has a default value of 1, which is different from stub argument default True
```

Is the there a way to silence that? Or should I just use `...` for that one? It's a bool argument but it using 1, probably for Reasons. :)

stdlib/distutils/dist.pyi
```diff
@@ -59,20 +59,20 @@ class Distribution:
     def __init__(self, attrs: Mapping[str, Any] | None = None) -> None: ...
     def get_option_dict(self, command: str) -> dict[str, tuple[str, str]]: ...
     def parse_config_files(self, filenames: Iterable[str] | None = None) -> None: ...
-    def get_command_obj(self, command: str, create: bool = ...) -> Command | None: ...
-    global_options: Incomplete
+    def get_command_obj(self, command: str, create: bool = True) -> Command | None: ...
```
AlexWaygood, line 62: You'll have to add an allowlist entry for this method. Stubtest is, strictly speaking, correct that the default value is `1` rather than `True`. But if we change the default value to `1`, then we have to change the annotation to `int`. And we don't want to change the annotation here, because users Really Should be passing `bool`s to this parameter -- it only has a default value of `1` for historical reasons.

The stdlib allowlist is here: https://github.com/python/typeshed/blob/main/tests/stubtest_allowlists/py3_common.txt

`setuptool` has its own allowlist, which is here: https://github.com/python/typeshed/blob/main/stubs/setuptools/%40tests/stubtest_allowlist.txt

AlexWaygood: We did a similar thing with some `tkinter` functions here, that had `1` instead of `True` as the default for historical reasons: #9637

stdlib/distutils/dist.pyi
```diff
@@ -59,20 +59,20 @@ class Distribution:
...
+    global_options: list[tuple[str, str | None, str, int] | tuple[str, str | None, str]]
     common_usage: str
-    display_options: Incomplete
-    display_option_names: Incomplete
-    negative_opt: Incomplete
+    display_options: list[tuple[str, str | None, str, int] | tuple[str, str | None, str]]
+    display_option_names: list[str]
+    negative_opt: dict[str, str]
```
AlexWaygood, line 67: If `global_options` and `display_options` are the same type, it might be worth adding a `TypeAlias` somewhere in the global scope:

```python
_OptionsList: TypeAlias = list[tuple[str, str | None, str, int] | tuple[str, str | None, str]]
```

And then both `global_options` and `display_options` could be annotated as `ClassVar[_OptionsList]`. It's a complicated type to repeat twice, and adding the type alias makes it easier to see at a glance that they're the same type.

```suggestion
    display_options: ClassVar[list[tuple[str, str | None, str, int] | tuple[str, str | None, str]]]
    display_option_names: ClassVar[list[str]]
    negative_opt: ClassVar[dict[str, str]]
```

AlexWaygood (commented):
Thanks! Haven't looked in depth at the `setuptools` changes yet, but here's a few points on the `distutils` changes -- I think a few of them will also apply to some of the `setuptools` changes

stdlib/distutils/dist.pyi
```diff
@@ -101,21 +103,21 @@ class Distribution:
     def print_commands(self) -> None: ...
     def get_command_list(self): ...
     def get_command_packages(self): ...
-    def get_command_class(self, command): ...
-    def reinitialize_command(self, command, reinit_subcommands: int = 0): ...
+    def get_command_class(self, command: str | Command) -> type[Command]: ...
```
JelleZijlstra, line 106: It doesn't look like this works with Command objects: https://github.com/python/cpython/blob/d025b1d2785ced8d436246aaf7627aa2c33493d4/Lib/distutils/dist.py#L807

stdlib/distutils/dist.pyi
```diff
@@ -101,21 +103,21 @@ class Distribution:
...
+    def reinitialize_command(self, command: str | Command, reinit_subcommands: int = 0) -> Command: ...
```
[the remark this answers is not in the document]
henryiii: I thought about this. I was slightly tempted to do that, but was on the side of "not worth it yet". I can, though.

stubs/setuptools/setuptools/_distutils/dist.pyi
```diff
@@ -54,12 +59,66 @@ class DistributionMetadata:
...
+    def find_config_files(self): ...
+    commands: Incomplete
+    def parse_command_line(self): ...
+    def finalize_options(self) -> None: ...
+    def handle_display_options(self, option_order): ...
+    def print_command_list(self, commands, header, max_length) -> None: ...
+    def print_commands(self) -> None: ...
+    def get_command_list(self): ...
+    def get_command_packages(self): ...
+    def get_command_class(self, command: str | Command) -> type[Command]: ...
```
JelleZijlstra, line 107:
```suggestion
    def get_command_class(self, command: str) -> type[Command]: ...
```

stubs/setuptools/setuptools/dist.pyi
```diff
@@ -1,36 +1,30 @@
...
+    def warn_dash_deprecation(self, opt: str, section: str) -> str: ...
+    def make_option_lowercase(self, opt: str, section: str) -> str: ...
+    def parse_config_files(self, filenames: Iterable[str] | None = ..., ignore_option_errors: bool = ...) -> None: ...
+    def fetch_build_eggs(self, requires: str | Iterable[str]): ...
+    def get_egg_cache_dir(self) -> str: ...
     def fetch_build_egg(self, req): ...
-    def get_command_class(self, command): ...
-    def print_commands(self): ...
-    def get_command_list(self): ...
+    def get_command_class(self, command: str | Command) -> type[Command]: ...
```
JelleZijlstra, line 20:
```suggestion
    def get_command_class(self, command: str) -> type[Command]: ...
```

stdlib/distutils/dist.pyi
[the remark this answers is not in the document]
AlexWaygood: Eh, why not, people rarely complain about their type checkers having inferences that are _too_ precise :p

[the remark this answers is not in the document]
henryiii: Okay, will do in a few minutes.

stubs/setuptools/setuptools/_distutils/dist.pyi
```diff
@@ -54,12 +61,69 @@ class DistributionMetadata:
...
+    def handle_display_options(self, option_order): ...
+    def print_command_list(self, commands, header, max_length) -> None: ...
+    def print_commands(self) -> None: ...
+    def get_command_list(self): ...
+    def get_command_packages(self): ...
+    def get_command_class(self, command: str) -> type[Command]: ...
+    @overload
+    def reinitialize_command(self, command: str, reinit_subcommands: int = 0) -> Command: ...
+    @overload
+    def reinitialize_command(self, command: _CommandT, reinit_subcommands: int = 0) -> _CommandT: ...
```
AlexWaygood, line 113:
```suggestion
    def reinitialize_command(self, command: str, reinit_subcommands: bool = False) -> Command: ...
    @overload
    def reinitialize_command(self, command: _CommandT, reinit_subcommands: bool = False) -> _CommandT: ...
```

stubs/setuptools/setuptools/dist.pyi
```diff
@@ -1,36 +1,29 @@
...
-    ext_modules: Incomplete
-    def exclude_package(self, package) -> None: ...
-    def has_contents_for(self, package): ...
+    def exclude_package(self, package: str) -> None: ...
+    def has_contents_for(self, package: str) -> bool | None: ...
     def exclude(self, **attrs) -> None: ...
-    def get_cmdline_options(self): ...
-    def iter_distribution_names(self) -> None: ...
+    def get_cmdline_options(self) -> dict[str, dict[str, str | None]]: ...
+    def iter_distribution_names(self) -> Generator[str, None, None]: ...
```
AlexWaygood, line 26: I'd prefer to use the more general `Iterator` type here -- it's more readable, and the fact that it's specifically a generator feels like an implementation detail

```suggestion
    def iter_distribution_names(self) -> Iterator[str]: ...
```

stubs/setuptools/setuptools/_distutils/dist.pyi
```diff
@@ -54,12 +61,69 @@ class DistributionMetadata:
...
+    def announce(self, msg, level: int = ...) -> None: ...
+    def run_commands(self) -> None: ...
+    def run_command(self, command: str | Command) -> None: ...
```
AlexWaygood, line 116:
```suggestion
    def run_command(self, command: str) -> None: ...
```

AlexWaygood (commented):
Thanks, nearly there! A _few_ more nits:

stdlib/distutils/dist.pyi
```diff
@@ -101,21 +105,24 @@ class Distribution:
...
-    def run_command(self, command) -> None: ...
-    def has_pure_modules(self): ...
-    def has_ext_modules(self): ...
-    def has_c_libraries(self): ...
-    def has_modules(self): ...
-    def has_headers(self): ...
-    def has_scripts(self): ...
-    def has_data_files(self): ...
-    def is_pure(self): ...
+    def run_command(self, command: str | Command) -> None: ...
```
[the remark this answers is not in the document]
henryiii: When I wrote it, I had typed `get_command_obj` with `str | Command`.

stubs/setuptools/setuptools/dist.pyi
henryiii: I'm a big fan of using `Generator`, as `Iterator` is not correct (https://github.com/python/typeshed/issues/2772), but I'm fine to stay with the current style, and this case won't affect anything.

AlexWaygood (approved):
LGTM, thanks!

PR #9896, opened 2023-03-16
chore: add generated methods

henryiii:
Pulled out from #9895. Not sure why stubtest can't see these, they've been there forever:

https://github.com/pypa/setuptools/blob/be6c0218bcba78dbd4ea0b5a8bb9acd5d5306240/setuptools/_distutils/dist.py#L151-L153

https://github.com/python/cpython/blob/1bf9cc509326bc42cd8cb1650eb9bf64550d817e/Lib/distutils/dist.py#L159-L162

Looks like subtest just needs this added to some sort of ignore list.


srittau (approved):
Thanks!

PR #9897, opened 2023-03-16
Fix types in pycocotools

ClementPinard:
:wave: Hello there

Could not figure out why my type checker (pyright for that matter) was failing in my own code.

It turns out that some typings in pycocotools are wrong.

Namely :

COCOeval.eval is an entirely different dict from `_EvaluationResult`, as it can be seen here : https://github.com/cocodataset/cocoapi/blob/master/PythonAPI/pycocotools/cocoeval.py#L411

As such, I renamed the original `_EvaluationResult` into `_ImageEvaluationResult` and defined a new typed dict with the original name. Note that we could keep the original typed dict as is and create a new typed dict named something like `_AccumulatedEvaluationResult`, I don't have a preference.

I also could see that Params.areaRng is not a list, it's a list of list, as it can be seen here : https://github.com/cocodataset/cocoapi/blob/master/PythonAPI/pycocotools/cocoeval.py#L509

Note that it's actually either an int or a float. not sure if I should put `int | float`  or simply `float` in that case.


stubs/pycocotools/pycocotools/cocoeval.pyi
```diff
@@ -53,7 +64,7 @@ class Params:
     recThrs: _NDArray
     # recThrs: npt.NDArray[np.float64]
     maxDets: list[int]
-    areaRng: list[int]
+    areaRng: list[list[int | float]]
```
srittau, line 67: Nit: For typing purposes, `float` is a super-type of `int`, so the following is identical and preferred:

```suggestion
    areaRng: list[list[float]]
```

srittau (changes requested):
Thanks!

Could you also change the return type of `evaluateImg()` to the new name?

ClementPinard:
Ah yes you're right, sorry for missing that.

thanks for the review !

srittau (approved):

PR #9899, opened 2023-03-17
setuptools: fill out extension's constructor

henryiii:
Filling out the extension constructor. It passes through *args, **kwargs, but the distutils constructor is fully specified.


srittau (approved):
LGTM, thanks!

PR #9900, opened 2023-03-17
distutils/setuptools: Don't use wildcards for allowlist entries

AlexWaygood:
These entries are causing more functions to be allowlisted than was intended: for example, none of these methods are generated methods, but they're all currently included by the `distutils.dist.Distribution.get_.*` entry:

https://github.com/python/typeshed/blob/2d990ee2f714c1902620586efed576b4f7edaba7/stdlib/distutils/dist.pyi#L106-L108

This caused a bit of confusion for me, as I was fiddling around with a stubtest patch, and expected stubtest to emit an error if I deleted this allowlist entry:

https://github.com/python/typeshed/blob/2d990ee2f714c1902620586efed576b4f7edaba7/tests/stubtest_allowlists/py3_common.txt#L89

...But stubtest didn't emit an error, because `distutils.dist.Distribution.get_command_obj` was _still_ allowlisted, even with that entry deleted, due to these regex entries here.

### DIFF
--- a/stubs/setuptools/@tests/stubtest_allowlist.txt
+++ b/stubs/setuptools/@tests/stubtest_allowlist.txt
@@ -31,7 +31,27 @@ pkg_resources.EggMetadata.loader
 setuptools._distutils.dist.Distribution.get_command_obj
 
 # Dynamically created in __init__
-setuptools._distutils.dist.Distribution.get_.*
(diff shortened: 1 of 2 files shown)


hauntsaninja (approved):
Thanks!

PR #9915, opened 2023-03-21
Support `SupportsIndex` in constructors for all `datetime` objects

sobolevn:
Note, that `fromisocalendar` does not support objects with `__index__` method.

### DIFF
--- a/stdlib/datetime.pyi
+++ b/stdlib/datetime.pyi
@@ -1,9 +1,8 @@
 import sys
-from _typeshed import SupportsIndex
 from abc import abstractmethod
 from time import struct_time
 from typing import ClassVar, NamedTuple, NoReturn, TypeVar, overload
-from typing_extensions import Literal, Self, TypeAlias, final
+from typing_extensions import Literal, Self, SupportsIndex, TypeAlias, final
 
 if sys.version_info >= (3, 11):
     __all__ = ("date", "datetime", "time", "timedelta", "timezone", "tzinfo", "MINYEAR", "MAXYEAR", "UTC")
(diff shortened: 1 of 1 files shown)


stdlib/datetime.pyi
```diff
@@ -1,4 +1,5 @@
 import sys
+from _typeshed import SupportsIndex
```
AlexWaygood, line 2:
```suggestion
from typing_extensions import SupportsIndex
```

stdlib/datetime.pyi
```diff
@@ -10,6 +11,7 @@ elif sys.version_info >= (3, 9):
     __all__ = ("date", "datetime", "time", "timedelta", "timezone", "tzinfo", "MINYEAR", "MAXYEAR")
 
 _D = TypeVar("_D", bound=date)
+_IntLike: TypeAlias = int | SupportsIndex
```
AlexWaygood, line 14: I don't think this is necessary -- `int`s have an `__index__` method:

```pycon
>>> x = 3
>>> x.__index__()
3
```

You can just use `SupportsIndex`, and it will allow `int`s

AlexWaygood: > I kept it for extra readability.

To me it makes it less readable, as it causes confusion -- I'm left wondering why there's a redundant union here, since I know that `int` is a subtype of `SupportsIndex`.

> `int` is the main data-type, `SupportsIndex` is the second option.

Sure, but that applies in 99% of cases where the runtime accepts anything with an `__index__` method, and we don't use a special type alias for any of those:

https://github.com/python/typeshed/blob/b6564d976b336ddff962870092ac873a60dfa9b2/stdlib/builtins.pyi#L291

https://github.com/python/typeshed/blob/b6564d976b336ddff962870092ac873a60dfa9b2/stdlib/builtins.pyi#L1171

https://github.com/python/typeshed/blob/b6564d976b336ddff962870092ac873a60dfa9b2/stdlib/builtins.pyi#L993-L1001

PR #9917, opened 2023-03-22
[minidom] add missing return types

FrancescElies:
Adds explicit return types for 4 functions


srittau (approved):
The new pytest primer hits are probably because `getElementsByTagName` should return `NodeList[Element]`, instead of `NodeList[Node]`. The new cwltool hit seems to point to a genuine problem.

PR #9918, opened 2023-03-22
Add stubs for translationstring

Daverball:
I've already had this mostly stubbed out privately, it's a very small package, so I thought I'd contribute the stubs.


stubs/translationstring/translationstring/__init__.pyi
```diff
@@ -0,0 +1,81 @@
...
+    ) -> str: ...
+
+class TranslationString(str):
+    domain: str | None
+    context: str | None
+    default: str
+    mapping: dict[str, Any] | None
+    def __new__(
+        self,
+        msgid: str | Self,
```
srittau, line 60: As `Self` is a subclass of `str`:

```suggestion
        msgid: str,
```

stubs/translationstring/translationstring/__init__.pyi
```diff
@@ -0,0 +1,81 @@
...
+        domain: str | None = ...,
+        default: str | None = ...,
+        mapping: dict[str, Any] | None = ...,
+        context: str | None = ...,
```
srittau, line 64: As we now allow and recommend simple defaults:

```suggestion
        domain: str | None = None,
        default: str | None = None,
        mapping: dict[str, Any] | None = None,
        context: str | None = None,
```

stubs/translationstring/translationstring/__init__.pyi
```diff
@@ -0,0 +1,81 @@
...
+    ) -> Self: ...
+    def __mod__(self, options: dict[str, Any]) -> Self: ...  # type:ignore[override]
```
srittau, line 66: Returning `Self` would mean that sub-classes would return the subclass. But in this case, a new `TranslationString` is explicitly created:

```suggestion
    def __mod__(self, options: dict[str, Any]) -> TranslationString: ...  # type:ignore[override]
```

stubs/translationstring/translationstring/__init__.pyi
```diff
@@ -0,0 +1,81 @@
...
+    def interpolate(self, translated: str | None = ...) -> str: ...
+    def __reduce__(self) -> tuple[type[Self], tuple[str, str | None, str, dict[str, Any], str | None]]: ...
+
+def TranslationStringFactory(factory_domain: str) -> _TranslationStringFactory: ...
+def ChameleonTranslate(translator: Callable[[TranslationString], str] | None) -> _ChameleonTranslate: ...
+def ugettext_policy(translations: NullTranslations, tstring: str, domain: str | None, context: str | None) -> str: ...
+def dugettext_policy(translations: NullTranslations, tstring: str, domain: str | None, context: str | None) -> str: ...
+def Translator(translations: NullTranslations | None = ..., policy: _TranslatorPolicy | None = ...) -> _Translator: ...
```
srittau, line 74:
```suggestion
def Translator(translations: NullTranslations | None = None, policy: _TranslatorPolicy | None = None) -> _Translator: ...
```

stubs/translationstring/translationstring/__init__.pyi
```diff
@@ -0,0 +1,81 @@
...
+        self,
+        msgid: str | TranslationString,
+        mapping: dict[str, Any] | None = ...,
+        context: str | None = ...,
+        target_language: str | None = ...,
+        default: str | None = ...,
+    ) -> TranslationString: ...
+
+class _TranslatorPolicy(Protocol):
+    def __call__(self, translations: NullTranslations, tstring: str, domain: str | None, context: str | None) -> str: ...
```
srittau, line 26: As these are called using a positional call (i.e. implementors can call the arguments whatever they want to):

```suggestion
    def __call__(self, __translations: NullTranslations, __tstring: str, __domain: str | None, __context: str | None) -> str: ...
```

srittau (changes requested):
Thanks, a few smaller suggestions below, mostly adding default values.

pyrightconfig.stricter.json
```diff
@@ -71,6 +71,7 @@
         "stubs/SQLAlchemy",
         "stubs/stripe",
         "stubs/tqdm",
+        "stubs/translationstring",
```
AlexWaygood, line 74: Doesn't look like we need to exclude these stubs from the stricter pyright settings, since all parameters and return types appear to already be annotated :)

```suggestion
```

stubs/translationstring/translationstring/__init__.pyi
Daverball: Absolutely correct, but since translationstring does special case instances of `TranslationString` in most function/method parameters that accept a string, I think it's a better to be explicit as a hint for the user, even if it is redundant as far as type checking is concerned.

stubs/translationstring/translationstring/__init__.pyi
Daverball: True, the reason I did not do this, is, because that's what the docstring for the `policy` argument calls for, so technically Pylons could enforce the name of the parameters in a future revision, without breaking the contract laid out by the docs.

Daverball:
@AlexWaygood We have use the package in a production 3.6, 3.7, 3.8, 3.9 and 3.10 environments without issue. I think the only reason it hasn't been updated is because there was not really any good reason to, since it still just works fine. Most of the complex functionality is provided by the gettext/babel NullTranslations/GNUTranslations object.

It's an optional component of the Pyramid Framework, but it's the recommended way to do I18n, rather than calling `gettext` etc.

srittau:
GitHub is overworked:

> The agent pool assigned to this job has hit their MacOs concurrency limits
> Requested labels: macos-latest

There shouldn't be any differences between MacOS and the other runners, though.

PR #9920, opened 2023-03-22
[win32api] add missing types

FrancescElies:


stubs/pywin32/win32/win32api.pyi
```diff
@@ -11,6 +11,20 @@ class _MonitorInfo(TypedDict):
     Flags: int
     Device: str
 
+class FileVersionInfo(TypedDict):
```
AlexWaygood, line 14:
```suggestion
class _FileVersionInfo(TypedDict):
```

stubs/pywin32/win32/win32api.pyi
```diff
@@ -91,7 +105,7 @@ def GetDomainName() -> str: ...
 def GetEnvironmentVariable(variable): ...
 def GetEnvironmentVariableW(Name) -> str: ...
 def GetFileAttributes(pathName: str): ...
-def GetFileVersionInfo(Filename: str, SubBlock: str) -> None: ...
+def GetFileVersionInfo(Filename: str, SubBlock: str) -> FileVersionInfo: ...
```
AlexWaygood, line 108:
```suggestion
def GetFileVersionInfo(Filename: str, SubBlock: str) -> _FileVersionInfo: ...
```

AlexWaygood (commented):
Making the TypedDict name private should fix the stubtest errors

stubs/pywin32/win32/win32api.pyi
```diff
@@ -11,6 +11,21 @@ class _MonitorInfo(TypedDict):
...
+    FileVersionMS: int
+    FileVersionLS: int
+    ProductVersionMS: int
+    ProductVersionLS: int
+    FileFlagsMask: int
+    FileFlags: int
+    FileOS: int
+    FileType: int
+    FileSubtype: int
+    FileDate: None | Any
```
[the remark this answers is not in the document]
AlexWaygood: We prefer to use `_typeshed.Incomplete` (which is just an alias for `Any`, but has a better name) for situations where we aren't sure about the precise type

```suggestion
    FileDate: None | Incomplete
```

AlexWaygood:
(We prefer to avoid force pushes in the typeshed repo, as they interact badly with the GitHub UI, making it hard to see what changed between commits. All PRs are squash-merged anyway, so there's no need to worry about keeping a clean git history in 99% of cases :-)

FrancescElies:
> (For future reference: we prefer to avoid force pushes in the typeshed repo, as they interact badly with the GitHub UI, making it hard to see what changed between commits. All PRs are squash-merged anyway, so there's no need to worry about keeping a clean git history in 99% of cases :-)

Alright, didn't know that, won't do that anymore :)

stubs/pywin32/win32/win32api.pyi
```diff
@@ -91,7 +105,7 @@ def GetDomainName() -> str: ...
 def GetEnvironmentVariable(variable): ...
 def GetEnvironmentVariableW(Name) -> str: ...
 def GetFileAttributes(pathName: str): ...
-def GetFileVersionInfo(Filename: str, SubBlock: str) -> None: ...
+def GetFileVersionInfo(Filename: str, SubBlock: str) -> _FileVersionInfo: ...
```
AlexWaygood, line 108: It looks like these parameters are actually positional-only, so let's fix that while we're here. We can't use PEP-570 syntax yet in typeshed as we still support Python 3.7, so this is the py37-compatible way of marking parameters as positional-only:

```suggestion
def GetFileVersionInfo(__Filename: str, __SubBlock: str) -> _FileVersionInfo: ...
```

stubs/pywin32/win32/win32api.pyi
```diff
@@ -249,10 +263,10 @@ def WinExec(cmdLine: str, arg) -> None: ...
...
 def WriteProfileSection(section: str, data: str, iniName: str | None = ...): ...
 def WriteProfileVal(section: str, entry: str, value: str, iniName: str | None = ...) -> None: ...
-def HIBYTE(val): ...
-def LOBYTE(val): ...
-def HIWORD(val): ...
-def LOWORD(val): ...
+def HIBYTE(val: int) -> int: ...
+def LOBYTE(val: int) -> int: ...
+def HIWORD(val: int) -> int: ...
+def LOWORD(val: int) -> int: ...
```
AlexWaygood, line 269:
```suggestion
def HIBYTE(__val: int) -> int: ...
def LOBYTE(__val: int) -> int: ...
def HIWORD(__val: int) -> int: ...
def LOWORD(__val: int) -> int: ...
```

AlexWaygood (approved):
I merged my suggested changes. Thanks!

PR #9921, opened 2023-03-22
openpyxl: add baseline stubs for new modules

AlexWaygood:
Some new modules were added in a recent release of `openpyxl`, which is contributing to CI failures in #9511 by @Avasam. This PR adds baseline stubs for the new modules.


srittau (approved):
Thanks!

PR #9924, opened 2023-03-22
Add types to urllib3.filepost stubs

adamtheturtle:
I copied the types from https://github.com/urllib3/urllib3/blob/main/src/urllib3/filepost.py, which is newer than the currently released `1.26.x`. I left `iter_fields` as that does not exist in `main` and doesn't have types in the `1.26.x` branch.


stubs/urllib3/urllib3/filepost.pyi
```diff
@@ -1,12 +1,15 @@
...
 
 from . import fields
 
 RequestField = fields.RequestField
 
 writer: Any
 
-def choose_boundary(): ...
-def iter_field_objects(fields): ...
+_TYPE_FIELDS_SEQUENCE = Sequence[tuple[str, fields._TYPE_FIELD_VALUE_TUPLE] | RequestField]
```
AlexWaygood, line 9: `fields._TYPE_FIELD_VALUE_TUPLE` exists on the `main` branch of `urllib3`, but doesn't exist in our stubs for `fields.pyi` yet (nor in any released version of `urllib3`). Feel free to add it as part of this PR though!

third_party/2and3/requests/packages/urllib3/fields.pyi
AlexWaygood, line 1: Hmm, we no longer have a `third_party/2and3` directory (we haven't for quite some time!). You probably want to edit the `stubs/urllib3/urllib3/fields.pyi` file :)

adamtheturtle: 🤦 Accidental addition!

adamtheturtle:
Thanks you @AlexWaygood for the comments! This now passes.

AlexWaygood (approved):
Thanks!

PR #9925, opened 2023-03-22
Add alternate type for cffi's from_buffer

rhelmot:
This matches the rest of the cdecl parameters and allows my working project to typecheck.


srittau (approved):
Thanks, I only have superficial experience with cffi, but this looks plausible to me.

PR #9929, opened 2023-03-23
Update UUID is_safe default value

aminalaee:
The `UUID` class `is_safe` parameter doesn't reflect the default value properly.
This is available in Python versions so I don't think this is a limitation, but a mismatch.

@AlexWaygood As it's related to https://github.com/python/typeshed/pull/9606 can you please check?


aminalaee:
So I guess this is on purpose, since doing `is_safe: SafeUUID = None` is also not possible.

PR #9936, opened 2023-03-24
Reflect `uuid.uuid{3,5}` changes in 3.12

sobolevn:
This PR https://github.com/python/cpython/pull/94709 allowed `bytes` to be passed as the second argument on 3.12+

CC @JelleZijlstra as the oriignal reviewer


srittau (approved):
Thanks!

PR #9937, opened 2023-03-24
pysftp: CnOpts.hostkeys should be optional

vad:
As stated by pysftp's documentation, hostkeys are optional:

> Host Key checking is enabled by default. It will use ~/.ssh/known_hosts by default. If you wish to disable host key checking (NOT ADVISED) you will need to modify the default CnOpts and set the .hostkeys to None.

```
import pysftp
cnopts = pysftp.CnOpts()
cnopts.hostkeys = None
with pysftp.Connection('host', username='me', password='pass', cnopts=cnopts):
    # do stuff here
```

See https://pysftp.readthedocs.io/en/release_0.2.9/cookbook.html#pysftp-cnopts


srittau (approved):
Thanks!

PR #9941, opened 2023-03-25
stdlib: Fix some shlex types

kpschoedel:
Change type of `shlex.eof` from `str` to `str | None`.
Reason: https://github.com/python/cpython/blob/3.11/Lib/shlex.py#L33

Change return type of `shlex.get_token()` from `str` to `str | None`.
Reason: https://github.com/python/cpython/blob/3.11/Lib/shlex.py#L121

Change return type of `shlex.read_token()` from `str` to `str | None`.
Reason: https://github.com/python/cpython/blob/3.11/Lib/shlex.py#L271

Change return type of `shlex.error_leader()` from `None` to `str`.
Reason: https://github.com/python/cpython/blob/3.11/Lib/shlex.py#L294

Change return type of `shlex.sourcehook()` from `tuple[str, TextIO]` to `tuple[str, TextIO] | None`
Reason: https://github.com/python/cpython/blob/3.11/Lib/shlex.py#L113-L114


stdlib/shlex.pyi
```diff
@@ -39,12 +39,12 @@ class shlex(Iterable[str]):
         posix: bool = False,
         punctuation_chars: bool | str = False,
     ) -> None: ...
-    def get_token(self) -> str: ...
+    def get_token(self) -> str | None: ...
     def push_token(self, tok: str) -> None: ...
-    def read_token(self) -> str: ...
+    def read_token(self) -> str | None: ...
     def sourcehook(self, newfile: str) -> tuple[str, TextIO]: ...
```
Akuli, line 45: I think the `sourcehook()` method should be updated as well. It is used here:

```
                spec = self.sourcehook(self.read_token())
                if spec:
                    (newfile, newstream) = spec
                    self.push_source(newstream, newfile)
```

kpschoedel: I see; I had only looked look at definitions and not uses. Evidently a subclass can override it to return the tuple _or_ anything falsy. I don't know typing well enough to know how to annotate that correctly, since there's no SupportsBool.

Akuli: `| None` is good enough for that. Presumably the `newfile` argument can also be `None`, because according to the new annotations `read_token()` can return `None`.

Akuli: Actually, I think `newfile` is never `None` for some reason. The definition of `sourcehook()` in `class shlex` accesses `newfile[0]` without checking for `None`. So, we only need:

```suggestion
    def sourcehook(self, newfile: str) -> tuple[str, TextIO] | None: ...
```

Akuli (approved):
Thanks!

PR #9946, opened 2023-03-26
Ran stubdefaulter for `stubs/openpyxl`

Avasam:
This will reduce changes in #9511 as well as [an upcoming PR for descriptors](https://github.com/python/typeshed/pull/9511#issuecomment-1483750794)

Only automated changes + 1 fix, to allow for easy and quick review. Correct argument types for descriptors will be filled in with desrciptors support.


AlexWaygood (approved):
Thank you!

PR #9947, opened 2023-03-26
Add type-hints for dataclass params

antonagestam:


AlexWaygood:
Looks like mypy maybe doesn't currently synthesise the `__dataclass_params__` attribute for dataclass classes, only the `__dataclass_fields__` attribute, which would make this unmergeable for now.

antonagestam:
@AlexWaygood Should this be raised as an issue with mypy?

AlexWaygood:
> @AlexWaygood Should this be raised as an issue with mypy?

The attribute is an internal implementation detail of the `dataclasses` module, so I'm not sure if mypy _should_ really be synthesising it or not. On the other hand, it already synthesises the `__dataclass_fields__` attribute (as do all other major type checkers), and that attribute is also an internal implementation detail of `dataclasses`. So idk.

What do you need this attribute for?

AlexWaygood:
As you can see from the pyright failures, it looks like pyright also does not currently synthesise the `__dataclass_params__` attribute

AlexWaygood:
If you need the `_DataclassParams` class definition for some reason, we could probably add that to `dataclasses.pyi` without adding the `__dataclass_params__` attribute to the protocol in `_typeshed`

antonagestam:
I was guided by [this section](https://github.com/python/typeshed/blob/main/CONTRIBUTING.md#what-to-include) in the contributing guidelines:

> Other objects may be included if they are being used in practice

I'm using this to determine whether a dataclass is frozen or not. It doesn't seem reasonable to only have type-hints for one of these attributes and not the other, as both seem equally interesting for runtime introspection.

> On the other hand, it already synthesises the __dataclass_fields__ attribute [...], and that attribute is also an internal implementation detail of dataclasses.

This was my reasoning too.

How do these dependencies work, between type-checkers and typeshed? My understanding was that type-checkers depend on typeshed, and not the other way around. Is this is a special case because type-checkers have special handling for dataclasses?

JelleZijlstra:
Yes, the special-casing for dataclasses means that we cannot merge this until major type checkers synthesize this attribute. (And I'm not sure that they should; you'll have to make a case that this attribute is useful commonly enough.)

AlexWaygood:
> I'm using this to determine whether a dataclass is frozen or not. It doesn't seem reasonable to only have type-hints for one of these attributes and not the other, as both seem equally interesting for runtime introspection.

If you need this for introspection, you could consider filing a feature request over at CPython for a `dataclasses.params()` function to be added at runtime. That would mirror the approach the runtime has taken with `dataclass.fields()` (keeping the attribute itself private and undocumented, but adding a public-API introspection function for getting the contents of the attribute).

It might be easier to get the public-API introspection function added at runtime rather than to go round all the major type checkers and persuade them to synthesise another private and undocumented `dataclasses` attribute so that we can add stubs for it here.

antonagestam:
@AlexWaygood @JelleZijlstra Thanks for your thoughtful responses, I'll close this issue for now. I might pursue the suggested runtime API instead.

PR #9948, opened 2023-03-27
Reflect the latest reprlib updates

asvetlov:
3.11 adds `fillvalue` attr.
3.12 adds `indent`, plus constructor accepts all settings as named arguments


stdlib/reprlib.pyi
```diff
@@ -1,12 +1,14 @@
...
 from collections.abc import Callable
-from typing import Any
+from typing import Any, TypeVar
 from typing_extensions import TypeAlias
 
 __all__ = ["Repr", "repr", "recursive_repr"]
 
-_ReprFunc: TypeAlias = Callable[[Any], str]
+_T = TypeVar("_T")
+_ReprFunc: TypeAlias = Callable[[_T], str]
```
AlexWaygood, line 11: What's the motivation for this change? It's the only thing making the tests fail :)

asvetlov: Ooop.
I thought that `recursive_repr` decorator should not change the type of accepted argument.
Sure, this change is not related to others. I'll revert it.

asvetlov: Done

asvetlov:
> We recently changed our policy to prefer including default values in many cases (they're useful to users of IDEs), so I've pushed to your branch to replace the ...s with the defaults as they exist at runtime

Thanks for letting me know.
Honestly, in my own open-source project I also prefer actual defaults over `...` for exactly the same reason.

PR #9950, opened 2023-03-27
Add stubs/recipe_scrapers

adamschmalhofer:
For most basic and common use of recipe_scrapers


AlexWaygood:
This package appears to ship with a `py.typed` file, so it shouldn't be necessary to add stubs to typeshed. Type checkers should just see the inline types from the runtime package.

PR #9951, opened 2023-03-27
Add stubs/argcomplete

adamschmalhofer:
For most basic and common use of argcomplete


AlexWaygood:
This package also ships with a `py.typed` file, so it shouldn't be necessary to add separate stubs to typeshed if you're using the latest version of the runtime package

PR #9952, opened 2023-03-27
Add defaults for third-party stubs A-D

AlexWaygood:
I ran stubdefaulter on our third-party stubs beginning with A-D. (I skipped D3DShot, as it's hard to install locally, and we're about to get rid of those stubs anyway.)


srittau (approved):

AlexWaygood:
3e69de1 shows some truly horrifying default values that I'm very glad flake8-pyi caught before they were committed into the stubs ☺️

stubs/babel/babel/messages/plurals.pyi
```diff
@@ -10,4 +10,4 @@ class _PluralTuple(tuple[int, str]):
     @property
     def plural_forms(self) -> str: ...
 
-def get_plural(locale: str = ...) -> _PluralTuple: ...
+def get_plural(locale: str = None) -> _PluralTuple: ...
```
AlexWaygood, line 13:
```suggestion
def get_plural(locale: str | None = None) -> _PluralTuple: ...
```

stubs/babel/babel/dates.pyi
```diff
@@ -143,15 +143,15 @@ time_ = time
...
+    time: _Instant, tzinfo: BaseTzInfo | None = None, type: Literal["selection"] | None = None, locale: Locale | str | None = ...
 ): ...
 
 class DateTimePattern:
     pattern: str
     format: DateTimeFormat
     def __init__(self, pattern: str, format: DateTimeFormat) -> None: ...
     def __mod__(self, other: DateTimeFormat) -> str: ...
-    def apply(self, datetime: _Instant, locale: Locale | str | None) -> str: ...
+    def apply(self, datetime: _Instant, locale: Locale | str | ...) -> str: ...
```
srittau, line 154: That doesn't look right.

AlexWaygood:
```suggestion
    def apply(self, datetime: _Instant, locale: Locale | str | None) -> str: ...
```

AlexWaygood:
Squashed the chaotic last few commits together, for easier reviewing.

srittau (approved):
I assume the babel changes change `None` back to `...` because of the platform-specific defaults? LGTM in that case.

AlexWaygood:
> I assume the babel changes change `None` back to `...` because of the platform-specific defaults? LGTM in that case.

Yup, the defaults are all `None` for me locally but are some specific locale in CI, causing stubtest to (correctly) complain about a mismatch.

Avasam:
Not even gonna bother with `D3DShot`. Stubs pretty much only exist for legacy support reasons.
If I don't have a good working alternative by the time the stub is removed from typeshed, I'll just fork the archived library and type it inline anyway.

PR #9953, opened 2023-03-27
contextvars: widen return value of ContextVar.get(default)

jhominal:
While working on a Python 3.6 project (I know, not supported anymore, but...) I encountered an issue with the type declaration of `ContextVar.get`, where the third-party stub does not match the standard library stub.

In order to fix that, I thought it would be better to split the `ContextVar.get` into two overloads, as is done in the corresponding stdlib stub.


jhominal:
@srittau Thank you for the very fast turnaround time and review!
I would just like to know, as I do not know at what frequency the `types-...` packages are released, when should I expect that change to be available in https://pypi.org/project/types-contextvars/?

srittau:
@jhominal *In theory*, every three hours, but GitHub Actions is sometimes a bit inconsistent. You can see the latest runs here: https://github.com/typeshed-internal/stub_uploader/actions/workflows/update_stubs.yml

JelleZijlstra:
Since contextvars has been in the stdlib since 3.7, we may want to delete the backport stubs.

jhominal:
@srittau Thank you, that is awesome to hear!

@JelleZijlstra I absolutely understand, I would take it as a personal favor if you would just let this one release go through 😅
(You may also want to remove `dataclasses` as it is also a Python<3.7 backport)

jhominal:
Right, I realize right now that it would also be possible for me to maintain my own stub file in my project, I just have not looked into it.

Again, thank you!

PR #9954, opened 2023-03-27
Add defaults for third-party stubs E-H

AlexWaygood:
I omitted `gdb` (for obvious reasons), and `hdbcli` (because I couldn't immediately install it locally and didn't want to spend too much time debugging it for now).


srittau (approved):
Blanket approval again if the tests pass.

Avasam:
I can take care of `hdbcli`, it's already installed on my machine. There's also only 2 defaults

PR #9955, opened 2023-03-27
Add defaults for third-party stubs I-L

AlexWaygood:
I skipped JACK-Client for now, as it's complicated to install.


AlexWaygood:
We appear to have some invalid annotations in `keyboard`, but it's not a library I'm especially familiar with, so I'll just revert the additions of the problematic defaults for now and I'll open a followup issue about the invalid annotations.

srittau (approved):

PR #9957, opened 2023-03-27
Add default values for third-party stubs beginning with 'P'

AlexWaygood:
We have a lot of stubs beginning with 'P'!

I skipped `pycurl` (couldn't install it locally) and `python-xlib` (@Avasam did it just the other day in #9642).


AlexWaygood:
Hmm, looks like there might be a pyright bug here. For now, I'll just revert the defaults that pyright is complaining about, and I'll file an issue over at pyright later.

Avasam:
I can check pycurl if you want. I still have my mid-2009 macbook pro I patched and upgraded for this exact reason.
The installation is a bit complex, but documented in METADATA.toml

For linux, it's a lot simpler.

Windows you need to build from source. But you wan use WSL. If you use VSCode there's https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.remote-wsl you can use to even straight up have all your tools in VSCode think you're on Linux (including os-specific intellisense and whatnot)

AlexWaygood:
> I can check pycurl if you want. I still have my mid-2009 macbook pro I patched and upgraded for this exact reason.
> The installation is a bit complex, but documented in METADATA.toml

Ah, that would be great if you could, thank you!

srittau (approved):

AlexWaygood:
> > python-xlib (@Avasam did it just the other day in #9642).
> 
> I think that was before more defaults were allowed. stubdefaulter adds a couple more tuple and dict defaults now

'Kay, I'll do it in a followup PR (or you can if you like!)

AlexWaygood:
> Hmm, looks like there might be a pyright bug here. For now, I'll just revert the defaults that pyright is complaining about, and I'll file an issue over at pyright later.

- https://github.com/microsoft/pyright/issues/4856

PR #9958, opened 2023-03-27
Bump pyright test dependency to 1.1.300

AlexWaygood:


pyproject.toml
```diff
@@ -62,4 +62,4 @@ all = true
 disable_all_dunder_policy = true
 
 [tool.typeshed]
-pyright_version = "1.1.295"
+pyright_version = "1.1.300"
```
AlexWaygood, line 65:
```suggestion
pyright_version = "1.1.299"
```

srittau:
It looks to me as if the 1.1.300 errors are related to annotations like:

```python
    foo: int = ...
```

I wouldn't mind if we'd "fix" them. (Actually I'm a bit surprised we don't have a flake8-pyi check.)

srittau:
PyCQA/flake8-pyi#98

PR #9959, opened 2023-03-27
Add defaults for third-party stubs Q-T

AlexWaygood:
I skipped `six`, because stubdefaulter crashed when parsing the stubs for `six` (the issue is with stubdefaulter's `typeshed_client` dependency, which is used for parsing the stubs, not stubdefaulter itself). I also skipped `termcolor`, because in order to add stubs for `termcolor` stubdefaulter needs to install `termcolor<2` (as that's what we have in the `METADATA.toml` file for `termcolor`), but stubdefaulter itself depends on `termcolor>=2`.


JelleZijlstra:
If you open an issue on typeshed-client with the crash report I'll fix it.

Avasam:
`termcolor==1.1.0`'s defaults are all `None` anyway. incoming PR

AlexWaygood:
> If you open an issue on typeshed-client with the crash report I'll fix it.

- https://github.com/JelleZijlstra/typeshed_client/issues/78

AlexWaygood:
I tried commenting out the problematic line and running stubdefaulter on `six`, but it turns out stubdefaulter actually makes 0 changes to our stubs for `six` `¯\_(ツ)_/¯`

PR #9960, opened 2023-03-27
Rerun updated stubdefaulter on python-xlib

Avasam:
Follow up to https://github.com/python/typeshed/pull/9957#issuecomment-1485571835

Omitted `Xlib/rdb.pyi`'s `get_display_opts`. `argv` parameter defaults to the current cli arguments

CC @AlexWaygood


AlexWaygood (approved):
Thanks! :D

PR #9965, opened 2023-03-27
`keyboard` : Update `stubtest_allowlist` and add comments for missing defaults

Avasam:
Follow up to https://github.com/python/typeshed/pull/9955#issuecomment-1485719217 CC @AlexWaygood 
Only Windows has the "double" mouse event type (ie: double clicks, on Darwin and Linux it's simply two fast left clicks)
The default still includes it because it's default listeners, so it'll include all mouse click events independent of OS.

Unless it messes with type-checker's interpretations of possible types, this should be accurate. If it is an issue, let's keep `...` instead as the type is accurate.
We can also err on the side of caution and not include the default. This module is considered obsolete anyway.


Avasam:
pyrigth can't be happy on all OSes (it'll raise "unused ignore"). I'm not gonna add os conditions in the stub that only differs by a  `# pyright: ignore` comment because of defaults in an obsolete module.

srittau (approved):

PR #9968, opened 2023-03-28
[cffi] missing types

FrancescElies:
getitem and setitem index is an int.


hauntsaninja (commented):
Thanks, I think technically `index` can be a slice as well: https://foss.heptapod.net/pypy/cffi/-/blob/branch/default/c/_cffi_backend.c#L2697

JelleZijlstra:
It can also be SupportsIndex, as the code ultimately calls `PyNumber_AsSsize_t`. I updated the PR to reflect that.

hauntsaninja (approved):
Thanks both!

PR #9969, opened 2023-03-28
Add retry parameters to redis.client.from_url

biggerfisch:
These parameters work and are passed down, but were not present in the type for from_url


srittau (approved):
Thanks!

PR #9970, opened 2023-03-28
Delete some unnecessary type aliases in tkinter

Akuli:
Having less type aliases should make IDE suggestions and type checker error messages etc more readable. I didn't consider it when I rewrote the tkinter stubs :)


AlexWaygood (approved):
Haven't done a line-by-line review, but I approve of the principle, and trust your ability to search-and-replace :)

PR #9971, opened 2023-03-28
Add defaults to third-party stubs U-Z

AlexWaygood:
This PR was a bit of a ride: the code stubdefaulter came up with initially for `zxcvbn` caused [black](https://github.com/psf/black/issues/3627), flake8, [mypy](https://github.com/python/mypy/issues/14970) and [IDLE](https://github.com/python/cpython/issues/103089) all to hang.


AlexWaygood:
Nice, I see GitHub refuses to even try to show the diff for 7c9921c0702ced05992eeaa59801a85911707c86 because it's so large

PR #9972, opened 2023-03-28
Bump pyright to 1.1.300

AlexWaygood:
There's a lot of new errors; opening a draft PR so that we can see exactly what they are and analyze/fix them where possible. Some may be bugs in pyright rather than typeshed.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -62,4 +62,4 @@ all = true
 disable_all_dunder_policy = true
 
 [tool.typeshed]
-pyright_version = "1.1.299"
+pyright_version = "1.1.300"


stubs/colorama/colorama/winterm.pyi
```diff
@@ -1,22 +1,23 @@
...
+        YELLOW: ClassVar[int]
+        GREY: ClassVar[int]
 
     class WinStyle:
-        NORMAL: int = ...
-        BRIGHT: int = ...
-        BRIGHT_BACKGROUND: int = ...
+        NORMAL: ClassVar[int]
+        BRIGHT: ClassVar[int]
+        BRIGHT_BACKGROUND: ClassVar[int]
```
AlexWaygood, line 20: These all look like they could _probably_ be `Final`, but that could cause false-positive errors if I'm incorrect about that, and I'm not confident enough to risk it

srittau (approved):

stubs/colorama/colorama/winterm.pyi
JelleZijlstra: Seems pretty clear these should be final. They're primitive enums.

PR #9973, opened 2023-03-28
Add default values to `pika.spec`

AlexWaygood:
See the first commit for the workaround I used for the [pyright issue](https://github.com/microsoft/pyright/issues/4856) I encountered in #9957. (The pyright issue has already been closed as expected behaviour by Eric).

### DIFF
--- a/stubs/pika/pika/spec.pyi
+++ b/stubs/pika/pika/spec.pyi
@@ -57,8 +57,8 @@ class Connection(Class):
             version_major: int = 0,
             version_minor: int = 9,
             server_properties: Incomplete | None = None,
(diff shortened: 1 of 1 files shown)


srittau:
I actually prefer the workaround as it's more explicit.

srittau (approved):

AlexWaygood:
> I actually prefer the workaround as it's more explicit.

Yes, same, and now that Eric's explained pyright's behaviour here, I think there's an argument that it's actually more correct than mypy's behaviour here. It's slightly unfortunate that none of our tests warned us before now that mypy and pyright interpreted the code in this file completely differently, but there we go.

PR #9976, opened 2023-03-29
Remove unnecessary ellipsis expressions

srittau:


srittau:
Fixing the namedtuples now

srittau:
I wish we had a decent, eslint-like linter for Python. :( We should probably just ignore error F821 as that should be caught by our type checker tests anyway.

AlexWaygood:
We could consider switching off F821 in our CI for `.pyi` files, though. I think it probably doesn't provide that much value for us, given how much linting we already get from our mypy and pyright CI checks.

stdlib/typing.pyi
```diff
@@ -173,14 +173,14 @@ def overload(func: _F) -> _F: ...
 # Otherwise, flake8 erroneously flags them as undefined.
 # `_SpecialForm` objects in typing.py that are not used elswhere in the same file
 # do not need the default value assignment.
```
AlexWaygood, line 175: Need to delete the multiline comment immediately above this line explaining the `= ...`s if we're getting rid of the `= ...`s

stdlib/random.pyi
```diff
@@ -102,7 +102,7 @@ class SystemRandom(Random):
 
 # ----- random function stubs -----
```
AlexWaygood, line 104: It's not relevant to this PR, but this fatuous comment really grinds my gears; I'd love it if we can get rid of it while we're here 😆

```suggestion
```

stubs/waitress/waitress/buffers.pyi
```diff
@@ -6,12 +6,12 @@ COPY_BYTES: int
...
 class FileBasedBuffer:
-    remain: int = ...
-    file: BytesIO = ...
+    remain: int
+    file: BytesIO
     def __init__(self, file: BytesIO, from_buffer: BytesIO | None = None) -> None: ...
     def __len__(self) -> int: ...
     def __nonzero__(self) -> bool: ...
-    __bool__: Callable[[], bool] = ...
+    __bool__: Callable[[], bool]
```
AlexWaygood, line 14: https://github.com/Pylons/waitress/blob/afc7b9de54f21b63bb1e8fd6b12c704876d9868a/src/waitress/buffers.py#L47-L48

```suggestion
    def __bool__(self) -> Literal[True]: ...
```

stubs/colorama/colorama/ansitowin32.pyi
```diff
@@ -29,15 +29,15 @@ _WinTermCall: TypeAlias = Callable[[int | None, bool, bool], None]
...
-    ANSI_OSC_RE: Pattern[str] = ...
-    wrapped: TextIO = ...
-    autoreset: bool = ...
-    stream: StreamWrapper = ...
-    strip: bool = ...
-    convert: bool = ...
-    win32_calls: _WinTermCallDict = ...
-    on_stderr: bool = ...
+    ANSI_CSI_RE: Pattern[str]
+    ANSI_OSC_RE: Pattern[str]
```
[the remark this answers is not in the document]
AlexWaygood: They're never modified on instances from what I can see, skimming the source code, so I think so:

```suggestion
    ANSI_CSI_RE: ClassVar[Pattern[str]]
    ANSI_OSC_RE: ClassVar[Pattern[str]]
```

AlexWaygood (approved):
Looks good other than the missing `ClassVar` and `Literal` imports that the CI is complaining about

AlexWaygood:
I just remembered that we added an unnecessary `= ...` the other day in #9890, to workaround the pyflakes issue:

https://github.com/python/typeshed/blob/8df767f163e465e95447d4f95e25968d70856ebb/stubs/setuptools/setuptools/build_meta.pyi#L37

It looks like it's not flagged in the CI for #9972 because pyright is only flagging the `= ...`s for stubs where we have the stricter pyright settings enabled. So there will be more unnecessary `= ...`s for the third-party stubs where we have not yet enabled the stricter pyright settings.

But it's fine to leave those for now, we can get to those later.

PR #9978, opened 2023-03-29
[fpdf2] Update to 2.7.1

srittau:
Closes: #9967


stubs/fpdf2/fpdf/fpdf.pyi
```diff
@@ -451,17 +466,15 @@ class FPDF:
         title: str | None = None,
         alt_text: str | None = None,
         dims: tuple[float, float] | None = None,
+        keep_aspect_ratio: bool = False,
     ) -> _Image: ...
+    def preload_image(self, name: str | Image.Image | BytesIO, dims: Incomplete | None = None) -> tuple[str, Any, ImageInfo]: ...
```
AlexWaygood, line 471: The docstring says:

>dims (Tuple[float]): optional dimensions as a tuple (width, height) to resize the image
>    before storing it in the PDF.

https://github.com/PyFPDF/fpdf2/blob/615db585a53dde9af5874fb0a58f33ee7f84aa2b/fpdf/fpdf.py#L3821-L3822

```suggestion
    def preload_image(self, name: str | Image.Image | BytesIO, dims: tuple[float, float] | None = None) -> tuple[str, Any, ImageInfo]: ...
```

stubs/fpdf2/fpdf/fpdf.pyi
```diff
@@ -506,3 +519,32 @@ class FPDF:
...
+        *,
+        align: str | Align = "CENTER",
+        borders_layout: str | TableBordersLayout = ...,
+        cell_fill_color: int | tuple[Incomplete, ...] | DeviceGray | DeviceRGB | None = None,
+        cell_fill_mode: str | TableCellFillMode = ...,
+        col_widths: int | tuple[int, ...] | None = None,
+        first_row_as_headings: bool = True,
+        headings_style: FontFace = ...,
+        line_height: int | None = None,
+        markdown=False,
```
AlexWaygood, line 535:
```suggestion
        markdown: bool = False,
```

stubs/fpdf2/fpdf/fpdf.pyi
```diff
@@ -506,3 +519,32 @@ class FPDF:
...
+        text_align: str | Align = "JUSTIFY",
+        width: int | None = None,
+    ) -> _GeneratorContextManager[Table]: ...
+    @overload
+    def output(  # type: ignore[misc]
+        self,
+        name: Literal[""] | None = "",
+        dest: Unused = "",
+        linearize: bool = False,
+        output_producer_class: Callable[[FPDF], Incomplete] = ...,
```
AlexWaygood, line 545: `.bufferize()` is called on the object returned by this callable:

https://github.com/PyFPDF/fpdf2/blob/60f5104e9a125d2ac0288a23960d0b1960e54329/fpdf/fpdf.py#L4755-L4756

I think it's meant to return an instance of `OutputProducer` (or an instance of a subclass of `OutputProducer`:

https://github.com/python/typeshed/blob/972958a406358cca95ccf0e9230f88878557abc5/stubs/fpdf2/fpdf/output.pyi#L183-L192

```suggestion
        output_producer_class: Callable[[FPDF], OutputProducer] = ...,
```

stubs/fpdf2/fpdf/fpdf.pyi
```diff
@@ -506,3 +519,32 @@ class FPDF:
...
+    ) -> None: ...
+    @overload
+    def output(
+        self, name: str, dest: Unused = "", linearize: bool = False, output_producer_class: Callable[[FPDF], Incomplete] = ...
```
AlexWaygood, line 549: same as above

```suggestion
        self, name: str, dest: Unused = "", linearize: bool = False, output_producer_class: Callable[[FPDF], OutputProducer] = ...
```

stubs/fpdf2/fpdf/html.pyi
```diff
@@ -58,6 +59,11 @@ class HTML2FPDF(HTMLParser):
     heading_above: float
     heading_below: float
     warn_on_tags_not_matching: bool
+
+    # Not initialized in __init__:
+    font_face: Incomplete
+    h: Incomplete
```
AlexWaygood, line 65: It's whatever is returned by `px2mm`: https://github.com/PyFPDF/fpdf2/blob/60f5104e9a125d2ac0288a23960d0b1960e54329/fpdf/html.py#L574

And `px2mm` returns a float: https://github.com/PyFPDF/fpdf2/blob/60f5104e9a125d2ac0288a23960d0b1960e54329/fpdf/html.py#L177-L178

```suggestion
    h: float
```

stubs/fpdf2/fpdf/image_parsing.pyi
```diff
@@ -1,15 +1,33 @@
...
 SUPPORTED_IMAGE_FILTERS: tuple[_ImageFilter, ...]
+TIFFBitRevTable: list[int]
 
 def load_image(filename): ...
+def is_iccp_valid(iccp, filename) -> bool: ...
 
 # Returned dict could be typed as a TypedDict.
-def get_img_info(img, image_filter: _ImageFilter = "AUTO", dims: Incomplete | None = None) -> dict[str, Any]: ...
+def get_img_info(
+    filename, img: Incomplete | None = None, image_filter: _ImageFilter = "AUTO", dims: Incomplete | None = None
```
AlexWaygood, line 20: Docstring: 

> img: optional `BytesIO` or `PIL.Image.Image` instance

https://github.com/PyFPDF/fpdf2/blob/60f5104e9a125d2ac0288a23960d0b1960e54329/fpdf/image_parsing.py#L111

```suggestion
    filename, img: BytesIO | Image.Image | None = None, image_filter: _ImageFilter = "AUTO", dims: Incomplete | None = None
```

stubs/fpdf2/fpdf/image_parsing.pyi
```diff
@@ -1,15 +1,33 @@
...
+) -> dict[str, Any]: ...
+
+class temp_attr:
+    obj: Any
+    field: str
+    value: Any
+    exists: bool  # defined after __enter__ is called
+    def __init__(self, obj: Any, field: str, value: Any) -> None: ...
+    def __enter__(self) -> None: ...
+    def __exit__(self, exctype: Unused, excinst: Unused, exctb: Unused) -> None: ...
```
AlexWaygood, line 30: I thought we agreed that we should use the "principled" annotations for these if the three arguments were listed out explicitly, even if one or more was unused!

```suggestion
    def __exit__(self, exctype: type[BaseException] | None, excinst: BaseException | None, exctb: TracebackType | None) -> None: ...
```

stubs/fpdf2/fpdf/table.pyi
```diff
@@ -0,0 +1,67 @@
...
+
+DEFAULT_HEADINGS_STYLE: FontFace
+
+class Table:
+    rows: list[Row]
+
+    def __init__(
+        self,
+        fpdf: FPDF,
+        rows: Iterable[Incomplete] = (),
```
AlexWaygood, line 23: Docstring: https://github.com/PyFPDF/fpdf2/blob/615db585a53dde9af5874fb0a58f33ee7f84aa2b/fpdf/table.py#L38

```suggestion
        rows: Iterable[Row | str] = (),
```

stubs/fpdf2/fpdf/table.pyi
```diff
@@ -0,0 +1,67 @@
...
+class Cell:
+    text: str
+    align: str | Align | None
+    style: FontFace | None
+    img: str | None
+    img_fill_width: bool
+    colspan: int
+
+    @abstractmethod
+    def write(self, text, align: Incomplete | None = None): ...
```
AlexWaygood, line 67: I'm not sure this is actually meant to be overridden on subclasses: it looks like the maintainers are raising `NotImplementedError` here because, well, they haven't implemented it yet: https://github.com/PyFPDF/fpdf2/blob/615db585a53dde9af5874fb0a58f33ee7f84aa2b/fpdf/table.py#L355

It's an unusual use of `NotImplementedError`, but it is a supported use of `NotImplementedError` according to the CPython docs! https://docs.python.org/3/library/exceptions.html#NotImplementedError

> In user defined base classes, abstract methods should raise this exception when they require derived classes to override the method, or while the class is being developed to indicate that the real implementation still needs to be added.

```suggestion
    def write(self, text, align: Incomplete | None = None): ...
```

AlexWaygood (commented):
Those fpdf2 maintainers never get tired of refactoring their source code 🙃

stubs/fpdf2/fpdf/fonts.pyi
```diff
@@ -1,4 +1,28 @@
...
+from .drawing import DeviceGray, DeviceRGB
+from .enums import TextEmphasis
+
+@dataclass
+class FontFace:
+    family: str | None
+    emphasis: TextEmphasis | None
+    size_pt: int | None
+    color: int | tuple[Incomplete, ...] | DeviceGray | DeviceRGB | None
+    fill_color: int | tuple[Incomplete, ...] | DeviceGray | DeviceRGB | None
```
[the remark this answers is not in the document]
AlexWaygood: Having said that, maybe `tuple[float, float, float]` would be better. Not sure.

JelleZijlstra:
@srittau could you address Alex's review?

stubs/fpdf2/fpdf/__init__.pyi
```diff
@@ -1,7 +1,7 @@
 from pathlib import Path
 
-from .enums import Align as Align, XPos as XPos, YPos as YPos
-from .fpdf import FPDF as FPDF, TitleStyle as TitleStyle
+from .enums import Align as Align, TextMode as TextMode, XPos as XPos, YPos as YPos
```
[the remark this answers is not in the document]
srittau: Not adding it to `__all__` looks more like an oversight in fpdf2. It's imported into `__init__.py`, but not used, so I assume that it's meant to be imported directly from `fpdf` (like the other enums).

stubs/fpdf2/fpdf/deprecation.pyi
```diff
@@ -1,5 +1,7 @@
 from types import ModuleType
+from typing import Any, NoReturn
 
 class WarnOnDeprecatedModuleAttributes(ModuleType):
-    def __getattr__(self, name: str): ...
-    def __setattr__(self, name: str, value) -> None: ...
+    def __call__(self) -> NoReturn: ...
+    def __getattr__(self, name: str) -> Any: ...
+    def __setattr__(self, name: str, value: Any) -> None: ...
```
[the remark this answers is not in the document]
srittau: Well, it's annotated now, I don't see the harm in keeping it.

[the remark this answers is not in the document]
AlexWaygood: I'd argue that the harm is that it's a maintenance burden for us to continually have to update the stubs every time the upstream maintainers make minor additions or deletions to a module that they probably consider to be an implementation detail?

stubs/fpdf2/fpdf/fonts.pyi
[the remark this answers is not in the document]
srittau: Considering that `Device*` is defined using `fpdf.drawing.Number` (int, float, Decimal), I think it's fairly likely that this accepted for these tuples as well. Probably the safest bet.

stubs/fpdf2/fpdf/table.pyi
AlexWaygood: Good catch, you're right!

AlexWaygood (approved):
Thanks!

PR #9981, opened 2023-03-29
Fix type of connection_class in ConnectionPool

### LINKED ISSUE #9979: [redis] Incorrect type in ConnectionPool
If I'm not mistaken, the type for `connection_class` should be `AbstractConnection` instead of `Connection` https://github.com/python/typeshed/blob/ed915c5cde48329e3d1e597e60c4baee27b7854a/stubs/redis/redis/connection.pyi#L257

It is causing an error for with `UnixDomainSocketConnection` (which is an `AbstractConnection` but not a `Connection`).

Rafiot:
Fix #9979

Due to  https://github.com/redis/redis-py/commit/7d474f90453c7b90bd06c94e0250b618120a599d


srittau (approved):
Thanks!

PR #9982, opened 2023-03-29
[redis] Add CredentialProvider types to __init__.pyi

### LINKED ISSUE #9980: [redis] CredentialProvider types missing from __init__.pyi
In Redis' `__init__.py`, `CredentialProvider` and `UsernamePasswordCredentialProvider` are imported and re-exported (see [here](https://github.com/redis/redis-py/blob/master/redis/__init__.py#L69-L71)).

However, in the type stubs these types are missing from the `__init__.pyi` (see [here](https://github.com/python/typeshed/blob/main/stubs/redis/redis/__init__.pyi#L12-L13)).

The types are defined in `credentials.py`, so if you use `redis.credentials.CredentialProvider` in your code it works fine, but if you try to use `redis.CredentialProvider` directly, like in the [docs](https://redis.readthedocs.io/en/stable/examples/connection_examples.html#), you get the error `error: Name "redis.CredentialProvider" is not defined`

nint8835:
Resolves #9980

Adds the missing classes `CredentialProvider` and `UsernamePasswordCredentialProvider` to `__init__.py`, so using `redis.CredentialProvider` like written in the redis.py docs does not result in a type error.

### DIFF
--- a/stubs/redis/redis/__init__.pyi
+++ b/stubs/redis/redis/__init__.pyi
@@ -1,4 +1,4 @@
-from . import client, connection, exceptions, sentinel, utils
+from . import client, connection, credentials, exceptions, sentinel, utils
 from .cluster import RedisCluster as RedisCluster
 
 __all__ = [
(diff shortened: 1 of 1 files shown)


srittau (approved):
Thanks!

PR #9987, opened 2023-03-30
Remove useless overloads from `sys.getsizeof`

sobolevn:
There's no need to use `@overload`, because signature is simple enough.

Btw, I am preparing a PR to CPython to convert this function to AC, so `__text_signature__` will be available :)


srittau (approved):
LGTM, this doesn't really have a default value, so `...` is fine here.

PR #9986, opened 2023-03-30
Remove `termcolor`

sobolevn:
See #9985
Closes #8752

PR #9988, opened 2023-03-30
Remove `prettytable`

### LINKED ISSUE #9028: Remove stubs for prettytable (not until April 2023)
`prettytable` has just  incorporated typeshed's stubs as inline annotations and added a `py.typed` file (see https://github.com/jazzband/prettytable/issues/203). We've just marked our stubs as obsolete in #9023; we should remove the stubs in April as per our [policy](https://github.com/python/typeshed/blob/master/CONTRIBUTING.md#third-party-library-removal-policy).

sobolevn:
Closes #9028


AlexWaygood (approved):
Looks good, will merge on 1 April

sobolevn:
I was never so excited for the Aprils Fools in my life! 😆

hugovk:
Hooray, that's two projects (with https://github.com/python/typeshed/pull/9986) I maintain removed from typeshed today!

Will have to remove Pillow next, but that's a bigger job...

AlexWaygood:
> Hooray, that's two projects (with #9986) I maintain removed from typeshed today!

It was a pleasure providing stubs for your projects, and an even greater pleasure deleting them 😄

PR #9989, opened 2023-03-30
Remove `Flask-SQLAlchemy`

### LINKED ISSUE #8885: Remove Flask-SQLAlchemy stubs (not until April 2023)
Version 3.0.1, released on 2022-10-11, added a `py.typed` file.

sobolevn:
Closes #8885


stubs/Flask-Migrate/METADATA.toml
```diff
@@ -1,6 +1,6 @@
 version = "4.0.*"
 # Requires a version of flask with a `py.typed` file
```
AlexWaygood, line 2:
```suggestion
# Requires versions of flask/Flask-SQLAlchemy with `py.typed` files
```

AlexWaygood (approved):
Looks good, will merge on 1 April

AlexWaygood:
Oh, looks like the way Flask-SQLAlchemy is packaged breaks some of the logic for dealing with non-types dependencies in `pytype_test.py` ://

sobolevn:
I will take a look later, there are two unrelated problems.
`pytype` fails because of `top_level.txt`
and `pyright` is just no happy with some types

AlexWaygood:
The pyright stuff is very strange. It's erroring out on some types in stubs/SQLAlchemy. But you haven't changed anything in that directory!

stubs/Flask-Migrate/METADATA.toml
```diff
@@ -1,6 +1,6 @@
 version = "4.0.*"
-# Requires a version of flask with a `py.typed` file
-requires = ["Flask>=2.0.0", "types-Flask-SQLAlchemy"]
+# Requires versions of flask/Flask-SQLAlchemy with `py.typed` files
+requires = ["Flask>=2.0.0", "Flask-SQLAlchemy>=3.0.1"]
```
AlexWaygood, line 3: I figured it out. Because you're specifying Flask-SQLAlchemy as a dependency of types-Flask-Migrate, that then pulls in SQLAlchemy as a transitive dependency of types-Flask-Migrate (because SQLAlchemy is a dependency of Flask-SQLAlchemy). That means pip now install SQLAlchemy in CI before pyright tests any of the stubs and, because it's unpinned, pip installs the latest version of SQLAlchemy available. Because `SQLAlchemy>=2` has inline types and a `py.typed` file, _that_ means that pyright is now pointing out the inconsistencies between our stubs for `SQLAlchemy<2` and the inline types in site-packages for `SQLAlchemy>=2`

AlexWaygood:
An easy way of solving this, given our Sqlalchemy stubs are already marked as obsolete, would be to just wait until we're ready to remove those stubs, and then remove them both together

stubs/Flask-Migrate/METADATA.toml
Avasam: We could just add the missing generics.

AlexWaygood: > We could just add the missing generics. So that the stubs are on par enough with the 2.0 version for this PR to pass.

It feels like a somewhat fragile solution -- if `SQLAlchemy` makes more stuff generic on the 2.0 branch that isn't generic in our stubs for `SQLAlchemy` 1.0, and they then make a release, our CI will break unexpectedly

Avasam: Then I think just waiting to remove them together, given their co-dependencies, just makes the most sense.

AlexWaygood:
Closing, as this can only be done at the same time as removing our `SQLAlchemy` stubs, and we've got #10389 open for that now

PR #9991, opened 2023-03-30
Improve `urlunsplit` and `urlunparse`

AlexWaygood:
- The first overload for `urlunsplit` and `urlunparse` is currently pointless. If you provide a tuple of a different length to these functions, type checkers simply fall back to the second overload, and don't emit an error: https://mypy-play.net/?mypy=latest&python=3.11&gist=966b9d881ce16ee67e220bff46d5f21c.
- Mypy currently infers `str` for `urlunsplit((None, None, None, None, None))`, but the returned value is `b""`: https://mypy-play.net/?mypy=latest&python=3.11&gist=dffd33f24d5e90b18cee1757d48149c6. Replace the first overload with an overload for `Iterable[None]`.
- Use `Iterable` instead of `Sequence`, since these functions accept any iterable that can be unpacked; it doesn't have to be a sequence.


hauntsaninja (approved):
Thanks, werkzeug error looks clearer (also I think the code might be genuinely buggy if someone passes empty dict)

Note, since we're caring more about the IDE setting now, this annotation will be a little less helpful, but I'm fine with it.

PR #9994, opened 2023-03-31
xml: `NodeList.length` is read-only

hamdanal:
Ref: https://github.com/python/cpython/blob/d97aef8ebfbbb275384b17f06945e583fb3189ea/Lib/xml/dom/minicompat.py#L57-L65

Also `EmptyNodeList.length` is always 0


stdlib/xml/dom/minicompat.pyi
```diff
@@ -8,11 +9,13 @@ _T = TypeVar("_T")
 StringTypes: tuple[type[str]]
 
 class NodeList(list[_T]):
-    length: int
+    @property
+    def length(self) -> int: ...
     def item(self, index: int) -> _T | None: ...
 
 class EmptyNodeList(tuple[Any, ...]):
```
AlexWaygood, line 16: What a bizarre class. If it's always meant to be, you know, empty, maybe this would be better?

```suggestion
class EmptyNodeList(tuple[()]):
```

hamdanal: Why not :)

AlexWaygood:
Thanks!

PR #9995, opened 2023-04-01
Remove stubs for `D3DShot`

AlexWaygood:
The update marking the stubs as no longer updated has been pushed to PyPI: https://pypi.org/project/types-D3DShot/


AlexWaygood:
Cc. @Avasam :)

srittau (approved):

PR #9996, opened 2023-04-01
Fix return type of protobuf.ScalarMap.get()

### LINKED ISSUE #9932: Incorrect return type on protobuf.ScalarMap; get() can return None
The following line:

https://github.com/python/typeshed/blob/a34da859660cdad9ab6cd427093a6cde9df09565/stubs/protobuf/google/protobuf/internal/containers.pyi#L70

should be:

```python
def get(self, key: _K, default: None = ...) -> _ScalarV | None: ...
```

It looks like MessageMap.get() later in the same file may have the same issue on its first overload.

spatinom:
ScalarMap and MessageMap get can return None. Fixes #9932


srittau (approved):
Looks sensible, thanks!

PR #9997, opened 2023-04-01
Add tensorflow legacy optimizers.

hmc-cs-mdrissi:
# Summary
Add stubs for 1 of 3 kinds of tensorflow optimizers.

# Discussion
Tensorflow's optimizer api has gone through revisions over years. tf.compat.v1.train.Optimizer which were deprecated in tensorflow v2. tf.keras.optimizers.legacy.Optimizer which became legacy in tf 2.11 (fairly recent) and newest tf.keras.optimizers.Optimizer. This pr focuses on second because codebase I mostly work on is still on tf 2.8. The other two should also be added eventually as all 3 are still used today and often deprecated things continue to be maintained for a long time in tensorflow under a new module name.


stubs/tensorflow/tensorflow/_internal.pyi
```diff
@@ -0,0 +1,5 @@
+# This includes internal types that are not part of the public API, but
+# are needed as annotations for the public API. Most commonly this is
+# internal base classes.
+
+class Trackable: ...
```
hmc-cs-mdrissi, line 5: I'm unsure on this choice. Trackable is a real class, but it's internal. Optimizer subclasses Trackable and there are other things that do subclass trackable and some public apis should have signature that looks like,

```python
def save_checkpoint(obj: Trackable) -> None
```

For a type like this is it preferred to keep it here or to add the real module it comes from even though that one is undocumented?

hmc-cs-mdrissi:
General 2 week poke. Ready as is with 1 small question.

stubs/tensorflow/tensorflow/_internal.pyi
JelleZijlstra: I would prefer to put it in the real module that it lives in, if possible.

stubs/tensorflow/tensorflow/__init__.pyi
```diff
@@ -282,4 +283,52 @@ class Module:
...
+    @overload
+    def gradient(
+        self,
+        target: ContainerTensors,
+        sources: ContainerTensors,
+        output_gradients: list[Tensor] | None = None,
+        unconnected_gradients: UnconnectedGradients = ...,
+    ) -> ContainerGradients: ...
+    @contextmanager
+    def stop_recording(self) -> Iterator[None]: ...
```
JelleZijlstra, line 328:
```suggestion
    def stop_recording(self) -> Generator[None, None, None]: ...
```
So that maybe one day we can make `@contextmanager` take only Generators.

stubs/tensorflow/tensorflow/keras/optimizers/legacy/__init__.pyi
```diff
@@ -0,0 +1,112 @@
...
+_Initializer: TypeAlias = str | Callable[[], tf.Tensor] | dict[str, Any]
+_Shape: TypeAlias = tf.TensorShape | Iterable[int | None]
+_Dtype: TypeAlias = tf.DType | str | None
+_LearningRate: TypeAlias = float | tf.Tensor | schedules.LearningRateSchedule | Callable[[], float | tf.Tensor]
+_GradientAggregator: TypeAlias = Callable[[list[tuple[Gradients, tf.Variable]]], list[tuple[Gradients, tf.Variable]]] | None
+_GradientTransformer: TypeAlias = (
+    Iterable[Callable[[list[tuple[Gradients, tf.Variable]]], list[tuple[Gradients, tf.Variable]]]] | None
+)
+
+# kwargs here and in other optimizers can be given better type after Unpack[TypedDict] is supported.
```
JelleZijlstra, line 21: Maybe mention PEP 692 for ease of grepping later.

stubs/tensorflow/tensorflow/_internal.pyi
hmc-cs-mdrissi: Added. I did notice that module it lives in got changed between tf 2.8 -> 2.12 silently, not mentioned in release notes.

stubs/tensorflow/tensorflow/__init__.pyi
hmc-cs-mdrissi: Done.

stubs/tensorflow/tensorflow/keras/optimizers/legacy/__init__.pyi
hmc-cs-mdrissi: Done.

PR #10001, opened 2023-04-03
Complete `openpyxl` base descriptors

Avasam:
Handling Descriptors separately from #9511 will reduce a lot of the changes and complexity from that PR.
That's a bigger task that I expected, so let's start with just those found in `descriptors/base.pyi`.
For now I've left the getter/setter of the Alias descriptor untyped. I don't think there's a good way to type it other than using generic type arguments for the setter arg and getter return types.

The type implementation was very test-driven (as you can see by the test cases), but also made me discover what seems to be a few false positives and negatives with mypy's handling of generic descriptors. Pyright seems accurate and acts as expected except for one specific case.
I have marked all of those with appropriate comments. Please validate if these are mypy issues or if I need to do something differently.

Ref to https://github.com/python/mypy/issues/13316 since it seems to affect this PR.

A lot of default arguments in class `__init__` would fail. I've added an overload when necessary. (all done in https://github.com/python/typeshed/commit/9cc16f0ad922ba31672cad2789d49ed650f14f35 )

I wouldn't mind first adding a typed heavy user of `openpyxl` or two to `mypy_primer` before this PR if there's suggestions.


Avasam:
I don't know how to fix that last pytype error

srittau (approved):
Only spot checked, but LGTM.

PR #10002, opened 2023-04-03
Add pyright testcases / regression tests to the runtests script

Avasam:


scripts/runtests.py
```diff
@@ -19,6 +19,9 @@ def colored(text: str, color: str | None = None, on_color: str | None = None, at
 
 
 _STRICTER_CONFIG_FILE = "pyrightconfig.stricter.json"
+_TESTS_CONFIG_FILE = "pyrightconfig.testcases.json"
```
AlexWaygood, line 22:
```suggestion
_TESTCASES_CONFIG_FILE = "pyrightconfig.testcases.json"
```

scripts/runtests.py
```diff
@@ -89,21 +92,45 @@ def main() -> None:
...
         print(pyright_result.stderr)
         pyright_returncode = pyright_result.returncode
         pyright_skipped = False
 
+    print(f"\nRunning Pyright regression tests for Python {_PYTHON_VERSION}...")
+    pyright_testcases_result = subprocess.run(
+        [
+            sys.executable,
+            "tests/pyright_test.py",
+            Path(path) / "@tests",
```
AlexWaygood, line 116: We don't at the moment, but it's possible we might start putting more `.py` files in the `@tests/` directories in the future; better to specifically test the `@tests/test_cases` directory.

```suggestion
            Path(path) / "@tests" / "test_cases",
```

scripts/runtests.py
```diff
@@ -89,21 +92,45 @@ def main() -> None:
...
+            "--pythonversion",
+            _PYTHON_VERSION,
+            "-p",
+            _TESTS_CONFIG_FILE,
```
AlexWaygood, line 120:
```suggestion
            _TESTCASES_CONFIG_FILE,
```

scripts/runtests.py
```diff
@@ -89,21 +92,45 @@ def main() -> None:
...
+        ],
+        stderr=subprocess.PIPE,
```
AlexWaygood, line 122: (No functional difference, but I find this more readable)

```suggestion
        capture_output=True,
```

AlexWaygood (commented):
Makes sense! A few minor points below.

Also, if there are no test cases for a stubs package, on my machine (without pyright installed) a message is printed to the terminal saying that the regression tests for pyright were _skipped_. But that's not really accurate if there are no regression tests to skip!

Avasam:
> Also, if there are no test cases for a stubs package, on my machine (without pyright installed) a message is printed to the terminal saying that the regression tests for pyright were _skipped_. But that's not really accurate if there are no regression tests to skip!

Skipped tests means no work was done towards determining whether it would fail or pass on CI. If the npx command fails (ie not installed), the test is skipped. So it's intended.

That being said, the lack of test_cases is an automatic success (0/0 tests passed). And we check for that early. Imo it would also be faster to do that additional check rather than letting pyright setup and run on 0 files.

I also just thought that I didn't try or account for stdlib.

scripts/runtests.py
Avasam: The output no longer appears in the console. And that was in line with the rest of the tests in this file. I've reverted it.

Avasam:
Updated to first check for the existance of the `test_cases` folder. It should also now work with stdlib.

AlexWaygood (approved):
Looks good!

This still doesn't do what I'd expect:

```
(.venv) typeshed>python scripts/runtests.py stdlib
usage: runtests.py [-h] [--run-stubtest] path
runtests.py: error: 'path' argument should be in format <folder>/<stub>.
```

But we have that behaviour on `main` as well; fixing it can wait for another PR

PR #10003, opened 2023-04-03
Added fixers to lib2to3

### LINKED ISSUE #2510: Lib2to3 stubs are incomplete
The stubs for `lib2to3` are incomplete. In particular they are missing the modules `lib2to3.fixer_base` and `lib2to3.main`.

The former is required to write custom fixers and the latter is required to write your own command line tool for running your fixes.

Avasam:
Did this a while ago and forgot to PR. Wanted to take a look at some of the oldest issues and this one didn't seem too hard.
Closes #2510 , the least recently updated issue in typeshed (as of starting this PR)!

Add all missing stubs in lib2to3. pyright-complete most of it.


JelleZijlstra:
Thanks for all the work, but given that this is a lot of new stubs and lib2to3 is deprecated (to be removed in 3.13), I'm not sure we should really be adding all this.

Avasam:
Up to you! There's still a while before 3.13. Most of the changes came from stubgen or some search&replace.

srittau:
Personally, I'm fine with adding this, although we could do a superficial review for the newly added files.

JelleZijlstra:
Yes, I suppose there's no harm in it. I will take a quick look and merge this.

stdlib/lib2to3/btm_matcher.pyi
```diff
@@ -0,0 +1,30 @@
...
+    nodes: Incomplete
+    fixers: Incomplete
+    logger: Incomplete
+    def __init__(self) -> None: ...
+    def add_fixer(self, fixer: BaseFix) -> None: ...
+    def add(self, pattern: SupportsGetItem[int | slice, Incomplete] | None, start: BMNode) -> list[BMNode]: ...
+    def run(self, leaves: Iterable[Leaf]) -> defaultdict[BaseFix, list[Node | Leaf]]: ...
+    def print_ac(self) -> None: ...
+
+def type_repr(type_num: _T) -> _T: ...
```
JelleZijlstra, line 30:
```suggestion
def type_repr(type_num: int) -> str | int: ...
```
It takes the number of a parse tree node and returns either its name or the number again if it isn't found.

stdlib/lib2to3/fixes/fix_future.pyi
```diff
@@ -0,0 +1,9 @@
+from typing import ClassVar
+from typing_extensions import Literal, LiteralString
+
+from .. import fixer_base
+
+class FixFuture(fixer_base.BaseFix):
+    BM_compatible: ClassVar[Literal[True]]
+    PATTERN: ClassVar[LiteralString]
```
JelleZijlstra, line 8: I don't think LiteralString makes much sense in cases like this. Let's avoid using it.

PR #10004, opened 2023-04-03
Accept int in tkinter.Canvas dash arguments

### LINKED ISSUE #7723: tkinter canvas dash
I have some code that does:

```
            self._canvas.create_rectangle(
                event.x,
                event.y,
                self._initial_drag_x,
                self._initial_drag_y,
                dash=5,
                outline="black",
                tags="selection_displayer_rect",
            )
```
It fails to type check, unless I comment out `dash=5`, but it works at runtime.

Akuli:
Fixes #7723


srittau (approved):
Didn't verify, but looks sensible, and you're our resident tkinter expert.

PR #10006, opened 2023-04-04
`dataclasses.make_dataclass` has a new parameter in 3.12

sobolevn:
Source https://github.com/python/cpython/blob/810d365b5eb2cf3043957ca2971f6e7a7cd87d0d/Lib/dataclasses.py#L1403


hauntsaninja (approved):
Thank you!

PR #10008, opened 2023-04-04
Various improvements to `Markdown` stubs

AlexWaygood:
- Delete various unnecessary `= ...`s
- Add `ClassVar` annotations where applicable
- Delete several attributes and classes that no longer appear to exist at runtime
- Improve various types


srittau (approved):
Thanks!

PR #10016, opened 2023-04-06
Add a stub for sys.thread_info

tomasr8:
Adds a stub for [sys.thread_info](https://docs.python.org/3/library/sys.html#sys.thread_info)

This property was added in 3.3, should I add a version check? I haven't found any check for versions this old..

### DIFF
--- a/stdlib/sys.pyi
+++ b/stdlib/sys.pyi
@@ -201,8 +201,8 @@ class _int_info(structseq[int], tuple[int, int, int, int]):
     @property
     def str_digits_check_threshold(self) -> int: ...
 
-_ThreadInfoName = Literal["nt", "pthread", "pthread-stubs", "solaris"]
-_ThreadInfoLock = Literal["semaphore", "mutex+cond"] | None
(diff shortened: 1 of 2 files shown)


JelleZijlstra:
No version check needed, as far as we're concerned Python 3.7 was the first ever version of Python.

Akuli:
You need to delete `sys.thread_info` from `tests/stubtest_allowlists/py3_common.txt`.

stdlib/sys.pyi
```diff
@@ -201,6 +201,20 @@ class _int_info(structseq[int], tuple[int, int, int, int]):
     @property
     def str_digits_check_threshold(self) -> int: ...
 
+_ThreadInfoName = Literal["nt", "pthread", "pthread-stubs", "solaris"]
+_ThreadInfoLock = Literal["semaphore", "mutex+cond"] | None
```
Akuli, line 205:
```suggestion
_ThreadInfoName: TypeAlias = Literal["nt", "pthread", "pthread-stubs", "solaris"]
_ThreadInfoLock: TypeAlias = Literal["semaphore", "mutex+cond"] | None
```

Akuli (approved):
Thanks!

PR #10019, opened 2023-04-07
Fix target argument of pathlib.Path.{sym,hard}link_to

viccie30:
The underlying `os.{sym,}link` functions take any PathLike object:
https://github.com/python/typeshed/blob/1c0500a57050815102c702efd053e09770a5ee88/stdlib/os/__init__.pyi#L749-L751, https://github.com/python/typeshed/blob/1c0500a57050815102c702efd053e09770a5ee88/stdlib/os/__init__.pyi#L700-L707.


srittau (changes requested):
Thanks! Just like in `os`, we should just use `_typeshed.StrOrBytesPath` here.

AlexWaygood:
> Thanks! Just like in `os`, we should just use `_typeshed.StrOrBytesPath` here.

I haven't looked too closely at this PR, but (unlike `os`), `pathlib` only supports string paths, not bytes paths, so it should probably be `_typeshed.StrPath` rather than `_typeshed.StrOrBytesPath` here?

AlexWaygood:
> The arguments to these functions skip any pathlib handling and just call the underlying functions, so bytes are accepted.

My worry is that even if bytes paths technically work for this method, you probably shouldn't be using bytes paths with anything pathlib-y, it's bound to bite you at some point.

But no strong opinion; happy for us to go with `StrOrBytesPath` since it is allowed by the implementation, at least at the moment.

viccie30:
I have made the requested change to `StrOrBytesPath`.

srittau (approved):
Thanks!

PR #10027, opened 2023-04-09
Update `croniter` to `1.3.10`

sobolevn:
Fixing https://github.com/python/typeshed/actions/runs/4647756507/jobs/8224937971
Closes https://github.com/python/typeshed/issues/10022


AlexWaygood (approved):
Thanks!

PR #10031, opened 2023-04-11
Add more correct type annotations for `tomllib`

wchistow:
I've added more correct type annotations for `tomllib` module, because it was `Any`, but there is a [table](https://docs.python.org/3/library/tomllib.html#conversion-table) of types in [documentation](https://docs.python.org/3/library/tomllib.html).

### DIFF
--- a/stdlib/tomllib.pyi
+++ b/stdlib/tomllib.pyi
@@ -1,10 +1,14 @@
 from _typeshed import SupportsRead
 from collections.abc import Callable
-from typing import Any
+from datetime import datetime, date, time
+from typing import Any, TypeAlias
 
 __all__ = ("loads", "load", "TOMLDecodeError")
 
 class TOMLDecodeError(ValueError): ...
 
-def load(__fp: SupportsRead[bytes], *, parse_float: Callable[[str], Any] = ...) -> dict[str, Any]: ...
-def loads(__s: str, *, parse_float: Callable[[str], Any] = ...) -> dict[str, Any]: ...
+TOMLValue: TypeAlias = int | float | str | bool | list['TOMLValue'] | \
+                       dict['TOMLValue', 'TOMLValue'] | datetime | date | time
+
+def load(__fp: SupportsRead[bytes], *, parse_float: Callable[[str], Any] = ...) -> dict[str, TOMLValue]: ...
+def loads(__s: str, *, parse_float: Callable[[str], Any] = ...) -> dict[str, TOMLValue]: ...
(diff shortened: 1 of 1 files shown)


AlexWaygood:
Note to reviewers: mypy_primer won't be much use on this PR, since we run mypy_primer on Python 3.10, and tomllib is new in Python 3.11.

AlexWaygood:
> Note to reviewers: mypy_primer won't be much use on this PR, since we run mypy_primer on Python 3.10, and tomllib is new in Python 3.11.

We now run mypy_primer on 3.11 following #9336. I've merged `main` into this branch so that we get a mypy_primer re-run.

stdlib/tomllib.pyi
```diff
@@ -1,10 +1,13 @@
 from _typeshed import SupportsRead
 from collections.abc import Callable
-from typing import Any
+from datetime import date, datetime, time
+from typing import Any, TypeAlias
```
AlexWaygood, line 4:
```suggestion
from typing import Any
from typing_extensions import TypeAlias
```

AlexWaygood:
Just taking the first pytest error in the mypy_primer diff as an example, they're doing this:

```py
config = tomllib.loads(toml_text)
result = config.get("tool", {}).get("pytest", {}).get("ini_options", None)
```

That seems like pretty idiomatic use of `tomllib` to me, but with this PR, mypy would complain loudly about it.

The stdlib `tomllib` is also exactly the same code as the third-party library `tomli`, which uses `dict[str, Any]` as its return types for these functions: https://github.com/hukkin/tomli/blob/1bd3345f97cba795d7e6075956815c0a52151ed0/src/tomli/_parser.py#L57-L132. I'm not sure it makes sense for typeshed to deviate from `tomli`'s annotations here, given most libraries will be only using `tomllib` on 3.11+, and will be falling back to `tomli` on Python <=3.10.

Lastly, even if we decided that this PR is worth the disruption, it's not currently correct. The docs state:

> _parse_float_ will be called with the string of every TOML float to be decoded. By default, this is equivalent to `float(num_str)`. This can be used to use another datatype or parser for TOML floats (e.g. [decimal.Decimal](https://docs.python.org/3/library/decimal.html#decimal.Decimal)).

That means that if the user provides a value to the `parse_float` parameter, you could have different types being returned. We could possibly represent this by using overloads and making `_TOMLValue` a generic-recursive type alias, but this would make the type annotation very complicated, and increase the risk of type checkers emitting false positives for users of `tomllib`.

srittau:
I agree that the current annotation of `dict[str, Any]` is probably the best we can do for now, until we get permissive unions (python/typing#566).

AlexWaygood:
Let's close this for now, then. Happy to reopen if @hauntsaninja or @hukkin disagree strongly with this decision :)

PR #10032, opened 2023-04-11
pywin32: add `HANDLEType` class

glyph:
per http://timgolden.me.uk/pywin32-docs/PyHANDLE.html these objects actually have methods, they are not ints

This probably needs a much more comprehensive trip through all the pywin32 stubs looking for all the places that are incorrectly documented to return `int`


srittau:
Thanks! I don't have access to a Windows installation to check, at the moment, but is `HANDLEType` really available at runtime? Stubtest doesn't seem to think so, but it isn't always right. Also, why is it called `PyHANDLE` in the documentation?

AlexWaygood:
> Thanks! I don't have access to a Windows installation to check, at the moment, but is `HANDLEType` really available at runtime? Stubtest doesn't seem to think so, but it isn't always right.

It's available on my machine locally:

```pycon
>>> import pywintypes
>>> h = pywintypes.HANDLEType
>>> h.Close
<method 'Close' of 'PyHANDLE' objects>
>>> h.close
<method 'close' of 'PyHANDLE' objects>
>>> h.Detach
<method 'Detach' of 'PyHANDLE' objects>
>>> h.__int__
<slot wrapper '__int__' of 'PyHANDLE' objects>
```

Looking at the CI errors, stubtest appears to be complaining about a property on the `HANDLEType` class being _missing_ from the stub, rather than things existing in the stub that don't exist at runtime.

@glyph, would you be able to add the missing `handle` property?

glyph:
> > Also, why is it called `PyHANDLE` in the documentation?
> 
> No idea why those docs refer to it as `PyHANDLE`, but there's definitely no such object in the `pywintypes` module at runtime:
> 
> ```
> >>> import pywintypes
> >>> pywintypes.PyHANDLE
> Traceback (most recent call last):
>   File "<stdin>", line 1, in <module>
> AttributeError: module 'pywintypes' has no attribute 'PyHANDLE'. Did you mean: 'HANDLE'?
> ```
> 
> Maybe the name changed at some point? `¯\_(ツ)_/¯`

I'm pretty sure it's always been this way, just following the standard naming convention for low-level types in Python, which is to name them something inscrutably, randomly different than what they're called at runtime to make them confusingly difficult to find in their appropriate place:

```
Python 3.11.2 (v3.11.2:878ead1ac1, Feb  7 2023, 10:02:41) [Clang 13.0.0 (clang-1300.0.29.30)] on darwin
>>> def foo():
...  pass
... 
>>> type(foo)
<class 'function'>
>>> function
NameError: name 'function' is not defined
>>> type(foo).__module__
'builtins'
>>> import builtins
>>> builtins.function
AttributeError: module 'builtins' has no attribute 'function'
>>> foo.__code__
<code object foo at 0x105b5fdb0, file "<stdin>", line 1>
>>> code
NameError: name 'code' is not defined
>>> foo.__code__.__class__.__module__
'builtins'
>>> builtins.code
AttributeError: module 'builtins' has no attribute 'code'
>>> builtins
<module 'builtins' (built-in)>
>>> type(builtins)
<class 'module'>
>>> module
NameError: name 'module' is not defined
>>> type(builtins).__module__
'builtins'
>>> builtins.module
AttributeError: module 'builtins' has no attribute 'module'
>>> import types
>>> types.ModuleType is type(builtins)
True
>>> types.FunctionType is type(foo)
True
>>> types.CodeType is type(foo.__code__)
True
```

I am pretty sure `import pywintypes` is literally named by analogy to `import types`.

stubs/pywin32/win32/lib/pywintypes.pyi
```diff
@@ -15,6 +15,12 @@ class error(Exception):
 class com_error(Exception): ...
 class UnicodeType(str): ...
 
+class HANDLEType:
+    def Close(self) -> None: ...
```
glyph, line 19:
```suggestion
    handle: int
    def Close(self) -> None: ...
```

stubs/pywin32/win32/lib/pywintypes.pyi
```diff
@@ -15,6 +15,13 @@ class error(Exception):
 class com_error(Exception): ...
 class UnicodeType(str): ...
 
+class HANDLEType:
+    handle: int
```
glyph, line 19:
```suggestion
    @property
    def handle(self) -> int: ...
```

glyph: oops, it's readonly

glyph:
> @glyph, would you be able to add the missing `handle` property?

Done. And fixed `__init__` as well.

stubs/pywin32/win32/lib/pywintypes.pyi
```diff
@@ -15,6 +16,15 @@ class error(Exception):
 class com_error(Exception): ...
 class UnicodeType(str): ...
 
+class HANDLEType:
+    def __init__(self) -> NoReturn: ...
```
AlexWaygood, line 20:
```suggestion
    def __init__(self, *args: Never, **kwargs: Never) -> NoReturn: ...
```

stubs/pywin32/win32/lib/pywintypes.pyi
```diff
@@ -2,6 +2,7 @@
 # "KeyError: 'pywintypes'"
 from _typeshed import Incomplete
 from datetime import datetime
+from typing import NoReturn
 from typing_extensions import Literal
```
AlexWaygood, line 6:
```suggestion
from typing_extensions import Literal, Never
```

AlexWaygood (commented):
Suggested fix for the new stubtest complaints:

Avasam:
> I'm pretty sure it's always been this way, just following the standard naming convention for low-level types in Python, which is to name them something inscrutably, randomly different than what they're called at runtime to make them confusingly difficult to find in their appropriate place

This. You'll see a lot of types I have added in a private stub because they are not exposed directly anywhere that I know of at runtime. Or only an alias is exposed. Which is confusing when you call `type()` and it returns a class name you can't import from anywhere.

Also pywin32's documentation is not to be trusted at face value. A LOT is missing or obsolete,

glyph:
Thanks!

PR #10033, opened 2023-04-12
zipfile.Path.open() signature fix

### LINKED ISSUE #9966: zipfile.Path.open() signature not correct
MRE:

```python
from typing import BinaryIO
from zipfile import Path as ZipPath


def load_results(zip_file: BinaryIO):
    zip_path = ZipPath(zip_file)
    for unpacked_file in zip_path.iterdir():
        if not unpacked_file.is_file():
            continue
        if unpacked_file.name == "results.json":
            with unpacked_file.open("r", encoding="utf8") as results_file:  # line 11
                print(type(results_file.read()))  # line 12
                continue
        if unpacked_file.name == "image.jpg":
            with unpacked_file.open("rb") as image:  # line 15
                print(type(image.read()))


if __name__ == '__main__':
    with open("example.zip", "rb") as f:
        load_results(f)
```
To run the script, please add an `example.zip` file with `results.json` (with any text in UTF-8, including Cyrillic symbols) and an `image.jpg` file to your working directory.

In PyCharm (which uses typeshed), I am receiving warnings about the following lines of code:

* Line 11: `"Unexpected argument"` regarding the encoding keyword argument.
* Line 15: `"Expected type 'Literal["r", "w"]', got 'Literal["rb"]' instead"` regarding the type of the file mode.

In both cases, the code appears to work correctly:
* Without the encoding keyword argument, I encounter a UnicodeDecodeError.
* The print statements show different types (`<class 'str'>` for line 12 and `<class 'bytes'>` for line 15) as expected when using `"r"` or `"rb"` file mode.

grigoriev-semyon:


stdlib/zipfile.pyi
```diff
@@ -245,7 +253,13 @@ if sys.version_info >= (3, 8):
...
         if sys.version_info >= (3, 9):
             def open(
-                self, mode: _ReadWriteBinaryMode = "r", *args: Any, pwd: bytes | None = None, **kwargs: Any
+                self,
+                mode: _ReadWriteBinaryMode = "r",
+                *args: Any,
+                pwd: bytes | None = None,
+                encoding: str | None = None,
+                force_zip64: bool = ...,
+                **kwargs: Any,
```
srittau, line 262: I just checked and it seems that `encoding` is only supported on Python 3.10+, so we would need another branch for that.

I don't see `force_zip64` supported in any version?

srittau (changes requested):
Thanks, remarks below.

stdlib/zipfile.pyi
grigoriev-semyon: Yes, `force_zip64` don't supported in versions >= 3.9.

`encoding` don't supported in versions < 3.9
`encoding` can be supplied as a kwarg in py3.9
Changed in version 3.10.10: The `encoding` parameter can be supplied as a positional argument

srittau (changes requested):
Oops, my bad. I misread the code. As `*args` is just passed to `TextIOWrapper` on Python 3.9 and the second argument is `encoding`, it can be passed by position on 3.9 as well. So we don't need the 3.9 branch after all and the 3.10 branch can be used for 3.9+.

srittau (changes requested):
Oops, my bad. I misread the code. As `*args` is just passed to `TextIOWrapper` on Python 3.9 and the second argument is `encoding`, it can be passed by position on 3.9 as well. So we don't need the 3.9 branch after all and the 3.10 branch can be used for 3.9+.

srittau (approved):

PR #10035, opened 2023-04-12
Update stubs with new ujson features

dbergamin:
Reference from ujson 5.7.0:
https://github.com/ultrajson/ultrajson/blob/5.7.0/python/objToJSON.c#L797 https://github.com/ultrajson/ultrajson/blob/5.7.0/tests/test_ujson.py

### DIFF
--- a/stubs/ujson/ujson.pyi
+++ b/stubs/ujson/ujson.pyi
@@ -2,7 +2,6 @@ from _typeshed import Incomplete
 from collections.abc import Callable
 from typing import IO, Any, AnyStr
 
-
 __version__: str
 
 def encode(
--- a/stubs/ujson/ujson.pyi
+++ b/stubs/ujson/ujson.pyi
@@ -1,4 +1,7 @@
+from _typeshed import Incomplete
+from collections.abc import Callable
 from typing import IO, Any, AnyStr
+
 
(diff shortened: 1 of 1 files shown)


stubs/ujson/ujson.pyi
```diff
@@ -10,6 +10,10 @@ def encode(
     escape_forward_slashes: bool = ...,
     sort_keys: bool = ...,
     indent: int = ...,
+    allow_nan: bool = ...,
+    reject_bytes: bool = ...,
+    default: Any = ...,
+    separators: Any = ...,
```
srittau, line 16:
```suggestion
    default: Callable[[Unknown], Unknown] | None = None,
    separators: tuple[str, str] | None = None,
```

Same for the other two instances below. Needs import of `_typeshed.Unknown` and possible `Callable`.

srittau (changes requested):
Thanks! A few further type improvements below:

stubs/ujson/ujson.pyi
srittau: Yikes, indeed.

dbergamin: thanks for the feedback! have improved the precision as suggested

Issue #10024, opened 2023-04-08
OpenSSL.SSL.SSLeay_version returns bytes, not str


wRAR:
The stub has `def SSLeay_version(type: int) -> str`, while in practice:

```
In [3]: OpenSSL.SSL.SSLeay_version(OpenSSL.SSL.SSLEAY_VERSION)
Out[3]: b'OpenSSL 3.0.7 1 Nov 2022'
```

The function itself, at least in this version, is just `_ffi.string(_lib.OpenSSL_version(type))`, and as OpenSSL_version(3) returns a `char*` it looks correct to get `bytes` in Python.

srittau:
Thanks! #10037 should fix this.

PR #10037, opened 2023-04-12
[pyOpenSSL] Fix return type of SSLeay_version

srittau:
Closes: #10024


srittau:
Primer hit actually already accounts for that: https://github.com/mitmproxy/mitmproxy/blob/be02b1e298dfb1b218007f824c4cf345c53b0cee/mitmproxy/utils/debug.py#L21-L23

```python
    openssl_version = SSL.SSLeay_version(SSL.SSLEAY_VERSION)
    if isinstance(openssl_version, bytes):
        openssl_version = openssl_version.decode()
```

AlexWaygood (approved):

PR #10047, opened 2023-04-14
Fix argument and return types of os.fspath

viccie30:
If passed an instance of a subclass of `str` or `bytes`, that same instance is returned, not a `str` or `bytes` object. The same is true for the return value of the argument's `__fspath__` method if the argument is `PathLike`.

This goes for both the C ([here](https://github.com/python/cpython/blob/be8903eb9d66ef1229f93a3a6036aeafc3bb0bda/Modules/posixmodule.c#L15214-L15257)) and Python ([here](https://github.com/python/cpython/blob/be8903eb9d66ef1229f93a3a6036aeafc3bb0bda/Lib/os.py#L1042-L1069)) implementations of `os.fspath`.


JelleZijlstra:
Covariance is meaningless in a function; it only makes a difference for classes. I don't think this fixes anything, and as you can see it introduces some new errors.

hauntsaninja:
You might want two typevars that are bound by str and bytes

viccie30:
Thanks for the feedback @JelleZijlstra. I'll see if I can come up with something using @hauntsaninja's suggestion.

PR #10051, opened 2023-04-16
Add new enums in PIL.ExifTags

dairiki:
This PR adds the `Base` and `GPS` enums to `PIL.ExifTags`.

These were added in Pillow 9.3.0 (via python-pillow/Pillow#6630).


srittau (approved):
Thanks!

PR #10052, opened 2023-04-16
tensorflow feature columns

hmc-cs-mdrissi:
# Summary
Add stubs for feature_column [api](https://www.tensorflow.org/api_docs/python/tf/feature_column). They fully cover documented api.

Most of functions return undocumented FeatureColumn classes. I've included stubs for FeatureColumn classes with some common easier methods, but those are incomplete and as they're all undocumented I did not put fallback getattr for each one.


stubs/tensorflow/tensorflow/io/__init__.pyi
```diff
@@ -0,0 +1,101 @@
...
+
+class SparseFeature(NamedTuple):
+    index_key: str | list[str]
+    value_key: str
+    dtype: _DTypeLike
+    size: int | list[int]
+    already_sorted: bool = False
+
+class RaggedFeature(NamedTuple):
+    class RowSplits(NamedTuple):
```
hmc-cs-mdrissi, line 75: This mypy dislikes and I'm unsure how to type this. The real code uses collections.namedtuple and does have nested [namedtuples](https://github.com/tensorflow/tensorflow/blob/0db597d0d758aba578783b5bf46c889700a45085/tensorflow/python/ops/parsing_config.py#L177). But mypy gives an error of,

```
Invalid statement in NamedTuple definition; expected "field_name: field_type [= default]"
```

because of them. pyright is happy with them. I could move the nested namedtuples out and having them marked private like, _RowSplits(NamedTuple). That does disagree with runtime as RaggedFeature.RowSplits is the real way to access them, but unsure how to deal with it. Or maybe just type: ignore[misc] for them?

stubs/tensorflow/tensorflow/python/feature_column/feature_column_v2.pyi
```diff
@@ -0,0 +1,271 @@
...
+    def parents(self) -> list[FeatureColumn | str]: ...
+
+class DenseColumn(FeatureColumn, metaclass=ABCMeta): ...
+class SequenceDenseColumn(FeatureColumn, metaclass=ABCMeta): ...
+
+# These classes are mostly subclasses of collections.namedtuple but we can't use
+# typing.NamedTuple because they use multiple inheritance with other non namedtuple classes.
+# _cls instead of cls is because collections.namedtuple uses _cls for __new__.
+class NumericColumn(DenseColumn):
+    key: str
```
hmc-cs-mdrissi, line 35: As these are attributes from collections.namedtuple subclassing they are read only and could maybe be marked property? Sorta neutral to being more precise here unsure namedtuple fields are properties either. Especially as right now it's little more lenient and all these classes aren't documented so being stricter seems unnecessary.

stubs/tensorflow/tensorflow/io/__init__.pyi
```diff
@@ -0,0 +1,103 @@
...
+
+    class ValueRowIds(NamedTuple):  # type: ignore[misc]
+        key: str
+
+    class UniformRowLength(NamedTuple):  # type: ignore[misc]
+        length: int
+    dtype: _DTypeLike
+    value_key: str | None = None
+    partitions: tuple[RowSplits | RowLengths | RowStarts | RowLimits | ValueRowIds | UniformRowLength, ...] = ()  # type: ignore[name-defined]
+    row_splits_dtype: _DTypeLike = ...
```
hmc-cs-mdrissi, line 97: This is a recent bug/regression in pyright. It passes on 1.1.295 but fails on 1.1.303. I'll try to minimize this and file a bug report to pyright.

hmc-cs-mdrissi: Hmm I can't type ignore this with current pyright config settings. We have reportUnnecessaryTypeIgnoreComment set to error in both pyrightconfig.json and pyrightconfig.stricter.json, but this only errors with pyrightconfig.stricter.json. If a line hits an error in stricter file but not looser one it becomes impossible to ignore. @Avasam

stubs/tensorflow/tensorflow/io/__init__.pyi
```diff
@@ -0,0 +1,104 @@
...
+    def close(self) -> None: ...
+    def __enter__(self) -> Self: ...
+    def __exit__(
+        self, exc_type: type[BaseException] | None, exc_val: BaseException | None, exc_tb: TracebackType | None
+    ) -> None: ...
+
+class FixedLenFeature(NamedTuple):
+    shape: _ShapeLike
+    dtype: _DTypeLike
+    default_value: _TensorCompatible | None = None
```
hmc-cs-mdrissi, line 56: @rchen152 

```
  File "/Users/pa-loaner/Snapchat/Dev/.venvs/typeshed/lib/python3.9/site-packages/pytype/ast/visitor.py", line 55, in _call_visitor
    return visitor(node)
  File "/Users/pa-loaner/Snapchat/Dev/.venvs/typeshed/lib/python3.9/site-packages/pytype/pyi/parser.py", line 477, in visit_AnnAssign
    raise ParseError(msg)
pytype.pyi.types.ParseError:   File: "stubs/tensorflow/tensorflow/io/__init__.pyi", line 56
    default_value: _TensorCompatible | None = None
   ^
ParseError: Default value for default_value: typing.Union can only be '...', got None
```

I think this error happens with any default value in NamedTuple in a stub file. A more minimal example that also triggers an error like this is,

```python
class Foo(NamedTuple):
  x: bool = False
```

I'll file a bug report on pytype as well. This only happens with .pyi files and namedtuple defaults work fine for .py files.

hmc-cs-mdrissi:
I think this is first typeshed pr I've had that found bug report for all 3 type checkers used in CI. A different bug for each one although all related sort of to NamedTuples.

stubs/tensorflow/tensorflow/io/__init__.pyi
hmc-cs-mdrissi: Already fixed by pyright. I'm fine waiting til next release and updating pyright version.

stubs/tensorflow/tensorflow/io/__init__.pyi
hmc-cs-mdrissi: This also isn't an error that would be good to pytype_exclude_list as right now it causes a parse error so any file that imports this one also fails transitively and I think many tensorflow files would wind up needing to be excluded.

hmc-cs-mdrissi: My workaround is just drop defaults for now and use `...` and I'll add them back later after pytype allows namedtuple default values.

stubs/tensorflow/tensorflow/io/__init__.pyi
Avasam: Pyright releases are pretty quick, I reckon you can probably wait for this to be fixed, and in the mean time have this PR complete while assuming it'll work.
Otherwise you could add `stubs/tensorflow/tensorflow/io/__init__.pyi` tot eh stricter configs exclusions, but I'm pretty sure the next release of Pyright will be soon enough that won't be necessary.

AlexWaygood: I believe Eric does a new pyright release every Tuesday, so yeah, let's just wait for the new release :)

AlexWaygood:
> I think this is first typeshed pr I've had that found bug report for all 3 type checkers used in CI. A different bug for each one although all related sort of to NamedTuples.

Wow, you really hit the jackpot on this one!

JelleZijlstra (approved):

stubs/tensorflow/tensorflow/io/gfile.pyi
```diff
@@ -0,0 +1,11 @@
+from _typeshed import Incomplete
+from collections.abc import Iterable
+
+def rmtree(path: str) -> None: ...
```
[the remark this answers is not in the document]
hmc-cs-mdrissi: Looks to be yes. Just tested and  listdir works with Path[str]. Will test each manually and update them.

JelleZijlstra:
Also looks like a bunch of the `# pyright: ignore`s can be removed.

pyrightconfig.json
```diff
@@ -7,7 +7,11 @@
     ],
     "exclude": [
         // test cases use a custom config file
-        "stubs/**/@tests/test_cases"
+        "stubs/**/@tests/test_cases",
+        // Bug in pyright that can't be ignored as it passes this one,
+        // but fails pyrightconfig.stricter.json. Can be removed after
+        // upgrading to 1.1.304 and removing ignores.
+        "stubs/tensorflow/tensorflow/io/__init__.pyi",
```
AlexWaygood, line 14: Can this be removed now?

stubs/tensorflow/tensorflow/io/gfile.pyi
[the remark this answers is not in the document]
hmc-cs-mdrissi: All of them support Path except for glob. Also discovered they all support both str/bytes paths.

pyrightconfig.json
hmc-cs-mdrissi: Yes it can be removed. I'll cleanup pyright ignores/this later today.

PR #10055, opened 2023-04-17
Add missing stubs for `logging.config`

tomasr8:
Noticed that the logging module was a few items short from being complete.
Most of the classes with the exception of `DictConfigurator` seem to be undocumented.


stdlib/logging/config.pyi
```diff
@@ -60,3 +87,57 @@ else:
...
+    def pop(self, key: Hashable, default: Any = None) -> Any: ...
+
+class ConvertingList(list[Any], ConvertingMixin):  # undocumented
+    @overload
+    def __getitem__(self, key: SupportsIndex) -> Any: ...
+    @overload
+    def __getitem__(self, key: slice) -> Any: ...
+    def pop(self, idx: SupportsIndex = -1) -> Any: ...
+
+class ConvertingTuple(tuple[Any], ConvertingMixin):  # undocumented
```
JelleZijlstra, line 107:
```suggestion
class ConvertingTuple(tuple[Any, ...], ConvertingMixin):  # undocumented
```

PR #10056, opened 2023-04-17
unittest.mock.__version__ was removed in 3.9

FasterSpeeding:
See the first entry in https://docs.python.org/3/whatsnew/3.9.html#removed


AlexWaygood (approved):
Thanks!

PR #10057, opened 2023-04-17
Add missing exception classes in `setuptools.errors`

maresb:
As part of the `distutils` deprecation implemented in #9795 by @Avasam, several `distutils`-related exceptions were removed from `setuptools.errors`. I think that some exceptions may have unintentionally been removed, so I have added them back in this PR.

Indeed, evidence that `setuptools.errors` should not be entirely deprecated can be found in [PEP 632](https://peps.python.org/pep-0632/#migration-advice), where `setuptools` is suggested as a substitute for `distutils.errors`.


Avasam:
> I think that some exceptions may have unintentionally been removed, so I have added them back in this PR.

Small correction:
I don't think anything was removed from `setuptools.errors` (https://github.com/python/typeshed/pull/9795/files#diff-7881e6b106cbf3dcded0f78acc68b6b1144200ed8d6f96d7bb6e8cdbffc9eda2), it's more like they've always been missing, because `setuptools/METADATA.toml` has `ignore_missing_stub = true` https://github.com/python/typeshed/blob/main/stubs/setuptools/METADATA.toml#L4

So thanks for adding missing stubs!

AlexWaygood:
> Small correction:
> I don't think anything was removed from `setuptools.errors` (https://github.com/python/typeshed/pull/9795/files#diff-7881e6b106cbf3dcded0f78acc68b6b1144200ed8d6f96d7bb6e8cdbffc9eda2), it's more like they've always been missing, because `setuptools/METADATA.toml` has `ignore_missing_stub = true` https://github.com/python/typeshed/blob/main/stubs/setuptools/METADATA.toml#L4

This is true, but it is also true, looking at it again, that many exception classes were removed from `setuptools/_distutils/errors.pyi`, meaning users can no longer import them from `setuptools._distutils.errors`: https://github.com/python/typeshed/pull/9795/files#diff-8a7c86f8b0bf8367e2bbc71a6168e13ae3b279440ab5775da788c36a502854eeR1-L19. These classes weren't then added to `setuptools/errors.pyi`, so users now have nowhere they _can_ import them from. This was probably an oversight :)

AlexWaygood (approved):
Thanks! I've tweaked the PR a little bit so that it's a bit more faithful to what happens at runtime. Will merge once CI passes.

Avasam:
Given the push for the deprecation, it'd be a good idea indeed to ensure that all stubs in distutils are accounted for. There aren't too many stubtest errors left. I should have a PR ready pretty soon.

maresb:
Awesome, thanks so much for the extremely fast response!!!

PR #10061, opened 2023-04-18
stdlib/xml: fix return types for toxml/toprettyxml methods

### LINKED ISSUE #10060: Incorrect return types for xml.dom.minidom.Node.{toxml,toprettyxml}
It looks like #9917 added the `-> str` return type to these functions, but the [documentation](https://docs.python.org/3/library/xml.dom.minidom.html#xml.dom.minidom.Node.toxml) indicates that _may_ return `bytes`:

>  Return a string or byte string containing the XML represented by the DOM node.
>
> With an explicit encoding 1 argument, the result is a byte string in the specified encoding. With no encoding argument, the result is a Unicode string, and the XML declaration in the resulting string does not specify an encoding. Encoding this string in an encoding other than UTF-8 is likely incorrect, since UTF-8 is the default encoding of XML.

This small example seems to confirm that it can return `bytes`:

```py
from xml.dom.minidom import Document

doc = Document()
doc.appendChild(doc.createElement("foo"))
print(doc.toprettyxml(indent="\t", encoding="UTF-8"))
# output: b'<?xml version="1.0" encoding="UTF-8"?>\n<foo/>\n'
```

wesleywright:
Fixes #10060.

This commit adds overloads for `xml.dom.minidom.Node.toxml` and `xml.dom.minidom.Node.toprettyxml` to return `str` when `encoding=None` and `bytes` otherwise. I also added return types when evaluating in Python < 3.9, which the stubs previously lacked; AFAICT the same return types should be valid there.

### DIFF
--- a/stdlib/xml/dom/minidom.pyi
+++ b/stdlib/xml/dom/minidom.pyi
@@ -34,33 +34,17 @@ class Node(xml.dom.Node):
         def toxml(self, encoding: str, standalone: bool | None = None) -> bytes: ...
         @overload
         def toxml(self, encoding: None = None, standalone: bool | None = None) -> str: ...
-        def toxml(self,
-            encoding: str | None = None,
-            standalone: bool | None = None,
-        ) -> str | bytes: ...
-
+        def toxml(self, encoding: str | None = None, standalone: bool | None = None) -> str | bytes: ...
         @overload
         def toprettyxml(
-            self,
-            indent: str = "\t",
-            newl: str = "\n",
-            encoding: str = ...,
-            standalone: bool | None = None,
+            self, indent: str = "\t", newl: str = "\n", encoding: str = ..., standalone: bool | None = None
         ) -> bytes: ...
         @overload
         def toprettyxml(
-            self,
-            indent: str = "\t",
-            newl: str = "\n",
-            encoding: None = None,
-            standalone: bool | None = None,
+            self, indent: str = "\t", newl: str = "\n", encoding: None = None, standalone: bool | None = None
         ) -> str: ...
         def toprettyxml(
-            self,
-            indent: str = "\t",
-            newl: str = "\n",
-            encoding: str | None = None,
-            standalone: bool | None = None,
+            self, indent: str = "\t", newl: str = "\n", encoding: str | None = None, standalone: bool | None = None
         ) -> str: ...
     else:
         @overload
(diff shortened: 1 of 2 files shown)


stdlib/xml/dom/minidom.pyi
```diff
@@ -30,13 +30,33 @@ class Node(xml.dom.Node):
...
+        def toprettyxml(
+            self, indent: str = "\t", newl: str = "\n", encoding: str = ..., standalone: bool | None = None
+        ) -> bytes: ...
+        @overload
+        def toprettyxml(
+            self, indent: str = "\t", newl: str = "\n", encoding: None = None, standalone: bool | None = None
+        ) -> str: ...
         def toprettyxml(
             self, indent: str = "\t", newl: str = "\n", encoding: str | None = None, standalone: bool | None = None
         ) -> str: ...
```
hmc-cs-mdrissi, line 48:
```suggestion
        ) -> str | bytes: ...
```

JelleZijlstra: In fact this should be deleted; stubs don't need an implementation, only the overloads.

stdlib/xml/dom/minidom.pyi
```diff
@@ -30,13 +30,33 @@ class Node(xml.dom.Node):
     def localName(self) -> str | None: ...
     def __bool__(self) -> Literal[True]: ...
     if sys.version_info >= (3, 9):
-        def toxml(self, encoding: str | None = None, standalone: bool | None = None) -> str: ...
+        @overload
+        def toxml(self, encoding: str, standalone: bool | None = None) -> bytes: ...
+        @overload
+        def toxml(self, encoding: None = None, standalone: bool | None = None) -> str: ...
+        def toxml(self, encoding: str | None = None, standalone: bool | None = None) -> str | bytes: ...
```
JelleZijlstra, line 37:
```suggestion
```

wesleywright:
After fixing the dangling implementation lines, I realized that my attempt at the stubs was too naïve due to how `encoding` can be passed to `toprettyxml()` as either a kwarg or a positional. I updated the stubs to account for that.

The "correct" way to stub this wasn't immediately obvious to me, so I wrote a small test case (`test_cases/check_xml.py`) to validate my stubs. I'm happy to delete those from the PR if the stubs are deemed to simple to be worth adding to test cases here.

test_cases/stdlib/check_xml.py
```diff
@@ -0,0 +1,35 @@
...
+# Because toprettyxml can mix positional and keyword variants of the "encoding" argument, which
+# determines the return type, the proper stub typing isn't immediately obvious. This is a basic
+# brute-force sanity check.
+# Test cases like toxml
+assert_type(document.toxml(), str)
+assert_type(document.toxml(encoding=None), str)
+assert_type(document.toxml(encoding="UTF8"), bytes)
+if sys.version_info >= (3, 9):
+    assert_type(document.toxml(standalone=True), str)
+    assert_type(document.toxml(encoding="UTF8", standalone=True), bytes)
```
AlexWaygood, line 28: Aren't these cases already covered by the cases on lines 9-16?

AlexWaygood (commented):
Thanks! This looks good to me. We hit this issue quite often in typeshed (where we need three overloads even though it _feels_ like we should be able to get away with two), but the test cases make it easier to verify that type checkers are making the correct inference here.

Just one question about some of the test cases -- looks to me like there might be some minor duplication right now?

test_cases/stdlib/check_xml.py
wesleywright: These are all meant to test toprettyxml; apparently I goofed that up at some point (probably when adding the version checks for 3.9). Will swap them to the correct method.

AlexWaygood (approved):
thanks!

PR #10062, opened 2023-04-18
Add type stub for whichcraft

thatfloflo:
Stubs for [whichcraft](https://pypi.org/project/whichcraft/).

Should be pretty straight-forward as the module simply provides a drop-in replacement for `shutil.which` on systems where that might not be available.


pyrightconfig.stricter.json
```diff
@@ -72,7 +72,8 @@
         "stubs/tqdm",
         "stubs/ttkthemes",
         "stubs/urllib3",
-        "stubs/vobject",
+        "stubs/vobject",,
+        "stubs/whichcraft"
```
Akuli, line 76: Our script has a bug that causes a syntax error when adding to the end of the script. It might be easy to fix, if you're interested :)

But that said, I think we don't need to add this to the stricter exclude list at all:

```suggestion
        "stubs/vobject",
```

stubs/whichcraft/whichcraft.pyi
```diff
@@ -0,0 +1,5 @@
+__author__: str
+__email__: str
+__version__: str
+
+def which(cmd: str, mode: int = ..., path: str | None = ...) -> str | None: ...
```
AlexWaygood, line 5: We now prefer to include default values where possible, but, again, our script to autogenerate stubs hasn't yet been updated to reflect our recent change in policy there :)

```suggestion
def which(cmd: str, mode: int = 1, path: str | None = None) -> str | None: ...
```

AlexWaygood (dismissed):
Thanks!

Akuli:
After looking at the [source code](https://github.com/cookiecutter/whichcraft/blob/master/whichcraft.py), I'm not sure if we should add this. It uses the standard-library `shutil.which()` when available. We only claim to support Python 3.7+ where `shutil.which()` is standard-library.

AlexWaygood:
> After looking at the [source code](https://github.com/cookiecutter/whichcraft/blob/master/whichcraft.py), I'm not sure if we should add this. It uses the standard-library shutil.which() when available. We only claim to support Python 3.7+ where shutil.which() is standard-library.

The `setup.py` file [states that the package supports Python 3.7](https://github.com/cookiecutter/whichcraft/blob/4318da2e2a4cb1ae3928446e5e98e3eb958a80f9/setup.py#L73), and in our `CONTRIBUTING.md` file, we state that we accept "stubs for any third-party packages" as long as "the package supports any Python version supported by typeshed":

https://github.com/python/typeshed/blob/0b17fafbe3a3a75ae11431ebdee72af5bf751a9c/CONTRIBUTING.md?plain=1#L126-L131

I agree that the package doesn't seem that useful if you're _only_ using Python 3.7+. But that's not how I read our policy.

Akuli:
I still can't think of a practical use case for the stubs. Type checkers no longer support Python 2, so if you care about type stubs in the first place, you must be using Python 3. And if you are using Python 3, there's no reason to use Python 3.1 or 3.2.

Is there an actual use case for these stubs that I just can't think of?

thatfloflo:
@Akuli yes, I think that's a fair objection in principle, as the module will always act as an alias to `shutil.which` on Python 3.7+ - so happy to go with you on this. The benefit I would see in including it is that it simplifies testing on codebases covering multiple versions of Python, potentially including those early ones where `from shutil import which` might fail. I agree there should be no reason for people to really support them anymore, but that doesn't mean they won't...

Akuli:
Do you know someone who needs a `which()` function in type-checked code that must also run on Python 3.2? :D

Akuli (approved):
I don't have a strong opinion about whether this should be merged or not. On the one hand it doesn't seem very useful, but on the other hand it would add very little maintenance effort.

AlexWaygood:
> I don't have a strong opinion about whether this should be merged or not. On the one hand it doesn't seem very useful, but on the other hand it would add very little maintenance effort.

basically my opinion too

thatfloflo:
No, to be fair any projects using it I'm currently aware of have dropped everything pre-3.6-ish. Even so, some still have whichcraft in the codebase (e.g. [Eel](https://pypi.org/project/Eel/ and various forks of it), so they have to do some config or `#type: ignore` stuff to avoid missing-import errors from mypy... (where really the answer is to just drop whichcraft imo).

I completely agree with you, the benefits are super tiny here, but so is the maintenance -- so personally happy either way.

AlexWaygood:
Hmmmm, if all we're doing is saving a few `type: ignore` comments for projects that really should just be migrating to `shutil.which()` anyway, then I guess I lean towards not adding these stubs :)

thatfloflo:
Okay, I've done a bit of poking around with code search across repos, which got around 500 results. A fair few of them seem to have now moved or be moving to direct shutil.which. While there are a few which are still "any Python 3 goes", none of the ones I looked at seem to be actually type checked in any obvious way. On second thought, the mypy error situation might also be a prompt for people to drop whichcraft if they don't really need it anymore. So following that, I'm with @Akuli and @AlexWaygood in that it's probably not worth actually including.

Seeing as there's nothing else going for it I'll close the PR...

PR #10070, opened 2023-04-21
Fix wrong connection argument name

kasium:
The name should be user and not username

### DIFF
--- a/stubs/hdbcli/hdbcli/dbapi.pyi
+++ b/stubs/hdbcli/hdbcli/dbapi.pyi
@@ -18,7 +18,7 @@ class Connection:
         self,
         address: str,
         port: int,
-        username: str,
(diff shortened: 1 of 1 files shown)


kasium:
Can someone please merge this?

JelleZijlstra (approved):
A bit hard to verify as this is apparently a closed-source C extension module and the docs don't cover the `Connection.__init__` interface, but `Connection(user=42)` complains about a wrong type while `Connection(username=42)`, so I think you're right.

kasium:
@JelleZijlstra thanks for checking. I found the issue while using hdbcli and yes, C doesn't make it easier :-)

JelleZijlstra:
I'm OK with C as long as I can actually see the source code :)

PR #10073, opened 2023-04-23
Add stubs for types.get_original_bases

Gobot1234:
ref: https://github.com/python/cpython#101827


stdlib/types.pyi
```diff
@@ -563,6 +566,9 @@ def prepare_class(
     name: str, bases: tuple[type, ...] = (), kwds: dict[str, Any] | None = None
 ) -> tuple[type, dict[str, Any], dict[str, Any]]: ...
 
+if sys.version_info >= (3, 12):
+    def get_orignal_bases(__cls: Any) -> tuple[Any, ...]: ...
```
JelleZijlstra, line 570:
```suggestion
    def get_original_bases(__cls: Any) -> tuple[Any, ...]: ...
```

Gobot1234: Oops, thanks

stdlib/types.pyi
```diff
@@ -563,6 +566,9 @@ def prepare_class(
     name: str, bases: tuple[type, ...] = (), kwds: dict[str, Any] | None = None
 ) -> tuple[type, dict[str, Any], dict[str, Any]]: ...
 
+if sys.version_info >= (3, 12):
+    def get_original_bases(__cls: Any) -> tuple[Any, ...]: ...
```
AlexWaygood, line 570: It needs to be passed a type, right?

```suggestion
    def get_original_bases(__cls: type) -> tuple[Any, ...]: ...
```

Gobot1234: I'm really not having a good night lol, yeah you're right

Gobot1234: Actually should that be type[Any] or does it not matter?

AlexWaygood: It shouldn't matter

Gobot1234: Oh yeah prepare_class doesn't subscript further

AlexWaygood (approved):
Thanks!

PR #10075, opened 2023-04-24
[dj-database-url] Fix CONN_MAX_AGE and conn_max_age types

lorddaedra:


sobolevn (approved):
It is true: https://docs.djangoproject.com/en/4.2/ref/settings/#std-setting-CONN_MAX_AGE

PR #10076, opened 2023-04-24
Add `cache_parameters` method to `_lru_cache_wrapper`

sobolevn:
The docs says that `cache_parameters()` function was added in 3.9: https://docs.python.org/3/library/functools.html#functools.lru_cache Source: https://github.com/python/cpython/blob/af530469954e8ad49f1e071ef31c844b9bfda414/Lib/functools.py#L512 But, `typeshed` does not have it.

Demo:

```python
>>> import functools
>>> functools.lru_cache(sum).cache_parameters()
{'maxsize': 128, 'typed': False}
```

### DIFF
--- a/stdlib/functools.pyi
+++ b/stdlib/functools.pyi
@@ -45,7 +45,7 @@ class _CacheInfo(NamedTuple):
     misses: int
     maxsize: int | None
     currsize: int
-        
+
 if sys.version_info >= (3, 9):
     class _CacheParameters(TypedDict):
         maxsize: int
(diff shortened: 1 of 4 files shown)


sobolevn:
Looks like `stubtest` needs a special-case for it.

stdlib/functools.pyi
```diff
@@ -46,12 +46,20 @@ class _CacheInfo(NamedTuple):
...
+        typed: bool
+
 @final
 class _lru_cache_wrapper(Generic[_T]):
     __wrapped__: Callable[..., _T]
     def __call__(self, *args: Hashable, **kwargs: Hashable) -> _T: ...
     def cache_info(self) -> _CacheInfo: ...
     def cache_clear(self) -> None: ...
+    if sys.version_info >= (3, 9):
+        def cache_parameters(self) -> _CacheParameters: ...
```
JelleZijlstra, line 61: Would be a good use case for inline TypedDict :)

JelleZijlstra (approved):

PR #5388, opened 2021-05-09
Add some missing logging features

spaghEddieDoyle:
This handles #5257 #5015 and changes a `type` to `Type[...]` for #343

PR #10077, opened 2023-04-24
Simplify `logging.config` type imports

sobolevn:
There's no need in `if sys.version` check here, all other places just use `typing_extensions` directly


Akuli (approved):
Apparently this was added in a PR that I reviewed: #5388

PR #10081, opened 2023-04-24
Loosen constraints on ujson file functions

tjstum:
Like `json.dump`, `ujson.dump` takes any object that has a string-accepting `.write` method.

Similarly, `ujson.load` works with any object with a `.read()` that returns either a string or a bytes


stubs/ujson/ujson.pyi
```diff
@@ -47,6 +47,6 @@ def dump(
 ) -> None: ...
 def decode(s: AnyStr, precise_float: bool = ...) -> Any: ...
 def loads(s: AnyStr, precise_float: bool = ...) -> Any: ...
-def load(fp: IO[AnyStr], precise_float: bool = ...) -> Any: ...
+def load(fp: SupportsRead[AnyStr], precise_float: bool = ...) -> Any: ...
```
AlexWaygood, line 50:
```suggestion
def load(fp: SupportsRead[str | bytes], precise_float: bool = ...) -> Any: ...
```

AlexWaygood (approved):
Thanks!

stubs/ujson/ujson.pyi
```diff
@@ -47,6 +47,6 @@ def dump(
 ) -> None: ...
 def decode(s: AnyStr, precise_float: bool = ...) -> Any: ...
```
tjstum, line 48: @AlexWaygood should I also remove these (since they aren't being used as a TypeVar)?

AlexWaygood: Yes, they look incorrect. I didn't want to ask you to change them in this PR since they are unrelated to the change you were making, but a followup PR fixing these would be welcome! :)

PR #10083, opened 2023-04-25
Fix paramiko client parameter types

xtrochu:
Use Mapping instead of dict for better covariance.

Right now, mypy rejects a `dict[str, list[str]]` for the `disabled_algorithms` parameter of `connect`, because of the invariance of the `dict` type.


AlexWaygood:
Thanks!

PR #10086, opened 2023-04-26
Convert the remaining ujson uses of AnyStr

tjstum:
Followup from #10081


tjstum:
@JelleZijlstra @AlexWaygood gentle ping! if there's anything else here you need from me, please let me know.

PR #10090, opened 2023-04-27
influxdb-client: fix stubtest

### LINKED ISSUE #10089: Daily tests failed on Thu Apr 27 2023

AlexWaygood:
Fixes #10089. The error was caused by the release of `urllib3==2.0.0`, which includes a `py.typed` file and inline type hints


hauntsaninja (approved):
Seems good! urllib3 is widely enough used that some people may need to pin an old version of types-influxdb-client, but that's fine

AlexWaygood:
Looks like we can't depend on `urllib3` rather than `types-urllib3` until our `types-urllib3` are actually removed, or we break pyright in CI :(

PR #10091, opened 2023-04-27
Add `object.__getstate__()` in Python 3.11+

christianbundy:
Background: Python 3.11 adds a default implementation of [`object.__getstate__()`](https://docs.python.org/3/library/pickle.html#object.__getstate__), which makes it available for every object.

Problem: This is currently missing from typeshed, which creates false positives in Mypy when calling the `__getstate__()` method on classes that don't explicitly define this method. 

Solution: Update `stdlib/builtins.pyi` to include this method in Python 3.11+.


christianbundy:
Thanks for the quick review!

PR #10096, opened 2023-04-28
Add stubs for cefpython3

demberto:
A few functions are documented but their docs say its not implemented yet. I have added the annotations for those functions but commented them out.


JelleZijlstra:
Apparently 66.1 only has Windows wheels: https://pypi.org/project/cefpython3/#files. Maybe we should provide stubs for 66.0 for now.

AlexWaygood:
The package doesn't support Python 3.10, and we run stubtest on 3.10 in CI.

I'd rather not start running stubtest on 3.9 in CI for all stubs.

It would probably be possible to have stubtest run on different Python versions for different stubs. But that would increase the complexity of our testing infrastructure.

The easiest thing to do for now would be to add `skip = true` to the `tool.stubtest` section of the `METADATA.toml` file for these stubs. But it's hardly a great solution.

AlexWaygood:
> The official package hasn't had an update or a communication from the author since 2021. However newer forked versions of it are available with wheels for 3.10.

(To be clear, it's absolutely fine to add stubs for a package that doesn't official support 3.10! It's just a matter of how exactly to make our CI happy with it 🙂)

JelleZijlstra:
Could you skip the stubs in stubtest for now as @AlexWaygood suggested?

stubs/cefpython3/cefpython3/cefpython.pyi
```diff
@@ -0,0 +1,741 @@
...
+KeyEventFlags: TypeAlias = int
+EVENTFLAG_NONE: int = ...
+EVENTFLAG_CAPS_LOCK_ON: int = ...
+EVENTFLAG_SHIFT_DOWN: int = ...
+EVENTFLAG_CONTROL_DOWN: int = ...
+EVENTFLAG_ALT_DOWN: int = ...
+EVENTFLAG_LEFT_MOUSE_BUTTON: int = ...
+EVENTFLAG_MIDDLE_MOUSE_BUTTON: int = ...
+EVENTFLAG_RIGHT_MOUSE_BUTTON: int = ...
+if sys.platform == "darwin":
```
JelleZijlstra, line 96: This check doesn't make much sense as the library only has Windows wheels.

stubs/cefpython3/cefpython3/cefpython.pyi
```diff
@@ -0,0 +1,741 @@
...
+WOD_SINGLETON_TAB: int = ...
+WOD_NEW_FOREGROUND_TAB: int = ...
+WOD_NEW_BACKGROUND_TAB: int = ...
+WOD_NEW_POPUP: int = ...
+WOD_NEW_WINDOW: int = ...
+WOD_SAVE_TO_DISK: int = ...
+WOD_OFF_THE_RECORD: int = ...
+WOD_IGNORE_ACTION: int = ...
+
+def CreateBrowserSync(
```
JelleZijlstra, line 142: This looks different from the arguments in https://github.com/cztomczak/cefpython/blob/5679f28cec18a57a56e298da2927aac8d8f83ad6/src/cefpython.pyx#L654

JelleZijlstra (commented):
Sorry for the long wait! It's difficult for me to review this, as the package only provides Windows wheels (I don't have a Windows system) and the package seems to have a complicated compilation setup. I glanced at a few things I could find but I'd prefer to leave the final review to someone who can check this on Windows.

JelleZijlstra:
Thanks for contributing! I'm closing this PR for now, because it still has unresolved feedback after over three months of inactivity. If you are still interested, please feel free to open a new PR (or ping us to reopen this one).

PR #10100, opened 2023-04-29
Mark `invoke` as obsolete since `2.1.2`

sobolevn:
New version of `invoke` contains type annotations and `py.typed` file: https://github.com/pyinvoke/invoke/tree/2.1.0/invoke

Released today (in my timezone): https://github.com/pyinvoke/invoke/releases/tag/2.1.0

We can stop updating stubs for it! 🎉 
Closes https://github.com/python/typeshed/pull/10098


AlexWaygood (commented):
Thanks! I think I'd rather wait for the automated stubsabot PR, though, in case there's a packaging bug that means that the `py.typed` file isn't included in wheels published on PyPI. This seems to happen fairly often.

stubs/invoke/METADATA.toml
```diff
@@ -1,5 +1,5 @@
 version = "2.0.*"
-obsolete_since = "2.1.0" # Released on 2023-04-29
+obsolete_since = "2.1.2" # Released on 2023-05-16
```
AlexWaygood, line 2: According to https://pypi.org/project/invoke, it was released on 2023-05-15. Stubsabot agrees in #10098 (it's auto-updated the PR to suggest marking the stubs as obsolete, now that the py.typed file is included in the wheel uploaded to PyPI)

sobolevn: timezone difference :)

PR #10109, opened 2023-04-29
paramiko: Improve various bytes-related types

### LINKED ISSUE #5736: paramiko stubs seem incorrect for python3

JelleZijlstra:
Part of #9006

Fixes #5736


srittau (approved):
I did not confirm the types against runtime types, but they all look plausible.

PR #10110, opened 2023-04-29
Restore error output in runtests.py

Avasam:
CC @AlexWaygood 
This was changed in https://github.com/python/typeshed/pull/10002#discussion_r1157443671, but results in the following:
image


AlexWaygood (approved):
Oops -- my bad, thanks for fixing! 🤦

PR #10114, opened 2023-04-30
Mark Pillow as complete for stubtest

Avasam:
Run stubgen + stubdefaulter on Pillow and fix issues.
A handful of int literals were easier to update by simply copying back the list from source.
Add platform-specific checks. (some were needed for stubtest to pass on both win32 and linux, so I did them all)


srittau (approved):
Thanks, spot checking looks good.

PR #10118, opened 2023-04-30
move some stuffs related to `_SimpleCData` from `ctypes/__init__.pyi` to `_ctypes.pyi`, as at runtime.

junkmd:
related to #8633 and https://github.com/python/typeshed/issues/8571#issuecomment-1221316973

#10111 and google/pytype#1411 (commit; https://github.com/google/pytype/commit/2e2088fcae3767802fe60f419499090db59f6c4f) have enabled us to move definitions from `__init__.pyi` to `_ctypes.pyi`.(https://github.com/python/typeshed/issues/8633#issuecomment-1528833567)
There are no longer causes CI to fail
So I'm going to contribute for the first time in several months.

I have moved the following stuffs.
- `_CDataMeta`
- `_CData`
- `_SimpleCData`
- `Array`

And I fixed `stubtest_allowlist` for `_ctypes.Array.`

### DIFF
--- a/stdlib/_ctypes.pyi
+++ b/stdlib/_ctypes.pyi
@@ -2,7 +2,7 @@ import sys
 from _typeshed import ReadableBuffer, WriteableBuffer
 from abc import abstractmethod
 from collections.abc import Iterable, Iterator, Mapping
-from ctypes import _CArgObject, _PointerLike, CDLL
+from ctypes import CDLL, _CArgObject, _PointerLike
 from typing import Any, Generic, TypeVar, overload
 from typing_extensions import Self, TypeAlias
 
--- a/stdlib/_ctypes.pyi
+++ b/stdlib/_ctypes.pyi
@@ -1,6 +1,16 @@
 import sys
-from ctypes import _CArgObject, _PointerLike
-from typing_extensions import TypeAlias
+from _typeshed import ReadableBuffer, WriteableBuffer
+from abc import abstractmethod
+from collections.abc import Iterable, Iterator, Mapping
+from ctypes import _CArgObject, _PointerLike, CDLL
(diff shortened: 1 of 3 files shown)


junkmd:
Thank you for the correction in https://github.com/python/typeshed/pull/10118/commits/727ceb8b714208cd5a3c94fc3877fb3a9956cceb.

AlexWaygood (approved):
Thank you!

AlexWaygood:
> 
> ```diff
> psycopg (https://github.com/psycopg/psycopg)
> + psycopg/psycopg/pq/pq_ctypes.py:1026: error: Returning Any from function declared to return "bytes"  [no-any-return]
> 
> discord.py (https://github.com/Rapptz/discord.py)
> - discord/opus.py:377: error: Value of type variable "_T" of "array" cannot be "bytes"  [type-var]
> ```

It looks like this interferes slightly with mypy's plugin for typechecking `ctypes` code. There's a bunch of places in the mypy codebase where they hardcode `if fullname == "ctypes.Array"` etc.:

https://github.com/python/mypy/blob/6b1fc865902bf2b845d3c58b6b9973b5a412241f/mypy/plugins/default.py#L41-L48

It should be fairly easy to fix; I can make a PR to mypy.

junkmd:
> It should be fairly easy to fix; I can make a PR to mypy.

Excellent. 👍 

It would be very helpful to have you do this.

AlexWaygood:
Hmm, actually, it might not be possible to fix the mypy plugin unless it's done after mypy syncs typeshed.

@hauntsaninja, do you reckon it's okay to merge this for now, and fix mypy after the next typeshed sync?

hauntsaninja:
Yeah, that sounds good!

PR #10119, opened 2023-04-30
Run mypy and pyright on our py312 stubs in CI

AlexWaygood:
Depends on #10117


tests/mypy_test.py
```diff
@@ -76,6 +76,13 @@ def valid_path(cmd_arg: str) -> Path:
     return path
 
 
+def remove_dev_suffix(version: str) -> str:
+    """Helper function for argument-parsing"""
+    if version.endswith("-dev"):
```
JelleZijlstra, line 81: `.removesuffix`? Or do we run this on overly old versions?

stdlib/configparser.pyi
```diff
@@ -29,6 +29,9 @@ __all__ = [
     "MAX_INTERPOLATION_DEPTH",
 ]
 
+if sys.version_info >= (3, 12):
+    __all__.remove("SafeConfigParser")
```
JelleZijlstra, line 33: Is that supported? I think we should instead .append it if we're below 3.12.

tests/mypy_test.py
AlexWaygood: We run this on py37, yeah :)

I can add a comment if you like? The reason why we run this on py37 is because if we run mypy with the `--python-version=3.7` argument _on py311_ (or whatever), pip will install a version of numpy that's incompatible with python 3.7 (because the latest version of numpy is incompatible with 3.7). This then causes mypy to complain about invalid syntax on some of the types it finds in `site-packages/` when it's run with the `--python-version=3.7` flag.

JelleZijlstra: Thanks! Let's just leave this as is then.

stdlib/configparser.pyi
AlexWaygood: I thought it was, but can't immediately find the reference, and I guess it's better to be safe than sorry :)

JelleZijlstra: At least typeshed-client only supports `+=`, and that's the only operation used in any existing stub in typeshed.

AlexWaygood: Looks like pyright definitely supports it: https://github.com/microsoft/pyright/blob/main/docs/typed-libraries.md#library-interface

But whereas I can see code in `mypy/semanal.py` that specifically deals with `__all__.append` and `__all__.extend`...
https://github.com/python/mypy/blob/6b1fc865902bf2b845d3c58b6b9973b5a412241f/mypy/semanal.py#L4881-L4895

...I can't find any equivalent handling of `__all__.remove` in mypy.

AlexWaygood: We should probably add a lint to flake8-pyi so that it forbids operations on `__all__` unless they're known to be supported by all major type checkers.

JelleZijlstra: Looking at the mypy code they also support `.append` and `.extend` but not `.remove`.

JelleZijlstra (approved):

stdlib/configparser.pyi
AlexWaygood: Opened:

- https://github.com/PyCQA/flake8-pyi/issues/369

PR #10120, opened 2023-04-30
Don't install `types-*` dependencies in pyright tests

### LINKED ISSUE #10112: pyright gives types found in `site-packages` higher priority than types found in typeshed
See this comment for context and example: https://github.com/python/typeshed/pull/10058#issuecomment-1528831457

I think this only affects `pyyaml` and `setuptools` at the moment, as they are they only `type-*` packages installed during tests.

Can probably be fixed by specifying the `--typeshed-path` cli argument. Note that this would have to be done in `pyright_test.py` and all Github actions. I have not tested this potential solution myself yet, and I didn't feel like doing this right now, hence this issue to remember.

Avasam:
One idea to fix #10112 .

This idea could probably be extended to pytype and mypy stubs type-checking tests to reduce dependencies installs. (rename `requirements-pyright.txt` to `requirements-stubs-typecheck.txt` or something)

Or I can go the route of using `grep` in CI to only include the dependencies needed to run `get_external_stub_requirements.py`, whilst keeping the version requirement.

Or I can try to use grep to filter out `types-*` dependencies.


AlexWaygood:
How about https://github.com/python/typeshed/pull/10121, as an alternative fix? Feels slightly cleaner imo

AlexWaygood:
Superseded by #10121

PR #10121, opened 2023-04-30
Alternative fix for pyright CI issues

AlexWaygood:
Helps with https://github.com/python/typeshed/issues/10112 (an alternative fix to #10120). Install 3rd-party stub dependencies into an isolated environment, and use that for running pyright on typeshed, rather than installing the 3rd-party stub requirements into the same environment all our test requirements are installed in.

As with #10120, this doesn't fix the issue encountered in #9989 or #10090, but it does unblock #10058.

### DIFF
--- a/pyrightconfig.json
+++ b/pyrightconfig.json
@@ -1,6 +1,7 @@
 {
     "$schema": "https://raw.githubusercontent.com/microsoft/pyright/main/packages/vscode-pyright/schemas/pyrightconfig.schema.json",
     "typeshedPath": ".",
+    "venv": ".venv",
     "include": [
         "stdlib",
         "stubs",
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -125,15 +125,23 @@ jobs:
           python-version: "3.10"
           cache: pip
           cache-dependency-path: requirements-tests.txt
-      - name: Install external dependencies for 3rd-party stubs
+      - name: Install typeshed test-suite requirements
+        # Install these so we can run `get_external_stub_requirements.py`
+        run: pip install -r requirements-tests.txt
+      - name: Create an isolated venv for testing
+        run: python -m venv .venv
+      - name: Install 3rd-party stub dependencies
         run: |
-          pip install -r requirements-tests.txt
           DEPENDENCIES=$(python tests/get_external_stub_requirements.py)
           if [ -n "$DEPENDENCIES" ]; then
+            source .venv/bin/activate
             echo "Installing packages: $DEPENDENCIES"
             pip install $DEPENDENCIES
           fi
-      - run: pip freeze --all
+      - name: List 3rd-party stub dependencies installed
+        run: |
+          source .venv/bin/activate
+          pip freeze --all
       - name: Get pyright version
         uses: SebRollen/toml-action@v1.0.2
         id: pyright_version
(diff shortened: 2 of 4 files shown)


Avasam:
Oh yeah that works too. Didn't think of pointing pyright to a venv. It does feel cleaner than my suggestions.

For #9989 you *could* (if you wanted) uninstall SQLAlchemy 2 from the venv. Just like we do for setuptools/distutils.
Even if you don't want to do that, it shows this solution is more flexible.

AlexWaygood:
> For #9989 you _could_ (if you wanted) uninstall SQLAlchemy 2 from the venv.

Ooh, interesting idea! Feels more like a _workaround_ than a _fix_ though -- the fundamental issue is that we should probably be testing each third-party stubs package in isolation, like we do with mypy, and it wouldn't do anything to fix that. For me, it's just a question of whether we can tolerate the extra complexity that doing that would entail (I'm not sure I can! I really like how simple running pyright via a GitHub Action is!).

pyrightconfig.json
```diff
@@ -1,6 +1,7 @@
 {
     "$schema": "https://raw.githubusercontent.com/microsoft/pyright/main/packages/vscode-pyright/schemas/pyrightconfig.schema.json",
     "typeshedPath": ".",
+    "venv": ".venv",
```
Akuli, line 4: Does this mean that it isn't going to work for people who prefer to name their venv something else? I usually name it `env`. It might not be a problem for me, because I don't run pyright locally very often.

AlexWaygood: Yes, I think it will mean that. FWIW, I also prefer `env` as a name for my local virtual environments (it's fewer characters to type), but went with `.venv` here since it's the name recommended by our CONTRIBUTING docs:

https://github.com/python/typeshed/blob/ae0c9f9dadd72ee823d289416943a9109e24fe52/CONTRIBUTING.md?plain=1#L59

Avasam: `.venv` is also what's configured for flake8, gitignore, and isort.
mypy_test and regr_test use a temp directory anyway, but it's also using `.venv`

So it's pretty established that the standard for typeshed is to name the virtual environment folder `.venv`

Akuli: I would still prefer passing it in a command-line option instead of hard-coding it in the config.

AlexWaygood: > I would still prefer passing it in a command-line option instead of hard-coding it in the config.

So would I, but if I'm reading the docs for pyright correctly, I don't think that's an option

Akuli: Apparently pyright has a `-v, --venv-path` option whose purpose is exactly this. From [the docs](https://github.com/microsoft/pyright/blob/main/docs/command-line.md):

> (3) This option is used in conjunction with configuration file, ... This allows a common config file to be checked in to the project and shared by everyone on the development team without making assumptions about the local paths to the venv directory on each developer’s computer.

AlexWaygood: The `venv-path` option can be set via the command line (and I do in this PR, via pyright-action). But the `venv-path` option is meant to point to a directory in which there might be multiple different virtual environments, if I'm reading the pyright docs correctly, meaning it's useless to set the `venv-path` option on the command line unless you also set the `venv` option in the config file

Akuli (approved):

PR #10122, opened 2023-05-01
Add core tensorflow.data stubs

hmc-cs-mdrissi:
# Summary
Add stubs for [tf.data.Dataset](https://www.tensorflow.org/api_docs/python/tf/data/Dataset), the main api used in tf.data. Includes a few other tf.data stubs/dependent types but mostly focused on that class.

### DIFF
--- a/stubs/tensorflow/tensorflow/__init__.pyi
+++ b/stubs/tensorflow/tensorflow/__init__.pyi
@@ -1,16 +1,26 @@
 from _typeshed import Incomplete, Unused
-from abc import ABCMeta
+from abc import ABC, ABCMeta, abstractmethod
 from builtins import bool as _bool
 from collections.abc import Callable, Generator, Iterable, Iterator, Mapping, Sequence
 from contextlib import contextmanager
 from enum import Enum
 from types import TracebackType
-from typing import Any, NoReturn, TypeVar, overload
+from typing import Any, Generic, NoReturn, TypeVar, overload
 from typing_extensions import ParamSpec, Self, TypeAlias
 
 import numpy
-from tensorflow import feature_column as feature_column, initializers as initializers, io as io, keras as keras, math as math
+from google.protobuf.message import Message
+from tensorflow import (
+    data as data,
+    experimental as experimental,
+    feature_column as feature_column,
+    initializers as initializers,
+    io as io,
+    keras as keras,
+    math as math,
+)
 from tensorflow._aliases import ContainerGradients, ContainerTensors, ContainerTensorsLike, Gradients, TensorLike
+from tensorflow.core.protobuf import struct_pb2
 
 # Explicit import of DType is covered by the wildcard, but
 # is necessary to avoid a crash in pytype.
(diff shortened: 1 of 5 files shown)


stubs/tensorflow/tensorflow/data/__init__.pyi
```diff
@@ -0,0 +1,210 @@
...
+    def skip(self, count: _ScalarTensorCompatible, name: str | None = None) -> Dataset[_T1]: ...
+    def take(self, count: _ScalarTensorCompatible, name: str | None = None) -> Dataset[_T1]: ...
+    def unbatch(self, name: str | None = None) -> Dataset[_T1]: ...
+    def unique(self, name: str | None = None) -> Dataset[_T1]: ...
+    def with_options(self, options: Options, name: str | None = None) -> Dataset[_T1]: ...
+    @staticmethod
+    def zip(datasets: tuple[Dataset[_T2], Dataset[_T3]], name: str | None = None) -> Dataset[tuple[_T2, _T3]]: ...
+    def __len__(self) -> int: ...
+    def __nonzero__(self) -> bool: ...
+    # Missing ones are ragged_batch, random, rejection_resample, sample_from_datasets,
```
hmc-cs-mdrissi, line 181: I'll try to fill rest in next day or two to have this class be type complete.

hmc-cs-mdrissi:
Hmm, the pytype crash looks to be caused by importing a protobuf generated stub in tensorflow stubs. I don't see a good way to avoid that as a lot of tensorflow types will depend on protobuf generated types. Any ideas or should I just add all tensorflow files to pytype exclusion? If I need to exclude all files is there a way to exclude an entire stub directory/can I add a constant,

PYTYPE_EXCLUDED_STUBS: set[str] = {"tensorflow"}

hmc-cs-mdrissi:
The shape_poly_test.py hit is correct. They have code that simplified looks like this,

```python
shapes = foo.output_shapes # TensorShape | list[TensorShape]
specs = foo.output_signature # TypeSpec | list[TypeSpec]
if not isinstance(shapes, list):
  shapes = [shapes
  specs = [specs]

zip(shapes, specs)
```

Shapes/specs are related in that if one is list the other will be too, but there's no way a type checker can know that relationship/narrow a different variable because of it. They just need to add an extra assert.

stubs/tensorflow/tensorflow/__init__.pyi
```diff
@@ -1,16 +1,26 @@
...
+    data as data,
+    experimental as experimental,
+    feature_column as feature_column,
+    initializers as initializers,
+    io as io,
+    keras as keras,
+    math as math,
+)
 from tensorflow._aliases import ContainerGradients, ContainerTensors, ContainerTensorsLike, Gradients, TensorLike
+from tensorflow.core.protobuf import struct_pb2
```
hmc-cs-mdrissi, line 23: This is the line that's leading to pytype crash (from inspecting protobuf types).

JelleZijlstra:
> Hmm, the pytype crash looks to be caused by importing a protobuf generated stub in tensorflow stubs. I don't see a good way to avoid that as a lot of tensorflow types will depend on protobuf generated types. Any ideas or should I just add all tensorflow files to pytype exclusion? If I need to exclude all files is there a way to exclude an entire stub directory/can I add a constant,
> 
> `PYTYPE_EXCLUDED_STUBS: set[str] = {"tensorflow"}` and adjust pytype_test accordingly?

cc @rchen152, do you know if this is easy to fix in pytype? The errors are in https://github.com/python/typeshed/actions/runs/4849039988/jobs/8640678344?pr=10122 (there's a lot of them).

rchen152:
> cc @rchen152, do you know if this is easy to fix in pytype? The errors are in https://github.com/python/typeshed/actions/runs/4849039988/jobs/8640678344?pr=10122 (there's a lot of them).

The errors about "typing.NotRequired" being missing should be easy to fix. The protobuf-related ones I'll look into in more detail on Thursday. I'm guessing that we'll end up having to exclude some files from pytype analysis as @hmc-cs-mdrissi suggests, but it would be nice if we could exclude only the problematic protobuf files and tell pytype to treat them as containing just `def __getattr__(name) -> Any: ...` when imported.

rchen152:
Ok, the immediate problem is that when a stub not in pytype_exclude_list imports one that is in pytype_exclude_list, then the latter gets loaded and - of course - generates a bunch of errors. This feels like it should be fixable by feeding the exclude list into pytype's internal list of missing modules here: https://github.com/python/typeshed/blob/6c54c74a84642faf8ed9834bd1adbc10c1877326/tests/pytype_test.py#L73. I'll try to get that working next week (I'm out tomorrow).

Somewhat orthogonal to that, the reason that pytype is failing on all of these auto-generated protobuf stubs is some buggy alias lookup logic. I have a very messy change that seems to more or less fix it, so longer term, I'll try to clean that up and get it submitted, which would let us take these stubs off the exclude list.

rchen152:
https://github.com/python/typeshed/pull/10187 was just merged and should fix the pytype errors in this PR.

JelleZijlstra:
Thanks Rebecca!

hmc-cs-mdrissi:
Thank you Rebecca.

I think this pr is complete now and ready for review.

stubs/tensorflow/tensorflow/experimental/__init__.pyi
```diff
@@ -0,0 +1,10 @@
+from _typeshed import Incomplete
+from abc import ABC
+from typing import Generic, TypeVar
+
+_T1 = TypeVar("_T1", covariant=True)
+
+class Optional(Generic[_T1], ABC):
```
JelleZijlstra, line 7:
```suggestion
_T_co = TypeVar("_T_co", covariant=True)

class Optional(Generic[_T_co], ABC):
```
Conventional name

PR #10123, opened 2023-05-01
Remove unnecessary F821 noqas

AlexWaygood:
We currently have F821 errors suppressed across the repo:

https://github.com/python/typeshed/blob/ae0c9f9dadd72ee823d289416943a9109e24fe52/.flake8#L37-L46

We might switch it back on if https://github.com/PyCQA/flake8-pyi/pull/364 is merged, but if it is merged, flake8 won't emit F821 on these lines anyway.


AlexWaygood:
This deals with all the `typeshed_primer` hits in https://github.com/PyCQA/flake8-pyi/pull/364#issuecomment-1529133908 except for the ones in generated protobuf stubs (which is most of them).

PR #10127, opened 2023-05-01
Add `stubdefaulter` to `create_baseline_stubs.py`

### LINKED ISSUE #10126: `create_baseline_stubs.py` script does not add default values
As title says. Now that typeshed prefers having default values, it would make sense if the `create_baseline_stubs.py` script added them.
Maybe that would be stubgen's job?
For now I'm running https://github.com/JelleZijlstra/stubdefaulter manually as I am moving stubs out of https://github.com/microsoft/python-type-stubs, but I don't think that's even documented/mentioned in the contributing guidelines.

Avasam:
Closes #10126

### DIFF
--- a/scripts/create_baseline_stubs.py
+++ b/scripts/create_baseline_stubs.py
@@ -158,7 +158,7 @@ def main() -> None:
 
     run_stubgen(package, stub_dir)
     run_stubdefaulter(stub_dir)
-    
+
     run_isort(stub_dir)
     run_black(stub_dir)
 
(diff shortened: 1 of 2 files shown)


scripts/create_baseline_stubs.py
```diff
@@ -153,6 +158,7 @@ def main() -> None:
 
     run_stubgen(package, stub_dir)
 
+    run_stubdefaulter(stub_dir)
```
Avasam, line 161:
```suggestion
    run_stubgen(package, stub_dir)
    run_stubdefaulter(stub_dir)
    
```

AlexWaygood (approved):
Thanks! LGTM

Issue #10093, opened 2023-04-27
sys.stdin, sys.stdout, sys.stderr are of type io.TextIOWrapper


bdrung:
Example code:
```python
#!/usr/bin/python3
import sys
print(type(sys.stdin))
print(type(sys.stdout))
print(sys.stdin.detach())
```
This results in following output:
```
$ ./stdin.py 
<class '_io.TextIOWrapper'>
<class '_io.TextIOWrapper'>
<_io.BufferedReader name='<stdin>'>
```
But mypy complains:
```
$ mypy stdin.py 
stdin.py:5: error: "TextIO" has no attribute "detach"  [attr-defined]
Found 1 error in 1 file (checked 1 source file)
```

bdrung:
Apport uses `sys.stdout.detach()` and it also redefines it with `sys.stdin = os.fdopen(int(fds[0]), "r")` at some point.

Akuli:
While it is common to redefine `sys.stdout = io.StringIO()` or `sys.stderr = io.StringIO()`, maybe doing that isn't so common for stdin and we could have a more specific `sys.stdin` type.

TeamSpen210:
Since it is expected that libraries could redefine them, perhaps this error is good since it'll force you to use `isinstance()` or `hasattr()`, and handle the case that `detach()` isn't present?

AlexWaygood:
We previously discussed this in https://github.com/python/typeshed/pull/5623.

I guess I'd be open to having a more specific type for `sys.stdin`, as @Akuli suggests. But it does seem a bit weird to have a different type for `sys.stdin` than the one we use for the other two. That feels like it might cause yet more user confusion.

Whatever we do here, we should probably add a comment to the stub, at the very least. It's hard to properly document the decisions we make at typeshed, but we can do better than what we have at the moment:

https://github.com/python/typeshed/blob/6f6b1c9135edbe7e6b16b06070f3b12c7bee3a5a/stdlib/sys.pyi#L60-L62

PR #10135, opened 2023-05-03
Document sys.std* typing

srittau:
Closes: #10093

### DIFF
--- a/stdlib/sys.pyi
+++ b/stdlib/sys.pyi
@@ -59,7 +59,7 @@ ps1: object
 ps2: object
 
 # TextIO is used instead of more specific types for the standard streams,
-# since they can be monkeypatched at runtime. At startup, the objects
+# since they are often monkeypatched at runtime. At startup, the objects
 # are initialized to instances of TextIOWrapper.
 #
 # To use methods from TextIOWrapper, use an isinstance check to ensure that
--- a/stdlib/sys.pyi
+++ b/stdlib/sys.pyi
@@ -57,9 +57,18 @@ if sys.version_info >= (3, 8):
     pycache_prefix: str | None
 ps1: object
 ps2: object
+
+# TextIO is used instead of more specific types for the following fields,
+# since they can be overridden at runtime. After startup, the fields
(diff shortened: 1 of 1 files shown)


stdlib/sys.pyi
```diff
@@ -57,9 +57,18 @@ if sys.version_info >= (3, 8):
     pycache_prefix: str | None
 ps1: object
 ps2: object
+
+# TextIO is used instead of more specific types for the following fields,
+# since they can be overridden at runtime. After startup, the fields
+# are set to TextIOWrapper. To use methods from TextIOWrapper, use an
+# isinstance check to ensure that the streams have not been overridden:
```
AlexWaygood, line 64:
```suggestion
# since they are often monkeypatched by 3rd-party libraries at runtime.
# At startup, these variables are set to instances of TextIOWrapper.
# To use methods from TextIOWrapper,
# use an isinstance() check to ensure that the streams have not been overridden:
```

stdlib/sys.pyi
```diff
@@ -57,9 +57,18 @@ if sys.version_info >= (3, 8):
     pycache_prefix: str | None
 ps1: object
 ps2: object
+
+# TextIO is used instead of more specific types for the following fields,
```
JelleZijlstra, line 61:
```suggestion
# TextIO is used instead of more specific types for the following objects,
```
I'd use "fields" only within classes.

stdlib/sys.pyi
```diff
@@ -57,9 +57,20 @@ if sys.version_info >= (3, 8):
     pycache_prefix: str | None
 ps1: object
 ps2: object
+
+# TextIO is used instead of more specific types for the standard streams,
+# since they can be monkeypatched at runtime. At startup, the objects
```
AlexWaygood, line 62: I would say "are often" rather than "can be", since any variable _could_ be monkeypatched at runtime, but for these variables, it's a common idiom that's sort-of officially blessed. (If it weren't, there wouldn't be any need for `sys.__stdin__`, `sys.__stdout__` and `sys.__stderr__`.)

```suggestion
# since they are often monkeypatched at runtime. At startup, the objects
```

Issue #10132, opened 2023-05-03
`ctypes._FuncPointer` doesn't exist at runtime and `_ctypes.CFuncPtr` doesn't exist in typestub


junkmd:
Quoting from https://github.com/python/typeshed/pull/9251#issuecomment-1324551303;

> * Our stub has a `_FuncPointer` class that doesn't exist at runtime. But in fact, the code returns a subclass of `_ctypes.CFuncPtr`, so why don't we use that class in the stubs?

In addition, Python's `ctypes` runtime namespace is defined by the following lines;

- `from _ctypes import CFuncPtr as _CFuncPtr` importing part
  https://github.com/python/cpython/blob/872cbc613245db7a1fc5e6656ed0135d2e115f50/Lib/ctypes/__init__.py#L10

  - Subclasses of `_ctypes.CFuncPtr` is created by the following lines

    - Definition of `CFUNCTYPE`, factory function that creates dynamically `CFunctionType` class
      https://github.com/python/cpython/blob/872cbc613245db7a1fc5e6656ed0135d2e115f50/Lib/ctypes/__init__.py#L73-L107

    - Definition of `PYFUNCTYPE`, factory function that creates dynamically `CFunctionType` class
      https://github.com/python/cpython/blob/872cbc613245db7a1fc5e6656ed0135d2e115f50/Lib/ctypes/__init__.py#L509-L514

    - Definition of `WINFUNCTYPE`, factory function that creates dynamically `WinFunctionType` class
      https://github.com/python/cpython/blob/872cbc613245db7a1fc5e6656ed0135d2e115f50/Lib/ctypes/__init__.py#L114-L134

    - Definition of `CDLL`, its' `__getitem__` and `__getattr__` return `_FuncPtr` class
      https://github.com/python/cpython/blob/872cbc613245db7a1fc5e6656ed0135d2e115f50/Lib/ctypes/__init__.py#L322-L400

---

I think the following way can be improved with less changes and less impact. 

1. move the definition of `_FuncPointer` to `_ctypes.pyi`.
2. change the class name to `CFuncPtr`.
3. add `from _ctypes import CFuncPtr as _CFuncPtr` line to `ctypes/__init__.pyi`.
4. add `class _FuncPointer(_CFuncPtr): ...` line.

The `ctypes._FuncPointer` will be conceptualized as "a dynamically generated subclass of `_ctypes.CFuncPtr`".

---

[...]

PR #10140, opened 2023-05-05
move and rename from `ctypes._FuncPointer` to `_ctypes.CFuncPtr`

junkmd:
- Please see #10132


junkmd:
Stubtests are failed.

Quoting from https://github.com/python/typeshed/actions/runs/4889040813/jobs/8727237786?pr=10140#step:5:19,

> error: _ctypes.CFuncPtr cannot be subclassed at runtime, but isn't marked with @final in the stub

However, the return value of `CFUNCTYPE` is a subclass of `_ctypes.CFuncPtr` at runtime.
https://github.com/python/cpython/blob/872cbc613245db7a1fc5e6656ed0135d2e115f50/Lib/ctypes/__init__.py#L73-L107

I wonder why it occurs.

JelleZijlstra:
Stubtest probably just tries to do something like:

```
>>> class X(ctypes._CFuncPtr): pass
... 
TypeError: class must define _flags_ which must be an integer
```

It can only be subclassed if the subclass sets fields like `_flags_`. I think we can simply add it to the allowlist (`tests/stubtest_allowlists/py3_common.txt` probably).

junkmd:
Both adding `_ctypes.CFuncPtr._flags_` to allowlists(6399ecd) and adding the `_flags_` attribute to `_ctypes.CFuncPtr`(caedbda) doesn't resolve this problem 😭.

AlexWaygood:
> Both adding `_ctypes.CFuncPtr._flags_` to allowlists([6399ecd](https://github.com/python/typeshed/commit/6399ecd425efade47a6c73ba830984f58ccad094)) and adding the `_flags_` attribute to `_ctypes.CFuncPtr`([caedbda](https://github.com/python/typeshed/commit/caedbdacd082071fda86cf1fa0e2d0d6ac7ae31c)) doesn't resolve this problem 😭.

You needed to add `_ctypes.CFuncPtr` to the allowlist rather than _ctypes.CFuncPtr._flags_` — I've corrected it for you :)

AlexWaygood (approved):

junkmd:
@AlexWaygood 
Thank you for the correction.

JelleZijlstra (approved):

PR #10151, opened 2023-05-06
Redis: Add missing exported names

mjpieters:
- `redis.asyncio.RedisCluster` / `redis.asyncio.CommandsParser` [added in 4.3.0](https://github.com/redis/redis-py/releases/tag/v4.3.0)
- `redis.default_backoff` / `redis.asyncio.default_backoff`, [added in 4.4.0](https://github.com/redis/redis-py/releases/tag/v4.4.0rc4)


AlexWaygood:
(Nit: we prefer to avoid force pushes where possible at typeshed, as it makes it hard to see what changed between commits, and doesn't play very well with the GitHub UI. All PRs are squash-merged at typeshed, so there's no need to worry about a clean git history on the PR :)

mjpieters:
> (Nit: we prefer to avoid force pushes where possible at typeshed, as it makes it hard to see what changed between commits, and doesn't play very well with the GitHub UI. All PRs are squash-merged at typeshed, so there's no need to worry about a clean git history on the PR :)

Ah, I'll keep that in mind!

Yes, I normally aim to provide a clean commit series, as a small mistake like forgetting to put `default_backoff` in the `__all__` list is not something a reviewer would need to care about. 😄 So, my motivation was to make it easier for reviewers, not about what gets merged. I'll try to not use force pushes here in future.

mjpieters:
The failing stubtest is already reported as #9127, and I _think_ now depends on a new release of Redis with redis/redis-py#2729 included?

AlexWaygood:
> The failing stubtest is already reported as #9127, and I _think_ now depends on a new release of Redis with [redis/redis-py#2729](https://github.com/redis/redis-py/pull/2729) included?

Yes, I think it would be fine just to add an allowlist entry for now!

mjpieters:
> Yes, I think it would be fine just to add an allowlist entry for now!

I saw the original `redis.asyncio.cluster.RedisCluster` location was already in that list, only the new export needed adding.

AlexWaygood (approved):
Thanks!

PR #10157, opened 2023-05-08
Allow marking stubs as partial

Avasam:
This is required for https://github.com/typeshed-internal/stub_uploader/pull/94, please read discussion over there for all the details. Both should be approved before merging.

I've updated the docs, but wording could be improved I'm sure.


CONTRIBUTING.md
```diff
@@ -191,6 +191,11 @@ supported:
 * `upload` (optional): This field is set to `false` to prevent automatic
   uploads to PyPI. This should only used in special cases, e.g. when the stubs
   break the upload.
+* `partial_stub` (optional): This field is used to let stub_uploader know whether
+  to distribute a type stub package as
+  [partial](https://peps.python.org/pep-0561/#partial-stub-packages). This is for
+  3rd-party stubs that don't cover the entirerity of the package's public API.
```
Avasam, line 197:
```suggestion
  3rd-party stubs that don't cover the entirety of the package's public API.
```

sobolevn (commented):
I agree that splitting these two things makes sense, because other use-case where stubs can be partial is when `allowlist` contains a lot of entries.

CONTRIBUTING.md
```diff
@@ -191,6 +191,11 @@ supported:
 * `upload` (optional): This field is set to `false` to prevent automatic
   uploads to PyPI. This should only used in special cases, e.g. when the stubs
   break the upload.
+* `partial_stub` (optional): This field is used to let stub_uploader know whether
+  to distribute a type stub package as
+  [partial](https://peps.python.org/pep-0561/#partial-stub-packages). This is for
+  3rd-party stubs that don't cover the entirety of the package's public API.
+  In most cases, this field follows `stubtest_requirements`.
```
srittau, line 198: Did you mean `ignore_missing_stub`? I would also not mention the stub uploader as that is mostly irrelevant for usual contributors. Also maybe word it as "... is identical to ..." Otherwise it sounds a bit as if this field must be placed after the other field physically.

```suggestion
* `partial_stub` (optional): This field marks the type stub package as
  [partial](https://peps.python.org/pep-0561/#partial-stub-packages). This is for
  3rd-party stubs that don't cover the entirety of the package's public API.
  In most cases, this field is identical to `ignore_missing_stub`.
```

(Sidenote: I just noticed that we document neither `ignore_missing_stub` nor `stubtest_requirements` in `CONTRIBUTING`.)

tests/parse_metadata.py
```diff
@@ -123,11 +123,22 @@ class StubMetadata:
     obsolete_since: Annotated[str, "A string representing a specific version"] | None
     no_longer_updated: bool
     uploaded_to_pypi: Annotated[bool, "Whether or not a distribution is uploaded to PyPI"]
+    partial_stub: Annotated[bool, "Whether stub_uploader should distribute as a partial type stub."]
```
srittau, line 126:
```suggestion
    partial_stub: Annotated[bool, "Whether this is a partial type stub package"]
```

srittau (changes requested):
Thanks! A few wording suggestions below.

Avasam:
> I agree that splitting these two things makes sense, because other use-case where stubs can be partial is when `allowlist` contains a lot of entries.

Exactly. If you know of any stub for which it is the case *right now*, let me know so I don't mark it as partial.

CONTRIBUTING.md
Avasam: oops thanks.

srittau (approved):

JelleZijlstra:
The CI failures look like GitHub bugs (there's an active issue, https://www.githubstatus.com/). Let's retrigger after GH is stable again.

PR #10161, opened 2023-05-09
Update `redis` to `4.5.5`

sobolevn:
Closes https://github.com/python/typeshed/pull/10159

PR #10162, opened 2023-05-09
[redis] Add new methods to `RedisClusterCommands`

sobolevn:
This is a follow up to https://github.com/python/typeshed/pull/10161

I've noticed that there are missing methods here: https://github.com/redis/redis-py/blob/f056118224e851915922de02ec40f2d16c9e4dd7/redis/commands/cluster.py#L644-L665

I won't bother adding correct type hints right now, but it is still better to have these methods.


AlexWaygood (approved):
These methods already have type annotations in the source code, and tbh I wouldn't mind just blindly trusting those. But this is also fine!

sobolevn:
I checked them, they look very suspicious: https://github.com/redis/redis-py/blob/f056118224e851915922de02ec40f2d16c9e4dd7/redis/asyncio/cluster.py#L68-L70

AlexWaygood:
> I checked them, they look very suspicious: https://github.com/redis/redis-py/blob/f056118224e851915922de02ec40f2d16c9e4dd7/redis/asyncio/cluster.py#L68-L70

Yes, good point :)

PR #10163, opened 2023-05-09
Add pyxdg stubs

lengau:


stubs/pyxdg/xdg/DesktopEntry.pyi
```diff
@@ -0,0 +1,67 @@
...
+from os import PathLike
+from typing import AnyStr
+from typing_extensions import Literal
+
+from xdg.IniFile import IniFile
+
+class DesktopEntry(IniFile):
+    defaultGroup: str
+    content: dict[str, dict[str, str]]
+    def __init__(self, filename: PathLike[AnyStr] | None = None) -> None: ...
```
JelleZijlstra, line 11: AnyStr doesn't make sense here because the class is not generic. I think you just want a union:
```suggestion
    def __init__(self, filename: PathLike[str | bytes] | None = None) -> None: ...
```
Same on the next line.

JelleZijlstra: Also, since `.new()` only supports str paths and this function delegates to new sometimes, maybe it's safer to also only allow str paths here.

stubs/pyxdg/xdg/DesktopEntry.pyi
```diff
@@ -0,0 +1,67 @@
...
+    def getType(self) -> str: ...
+    def getVersion(self) -> float: ...
+    def getVersionString(self) -> str: ...
+    def getName(self) -> str: ...
+    def getGenericName(self) -> str: ...
+    def getNoDisplay(self) -> bool: ...
+    def getComment(self) -> str: ...
+    def getIcon(self) -> str: ...
+    def getHidden(self) -> bool: ...
+    def getOnlyShowIn(self) -> str: ...
```
JelleZijlstra, line 23: This calls `get()` with `list=True`, so I think it returns a `list[str]`.

stubs/pyxdg/xdg/IconTheme.pyi
```diff
@@ -0,0 +1,59 @@
+from _typeshed import Unused
+from collections.abc import Collection
+from os import PathLike
+from typing import AnyStr
+
+from xdg.IniFile import IniFile
+
+basedir: Unused  # This is something that shouldn't be included since it's
```
JelleZijlstra, line 8: Incomplete sentence. And if it shouldn't be included, maybe we should just omit it from the stub?

stubs/pyxdg/xdg/IconTheme.pyi
```diff
@@ -0,0 +1,59 @@
...
+
+class IconTheme(IniFile):
+    def __init__(self) -> None: ...
+    dir: str
+    name: str
+    comment: str
+    directories: list[str]
+    type: str
+    def parse(self, file: PathLike[AnyStr]) -> None: ...  # type: ignore[override]
```
JelleZijlstra, line 17: I don't think this works with bytes paths. `self.name` comes out of the file through some `os.path` operations, and it's returned from `__repr__`, but `__repr__` is only allowed to return `str`.
```suggestion
    def parse(self, file: PathLike[str]) -> None: ...  # type: ignore[override]
```

stubs/pyxdg/xdg/IconTheme.pyi
```diff
@@ -0,0 +1,59 @@
...
+    def getDir(self) -> str: ...
+    def getName(self) -> str: ...
+    def getComment(self) -> str: ...
+    def getInherits(self) -> list[str]: ...
+    def getDirectories(self) -> list[str]: ...
+    def getScaledDirectories(self) -> list[str]: ...
+    def getHidden(self) -> bool: ...
+    def getExample(self) -> str: ...
+    def getSize(self, directory: PathLike[AnyStr]) -> int: ...
```
JelleZijlstra, line 26: It doesn't look like these would work well with bytes. Let's use str instead of AnyStr.

stubs/pyxdg/xdg/IconTheme.pyi
```diff
@@ -0,0 +1,59 @@
...
+    def checkKey(self, key: str, value: str, group: str) -> None: ...
+
+class IconData(IniFile):
+    def __init__(self) -> None: ...
+    def parse(self, file: PathLike[AnyStr]) -> None: ...  # type: ignore[override]
+    def getDisplayName(self) -> str: ...
+    def getEmbeddedTextRectangle(self) -> list[int]: ...
+    def getAttachPoints(self) -> list[tuple[int, int]]: ...
+    def checkExtras(self) -> None: ...
+    def checkGroup(self, group) -> None: ...
```
JelleZijlstra, line 44:
```suggestion
    def checkGroup(self, group: str) -> None: ...
```

stubs/pyxdg/xdg/BaseDirectory.pyi
```diff
@@ -0,0 +1,19 @@
...
+from typing import AnyStr
+
+xdg_data_home: str
+xdg_data_dirs: list[str]
+xdg_config_home: str
+xdg_config_dirs: list[str]
+xdg_cache_home: str
+xdg_state_home: str
+
+def save_config_path(*resource: PathLike[AnyStr]) -> str: ...
```
JelleZijlstra, line 12: AnyStr is correct for these functions

JelleZijlstra (commented):
Thanks! I looked at the files up to `IconTheme.pyi` and noticed a few issues. Mostly:

- AnyStr is used too much, often in places where a type variable doesn't really make sense, and for places where bytes paths don't really work.
- Some of the `.get()` calls return the wrong type.

stubs/pyxdg/xdg/IconTheme.pyi
lengau: I'm new to making stubs — when I omitted it, stubtest started complaining. I can only see a way to disable this for the whole project at a command-line level, not for just a single item. What's the best way to handle this?

(The rest of the sentence is that it's not part of the public API but just a result of a for loop in the file)

JelleZijlstra: You can add an allowlist file that lists specific names; you'll see examples of that in several other third-party stubs.

AlexWaygood: E.g. https://github.com/python/typeshed/blob/main/stubs/SQLAlchemy/%40tests/stubtest_allowlist.txt as an example :)

stubs/pyxdg/xdg/BaseDirectory.pyi
lengau: You got me double-checking and turns out it's not :-) They do joins from `os.environ` rather than `os.environb`, so they all need to be strings

stubs/pyxdg/xdg/Menu.pyi
```diff
@@ -0,0 +1,168 @@
...
+    def handle_moves(self, menu: Menu) -> None: ...
+
+class MenuEntryCache:
+    cacheEntries: dict[str, list[MenuEntry]]
+    cache: dict[str, list[MenuEntry]]
+    def __init__(self) -> None: ...
+    def add_menu_entries(self, dirs: Iterable[str], prefix: str = "", legacy: bool = False) -> None: ...
+    def get_menu_entries(self, dirs: Collection[str], legacy: bool = True) -> list[MenuEntry]: ...
+
+def parse(filename: AnyStr | None = None, debug: bool = False) -> XMLMenuBuilder: ...
```
AlexWaygood, line 168: `AnyStr` looks incorrect here -- it's only used once in the signature, so it's not a generic function. That means using a `TypeVar` has little meaning here; you should either use `str`, `bytes` or `str | bytes`

stubs/pyxdg/xdg/MenuEditor.pyi
```diff
@@ -0,0 +1,146 @@
...
+_MenuItem: TypeAlias = Menu | MenuEntry | Separator
+
+class MenuEditor:
+    menu: Menu
+    filename: str
+    tree: Incomplete
+    parser: XMLMenuBuilder
+    filenames: list[str]
+    def __init__(self, menu: Menu | None = None, filename: PathLike[AnyStr] | None = None, root: bool = False) -> None: ...
+    def parse(self, menu: Menu | None = None, filename: PathLike[AnyStr] | None = None, root: bool = False) -> None: ...
```
AlexWaygood, line 17: This class isn't generic so `AnyStr` doesn't mean much here either

AlexWaygood (commented):
Spotted two more incorrect uses of AnyStr :)

stubs/pyxdg/xdg/IconTheme.pyi
lengau: Ahh never mind, got it. Just needed to add it to the allow list.

stubs/pyxdg/xdg/IconTheme.pyi
JelleZijlstra: Bytes paths are often a bit of an afterthought and APIs only really work thoroughly for them if the API author has been careful to preserve support. It sounds like pyxdg didn't really do that. Few real-world users are likely to pass bytes paths, so it's probably not a problem for them.

stubs/pyxdg/xdg/MenuEditor.pyi
```diff
@@ -0,0 +1,147 @@
+from os import PathLike
+from typing import overload
+from typing_extensions import TypeAlias
+
+from stdlib.xml.etree.ElementTree import ElementTree
```
AlexWaygood, line 5: We do imports in the stdlib in typeshed the same as you'd do them at runtime :) Type checkers won't recognise a "stdlib" package

```suggestion
from xml.etree.ElementTree import ElementTree
```

AlexWaygood, line 5: We do imports in the stdlib in typeshed the same as you'd do them at runtime :) Type checkers won't recognise a "stdlib" package

```suggestion
from xml.etree.ElementTree import ElementTree
```

lengau: Yeah that was a mistake in trusting pycharm without checking it :-)

stubs/pyxdg/xdg/BaseDirectory.pyi
```diff
@@ -8,11 +8,11 @@ xdg_config_dirs: list[str]
...
-def load_config_paths(*resource: PathLike[str]) -> Iterator[str]: ...
-def load_first_config(*resource: PathLike[str]) -> str: ...
-def load_data_paths(*resource: PathLike[str]) -> Iterator[str]: ...
+def save_config_path(*resource: PathLike[str] | str) -> str: ...
+def save_data_path(*resource: PathLike[str] | str) -> str: ...
+def save_cache_path(*resource: PathLike[str] | str) -> str: ...
+def save_state_path(*resource: PathLike[str] | str) -> str: ...
+def load_config_paths(*resource: PathLike[str] | str) -> Iterator[str]: ...
+def load_first_config(*resource: PathLike[str] | str) -> str: ...
+def load_data_paths(*resource: PathLike[str] | str) -> Iterator[str]: ...
```
AlexWaygood, line 17: Use our `_typeshed.StrPath` alias for cases like these:

https://github.com/python/typeshed/blob/ecf4e708bff2f99d3297f801e52f0ba1d7ca0436/stdlib/_typeshed/__init__.pyi#L134

lengau:
Thanks @JelleZijlstra and @AlexWaygood for the reviews! I noted some other errors, but I think I've got things fixed up now.

stubs/pyxdg/xdg/Menu.pyi
```diff
@@ -0,0 +1,167 @@
...
+
+TYPE_DIR: Literal[0] = 0
+TYPE_FILE: Literal[1] = 1
+
+class XMLMenuBuilder:
+    debug: bool
+    def __init__(self, debug: bool = False) -> None: ...
+    cache: MenuEntryCache
+    def parse(self, filename: str | None = None) -> Menu: ...
+    def parse_menu(self, node: xml.dom.Node, filename: str): ...
```
JelleZijlstra, line 137:
```suggestion
    def parse_menu(self, node: xml.dom.Node, filename: str) -> Menu: ...
```

stubs/pyxdg/xdg/Menu.pyi
```diff
@@ -0,0 +1,167 @@
...
+    Directory: Menu | None
+    Entries: list[str]
+    Doc: str
+    Filename: str
+    Depth: int
+    Parent: Menu | None
+    NotInXml: bool
+    Show: bool
+    Visible: int
+    AppDirs: list[str | bytes]
```
JelleZijlstra, line 28: Are you sure this can contain bytes?

stubs/pyxdg/xdg/RecentFiles.pyi
```diff
@@ -0,0 +1,28 @@
...
+    def __init__(self) -> None: ...
+    def parse(self, filename: StrPath | None = None) -> None: ...
+    def write(self, filename: StrOrBytesPath | None = None) -> None: ...
+    def getFiles(
+        self, mimetypes: Iterable[str] | None = None, groups: Iterable[Incomplete] | None = None, limit: int = 0
+    ) -> list[StrPath]: ...
+    def addFile(
+        self, item: StrPath, mimetype: str, groups: Iterable[Incomplete] | None = None, private: bool = False
+    ) -> None: ...
+    def deleteFile(self, item: StrPath) -> None: ...
```
JelleZijlstra, line 16: I think this take a `RecentFile` in addition to `StrPath`.

stubs/pyxdg/xdg/util.pyi
```diff
@@ -0,0 +1,14 @@
...
+from _typeshed import StrPath
+from typing_extensions import Literal, TypeVar
+
+_StrPathT = TypeVar("_StrPathT", bound=StrPath)
+
+PY3: Literal[True]
+
+def u(s: str) -> str: ...
+
+if sys.version_info >= (3, 3):
```
JelleZijlstra, line 11: We don't care about 3.2, so we can just unconditionally re-export `shutil.which`.

JelleZijlstra (commented):
Thanks, looked over the rest of the PR now.

stubs/pyxdg/xdg/IniFile.pyi
```diff
@@ -0,0 +1,87 @@
...
+    fileExtension: str
+    filename: str
+    tainted: bool
+    content: dict[str, dict[str, str]]
+    warnings: list[str]
+    errors: list[str]
+    def __init__(self, filename: str | None = None) -> None: ...
+    def __cmp__(self, other: IniFile) -> bool: ...
+    def parse(self, filename: str, headers: Iterable[str] | None = None) -> None: ...
+    @overload
```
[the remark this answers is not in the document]
JelleZijlstra: We still need to get rid of the `str | list[str]` return types. Union return types should be avoided in general, because they are hard to work with (they force users to follow any call with an `isinstance` or similar to narrow the type).

If this method isn't expected to be used by external users, it's fine to simply have it return Any.

stubs/pyxdg/xdg/IniFile.pyi
```diff
@@ -0,0 +1,87 @@
...
+        self, key: str, group: str | None, locale: bool, type: Literal["boolean"], list: bool = False, strict: bool = False
+    ) -> bool | list[bool]: ...
+    @overload
+    def get(
+        self, key: str, group: str | None, locale: bool, type: Literal["integer"], list: bool = False, strict: bool = False
+    ) -> int | list[int]: ...
+    @overload
+    def get(
+        self, key: str, group: str | None, locale: bool, type: Literal["numeric"], list: bool = False, strict: bool = False
+    ) -> float | list[bool]: ...
```
JelleZijlstra, line 34: list[float] surely?

AlexWaygood:
@lengau could you possibly address Jelle's latest round of review on this? :-)

No worries if you've been busy, just checking in on the status here!

lengau:
@AlexWaygood Yeah sorry, I've been super busy and probably will be for the rest of the month. However, I do fully intend to rebase and adjust in August.

stubs/pyxdg/@tests/test_cases/check_IniFile.py
```diff
@@ -0,0 +1,277 @@
...
+from xdg.IniFile import IniFile
+
+ini_file = IniFile()
+
+# The "get" method is quite complex with many overloads. Check that many forms
+# are valid.
+# The function definition is:
+# def get(self, key, group=None, locale=False, type="string", list=False, strict=False):
+
+# Get str
```
AlexWaygood, line 12: You could use `typing_extensions.assert_type()` to test that the type is inferred correctly by type checkers

stubs/pyxdg/xdg/IniFile.pyi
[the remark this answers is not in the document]
lengau: I've modified these and added a series of checks that I believe are exhaustive.

stubs/pyxdg/@tests/test_cases/check_IniFile.py
AlexWaygood: Examples of how to use it:

https://github.com/python/typeshed/blob/3a56e0034c98cadf935ab1f57ca852702cbcfed2/test_cases/stdlib/builtins/check_pow.py#L6-L31

lengau: Done!

PR #10166, opened 2023-05-10
Fix return type of FPDF.output()

srittau:


JelleZijlstra (approved):
Confirmed in the code that it returns `None` if a name is given and a buffer otherwise.

Issue #10149, opened 2023-05-06
Remove stubs for pyvmomi (not until November 2023)


hauntsaninja:
https://github.com/python/typeshed/pull/10148

intgr:
Just for information, I tried the upstream type hints and they are not usable yet. With the initial pyvmomi release shipping type stubs (8.0.1.0), there are syntax errors a few of the files, preventing their use completely with mypy (https://github.com/vmware/pyvmomi/issues/1015), and beyond that it looks like there will be other issues as well.

I'm sure the upstream stubs will improve over time, just need to wait some more. November deadline seems reasonable.

The upside is that upstream type stubs look like they're autogenerated and very extensive, they cover all of the API surface, not just a few object types like `types-pyvmomi`.

AlexWaygood:
Ah, thanks for letting us know. If the upstream type hints are completely unusable, maybe it would be better to revert #10148 for now. (With the stubs marked as obsolete, the description on PyPI tells users not to use `types-pyvmomi` anymore, which is obviously incorrect advice right now.) This is a pretty unusual situation, though.

AlexWaygood:
Okay, for now we've reverted marking the stubs as obsolete. We can mark `types-pyvmomi` as obsolete again once the upstream types are usable.

intgr:
With the latest pyvmomi version 8.0.1.0.1 (May 25), the upstream type stubs work well with mypy for my use cases.

PyCharm still struggles with upstream types, but I have not investigated that in detail.

AlexWaygood:
"Upstream types don't work perfectly with all type checkers" feels like it falls into the category of "the stubs are somewhat buggy" rather than "the stubs are fundamentally broken". So I'd vote for marking `types-pyvmomi` as obsolete once more, in that case. Thanks for the update @intgr!

AlexWaygood:
Now marked as obsolete since 8.0.1.0.1: https://github.com/python/typeshed/commit/2ec6ba979f98c3c49d3a7ac2b11f1b3459e8e95a

PR #10171, opened 2023-05-10
Revert "[stubsabot] Mark pyvmomi as obsolete since 8.0.1.0 (#10148)"

AlexWaygood:
This reverts commit 853d01d55ef5b5c7f0733e62c2e214be1d472657.

As per https://github.com/python/typeshed/issues/10149#issuecomment-1542015902, it seems like the stubs bundled with the upstream `pyvmomi` package aren't currently usable due to syntax errors. Given that marking the stubs as obsolete causes a line to be added to the PyPI description instructing users to uninstall `types-pyvmomi`, it seems like we should probably _not_ mark the stubs as obsolete until the upstream types are free of syntax errors.

### DIFF
--- a/stubs/pyvmomi/METADATA.toml
+++ b/stubs/pyvmomi/METADATA.toml
@@ -1,5 +1,4 @@
 version = "8.0.0.*"
-obsolete_since = "8.0.1.0" # Released on 2023-05-05
 partial_stub = true
 
 [tool.stubtest]


AlexWaygood:
Note that if this is merged, stubsabot will quickly propose a PR to mark the stubs as obsolete again. We'll just have to sit on that PR until pyvmomi cut a new release that doesn't have syntax errors in the stub files.

JelleZijlstra (approved):

Issue #8647, opened 2022-08-29
`builtins.slice` should be generic


randolf-scholz:
This is a port of https://github.com/python/typing/issues/159, which was closed as being out of scope, with https://github.com/srittau suggesting it maybe be more suitable for [typeshed](https://github.com/python/typeshed) instead.

It would be useful to have `slice` being generic and allowing type hints a la

- `slice[start_type, stop_type, step_type]`
- `slice[S, T]` as shorthand for `slice[S, S, T]` or `slice[S, T, Any]`
- `slice[T]` as shorthand for `slice[T, T, T]` or `slice[T, T, Any]` or `slice[Any, T, Any]`
- `slice` as shorthand for `slice[Any, Any, Any]`

Syntax wise, I would advocate for the latter options `slice[S, T]=slice[S, T, Any]`, `slice[T]=slice[Any, T, Any]` for the following reasons:

1. It resembles the regular usage of `slice`: 
    - `slice(x) = slice(None, x, None)`, 
    - `slice(a, b) = slice(a, b, None)`
2. It has better compatible with `datetime`/`timedelta` variables, which might be a common use-case.
   - `slice[datetime]=slice[datetime, datetime, datetime]` would be an error, because `datetimes cannot be added, only timedeltas can be added to datetimes.

### Real-world examples

- Datetime based slicing: Here, `start` and `stop` are `datetime`-like and `step` is `timedelta`-like (e.g. `datetime.datetime` / `datetime.timedelta`, `numpy.datetime64`/`numpy.timedelta64` or `pandas.Timestamp`/`pandas.Timedelta`)
- Label based slicing via strings, for example, `pandas` allows something like `frame.loc["2017-07-01":"2017-10-31"]` for a `DataFrame` that is indexed by a `DatetimeIndex`, this will first convert the strings to `datetime` objects and then select all rows of the table that lie between the given dates.
- Similarly, one might want to select all rows of a table that lie between two float values.

TeamSpen210:
An example of heterogeneous types (although a little cute) is to do `obj["keyword": value]`, effectively producing keyword arguments.

Gobot1234:
Related https://peps.python.org/pep-0696/#using-another-typevarlike-as-the-default which were my thoughts on the matter

AlexWaygood:
Marking as deferred until and unless PEP-696 is accepted by the Steering Council and implemented by all major type checkers

lorenzo-w:
### Little status update:

[...]

PR #10174, opened 2023-05-12
Make the built-in `slice` type generic in terms of its `start`, `stop`, and `step`

daniil-berg:
## Motivation

The idea for `builtins.slice` to be generic goes back as far as 2015 (see python/typing#159) and the discussion was ported to issue #8647 more recently. See the latter issue as well as the related discussion in python/mypy#2410 for some great points about _why_ this may be useful.

In short, many libraries that provide some _sequence-like_ class offer a slicing mechanism with non-integer parameters, such as for example a time series object containing `datetime` items that can be sliced by **starting** and/or **stopping** with a specific `datetime` and potentially restricting the slice **step** to a specific `timedelta`.

## Implementation

As it is right now, it is fairly straightforward. `builtins.slice` is made **generic** and parameterized by three type variables without an upper bound. The `__init__` signatures and the read-only `start`, `stop`, and `step` properties are adjusted accordingly.

### Missing type arguments

Since [`slice`](https://docs.python.org/3/library/functions.html#slice) allows initialization with just _one_ argument and such a call only binds the `stop` type argument, initializing it as e.g. `slice(5)` will leave the `start` and `step` types undefined, which in practice will probably have the type of the slice be inferred as `slice[Any, int, Any]`. But that will still cause `mypy --stict` so complain that the assignment needs to be clarified with a specific type annotation like e.g. `s: slice[int, int, int] = slice(5)`.

Likewise, initialization with just two arguments is also possible and would leave the `step` type undefined.

### No shorthand notation

With this naive implementation, the `slice` type can only ever be used with exactly three type arguments. Some people have suggested that it would be very convenient to have **shorthand** notations for implicitly setting one or two of the three type arguments.

`slice[S, T]` would be a reasonable shorthand for `slice[S, T, Any]`.

`slice[T]` would be a reasonable shorthand either for `slice[Any, T, Any]` or `slice[T, T, Any]`. I tend to favor the latter because it makes sense to assume the `start` type to match the `stop` type, if only the `stop` type is specified.

In any case, no shorthand is currently supported.

## Future improvements

The type variable defaults proposed in [PEP 696](https://peps.python.org/pep-0696) would greatly improve a generic `slice` type by _partially_ solving the limitations mentioned above. Specifically, being able to [use another type variable as a default](https://peps.python.org/pep-0696/#using-another-typevarlike-as-default) picks up this exact issue.

I still thought it best to provide one imperfect solution now and incorporate those improvements later on. I am willing to pick up any future work directly related to this change, if this is deemed desirable at all.

## Miscellaneous

I went ahead and modified every existing `slice` annotation in `stdlib`/`stubs` to `slice[Any, Any, Any]` for now (except the one in `pyinstaller` because there was an explicit comment indicating `slice[None, None, None]` may be more appropriate). I am pretty sure this is not optimal because at least in a few cases a more restrictive annotation (e.g. `slice[int, int, int]`) may be appropriate/obvious. But I thought I would put forward the proposal first and fine-tune it later, if the maintainers think it is worthwhile.

Running the CI tests I only got errors from `pytype` in all places, where I passed type arguments to `slice`:
```
pytype.pytd.visitors.ContainerError: Class builtins.slice is not a container
```

I am not sure why `pytype` is ignoring the typeshed stubs. But the rest of the tests seemed to pass normally.

I did not run the Mypy primer script locally. I am curious to see the results of the workflow triggered by this PR. I expect quite a few [`type-arg`](https://mypy.readthedocs.io/en/stable/error_code_list2.html#check-that-type-arguments-exist-type-arg) errors for obvious reasons.


srittau:
While I agree that the slice type should be generic, doing it before we have type var defaults is just too disruptive, as the primer hits show.

AlexWaygood:
I agree with @srittau. Without PEP 696, this would just be too much of a compatibility break, unfortunately. It sounds like [we might have to wait for a year or so](https://discuss.python.org/t/pep-696-type-defaults-for-typevarlikes/22569/6?u=alexwaygood) before we get a decision from the Steering Council on PEP 696, so I'm closing this PR for now. But thank you for it -- we can reconsider once a decision on PEP 696 has been made.



> Running the CI tests I only got errors from `pytype` in all places, where I passed type arguments to `slice`:
> 
> ```
> pytype.pytd.visitors.ContainerError: Class builtins.slice is not a container
> ```
> 
> I am not sure why `pytype` is ignoring the typeshed stubs.

Pytype has its own version of `builtins.pyi` and `typing.pyi`; it doesn't pay attention to what typeshed has for these modules. This often causes a bit of confusion in our CI :)

srittau:
@AlexWaygood Should we mark this PR as deferred and leave it open?

AlexWaygood:
> @AlexWaygood Should we mark this PR as deferred and leave it open?

I'm not really a fan of leaving PRs open for extended periods of time. They tend to be hard to keep up to date (merge conflicts tend to stack up), so it's generally easier just to recreate them when the time comes. And it sounds like the Steering Council is resolute in not taking a decision on PEP 696 for quite a while (unfortunately). We already have #8647 open to remind us to look into this once PEP 696 is (hopefully) accepted.

daniil-berg:
Fair enough. Thank you for taking the time to respond and elaborate. I guess it makes sense not to rely on people keeping their PR branches up to date for extended periods of time.

For what it's worth, I agree with your position on the order of PEP 695 vs. PEP 696 and wish the SC had come to a different decision.

With a bit of luck I'll be able to keep this up to date and ready for when PEP 696 lands. Then maybe you can reopen this in the future.

AlexWaygood:
> With a bit of luck I'll be able to keep this up to date and ready for when PEP 696 lands. Then maybe you can reopen this in the future.

@daniil-berg absolutely. Very happy to reconsider when that time comes :)

jorwoods:
PEP 696 has been accepted and will be in Python 3.13.

daniil-berg:
Hi @AlexWaygood and @srittau. Since PEP 696 has been accepted, I overhauled the branch to incorporate type parameter defaults for the start, stop, and step parameters. I just rebased the branch on top of the current `main`. The [commit message](https://github.com/python/typeshed/compare/main...daniil-berg:typeshed:slice-generic) basically spells out what I changed.

The type parmeters default to `Any`. I introduced the `IndexSlice` type alias for `slice[SupportsIndex | None, SupportsIndex | None, SupportsIndex | None]` for convenience because I use that type a lot throughout the stubs.

Basically, wherever there is an overload for a method that accepts either a `SupportsIndex` or a `slice` argument, I replaced `slice` with `IndexSlice`. There were a few instances throughout the stubs, where similarly overloaded methods accepted `int` or `slice`. (Example: [`SupportsGetItemBuffer`](https://github.com/python/typeshed/blob/d326c9bd424ad60c2b63c2ca1c5c1006c61c3562/stdlib/_typeshed/__init__.pyi#L291-L294)) Since I wanted my changes to be minimally invasive, I took that at face value and replaced `slice` with `slice[int | None, int | None, int | None]` instead of my new `IndexSlice`. But I have a feeling that those `int` annotations were an oversight, so I can change that, if you like.

There was also some weird stuff going on in the AST modules, where `slice` was being shadowed. I did not touch those.

The `slice.__new__` method returns either `slice[None, _T, None]` or `slice[_T1, _T2, _T3]` depending on the overload signature.

I humbly ask for this PR to be re-opened and my proposal to be reconsidered.

srittau:
Unfortunately, the PR can't be reopened due to the rebase. Please open a new PR. Thanks for working on this.

obskyr:
If the author wants to make a new PR, I know I'd like this feature! Having a native slice type that lets you represent any range&nbsp;– for example, start and end times in a video file&nbsp;– would be wonderful.

obskyr:
If anyone else finds this like I did, here’s the situation at the time of writing: `slice` *is* now generic (https://github.com/python/typeshed/pull/11637), but not yet at runtime (https://github.com/python/cpython/pull/128336).

PR #10176, opened 2023-05-13
Update _SSLProtocolTransport._ssl_protocol

Dreamsorcerer:
This change just caught us out:
https://github.com/python/cpython/pull/98540/files

### DIFF
--- a/stdlib/asyncio/sslproto.pyi
+++ b/stdlib/asyncio/sslproto.pyi
@@ -66,7 +66,7 @@ class _SSLProtocolTransport(transports._FlowControlMixin, transports.Transport):
     _sendfile_compatible: ClassVar[constants._SendfileMode]
 
     _loop: events.AbstractEventLoop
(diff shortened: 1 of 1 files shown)


stdlib/asyncio/sslproto.pyi
```diff
@@ -66,7 +66,10 @@ class _SSLProtocolTransport(transports._FlowControlMixin, transports.Transport):
     _sendfile_compatible: ClassVar[constants._SendfileMode]
 
     _loop: events.AbstractEventLoop
-    _ssl_protocol: SSLProtocol
+    if sys.version_info >= (3, 11, 1):
```
JelleZijlstra, line 69:
```suggestion
    if sys.version_info >= (3, 11):
```
We don't do patch releases on typeshed, we pretend everything is as it was in the most recent release of the series.

JelleZijlstra (approved):

Issue #10116, opened 2023-04-30
`asyncio` no longer supports generator-based coroutines in 3.12+


AlexWaygood:
What the title says.

On Python 3.11:

```pycon
>python -m asyncio
asyncio REPL 3.11.2 (tags/v3.11.2:878ead1, Feb  7 2023, 16:38:35) [MSC v.1934 64 bit (AMD64)] on win32
Use "await" directly instead of "asyncio.run()".
Type "help", "copyright", "credits" or "license" for more information.
>>> import asyncio
>>> def foo():
...     yield from []
...
>>> asyncio.create_task(foo())
<Task pending name='Task-1' coro=<foo() running at <console>:1>>
>>> exit()
```

On Python 3.12:

```pycon
>python -m asyncio
Running PGUpdate|x64 interpreter...
asyncio REPL 3.12.0a7+ (heads/main-dirty:79b9db9295, Apr 28 2023, 22:41:05) [MSC v.1932 64 bit (AMD64)] on win32
Use "await" directly instead of "asyncio.run()".
Type "help", "copyright", "credits" or "license" for more information.
>>> import asyncio
>>> def foo():
...     yield from []
...
>>> asyncio.create_task(foo())
Traceback (most recent call last):
  File "C:\Users\alexw\coding\cpython\Lib\concurrent\futures\_base.py", line 449, in result
    return self.__get_result()
           ^^^^^^^^^^^^^^^^^^^
  File "C:\Users\alexw\coding\cpython\Lib\concurrent\futures\_base.py", line 401, in __get_result
    raise self._exception
  File "C:\Users\alexw\coding\cpython\Lib\asyncio\__main__.py", line 34, in callback
    coro = func()
           ^^^^^^
  File "<console>", line 1, in <module>
  File "C:\Users\alexw\coding\cpython\Lib\asyncio\tasks.py", line 377, in create_task
    task = loop.create_task(coro)
           ^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\alexw\coding\cpython\Lib\asyncio\base_events.py", line 436, in create_task
    task = tasks.Task(coro, loop=self, name=name, context=context)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: a coroutine was expected, got <generator object foo at 0x000001E6821AA200>
```

@kumaraditya303 tells me that the same change has been made across `asyncio` on Python 3.12.

[...]

PR #10177, opened 2023-05-13
[asyncio] Make `create_task` work with coroutines only in 3.12

sobolevn:
I've decided to start small and change the most obvious cases of `Generator` usage in `asyncio`.

Refs #10116


JelleZijlstra:
Would it be better to add an alias that expands to `Coroutine` only in 3.12 and `Coroutine | Generator` (plus typevars) in 3.11 and lower?

stdlib/asyncio/tasks.pyi
```diff
@@ -46,6 +46,11 @@ _FT = TypeVar("_FT", bound=Future[Any])
 _FutureLike: TypeAlias = Future[_T] | Generator[Any, None, _T] | Awaitable[_T]
 _TaskYieldType: TypeAlias = Future[object] | None
 
+if sys.version_info >= (3, 12):
+    _CoroutineLike: TypeAlias = Coroutine[Any, Any, _T]
```
AlexWaygood, line 50: Maybe this alias should go in `asyncio/__init__.pyi`, so that it can be easily imported from anywhere in the `asyncio/` directory

sobolevn: I've tried it locally and it looks a bit wierd there:
Снимок экрана 2023-05-14 в 11 01 08

I like it more in `tasks.py` :)

As you can see in this PR, importing it is rather simple as-is.

AlexWaygood: > I've tried it locally and it looks a bit wierd there:

It doesn't look _too_ weird to me, and we do a similar thing in `email/__init__.pyi`:

https://github.com/python/typeshed/blob/9c212cdf5c1573f71d4fa55f02cfd058e2a4a6e3/stdlib/email/__init__.pyi#L1-L14

Putting the alias in `asyncio/tasks.pyi` just feels quite obscure to me, considering we might be using it in lots of places which probably don't have much to do with tasks.

Alternatively, we could do something like this: https://github.com/python/typeshed/commit/04e02a1da8220c81b319bf168d2a31fa25f688ec, and then add the alias to `_typeshed.asyncio`. I can turn that into a PR if you think it's worth it? :)

sobolevn: Ok, I moved it to `__init__.pyi` 👍

AlexWaygood: Thank you!

stdlib/asyncio/__init__.pyi
```diff
@@ -1,4 +1,14 @@
...
+from collections.abc import Coroutine, Generator
+from typing import Any, TypeVar
+from typing_extensions import TypeAlias
+
+_T = TypeVar("_T")
+
+if sys.version_info >= (3, 12):
+    _CoroutineLike: TypeAlias = Coroutine[Any, Any, _T]
+else:
+    _CoroutineLike: TypeAlias = Generator[Any, None, _T] | Coroutine[Any, Any, _T]
```
AlexWaygood, line 11: It's going to make flake8-pyi angry:

```suggestion
# Aliases imported by multiple submodules in typeshed 
if sys.version_info >= (3, 12):
    _CoroutineLike: TypeAlias = Coroutine[Any, Any, _T]  # noqa: Y047
else:
    _CoroutineLike: TypeAlias = Generator[Any, None, _T] | Coroutine[Any, Any, _T]  # noqa: Y047
```

AlexWaygood (approved):

PR #10184, opened 2023-05-15
Change `asyncio.Task.__init__` signature

sobolevn:
Previously it was possible to pass `Awaitable` objects and `Generator`s into `Task.__init__`:

```python
>>> from asyncio import Task
>>> def gen():
...    yield
... 
>>> Task(gen())
<stdin>:1: DeprecationWarning: There is no current event loop
<Task pending name='Task-1' coro=<gen() running at <stdin>:1>>
```

Starting from `3.12` Python does not allow anything but coroutines in `Task.__init__`.

Proof:

```python
Python 3.12.0a7+ (heads/main-dirty:3cba61f111, May 15 2023, 12:56:55) [Clang 11.0.0 (clang-1100.0.33.16)] on darwin
>>> from asyncio import Task
>>> def gen():
...    yield 
... 
>>> Task(gen())
<stdin>:1: DeprecationWarning: There is no current event loop
TypeError: a coroutine was expected, got <generator object gen at 0x10bd47930>
```

Also changes:

```python
>>> Task(coro()).get_coro()
<coroutine object coro at 0x10bd47a00>
```

I am keeping this type alias local, because right now it is not used anywhere else. However, I see a clear difference between `Task.__init__` and `create_task()`, I think that we need to go deeper in the next PRs and find, which one is more correct.

Refs https://github.com/python/typeshed/issues/10116

### DIFF
--- a/stdlib/asyncio/tasks.pyi
+++ b/stdlib/asyncio/tasks.pyi
@@ -275,11 +275,10 @@ else:
     ) -> tuple[set[Task[_T]], set[Task[_T]]]: ...
     async def wait_for(fut: _FutureLike[_T], timeout: float | None, *, loop: AbstractEventLoop | None = None) -> _T: ...
 
-
 if sys.version_info >= (3, 12):
-    _TaskCompatibleCoro: TypeAlias =  Coroutine[Any, Any, _T_co]
+    _TaskCompatibleCoro: TypeAlias = Coroutine[Any, Any, _T_co]
 else:
-    _TaskCompatibleCoro: TypeAlias =  Generator[_TaskYieldType, None, _T_co] | Awaitable[_T_co]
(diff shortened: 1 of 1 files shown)


sobolevn:
One more thing, `Awaitable` looks suspicious here. See:

```python
Python 3.10.0 (default, Nov  1 2021, 10:24:06) [Clang 11.0.0 (clang-1100.0.33.16)] on darwin
>>> async def coro(): ...
... 
>>> from asyncio import Task
>>> class Some:
...    def __await__(self):
...       return coro().__await__()
... 
>>> Task(Some())
<stdin>:1: DeprecationWarning: There is no current event loop
TypeError: a coroutine was expected, got <__main__.Some object at 0x1024b2440>
```

This needs further work in the future PR. I will read the source code to analyze what is going on here. Probably, this needs to be changed to `Coroutine | Future` or something.

sobolevn:
Looks like GitHub has some problems with Actions again.

srittau (approved):

PR #10187, opened 2023-05-16
Treat entries on pytype_exclude_list as missing modules for pytype_test.

rchen152:
This works around the pytype errors in https://github.com/python/typeshed/pull/10122.

### DIFF
--- a/tests/pytype_test.py
+++ b/tests/pytype_test.py
@@ -177,7 +177,7 @@ def get_missing_modules(files_to_test: Sequence[str]) -> Iterable[str]:
                 # Skips comments, empty lines, and stdlib files, which are in
                 # the exclude list because pytype has its own version.
                 continue
-            unused_stubs_prefix, unused_pkg, mod_path = fi.split(os.path.sep, 2)  # pyright: ignore [reportUnusedVariable]
+            unused_stubs_prefix, unused_pkg, mod_path = fi.split("/", 2)  # pyright: ignore [reportUnusedVariable]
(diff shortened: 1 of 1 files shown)


tests/pytype_test.py
```diff
@@ -165,6 +168,17 @@ def get_missing_modules(files_to_test: Sequence[str]) -> Iterable[str]:
...
+    test_dir = os.path.dirname(__file__)
+    exclude_list = os.path.join(test_dir, "pytype_exclude_list.txt")
+    with open(exclude_list) as f:
+        excluded_files = f.readlines()
+        for fi in excluded_files:
+            if not fi.startswith(f"stubs{os.path.sep}"):
+                # Skips comments, empty lines, and stdlib files, which are in
+                # the exclude list because pytype has its own version.
+                continue
+            unused_stubs_prefix, unused_pkg, mod_path = fi.split(os.path.sep, 2)  # pyright: ignore [reportUnusedVariable]
```
srittau, line 180: Nit: `pytype_exclude_list.txt` always uses `/` as separator. While Windows is not supported by pytype, I think it's safer to hardcode the slash:

```suggestion
            unused_stubs_prefix, unused_pkg, mod_path = fi.split("/", 2)  # pyright: ignore [reportUnusedVariable]
```

srittau (changes requested):
Thanks! One nit below.

tests/pytype_test.py
rchen152: Done, thanks!

srittau (approved):

PR #10196, opened 2023-05-22
[PyYAML] Improve dump/serialize types

srittau:
* Add overloads to return `str` or `bytes`, instead of `Any`.
* Mark some arguments as keyword-only.
* Add missing default values.


srittau:
An additional reason that I marked the arguments as keyword-only is to avoid adding an additional overload.

PR #10197, opened 2023-05-22
weakref.WeakMethod callback arg type

VoodaGod:
see
https://github.com/python/cpython/blob/e295d8605699ad3d8ec46c8d55a5e47da05b20c6/Lib/weakref.py#LL60C35-L60C35
the argument of the callback is in fact a weakref to the callable type


JelleZijlstra:
Seems like it's actually the WeakMethod itself that gets passed to the callback. This is visible in the tests: https://github.com/python/cpython/blob/3a5be878be6f89ee98d0ef9a1274e6a9d9ccbc37/Lib/test/test_weakref.py#L1108.

stdlib/weakref.pyi
```diff
@@ -41,7 +41,7 @@ _P = ParamSpec("_P")
 ProxyTypes: tuple[type[Any], ...]
 
 class WeakMethod(ref[_CallableT], Generic[_CallableT]):
-    def __new__(cls, meth: _CallableT, callback: Callable[[_CallableT], object] | None = None) -> Self: ...
+    def __new__(cls, meth: _CallableT, callback: Callable[[ref[_CallableT]], object] | None = None) -> Self: ...
```
JelleZijlstra, line 44:
```suggestion
    def __new__(cls, meth: _CallableT, callback: Callable[[Self], object] | None = None) -> Self: ...
```

PR #10200, opened 2023-05-23
Update typing-extensions; some 3.12 updates

JelleZijlstra:

PR #10201, opened 2023-05-23
Update ast for 3.12

JelleZijlstra:
Split out from #10200 since this part appears to cause some pytype failures.


stdlib/ast.pyi
```diff
@@ -168,7 +168,7 @@ class NodeTransformer(NodeVisitor):
     #       The usual return type is AST | None, but Iterable[AST]
     #       is also allowed in some cases -- this needs to be mapped.
 
-_T = TypeVar("_T", bound=AST)
+_T = _TypeVar("_T", bound=AST)
```
AlexWaygood, line 171: Does pytype like it any more if we do `_T = typing.TypeVar("_T")`? (FWIW that would also be friendlier to flake8-pyi, which will recognise `typing.TypeVar("_T")` as a TypeVar declaration, but won't recognise `_TypeVar("_T")` as a TypeVar declaration.)

stdlib/_ast.pyi
```diff
@@ -59,27 +60,39 @@ class Expression(mod):
 class stmt(AST): ...
 
 class FunctionDef(stmt):
-    if sys.version_info >= (3, 10):
+    if sys.version_info >= (3, 12):
+        __match_args__ = ("name", "type_params", "args", "body", "decorator_list", "returns", "type_comment")
```
AlexWaygood, line 64: I didn't realise you were adding type_params to the__match_args__ for various AST classes. I guess that makes sense, but it feels like it could be quite a breaking change if anybody had been matching FuncDef or ClassDef nodes with "positional arguments" in their `case` statements, e.g.

```python
match node:
    case FunctionDef("foo", ast.arguments(), []):  # <- would have matched on 3.11, not on 3.12!
        ...
```

Feels like it would be much less breaking if type_params was added to the end of `__match_args__` rather than inserted into the middle?

AlexWaygood: At the very least, seems like this change to `__match_args__` should be clearly documented in https://github.com/python/cpython/pull/104642

AlexWaygood: Sure, I'll open an issue

AlexWaygood: I filed:

- https://github.com/python/cpython/issues/104799

JelleZijlstra:
Alex suggested that, not me :)

Thanks! I can try switching to `typing.TypeVar` for now.

rchen152:
I have a fix for the pytype issue out for review. With any luck, I'll get it into this week's release.

rchen152:
I think pytype should be happy with `from typing import TypeVar as _TypeVar` with today's release (2023.5.24). Could you give it a try?

rchen152:
Oh drat, it still seems broken =/ I'll look into it tomorrow.

JelleZijlstra:
Thanks! No hurry, we'll probably have to change this anyway once python/cpython#104799 gets resolved and I can always try the `typing.TypeVar` workaround.

rchen152:
Ok, should actually be fixed now XD And I made sure to run pytype_test locally this time to check that the fix works. Will be in next week's pytype release.

AlexWaygood (changes requested):
Needs to be updated now that https://github.com/python/cpython/pull/104974 is merged ;)

AlexWaygood (approved):
Looks good!

rchen152:
The pytype failure should actually be fixed now, in version 2023.6.2.

JelleZijlstra:
Thanks @rchen152!

PR #10203, opened 2023-05-24
Change `_EnumTypeWrapper` to inherit `type`

vbkaisetsu:
This change addresses the issue reported in https://github.com/protocolbuffers/protobuf/issues/10240

For example, I create the following `foo.proto`:
```proto
syntax = "proto3";

package prototest;

message Foo {
  enum Bar {
    X = 0;
    Y = 1;
  }

  Bar bar = 1;
}
```

`protoc foo.proto --pyi_out . --python_out .` generates the following `foo_pb2.pyi`:
```python
from google.protobuf.internal import enum_type_wrapper as _enum_type_wrapper
from google.protobuf import descriptor as _descriptor
from google.protobuf import message as _message
from typing import ClassVar as _ClassVar, Optional as _Optional, Union as _Union

DESCRIPTOR: _descriptor.FileDescriptor

class Foo(_message.Message):
    __slots__ = ["bar"]
    class Bar(int, metaclass=_enum_type_wrapper.EnumTypeWrapper):
        __slots__ = []
    BAR_FIELD_NUMBER: _ClassVar[int]
    X: Foo.Bar
    Y: Foo.Bar
    bar: Foo.Bar
    def __init__(self, bar: _Optional[_Union[Foo.Bar, str]] = ...) -> None: ...
```

When I use the above library in `a.py` as follows:
```python
import foo_pb2

for x in foo_pb2.Foo.Bar.keys():
    print(x)
```

`mypy` reports the following error:
```
$ mypy a.py
foo_pb2.pyi:10: error: Metaclasses not inheriting from "type" are not supported  [misc]
foo_pb2.pyi:11: error: Need type annotation for "__slots__" (hint: "__slots__: List[<type>] = ...")  [var-annotated]
a.py:3: error: "Type[Bar]" has no attribute "keys"  [attr-defined]
```

After this change:
```
$ mypy a.py
foo_pb2.pyi:11: error: Need type annotation for "__slots__" (hint: "__slots__: List[<type>] = ...")  [var-annotated]
```
(NOTE: Perhaps a modification is necessary to add a type annotation of `__slots__` to `protoc` as well.)


sobolevn (approved):
It is used as a metaclass (from what I can tell), so `type` base class seems correct.

JelleZijlstra:
Seems like it doesn't actually inherit from `type`:

```
In [7]: google.protobuf.internal.enum_type_wrapper.EnumTypeWrapper.mro()
Out[7]: [google.protobuf.internal.enum_type_wrapper.EnumTypeWrapper, object]
```

Still, adding `type` as a base might be the best thing to do pragmatically.

stubs/protobuf/google/protobuf/internal/enum_type_wrapper.pyi
```diff
@@ -6,7 +6,7 @@ _V = TypeVar("_V", bound=int)
 
 # Expose a generic version so that those using mypy-protobuf
 # can get autogenerated NewType wrapper around the int values
-class _EnumTypeWrapper(Generic[_V]):
+class _EnumTypeWrapper(type, Generic[_V]):
```
AlexWaygood, line 9:
```suggestion
# NOTE: this doesn't actually inherit from type,
# but mypy doesn't support metaclasses that don't inherit from type,
# so we pretend it does in the stubs...
class _EnumTypeWrapper(type, Generic[_V]):
```

PR #10205, opened 2023-05-24
Clean up stubtest allowlist for typing

JelleZijlstra:
Going to do some CI-driven development here


JelleZijlstra:
Yes, going through it now. There's also some missing entries in `typing_extensions.__all__`.

tests/stubtest_allowlists/py3_common.txt
```diff
@@ -407,9 +407,67 @@ pickle.Unpickler.memo  # undocumented implementation detail, has different type
...
+typing\.AsyncGenerator
+typing\.Annotated
+typing\.AbstractSet
+
+# We lie in the stub and say it's read-only
+typing_extensions\.ParamSpec(Args|Kwargs)\.__origin__
+
+# Internal attributes
+.*\.__protocol_attrs__
+.*\.__callable_proto_members_only__
```
AlexWaygood, line 458: https://github.com/python/mypy/pull/15294 means we should be able to remove these, come the next mypy release

(No change requested)

tests/stubtest_allowlists/py37.txt
```diff
@@ -159,3 +159,6 @@ typing_extensions\.reveal_type
 
 # Doesn't exist at runtime
 typing\.Protocol
+
+# Function at runtime; we pretend it's already a class
```
AlexWaygood, line 163: This comment isn't accurate. `typing_extensions.NewType` _is_ a class on all versions of Python, as of v4.6.0:

```pycon
Python 3.8.16 (default, Jan 17 2023, 22:25:28) [MSC v.1916 64 bit (AMD64)] :: Anaconda, Inc. on win32
>>> from typing_extensions import NewType
>>> NewType
<class 'typing_extensions.NewType'>
```

JelleZijlstra: I see, the issue is that in the typing_extensions stubs we always just re-export typing.NewType, which is a function in the stubs below 3.10. I suppose I should change the stub instead.

AlexWaygood: Yup, that's the conclusion I just came to as well.

AlexWaygood (approved):
hooray!

PR #10206, opened 2023-05-24
SysLogHandler facility could be a string

doc-sheet:
It is possible to pass syslog facility as string to `__init__` too


JelleZijlstra:
Thanks!

PR #10211, opened 2023-05-25
builtins: updates for py312

hauntsaninja:


stdlib/builtins.pyi
```diff
@@ -188,6 +188,8 @@ class type:
     if sys.version_info >= (3, 10):
         def __or__(self, __value: Any) -> types.UnionType: ...
         def __ror__(self, __value: Any) -> types.UnionType: ...
+    if sys.version_info >= (3, 12):
+        __type_params__: tuple[object, ...]
```
AlexWaygood, line 192: This also needs to be added to `builtins.function` and `types.FunctionType`

JelleZijlstra: And we can ideally type it as `tuple[TypeVar | ParamSpec | TypeVarTuple, ...]`, though it's possible some type checkers will get upset.

PR #10213, opened 2023-05-25
math: updates for py312

hauntsaninja:


stdlib/math.pyi
```diff
@@ -130,6 +133,10 @@ def radians(__x: _SupportsFloatOrIndex) -> float: ...
 def remainder(__x: _SupportsFloatOrIndex, __y: _SupportsFloatOrIndex) -> float: ...
 def sin(__x: _SupportsFloatOrIndex) -> float: ...
 def sinh(__x: _SupportsFloatOrIndex) -> float: ...
+
+if sys.version_info >= (3, 12):
+    def sumprod(__p: Iterable[_SupportsFloatOrIndex], __q: Iterable[_SupportsFloatOrIndex]) -> float: ...
+
```
AlexWaygood, line 139: Seems like you need types that support multiplication:

```pycon
>>> import math as m
>>> class Floatable:
...     def __float__(self):
...         return 42.0
...
>>> class Indexable:
...     def __index__(self):
...         return 42
...
>>> indexes = [Indexable(), Indexable()]
>>> floats = [Floatable(), Floatable()]
>>> m.sumprod(indexes, indexes)
TypeError: unsupported operand type(s) for *: 'Indexable' and 'Indexable'
>>> m.sumprod(floats, floats)
TypeError: unsupported operand type(s) for *: 'Floatable' and 'Floatable'
>>> m.sumprod(indexes, floats)
TypeError: unsupported operand type(s) for *: 'Indexable' and 'Floatable'
>>> m.sumprod(floats, indexes)
TypeError: unsupported operand type(s) for *: 'Floatable' and 'Indexable'
```

stdlib/math.pyi
```diff
@@ -112,7 +112,10 @@ def log1p(__x: _SupportsFloatOrIndex) -> float: ...
 def log2(__x: _SupportsFloatOrIndex) -> float: ...
 def modf(__x: _SupportsFloatOrIndex) -> tuple[float, float]: ...
 
-if sys.version_info >= (3, 9):
+if sys.version_info >= (3, 12):
+    def nextafter(__x: _SupportsFloatOrIndex, __y: _SupportsFloatOrIndex, *, steps_after: int | None = None) -> float: ...
```
AlexWaygood, line 116: - The new keyword argument is `steps`, not `steps_after`
- It takes anything that has an `__index__` method, not just `int`

```suggestion
    def nextafter(__x: _SupportsFloatOrIndex, __y: _SupportsFloatOrIndex, *, steps: SupportsIndex | None = None) -> float: ...
```

hauntsaninja: Oops, not sure what happened there, I've been testing most of these specifically too. Thanks for catching!

stdlib/math.pyi
hauntsaninja: Rather than make a mul and add protocol where I'm not quite sure of the parameter types, I opted to just go for float. If you're using e.g. numpy types you're not going to be using math.sumprod anyway.

AlexWaygood: Yeah, good call, I think — BinOp protocols are always annoyingly complex to get right anyway

AlexWaygood (approved):

PR #10214, opened 2023-05-25
random: update for py312

hauntsaninja:


stdlib/random.pyi
```diff
@@ -79,8 +81,15 @@ class Random(_random.Random):
 
     def uniform(self, a: float, b: float) -> float: ...
     def triangular(self, low: float = 0.0, high: float = 1.0, mode: float | None = None) -> float: ...
+    if sys.version_info >= (3, 12):
+        def binomialvariate(self, n: int = 1, p: float = 0.5) -> float: ...
```
AlexWaygood, line 85: The docstring says:

> Returns an integer in the range:   0 <= X <= n

https://github.com/python/cpython/blob/278030a17d6889731d91a55e9f70cbec0b07dee8/Lib/random.py#L767

At first glance, that looks correct

```suggestion
        def binomialvariate(self, n: int = 1, p: float = 0.5) -> int: ...
```

AlexWaygood (approved):

PR #10215, opened 2023-05-25
shutil.unpack_archive: update for py312

hauntsaninja:


AlexWaygood (approved):
shutil.rmtree also has a new parameter in py312, but we can do that in a separate PR

PR #10216, opened 2023-05-25
calendar: update for py312

hauntsaninja:


hauntsaninja:
Not sure what pyright wants here, bug?

AlexWaygood:
> Not sure what pyright wants here, bug?

Oh, no idea what's going on there. We're a few pyright releases behind; could try updating to the latest and see if it still reproduces?

JelleZijlstra (approved):
Looks good, the enum issue is mysterious to me too.

I would suggest it's an import cycle issue, but nothing in the stdlib relies on `calendar`.

AlexWaygood:
Yeah, seems like the failures persist with the latest pyright

hauntsaninja:
cc @erictraut any idea what's going wrong here?

erictraut:
This is caused by a circular dependency involving the `EnumMeta` class.

The `ReprEnum` class derives from `Enum` which uses the metaclass `EnumMeta`. The `EnumMeta` has a `__new__` method that describes a keyword parameter `boundary` of type `FlagBoundary`. This type derives from `StrEnum` which derives from `ReprEnum`, and we're back to the start.

This circularity can be resolved if analysis is done in precisely the right order. Since pyright is a lazy (just-in-time) type analyzer, the standard ordering can easily change. I suspect that the change in this PR perturbed things just enough to change the evaluation ordering. I have logic in place that's designed to detect circularities like this and force them to resolve in a satisfactory order, but the logic is complex, and it appears that it's not handling this case. I'll investigate it further. 

I've filed [this bug](https://github.com/microsoft/pyright/issues/5232) in the pyright issue tracker.

erictraut:
This will be fixed in the next pyright release (1.1.312).

PR #10217, opened 2023-05-25
statistics: update for py312

hauntsaninja:


stdlib/statistics.pyi
```diff
@@ -114,8 +114,13 @@ if sys.version_info >= (3, 8):
         def __rsub__(self, x2: float | NormalDist) -> NormalDist: ...
         __rmul__ = __mul__
 
-if sys.version_info >= (3, 10):
+if sys.version_info >= (3, 12):
+    def correlation(__x: Sequence[_Number], __y: Sequence[_Number], *, method: str = "linear") -> float: ...
```
AlexWaygood, line 118: I think using `Literal` would be appropriate here — it has the potential to lead to false-positive errors, _but_ Raymond has previously complained to us about similar situations in `statistics.pyi` where we _weren't_ using Literal types:

- #7462

```suggestion
    def correlation(__x: Sequence[_Number], __y: Sequence[_Number], *, method: Literal["linear", "ranked"] = "linear") -> float: ...
```

AlexWaygood (approved):

PR #10219, opened 2023-05-25
`shutil.rmtree`: Add `onexc` parameter; add parameter defaults

AlexWaygood:


stdlib/shutil.pyi
```diff
@@ -88,21 +88,34 @@ else:
     ) -> _PathReturn: ...
 
 _OnErrorCallback: TypeAlias = Callable[[Callable[..., Any], Any, Any], object]
+_OnExcCallback: TypeAlias = Callable[[Callable[..., Any], Any, Exception], object]
```
hauntsaninja, line 91: I think the second argument is `str`

hauntsaninja, line 91:
```suggestion
_OnExcCallback: TypeAlias = Callable[[Callable[..., Any], str, Exception], object]
```

stdlib/shutil.pyi
```diff
@@ -88,21 +88,34 @@ else:
     ) -> _PathReturn: ...
 
 _OnErrorCallback: TypeAlias = Callable[[Callable[..., Any], Any, Any], object]
```
AlexWaygood, line 90:
```suggestion
_OnErrorCallback: TypeAlias = Callable[[Callable[..., Any], str, Any], object]
```

AlexWaygood:
I guess it would be better to use overloads here, actually. You shouldn't be providing `onexc` _and_ `onerror` on 3.12+.

hauntsaninja (approved):

PR #10223, opened 2023-05-26
Update pyscreeze

### LINKED ISSUE #10222: Daily tests failed on Fri May 26 2023

JelleZijlstra:
Fixes #10222


stubs/PyScreeze/pyscreeze/__init__.pyi
```diff
@@ -142,6 +142,8 @@ def locateCenterOnScreen(
     step: int = 1,
     confidence: None = None,
 ) -> Point | None: ...
+def locateOnScreenNear(image: str | Image.Image | _Mat, x: int, y: int) -> Point: ...
```
srittau, line 145: From my reading of the code, this should be:

```suggestion
def locateOnScreenNear(image: str | Image.Image | _Mat, x: int, y: int) -> Box: ...
```

srittau (changes requested):

stubs/PyScreeze/pyscreeze/__init__.pyi
JelleZijlstra: You're right, thanks

srittau (approved):

PR #10225, opened 2023-05-26
Use PEP 688

### LINKED ISSUE #10224: Implement PEP 688
[PEP 688](https://peps.python.org/pep-0688/#equivalent-for-older-python-versions) has been accepted and implemented in Python 3.12. For Python 3.12 and up, this provides a clean, protocol-based way to implement buffers.

Unfortunately, we still can't directly represent buffer types on earlier Python versions. I'd like to propose the following approach:
- We pretend that `typing_extensions.Buffer` is a Protocol with a `__buffer__` method even before 3.12.
- We create `__buffer__` methods on buffer classes without a version condition, even though the method does not actually exist before 3.12.
- Now, we can define protocols that combine `__buffer__` with other methods, such as the current `_typeshed.SliceableBuffer`.

The downside is that we're lying about the existence of the `__buffer__` method, so that type checkers will not catch places where users call the method on Python < 3.12. However, hopefully this will be rare, and it is outweighed by the benefit that we can now talk about arbitrary buffer types, including third-party ones, without the big unions we're currently using in `_typeshed`.

JelleZijlstra:
Fixes #10224


stdlib/typing_extensions.pyi
```diff
@@ -398,4 +398,7 @@ else:
             def __or__(self, right: Any) -> _SpecialForm: ...
             def __ror__(self, left: Any) -> _SpecialForm: ...
 
-    class Buffer(abc.ABC): ...
+    class Buffer(Protocol):
```
AlexWaygood, line 401:
```suggestion
    @runtime_checkable
    class Buffer(Protocol):
```

Or type checkers won't let users call isinstance() against it

tests/stubtest_allowlists/py3_common.txt
```diff
@@ -774,3 +774,7 @@ pkgutil.ImpLoader.get_source
 pkgutil.ImpLoader.is_package
 pkgutil.ImpLoader.load_module
 pkgutil.ImpLoader.source
+
+# We lie about the existence of these methods
+.*.__buffer__
+.*.__release_buffer__
```
AlexWaygood, line 780: Not thrilled about this, as it means our CI won't fail if we have a typo in a `__buffer__` method that actually exists at runtime. But I guess I can live with it!

AlexWaygood (approved):

tests/stubtest_allowlists/py3_common.txt
JelleZijlstra: Would be nice if the stubtest allowlist would only silence issues about the method not existing, not issues about the signature if it does exist.

AlexWaygood: I guess what we're saying is we're both +1 for https://github.com/python/mypy/issues/13703 :)

PR #10226, opened 2023-05-27
Fix groups type in argparse.HelpFormatter

hamdanal:
The `groups` argument to the `add_usage` method must be an iterable of `_MutuallyExclusiveGroup` instead of `_ArgumentGroup`.

This can be seen here:
- https://github.com/python/cpython/blob/6e62eb2e70a9f2a8099735989a58e8c8cfb4a2f2/Lib/argparse.py#L419
- https://github.com/python/cpython/blob/6e62eb2e70a9f2a8099735989a58e8c8cfb4a2f2/Lib/argparse.py#L1858-L1859
- https://github.com/python/cpython/blob/6e62eb2e70a9f2a8099735989a58e8c8cfb4a2f2/Lib/argparse.py#L2585-L2586
- https://github.com/python/cpython/blob/6e62eb2e70a9f2a8099735989a58e8c8cfb4a2f2/Lib/argparse.py#L2593-L2594


srittau (approved):
Thanks!

PR #10228, opened 2023-05-28
Change the `asyncio.Task` constructor type

sobolevn:
The signature of `Task` was changed in `3.9`
Now, only `Coroutine` objects are supported: https://github.com/python/cpython/blob/3.9/Lib/asyncio/tasks.py#L121-L125

Before `3.9` is was possible to send in any awaitable objects, see: https://github.com/python/cpython/blob/3.8/Lib/asyncio/tasks.py#L98

Refs https://github.com/python/typeshed/issues/10116


sobolevn:
@kumaraditya303 @AlexWaygood friendly ping :)

AlexWaygood:
I looked at this a bit the other day, and I wasn't able to come up with any examples on Python 3.7/3.8 of cases where a custom awaitable (that wasn't a coroutine or a generator) was accepted by the `Task` constructor. Are you able to come up with any?

kumaraditya303:
See https://github.com/python/typeshed/issues/10236

PR #10229, opened 2023-05-28
Use a Protocol as the argument for io.TextIOWrapper

JelleZijlstra:
I ran into a false positive where gzip.GzipFile couldn't be assigned
to io.TextIOWrapper.


JelleZijlstra:
The Protocol is based on reading https://github.com/python/cpython/blob/main/Modules/_io/textio.c, specifically how it uses `self->buffer`.

srittau:
Considering the (valid) primer hits: [The documentation](https://docs.python.org/3/library/io.html#io.TextIOWrapper) says:  "A buffered text stream providing higher-level access to a [BufferedIOBase](https://docs.python.org/3/library/io.html#io.BufferedIOBase) buffered binary stream." Would it make sense to use that type? I didn't investigate yet.

JelleZijlstra:
Looks like `gzip.GzipFile` inherits from `BufferedIOBase`, so that would at least fix my problem. I'll see what primer thinks.

JelleZijlstra:
OK, that's worse for mypy-primer.

srittau:
Looking at the primer hits:

1. pytest #1: Same problem as before. Maybe when we get typevar defaults, we can make `TextIOWrapper` generic.
2. pytest #2: A real problem. The result of `tempfile.TemporaryFile()`, which is an `typing.IO` object, is passed to `TextIOWrapper`. We should probably tighten our return type.
3. pytest #3: Test code, ignoring.
4. pylint: The result of `TextIOWrapper.detach()` is passed in, which is not defined in our stubs. The base class's `TextIOBase.detach()` returns a `BinaryIO`. In the implementation `TextIOBase` is not implemented, but `TextIOWrapper` is, hinting that we could tighten the return type in the stubs as well.
5. rich: Passes in a custom `_Reader` class, which derives from `RawIOBase`. Points to a genuine problem, at least going by the Python docs.
6. urllib3: Needs to update its explicit type annotation.
7. black: Passes in `sys.stdout.buffer`. Would need an assertion, which can't hurt in this case.
8. cwltool #1: See black.
9. cwltool #2: Would need to update its cast, but points to a potentially real problem.

Overall I would say that we should fix the other two problems in typeshed first and then do this change. It's a bit disruptive, but I think overall worth it for finding more potential problems and further cleanup of the I/O types.

srittau:
Actually, the `TemporaryFile()` annotation were still incorrect. #10284.

srittau:
To recap the primer output:

## Category 1: Passes in incompatible type

These seem to work at runtime, but still I think that a type checker pointing out that this could be problematic is actually an advantage. Explicitly casting away the problem would acknowledge this in the implementation.

* pytest no 2 (src/_pytest/capture.py:466)
* rich

## Category 2: Passes in `sys.stdout.buffer

These have the usual "std" types problems. Using an explicit assert on `sys.stdout`'s type would be a good idea in any case.

* black
* cwltool no 2 (cwltool/main.py)

## Category 3: Acknowledged as problematic in the source code

I see adding a cast here as acceptable.

* pytest no 1 (src/_pytest/capture.py:174)

## Category 4: Needs to update its explicit type annotation/cast

* urllib3
* cwltool no 1 (cwltool/cwlprov/writablebagfile.py)

As written before, despite the primer output I think this change is worthwhile as is, for the reasons outlined above.

srittau (dismissed):

JelleZijlstra:
I think I'd prefer to go back to a protocol, even if the documentation explicitly calls for a `BufferedIOBase`. The runtime implementation doesn't care about the nominal type, so I think we should also use a protocol.

srittau:
I consider this solution worse. For example, the primer hit at src/_pytest/capture.py:546 is a valid use that's now rejected. While this particular instance can be fixed by adjusting the protocol, we can't fix all valid uses without the protocol actually mimicking `BufferedIOBase`.

The best solution would be to make `TextIOWrapper` generic, bound to the protocol and defaulting to `BufferedIOBase`, but this can't be implemented, until we have type var defaults.

JelleZijlstra:
Just wanted to confirm the primer output was still the same, closing in favor of #11420.

PR #10232, opened 2023-05-30
Tighten tempfile() return types

srittau:


srittau:
A first stab. We also have to split `_TextMode` and `_BinaryMode`, similar to how they are split in `builtins.pyi`. We probably don't need to handle the read case, as a read-only temp file makes no sense. This could even help catch further bugs.

stdlib/tempfile.pyi
```diff
@@ -30,13 +30,14 @@ TMP_MAX: int
 tempdir: str | None
 template: str
 
-_StrMode: TypeAlias = Literal["r", "w", "a", "x", "r+", "w+", "a+", "x+", "rt", "wt", "at", "xt", "r+t", "w+t", "a+t", "x+t"]
-_BytesMode: TypeAlias = Literal["rb", "wb", "ab", "xb", "r+b", "w+b", "a+b", "x+b"]
+_TextMode: TypeAlias = Literal["r", "w", "a", "x", "r+", "w+", "a+", "x+", "rt", "wt", "at", "xt", "r+t", "w+t", "a+t", "x+t"]
```
AlexWaygood, line 33: Does `_typeshed.OpenTextModeUpdating` not work here?

srittau: Probably, I want to look into reusing the literals in `builtins.pyi`.

srittau:
Depends on #10233.

srittau:
This is the last prerequisite for trying #10229 again.

PR #10233, opened 2023-05-31
[aiofiles] Use file mode alias from _typeshed

srittau:


JelleZijlstra:
This allows a lot more modes.

srittau:
Which should be ok, since the `mode` parameter is just passed on to `open()`. We could argue not to allow read-only modes, but those were allowed previously, and I wanted to avoid semantic changes for now.

PR #10238, opened 2023-06-01
fix psutil.cpu_percent with percpu

pjknkda:
Fix `psutil.cpu_percent` when `percpu` is set to True.


stubs/psutil/psutil/__init__.pyi
```diff
@@ -231,7 +231,12 @@ def wait_procs(
 ) -> tuple[list[Process], list[Process]]: ...
 def cpu_count(logical: bool = True) -> int: ...
 def cpu_times(percpu: bool = False): ...
-def cpu_percent(interval: float | None = None, percpu: bool = False) -> float: ...
+@overload
+def cpu_percent(interval: float | None = None, percpu: Literal[False] = False) -> float: ...
+@overload
+def cpu_percent(interval: float | None, percpu: Literal[True]) -> list[float]: ...
+@overload
+def cpu_percent(interval: float | None = None, *, percpu: Literal[True]) -> list[float]: ...
```
srittau, line 239: Thanks! This can be slightly simplified, since the case where `interval` is specified is already covered by overload 2:

```suggestion
@overload
def cpu_percent(*, percpu: Literal[True]) -> list[float]: ...
```

srittau (changes requested):

srittau (approved):

PR #10239, opened 2023-06-02
Unpin micro-versions

sobolevn:
I think that it should be good right now, since new versions were released since Feb and GitHub actions now pick them up, example: https://github.com/python/typeshed/actions/runs/5144937782/jobs/9261882339#step:3:24

Refs https://github.com/python/typeshed/pull/9742


sobolevn:
Looks like `pdb.getsourcelines` was removed in https://github.com/python/cpython/commit/d4f8fafd2df3a6a1828c96f9347a14595ec438d1 as a bugfix, so I will add it to `ignorelist` 👍

stdlib/pdb.pyi
```diff
@@ -168,7 +167,6 @@ class Pdb(Bdb, Cmd):
 def find_function(funcname: str, filename: str) -> tuple[str, str, int] | None: ...
 def main() -> None: ...
 def help() -> None: ...
-def getsourcelines(obj: _SourceObjectType) -> tuple[list[str], int]: ...
```
AlexWaygood, line 171: Why are we deleting this entirely? Some users might be depending on this being in the stub. Can't we just put the definition inside an `if sys.version_info < (3, 10)` branch?

tests/stubtest_allowlists/py310.txt
```diff
@@ -27,6 +27,7 @@ ipaddress.IPv6Interface.hostmask
 ipaddress._BaseNetwork.broadcast_address
 ipaddress._BaseNetwork.hostmask
 multiprocessing.spawn._main
+# pdb.getsourcelines  # Undocumented helper, was removed in 3.10.11
```
AlexWaygood, line 30: I don't understand what this comment is for ☹️

AlexWaygood (approved):

PR #10243, opened 2023-06-02
Update some dependencies

JelleZijlstra:


JelleZijlstra:
Problem:

```
ERROR: Cannot install -r requirements-tests.txt (line 12), -r requirements-tests.txt (line 2) and pathspec>=0.10.3 because these package versions have conflicting dependencies.

The conflict is caused by:
    The user requested pathspec>=0.10.3
    black 23.3.0 depends on pathspec>=0.9.0
    pycln 2.1.4 depends on pathspec<0.10.0 and >=0.9.0
```

We explicitly want `pathspec>=0.10.3`, but pycln doesn't want it. Commented on this at hadialqattan/pycln#196. For now I'll revert to pycln 2.1.3.

AlexWaygood (approved):

PR #10244, opened 2023-06-03
py312: Add `delete=True` parameter to `tempfile.TemporaryDirectory`

sobolevn:
Source: https://github.com/python/cpython/blob/3.12/Lib/tempfile.py#L864


AlexWaygood (approved):
`NamedTemporaryFile` and `_TemporaryFileWrapper` also get a new `delete_on_close` parameter in Python 3.12, but we can add these in a separate PR.

PR #10245, opened 2023-06-03
Update sys.pyi for Python3.12

sobolevn:
Source: https://github.com/python/cpython/blob/3.12/Python/sysmodule.c
Docs: https://github.com/python/cpython/blob/3.12/Doc/library/sys.rst


hauntsaninja:
(Let's merge in #10253 first)

JelleZijlstra:
Looks like you get to remove some allowlist entries :)

stdlib/sys.pyi
```diff
@@ -359,3 +359,10 @@ if sys.version_info < (3, 8):
 # as part of the response to CVE-2020-10735
 def set_int_max_str_digits(maxdigits: int) -> None: ...
 def get_int_max_str_digits() -> int: ...
+
+if sys.version_info >= (3, 12):
+    def getunicodeinternedsize() -> int: ...
+    if sys.platform == "linux":
+        def activate_stack_trampoline(__backend: str) -> None: ...
+        def deactivate_stack_trampoline() -> None: ...
+        def is_stack_trampoline_active() -> bool: ...
```
AlexWaygood, line 368: According to stubtest, these exist on all platforms, not just Linux

sobolevn: Ok, here's what is going on:

```python
>>> import sys
>>> sys.activate_stack_trampoline('perf')
ValueError: perf trampoline not available
```

Source: https://github.com/python/cpython/blob/3.12/Python/sysmodule.c#L2166-L2168

I think that I can add a `def activate_stack_trampoline(__backend: str) -> NoReturn` for non-linux versions.

AlexWaygood: I'd actually be okay with keeping the sys.version_info branch and adding allowlist entries on Windows/Darwin in that case! (We should add a comment to the allowlist saying why they're not included, though).

sobolevn: I think that adding it is better, users might do something like:

```python
try:
   sys.activate_stack_trampoline('perf')
except ValueError:
   some_other_case()
```

And it is quite easy to do with conditional definition.

AlexWaygood (approved):
Thanks! LGTM

stdlib/sys.pyi
AlexWaygood: Yeah, makes sense!

Issue #9735, opened 2023-02-14
cffi: FFI.string invalid signature


senyai:
https://github.com/python/typeshed/blob/140bba3425229a69cf4d5ccf7f5348ff85f66c16/stubs/cffi/cffi/api.pyi#L42

I know that `maxlen` agument [is optional](https://cffi.readthedocs.io/en/latest/ref.html#ffi-string-ffi-unpack).

senyai:
After installing `python -m pip install types-cffi --user` (1.15.1.15) I got this file structure in site-packages:
```types_cffi-1.15.1.5.dist-info/RECORD
types_cffi-1.15.1.5.dist-info/INSTALLER
types_cffi-1.15.1.5.dist-info/WHEEL
types_cffi-1.15.1.5.dist-info/METADATA
types_cffi-1.15.1.5.dist-info/top_level.txt
types_cffi-1.15.1.5.dist-info/REQUESTED
_cffi_backend-stubs/__init__.pyi
_cffi_backend-stubs/METADATA.toml
cffi-stubs/recompiler.pyi
cffi-stubs/error.pyi
cffi-stubs/__init__.pyi
cffi-stubs/lock.pyi
cffi-stubs/ffiplatform.pyi
cffi-stubs/vengine_gen.pyi
cffi-stubs/model.pyi
cffi-stubs/cparser.pyi
cffi-stubs/api.pyi
cffi-stubs/commontypes.pyi
cffi-stubs/cffi_opcode.pyi
cffi-stubs/backend_ctypes.pyi
cffi-stubs/setuptools_ext.pyi
cffi-stubs/vengine_cpy.pyi
cffi-stubs/verifier.pyi
cffi-stubs/pkgconfig.pyi
cffi-stubs/METADATA.toml
```

where `cffi-stubs/api.pyi` has the right signature and `_cffi_backend-stubs/__init__.pyi` has the wrong one: `def string(self, cdata: CData, maxlen: int) -> bytes | str: ...` What is going on?

srittau:
This looks to be a bug in `stubs/cffi/_cffi_backend.pyi`. PR welcome!

PR #10253, opened 2023-06-04
Run stubtest for Python 3.12

hauntsaninja:


stdlib/builtins.pyi
```diff
@@ -53,7 +53,7 @@ from typing import (  # noqa: Y022
     overload,
     type_check_only,
 )
-from typing_extensions import (
+from typing_extensions import (  # type: ignore
```
hauntsaninja, line 56: These type ignores look like some pretty weird mypy bugginess surrounding LiteralString. I thought probably makes sense to just ignore for now and figure it out later, so we can test 3.12 effectively

JelleZijlstra (approved):

tests/stubtest_allowlists/py310.txt
```diff
@@ -183,3 +194,17 @@ typing.NewType.__call__
 typing_extensions\.Final
 typing\.NamedTuple
 typing\.Annotated
+
+# We lie about the existence of these methods
+.*.__buffer__
+.*.__release_buffer__
+
+# Removed in 3.12
+distutils\..*
```
AlexWaygood, line 203: I would have preferred to not do this. I feel like whenever we've used wildcard regex allowlist entries in the past, it's ended up causing us pain and confusion further down the line (https://github.com/python/typeshed/pull/9900; https://github.com/python/typeshed/pull/10205; there have been others as well).

If we _have_ to do this, then we may as well revert the changes that were made to `stubtest_stdlib.py` in #9734. The `venv` module is really slow on Windows, so those changes basically doubled the time it takes for me to run the test locally.

hauntsaninja: I agree in general, but I have basically no further interest in the accuracy of the distutil stubs, especially since most users aren't even going to get distutils when they `import distutils`. So that points to undo #9734. But I definitely wouldn't say "we _have_ to do this"... let me know which one you want and I'll make the PR.

AlexWaygood: Yeah, that all makes sense :-) let's go for undoing #9735, then.

AlexWaygood: Longer term, I guess we might want to look at getting type checkers to somehow recognise that if a user does `import distutils` with `setuptools` installed, they'll actually be getting `setuptools._distutils`... but the idea of adding that kind of special-casing to type checkers *really* doesn't fill me with joy.....

stdlib/builtins.pyi
AlexWaygood: Were you using mypy `master` when you ran tests locally for this PR? This looks like the regression I reported in https://github.com/python/mypy/pull/15161#issuecomment-1531596612

hauntsaninja: Ah I'd missed that, it's definitely the same thing. Hmm, I'm not sure what the best way to proceed is.

PR #10254, opened 2023-06-04
Update pyright to 1.1.311

hauntsaninja:


hauntsaninja:
See if it magically helps with #10216, otherwise I'll tag Eric

PR #10257, opened 2023-06-04
`pathlib`: `is_mount()` is implemented on Windows on py312+

AlexWaygood:
It was implemented in https://github.com/python/cpython/commit/29650fea9605bf1f48320487c6d5d6d70d97ad95

### DIFF
--- a/stdlib/pathlib.pyi
+++ b/stdlib/pathlib.pyi
@@ -160,7 +160,8 @@ class Path(PurePath):
         def owner(self) -> str: ...
         def group(self) -> str: ...
 
-    # This methods does "exist" on Windows, but always raises NotImplementedError on <3.12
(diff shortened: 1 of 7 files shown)


stdlib/pathlib.pyi
```diff
@@ -159,6 +159,9 @@ class Path(PurePath):
         # so it's safer to pretend they don't exist
         def owner(self) -> str: ...
         def group(self) -> str: ...
+
+    # This methods does "exist" on Windows, but always raises NotImplementedError on <3.12
```
AlexWaygood, line 163:
```suggestion
    # This method does "exist" on Windows on <3.12, but always raises NotImplementedError
    # On py312+, it works properly on Windows, as with all other platforms
```

hauntsaninja (approved):
Great!

PR #10259, opened 2023-06-04
Remove venv for stubtest stdlib

hauntsaninja:
Requested by Alex in #10253


AlexWaygood (approved):
Looks great, thank you!

PR #10261, opened 2023-06-04
pathlib: more changes for py312

hauntsaninja:


stdlib/pathlib.pyi
```diff
@@ -70,6 +74,9 @@ class PurePath(PathLike[str]):
     if sys.version_info >= (3, 9) and sys.version_info < (3, 11):
         def __class_getitem__(cls, type: Any) -> GenericAlias: ...
 
+    if sys.version_info >= (3, 12):
+        def with_segments(self: Self, *args: StrPath) -> Self: ...
```
AlexWaygood, line 78: nit:

```suggestion
        def with_segments(self, *args: StrPath) -> Self: ...
```

AlexWaygood (approved):

stdlib/pathlib.pyi
hauntsaninja: Oh huh, I thought I tried that, ran into an error and removed it thinking it was an old-style Self (without checking). But I must have messed something up.

PR #10263, opened 2023-06-04
Make stubtest work better on macOS

hauntsaninja:
Apparently my laptop is not the same as the Github Actions runners

### DIFF
--- a/tests/stubtest_allowlists/py3_common.txt
+++ b/tests/stubtest_allowlists/py3_common.txt
@@ -424,8 +424,8 @@ typing\.AsyncGenerator
 typing\.AbstractSet
 
 # Internal attributes
-.*\.__protocol_attrs__
-.*\.__callable_proto_members_only__
+# .*\.__protocol_attrs__
+# .*\.__callable_proto_members_only__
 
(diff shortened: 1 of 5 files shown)


tests/stubtest_allowlists/py3_common.txt
```diff
@@ -424,8 +424,8 @@ typing\.AsyncGenerator
 typing\.AbstractSet
 
 # Internal attributes
-.*\.__protocol_attrs__
-.*\.__callable_proto_members_only__
+typing_extensions\..*\.__protocol_attrs__
+typing_extensions\..*\.__callable_proto_members_only__
```
AlexWaygood, line 428: Not sure there's much point worrying about these too much, since the next mypy release will include https://github.com/python/mypy/commit/b8f8d800fc1d72452af67aa0ae64c48bfacb0c92

hauntsaninja: Yeah, I realised this was because I had latest mypy installed

AlexWaygood (approved):
Haven't got any way of testing, but I trust that this is correct!

PR #10266, opened 2023-06-06
Tighten TextIOWrapper buffer types

srittau:
This matches the documentation, which says:

> A buffered text stream providing higher-level access to a BufferedIOBase
> buffered binary stream.

Cf #10229

### DIFF
--- a/stdlib/io.pyi
+++ b/stdlib/io.pyi
@@ -147,7 +147,7 @@ class TextIOBase(IOBase):
 class TextIOWrapper(TextIOBase, TextIO):  # type: ignore[misc]  # incompatible definitions of write in the base classes
     def __init__(
         self,
-        buffer: IO[bytes],
+        buffer: BufferedIOBase,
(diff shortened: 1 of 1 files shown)


srittau:
Also, the current annotations are unsound. You can pass in any `IO[bytes]`, but are guaranteed to get back a `BinaryIO` from `.buffer` and `.detach()`, although in reality you just get back what you passed in.

AlexWaygood:
I'd be interested to see the mypy_primer results if we tried making it generic. It might be very disruptive, but it also might not.

srittau:
Anyway, the merge order is #10233 -> #10232 -> potentially this or #10229

srittau:
I suggest we merge the two other PRs first, rerun the tests here and then I'll also do a test with a generic `TextIOWrapper`.

srittau:
I merged the additional changes into #10229, closing this here.

PR #10267, opened 2023-06-06
Make all tests executable

srittau:
This allows tests to be executed using `./tests/...`, without needing to type `python/tests/...`. It's also useful for bash completion.


AlexWaygood (approved):
I think something weird happened when I edited some of these on Windows at some point, and it caused a regression here. I've no idea what I did, or how to stop it happening again, but thanks for fixing!

AlexWaygood:
I wonder if there's a pre-commit hook we could add to stop this happening...

srittau:
I don't think it's terribly important. In CI we use "python" explicitly and it's easy enough to spot and work around locally.

PR #10269, opened 2023-06-06
Update functools.pyi for Python3.12

sobolevn:
I am not sure that my `_assigned` hack is correct, another option is to use `if` for the whole function (but, I don't like this idea).


AlexWaygood:
> another option is to use `if` for the whole function (but, I don't like this idea).

Why aren't you a fan of that? I'll grant that it's annoyingly repetitive, but it's no more repetitive than when we have to repeat a whole function across two `sys.version_info` branches just because a new keyword-only argument was added at the end.

AlexWaygood (approved):
Thanks!

PR #10270, opened 2023-06-07
Add stubs for `inifile`

dairiki:
Here are stubs for `inifile` ([github](https://github.com/mitsuhiko/python-inifile), [pypi](https://pypi.org/project/inifile/)).

It's old and not particularly widely used, but we use it in [Lektor](https://github.com/lektor/lektor) and its plugins and are starting to work on type annotations there.


stubs/inifile/@tests/stubtest_allowlist.txt
```diff
@@ -0,0 +1,8 @@
+# These are internal use and python 2 compatibility variables and functions
```
JelleZijlstra, line 1: good call on allowlisting those

JelleZijlstra (approved):

PR #10273, opened 2023-06-07
inifile: add missing attributes

dairiki:
This adds a few attributes that a missed on the first go-round (#10270):

- IniData.dialect
- IniFile.filename
- IniFile.encoding
- IniFile.is_new

It also narrows the type of the `filename` parameter to `IniFile` from `StrOrBytesPath | int` to `StrPath`:
- The `| int`  was erroneous: passing a file descriptor will not work.
- The narrowing from `StrOrBytesPath` to `StrPath` is not strictly correct, however:
   - `StrPath` is the common and probably the intended usage. (The `AppIniFile` subclass only accepts `StrPath`.)
   - It avoids the need to convert `IniFile` to a generic (which would be required to get the type of `IniFile.filename` correct).


stubs/inifile/inifile.pyi
```diff
@@ -99,7 +100,10 @@ class IniData(MutableMapping[str, str]):
     def __delitem__(self, name: str) -> None: ...
 
 class IniFile(IniData):
-    def __init__(self, filename: StrOrBytesPath | int, encoding: str | None = ..., dialect: Dialect | None = ...) -> None: ...
+    def __init__(self, filename: StrPath, encoding: str | None = ..., dialect: Dialect | None = ...) -> None: ...
+    filename: str
```
AlexWaygood, line 104: Shouldn't this be the same type as the `__init__` parameter?

```suggestion
    filename: StrPath
```

dairiki: @AlexWaygood Maybe.  `Filename` does get coerced to a `str` [here](https://github.com/mitsuhiko/python-inifile/blob/dbdf3f319e6ef76dfa70b536defe916286a84f78/inifile.py#L500).

That brings up a somewhat related question:

Probably, `IniFile.filename` (as well as most or all of the other class attributes annotated in this module) should be treated as read-only.  To reflect that, I tried adding them as `@property`s, but `stubtest` didn't like that.
What's the best practice in this case?

AlexWaygood: > @AlexWaygood Maybe. `Filename` does get coerced to a `str` [here](https://github.com/mitsuhiko/python-inifile/blob/dbdf3f319e6ef76dfa70b536defe916286a84f78/inifile.py#L500).

Oh, you're quite right! Sorry, I misread some code somewhere.

> To reflect that, I tried adding them as `@property`s, but `stubtest` didn't like that.
> What's the best practice in this case?

If necessary, we can just add allowlist entries to tell stubtest to be quiet about these "white lies" :) In general, we try to keep allowlist entries to a minimum, but if there's a good reason to tell stubtest to shut up in a certain situation, it's an option.

For an example of a third-party stubs package in typeshed with an allowlist, see https://github.com/python/typeshed/blob/main/stubs/colorama/%40tests/stubtest_allowlist.txt

AlexWaygood (approved):
Happy to merge this if you're happy, but I'm also fine if you want to convert some attributes to properties and add allowlist entries

dairiki:
> Happy to merge this if you're happy, but I'm also fine if you want to convert some attributes to properties and add allowlist entries

Thank you.  I wasn't sure how acceptable is was to allowlist in that case.

I will convert to properties.  Stand by.

dairiki:
Okay, I think this is ready to merge.

AlexWaygood:
Thanks for the great commit message explaining why these attributes are being made read-only in the stub. That's going to be super helpful if we're ever digging through `git blame` in the future to try to figure out why we made this decision :)

dairiki:
@AlexWaygood  Thank you!

PR #10275, opened 2023-06-07
[psutil] add settable info property

### LINKED ISSUE #10195: Process has no attribute info - process_iter from psutil
`Proccess` object do not have `info` attribute, but code form `ps.py` is working fine.

**ps.py**
```python
from psutil import process_iter

def proc_is_running(name: str) -> int:
    for proc in process_iter(['pid', 'name']):
        if name in proc.info['name']:
            return proc.info['pid']
    return 0
```
**pip list**
```shell
mypy                      1.3.0
mypy-extensions           1.0.0
types-psutil              5.9.5.12
```
**mypy ps.py**
```shell
ps.py:6: error: "Process" has no attribute "info"  [attr-defined]
ps.py:7: error: "Process" has no attribute "info"  [attr-defined]
```

ringohoffman:
Update depending on https://github.com/giampaolo/psutil/pull/2267.

Fixes #10195


sobolevn (commented):
This would have to wait for https://github.com/giampaolo/psutil/pull/2267 to be released.

Refs https://github.com/python/typeshed/issues/10195

JelleZijlstra:
If I understand correctly, this attribute already sometimes exists. If so, I'd be OK with adding it to the stub with a comment, whether or not the corresponding runtime change gets accepted.

sobolevn (approved):
Yes, it is sometimes set here: https://github.com/giampaolo/psutil/blob/a2c543f95ce06f8cc4ee77061dd48c7eaaf402f8/psutil/__init__.py#L1438

I think that it is okay to add with a comment.

PR #10277, opened 2023-06-07
`datetime`: Remove unused stubtest allowlist entry for py312

sobolevn:
I don't quite understand what is the case with `__all__` here. So, I want to see the results.
Related https://github.com/python/cpython/pull/104920

This will be fixed automatically when this new commit by @hauntsaninja is released.


AlexWaygood:
Blocked by:
- https://github.com/actions/python-versions/pull/235

AlexWaygood:
We're using the latest beta of Python 3.12 in CI now, but the CI is still failing on this PR -- @sobolevn, looks like some updates might need to be made to `datetime.pyi`?

sobolevn:
Hm, I don't think so: 
1. Source: https://github.com/python/cpython/blob/3.12/Lib/datetime.py#L8-L9 `("date", "datetime", "time", "timedelta", "timezone", "tzinfo", "MINYEAR", "MAXYEAR", "UTC")`
2. Stubs: https://github.com/python/typeshed/blob/main/stdlib/datetime.pyi#L8 `("date", "datetime", "time", "timedelta", "timezone", "tzinfo", "MINYEAR", "MAXYEAR")` (no `"UTC"`)

Looks like https://github.com/python/cpython/pull/104920 was not backported to `3.12`.

AlexWaygood:
> Looks like [python/cpython#104920](https://github.com/python/cpython/pull/104920) was not backported to `3.12`.

I've triggered the backport on that PR, but I'll wait a little bit to check that Paul doesn't object before merging it :)

srittau:
The backport has now been merged, so this should be fixed in the next beta.

PR #10290, opened 2023-06-09
Update ctypes stubs for Python3.12

sobolevn:
Source: https://github.com/python/cpython/blob/3.12/Lib/ctypes/__init__.py#L572-L577

Basically, it depends on `SIZEOF_TIME_T` value to set a proper alias type.

### DIFF
--- a/stdlib/ctypes/__init__.pyi
+++ b/stdlib/ctypes/__init__.pyi
@@ -182,7 +182,7 @@ if sys.platform == "win32":
     class HRESULT(_SimpleCData[int]): ...  # TODO undocumented
 
 if sys.version_info >= (3, 12):
-    c_time_t: TypeAlias = c_int32 | c_int64  # noqa: Y042
+    c_time_t: type[c_int32 | c_int64]
 
 class py_object(_CanCastTo, _SimpleCData[_T]): ...
 class BigEndianStructure(Structure): ...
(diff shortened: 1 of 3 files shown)


JelleZijlstra:
Stubtest complains it's not a union at runtime. (There are also a bunch of other stubtest complains, maybe it picked up the new beta?)

sobolevn:
@JelleZijlstra you are right:
1. I've sent https://github.com/python/typeshed/pull/10291 to fix the CI for the new beta
2. It is not a `Union` on any concrete system, but it is a "meta-union", because on different systems it will be a different thing. So, I guess it is safe to allowlist this type

stdlib/ctypes/__init__.pyi
```diff
@@ -181,6 +181,9 @@ class c_bool(_SimpleCData[bool]):
 if sys.platform == "win32":
     class HRESULT(_SimpleCData[int]): ...  # TODO undocumented
 
+if sys.version_info >= (3, 12):
+    c_time_t: TypeAlias = c_int32 | c_int64  # noqa: Y042
```
AlexWaygood, line 185: Unless this is literally an instance of `types.UnionType` at runtime, wouldn't it be better to do something like

```suggestion
    c_time_t: type[c_int32 | c_int64]
```

?

AlexWaygood:
you now need to remove the unused allowlist entry for `c_time_t`

AlexWaygood:
Thanks!

PR #10291, opened 2023-06-09
Fix CI for the python@3.12.0-beta.2

sobolevn:
Refs https://github.com/python/typeshed/pull/10290


JelleZijlstra (approved):
On your PR I also saw some winapi failures, let's see if those reproduce.

AlexWaygood:
> On your PR I also saw some winapi failures, let's see if those reproduce.

Yeah, looks like we need to add 24 new entries to the `win32-py312.txt` allowlist. (Or we could just add the missing constants in this PR as well -- they're all `int`s.)

AlexWaygood (approved):

PR #10292, opened 2023-06-09
Improve the signature of `typing.NamedTuple.__init__`

AlexWaygood:
Both `typename` and `fields` should be positional-only. Here's what happens if you try to specify both as keyword arguments at runtime:

```pycon
>>> from typing import NamedTuple
>>> Point = NamedTuple(typename="Point", fields=[("x", int), ("y", int)])
TypeError: NamedTuple() missing 1 required positional argument: 'typename'
```

If you specify just `fields` as a keyword argument, there's no error at runtime. However, it probably doesn't do what you wanted it to do: the user gets a `NamedTuple` class that has a single field named `fields`:

```pycon
>>> Point = NamedTuple('Point', fields=[("x", int), ("y", int)])
>>> Point(1, 2)
TypeError: Point.__new__() takes 2 positional arguments but 3 were given
>>> Point(x=4)
TypeError: Point.__new__() got an unexpected keyword argument 'x'
>>> Point(fields=4)
Point(fields=4)
```

Additionally, the first overload shouldn't have a default for the `fields` argument. It's only okay to not specify `fields` if you specify keyword arguments. (Though we'll [probably be deprecating that as well](https://github.com/python/cpython/issues/105566).)


AlexWaygood:
Well, the reason why I originally wrote this PR is that pyright currently allows this, and interprets it (incorrectly) as specifying the fields of the named tuple class:

```py
Point = NamedTuple("Point", fields=[("x", int), ("y", int)])
```

But mypy has a strange feature where you can use bare `NamedTuple` as an annotation, and (from the primer results) it looks like this change interferes with that feature a little.

I feel like mypy is in the wrong here, though -- `NamedTuple.__init__` shouldn't be interpreted as the `__init__` method of the base class of all named tuples. That's not how named tuples work.

AlexWaygood:
Pydantic and koda-validate are also just doing very dynamic things with named tuples, which I don't think they should reasonably expect to type check honestly.

srittau (approved):
They are also marked as positional-only in the implementation:

https://github.com/python/cpython/blob/d4fa52934a282df51cff800eee5caeb94a229547/Lib/typing.py#L2756

PR #10294, opened 2023-06-09
Remove unnecessary union in the default type in .get() and .pop() methods

### LINKED ISSUE #10293: The type of __default= in dict.pop and dict.get doesn't look right
`dict.pop` is currently defined as 

https://github.com/python/typeshed/blob/18514235198637fd2f9a7a92084a73d630da8f17/stdlib/builtins.pyi#L1086-L1089

Note the union with two type parameters as a type of `__default=`. This type is ambiguous if `dict.pop` is called with a value of type `_VT`. For example, given

```
d: dict[str, int]
d.pop("foo")
```

a type checker can choose any value for `_T`, since `str` is a subtype of `str | _T` regardless of what `_T` is. 

It seems the union was introduced in 2fdcd2e9e8e51b138f03df8e0d96e88c4b280110, which unfortunately does not seem to give any reasoning for that. I think we should drop it and use `_T` instead in both `dict.pop` and `dict.get` (and similarly for `MutableMapping`), i.e.

```
class dict(Generic[_KT, _VT]):
   # ...
   @overload
   def pop(self, __key: _KT, __default: _T) -> _T: ...
```

superbobry:
Fixes #10293.


superbobry:
There are more usages in `stubs/` which I did not update because I'm not familiar with these libraries enough.

JelleZijlstra:
We won't be able to accept a change that introduces this many new errors. Might be worth looking into some of these to see if there's any pattern in the new errors that we can support better.

It would also be good to add some more test cases to cover `dict.get`; it's been a very tricky function to cover in the stubs.

superbobry:
My guess would be that these errors were previously masked by the overparameterized union type, but I will have a closer look and summarize some of the causes here.

superbobry:
The two failure in pypa/pip, for example, look reasonable:

* `backend_path` in https://github.com/pypa/pip/blob/2d168e6e13cf50bfe10a5cf63aa758799de942da/src/pip/_internal/pyproject.py#L162 needs a type annotation because a type checker cannot assign a type to `[]` used as a default. Prior to this change it had an option of choosing `Any`, since `_VT = Any` and thus `__default` is effectively `Any | _T`. 
* `hashes` in https://github.com/pypa/pip/blob/2d168e6e13cf50bfe10a5cf63aa758799de942da/src/pip/_internal/models/link.py#L266 have a similar problem, the default is an empty `{}` and `_VT = Any`.

I think a type checker could use `list[Any]` and `dict[Any, Any]` in these cases. Should I file a mypy bug?

superbobry:
Here is another example: PyCQA/isort

* `with_comments` expects `Optional[List[str]]` as the first argument in https://github.com/PyCQA/isort/blob/8f2cd6bd3249ee5439ba6559f5b7f60ad42aee15/isort/output.py#L328, yet the type of `comments` is `Union[Any, Tuple[()]]`, because `.pop()` is called with `()` as the default value. This looks like a genuine bug.
* `use_comments` in https://github.com/PyCQA/isort/blob/8f2cd6bd3249ee5439ba6559f5b7f60ad42aee15/isort/output.py#LL456C29-L456C41 needs a type annotation, because mypy cannot infer the type argument of a list literal.
* https://github.com/PyCQA/isort/blob/8f2cd6bd3249ee5439ba6559f5b7f60ad42aee15/isort/output.py#L666 is effectively the same. mypy does not allow `{}.pop("foo", ...)`, because it infers the type argument to `<nothing>`.

Should this be fixed in mypy?

JelleZijlstra:
> Should this be fixed in mypy?

Possibly, but you'll have to be more precise about what needs to change in mypy (and possibly other type checkers).

I haven't looked at the mypy-primer output in much detail, but it will be hard sell to merge this without some type checker changes that make the input on typed codebases less disruptive.

superbobry:
I think the necessary fix here is to specialize empty collections to `Any` if it is unable to otherwise infer their type. For example,

```
def test(d: dict[str, Any]) -> None:
  x = d.get("foo", {})
  reveal_type(x)
```

currently emits `Any` without this PR and `Any | dict[<unknown>, <unknown>]` with this PR. I think a more useful behavior (which is also consistent with PEP-484) is to emit `Any | dict[Any, Any]`.

Akuli:
> what do you think about the following process?

I like it in many cases, and in fact we already use it. Here is the list of everything that mypy currently reverts: https://github.com/python/mypy/blob/cfec71798175fcbf030d6d114750d6fac454b3c9/misc/sync-typeshed.py#L182-L184

However, these are all relatively small improvements, and the typeshed+mypy combination is usable even without the reverts.

Typeshed follows "practicality beats purity" in many situations. It just isn't practical to break the most commonly used dict methods in the most popular type checker, even if it would be the "correct" thing to do or if we aren't "responsible" for that. Mypy could easily revert it, but it would make things difficult for people who for whatever reason want to point mypy at a custom clone of typeshed, e.g. because the stubs in the latest mypy version are too old.

superbobry:
I see where you are coming from, but I hope we can disentangle the notion of practicality from any particular type checker. The popularity of mypy does not seem relevant here. 

`dict` is ubiquitous in Python code, so it is important to get this right. At least one of the failures described above is caused by a true type error, which was previously masked by the way `dict.get` is annotated. Surely, it isn't practical to keep the annotations which are known to produce false negatives in user code?

Note also that I do not propose for mypy to revert this change, but rather to find a reasonable way to accommodate it. For example, by implementing python/mypy#15433.

AlexWaygood:
FYI I think this would also have a very high impact on pyright's typechecking, not just mypy. You can see in our CI that this PR causes the job where we run pyright on our own test suite to fail, as pyright can no longer infer the returned type from various calls to `dict.get()`

superbobry:
Yes, `pyright` also flags occurrences of collection literals where the type arguments cannot be inferred from the surrounding code. This is not specific to `dict.get` and will fail, e.g. given

```
from typing import TypeVar
T = TypeVar("T")
def f(x: T) -> T: ...
def test():
  return f({})
```

eltoder:
Maybe this can be helped with an extra overload:
```python
    @overload
    def get(self, __key: _KT, __default: _VT) -> _VT: ...
    @overload
    def get(self, __key: _KT, __default: _T) -> _VT | _T: ...
```
Now if mypy can prove that the default has a type acceptable for `_VT`, it does not need to infer unknown type variables, so empty collection literals should work. I think this similar to how it works now.

Btw, do we really need the one argument overload for get? Can it just be
```python
    @overload
    def get(self, __key: _KT, __default: _T = None) -> _VT | _T: ...
```
We do need it for pop, but get and pop have different behavior when called with one argument.

AlexWaygood:
Superseded by #10501, which fixes the issue in a much less disruptive way

PR #10296, opened 2023-06-10
Update `http/__init__.pyi` for Python3.12

sobolevn:
I would prefer to leave `http.client` for later when https://github.com/python/cpython/pull/105628 is merged / rejected.

Source: https://github.com/python/cpython/blame/3.12/Lib/http/__init__.py#L34-L52


AlexWaygood (approved):

AlexWaygood:
Oops, screwed up the commit message a bit there...

sobolevn:
Btw, if you have time to check https://github.com/python/cpython/pull/105628 I would appreciate your review.

PR #10297, opened 2023-06-10
Add missing __eq__ to Mapping

Dreamsorcerer:
https://github.com/python/cpython/blob/main/Lib/_collections_abc.py#L831


stdlib/typing.pyi
```diff
@@ -631,6 +631,7 @@ class Mapping(Collection[_KT], Generic[_KT, _VT_co]):
     def keys(self) -> KeysView[_KT]: ...
     def values(self) -> ValuesView[_VT_co]: ...
     def __contains__(self, __key: object) -> bool: ...
+    def __eq__(self, other: object) -> bool: ...
```
Dreamsorcerer, line 634: I think mypy requires this to be `object -> bool`. I'm guessing it doesn't have any logic to deal with `Notimplemented`?
Otherwise, this should actually be `Mapping[Any, Any] -> bool` and `object -> NotImplemented`.

AlexWaygood: Type checkers don't understand `NotImplemented`. The way you've typed it is correct.

Dreamsorcerer: That's fine. I guess if they did understand it though, then they would be able to give an error when you try to compare Mapping to another object with `__eq__()`, realising that they can't be compared together (although in practice, I struggle to find an object that doesn't compare...).

AlexWaygood: The heuristics mypy uses for its overlapping-equality check turn out to be pretty good in practice, so I don't think that's a huge issue. Though I agree it can be somewhat surprising when writing stubs that adding `__eq__` methods with the exact same signature to subclasses can have an impact on typechecking

Dreamsorcerer:
The new error from mypy primer is because they've typed it with `Self[K]`, should be fine for them to type it as `object` like everything else:
https://github.com/ibis-project/ibis/blob/master/ibis/common/egraph.py#L101

The removed errors from streamlit are exactly the false positives that caused me to look at this (but, in a different library).

stdlib/typing.pyi
AlexWaygood, line 634: We have lots of classes in typeshed that we pretend inherit from `Mapping` in the stubs, even though they don't at runtime. This is because `Mapping` is an ABC not a protocol, so it's the only way to get type checkers to recognise these classes as being mappings.

A consequence of this is that a bunch of classes that fake-inherit from `Mapping` in the stubs will only allow the second argument to be passed as a positional argument. So I think we should mark it as positional-only in the stubs, even though that's not actually true at runtime, to reduce false negatives for those classes.

We already do a similar thing for lots of other methods in `typing.pyi` — these classes are very special in the typing system!

```suggestion
    def __eq__(self, __other: object) -> bool: ...
```

AlexWaygood (commented):
This looks good, just one comment — see above

stdlib/typing.pyi
AlexWaygood: (Also, to note, if type checkers _did_ understand `NotImplemented` it would be _very_ tedious for us, as _every_ BinOp method would have to be typed using overloads 😄)

AlexWaygood (approved):
Thanks!

PR #10298, opened 2023-06-10
Complete openpyxl's nested descriptors

Avasam:
Follow-up to #10001 , this time tackling descriptors in `stubs\openpyxl\openpyxl\descriptors\nested.pyi`. Similarly to the previous PR, this one is very test-driven (and lands on a lot of the same mypy-false-negatives, at least I can rely on pyright here, ref https://github.com/python/mypy/issues/13316 )


stubs/openpyxl/openpyxl/packaging/core.pyi
```diff
@@ -21,38 +22,38 @@ class QualifiedDateTime(NestedDateTime):
...
-        language: Incomplete | None = None,
+        creator: object = "openpyxl",
+        description: object = None,
+        identifier: object = None,
+        language: object = None,
         modified=None,
-        subject: Incomplete | None = None,
-        title: Incomplete | None = None,
+        subject: object = None,
+        title: object = None,
```
srittau, line 58: You have converted some `Incomplete | None` annotations to `object`. (Here and elsewhere.) I didn't check with the implementation, but it seems to me that supplying arbitrary objects to, e.g., `title` can't be right.

srittau (commented):
Thanks! Skimming these, everything LGTM, except one concern below.

stubs/openpyxl/openpyxl/packaging/core.pyi
Avasam: There are "String Convertible" descriptors that converts whatever value you give to a string, so they can really be given anything.

"Boolean convertible" do have some intended values (even if in theory you could pass anything, same reason we type params as boolean even if the truthyness is checked). But I haven't found such intended values for "String Convertible"

srittau (approved):

Avasam:
One question I had that I forgot about:
Should I make more Nested Descriptor versions of type aliases to alleviate the parameters and setters long unions? Like I did with `_NestedNoneSetParam`.

Maybe a `_NestedDescriptorParam[_SomeType | None]` instead of `_HasTagAndGet[_SomeType | None] | _SomeType | None`

srittau:
I personally like type aliases in those cases. But I'm not sure if that will really be an advantage for our users and not just unnecessary churn for us to change now.

PR #10299, opened 2023-06-11
fix: use correct base classes for classes in email.generator

hoefling:
I noticed that `email.feedparser.BytesFeedParser` and `email.generator.BytesGenerator` do not inherit from `email.feedparser.FeedParser` and `email.generator.Generator`, respectively, although they do in the actual stdlib module implementation. This PR amends that.

Update: reverted changes in `email.feedparser`, will try it again in a different PR.

### DIFF
--- a/stdlib/email/feedparser.pyi
+++ b/stdlib/email/feedparser.pyi
@@ -1,24 +1,25 @@
 from collections.abc import Callable
 from email.message import Message
 from email.policy import Policy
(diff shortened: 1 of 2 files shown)


hoefling:
Looks like this is related to https://github.com/python/typeshed/issues/3968.

AlexWaygood (approved):

PR #10300, opened 2023-06-11
Add pexpect stubs

OscarCreator:
I generate stubs for [pexpect](https://github.com/pexpect/pexpect) but I am not so sure about all types. 
I tried to replace some `Incomplete` by looking in the repo's source an tests but would like some help.


stubs/pexpect/pexpect/pty_spawn.pyi
```diff
@@ -0,0 +1,74 @@
+from _typeshed import Incomplete
+from collections.abc import Callable, Sequence
+from io import TextIOWrapper
+from os import _Environ
+from typing import AnyStr
```
srittau, line 5: `AnyStr` is not an alias for `str | bytes`, which is most likely what you want to use instead below.

stubs/pexpect/pexpect/pty_spawn.pyi
```diff
@@ -0,0 +1,74 @@
...
+    def close(self, force: bool = True) -> None: ...
+    def isatty(self): ...
+    def waitnoecho(self, timeout: int = -1): ...
+    def getecho(self): ...
+    def setecho(self, state: bool): ...
+    def read_nonblocking(self, size: int = 1, timeout: int | None = -1): ...
+    def write(self, s: AnyStr) -> None: ...
+    def writelines(self, sequence: Sequence[AnyStr]) -> None: ...
+    def send(self, s: AnyStr): ...
+    def sendline(self, s: str = ""): ...
```
srittau, line 48:
```suggestion
    def sendline(self, s: str | bytes = ""): ...
```

stubs/pexpect/pexpect/_async.pyi
```diff
@@ -0,0 +1,19 @@
+import asyncio
+from _typeshed import Incomplete
+
+@asyncio.coroutine
+def expect_async(expecter, timeout: Incomplete | None = None): ...
+@asyncio.coroutine
+def repl_run_command_async(repl, cmdlines, timeout: int = -1): ...
```
srittau, line 7: Using `@coroutine` doesn't work anymore:

```suggestion
async def expect_async(expecter, timeout: Incomplete | None = None): ...
async def repl_run_command_async(repl, cmdlines, timeout: int = -1): ...
```

This should also get rid of the mypy and pyright failures.

stubs/pexpect/pexpect/pty_spawn.pyi
```diff
@@ -0,0 +1,74 @@
...
+    def flag_eof(self, value: bool) -> None: ...
+    def eof(self): ...
+    def terminate(self, force: bool = False): ...
+    status: int | None
+    exitstatus: bool | None
+    signalstatus: int | None
+    terminated: bool
+    def wait(self): ...
+    def isalive(self): ...
+    def kill(self, sig, int) -> None: ...
```
srittau, line 64:
```suggestion
    def kill(self, sig: int) -> None: ...
```

stubs/pexpect/pexpect/utils.pyi
```diff
@@ -0,0 +1,10 @@
+from _typeshed import Incomplete
+
+InterruptedError: Incomplete
```
srittau, line 3: I think we can just remove this for now. It's just a Python 2 compatibility stub in the implementation, which I'm not sure we need. Alternatively you could import it from builtins:

```python
from builtins import InterruptedError as InterruptedError
```

srittau (changes requested):
Thanks! I did not verify the types against the implementation, but most annotations looks fine. A few remarks below, though.

OscarCreator:
Hi, do you have any idea why I get `error: Returning Any from function declared to return "spawn"` from mypy_primer?
This is the code from pytest:
```python
def spawn(self, cmd: str, expect_timeout: float = 10.0) -> "pexpect.spawn":
        ...

        child = pexpect.spawn(cmd, logfile=logfile, timeout=expect_timeout)
        self._request.addfinalizer(logfile.close)
        return child # complain is here
```
Is `child` not the `spawn` object? Or am I thinking wrong here?

srittau:
The new primer hit in pytest is most likely due to the weird way that `pexpect` is imported:

```python
    def spawn(self, cmd: str, expect_timeout: float = 10.0) -> "pexpect.spawn":
        ...
        pexpect = importorskip("pexpect", "3.0")
        ...
        child = pexpect.spawn(cmd, logfile=logfile, timeout=expect_timeout)
        self._request.addfinalizer(logfile.close)
        return child
```

I.e., mypy has no idea that `pexpect` in this context is the `pexpect` module.

PR #10301, opened 2023-06-11
fix: use Self to ensure correct return type for EmailPolicy cloning

hoefling:
At the moment `email.policy.EmailPolicy.clone()` returns a `Policy` type, as can be reproduced with e.g.
```
$ mypy -c 'from email.policy import EmailPolicy; reveal_type(EmailPolicy().clone(utf8=True))'
<string>:1: note: Revealed type is "email.policy.Policy"
```
This differs at runtime:
```
$ python -c 'from email.policy import EmailPolicy; print(type(EmailPolicy().clone(utf8=True)))'
<class 'email.policy.EmailPolicy'>
```
This PR amends that.


AlexWaygood (approved):
Thanks!

PR #10304, opened 2023-06-12
Update zoneinfo for Python3.12

sobolevn:
Sources:
- https://github.com/python/cpython/blame/3.12/Lib/zoneinfo/__init__.py
- https://github.com/python/cpython/blame/3.12/Lib/zoneinfo/_zoneinfo.py


srittau (approved):

tests/stubtest_allowlists/py312.txt
```diff
@@ -259,6 +258,9 @@ posixpath.join
 ntpath.join
 os.path.join
 
+# Pos-only argument had different "names" in different Python versions:
```
sobolevn, line 261:
```suggestion
# Pos-only parameters had different "names" in different Python versions:
```

Issue #9836, opened 2023-03-03
Openpyxl typeshed errors


tritemio:
In an internal project, updating to 3.0.4.7 causes mypy in strict mode to generate the following error:

```python
ocra.py:120: error: Value of type "_WorkbookChild" is not indexable  [index]
ocra.py:124: error: "_WorkbookChild" has no attribute "iter_rows"  [attr-defined]
```

The code for line 120 is:

```python
...
wb = openpyxl.load_workbook(path)
ws = wb.active
header = [cell.value for cell in ws[1]]  # line 120
...
```

and for line 124:

```python
ws.iter_rows(min_row=2)  # line 124
```

The same code checked using  3.0.4.6 passes all mypy checks in strick mode.

It seems like  #9764 introduced a regression?

srittau:
#9764 actually made `load_workbook()` return a `Workbook`. Previously, it returned `Any`. The problem is that `wb.active` can return other workbook children (and even None), not only `Worksheet`s. The easiest way to work around this problem is to use:

```python
wb = openpyxl.load_workbook(path)
ws = wb.active
assert isinstance(ws, Worksheet)
header = [cell.value for cell in ws[1]]  # line 120
```

Cc @mynhardtburger

Avasam:
Even in #9511 I made `load_workbook` return a `Workbook`, the type comes from: https://foss.heptapod.net/openpyxl/openpyxl/-/blob/branch/3.1/openpyxl/reader/workbook.py#L31

tritemio:
Thanks for the quick answer. I'll apply the workaround to fix the code.

Maybe it could a good idea to write in the changelog that the type stubs are now more accurate and this may require changes to existing code bases that implicitly relied on wb.active to be `Any`. A link to the fix in this thread could also be useful.

srittau:
@tritemio Unfortunately, our changelogs are currently auto-generated from git commit messages, which is why they are a bit messy and we can't edit them retroactively.

Avasam:
Specifically for `wb.active`, "the Any trick" could be used? This was my first thought when I saw #11094

PR #10306, opened 2023-06-12
Fix openpyxl's workbook.pyi

e2t:
Preperty `active` returns `openpyxl.worksheet.worksheet.Worksheet` or None. This can be seen from the source code:

```python
    @property
    def active(self):
        """Get the currently active sheet or None

        :type: :class:`openpyxl.worksheet.worksheet.Worksheet`
        """
        try:
            return self._sheets[self._active_sheet_index]
        except IndexError:
            pass
```
_WorkbookChild does not have many of the attributes that Worksheet has.


srittau:
Thank you for your contribution. Unfortunately, according to the implementation (and unlike what the docstring says), `Workbook.active` can be any kind of workbook child: `Worksheet`, `WriteOnlyWorksheet`, or `Chartsheet`. See also #9836.

PR #10308, opened 2023-06-12
`openpyxl`: type `tagname`, `namespace` and `__doc__` ClassVar and params

Avasam:
`tagname` is always set as a `ClassVar`. The only moment it's not is when it's not been overridden, in which case it's a property that throws `NotImplementedError` but shouldn't make the class abstract.

`namespace` is set as a `ClassVar` (either as some literal string or None), except for `Nested` descriptors, where it it an optional param in `__init__`.


stubs/openpyxl/@tests/test_cases/check_nested_descriptors.py
```diff
@@ -32,7 +32,7 @@
 
 
 class WithDescriptors(Serialisable):
-    descriptor = Nested[str]()
+    descriptor = Nested(expected_type=str)
```
srittau, line 35: This can't really work as a type annotation, can it? (Same for the existing attributes below.)

Avasam: It"s not a type annotation, it's an actual assignement.

srittau (approved):
Good to go, once the merge conflicts are resolved.

PR #10309, opened 2023-06-12
Split requirements files into sections

srittau:
Also note that type checkers and linters should be pinned to a specific version.

See typeshed-internal/stub_uploader#96 for context.

### DIFF
--- a/requirements-tests.txt
+++ b/requirements-tests.txt
@@ -1,4 +1,6 @@
-aiohttp==3.8.4; python_version < "3.12"           # aiohttp can't be installed on 3.12 yet
+# Type checkers and other linters that we test our stubs against. These should always
+# be pinned to a specific version to make failure reproducible. See also the
+# "tool.typeshed" section in pyproject.toml for additional type checkers.
 black==23.3.0                                     # must match .pre-commit-config.yaml
 flake8==6.0.0; python_version >= "3.8"            # must match .pre-commit-config.yaml
 flake8-bugbear==23.5.9; python_version >= "3.8"   # must match .pre-commit-config.yaml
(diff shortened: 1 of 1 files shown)


AlexWaygood:
We'll need to get rid of the `requirements-txt-fixer` pre-commit hook if we want to do this 😄

requirements-tests.txt
```diff
@@ -8,6 +13,8 @@ isort==5.12.0; python_version >= "3.8"            # must match .pre-commit-confi
 mypy==1.3.0
 packaging==23.1
 pathspec>=0.11.1
+
+# Git support.
 pre-commit-hooks==4.4.0                           # must match .pre-commit-config.yaml
 pycln==2.1.5                                      # must match .pre-commit-config.yaml
 pytype==2023.6.2; platform_system != "Windows" and python_version < "3.11"
```
Avasam, line 20: Small comment: Given pycln is a formatter and pytype a type-checker, I would've expected to see them in the above section with black and mypy.

requirements-tests.txt
```diff
@@ -1,21 +1,30 @@
...
 isort==5.12.0; python_version >= "3.8"            # must match .pre-commit-config.yaml
 mypy==1.3.0
-packaging==23.1
-pathspec>=0.11.1
-pre-commit-hooks==4.4.0                           # must match .pre-commit-config.yaml
 pycln==2.1.5                                      # must match .pre-commit-config.yaml
 pytype==2023.6.2; platform_system != "Windows" and python_version < "3.11"
+
+# Git support.
+pre-commit-hooks==4.4.0                           # must match .pre-commit-config.yaml
```
AlexWaygood, line 15: I'd put this in the "linters" category -- it's only there so that people can run these hooks manually in their local environments, if they want to:

https://github.com/python/typeshed/blob/9ca472420429c36317f280cad9db81eb72e2f5ac/.pre-commit-config.yaml#L5-L12

requirements-tests.txt
Avasam: oh lol, I see.

Might be worth requesting "sort only within groups" as a feature. It feels like a common use-case (I myself am making groups in my requirements file and not using the fixer in my pre-commit for that exact reason).

Removing it atm seems reasonable.

PR #10312, opened 2023-06-12
Type all openpyxl constants

Avasam:
As title says. Has overlap with #9511 so this will reduce the number of changes in that PR.


srittau (approved):
Spot checked: LGTM

PR #10313, opened 2023-06-12
Type openpyxl incomplete dunders

Avasam:
As title says. A small and simple PR scoped to completing dunders types in openpyxl. Has overlap with #9511 so this will reduce the number of changes in that PR.


stubs/openpyxl/openpyxl/styles/alignment.pyi
```diff
@@ -14,7 +15,7 @@ vertical_aligments: tuple[_VerticalAlignmentsType, ...]
 
 class Alignment(Serialisable):
     tagname: str
-    __fields__: Incomplete
+    __fields__: ClassVar[tuple[str, ...]]
```
srittau, line 18: I can't seem to find `__fields__` in the implementation, looking at git HEAD. Same for `Side` and `Border` below.

Avasam: Tag 3.1.2 https://foss.heptapod.net/openpyxl/openpyxl/-/blob/3.1.2/openpyxl/styles/alignment.py#L21

Seems like it will disappear in 3.2

srittau:
There are some conflicts now, partly due to the removal/change to `__doc__`.

srittau (approved):

PR #10322, opened 2023-06-16
Type openpyxl `functions` module and its usages

Avasam:
Split off from, and improved over, #9511

- Typed openpyxl `functions` module and its usages
    Typing this module is kind of a real mess, in-code comment should explain it well. Since we can't really use any other xml libraries in these stubs anyway for various reasons (untyped, not a dependency, ...), `xml` is the only compatible one for typing openpyxl.
- Still I tried to use sensible Protocols when possible. For that reason I also moved some shared protocols from `nested.pyi`.
- Created `_functions_overloads.pyi` to abstract away most of the complexity and re-implement overloads of the different possible libraries.
- Updated the stubtest_allowlist


stubs/openpyxl/openpyxl/worksheet/_writer.pyi
```diff
@@ -1,15 +1,16 @@
...
 
 def create_temporary_file(suffix: str = ""): ...
 
 class WorksheetWriter:
     ws: Incomplete
-    out: Incomplete
+    out: IO[str] | str
     xf: Incomplete
-    def __init__(self, ws, out: Incomplete | None = None) -> None: ...
+    def __init__(self, ws, out: IO[str] | str | None = None) -> None: ...
```
srittau, line 13: We try to avoid `IO` etc. in typeshed if possible. Is it possible to use a protocol here?

Avasam: Yeah, the `close` and `write` methods are used. (Maybe more I'll double check usage in other parts of the code)
If the `out` parameter is None, it seems to default to a `BufferedWriter` and `IO` was the top parent with both methods.

Avasam: I simplified it down to `StrPath | str`, with an additional union to `BytesIO` because there's an explicit `if isinstance(self.out, BytesIO)` branch in `WorksheetWriter.read`

AlexWaygood: `StrPath` is a union that already includes `str`, no?

Avasam: uh, yes, lemme simplify that too real quick ^^

Avasam: Looking at the `et_xmlfile` implementation of `xmlfile`, if not passing directly a string, it needs the `close` and `write` method. But that one isn't restricted to `BytesIO`. Maybe I should just replace `BytesIO` in that union with a protocol that has both methods and fits BytesIO.

stubs/openpyxl/openpyxl/worksheet/_writer.pyi
```diff
@@ -1,15 +1,26 @@
...
+from typing_extensions import TypeAlias
+
+# WorksheetWriter.read has an explicit BytesIO branch. Let's make sure this protocol is viable for BytesIO too.
+class _SupportsCloseAndWrite(Protocol):
+    def write(self, __buffer: ReadableBuffer) -> Unused: ...
+    def close(self) -> Unused: ...
+
+# et_xmlfile.xmlfile accepts a str | _SupportsCloseAndWrite
+# lxml.etree.xmlfile should accept a StrPath | _SupportsClose https://lxml.de/api/lxml.etree.xmlfile-class.html
+_OutType: TypeAlias = _SupportsCloseAndWrite | StrPath
```
srittau, line 13: Yikes, that code is a mess in the original source code. So if I read this correctly, `WorksheetWriter.read()` only works with `BytesIO` and whatever can be passed to `open()` (str-like paths). `.cleanup()` work only with str-like paths. `.get_stream()` calls `xmlfile()`, where at least the lxml variant needs a "file-like object" or a string, according to the docs. And, if not supplied to the constructor, it's initialized to `NamedTemporaryFile()` (or its return value), which is neither a str-like path nor a `BytesIO`. Year, let's just use your solution for now.

srittau (approved):
LGTM, at least as a first stab. Original code is a mess, so let's see whether there are any problems in the future.

PR #10323, opened 2023-06-16
add shipping_rate type

SergioB-dev:
*DRAFT - Will do a couple more passes to make sure all requirements are met*

### DIFF
--- a/stubs/stripe/stripe/api_resources/shipping_rate.pyi
+++ b/stubs/stripe/stripe/api_resources/shipping_rate.pyi
@@ -1,0 +1,14 @@
+from _typeshed import Incomplete
+from typing import NoReturn
+
+from stripe import api_requestor as api_requestor
+from stripe.api_resources.abstract import (
+    CreateableAPIResource as CreateableAPIResource,
+    ListableAPIResource as ListableAPIResource,
+    UpdateableAPIResource as UpdateableAPIResource,
+)
+
+class ShippingRate(CreateableAPIResource, ListableAPIResource, UpdateableAPIResource):
+    OBJECT_NAME: str
+    @classmethod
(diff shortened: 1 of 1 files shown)


stubs/stripe/stripe/api_resources/shipping_rate.pyi
```diff
@@ -0,0 +1,14 @@
...
+from stripe.api_resources.abstract import (
+    CreateableAPIResource as CreateableAPIResource,
+    ListableAPIResource as ListableAPIResource,
+    UpdateableAPIResource as UpdateableAPIResource,
+)
+
+class ShippingRate(CreateableAPIResource, ListableAPIResource, UpdateableAPIResource):
+    OBJECT_NAME: str
+    @classmethod
+    def retrieve(cls, id, api_key: Incomplete | None = None, **params) -> NoReturn: ...
```
AlexWaygood, line 14: Where is this method defined? I can't see it in the source code: https://github.com/stripe/stripe-python/blob/69a9e91d784353ffb2d61644b1293d72082b9a88/stripe/api_resources/shipping_rate.py#L9-L14

Also, `-> NoReturn` means that the method always raises an Exception, or enters an infinite loop. Is that what you meant, or did you mean `-> None`? :)

SergioB-dev: > Where is this method defined?

So this method is defined for the class [APIResource](https://github.com/python/typeshed/blob/main/stubs/stripe/stripe/api_resources/abstract/api_resource.pyi) which this new class `ShippingRate` already inherits from. So adding it here was redundant. I've removed that method which also takes care of the second problem.

It was not my intention to always raise an Exception so thanks for the guidance there!

PR #10325, opened 2023-06-16
unittest.mock: use ParamSpec in patch

### LINKED ISSUE #10324: Using `mock.patch` as a decorator doesn't properly forward the underlying function signature (even with a `new=` argument)
The following seems to type check fine (both mypy and pyright) even though it should not (I would hope):

```py
from unittest.mock import Mock, patch

@patch("sys.exit", new=Mock())
def f(i: int) -> None:
    ...

f("a")
```

Tried with `types-mock==5.0.0.6`.

hauntsaninja:
Fixes #10324

### DIFF
--- a/stdlib/unittest/mock.pyi
+++ b/stdlib/unittest/mock.pyi
@@ -3,13 +3,14 @@ from collections.abc import Awaitable, Callable, Coroutine, Iterable, Mapping, S
 from contextlib import _GeneratorContextManager
 from types import TracebackType
 from typing import Any, Generic, TypeVar, overload
-from typing_extensions import Final, Literal, Self, TypeAlias
+from typing_extensions import Final, Literal, ParamSpec, Self, TypeAlias
 
 _T = TypeVar("_T")
 _TT = TypeVar("_TT", bound=type[Any])
 _R = TypeVar("_R")
 _F = TypeVar("_F", bound=Callable[..., Any])
 _AF = TypeVar("_AF", bound=Callable[..., Coroutine[Any, Any, Any]])
+_P = ParamSpec("_P")
 
 if sys.version_info >= (3, 8):
     __all__ = (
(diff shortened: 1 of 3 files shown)


stdlib/unittest/mock.pyi
```diff
@@ -234,7 +235,7 @@ class _patch(Generic[_T]):
     @overload
     def __call__(self, func: _TT) -> _TT: ...
```
srittau, line 236: I don't think we need this overload anymore as types should be callable. But can't hurt to add another test for that.

hauntsaninja: I think we still do, so that you can resolve attributes of the class. Note that this only matters when using pyright, since mypy sort of ignores class decorators. I'll add a test case though!

srittau (approved):

AlexWaygood:
Pyright now seems broken on `main` -- it looks like for some reason, pre-commit was the only check that was run on the last commit on this PR? (We didn't get any GitHub Actions on it for some reason -- I assume this was due to GitHub flakiness?)

AlexWaygood:
https://github.com/python/typeshed/actions/runs/5322153767/jobs/9638207422

srittau:
The error message:

> /home/runner/work/typeshed/typeshed/test_cases/stdlib/check_unittest.py:117:13 - error: "assert_type" mismatch: expected "() -> int" but received "() -> int"

This doesn't make much sense to me. @erictraut Could you have a look?

erictraut:
@srittau, this was a bug in pyright. It will be addressed in pyright 1.1.315, which I'll release this evening. The problem is that when pyright creates a type based on `Callable` it adds an implicit "positional-only separator" parameter to the end. It shouldn't do this if there are zero parameters specified.

tests/utils.py
AlexWaygood, line 1: Were the changes to this file just so that the stdlib is always tested first in CI?

(I don't object, just curious! :)

hauntsaninja: Yeah, I wanted stdlib tested first for lower latency local dev. Makes sense given that it's where most of our test cases are / probably a plurality of changes

AlexWaygood: Makes sense! You may already know this, but you can also run `python tests/regr_test.py stdlib` to run just the stdlib tests :-)

srittau:
Thanks again, @erictraut, for the super fast response!

PR #10327, opened 2023-06-18
Add _file in class SpooledTemporaryFile in tempfile.pyi

### LINKED ISSUE #10286: `tempfile.SpooledTemporaryFile` missing `_file` attribute
According to the python docs, `SpooledTemporaryFile._file` is an attribute: "The returned object is a file-like object whose _file attribute is either an [io.BytesIO](https://docs.python.org/3/library/io.html#io.BytesIO) or [io.TextIOWrapper](https://docs.python.org/3/library/io.html#io.TextIOWrapper) object (depending on whether binary or text mode was specified)"

So the `_file` attribute should be added to `tempfile.pyi` for the `SpooledTemporaryFile` class.

heysujal:
@Dr-Irv This is a draft  PR that attempts to close #10286
I have _file attribute in class SpooledTemporaryFile using hints given by @Akuli in #10286 
So, please provide feedback and let me know if there are any changes to make.
On further looking into documentation I found that there can also be a [true file object that needs to be returned depending on whether rollover() has been called](https://docs.python.org/3/library/tempfile.html?highlight=spooled#tempfile.SpooledTemporaryFile) 
I can add it later once I get feedback or might make another PR if required.

P.S. I ran the tests before making a PR and it was all good. But they are failing here for some reason.


stdlib/tempfile.pyi
```diff
@@ -547,6 +553,10 @@ class SpooledTemporaryFile(IO[AnyStr], _SpooledTemporaryFileBase):
     def tell(self) -> int: ...
     def truncate(self, size: int | None = None) -> None: ...  # type: ignore[override]
     @overload
+    def _file(self) -> io.BytesIO: ...
+    @overload
+    def _file(self) -> io.TextIOWrapper: ...
```
Akuli, line 558: This looks better, although it is missing two things:
- Specifying which of the two overloads a type checker should choose, similar to `write()` below
- The `@property` decorator

That said, I just realized it is not possible to use `@overload` and `@property` together like I imagined. This unfortunately means there's no way to say "this attribute is `BytesIO` for `SpooledTemporaryFile[bytes]` and `TextIOWrapper` for `SpooledTemporaryFile[str]`".

I think the best we can do is using `IO` with a normal attribute, not a property:

```suggestion
    _file: IO[AnyStr]
```

Here `IO[AnyStr]` means that it will be `IO[str]` for `SpooledTemporaryFile[str]`, and `IO[bytes]` for `SpooledTemporaryFile[bytes]`. This will be good enough for most use cases, but not ideal, because `IO` is not as specific as `TextIOWrapper`/`BytesIO`. For example, `some_string_tempfile._file.buffer` will not type check, because even though `TextIOWrapper` has a `buffer` attribute, the more general `IO[str]` does not have it and the type checker complains.

Akuli (commented):
This was trickier than I thought, but I figured it out :)

stdlib/tempfile.pyi
Akuli: Neat. I will accept this PR with a simple `_file: IO[str]` or with the descriptor workaround, since either way it is much better than not having a `_file` attribute at all :)

stdlib/tempfile.pyi
```diff
@@ -546,6 +546,7 @@ class SpooledTemporaryFile(IO[AnyStr], _SpooledTemporaryFileBase):
     def seek(self, offset: int, whence: int = ...) -> int: ...
     def tell(self) -> int: ...
     def truncate(self, size: int | None = None) -> None: ...  # type: ignore[override]
+    _file: IO[AnyStr]
```
Akuli, line 549: One more small nit: I would prefer this to be in the beginning of the class, similarly to the `name` attribute in `TemporaryDirectory` below.

Akuli (commented):
Thanks!

Akuli (approved):

PR #10329, opened 2023-06-18
inspect: getmodulename takes a PathLike

### LINKED ISSUE #10328: inspect.getmodulename accepts pathlib.Path
Consider this simple snippet:

```python
# mod.py
import inspect
import pathlib

inspect.getmodulename(pathlib.PurePosixPath('foo.py'))
```

Run it with Python 3.8 or later and it runs fine (I don't have Python 3.7 due to ARM).

Check it with mypy and it fails:

```
 draft $ mypy mod.py
mod.py:4: error: Argument 1 to "getmodulename" has incompatible type "PurePosixPath"; expected "str"  [arg-type]
```

Indeed the [definition](https://github.com/python/typeshed/blob/df795220a376ae9371afdc60dbf160392b93b6c2/stdlib/inspect.pyi#L180) indicates that only `str` is allowed for input.

The [implementation](https://github.com/python/cpython/blob/bc07c8f096791d678ca5c1e3486cb9648f7a027b/Lib/inspect.py#L936-L946) passes the input through to `os.path.basename`, which accepts `pathlib.Path`. Probably the typeshed should reflect this behavior, especially considering that the docs suggest that the input is a "path".

hauntsaninja:
Fixes #10328


stdlib/inspect.pyi
```diff
@@ -177,7 +178,7 @@ if sys.version_info >= (3, 11):
     @overload
     def getmembers_static(object: object, predicate: _GetMembersPredicate | None = None) -> _GetMembersReturn: ...
 
-def getmodulename(path: str) -> str | None: ...
+def getmodulename(path: PathLike[str]) -> str | None: ...
```
AlexWaygood, line 181: It also accepts `str`. We should use `_typeshed.StrPath`

```suggestion
def getmodulename(path: StrPath) -> str | None: ...
```

AlexWaygood (changes requested):

stdlib/inspect.pyi
hauntsaninja: Oh oof, thanks for the review!

jaraco:

There's code this will affect that's expected to land in `importlib_metadata` soon. Will that help?

AlexWaygood (approved):
LGTM

AlexWaygood:
> > According to [mypy_primer](https://github.com/hauntsaninja/mypy_primer), this change has no effect on the checked open source code. 🤖🎉
> 
> There's code this will affect that's expected to land in `importlib_metadata` soon. Will that help?

It's okay, we agree that `inspect.getmodulename` takes path-like objects; I think we're good to go on this :-)

hauntsaninja:
Thanks, I'll add importlib_metadata to the primer corpus! I was just a little sad my initial mistake wasn't flagged

jaraco:
> Thanks, I'll add importlib_metadata to the primer corpus! I was just a little sad my initial mistake wasn't flagged

Good point. Does that mean that there are no uses in the primer of `getmodulename(str)`? That's a little surprising. The `importlib_metadata` changes won't help with that as they'll be passing a subclass of `pathlib.PurePosixPath`.

jaraco:
Big thanks for the quick turnaround!

PR #10330, opened 2023-06-19
Rename Generator-like type params to be more obvious

Gobot1234:


Gobot1234:
Didn't realise how out of date my remote was hehe

AlexWaygood (commented):
The CI isn't running because there's quite a few merge conflicts — could you take a look?

stdlib/typing.pyi
```diff
@@ -467,19 +474,21 @@ class AsyncIterator(AsyncIterable[_T_co], Protocol[_T_co]):
...
         self, __typ: Type[BaseException], __val: BaseException | object = None, __tb: TracebackType | None = None
-    ) -> Awaitable[_T_co]: ...
+    ) -> Awaitable[_YieldT_co]: ...
     @overload
     @abstractmethod
-    def athrow(self, __typ: BaseException, __val: None = None, __tb: TracebackType | None = None) -> Awaitable[_T_co]: ...
+    def athrow(self, __typ: BaseException, __val: None = None, __tb: TracebackType | None = None) -> Awaitable[_YieldT_co]: ...
     def aclose(self) -> Awaitable[None]: ...
+    @abstractmethod
+    def __aiter__(self) -> AsyncGenerator[_YieldT_co, _SendT_contra]: ...
```
AlexWaygood, line 491: Why has this method been added?

stdlib/types.pyi
```diff
@@ -380,21 +384,21 @@ class AsyncGeneratorType(AsyncGenerator[_T_co, _T_contra]):
         @property
         def ag_suspended(self) -> bool: ...
 
-    def __aiter__(self) -> AsyncGeneratorType[_T_co, _T_contra]: ...
-    def __anext__(self) -> Coroutine[Any, Any, _T_co]: ...
-    def asend(self, __val: _T_contra) -> Coroutine[Any, Any, _T_co]: ...
+    def __aiter__(self) -> AsyncGeneratorType[_YieldT_co, _SendT_contra]: ...
```
AlexWaygood, line 387: Does this also just return self? If so, maybe:

```suggestion
    def __aiter__(self) -> Self: ...
```

AlexWaygood: Confirmed:

```pycon
>>> async def foo():
...   yield
...   
>>> f = foo()
>>> f.__aiter__() is f
True
```

stdlib/types.pyi
```diff
@@ -351,27 +348,31 @@ class ModuleType:
     # using `builtins.__import__` or `importlib.import_module` less painful
     def __getattr__(self, name: str) -> Any: ...
 
+_YieldT_co = TypeVar("_YieldT_co", covariant=True)
```
AlexWaygood, line 351: Our usual style is to put all the TypeVars at the top of the file, but I actually quite like way you've done it here, as the meaning of the TypeVars is easier to understand if it's closer to the definition. Will wait a little bit to see if any other maintainers have objections, though

AlexWaygood (approved):

stdlib/types.pyi
srittau: Speaking generally: I think type vars close to their use can make sense in more specific type vars, especially those with bounds or named `_FlooWidgetT` instead of just `_T`.

PR #10331, opened 2023-06-19
Add is_cli param to pydoc.pyi

Eclips4:
As introduced in [this PR](https://github.com/python/cpython/pull/102614), `pydoc.doc` and `Helper.help` now has an `is_cli` keyword parameter, which defaults to `False`.
It's has been backported to 3.11 branch, so I added a check for the version.


hauntsaninja (approved):
Thanks! We'll merge once new versions have been released and Github Actions picks them up

AlexWaygood:
Now that Python 3.12b3 has made its way to GitHub Actions, the _lack_ of this parameter is causing our CI to fail on `main` (e.g. https://github.com/python/typeshed/actions/runs/5334685391/jobs/9666793887?pr=10277). But it'll still be a while until we get a 3.11 patch release that includes this parameter.

For now, I'm going to edit this PR so that it just includes the parameter on py312+. We can change the stub so that the parameter is also included on py311 when we get a py311 release that includes the parameter :)

PR #10335, opened 2023-06-20
Replace static assertion that fails with pyright

AlexWaygood:
See https://github.com/python/typeshed/pull/10325#issuecomment-1598684045

### DIFF
--- a/test_cases/stdlib/check_unittest.py
+++ b/test_cases/stdlib/check_unittest.py
@@ -113,3 +113,4 @@ class TestXYZ(unittest.TestCase):
 
 
 assert_type(TestXYZ.attr, int)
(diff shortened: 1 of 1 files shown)


test_cases/stdlib/check_unittest.py
```diff
@@ -114,4 +113,3 @@ def method() -> int:
 
 
 assert_type(TestXYZ.attr, int)
```
JelleZijlstra, line 115:
```suggestion
assert_type(TestXYZ.attr, int)
assert_type(TestXYZ.method(), int)
```
Keep some of the intent of the test

JelleZijlstra (approved):

PR #10336, opened 2023-06-20
CI: Pin Python to 3.7.16 for stubtest runs on macos

### LINKED ISSUE #10333: Stdlib stubtest failing on Python 3.7, macOS

AlexWaygood:
Temporary workaround to resolve #10333.

Also, while we're here:
- Run stubtest on 3.12 as part of the daily tests, not just on PR runs
- Use the new `allow-prereleases` key with the setup-python action, so we won't need to manually drop the `-dev` suffix when 3.12.0 is released in the autumn.

### DIFF
--- a/.github/workflows/daily.yml
+++ b/.github/workflows/daily.yml
@@ -38,18 +38,28 @@ jobs:
         os: ["ubuntu-latest", "windows-latest", "macos-11"]
         # TODO: unpin the micro version of 3.7 when this issue is resolved:
         # https://github.com/actions/setup-python/issues/682
-        python-version: ["3.7.16", "3.8", "3.9", "3.10", "3.11", "3.12"]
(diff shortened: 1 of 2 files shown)


AlexWaygood:
*Sigh*, now the Windows run is failing on Python 3.7.

.github/workflows/daily.yml
```diff
@@ -36,17 +36,30 @@ jobs:
       matrix:
         # tkinter doesn't import on macOS-12
         os: ["ubuntu-latest", "windows-latest", "macos-11"]
-        python-version: ["3.7", "3.8", "3.9", "3.10", "3.11"]
+        # TODO: unpin the micro version of 3.7 when this issue is resolved:
+        # https://github.com/actions/setup-python/issues/682
```
AlexWaygood, line 40:
```suggestion
```

hauntsaninja (approved):
yikes, thanks!

PR #5052, opened 2021-02-22
assert_has_awaits should take Iterable[_Call] instead of _CallList

JonathanMeans:
This addresses issue #5012

PR #10337, opened 2023-06-21
Take an iterable of calls for assert_has_awaits in mock lib

FasterSpeeding:
This copies over the fix from https://github.com/python/typeshed/pull/5052 to the unittest.mock backport library's stubs


hauntsaninja (approved):
Thanks again!

PR #10338, opened 2023-06-21
use ParamSpec in patch in mock stubs

FasterSpeeding:
This copies over the change from https://github.com/python/typeshed/pull/10325 to the unittest.mock backport library's stubs


hauntsaninja (approved):
Thank you!

PR #10341, opened 2023-06-21
Bump croniter to 1.4

srittau:
Also require `bytes` for all `hash_id`s for consistency, even
if some methods might work with other types as well. But
using other types might hint at some inconsistent type
usage and be a source of possible bugs.

Improve annotations of some final attributes.

Closes: #10321


stubs/croniter/croniter/croniter.pyi
```diff
@@ -32,32 +36,44 @@ class CroniterNotAlphaError(CroniterError): ...
...
+            Literal[30],
+            Literal[31],
+            Literal[30],
+            Literal[31],
+            Literal[31],
+            Literal[30],
+            Literal[31],
+            Literal[30],
+            Literal[31],
+        ]
```
AlexWaygood, line 55: Honestly, I'd be inclined to go for a simpler type here (but I don't feel strongly about it!):

```suggestion
    DAYS: Final[tuple[Literal[28, 30, 31], ...]]
```

stubs/croniter/croniter/croniter.pyi
```diff
@@ -84,9 +101,9 @@ class croniter(Iterator[Any]):
     def iter(self, ret_type: _RetType | None = ...) -> Iterator[Any]: ...
     def is_leap(self, year: int) -> bool: ...
     @classmethod
-    def expand(cls, expr_format: str, hash_id: str | bytes | None = None) -> tuple[list[list[str]], dict[str, Any]]: ...
+    def expand(cls, expr_format: str, hash_id: bytes | None = None) -> tuple[list[list[str]], dict[str, Any]]: ...
```
AlexWaygood, line 104: I think the second item of the return tuple should be the same type as the `nth_weekday_of_month` attribute annotation

```suggestion
    def expand(cls, expr_format: str, hash_id: bytes | None = None) -> tuple[list[list[str]], dict[str, set[int]]]: ...
```

stubs/croniter/croniter/croniter.pyi
srittau: I like having a fixed-size tuple. On the other hand, I don't think the concrete numbers are terribly important:

```python
DAYS: Final[tuple[int, int, int, int, int, int, int, int, int, int, int, int]]
```

AlexWaygood: Let's just leave it as it is, then -- I quite like the concrete numbers, since we know for a certainty that no month has a number of days that's not `28`, `30` or `31`, and it feels nice to me to reflect that in the stubs 😄

PR #10345, opened 2023-06-21
Update OpenSSL/crypto.pyi

jvanasco:
Adds typing for missing constants `TYPE_DH` and `TYPE_EC`

See https://github.com/pyca/pyopenssl/blob/main/src/OpenSSL/crypto.py#L94-L95


AlexWaygood (approved):
Thanks!

PR #10346, opened 2023-06-22
Fix return type of start_tls()

Dreamsorcerer:
This just returns `_app_transport`:
https://github.com/python/cpython/blob/dba72175116373c1d15e25d84c88b516daf9f5c4/Lib/asyncio/base_events.py#L1288
Which will be set to `None` on init and after connection lost:
https://github.com/python/cpython/blob/dba72175116373c1d15e25d84c88b516daf9f5c4/Lib/asyncio/sslproto.py#L412

Related bug report:
https://github.com/aio-libs/aiohttp/issues/3355

### DIFF
--- a/stdlib/asyncio/base_events.pyi
+++ b/stdlib/asyncio/base_events.pyi
@@ -263,7 +263,7 @@ class BaseEventLoop(AbstractEventLoop):
             server_hostname: str | None = None,
             ssl_handshake_timeout: float | None = None,
             ssl_shutdown_timeout: float | None = None,
-        ) -> Transport: ...
+        ) -> Transport | None: ...
         async def connect_accepted_socket(
             self,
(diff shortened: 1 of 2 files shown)


Dreamsorcerer:
I've opened a cpython bug report, maybe this should be considered a bug rather than incorrect typing:
https://github.com/python/cpython/issues/105993

AlexWaygood:
> I've opened a cpython bug report, maybe this should be considered a bug rather than incorrect typing:
> [python/cpython#105993](https://github.com/python/cpython/issues/105993)

Let's wait for the asyncio experts over at CPython to decide whether or not it's a bug in CPython before we merge this 👍

AlexWaygood:
The CPython docs PR has been merged, so I'm fan with landing this now. Thanks!

PR #10352, opened 2023-06-23
Bump pyright to 1.1.315

AlexWaygood:


AlexWaygood:
With the latest update, pyright is failing on a bunch of our test cases for openpyxl descriptors. @Avasam, could there be a bug in our openpyxl stubs/tests, or do you think this may be a pyright regression? :)

Avasam:
If it ain't too much of a hurry I'll take a deeper look at it tomorrow. There's two specific descriptors failing, and only for setters tests.
If this was a regression I'd expect more than that.

Maybe I didn't override the `__set__` correctly.

AlexWaygood:
> If it ain't too much of a hurry I'll take a deeper look at it tomorrow.

No hurry at all, take your time!!

Avasam:
There you go: #10356

PR #10356, opened 2023-06-23
Bump pyright to 1.1.315 and fix openpyxl tests

Avasam:
Only changing the generic type from `bool` to `_T` in `nested.pyi` was necessary. But figured may do it in `base.pyi` as well.
Still unsure whether this counts as a pyright regression, I feel like it does, but at least it works and it's consistant now.

Also fixed an incorrect self type in `NestedText` that didnt really affect typing (but it was incorrect so may as well fix it).

Closes #10352 by superseeding it.


AlexWaygood:
Thanks @Avasam for fixing and @JelleZijlstra for merging!

PR #10372, opened 2023-06-27
s2clientprotocol stubs

### LINKED ISSUE #10371: Create third-party stubs for s2clientprotocol
## Background

Blizzard originally published the [s2client-proto](https://github.com/Blizzard/s2client-proto) Protocol Buffers in [2017](https://github.com/Blizzard/s2client-proto/commit/dca8b6831a84747c2cd6e0c33d6416e14838d886). The Protocol Buffers themselves are largely frozen in time, as updates to the StarCraft II API ceased with the end of DeepMind's involvement - but Blizzard automation continues to package the Python modules generated from the Protocol Buffers and publish this package to PyPi as [s2clientprotocol](https://pypi.org/project/s2clientprotocol/). This work predates [mypy-protobuf](https://github.com/nipunn1313/mypy-protobuf), which leaves the package without stubs.

## Proposal

Given the sc2client-proto repository is in maintenance mode, I would like to generate stubs from the Protocol Buffers using:

- The [protoc executable](https://github.com/Blizzard/s2client-proto/blob/master/install_protoc.sh) used by s2client-proto to generate the Python files, in conjunction with
- The [mypy-protobuf](https://pypi.org/project/mypy-protobuf/) plugin

The generated stubs could then be published here on typedshed under `stubs/s2clientprotocol`.

I have already spiked the issue, and the stubs are correctly generated with the caveat that they will fail some of `typeshed`'s stub tests (I believe this is a consequence of stubbing Protocol Buffers in general with `mypy-protobuf`, but the stubs themselves appear to be functional with both `pyright` and `mypy`).

## Alternatives

I have already opened a [PR](https://github.com/Blizzard/s2client-proto/pull/204) to have the stubs generated and published with the `s2clientprotocol` package - but I have my concerns regarding the publication environment and how much effort the automation maintainers are willing to invest to ensure my changes continue to work. In the case that the PR goes nowhere, I intend to open a PR against `typeshed` to have the stubs published here instead. However, given the stubs do not pass `typeshed` stub tests on the merits of them being stubs for generated Python via `mypy-protobuf`, I am equally concerned for their uptake here.

jrtknauer:
## Motivation

Fixes #10371. Should the `s2client-proto` maintainers respond to and merge my PR as part of the package, then inclusion of the stubs here would no longer be necessary. However, the maintainers do not seem to be particularly active, so I believe it would be better to get the stubs working here first (if it is possible to overlook the discrepencies in the third party stub test).


stubs/s2clientprotocol/METADATA.toml
```diff
@@ -0,0 +1,2 @@
+version = "5.*"
+requires = ["protobuf<4", "types-protobuf"]
```
AlexWaygood, line 2: `protobuf` isn't `py.typed` (hence why we have stubs for it in typeshed!), so I don't think having it as a dependency _as well as_ `types-protobuf` gets us anything here? Removing it should also fix the `stub_uploader` errors in CI:

```suggestion
requires = ["types-protobuf"]
```

AlexWaygood:
(Nit: no need to force-push on a typeshed PR. We merge all PRs using GitHub's "squash and merge" button, so we don't really care about a clean commit history, and force pushes tend to make it harder for us to see what changed in between commits :)

AlexWaygood (approved):
Thanks, and sorry for the delay! Stubtest struggles a bit with this package, so I've had to add all the generated files to the stubtest allowlist. We already do this quite extensively for other stubs packages that we generate using mypy-protobuf, though, so I'm fine with that. mypy-protobuf generally generates pretty good stubs that users seem to be happy with!

jrtknauer:
No worries. In the interim I have been building the protos and stubs in-project, but there are other `s2clientprotocol` users who could benefit from having access to stubs.

PR #10376, opened 2023-06-28
Bump PyMySQL to `1.1.*`

sobolevn:
Docs for `collation`: https://github.com/PyMySQL/PyMySQL/blob/dbf1ff52a695278cd80e179641f67bb6e2a83326/pymysql/connections.py#L116

Closes https://github.com/python/typeshed/pull/10365


sobolevn:
`PyMySQL.connections.Connection` is a good example of `TypeVar` default use case: you can pass `_C` cursor type or use `None`, then `Cursor` will be used by default.

srittau (approved):
Thanks!

PR #10380, opened 2023-06-28
Add stubs for qrcode

srittau:
This is best reviewed commit by commit. The first commit is just the result of create_baseline_stubs. The other commit were the minimum required to make the tests pass.

I plan to go over the stubs manually in a future PR, but I tried to keep this initial PR as minimal as possible, trusting the existing annotations and the tools.


AlexWaygood:
It seems like the runtime package is already quite heavily annotated. Is it worth asking upstream if they'd consider adding a `py.typed` file, before we add stubs to typeshed?

srittau:
> It seems like the runtime package is already quite heavily annotated. Is it worth asking upstream if they'd consider adding a `py.typed` file, before we add stubs to typeshed?

I would call it "lightly annotated", so I think there is quite some room for improvement in the stubs for now.

AlexWaygood (commented):
The PR looks fine, but I still feel like this might be an unnecessary duplication of effort, given that it looks like the runtime package has several PRs open to improve their type annotations. I don't want to block the PR if others disagree with me, however :)

srittau:
Considering that the last upstream commit was in February and the PR has been open since March, I would prefer to have the types in typeshed for now.

AlexWaygood:
No other maintainer has chimed in over the last three weeks, and I don't feel strongly about it, so feel free to merge.

AlexWaygood (approved):

PR #10389, opened 2023-06-30
Remove `SQLAlchemy` stubs

### LINKED ISSUE #8885: Remove Flask-SQLAlchemy stubs (not until April 2023)

### LINKED ISSUE #9607: Remove stubs for SQLAlchemy (not until at least July 2023)
SQLAlchemy 2.0 has just been released, making our stubs obsolete: https://github.com/python/typeshed/commit/808a431a57e7efb2ec8afa97840c368e760d2e11.

However, we can continue to make improvements to our SQLAlchemy stubs in the meantime, and we probably don't need to see this as a _strict_ deadline. See discussion in #9595 for some reasons why the situation with SQLAlchemy is a little unique.

sobolevn:
I don't feel quite comfortable with this PR for several reasons (and not because it is 30 Jun and not 1 Jul today 😆):
1. I remember https://github.com/python/typeshed/pull/9989 So, maybe we should delete them together?
2. I think that these stubs are widely used and most people are stuck with `SQLAlchemy@1` with no chances to update to `v2`. Do we want to leave them with all the existing typing imperfections?
3. There are still some dependent packages. For example, `Flask-Migrate`

Refs https://github.com/python/typeshed/pull/9595
Supersedes https://github.com/python/typeshed/pull/9989
Closes #9607
Closes #8885
Closes #9597


AlexWaygood:
> 1. I remember [Remove `Flask-SQLAlchemy` #9989](https://github.com/python/typeshed/pull/9989) So, maybe we should delete them together?

Judging from the pytype errors, it looks like we'll have to

AlexWaygood:
`pytype_test.py` is failing because:
- `Flask-SQLAlchemy` becomes a non-types dependency of `Flask-Migrate` in this PR
- `pytype_test.py` tries to inspect the `top_level.txt` file of all non-types dependencies
- `Flask-SQLAlchemy` uses pdm as its build backend, and pdm does not create `top_level.txt` files in the same way that setuptools does

AlexWaygood:
Well, I had _hoped_ that https://github.com/python/typeshed/pull/10391 would resolve the CI failures here, but it doesn't look like that's the case. It seems like the dictionary returned by `importlib.metadata.packages_distributions()` in CI doesn't include `flask_sqlalchemy` as a key when `Flask-SQLAlchemy` is installed. This is strange, as it does for me when I run the same commands locally on Windows. I don't understand what's causing this issue.

sobolevn:
> I don't understand what's causing this issue.

Me neither. 
CC @rchen152 as pytype expert

AlexWaygood:
All green now after https://github.com/python/typeshed/pull/10393 🥳🥳

sobolevn:
@JelleZijlstra @srittau @AlexWaygood @hauntsaninja I think that this is important enough to ask for all opinions :)

AlexWaygood (approved):
I'm in favour of this, but I'm also okay with waiting a while longer if others want to

srittau (approved):
Upgrading to SQLAlchemy 2 is hard, but well worth it if using type annotations, as its built-in typing support is much better than anything pure stubs can achieve. According to https://pypistats.org/, this package is not very widely used, at least as compared to some other type packages. Also, the package itself doesn't go away, although we can't update it. +1 for removing the stubs now.

Issue #10384, opened 2023-06-29
``auto`` for ``StrEnum`` is incorrectly subclassing ``IntFlag``


DanielNoord:
```python
from enum import StrEnum, auto
from typing import reveal_type


class Test(StrEnum):
    test = auto()


print(type(Test.test.value))


def func(x: Test):
    print(reveal_type(x.value))


func(Test.test)
```

On ``3.11`` this returns:
```console
❯ python test.py
<class 'str'>
Runtime type is 'str'
test
```

However, both `pyright` and `mypy` reveal the type as being `int` because of:
https://github.com/python/typeshed/blob/201940cdc12e61e3bf5bc3ee4c5fd71bcc23fdf9/stdlib/enum.pyi#L289-L290

This was introduced all the way back in https://github.com/python/typeshed/pull/1331.

Well it might be "mostly" correct, with `StrEnum` this is clearly false and leads to cases where we couldn't use `StrEnum` without adding `type: ignores` anyway.
I don't really know what the appropriate fix is so I sadly can't contribute one.

srittau:
Our annotations of `auto()` predate `StrEnum`, and the current hack used to work (and still works) for the other enum types. But due to the highly dynamic nature of `auto` (and enums in general), this is something that should be implemented in the type checkers. typeshed's hack doesn't work anymore.

In light of the introduction of `StrEnum` and to prevent backwards incompatibilities, I think the best way forward is to ask type checker implementors to special case `auto()` and then eventually drop our hack from typeshed.

Cc @ilevkivskyi @JukkaL @erictraut for awareness.

sobolevn:
I can add a plugin for this to mypy tomorrow :)

PR #10390, opened 2023-06-30
Add `_generate_next_value_` to `StrEnum`

sobolevn:
This change is enough to (at least partially) fix https://github.com/python/typeshed/issues/10384

Why? Because `mypy` already had this plugin in place: https://github.com/python/mypy/blob/b995e1620789a3ab2fc5dbcf0698a9077b0f5731/mypy/plugins/enums.py#L85-L90

Source: https://github.com/python/cpython/blob/319de0b578ec3dec8d30610caee3aa93f51060c7/Lib/enum.py#L1279

So, adding it locally produces:

```python
from enum import StrEnum, auto

class A(StrEnum):
    a = auto()

reveal_type(type(A.a))  # N: Revealed type is "Type[Literal[ex.A.a]?]"
reveal_type(type(A.a.value))  # N: Revealed type is "Type[builtins.str]"
```

Closes https://github.com/python/typeshed/issues/10384


tests/stubtest_allowlists/py311.txt
```diff
@@ -17,6 +17,7 @@ configparser.LegacyInterpolation.__init__
 configparser.ParsingError.filename
 enum.Enum.__init__
 enum.Enum._generate_next_value_
+enum.StrEnum._generate_next_value_
```
AlexWaygood, line 20: Can you add a comment here to say why we're allowlisting it?

JelleZijlstra (approved):

stdlib/enum.pyi
```diff
@@ -237,6 +237,8 @@ if sys.version_info >= (3, 11):
         _value_: str
         @_magic_enum_attr
         def value(self) -> str: ...
+        @staticmethod
+        def _generate_next_value_(name: str, start: Unused, count: Unused, last_values: Unused) -> str: ...
```
AlexWaygood, line 241: The idea of `_generate_next_value_` is that you can override it in subclasses of `Enum` to provide custom behaviour for how `auto()` works. As such, I don't think we should use `object` annotations (which is what `Unused` is) for the parameters, or subclassers won't get the correct signature if they do actually want to use the parameters.

```suggestion
        def _generate_next_value_(name: str, start: int, count: int, last_values: list[str]) -> str: ...
```

tests/stubtest_allowlists/py311.txt
```diff
@@ -17,6 +17,8 @@ configparser.LegacyInterpolation.__init__
 configparser.ParsingError.filename
 enum.Enum.__init__
 enum.Enum._generate_next_value_
+# Python3.11 used to dynamically add this method:
```
AlexWaygood, line 20:
```suggestion
# Not strictly speaking a staticmethod on 3.11, but it acts like one:
```

AlexWaygood (approved):
Thanks!

PR #10391, opened 2023-06-30
pytype_test.py: Use importlib.metadata instead of pkg_resources

AlexWaygood:
My hope is that this will resolve the pytype failures in #10389. It's probably a good idea to do this anyway, though, since `pkg_resources` is deprecated.

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -72,7 +72,7 @@ for this script.
 
 Note: this test cannot be run on Windows
 systems unless you are using Windows Subsystem for Linux.
-It also requires a Python version < 3.11 as pytype does not yet support
+It can currently only be run on Python 3.10 as pytype does not yet support
 Python 3.11 and above.
 
 Run using:
--- a/tests/README.md
+++ b/tests/README.md
@@ -19,7 +19,7 @@ objects at runtime.
 in the `tests` and `scripts` directories.
 
 To run the tests, follow the [setup instructions](../CONTRIBUTING.md#preparing-the-environment)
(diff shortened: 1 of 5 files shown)


tests/README.md
```diff
@@ -72,7 +72,7 @@ for this script.
 
 Note: this test cannot be run on Windows
 systems unless you are using Windows Subsystem for Linux.
-It also requires a Python version < 3.11 as pytype does not yet support
+It can currently only be run on Python 3.10 as pytype does not yet support
```
AlexWaygood, line 75: We could avoid this if we added the `importlib_metadata` backport as a test dependency

tests/pytype_test.py
```diff
@@ -165,15 +165,17 @@ def get_missing_modules(files_to_test: Sequence[str]) -> Iterable[str]:
         except ValueError:
             continue
         stub_distributions.add(parts[idx + 1])
+
+    dist_to_pkg_map = defaultdict(list)
+    for dist, pkg_list in importlib.metadata.packages_distributions().items():
+        for pkg in pkg_list:
+            dist_to_pkg_map[dist].append(pkg)
```
AlexWaygood, line 172:
```suggestion
            dist_to_pkg_map[pkg].append(dist)
```

tests/README.md
srittau: I prefer the narrow requirements in this case. I would be surprised if pytype_test is much run manually and an extra requirement can always be a bit of a pain.

AlexWaygood: I agree, though I don't have a strong opinion at all

hauntsaninja (approved):

PR #10392, opened 2023-06-30
Fix signature of `http.client.HTTPSConnection` for Python3.12

sobolevn:
Source: https://github.com/python/cpython/blob/3.12/Lib/http/client.py#L1441-L1443

I will change `http.client.HTTPConnection.get_proxy_response_headers` after we will decide on https://github.com/python/cpython/pull/105628

### DIFF
--- a/stdlib/http/client.pyi
+++ b/stdlib/http/client.pyi
@@ -175,19 +175,31 @@ class HTTPConnection:
 class HTTPSConnection(HTTPConnection):
     # Can be `None` if `.connect()` was not called:
     sock: ssl.SSLSocket | Any
-    def __init__(
-        self,
-        host: str,
-        port: int | None = None,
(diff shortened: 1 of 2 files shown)


AlexWaygood (approved):

stdlib/http/client.pyi
```diff
@@ -175,19 +176,31 @@ class HTTPConnection:
...
-        *,
-        context: ssl.SSLContext | None = None,
-        check_hostname: bool | None = None,
-        blocksize: int = 8192,
-    ) -> None: ...
+    if sys.version_info >= (3, 12):
+        def __init__(
+            self,
+            host: str,
+            port: str | None = None,
```
allisonkarlitskaya, line 183: Was it intentional to change this to `str`?  I'm fairly sure this is incorrect...

PR #10393, opened 2023-06-30
Further improve non-types dependency handling in `pytype_test.py`

AlexWaygood:
Another attempt at resolving the pytype CI errors in #10389. I've done a lot of testing via GitHub Actions on my fork, and I'm _pretty_ sure this gets the CI green on #10389. I don't know why these importlib APIs seem to be able to find the packages associated with `Flask-SQLAlchemy` when `importlib.metadata.packages_distributions()` doesn't seem able to, but that does seem to be the case.

Happily, these changes mean that our test suite can once again be run on Python 3.9, which is a nice-to-have.

### DIFF
--- a/tests/pytype_test.py
+++ b/tests/pytype_test.py
@@ -16,10 +16,10 @@ from __future__ import annotations
 
 import argparse
 import importlib.metadata
+import inspect
 import os
 import sys
 import traceback
-from collections import defaultdict
 from collections.abc import Iterable, Sequence
 
 from packaging.requirements import Requirement
@@ -29,8 +29,8 @@ from parse_metadata import read_dependencies
 if sys.platform == "win32":
     print("pytype does not support Windows.", file=sys.stderr)
     sys.exit(1)
-if sys.version_info[:2] != (3, 10):
-    print("pytype_test.py can currently only be run on Python 3.10.", file=sys.stderr)
+if sys.version_info >= (3, 11):
+    print("pytype does not support Python 3.11+ yet.", file=sys.stderr)
     sys.exit(1)
 
 # pytype is not py.typed https://github.com/google/pytype/issues/1325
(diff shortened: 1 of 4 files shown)


tests/pytype_test.py
```diff
@@ -149,6 +149,17 @@ def find_stubs_in_paths(paths: Sequence[str]) -> list[str]:
...
 
+def _get_pkgs_associated_with_requirement(req_name: str) -> list[str]:
+    dist = importlib.metadata.distribution(req_name)
+    toplevel_txt_contents = dist.read_text("top_level.txt")
+    if toplevel_txt_contents is not None:
+        return toplevel_txt_contents.split()
+    if dist.files is None:
+        raise RuntimeError("Can't read find the packages associated with requirement {req_name!r}")
+    maybe_modules = [f.parts[0] if len(f.parts) > 1 else inspect.getmodulename(f) for f in dist.files]
+    return [name for name in maybe_modules if name is not None and "." not in name]
```
AlexWaygood, line 160: I drew on the implementation of `packages_distributions()` heavily when writing this function: https://github.com/python/cpython/blob/904aef994262383ae916545908f0578c2d53cf31/Lib/importlib/metadata/__init__.py#L933-L964

tests/typecheck_typeshed.py
```diff
@@ -63,6 +63,8 @@ def run_mypy_as_subprocess(directory: str, platform: str, version: str) -> Retur
         "possibly-undefined",
         "--enable-error-code",
         "redundant-expr",
+        "--custom-typeshed-dir",
+        ".",
```
AlexWaygood, line 67: This is now necessary to get our test suite to typecheck, since we're now passing an `os.PathLike` object into `inspect.getmodulename()` in `pytype_test.py`, and the recent update to our `inspect` stubs in #10329 hasn't made it into mypy yet

JelleZijlstra (approved):

PR #10398, opened 2023-07-03
Update `inspect` for Python3.12

sobolevn:
Source: https://github.com/python/cpython/blob/3.12/Lib/inspect.py


srittau (approved):
Thanks!

PR #10399, opened 2023-07-03
[Pillow] Update TiffImagePlugin for Python 3.11

srittau:
Also type the methods forwarded from Fraction and replace
`Any` with `Incomplete`.

Preparation for updating to Pillow 10.


srittau (commented):
./.

stubs/Pillow/PIL/TiffImagePlugin.pyi
```diff
@@ -71,34 +75,100 @@ class IFDRational(Rational):
...
-    __trunc__: Any
-    __lt__: Any
-    __gt__: Any
-    __le__: Any
-    __ge__: Any
-    __bool__: Any
-    __ceil__: Any
-    __floor__: Any
-    __round__: Any
+    # The following methods are aliased from fractions.Fraction:
```
AlexWaygood, line 78: I think we could just do

```py
    __add__ = Fraction.__add__
    __radd__ = Fraction.__radd__
```

etc.

srittau: I tried this at first, but got overload incompatible type errors with the overloaded functions. While we could probably just ignore those, I thought it would be safer to just copy the signature.

AlexWaygood: Okay, let's keep it as you have it then

Issue #10255, opened 2023-06-04
`distutils` and `setuptools._distutils` stubs are different


sobolevn:
As far as I understand, `setuptools._distutils` is just a vendored copy of `distutils`. 
But, right now they are different. For example: 
- https://github.com/python/typeshed/blob/c1a857a41a1cbc62df5c6fdac52d0ac79fdf2824/stdlib/distutils/core.pyi
- https://github.com/python/typeshed/tree/c1a857a41a1cbc62df5c6fdac52d0ac79fdf2824/stubs/setuptools/setuptools/_distutils (no `core.pyi`)

Here's the source code for `core.py`:
- https://github.com/python/cpython/blob/3.11/Lib/distutils/core.py
- https://github.com/pypa/setuptools/blob/b545fc778583f644d6c331773dbe0ea53bfa41af/setuptools/_distutils/core.py

So, we need to work on them! 
Refs https://github.com/python/typeshed/pull/10249

AlexWaygood:
Note that we deliberately removed much of  our `setuptools/_distutils` stubs in #9795, on the grounds that setuptools appeared to view the directory as an implementation detail; it was a significant maintenance burden to keep our stubs for the directory up to date with the runtime package; and users should probably be importing things from `setuptools` instead of `setuptools._distutils` wherever possible.

FWIW I'm open to revisiting that decision if there's a good reason to, but we should think carefully about it. Prior to that change, it felt like stubtest was breaking every month due to the rate of changes setuptools was making to `setuptools._distutils`.

hauntsaninja:
My opinions:

stdlib `distutils` is gone. Most users already do not get it when they `import distutils`. I have no interest in seeing further improvements made to its stubs.

`setuptools._distutils` seems [not infrequently used in the wild](https://grep.app/search?q=setuptools._distutils), and I imagine it will be used more by projects looking to support 3.12 with the fewest changes possible. Given this, I'm supportive of improvements to these stubs.

Regarding maintenance burden from setuptools development of `setuptools._distutils`: I default to the position that users are worse in the world where we don't have stubs at all (certain false positive) than in the world where stubs exist + we stubtest allowlist them (possible incorrect behaviour, but likely only if depending on implementation detail)

Of course, if there's no usage of the stubs then better off staying deleted, e.g. I don't see much point in bringing back the `setuptools._distutils.command.*` removed in #9795 

So all of this adds up to the following possible plan of action:
[...]

PR #10401, opened 2023-07-03
Bring back a few setuptools._distutils files

hauntsaninja:
Linking #10255

There are cases of these being used, including in mypy This is reverts a small part of #9795


srittau (approved):

srittau:
I did not review the files as this is a revert.

PR #10403, opened 2023-07-04
Redis: Add missing type for id in xadd

Kontrolix:
According to [documentation](https://redis-py.readthedocs.io/en/stable/commands.html#redis.commands.cluster.RedisClusterCommands.xadd) xadd  support 4 types for param id.


sileht (approved):

srittau (approved):
Thanks!

PR #10404, opened 2023-07-04
tkinter: Add missing "angle" argument to Canvas create_text

simaoafonso-pwt:
This is supported upstream.

https://tcl.tk/man/tcl8.6/TkCmd/canvas.htm#M156

Tested all these types and they all work.


Akuli (approved):
Thanks!

PR #10407, opened 2023-07-05
Pillow: Add `tuple[int, int]` to `_Color` TypeAlias

JohnVillalovos:
The `color` value can be a `tuple[int, int]` as seen in the test case [1] from the Pillow project.

For example when using the `LA` mode in a call to `PIL.Image.new()`:
  PIL.Image.new("LA", size=(200, 200), color=(0xFF, 0))

`color` can be a `tuple[int, int]`

[1] https://github.com/python-pillow/Pillow/blob/3ee925915085ee0f7ad72fbede5d9a9e6ea902a7/Tests/test_image_access.py#L376


Akuli (approved):
It might be possible to make `_Color` depend on the mode, so that you can't use a two-tuple with RGB for example, but that's beyond the scope of this PR.

PR #10408, opened 2023-07-05
Silence subclassing `Any` error for `_NotImplementedType`

ngnpope:
I somehow managed to encounter this issue bubbling up when using `mypy --strict` which enables `--disallow-subclassing-any`.

Am using the `pytest-mypy-plugins` package to test a new stubs-only package I'm writing, so I don't expect that this usually occurs in normal usage of `mypy` on the command line - at least it's been this way a long time and I've never seen it 🤔 

Example output:

```
.../tests/typecheck/test_all_imports.yml:3:
E   pytest_mypy_plugins.utils.TypecheckAssertionError: Output is not expected:
E   Actual:
E     .../lib/python3.11/site-packages/mypy/typeshed/stdlib/builtins.pyi:1114: error: Class cannot subclass "Any" (has type "Any")  [misc] (diff)
E   Expected:
E     (empty)
```

I've been working around this with the following:

```ini
[mypy]
strict = true
...

[mypy-builtins]
disallow_subclassing_any = false
```

### DIFF
--- a/stdlib/builtins.pyi
+++ b/stdlib/builtins.pyi
@@ -1221,7 +1221,7 @@ class property:
     def __delete__(self, __instance: Any) -> None: ...
 
 @final
-class _NotImplementedType(Any):
+class _NotImplementedType(Any):  # type: ignore[misc]
     # A little weird, but typing the __call__ as NotImplemented makes the error message
     # for NotImplemented() much better
     __call__: NotImplemented  # type: ignore[valid-type]  # pyright: ignore[reportGeneralTypeIssues]


AlexWaygood:
The `type: ignore`s in typeshed are purely for our internal tests; we prefer not to introduce unused ignores into our codebase if possible.

Typing errors in typeshed are supposed to not be visible to users of external tools which bundle typeshed. If pytest-mypy-plugins is showing you an error from typeshed in your CI, that means there's probably a bug in pytest-mypy-plugins, unfortunately (it _should_ be silencing any errors that originate from typeshed). Have you tried filing a bug report with them?

AlexWaygood:
Aside from anything else, there are a fair few other places where we subclass Any in typeshed (not because we like doing so, but because there's sometimes no better option). It would be a bit inconsistent to `type: ignore` this one but not all the other ones :)

PR #10409, opened 2023-07-05
[Pillow] Replace `Any` with `Incomplete`

srittau:
Only left instances where it was obvious why `Any` was used.

Depends on #10399


AlexWaygood (approved):
LGTM once #10399 has gone in

PR #10410, opened 2023-07-05
Mark `types-typed-ast` as no longer updated

AlexWaygood:
The [`typed-ast` repo](https://github.com/python/typed_ast) has been archived, and has a notice on its README saying that it is no longer maintained. Mypy no longer depends on typed-ast, and black is about to drop its dependency (https://github.com/psf/black/pull/3765).


JelleZijlstra (approved):

hauntsaninja:
🔪

PR #10416, opened 2023-07-06
Stripe: add shipping_rate to `__init__.py`

SergioB-dev:
This is a quick follow up to #10323. I failed to add the new type to `__init__.py` preventing the new type to be exposed for type checking.


AlexWaygood:
Thanks!

PR #10420, opened 2023-07-07
contextvars: rm -rf

### LINKED ISSUE #10418: contextvars are in both stubs and stdlib
There are two different contextvars pyi, one in stdlib and other in stubs.
If there is no good reason to keep both, I suggest to remove the one from stubs.

hauntsaninja:
Closes #10418


AlexWaygood (approved):
LGTM (once the `no_longer_updated` update has been pushed to PyPI)

AlexWaygood:
Marked as no longer updated on PyPI: https://pypi.org/project/types-contextvars/

PR #10422, opened 2023-07-08
Replace aliases by `builtins.*` in `uuid`

hamdanal:
These aliases were used because the class has properties called `int` and `bytes`. I think it is better to use the shadowed types from the builtins module. This should not change anything from the type checking side but it improves the IDE experience[^1].

[^1]: I was using uuid recently and the aliases showed up in vscode, I thought they were uuid special types and had to look in the stub to know what they are.


AlexWaygood (approved):

hamdanal:
Thanks Alex

PR #10423, opened 2023-07-08
Add stubs for `pkg_resources._vendor.packaging`

### LINKED ISSUE #10369: Error on stubs/setuptools/pkg_resources: Cannot access member "specifier" for type "Requirement"
There is no `specifier` attribute in `class Requirement` but when I run my code it is there and the program works. Pyright is reporting an error: `Cannot access member "specifier" for type "Requirement" Member "specifier" is unknown (reportGeneralTypeIssues)`

See: [typeshed/stubs/setuptools/pkg_resources/__init__.pyi](https://github.com/python/typeshed/blob/575dd0c448fed4e650b4a863a4d9160f6e092267/stubs/setuptools/pkg_resources/__init__.pyi#L86) 

My Python 3.11.3 code is:

```python
import pkg_resources     # Python 3.11.3
from pkg_resources import Requirement
. . . 
def check_missing_packages() -> bool:
    requirements_file = 'requirements.txt' # sys.argv[1]
    with open(requirements_file, "r") as f:
        required_packages = [
            line.strip().split("#")[0].strip() for line in f.readlines()
        ]
    working_set = pkg_resources.working_set
    installed_packages = {pkg.key: pkg.version for pkg in working_set}
    missing_packages = []
    print('-------------------- Requirements --------------------')
    for required_package in required_packages:
        if not required_package:  # Skip empty lines
            continue
        pkg: Requirement = Requirement.parse(required_package)
        pkg_specifier = pkg.specifier                           # ERROR REPORTED IN THIS LINE
        suffix_message = f', version: {pkg_specifier}' if len(str(pkg_specifier)) > 1 else ''
        print(f'pkg = {pkg.key}' + suffix_message)
        if (pkg.key not in installed_packages
            or pkg_resources.parse_version(
                    installed_packages[pkg.key]) not in pkg_specifier):
            missing_packages.append(str(pkg))
    if missing_packages:
        print("Missing packages Error: Aborting")
        print(", ".join(missing_packages))
        return False
    else:
        print("All packages are installed.")
        return True
```

**Additional information**
VSCode extension: Pyright - v1.1.315
pyright CLI : pyright - 1.1.315

AlexWaygood:
Really icky, but the only way to do it, since `pkg_resources` vendors `packaging` at runtime. Fixes #10369


hauntsaninja (approved):
Only as icky as any vendoring is!

PR #10425, opened 2023-07-09
Add __hash__ stub to classes that are hashable and define __eq__

### LINKED ISSUE #10424: Builtins are incorrectly hinted as unhashable
I believe this was partially introduced by #8465. I ran into this issue while looking into https://github.com/microsoft/pyright/issues/5446.

According to [the python docs](https://docs.python.org/3/reference/datamodel.html#object.__hash__:~:text=A%20class%20that%20overrides%20__eq__()%20and%20does%20not%20define%20__hash__()%20will%20have%20its%20__hash__()%20implicitly%20set%20to%20None), a type that redefines `__eq__` but not `__hash__` is unhashable.

In the current typeshed hints, many types explicitly define `__eq__` ([int for example](https://github.com/python/typeshed/blob/main/stdlib/builtins.pyi#L310)) but do not define hash. A type checker could incorrectly assume that the type is unhashable because of this.

TheTripleV:
Fixes #10424.

I only checked the types in builtins.pyi


AlexWaygood:
Thanks! Since #8465 didn't fix any known problems, however (just fixed some stylistic nits that we had), I'm inclined to revert the whole PR: see #10426. You're right that we missed this important detail of Python's semantics when it comes to determining whether an object is hashable or not :-)

PR #10426, opened 2023-07-09
Revert "Remove un-needed `__hash__` methods from `stdlib` (#8465)"

### LINKED ISSUE #10424: Builtins are incorrectly hinted as unhashable

AlexWaygood:
Reverts #8465
Fixes #10424
Closes #10425

#8465 caused regressions: see #10424 and https://github.com/python/mypy/issues/13800. Since it didn't fix any known problems (just some stylistic nits that we had), let's just revert the PR.


JelleZijlstra:
Seems like the mypy-primer hit is an example of python/mypy#13800. I wonder if it's because there's some change in whether mypy thinks a subclass of str and dict can exist.

PR #10427, opened 2023-07-09
Add `defaultdict.__or__`; improve `ChainMap.__or__` and `UserDict.__or__`

eltoder:


stdlib/collections/__init__.pyi
```diff
@@ -391,6 +391,9 @@ class defaultdict(dict[_KT, _VT], Generic[_KT, _VT]):
     def __missing__(self, __key: _KT) -> _VT: ...
     def __copy__(self) -> Self: ...
     def copy(self) -> Self: ...
+    if sys.version_info >= (3, 9):
+        def __or__(self, other: Mapping[_T1, _T2]) -> defaultdict[_KT | _T1, _VT | _T2]: ...
+        def __ror__(self, other: Mapping[_T1, _T2]) -> defaultdict[_KT | _T1, _VT | _T2]: ...
```
AlexWaygood, line 396: If you have a subclass of `defaultdict`, then `__or__` will actually return an instance of the subclass (note this _differs_ from how `dict.__or__` behaves):

```pycon
>>> from collections import defaultdict
>>> class Foo(defaultdict): ...
...
>>> type(Foo(int) | defaultdict(int))
<class '__main__.Foo'>
```

We can't express this fully without higher-kinded typevars (https://github.com/python/typing/issues/548), but we _can_ give a more specific return type in the specific case where a mapping that has the same type parameters is being `__or__`'d.

Also: the second parameter should be named `value`, not `other`:

```pycon
>>> from collections import defaultdict
>>> help(defaultdict.__or__)
Help on wrapper_descriptor:

__or__(self, value, /)
    Return self|value.
```

And it should be positional-only:

```pycon
>>> from collections import defaultdict
>>> x = defaultdict(int)
>>> x.__or__(value={})
TypeError: wrapper __or__() takes no keyword arguments
```

Putting all this together, I think the signature should be:

```suggestion
        @overload
        def __or__(self, __value: Mapping[_KT, _VT]) -> Self: ...
        @overload
        def __or__(self, __value: Mapping[_T1, _T2]) -> defaultdict[_KT | _T1, _VT | _T2]: ...
        @overload
        def __or__(self, __value: Mapping[_KT, _VT]) -> Self: ..
        @overload
        def __ror__(self, __value: Mapping[_T1, _T2]) -> defaultdict[_KT | _T1, _VT | _T2]: ...
```

AlexWaygood (commented):
Thanks! Looks pretty good, but a few nitpicks:

Akuli (approved):
I don't know why this works, but it works. Mypy (with this PR) and Python somehow decide that `defaultdict(int, {"foo": 1}) | {"bar": 2}` and `{"bar": 2} | defaultdict(int, {"foo": 1})` are defaultdicts, even though `dict` and `defaultdict` both define `__or__` and `__ror__`.

eltoder:
@Akuli defaultdict is a subclass of dict, so python prefers its method over dict's.

AlexWaygood:
Here's a full description of the nightmarish complexity Python goes through to evaluate `x | y`: https://snarky.ca/unravelling-binary-arithmetic-operations-in-python/

Akuli:
Thanks. I suspected it might have something to do with subclassing, but was somehow too lazy to look it up :)

stdlib/collections/__init__.pyi
JelleZijlstra: Probably the type of the other parameter should be `_typeshed.SupportsKeysAndGetItem`, not `Mapping`, right?

AlexWaygood: > Probably the type of the other parameter should be `_typeshed.SupportsKeysAndGetItem`, not `Mapping`, right?

Doesn't look like it?

```pycon
>>> from collections import defaultdict
>>> class Foo:
...     def __init__(self):
...         self.data = {}
...     def keys(self):
...         return self.data.keys()
...     def __getitem__(self, key):
...         return self.data[key]
...
>>> defaultdict(int) | Foo()
TypeError: unsupported operand type(s) for |: 'collections.defaultdict' and 'Foo'
```

eltoder: @JelleZijlstra @AlexWaygood defaultdict actually only supports subclasses of dict: https://github.com/python/cpython/blob/388b5daa523b828dc0f7e2a1a6886bebc20833ba/Modules/_collectionsmodule.c#L2136
However, we cannot declare it here as `dict`, because `dict` is declared to accept `Mapping` in typeshed, so this causes a violation of Liskov substitution. This also seems wrong. The code only accept dicts: https://github.com/python/cpython/blob/388b5daa523b828dc0f7e2a1a6886bebc20833ba/Objects/dictobject.c#L3606
Quick demo:
```python
>>> from collections.abc import Mapping
>>>> class D(Mapping):
    def __getitem__(self, key):
        raise KeyError
    def __len__(self):
        return 0
    def __iter__(self):
        return iter(())

>>> {"a": 1} | D()
TypeError: unsupported operand type(s) for |: 'dict' and 'D'
```

AlexWaygood: > @JelleZijlstra @AlexWaygood defaultdict actually only supports subclasses of dict:

Yes we had just come to the same conclusion :)

> However, we cannot declare it here as `dict`, because `dict` is declared to accept `Mapping` in typeshed, so this causes a violation of Liskov substitution. This also seems wrong. The code only accept dicts: https://github.com/python/cpython/blob/388b5daa523b828dc0f7e2a1a6886bebc20833ba/Objects/dictobject.c#L3606

Good catch! Want to update `dict.__(r)or__` in this PR as well?

eltoder:
@AlexWaygood It looks like the correct type for `dict.__or__` has an unfortunate side-effect: now there's no nice way to type a function that does `dict | os.environ`. This works because _Environ defines `__ror__`, but the type is not public and one wouldn't want to put it in the function signature. IOW:
```python
def fn1(lhs: dict[str, str], rhs: Mapping[str, str]) -> None:
    lhs | rhs  # does not type check now

def fn2(lhs: dict[str, str], rhs: dict[str, str]) -> None:
    lhs | rhs  # type checks, but cannot call with rhs=os.environ
```

Also, I got an error from mypy:
```
stdlib/collections/__init__.pyi:408: error: Signatures of "__ror__" of "defaultdict" and "__or__" of "dict[_T1, _T2]" are unsafely overlapping  [misc]
```
which did not make sense to me, so I added ignore.

I'm thinking to revert the change to dict. WDYT?

eltoder:
Maybe a better change would be to propose a change in cpython to use PyMapping_Check instead of PyDict_Check in dict and defaultdict. This is what almost all Python implemented `__or__` methods do. This is inconsistent with set, though.

JelleZijlstra:
> Maybe a better change would be to propose a change in cpython to use PyMapping_Check instead of PyDict_Check in dict and defaultdict. This is what almost all Python implemented `__or__` methods do. This is inconsistent with set, though.

That thought crossed my mind too, but it's not clear if it's right, and in any case we'd only make this change in 3.13+, so it doesn't really solve the problem of what to do in typeshed right now.

AlexWaygood:
I'll try something by pushing to your PR branch and see if it fixes some of the primer errors

JelleZijlstra:
As for `environ`, we do have stubs for the `os._Environ` type, and they define an `__or__` that accepts Mapping. I haven't checked whether that's exactly correct.

I'd be OK with just accepting `Mapping` for now in `defaultdict.__or__` and fixing the type in a separate PR.

AlexWaygood:
That didn't help anything, and it's not surprising that it didn't.

The issue isn't anything to do with our stubs for `os._Environ.__(r)or__`. The issue is that mypy has no way of knowing that the object being `|`'d is an instance of `os._Environ`, as they're just using `Mapping` as the type hint: https://github.com/systemd/mkosi/blob/84ba040dc048971f675c47f45b9b154e087e342e/mkosi/run.py#L227-L246C11

Arguably this is a true positive, as `__or__` and `__ror__` aren't included in the `Mapping` interface, so you can't guarantee that an arbitrary mapping can be `|`'d with a dictionary or an instance of `os._Environ`. But perhaps it makes it more annoying for users, for little gain. So maybe we should just stick with `Mapping` for `dict.__(r)or__` and `defaultdict.__(r)or__` for now.

eltoder:
@AlexWaygood it is a true positive, but it is annoying to properly type a [function like run](https://github.com/systemd/mkosi/blob/84ba040dc048971f675c47f45b9b154e087e342e/mkosi/run.py#L227).  `env: Mapping[str, PathString]` seems logical, but actually needs to be "a type that can be or'ed to a dict", which is "anything that accepts dict in `__ror__`". I think we can do this with a protocol and stick it into typing_extensions or something, but it's a lot of work. A more conservative type of `dict` will disallow `os.environ`. A quick fix for that is `dict | os._Environ`, but that seems ugly.

eltoder:
Done.

stdlib/weakref.pyi
```diff
@@ -119,7 +125,13 @@ class WeakKeyDictionary(MutableMapping[_KT, _VT]):
     @overload
     def pop(self, key: _KT, default: _VT | _T = ...) -> _VT | _T: ...
     if sys.version_info >= (3, 9):
+        @overload
+        def __or__(self, other: Mapping[_KT, _VT]) -> Self: ...
+        @overload
         def __or__(self, other: Mapping[_T1, _T2]) -> WeakKeyDictionary[_KT | _T1, _VT | _T2]: ...
+        @overload
+        def __ror__(self, other: Mapping[_KT, _VT]) -> Self: ...
+        @overload
```
AlexWaygood, line 134: Adding the new overloads here is incorrect: `weakref.WeakKeyDictionary.__or__` always returns `WeakKeyDictionary` instances, even in a subclass:

```pycon
>>> from weakref import WeakKeyDictionary
>>> class Foo(WeakKeyDictionary): ...
...
>>> type(Foo() | Foo())
<class 'weakref.WeakKeyDictionary'>
```

We should leave `WeakKeyDictionary` as it is. The same goes for `WeakValueDictionary`.

eltoder: Ah, you are right. It calls copy, but it's copy does not return Self.

AlexWaygood (approved):
LGTM. Thanks!

AlexWaygood:
Btw, workflow nit for future typeshed PRs: we _generally_ prefer to avoid force pushes at typeshed -- they don't work so well with GitHub's UI, making it hard to see what changed in between commits and thereby making it harder to do incremental reviews. We squash everything into a single commit before merging, anyway, so there's no need to worry about a messy commit history in a PR :)

PR #10432, opened 2023-07-10
Adds stubs for uWSGI

Daverball:
This adds stubs for the uWSGI Python API.

Similarly to GDB the Python API is only accessible within a uWSGI process, some parts of the API also only exist if certain configuration options are enabled. This makes running stubgen/stubdefaulter and most importantly stubtest a bit of pain.

For now the stubs have been created partially through the help of stubgen, but mostly by inspecing the C/Python code of the uwsgi API. Most of the C API code is located here: https://github.com/unbit/uwsgi/blob/2.0.21/plugins/python/uwsgi_pymodule.c but some parts of it reside in other files in this directory. The decorators submodule is written in pure python and can be referenced here: https://github.com/unbit/uwsgi/blob/2.0.21/uwsgidecorators.py
The [API docs](https://uwsgi-docs.readthedocs.io/en/latest/PythonModule.html) are also helpful to some degree, but they contain some functions that have been removed from the API a long time ago.

I will try to create a script to run stubtest locally at least once before taking this out of the Draft state. Depending on how that goes I may extend `tests/stubtest_thirdparty` with a special case for uWSGI, so we can have greater peace of mind that the stubs are complete going forward. But it may add too much complexity or not be helpful enough at catching errors, considering that some parts of the API depend on uWSGI configuration, which means new API commands could still  be missed by stubtest, if they require a new configuration value to be set.


Daverball:
@JelleZijlstra Can you take a look at the changes that were necessary to make `tests/stubtest_third_party.py` work with uWSGI? Would something like this be acceptable? It can and should probably still be cleaned up a bit.

tests/stubtest_third_party.py
```diff
@@ -97,8 +98,61 @@ def run_stubtest(dist: Path, *, verbose: bool = False, specified_platforms_only:
         if platform_allowlist.exists():
             stubtest_cmd.extend(["--allowlist", str(platform_allowlist)])
 
+        # Perform some black magic in order to run stubtest inside uWSGI
+        # we have to write the exit code from stubtest to a surrogate file
+        # because uwsgi --pyrun does not exit with the exitcode from the
+        # python script.
```
AlexWaygood, line 104: I'd definitely prefer it if you could factor out this special-casing into a separate function.

JelleZijlstra: Yes, was going to say the same thing

tests/stubtest_third_party.py
```diff
@@ -97,8 +98,61 @@ def run_stubtest(dist: Path, *, verbose: bool = False, specified_platforms_only:
...
+                    )
+                )
+
+            # FIXME: this is a bit fragile, can we think of a better
+            #        way of finding the executable inside the venv?
+            #        Is it maybe sufficient to prepend venv_dir to PATH
+            #        and then run shutil.which?
+            uwsgi_exe_name = "uwsgi.exe" if sys.platform == "win32" else "uwsgi"
+            stubtest_cmd = [
+                str(venv_dir / "bin" / uwsgi_exe_name),
```
Akuli, line 135: There is a `Scripts` folder instead of a `bin` folder on Windows.

tests/stubtest_third_party.py
Daverball: Yes, that seems sensible, it's a bit unfortunate that the hack with the exit code can't be factored out quite as nicely. Can you think of a better way? If this solution only has to work on linux/darwin I could create a small sh script which could be run instead of calling uwsgi directly.

Daverball: You are right. That seems like the easiest solution.

Daverball:
@JelleZijlstra @AlexWaygood @Akuli Alright this should be a lot better now.

tests/stubtest_third_party.py
```diff
@@ -128,6 +137,84 @@ def run_stubtest(dist: Path, *, verbose: bool = False, specified_platforms_only:
...
+                from mypy.stubtest import main
+
+                sys.argv = json.loads(os.environ.get("STUBTEST_ARGS"))
+                exit_code = main()
+                with open("{exit_code_surrogate}", mode="w") as fp:
+                    fp.write(str(exit_code))
+                sys.exit(exit_code)
+                """
+            )
+        )
```
AlexWaygood, line 171: You can use the `Path.write_text` method here, which handles opening and closing the file for you:

```suggestion
    uwsgi_script_contents = dedent(
        f"""
        import json
        import os
        import sys
        from mypy.stubtest import main

        sys.argv = json.loads(os.environ.get("STUBTEST_ARGS"))
        exit_code = main()
        with open("{exit_code_surrogate}", mode="w") as fp:
            fp.write(str(exit_code))
        sys.exit(exit_code)
        """
    )
    uwsgi_script.write_text(uwsgi_script_contents)
```

tests/stubtest_third_party.py
```diff
@@ -128,6 +137,84 @@ def run_stubtest(dist: Path, *, verbose: bool = False, specified_platforms_only:
...
+                    "--pyrun",
+                    "{uwsgi_script}",
+                ]
+                ret = subprocess.run(uwsgi_cmd, env=stubtest_env, capture_output=True)
+                print(ret.stdout.decode(), end="", file=sys.stdout)
+                with open("{exit_code_surrogate}", mode="r") as fp:
+                    sys.exit(int(fp.read()))
+                """
+            )
+        )
```
AlexWaygood, line 210: Same here:

```suggestion
    wrapper_script_contents = dedent(
        f"""
        import json
        import os
        import subprocess
        import sys

        stubtest_env = os.environ | {{"STUBTEST_ARGS": json.dumps(sys.argv)}}
        uwsgi_cmd = [
            "{uwsgi_exe}",
            "--ini",
            "{uwsgi_ini}",
            "--spooler",
            "{venv_dir}",
            "--pyrun",
            "{uwsgi_script}",
        ]
        ret = subprocess.run(uwsgi_cmd, env=stubtest_env, capture_output=True)
        print(ret.stdout.decode(), end="", file=sys.stdout)
        with open("{exit_code_surrogate}", mode="r") as fp:
            sys.exit(int(fp.read()))
        """
    )
    wrapper_script.write_text(wrapper_script_contents)
```

tests/stubtest_third_party.py
```diff
@@ -97,6 +98,14 @@ def run_stubtest(dist: Path, *, verbose: bool = False, specified_platforms_only:
         if platform_allowlist.exists():
             stubtest_cmd.extend(["--allowlist", str(platform_allowlist)])
 
+        # Perform some black magic in order to run stubtest inside uWSGI
```
AlexWaygood, line 101:
```suggestion
        # Perform some black magic in order to run stubtest inside uWSGI:
```

tests/stubtest_third_party.py
```diff
@@ -128,6 +137,80 @@ def run_stubtest(dist: Path, *, verbose: bool = False, specified_platforms_only:
     return True
 
 
+# Performs some black magic in order to run stubtest inside uWSGI
+# we have to write the exit code from stubtest to a surrogate file
+# because uwsgi --pyrun does not exit with the exitcode from the
+# python script. We have a second wrapper script that passed the
+# arguments along to the uWSGI script and retrieves the exit code
+# from the file, so it behaves like running stubtest normally would
+def setup_uwsgi_stubtest_command(dist: Path, venv_dir: Path, stubtest_cmd: list[str]) -> bool:
```
AlexWaygood, line 146: Multiline comment -> docstring

```suggestion
def setup_uwsgi_stubtest_command(dist: Path, venv_dir: Path, stubtest_cmd: list[str]) -> bool:
    """Perform some black magic in order to run stubtest inside uWSGI.

    We have to write the exit code from stubtest to a surrogate file
    because uwsgi --pyrun does not exit with the exitcode from the
    python script. We have a second wrapper script that passed the
    arguments along to the uWSGI script and retrieves the exit code
    from the file, so it behaves like running stubtest normally would.
    """
```

tests/stubtest_third_party.py
```diff
@@ -128,6 +137,80 @@ def run_stubtest(dist: Path, *, verbose: bool = False, specified_platforms_only:
...
+        uwsgi_exe = venv_dir / "Scripts" / "uwsgi.exe"
+    else:
+        uwsgi_exe = venv_dir / "bin" / "uwsgi"
+
+    if not uwsgi_exe.exists():
+        print_error("Did not find a uwsgi executable")
+        return False
+
+    # this wrapper script only forwards stdout, since stderr contains
+    # a bunch of spam from uWSGI, if you want to debug the internal script
```
AlexWaygood, line 181:
```suggestion
    # This wrapper script only forwards stdout, since stderr contains
    # a bunch of spam from uWSGI. If you want to debug the internal script,
```

tests/stubtest_third_party.py
```diff
@@ -128,6 +137,80 @@ def run_stubtest(dist: Path, *, verbose: bool = False, specified_platforms_only:
...
+        ]
+        ret = subprocess.run(uwsgi_cmd, env=stubtest_env, capture_output=True)
+        print(ret.stdout.decode(), end="", file=sys.stdout)
+        with open("{exit_code_surrogate}", mode="r") as fp:
+            sys.exit(int(fp.read()))
+        """
+    )
+    wrapper_script.write_text(wrapper_script_contents)
+
+    # modify stubtest_cmd to run our wrapper script instead
```
AlexWaygood, line 208: It would be nice for this comment to be a bit more expansive here: say what `stubtest_cmd[1]` is, and what we're replacing it with

tests/stubtest_third_party.py
```diff
@@ -128,6 +137,80 @@ def run_stubtest(dist: Path, *, verbose: bool = False, specified_platforms_only:
...
+    del stubtest_cmd[1]
+    stubtest_cmd[1] = str(wrapper_script)
```
AlexWaygood, line 210: Slicing might be slightly more elegant here:

```suggestion
    stubtest_cmd[1:3] = [str(wrapper_script)]
```

tests/stubtest_third_party.py
Daverball, line 208:
```suggestion
    # replace "-m mypy.stubtest" in stubtest_cmd with the path to our wrapper script
```

tests/stubtest_third_party.py
```diff
@@ -128,6 +134,81 @@ def run_stubtest(dist: Path, *, verbose: bool = False, specified_platforms_only:
...
+            "--pyrun",
+            "{uwsgi_script}",
+        ]
+        ret = subprocess.run(uwsgi_cmd, env=stubtest_env, capture_output=True)
+        print(ret.stdout.decode(), end="", file=sys.stdout)
+        with open("{exit_code_surrogate}", mode="r") as fp:
+            sys.exit(int(fp.read()))
+        """
+    )
+    wrapper_script.write_text(wrapper_script_contents)
```
AlexWaygood, line 205: What's the advantage of dynamically creating this script here and then executing it, as opposed to just adding a new script to our `tests/` directory and running that script from a subprocess in this script? :)

Daverball: I suppose there is no real advantage. Since stubtest_third_party was already creating temporary venvs it just seemed natural to make the scripts temporary as well. And the scripts are short enough, that I'm not too worried yet about missing out on the linting / static analysis.

But if you feel strongly enough about it I can certainly be persuaded to restructure this accordingly.

Daverball: Moving it into a static script would also add some minor overhead, if you still wanted to be able to run the script outside the temporary venv created by stubtest_third_party.

AlexWaygood: It struck me that, at this point, we were running stubtest from a script being run from a subprocess triggered from a script dynamically generated by `stubtest_third_party.py`. Turtles all the way down!

But I don't feel too strongly about it :) this is all much more readable now that it's factored out into a separate function! I think I'm basically fine with the structure of your changes as they are now.

tests/stubtest_third_party.py
```diff
@@ -128,6 +134,81 @@ def run_stubtest(dist: Path, *, verbose: bool = False, specified_platforms_only:
...
 
+def setup_uwsgi_stubtest_command(dist: Path, venv_dir: Path, stubtest_cmd: list[str]) -> bool:
+    """Perform some black magic in order to run stubtest inside uWSGI.
+
+    We have to write the exit code from stubtest to a surrogate file
+    because uwsgi --pyrun does not exit with the exitcode from the
+    python script. We have a second wrapper script that passed the
+    arguments along to the uWSGI script and retrieves the exit code
+    from the file, so it behaves like running stubtest normally would.
+    """
```
AlexWaygood, line 145: It wasn't immediately obvious to me that the scripts being generated would automatically be cleaned up:

```suggestion

    Both wrapper scripts are created inside `venv_dir`,
    which itself is a subdirectory inside a temporary directory,
    so both generated scripts will be cleaned up after this function
    has been executed.
    """
```

stubs/uWSGI/METADATA.toml
```diff
@@ -0,0 +1,8 @@
+version = "2.0.*"
+extra_description = """\
+  Type hints for uWSGI's \
+  [Python API](https://uwsgi-docs.readthedocs.io/en/latest/PythonModule.html). \
+  Note that this API is available only when running python code inside a uWSGI process \
+  and some parts of the API are only present when corresponding configuration options \
+  have been enabled.
+"""
```
AlexWaygood, line 8:
```suggestion
"""

[tool.stubtest]
# Run stubtest on Windows as well,
# to check that the uWSGI-specific parts of stubtest_third_party.py
# also work on Windows
platforms = ["linux", "win32"]
```

stubs/uWSGI/uwsgi.pyi
```diff
@@ -0,0 +1,241 @@
...
+def connect(__socket_name: str, timeout: int = 0) -> int: ...
+def connection_fd() -> int: ...
+def disconnect() -> None: ...
+def embedded_data(__name: str) -> bytes: ...
+def extract(__name: str) -> bytes | None: ...
+def farm_get_msg() -> bytes | None: ...
+def farm_msg(__farm_name: str, __message: str | bytes | ReadOnlyBuffer) -> None: ...
+def get_logvar(__key: str | bytes) -> bytes | None: ...
+def green_schedule() -> Literal[True]: ...
+def i_am_the_lord() -> bool: ...
```
Akuli, line 137: I'm not going to check each function in detail, but I checked this one because the name sounded funny :D

I found [this doc](https://uwsgi-docs.readthedocs.io/en/latest/Legion.html) which says it should take a name, maybe like this:

```suggestion
def i_am_the_lord(__name: str) -> bool: ...
```

I think stubtest and stubgen don't work very well with some functions implemented in C, if the function doesn't have the right info about its parameters. Maybe check the docs for all functions that take no arguments?

Akuli (commented):
This is almost ready for merging, but we can't merge yet because [stub_uploader](https://github.com/typeshed-internal/stub_uploader) has this code and would probably break:

```
    @property
    def upstream_distribution(self) -> Optional[str]:
        # TODO: add a field to METADATA.toml if the stubs are for a package
        # that does not exist on PyPI
        if self._alleged_upstream_distribution == "gdb":
            return None
        return self._alleged_upstream_distribution
```

This must be fixed first, either by checking for `"uwsgi"` (or `"uWSGI"`? not sure), or even better, by doing what the TODO comment says.

stubs/uWSGI/uwsgi.pyi
AlexWaygood: > I'm not going to check each function in detail, but I checked this one because the name sounded funny :D

I enjoyed the error message we got at the end of a stubtest run on https://github.com/python/typeshed/pull/10432/commits/39acdfe88af6948dfb96b2e9e1f66734a08c39bd:

```
note: unused allowlist entry uwsgi.i_am_the_lord
note: unused allowlist entry uwsgi.lord_scroll
note: unused allowlist entry uwsgi.scrolls
Found 3 errors (checked 2 modules)
worker 1 buried after 0 seconds
goodbye to uWSGI.
Sat Jul 22 18:26:07 2023 - uWSGI spooler 1947 screams: UAAAAAAH my master disconnected: i will kill myself !!!
```

stubs/uWSGI/METADATA.toml
```diff
@@ -0,0 +1,15 @@
+version = "2.0.*"
+upstream_repository = "https://github.com/unbit/uwsgi"
+extra_description = """\
+  Type hints for uWSGI's \
+  [Python API](https://uwsgi-docs.readthedocs.io/en/latest/PythonModule.html). \
+  Note that this API is available only when running python code inside a uWSGI process \
```
JelleZijlstra, line 6:
```suggestion
  Note that this API is available only when running Python code inside a uWSGI process \
```

stubs/uWSGI/uwsgidecorators.pyi
```diff
@@ -0,0 +1,180 @@
+from collections.abc import Callable
+from typing import Any, Generic, TypeVar, overload
+from typing_extensions import Literal, ParamSpec
+
+from uwsgi import _RPCCallable
+
+_T = TypeVar("_T")
+_T2 = TypeVar("_T2")
+_SR = TypeVar("_SR", bound=Literal[0, -1, -2] | None)
+_SignalCallback = TypeVar("_SignalCallback", bound=Callable[[int], Any])
```
JelleZijlstra, line 10: Suggest renaming this to `_SignalCallbackT` so it's clear that it's a TypeVar, not an alias.

stubs/uWSGI/uwsgi.pyi
Daverball: You are correct. That probably was my bad and I just skipped over the call to `PyArgParseTuple` when I was adding annotations (possibly because I was elated to see `Py_False`, rather than their weird `True` or `None` situation they've got going on everywhere). But I will double check that all the functions that take no arguments actually take no arguments, in case stubgen has some weird corner case there.

IIRC stubgen just punts and annotates everything with `*args, **kwargs` unless the docstring contains the callspec and then it will just trust that the docstring is correct, without checking it. It doesn't seem like it performs any static analysis on the C code, otherwise it would be possible to at least derive partial information from the method flags.

Daverball:
> This is almost ready for merging, but we can't merge yet because [stub_uploader](https://github.com/typeshed-internal/stub_uploader) has this code and would probably break:
> 
> ```python
>     @property
>     def upstream_distribution(self) -> Optional[str]:
>         # TODO: add a field to METADATA.toml if the stubs are for a package
>         # that does not exist on PyPI
>         if self._alleged_upstream_distribution == "gdb":
>             return None
>         return self._alleged_upstream_distribution
> ```
> 
> This must be fixed first, either by checking for `"uwsgi"` (or `"uWSGI"`? not sure), or even better, by doing what the TODO comment says.

I'm not sure that's true, the uWSGI distribution does exist on pypi. It just does not contain a `uwsgi` package, unless the code is executed by `uwsgi`. But I will double check and look at what this code does, to judge whether or not it actually will cause problems.

stubs/uWSGI/uwsgi.pyi
```diff
@@ -0,0 +1,241 @@
...
+def micros() -> int: ...
+def mule_get_msg(signals: bool = True, farms: bool = True, buffer_size: int = 65536, timeout: int = -1) -> bytes: ...
+def mule_id() -> int: ...
+@overload
+def mule_msg(__mesage: str | bytes | ReadOnlyBuffer) -> bool: ...
+@overload
+def mule_msg(__mesage: str | bytes | ReadOnlyBuffer, __mule_id: int) -> bool: ...
+@overload
+def mule_msg(__mesage: str | bytes | ReadOnlyBuffer, __farm_name: str) -> bool: ...
+def offload() -> Literal[b""]: ...
```
Daverball, line 167: This has been commented out, since it presumably isn't necessary anymore, so it's a no-op, but if we want to be backwards compatible, we should probably annotate the arguments it was going to parse, which was `__filename: str, __len: int = 0`

stubs/uWSGI/uwsgi.pyi
```diff
@@ -0,0 +1,241 @@
...
+def connect(__socket_name: str, timeout: int = 0) -> int: ...
+def connection_fd() -> int: ...
+def disconnect() -> None: ...
+def embedded_data(__name: str) -> bytes: ...
+def extract(__name: str) -> bytes | None: ...
+def farm_get_msg() -> bytes | None: ...
+def farm_msg(__farm_name: str, __message: str | bytes | ReadOnlyBuffer) -> None: ...
+def get_logvar(__key: str | bytes) -> bytes | None: ...
+def green_schedule() -> Literal[True]: ...
+def i_am_the_lord(__name: str) -> bool: ...
```
Daverball, line 137: Let's call this `__legion_name` for consistency's sake.

Daverball (commented):
The rest of the calls without arguments seemed fine, apart from `offload`

Akuli:
I found this thing: https://uwsgiconf.readthedocs.io/en/latest/uwsgi_stub.html

It is apparently "stubs" in a `.py` file. A bit weird/non-standard, and doesn't seem like it would work with type checkers. They don't use `typing.TYPE_CHECKING` to detect whether to bring in the stub or the real uwsgi, but instead assume that `if False: from .uwsgi_stub import *` will be picked up by the IDE or type checker.

I think this PR is still a good idea, because:
- Stubs from this PR will "just work" without changing your imports.
- Stubs from this PR will work with type checkers.

Their `.py` stub file is here: https://github.com/idlesign/uwsgiconf/blob/master/uwsgiconf/uwsgi_stub.py We could compare it to the stubs in this PR to see if anything needs improving.

I think after that we can merge :)

Daverball:
Interesting, it looks like the reason it isn't just a stub file, is so that you can still use the symbols at runtime, regardless of whether the application is run inside uWSGI and instead relies on `is_stub` to detect whether uWSGI is really there or if it's just a stub module where none of the functions do anything.

This seems like kind of a bad idea to me. At the very least I would expect most of the function implementations that intend to modify some uWSGI internal state to raise `NotImplementedError`, otherwise it would  be very easy to fool yourself into thinking something is working, when it actually isn't.

I took a quick look through the stubs and I couldn't spot any differences besides the fact that they're a little bit more loose, they incorrectly use named parameters, when actually most of these are positional only, they also incorrectly use a bool return type for a lot of these functions when it is actually `Literal[True] | None`, which is strange, but it's what they return. Also it looks like they incorrectly used HEAD as a reference for some of the stubs, rather than the latest version, because `request_context` is an upcoming parameter (and type) that has not been added yet.

Daverball:
I took a more detailed look and did find a few mistakes, so this was a good shout. The other discrepancies are either their mistake or them using HEAD as reference, rather than the release tag for the current version.

stubs/uWSGI/uwsgi.pyi
```diff
@@ -0,0 +1,241 @@
...
+_TrueOrNone: TypeAlias = Literal[True] | None
+
+class _RPCCallable(Protocol):
+    def __call__(self, *args: bytes) -> bytes | None: ...
+
+# FIXME: Technically we know the exact layout of _AppsDict and _WorkerDict
+#        but TypedDict does not support bytes keys, so for now we use type
+#        aliases to a more generic dict
+_AppsDict: TypeAlias = dict[bytes, int | bytes]
+_WorkerDict: TypeAlias = dict[bytes, int | bytes | tuple[_AppsDict, ...]]
```
Akuli, line 19: These will be a bit annoying to use, because you need to do type narrowing with `assert`:

```python
value = worker_dict[b"foo"]
assert isinstance(value, int)
# use the value
```

So I think this is the best we can do:

```suggestion
_WorkerDict: TypeAlias = dict[bytes, Any]
```

stubs/uWSGI/uwsgi.pyi
```diff
@@ -0,0 +1,241 @@
...
+applications: dict[str, WSGIApplication | str] | None
+buffer_size: int
+cores: int
+has_threads: int
+hostname: bytes
+is_a_reload: bool
+loop: bytes | None
+magic_table: dict[bytes, bytes]
+numproc: int
+opt: dict[str, bytes | Literal[True] | list[bytes | Literal[True]]]
```
Akuli, line 33: Is this dict meant to be used for setting options or for getting their values? If it's meant mostly for getting, this too could be better with `Any`. Same goes for the `applications` dict.

Akuli (approved):
Found a couple more nits, but IMO this could already be merged. I haven't checked everything, but the stubtest setup looks good and the `uwsgiconf.uwsgi_stub` comparison gives more confidence.

stubs/uWSGI/uwsgi.pyi
Daverball: Yeah I've debated with myself about whether or not I should use a union here. I'm fine with going this route for now. Once either `AnyOf` or `TypedDict`/`TypedMapping` with byte keys become a thing we can make it more specific.

stubs/uWSGI/uwsgi.pyi
Daverball: As far as I can tell it's purely for reading options, yes. I'm fine with changing it, but it seemed helpful to know that the values would be bytes, even though the key is a regular string, which is usually not the case in uWSGI.

Akuli:
Thanks!

Issue #10438, opened 2023-07-11
wrong condition for EQFULL constant within stdlib/errno.pyi


s04lg934jkg:
EQFULL is defined not only for MacOS in this code
https://github.com/python/typeshed/blob/f812d71ffa2662a0291f8e5e32a7e2ff082a8af1/stdlib/errno.pyi#L117

if sys.platform != "win32":
    ...
    if sys.platform != "darwin" or sys.version_info >= (3, 11):
        EQFULL: int  # undocumented

It looks like the intention was to defined EQFULL for MacOS starting from python 3.11, but in fact it is also defined for any system that is not Windows or Mac.

srittau:
I can confirm that at least with Ubuntu and Python 3.11 the errno does not exist. Cc @AlexWaygood who submitted the original PR.

AlexWaygood:
> Cc @AlexWaygood who submitted the original PR.

Prior to #8142 it was defined unconditionally on all platforms and all versions. I changed the availability on Windows and macOS in #8142, but left the availability on Linux as it was as I don't have a Linux box to test on.

AlexWaygood:
The original PR in which it was added was #7553 by @JelleZijlstra

Akuli:
Doesn't exist on linux for me either. Tried Python 3.8, 3.9, 3.10 and 3.11.

srittau:
I'll send a PR to update errno on Linux later.

AlexWaygood:
Fixed by #10440

PR #10440, opened 2023-07-11
Mark some errnos as darwin-only

srittau:
Remove errnos that don't seem to exist on any platform

### DIFF
--- a/stdlib/errno.pyi
+++ b/stdlib/errno.pyi
@@ -115,7 +115,7 @@ if sys.platform == "darwin":
     ERPCMISMATCH: int
     ESHLIBVERS: int
     if sys.version_info >= (3, 11):
-        EQFULL: int  # undocumented
+        EQFULL: int
 
 if sys.platform != "darwin":
     EDEADLOCK: int
--- a/stdlib/errno.pyi
+++ b/stdlib/errno.pyi
@@ -91,6 +91,10 @@ ECANCELED: int  # undocumented
 ENOTRECOVERABLE: int  # undocumented
 EOWNERDEAD: int  # undocumented
 
+if sys.platform == "sunos5" or sys.platform == "solaris":
+    ELOCKUNMAPPED: int
+    ENOTACTIVE: int
+
 if sys.platform != "win32":
     ENOTBLK: int
     EMULTIHOP: int
--- a/stdlib/errno.pyi
+++ b/stdlib/errno.pyi
@@ -94,6 +94,8 @@ EOWNERDEAD: int  # undocumented
 if sys.platform != "win32":
     ENOTBLK: int
     EMULTIHOP: int
+
+if sys.platform == "darwin":
     # All of the below are undocumented
     EAUTH: int
     EBADARCH: int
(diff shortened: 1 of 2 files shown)


srittau:
I checked by doing "import errnos; dir(errnos)" on all Python versions between 3.7 and 3.12 (there were no differences) and then comparing to errnos.pyi.

srittau:
Closes: #10438

stdlib/errno.pyi
```diff
@@ -164,9 +165,6 @@ if sys.platform != "win32" and sys.platform != "darwin":
     ENOKEY: int
     ENOMEDIUM: int
     ERFKILL: int
-    EL: int
-    ELOCKUNMAPPED: int
-    ENOTACTIVE: int
```
AlexWaygood, line 169: According to comments in the source code, `ELOCKUNMAPPED` and `ENOTACTIVE` do exist, but only on Solaris: https://github.com/python/cpython/blob/388b5daa523b828dc0f7e2a1a6886bebc20833ba/Modules/errnomodule.c#L860-L878

I can't find any reference to `EL` in the source code, however.

AlexWaygood: We should probably do something like:

```py
if sys.platform != "win32" and sys.platform != "darwin" and sys.platform != "linux":
    # these are Solaris-specific:
    ELOCKUNMAPPED: int
    ENOTACTIVE: int
```

srittau: I don't even know which `sys.platform` Solaris uses. ("solaris"? "sunos"?) I can re-add them, but we have no real support for Solaris in typeshed outside these errnos.

AlexWaygood: > I don't even know which `sys.platform` Solaris uses. ("solaris"? "sunos"?)

Nor do I!

> I can re-add them, but we have no real support for Solaris in typeshed outside these errnos.

We have plenty of other code in typeshed that's hardly tested (our gdb stubs come to mind, which we have to skip stubtest for), so the fact that we don't test on solaris doesn't feel like a great argument for _removing_ these errnos, since they're already here :)

srittau: I'd prefer a specific platform tag in that case. I'm not sure where to look, though. The `PLATFORM` C define is used to set `sys.platform`, but it's not defined directly in the source files. I assume it's set by the compiler and passed using the autotools. I'm looking into it.

srittau: I've found both `solaris` and `sunos5` (which was renamed "Solaris") in Python's source code, so I'm using those for now.

stdlib/errno.pyi
```diff
@@ -91,9 +91,15 @@ ECANCELED: int  # undocumented
 ENOTRECOVERABLE: int  # undocumented
 EOWNERDEAD: int  # undocumented
 
+if sys.platform == "sunos5" or sys.platform == "solaris":
```
AlexWaygood, line 94: flake8-pyi is just flagging this because it doesn't know about these platforms. I _think_ they should work fine with type checkers, so we should just `noqa` them?

```suggestion
if sys.platform == "sunos5" or sys.platform == "solaris":  # noqa: Y008
```

JelleZijlstra: We can also add these platforms to the list of supported platforms in flake8-pyi. I think I came up with that list years ago purely based on what was in use in typeshed at the time.

srittau: I'll noqa them for now. If we add them to flake8-pyi, we should double check whether "solaris" is actually a valid platform. It's only used in CPython source code, in a tools, while "sunos5" is used multiple times.

AlexWaygood (approved):
Thanks!

PR #10441, opened 2023-07-11
Allow exception value as first arg for format_exception_only in 3.10

Numerlor:
Passing in an exception object directly as the only arg is valid since 3.10
https://docs.python.org/3/library/traceback.html#traceback.format_exception_only https://github.com/python/cpython/blob/3.10/Lib/traceback.py#L139-L156


srittau:
I haven't looked at it, but in that case, would it make sense to have two overloads, one which *only* takes an exception and one where it takes an exception type and a second argument?

Numerlor:
> I haven't looked at it, but in that case, would it make sense to have two overloads, one which _only_ takes an exception and one where it takes an exception type and a second argument?

Yeah that'd work too, the function is currently weird in that it's pretty much a single arg func that can be called as func(arg) and func(_, arg). I was thinking the frist arg in the two arg form could be object as it's just ignored but that would probably just add confusion

srittau:
> Yeah that'd work too, the function is currently weird in that it's pretty much a single arg func that can be called as func(arg) and func(_, arg). I was thinking the frist arg in the two arg form could be object as it's just ignored but that would probably just add confusion

We have a special type alias `_typeshed.Unused` to `object` exactly to signal the fact to the user that an argument is ignored.

srittau (approved):
Thanks!

PR #10411, opened 2023-07-05
Bump Pillow to 10.0.*

srittau:
Depends on #10409.

Closes: #10394

PR #10445, opened 2023-07-12
Pillow: mark MAX_IMAGE_PIXELS as non-final

DMRobertson:
I don't think it was correct for #10411 to mark MAX_IMAGE_PIXELS as final. Quoting my reasoning from https://github.com/python/typeshed/pull/10411#discussion_r1261206232:

> According to [pillow.readthedocs.io/en/stable/reference/Image.html#PIL.Image.open](https://pillow.readthedocs.io/en/stable/reference/Image.html#PIL.Image.open):
>
>>    This threshold can be changed by setting [MAX_IMAGE_PIXELS](https://pillow.readthedocs.io/en/stable/reference/Image.html#PIL.Image.MAX_IMAGE_PIXELS). It can be disabled by setting Image.MAX_IMAGE_PIXELS = None.
>
> I can't see any mention of this [MAX_IMAGE_PIXELS] in [pillow.readthedocs.io/en/stable/releasenotes/10.0.0.html](https://pillow.readthedocs.io/en/stable/releasenotes/10.0.0.html) either.

### DIFF
--- a/stubs/Pillow/PIL/Image.pyi
+++ b/stubs/Pillow/PIL/Image.pyi
@@ -3,7 +3,7 @@ from collections.abc import Callable, Iterable, Iterator, MutableMapping, Sequen
 from enum import IntEnum
 from pathlib import Path
 from typing import Any, ClassVar, Protocol, SupportsBytes
-from typing_extensions import Final, Literal, Self, TypeAlias, TypeGuard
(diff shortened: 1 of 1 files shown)


AlexWaygood (approved):
Thanks!

PR #10446, opened 2023-07-12
add vtk stub files

### LINKED ISSUE #10442: Can we add vtk stub file ?
I am asking because the python package doesn't technically ship its stub files but allows us to generate them with a python command (see [vtkmodules/generate_pyi.py](https://docs.vtk.org/en/latest/api/python/vtkmodules/vtkmodules.generate_pyi.html))

So I don't know if it respect the third point in the requirements:
> the package does not ship with its own stubs or type annotations. 

Their module also don't generate stub files for all .py files (and in particular for `__init__.py`)

Gryfenfer97:
These are the stub files generated by the vtk script [generate_pyi.py](https://docs.vtk.org/en/latest/api/python/vtkmodules/vtkmodules.generate_pyi.html)
Should fix  #10442 

I am cleaning them and test them to be sure that it works
TODO:

**Cleaning**:
- [ ] vtkmodules

**Generating**
- [ ] gtk/
- [ ] numpy_interface/
- [ ] qt/
- [ ] tk/
- [ ] util/
- [ ] web/
- [ ] wx/
- [ ] tests


AlexWaygood:
pre-commit.ci autofix

Gryfenfer97:
I may need some help for this because the stub file generated doesn't seem to match valid python code and this generate an error
I have this piece of code in the generated stub file (the comment is the error generated by mypy):
```py
class vtkTransform(vtkLinearTransform):
    @overload  # Overload does not consistently use the "@staticmethod" decorator on all function signatures.
    def GetOrientation(self, orient: MutableSequence[float]) -> None: ...
    @overload
    def GetOrientation(self) -> tuple[float, float, float]: ...
    @overload
    @staticmethod
    def GetOrientation(orient: MutableSequence[float], matrix: vtkmodules.vtkCommonMath.vtkMatrix4x4) -> None: ...
```

Which, I guess, should correspond to the following python code
```py
class vtkTransform(vtkLinearTransform):
    def GetOrientation(self, orient = None):
        pass

    @staticmethod
    def GetOrientation(orient, matrix) -> None:
        pass
```

But this isn't valid python code, I have made this piece of code to test the behavior of the python interpreter (the comment is the error generated by my python interpreter):
```py
class Foo:
    def bar(self) -> int: 
        return 1

    @staticmethod
    def bar(i: str) -> None:
        pass


print(type(Foo.bar("abc")))
print(type(Foo().bar()))  # TypeError: Foo.bar() missing 1 required positional argument: 'i'
```

So I don't understand how they manage to do this since both methods works in the case of vtk (comment correspond to the output of the script):
```py
import vtk.vtkCommonTransforms
import vtk.vtkCommonMath

transform = vtk.vtkCommonTransforms.vtkTransform()
print(transform.GetOrientation())  # (0.0, -0.0, 0.0)
print(vtk.vtkCommonTransforms.vtkTransform.GetOrientation([0, 0, 0], vtk.vtkCommonMath.vtkMatrix4x4()))  # None
```
Can the behavior of python be different between `.py` scripts and generated c++ libraries ? Or did I wrongly assume the python code behind this ? and if so what could I do to fix the error ?

JelleZijlstra:
I started looking at this with an eye towards getting CI to pass, since this is one of the oldest open PRs. Unfortunately, there is a really huge amount of stubtest errors, more than I can deal with. This probably needs someone with deeper familiarity with the library.

JelleZijlstra:
Thanks for contributing! I'm closing this PR for now, because it still fails some tests
after more than a year of inactivity. If you are still interested, please feel free to open

PR #10455, opened 2023-07-13
tkinter: accept Iterable[int] for grid_columnconfigure, grid_rowconfigure

### LINKED ISSUE #9784: tkinter: grid_rowconfigure and grid_columnconfigure can take a list/tuple
This creates a row of 10 buttons that stay evenly spaced as the window is resized:

```py
import tkinter

root = tkinter.Tk()
for i in range(10):
    tkinter.Button(root, text="hi").grid(row=0, column=i)
root.grid_columnconfigure(list(range(10)), weight=1)
root.mainloop()
```

But mypy doesn't like passing a list to `grid_columnconfigure()`: `asd.py:6: error: Argument 1 to "grid_columnconfigure" of "Misc" has incompatible type "List[int]"; expected "Union[int, str]"  [arg-type]`

Also applies to `grid_rowconfigure`. The code also works if I change `list` to `tuple`, but similar to many other things in tkinter, it needs a list or tuple rather than an arbitrary sequence (so `grid_columnconfigure(range(10), ...)` doesn't work).

brccabral:
missing typing hint Iterable[int] for _GridIndex

Fixes #9784


stdlib/tkinter/__init__.pyi
```diff
@@ -179,7 +179,7 @@ _Compound: TypeAlias = Literal["top", "left", "center", "right", "bottom", "none
 _Cursor: TypeAlias = str | tuple[str] | tuple[str, str] | tuple[str, str, str] | tuple[str, str, str, str]
 # example when it's sequence:  entry['invalidcommand'] = [entry.register(print), '%P']
 _EntryValidateCommand: TypeAlias = str | list[str] | tuple[str, ...] | Callable[[], bool]
-_GridIndex: TypeAlias = int | str
+_GridIndex: TypeAlias = int | str | Iterable[int]
```
Akuli, line 182: As the name suggests, `_GridIndex` represents one index, not a sequence of indexes. Instead, you should change `grid_rowconfigure()` and `grid_columnconfigure()`, because it's just those two methods that accept lists and tuples. I explained this in more detail on #9784.

Also, my "only works with lists and tuples" comment in your other PR applies here too.

stdlib/tkinter/__init__.pyi
```diff
@@ -500,7 +500,7 @@ class Misc:
     bbox = grid_bbox
     def grid_columnconfigure(
         self,
-        index: _GridIndex,
+        index: _GridIndex | list[str] | list[int] | list[str | int] | tuple[str | int, ...],
```
Akuli, line 503: The [manual page for `grid`](https://www.tcl.tk/man/tcl/TkCmd/grid.html) says this:

> Indices may be integers, window  names or  the  keyword `all`. For `all` the options apply to all rows currently occupied by content windows. For a window name, that window  must  be a content window of this container and the optionsapply to all rows currently occupied by  the  container  window.

So because the only acceptable string  is `"all"`, it doesn't make sense to create a list of strings for this.

```suggestion
        index: _GridIndex | list[int] |tuple[int, ...],
```

Or do you have some use case where you actually want a list of strings?

Apparently these also accept widgets. (Tkinter widgets are wrappers for Tcl's window names, so you can just imagine "widget" when it says "window name".) We don't have to do it in this PR though.

stdlib/tkinter/__init__.pyi
```diff
@@ -510,7 +510,7 @@ class Misc:
     ) -> _GridIndexInfo | Any: ...  # can be None but annoying to check
     def grid_rowconfigure(
         self,
-        index: _GridIndex,
+        index: _GridIndex | list[str] | list[int] | list[str | int] | tuple[str | int, ...],
```
Akuli, line 513:
```suggestion
        index: _GridIndex | list[int] | tuple[int, ...],
```

Akuli (approved):
Thanks again!

(For other maintainers, feel free to merge this when the CI is green if you get here before me)

PR #10456, opened 2023-07-13
Accept ints for tkinter.ttk.Treeview columns

brccabral:
added Iterable[int] to Treeview columns typing hint

### DIFF
--- a/stdlib/tkinter/ttk.pyi
+++ b/stdlib/tkinter/ttk.pyi
@@ -2,9 +2,9 @@ import _tkinter
 import sys
 import tkinter
 from _typeshed import Incomplete
-from collections.abc import Callable
+from collections.abc import Callable, Iterable
 from tkinter.font import _FontDescription
-from typing import Any, overload
+from typing import Any, Iterable, overload
 from typing_extensions import Literal, TypeAlias, TypedDict
 
 __all__ = [
@@ -961,7 +961,7 @@ class Treeview(Widget, tkinter.XView, tkinter.YView):
         master: tkinter.Misc | None = None,
         *,
         class_: str = ...,
-        columns: str | list[str] | tuple[str, ...] = ...,
+        columns: str | Iterable[int] | list[str] | tuple[str, ...] = ...,
         cursor: tkinter._Cursor = ...,
         displaycolumns: str | list[str] | tuple[str, ...] | list[int] | tuple[int, ...] = ...,
         height: int = ...,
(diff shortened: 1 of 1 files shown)


stdlib/tkinter/ttk.pyi
```diff
@@ -961,7 +961,7 @@ class Treeview(Widget, tkinter.XView, tkinter.YView):
         master: tkinter.Misc | None = None,
         *,
         class_: str = ...,
-        columns: str | list[str] | tuple[str, ...] = ...,
+        columns: str | Iterable[int] | list[str] | tuple[str, ...] = ...,
```
Akuli, line 964: This doesn't really do the right thing with all iterables. For example, `collections.deque` is an iterable, but it does this:

```python
>>> from tkinter.ttk import Treeview
>>> from collections import deque
>>> tv = Treeview(columns=deque([1, 2, 3]))
>>> tv['columns']
('deque([1,', '2,', '3])')
```

So tkinter converts the `deque` instance into a string, and then splits that string by spaces. Instead of a column named `"1"`, you got a column named `"deque([1,"`.

In fact, only lists and tuples work. This is very common in tkinter. How about:
```suggestion
        columns: str | list[str] | list[int] | list[str | int] | tuple[str | int, ...] = ...,
```

You would expect `list[str | int]` to cover both `list[str]` and `list[int]`, but that's unfortunately not how it works. The reason is basically that `.append(int)` is a valid operation for `list[str | int]`, but not for `list[str]`, so a function that takes `list[str | int]` could do bad things when given a `list[str]`. With tuples it is enough to specify `tuple[str | int, ...]`, because you cannot append to (or otherwise modify) a tuple.

brccabral: Probably the same happens in #10455 , right?

stdlib/tkinter/ttk.pyi
```diff
@@ -983,7 +983,7 @@ class Treeview(Widget, tkinter.XView, tkinter.YView):
         self,
         cnf: dict[str, Any] | None = None,
         *,
-        columns: str | list[str] | tuple[str, ...] = ...,
+        columns: str | Iterable[int] | list[str] | tuple[str, ...] = ...,
```
Akuli, line 986: This needs updating too.

Akuli (approved):
Thanks!

stdlib/tkinter/ttk.pyi
Akuli: Yes, the same thing happens there too.

PR #10458, opened 2023-07-13
Add some ruff autofixes to CI

AlexWaygood:
The aim of this PR is to make it easier for new contributors to typeshed to comply with our -- at times fairly stringent! -- style guide. For large PRs (e.g. #10446, or #10355), this can be quite difficult. Ruff has a number of autofixes that we can use to help out with this task, and it's extremely fast.

Unlike most projects, we're not likely to ever _switch_ from flake8 to ruff -- even if ruff implemented the full set of flake8-pyi rules and had exactly the same behaviour as flake8-pyi for all of them, it's useful for us to be able to add new rules to our own linter whenver we want to. However, there's nothing stopping us from having ruff as part of our CI _as well as_ flake8.

Since (in my opinion) we're _only_ interested in the autofixes here, I've only enabled a few rules that have relatively safe autofixes relevant to stub files.

Since ruff includes isort as one of its autofixes, having isort as a test dependency is now redundant, so I've removed it in favour of ruff's I001 rule.

### DIFF
--- a/scripts/create_baseline_stubs.py
+++ b/scripts/create_baseline_stubs.py
@@ -60,9 +60,9 @@ def run_black(stub_dir: str) -> None:
     subprocess.run(["black", stub_dir])
 
 
-def run_isort(stub_dir: str) -> None:
-    print(f"Running isort: isort {stub_dir}")
-    subprocess.run([sys.executable, "-m", "isort", stub_dir])
+def run_ruff(stub_dir: str) -> None:
+    print(f"Running ruff: ruff {stub_dir}")
+    subprocess.run([sys.executable, "-m", "ruff", stub_dir])
 
 
 def create_metadata(stub_dir: str, version: str) -> None:
@@ -110,7 +110,7 @@ def add_pyright_exclusion(stub_dir: str) -> None:
 def main() -> None:
     parser = argparse.ArgumentParser(
         description="""Generate baseline stubs automatically for an installed pip package
-                       using stubgen. Also run black and isort. If the name of
+                       using stubgen. Also run black and ruff. If the name of
                        the project is different from the runtime Python package name, you may
                        need to use --package (example: --package yaml PyYAML)."""
     )
@@ -159,7 +159,7 @@ def main() -> None:
     run_stubgen(package, stub_dir)
     run_stubdefaulter(stub_dir)
 
-    run_isort(stub_dir)
+    run_ruff(stub_dir)
     run_black(stub_dir)
 
     create_metadata(stub_dir, version)
--- a/scripts/create_baseline_stubs.py
+++ b/scripts/create_baseline_stubs.py
@@ -164,8 +164,8 @@ def main() -> None:
     run_stubgen(package, stub_dir)
     run_stubdefaulter(stub_dir)
 
-    run_isort(stub_dir)
     run_ruff(stub_dir)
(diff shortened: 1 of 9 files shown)


srittau:
I'm not a fan of replacing a well-tested, easily usable tool like isort with a new tool that's still under heavy development, like ruff. (And while I'm not a fan of flake8, I'm also not a fan of implementing a linter in a different language for marginal speed wins.)

srittau:
And just to clarify: I'm not opposed to using ruff auto-fixes, except for those handled by isort. But I would like to keep isort, at least for now.

AlexWaygood:
I have some quibbles with your characterisations of ruff vs isort, but I also don't feel strongly at all about replacing isort with ruff (I just thought it would be nice to reduce the number of tools we rely on in CI). So I'll revert replacing isort with ruff, and concentrate on just adding some new autofixes to CI.

srittau:
One of the advantages of using isort is that tooling already works, for example, VSCode has isort support integrated. And ruff is quite early in its development with a <1 version number, so I wouldn't expect it to be stable in the fixes it outputs. Maybe in the future it makes sense to replace isort, because in general I support the idea to use less tools.

AlexWaygood:
> One of the advantages of using isort is that tooling already works, for example, VSCode has isort support integrated. And ruff is quite early in its development with a <1 version number, so I wouldn't expect it to be stable in the fixes it outputs. Maybe in the future it makes sense to replace isort, because in general I support the idea to use less tools.

Sure. After my last push, this PR no longer proposes replacing isort with ruff, just adding a few new ruff autofixes on top of isort.

Nonetheless, to flesh out my quibbles for you: my impression is that ruff's import-sorting is reasonably stable at this point, and that they're very quick to fix bugs when they're filed against the project. I also like that ruff has very regular releases, whereas isort releases are very irregular (and I'd wager they're likely to become increasingly irregular, since the creator of isort has said that he's a fan of ruff these days). And isort [definitely isn't bug-free](https://github.com/PyCQA/isort/issues/2154) (which I know isn't a claim you ever made, but I think speaks to the value of using a tool with regular releases). Personally I find ruff to be just as easily usable as isort -- but then again, I'm not a heavy user of VSCode, so I can't really speak to that aspect.

FWIW, I was also pretty sceptical initially about the idea of a Python development tool written in a different language, just for a bit of a speedup. But I think my scepticism has been proven wrong there, at least for now -- they seem to be getting a lot of PRs from the community, and it seems to be a pretty thriving project. We'll see how it plays out in the long term, of course.

scripts/create_baseline_stubs.py
```diff
@@ -160,6 +165,7 @@ def main() -> None:
     run_stubdefaulter(stub_dir)
 
     run_isort(stub_dir)
+    run_ruff(stub_dir)
```
sobolevn, line 168:
```suggestion
    run_ruff(stub_dir)
    run_isort(stub_dir)
```

I think that we should call `ruff` as the first one. Because it change semantics, not only style.

By the way, `isort` can also be later replaced with `ruff`.

sobolevn: Oh, I see https://github.com/python/typeshed/pull/10458#issuecomment-1642008989

srittau (approved):

Issue #10457, opened 2023-07-13
incorrect types for functions decorated with `@mock.patch`


svalentin:
Looks like we have a typeshed bug:
```
from unittest.mock import patch

def addition(a: int, b:int) -> int:
    return a+b

@patch('test.addition')
def patch_test(mock):
    print(mock)

if __name__ == "__main__":
    patch_test()
```

In this simple example mypy says:
```
test.py:12: error: Missing positional argument "mock" in call to "patch_test"  [call-arg]
```

Undoing https://github.com/python/typeshed/commit/9e86c6026a00a7bcc6c77ec8f7925a99930ee983 fixes it

PR #10459, opened 2023-07-13
Fix @patch when `new` is missing

srittau:
Closes: #10457

### DIFF
--- a/stdlib/unittest/mock.pyi
+++ b/stdlib/unittest/mock.pyi
@@ -2,7 +2,7 @@ import sys
 from collections.abc import Awaitable, Callable, Coroutine, Iterable, Mapping, Sequence
 from contextlib import _GeneratorContextManager
 from types import TracebackType
-from typing import Any, Concatenate, Generic, TypeVar, overload
+from typing import Any, Generic, TypeVar, overload
 from typing_extensions import Final, Literal, ParamSpec, Self, TypeAlias
 
 _T = TypeVar("_T")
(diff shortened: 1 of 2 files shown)


JukkaL (approved):

svalentin:
Thank you!

stdlib/unittest/mock.pyi
```diff
@@ -257,6 +259,22 @@ class _patch(Generic[_T]):
...
+else:
+    _Mock: TypeAlias = MagicMock
+
+# This class does not exist at runtime, it's a hack to make this work:
+#     @patch("foo")
+#     def bar(..., mock: MagicMock) -> None: ...
+class _patch_default_new(_patch[_Mock]):
+    @overload
+    def __call__(self, func: _TT) -> _TT: ...
+    # Can't use the following as ParamSpec is only allowed as last parameter:
```
AlexWaygood, line 273:
```suggestion
    # Can't use the following as ParamSpec is only allowed as the last parameter to Concatenate[]:
```

test_cases/stdlib/check_unittest.py
```diff
@@ -94,13 +94,20 @@ def __gt__(self, other: Bacon) -> bool:
...
+def f_explicit_new(i: int) -> str:
     return "asdf"
 
 
-assert_type(f(1), str)
-f("a")  # type: ignore
+assert_type(f_default_new(1), str)
+f_default_new("a")  # Not an error due to ParamSpec limitations
+assert_type(f_explicit_new(1), str)
+f_explicit_new("a")  # type: ignore[arg-type]
```
AlexWaygood, line 110: Nit: we currently don't use mypy error codes for `type: ignore`s in the `test_cases` directory: (see rationale in #8083)

```suggestion
f_explicit_new("a")  # type: ignore
```

AlexWaygood (approved):

PR #10461, opened 2023-07-14
[stripe] Add type for PaymentLink

anniel-stripe:
Adds type checking for the [PaymentLink](https://stripe.com/docs/api/payment_links/payment_links) resource, which is defined in stripe-python [here](https://github.com/stripe/stripe-python/blob/4829b884cf817dacd4d3217e40560693e26a7f8f/stripe/api_resources/payment_link.py).

Similar to https://github.com/python/typeshed/pull/10323 and https://github.com/python/typeshed/pull/10416

### DIFF
--- a/stubs/stripe/stripe/api_resources/payment_link.pyi
+++ b/stubs/stripe/stripe/api_resources/payment_link.pyi
@@ -11,7 +11,14 @@ class PaymentLink(CreateableAPIResource, ListableAPIResource, UpdateableAPIResou
 
     @overload
     @classmethod
-    def list_line_items(cls, payment_link, api_key=None, stripe_version=None, stripe_account=None, **params): ...
+    def list_line_items(
+        cls,
+        payment_link: str,
+        api_key: str | None = None,
+        stripe_version: str | None = None,
+        stripe_account: str | None = None,
+        **params,
+    ): ...
     @overload
     @classmethod
-    def list_line_items(cls, idempotency_key=None, **params): ...
+    def list_line_items(cls, idempotency_key: str | None = None, **params): ...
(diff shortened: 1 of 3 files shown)


stubs/stripe/stripe/api_resources/payment_link.pyi
```diff
@@ -0,0 +1,10 @@
+from stripe.api_resources.abstract import (
+    CreateableAPIResource as CreateableAPIResource,
+    ListableAPIResource as ListableAPIResource,
+    UpdateableAPIResource as UpdateableAPIResource,
+)
+
+class PaymentLink(CreateableAPIResource, ListableAPIResource, UpdateableAPIResource):
+    OBJECT_NAME: str
+
+    def list_line_items(self, idempotency_key=None, **params): ...
```
AlexWaygood, line 10: This looks like it might be one of those `stripe` methods that's sometimes an instance method and sometimes a classmethod. Am I correct?

anniel-stripe: Yes it is! Is there a preferred way to represent that? I looked through some of the other resources types and didn't find other special handling of `list_line_items` or similar methods.

AlexWaygood: There's no perfect way of representing these in the stubs, unfortunately, as it's a bit too dynamic for type checkers to understand properly! #9814 is a good example of the hack we've been using for some of the other methods `stripe` has that are like this, though.

(As with #9814, you'll have to add an entry to the stubtest allowlist if you want to use that kind of hack, or you'll get a CI error.)

anniel-stripe: Oh very good to know, thank you! Just updated.

AlexWaygood (approved):
Thanks!

PR #10465, opened 2023-07-15
Overload bool.__new__ for Literals

Gobot1234:
Inspired by https://github.com/microsoft/pyright/issues/5515


Gobot1234:
Is there a way to see the projects this affects using mypy-primer using my mypy branch?

AlexWaygood:
> Is there a way to see the projects this affects using mypy-primer using my mypy branch?

The easiest way would be to temporarily push a commit to your mypy PR that makes these edits to `mypy/typeshed/stdlib/builtins.pyi`

stdlib/builtins.pyi
```diff
@@ -214,6 +214,8 @@ _NegativeInteger: TypeAlias = Literal[-1, -2, -3, -4, -5, -6, -7, -8, -9, -10, -
 _LiteralInteger = _PositiveInteger | _NegativeInteger | Literal[0]  # noqa: Y026  # TODO: Use TypeAlias once mypy bugs are fixed
 
 class int:
+    @overload
+    def __new__(cls) -> Literal[0]: ...
```
AlexWaygood, line 218: Not sure if this is actually safe?

```pycon
>>> class Foo(int): ...
...
>>> type(Foo())
<class '__main__.Foo'>
```

In general our approach is to give more precise signatures for methods on classes, even if they're not so precise for subclasses. But these overloads would lead to typecheckers making inferences that are flat-out wrong when it comes to the return type of these methods on subclasses.

AlexWaygood: The changes you're making to `bool` are probably okay, since `bool` can't be subclassed

Gobot1234: Oh that's fair, maybe this can be re-looked at if we get an Intersection then

AlexWaygood: > maybe this can be re-looked at if we get an Intersection then

Thinking about a hypothetical type like `Self & Literal[0]` makes my brain go fuzzy, but sure 😆

stdlib/builtins.pyi
```diff
@@ -898,8 +898,21 @@ class memoryview(Sequence[int]):
...
 @final
 class bool(int):
+    @overload
+    def __new__(cls) -> Literal[False]: ...
+    @overload
+    def __new__(cls, __o: _Truthy) -> Literal[True]: ...
+    @overload
+    def __new__(cls, __o: _Falsy) -> Literal[False]: ...
+    @overload
     def __new__(cls, __o: object = ...) -> Self: ...
```
AlexWaygood, line 916: Slightly simpler:

```suggestion
    @overload
    def __new__(cls, __o: _Truthy) -> Literal[True]: ...
    @overload
    def __new__(cls, __o: _Falsy = ...) -> Literal[False]: ...
    @overload
    def __new__(cls, __o: object) -> Self: ...
```

Gobot1234: pyright's not a fan Screenshot 2023-07-19 at 11 47 32

AlexWaygood: I think pyright is incorrect (and we have that lint switched off in our CI due to the number of false positives it caused). But if you like, you can add some test cases to prove me wrong :)

Gobot1234:
https://github.com/python/mypy/pull/15687#issuecomment-1641863435 changes don't appear to be causing any problems

srittau:
Still waiting for python/mypy#15687.

Gobot1234:
Is there a way to run this with mypy dev? @srittau

srittau:
@Gobot1234 You could just temporarily change the mypy version in `requirements-tests.txt` to point to a git reference in this PR I think we did this before, but couldn't find it in `git log`.

srittau:
I've removed the "deferred" marker.

We might be able to suppress mypy's warnings, although we should add some tests to ensure that mypy still works correctly despite that.

Gobot1234:
> I've removed the "deferred" marker.
> 
> 
> 
> We might be able to suppress mypy's warnings, although we should add some tests to ensure that mypy still works correctly despite that.
> 
> 

Where do you want these?

AlexWaygood:
> > I've removed the "deferred" marker.
> > We might be able to suppress mypy's warnings, although we should add some tests to ensure that mypy still works correctly despite that.
> 
> Where do you want these?

You could add a `check_bool.py` file to this directory: https://github.com/python/typeshed/tree/main/stdlib/%40tests/test_cases/builtins. Docs on how those tests work can be found at https://github.com/python/typeshed/blob/main/tests/REGRESSION.md.

PR #10468, opened 2023-07-17
Treeview identifiers can be int

brccabral:
Treeview identifiers can be int


stdlib/tkinter/ttk.pyi
```diff
@@ -961,9 +961,9 @@ class Treeview(Widget, tkinter.XView, tkinter.YView):
         master: tkinter.Misc | None = None,
         *,
         class_: str = ...,
-        columns: str | list[str] | list[int] | list[str | int] | tuple[str | int, ...] = ...,
+        columns: _TreeviewColumnId | list[str] | list[int] | list[str | int] | tuple[str | int, ...] = ...,
```
Akuli, line 964: I think `int` does not make sense for `columns`. Why would you do `Treeview(columns=10)`? This will define one column with the name `10`, but someone will surely think it makes 10 different columns.

```suggestion
        columns: str | list[str] | list[int] | list[str | int] | tuple[str | int, ...] = ...,
```

stdlib/tkinter/ttk.pyi
```diff
@@ -961,9 +961,9 @@ class Treeview(Widget, tkinter.XView, tkinter.YView):
...
         cursor: tkinter._Cursor = ...,
-        displaycolumns: str | list[str] | tuple[str, ...] | list[int] | tuple[int, ...] = ...,
+        displaycolumns: _TreeviewColumnId | list[str] | tuple[str, ...] | list[int] | tuple[int, ...] = ...,
```
Akuli, line 966: I think we should delete `_TreeviewColumnId` and just use `int | str` everywhere. You can do it in this PR or in another PR or not at all, however you want :)

stdlib/tkinter/ttk.pyi
```diff
@@ -983,9 +983,9 @@ class Treeview(Widget, tkinter.XView, tkinter.YView):
         self,
         cnf: dict[str, Any] | None = None,
         *,
-        columns: str | list[str] | list[int] | list[str | int] | tuple[str | int, ...] = ...,
+        columns: _TreeviewColumnId | list[str] | list[int] | list[str | int] | tuple[str | int, ...] = ...,
```
Akuli, line 986:
```suggestion
        columns: str | list[str] | list[int] | list[str | int] | tuple[str | int, ...] = ...,
```

stdlib/tkinter/ttk.pyi
```diff
@@ -1023,13 +1023,19 @@ class Treeview(Widget, tkinter.XView, tkinter.YView):
...
-    def delete(self, *items: str) -> None: ...
-    def detach(self, *items: str) -> None: ...
-    def exists(self, item: str) -> bool: ...
+    @overload
+    def delete(self, items: str | list[str] | tuple[str, ...] | int | list[int] | tuple[int, ...]) -> None: ...
+    @overload
+    def delete(self, *items: str | int) -> None: ...
+    @overload
+    def detach(self, items: str | list[str] | tuple[str, ...] | int | list[int] | tuple[int, ...]) -> None: ...
+    @overload
```
Akuli, line 1032:
```suggestion
```

stdlib/tkinter/ttk.pyi
```diff
@@ -1058,7 +1064,7 @@ class Treeview(Widget, tkinter.XView, tkinter.YView):
     def identify_column(self, x: int) -> str: ...
     def identify_region(self, x: int, y: int) -> Literal["heading", "separator", "tree", "cell", "nothing"]: ...
     def identify_element(self, x: int, y: int) -> str: ...  # don't know what possible return values are
-    def index(self, item: str) -> int: ...
+    def index(self, item: str | int) -> int: ...
     def insert(
         self,
         parent: str,
```
Akuli, line 1070: Just below this line (can't leave a comment there unfortunately) there is `id: str = ...` and `iid: str | None = None`. Those should also accept ints.

stdlib/tkinter/ttk.pyi
```diff
@@ -1098,27 +1104,46 @@ class Treeview(Widget, tkinter.XView, tkinter.YView):
...
     if sys.version_info >= (3, 8):
         def selection(self) -> tuple[str, ...]: ...
     else:
         def selection(self, selop: Incomplete | None = ..., items: Incomplete | None = None) -> tuple[str, ...]: ...
 
-    def selection_set(self, items: str | list[str] | tuple[str, ...]) -> None: ...
-    def selection_add(self, items: str | list[str] | tuple[str, ...]) -> None: ...
-    def selection_remove(self, items: str | list[str] | tuple[str, ...]) -> None: ...
-    def selection_toggle(self, items: str | list[str] | tuple[str, ...]) -> None: ...
+    if sys.version_info >= (3, 6):
```
Akuli, line 1118: We don't support old Python versions, so you don't need this check. The 3.6+ code by itself is enough.

Akuli (commented):
Thanks!

stdlib/tkinter/ttk.pyi
```diff
@@ -998,23 +996,23 @@ class Treeview(Widget, tkinter.XView, tkinter.YView):
     @overload
     def configure(self, cnf: str) -> tuple[str, str, str, Any, Any]: ...
     config = configure
-    def bbox(self, item, column: _TreeviewColumnId | None = None) -> tuple[int, int, int, int] | Literal[""]: ...  # type: ignore[override]
-    def get_children(self, item: str | None = None) -> tuple[str, ...]: ...
-    def set_children(self, item: str, *newchildren: str) -> None: ...
+    def bbox(self, item: str | int, column: str | int | None = None) -> tuple[int, int, int, int] | Literal[""]: ...  # type: ignore[override]
+    def get_children(self, item: str | int | None = None) -> tuple[str, ...]: ...
+    def set_children(self, item: str | int, *newchildren: str) -> None: ...
```
Akuli, line 1001:
```suggestion
    def set_children(self, item: str | int, *newchildren: str | int) -> None: ...
```

stdlib/tkinter/ttk.pyi
```diff
@@ -1023,29 +1021,35 @@ class Treeview(Widget, tkinter.XView, tkinter.YView):
         anchor: tkinter._Anchor = ...,
         # id is read-only
     ) -> _TreeviewColumnDict | None: ...
-    def delete(self, *items: str) -> None: ...
-    def detach(self, *items: str) -> None: ...
-    def exists(self, item: str) -> bool: ...
+    @overload
+    def delete(self, items: str | int) -> None: ...
```
Akuli, line 1025: `treeview.delete(items=foo)` will not work, so the `items` argument should be marked as positional-only. Then its name does not matter, so we might as well make it obvious that you are specifying just one item in this overload.

```suggestion
    def delete(self, __item: str | int) -> None: ...
```

stdlib/tkinter/ttk.pyi
```diff
@@ -1023,29 +1021,35 @@ class Treeview(Widget, tkinter.XView, tkinter.YView):
...
+    @overload
+    def delete(self, *items: str | int) -> None: ...
+    @overload
+    def detach(self, items: str | int) -> None: ...
```
Akuli, line 1029:
```suggestion
    def detach(self, __item: str | int) -> None: ...
```

stdlib/tkinter/ttk.pyi
```diff
@@ -1098,27 +1102,39 @@ class Treeview(Widget, tkinter.XView, tkinter.YView):
...
     else:
         def selection(self, selop: Incomplete | None = ..., items: Incomplete | None = None) -> tuple[str, ...]: ...
 
-    def selection_set(self, items: str | list[str] | tuple[str, ...]) -> None: ...
-    def selection_add(self, items: str | list[str] | tuple[str, ...]) -> None: ...
-    def selection_remove(self, items: str | list[str] | tuple[str, ...]) -> None: ...
-    def selection_toggle(self, items: str | list[str] | tuple[str, ...]) -> None: ...
     @overload
-    def set(self, item: str, column: None = None, value: None = None) -> dict[str, Any]: ...
+    def selection_set(self, items: str | int | list[str] | tuple[str, ...] | list[int] | tuple[int, ...]) -> None: ...
```
Akuli, line 1117:
```suggestion
    def selection_set(self, __items: str | int | list[str] | tuple[str, ...] | list[int] | tuple[int, ...]) -> None: ...
```

stdlib/tkinter/ttk.pyi
```diff
@@ -1098,27 +1102,39 @@ class Treeview(Widget, tkinter.XView, tkinter.YView):
...
+    @overload
+    def selection_set(self, *items: str | int) -> None: ...
+    @overload
+    def selection_add(self, items: str | int | list[str] | tuple[str, ...] | list[int] | tuple[int, ...]) -> None: ...
```
Akuli, line 1121:
```suggestion
    def selection_add(self, __items: str | int | list[str] | tuple[str, ...] | list[int] | tuple[int, ...]) -> None: ...
```

stdlib/tkinter/ttk.pyi
```diff
@@ -1098,27 +1102,39 @@ class Treeview(Widget, tkinter.XView, tkinter.YView):
...
+    @overload
+    def selection_add(self, *items: str | int) -> None: ...
+    @overload
+    def selection_remove(self, items: str | int | list[str] | tuple[str, ...] | list[int] | tuple[int, ...]) -> None: ...
```
Akuli, line 1125:
```suggestion
    def selection_remove(self, __items: str | int | list[str] | tuple[str, ...] | list[int] | tuple[int, ...]) -> None: ...
```

stdlib/tkinter/ttk.pyi
```diff
@@ -1098,27 +1102,39 @@ class Treeview(Widget, tkinter.XView, tkinter.YView):
...
+    @overload
+    def selection_remove(self, *items: str | int) -> None: ...
+    @overload
+    def selection_toggle(self, items: str | int | list[str] | tuple[str, ...] | list[int] | tuple[int, ...]) -> None: ...
```
Akuli, line 1129:
```suggestion
    def selection_toggle(self, __items: str | int | list[str] | tuple[str, ...] | list[int] | tuple[int, ...]) -> None: ...
```

Akuli (commented):
The last nits...

stdlib/tkinter/ttk.pyi
```diff
@@ -1023,29 +1021,35 @@ class Treeview(Widget, tkinter.XView, tkinter.YView):
...
-    def detach(self, *items: str) -> None: ...
-    def exists(self, item: str) -> bool: ...
+    @overload
+    def delete(self, __item: str | int) -> None: ...
+    @overload
+    def delete(self, *items: str | int) -> None: ...
+    @overload
+    def detach(self, __item: str | int) -> None: ...
+    @overload
+    def detach(self, *items: str | int) -> None: ...
```
AlexWaygood, line 1031: @brccabral and @Akuli, do we need the overloads here? Are there any valid calls that would be disallowed if we just did:

```suggestion
    def delete(self, *items: str | int) -> None: ...
    def detach(self, *items: str | int) -> None: ...
```

brccabral: probably not, I had created the overload to accept lists, but they are not valid... I'll edit a new commit

stdlib/tkinter/ttk.pyi
```diff
@@ -1098,27 +1096,39 @@ class Treeview(Widget, tkinter.XView, tkinter.YView):
...
     else:
         def selection(self, selop: Incomplete | None = ..., items: Incomplete | None = None) -> tuple[str, ...]: ...
 
-    def selection_set(self, items: str | list[str] | tuple[str, ...]) -> None: ...
-    def selection_add(self, items: str | list[str] | tuple[str, ...]) -> None: ...
-    def selection_remove(self, items: str | list[str] | tuple[str, ...]) -> None: ...
-    def selection_toggle(self, items: str | list[str] | tuple[str, ...]) -> None: ...
     @overload
-    def set(self, item: str, column: None = None, value: None = None) -> dict[str, Any]: ...
+    def selection_set(self, __items: str | int | list[str] | tuple[str, ...] | list[int] | tuple[int, ...]) -> None: ...
```
Akuli, line 1111: Might be slightly simpler if the first overloads only accepted lists and tuples. Like this, for example:

```suggestion
    def selection_set(self, __items: list[str] | tuple[str, ...] | list[int] | tuple[int, ...]) -> None: ...
```

If you pass one `int` or `str`, it will match the overload with `*items` anyway.

Akuli (commented):
Last nit:

Akuli (approved):
Thanks for your work on our tkinter stubs :)

PR #10472, opened 2023-07-18
Update `mock` to `5.1.*`

sobolevn:
Closes https://github.com/python/typeshed/pull/10443


sobolevn:
This looks like a bug in `stubtest`:

```
error: mock.__all__ names exported from the stub do not correspond to the names exported at runtime. This is probably due to things being missing from the stub or an inaccurate `__all__` in the stub
Stub: in file /Users/sobolev/Desktop/typeshed/stubs/mock/mock/__init__.pyi
Names exported in the stub but not at runtime: []
Runtime:
Names exported at runtime but not in the stub: ['__version__']
```

I have `'__version__'` in `__all__` 🤔

AlexWaygood:
Looks like you have it included in `__all__` but not actually included in the module: https://github.com/python/typeshed/actions/runs/5586229512/jobs/10210010596?pr=10472

AlexWaygood:
I don't think there's a stubtest bug here. The error message says:

> error: `mock.__all__` names exported from the stub do not correspond to the names exported at runtime. This is probably due to things being missing from the stub or an inaccurate `__all__` in the stub

`__all__` is accurate in the stub here, but there _are_ names included in `__all__` at runtime that aren't currently exported from the stub (because they don't exist in the stub at the moment!).

The error message _used_ to be:

> error: `mock` names exported from the stub do not correspond to the names exported at runtime. This is probably due to things being missing from the stub or an inaccurate `__all__` in the stub

But _you_ were the one who changed it in https://github.com/python/mypy/commit/ceb976f1b9bc4c3ae7cd07a5222ec076f6c7ab72 so that the `object_path` always includes `__all__` ;)

stubs/mock/mock/mock.pyi
```diff
@@ -378,3 +379,13 @@ class PropertyMock(Mock):
     def __set__(self, obj: Any, value: Any) -> None: ...
 
 def seal(mock: Any) -> None: ...
+
+class ThreadingMixin(Base):
+    DEFAULT_TIMEOUT: float | None
```
AlexWaygood, line 384: The docstring says:

> You can change the default timeout of all instances via the `ThreadingMock.DEFAULT_TIMEOUT` attribute.

https://github.com/testing-cabal/mock/blob/d344fa2794b3b1ae7e4a4dbf265fb040d6f41d1f/mock/mock.py#L3137C1-L3138C47

```suggestion
    DEFAULT_TIMEOUT: ClassVar[float | None]
```

stubs/mock/mock/mock.pyi
```diff
@@ -378,3 +379,13 @@ class PropertyMock(Mock):
...
+
+    def __init__(self, *args: Any, timeout: float | None = ..., **kwargs: Any) -> None: ...
+    def reset_mock(self, *args: Any, **kwargs: Any) -> None: ...
```
AlexWaygood, line 387: This calls `super().reset_mock()`: https://github.com/testing-cabal/mock/blob/d344fa2794b3b1ae7e4a4dbf265fb040d6f41d1f/mock/mock.py#L3081

When this class is used as a superclass for `ThreadingMock`, that `super()` call ends up calling `NonCallableMock.reset_mock()`. So perhaps the signature of this method should match that method?

```suggestion
    def reset_mock(self, visited: Any = None, *, return_value: bool = False, side_effect: bool = False) -> None: ...
```

stubs/mock/mock/mock.pyi
sobolevn: Yeah, this is a preferable way: how it should be used.
Right now it is annotated the way it can be used: https://github.com/testing-cabal/mock/blob/d344fa2794b3b1ae7e4a4dbf265fb040d6f41d1f/mock/mock.py#L3071

stubs/mock/mock/mock.pyi
sobolevn: In theory, it can be used with something else, since it is a mixin.
For example, `AsyncMockMixin` in the same file uses just `*args, **kwargs`

AlexWaygood: I don't think this class is intended to be used as a mixin in conjunction with any classes outside `unittest.mock` or the third-party `mock` package, so I think this is just going to lead to unnecessarily imprecise typechecking. But you are strictly speaking correct.

As a compromise, what if we overrode `reset_mock()` on the `ThreadingMock` class (even though that's not what happens at runtime), with the more precise signature that you see on `NonCallableMock.reset_mock()`?

sobolevn: > if we overrode reset_mock() on the ThreadingMock class (even though that's not what happens at runtime), with the more precise signature that you see on NonCallableMock.reset_mock()?

Yes, that's a good idea. I can do the same for `AsyncMockMixin` in the next PR.

AlexWaygood (approved):
Thanks!

Issue #10113, opened 2023-04-29
Drop support for Python 3.7 (not until January 2024)


AlexWaygood:
Python 3.7 will reach end-of-life on 2023-06-27. For 3.6, we waited 6 months after EOL before dropping support. Do we want to do something similar this time, or are we happy with a shorter grace period?

3.7 is still a [very popular Python version](https://hugovk.github.io/pypi-tools/charts). On the other hand, large parts of the Python ecosystem have already dropped support for Python 3.7 in their latest versions (flake8, isort, numpy, pandas, Django).

hauntsaninja:
6 months seems fine to me.

JelleZijlstra:
Agree, I don't think we need to hurry to drop old versions. The maintenance load from old versions in typeshed isn't that great, and lots of people still use 3.7.

AlexWaygood:
I don't have a strong opinion either way, I just wanted to have the discussion. 6 months is fine for me!

Avasam:
6 months seems reasonable, it's also in line with typeshed's already established time frame of supporting obsolete stubs. And just like obsolete stubs, older versions will still be available on PyPI.
I'm really looking forward to using PEP 570 and waiting for it to properly type pywin32's C-modules.

AlexWaygood:
> I'm really looking forward to using PEP 570 and waiting for it to properly type pywin32's C-modules.

Hmm, can't you use the `__` prefix for arguments? I expect we'll do some kind of codemod once we drop support for Python 3.7, that automatically converts the old-style syntax for positional-only arguments to PEP-570 syntax.

Avasam:
> Hmm, can't you use the `__` prefix for arguments? I expect we'll do some kind of codemod once we drop support for Python 3.7, that automatically converts the old-style syntax for positional-only arguments to PEP-570 syntax.

Short answer is: I could, but I'd rather wait.

I didn't think of a codemod, that removes the concern of having to do additional work to migrate.
But I still think it'll be easier to review if only a `/` is added in hundreds of methods, rather than all param names changing. It'll be easier and faster to do too.

AlexWaygood:
[...]

PR #10473, opened 2023-07-18
Add stubs for `pluggy==1.2.0`

hoefling:
This PR suggests to add stubs for the `pluggy` package of the current version 1.2.0. The reason for this is that although `pluggy` itself is (almost) completely typed, the package ships no PEP 561 marker yet, as there is no consensus about what types to expose, see https://github.com/pytest-dev/pluggy/pull/414.


stubs/pluggy/pluggy/_hooks.pyi
```diff
@@ -0,0 +1,127 @@
...
+
+if sys.version_info >= (3, 8):
+    from typing import Final, TypedDict
+else:
+    from typing_extensions import Final, TypedDict
+
+if sys.version_info >= (3, 10):
+    from typing import TypeAlias
+else:
+    from typing_extensions import TypeAlias
```
AlexWaygood, line 16: We generally just import these from `typing_extensions` unconditionally. Type checkers think typing_extensions is part of the stdlib, so it doesn't have any impact on the dependencies of the stubs package if it's "unnecessarily" importing something from typing_extensions on Python 3.10, even though it "could" be importing the thing from typing:

```suggestion
from typing_extensions import Final, TypedDict, TypeAlias
```

stubs/pluggy/pluggy/_result.pyi
```diff
@@ -0,0 +1,28 @@
+import sys
+from collections.abc import Callable
+from types import TracebackType
+from typing import Generic, TypeVar
+
+if sys.version_info >= (3, 10):
+    from typing import TypeAlias
+else:
+    from typing_extensions import TypeAlias
```
AlexWaygood, line 9:
```suggestion
from typing_extensions import TypeAlias
```

stubs/pluggy/pluggy/_tracing.pyi
```diff
@@ -0,0 +1,25 @@
+import sys
+from collections.abc import Callable
+from typing import Any
+
+if sys.version_info >= (3, 10):
+    from typing import TypeAlias
+else:
+    from typing_extensions import TypeAlias
```
AlexWaygood, line 8:
```suggestion
from typing_extensions import TypeAlias
```

stubs/pluggy/pluggy/_manager.pyi
```diff
@@ -0,0 +1,72 @@
...
+else:
+    from typing_extensions import TypeAlias
+
+if sys.version_info >= (3, 8):
+    from importlib.metadata import Distribution
+    from typing import Final
+else:
+    from typing_extensions import Final
+
+    Distribution: TypeAlias = Any
```
AlexWaygood, line 31:
```suggestion
from typing_extensions import Final, TypeAlias

if sys.version_info >= (3, 8):
    from importlib.metadata import Distribution
else:
    Distribution: TypeAlias = Any
```

stubs/pluggy/pluggy/_hooks.pyi
hoefling: @AlexWaygood thank you! I remember now either you or another maintainer already pointed me at this once, but of course it didn't last long before I forgot about it :woman_facepalming:

stubs/pluggy/pluggy/__init__.pyi
```diff
@@ -0,0 +1,3 @@
+from ._hooks import HookimplMarker as HookimplMarker, HookspecMarker as HookspecMarker
+from ._manager import PluginManager as PluginManager, PluginValidationError as PluginValidationError
+from ._result import HookCallError as HookCallError
```
AlexWaygood, line 3: Should get rid of a few of the primer hits:

```suggestion
from ._result import HookCallError as HookCallError

__version__: str
```

stubs/pluggy/pluggy/_hooks.pyi
```diff
@@ -0,0 +1,117 @@
...
+from types import ModuleType
+from typing import Any, TypeVar, overload
+from typing_extensions import Final, TypeAlias, TypedDict
+
+from ._result import _Result
+
+_T = TypeVar("_T")
+_F = TypeVar("_F", bound=Callable[..., object])
+_Namespace: TypeAlias = ModuleType | type
+_Plugin = object
```
AlexWaygood, line 11: This is only ever used as a type alias:

```suggestion
_Plugin: TypeAlias = object
```

stubs/pluggy/pluggy/_hooks.pyi
```diff
@@ -0,0 +1,117 @@
...
+class _HookImplOpts(TypedDict):
+    wrapper: bool
+    hookwrapper: bool
+    optionalhook: bool
+    tryfirst: bool
+    trylast: bool
+    specname: str | None
+
+class HookspecMarker:
+    project_name: Final[str]
```
AlexWaygood, line 29: `Final` actually has slightly different semantics in a stub file than it does at runtime. While it's okay at runtime to have a "final instance variable", which is what `pluggy` does with this variable:

```py
class HookspecMarker:
    def __init__(self, project_name: str) -> None:
        self.project_name: Final = project_name
```

...in a stub, all variable annotations are put in the class scope, and PEP-591 states:

> Type checkers should infer a final attribute that is initialized in a class body as being a class variable. Variables should not be annotated with both ClassVar and Final.

As such, it doesn't seem to be possible to express the idea of a "final instance variable" in a stub file. Maybe we should just do:

```suggestion
    project_name: str
```

Or maybe we should add a comment:

```suggestion
    # actually a "Final instance variable" at runtime,
    # but that's not expressible in a stub,
    # where all Final variables are implicit ClassVars
    project_name: Final[str]
```

stubs/pluggy/pluggy/_hooks.pyi
```diff
@@ -0,0 +1,117 @@
...
+    def __init__(self, project_name: str) -> None: ...
+    @overload
+    def __call__(self, function: _F, firstresult: bool = ..., historic: bool = ..., warn_on_impl: Warning | None = ...) -> _F: ...
+    @overload
+    def __call__(
+        self, function: None = ..., firstresult: bool = ..., historic: bool = ..., warn_on_impl: Warning | None = ...
```
AlexWaygood, line 35: Our policy has changed recently to prefer including default values wherever possible (it's not necessary for typecheckers, but it makes for a better experience for users of IDEs). [`stubdefaulter`](https://github.com/JelleZijlstra/stubdefaulter) can add most them automatically for you, so I won't bother commenting on each signature that's missing default values :-)

stubs/pluggy/pluggy/_hooks.pyi
```diff
@@ -0,0 +1,117 @@
...
+    ) -> Callable[[_F], _F]: ...
+
+class HookimplMarker:
+    project_name: Final[str]
```
AlexWaygood, line 39: Same comment above about the "Final instance variable" issue

stubs/pluggy/pluggy/_hooks.pyi
```diff
@@ -0,0 +1,117 @@
...
+    ) -> Callable[[_F], _F]: ...
+
+def normalize_hookimpl_opts(opts: _HookImplOpts) -> None: ...
+def varnames(func: object) -> tuple[tuple[str, ...], tuple[str, ...]]: ...
+
+class _HookRelay:
+    def __getattr__(self, name: str) -> _HookCaller: ...
+
+class _HookCaller:
+    name: Final[str]
```
AlexWaygood, line 71: And again

stubs/pluggy/pluggy/_hooks.pyi
```diff
@@ -0,0 +1,117 @@
...
+    def is_historic(self) -> bool: ...
+    def get_hookimpls(self) -> list[HookImpl]: ...
+    def __call__(self, **kwargs: object) -> Any: ...
+    def call_historic(
+        self, result_callback: Callable[[Any], None] | None = ..., kwargs: Mapping[str, object] | None = ...
+    ) -> None: ...
+    def call_extra(self, methods: Sequence[Callable[..., object]], kwargs: Mapping[str, object]) -> Any: ...
+
+class _SubsetHookCaller(_HookCaller):
+    def __init__(self, orig: _HookCaller, remove_plugins: AbstractSet[_Plugin]) -> None: ...
```
AlexWaygood, line 91: Is it necessary to include this class in the stubs? We generally prefer to stub only the public API in typeshed, and this class doesn't seem to be referenced anywhere else in this PR. Getting rid of it would also half the length of the stubtest allowlist :-)

stubs/pluggy/pluggy/_manager.pyi
```diff
@@ -0,0 +1,65 @@
...
+    _Plugin,
+    normalize_hookimpl_opts as normalize_hookimpl_opts,
+)
+from ._result import _Result as _Result
+from ._tracing import TagTracerSub
+
+if sys.version_info >= (3, 8):
+    from importlib.metadata import Distribution
+else:
+    Distribution: TypeAlias = Any
```
AlexWaygood, line 24: We actually allow our stubs packages to have non-stubs packages as dependencies these days. So you could potentially add `importlib_metadata` as a dependency to this stubs package on Python 3.7. But also, Python 3.7 is now end-of-life and we'll slowly be phasing support out from typeshed over the coming months (see #10113). So maybe it's not worth the bother. I'm easy either way; up to you.

stubs/pluggy/pluggy/_result.pyi
```diff
@@ -0,0 +1,23 @@
+from collections.abc import Callable
+from types import TracebackType
+from typing import Generic, TypeVar
+from typing_extensions import TypeAlias
+
+_ExcInfo: TypeAlias = tuple[type[BaseException], BaseException, TracebackType]
```
AlexWaygood, line 6: You can replace this with `from _typeshed import ExcInfo`:

https://github.com/python/typeshed/blob/032f9195f9e3788a6e5c7ecb086f173a3ac92a95/stdlib/_typeshed/__init__.pyi#L254

stubs/pluggy/@tests/stubtest_allowlist.txt
```diff
@@ -0,0 +1,2 @@
+pluggy._hooks._HookRelay.__getattr__
```
AlexWaygood, line 1: Please add a comment here saying why the allowlist entry exists!

AlexWaygood (commented):
Thanks! Looks pretty good; a few remarks below.

Seems a bit unfortunate to add these stubs given how heavily typed the runtime package is, but if they don't want to add a `py.typed` file, I guess I'm okay with adding these.

stubs/pluggy/pluggy/_tracing.pyi
```diff
@@ -0,0 +1,20 @@
+from collections.abc import Callable
+from typing import Any
+from typing_extensions import TypeAlias
+
+_Writer: TypeAlias = Callable[[str], None]
```
AlexWaygood, line 5: The upstream package has a bad type annotation here. Since the value returned from `TagTracer._writer` is always discarded, this should be

```suggestion
_Writer: TypeAlias = Callable[[str], object]
```

That will get rid of a couple of the mypy_primer errors

hoefling:
@AlexWaygood 
> Seems a bit unfortunate to add these stubs given how heavily typed the runtime package is, but if they don't want to add a `py.typed` file, I guess I'm okay with adding these.

TBH I was expecting the package to expose type hints in 1.2.0, but since it is postponed, I'd rather have a `types-pluggy` for a year or so out there to not to carry around my own stubs along in every `pytest` plugin repo. Once the type hints are exposed, I think I'll be the first one to drop the stubs again and carry one less `types-*` dependency around.

Another (not so obvious and rather a personal) benefit from this PR is that thanks to you, I'm getting a great sync on the latest `typeshed` development progress ;-)

stubs/pluggy/pluggy/_manager.pyi
hoefling: @AlexWaygood I think something is still slipping away from me here - if I apply this patch
```diff
--- a/stubs/pluggy/METADATA.toml
+++ b/stubs/pluggy/METADATA.toml
@@ -1 +1,2 @@
 version = "1.2.0"
+requires = ["importlib-metadata;python_version<'3.8'"]
--- a/stubs/pluggy/pluggy/_manager.pyi
+++ b/stubs/pluggy/pluggy/_manager.pyi
@@ -21,7 +21,7 @@ from ._tracing import TagTracerSub
 if sys.version_info >= (3, 8):
     from importlib.metadata import Distribution
 else:
-    Distribution: TypeAlias = Any
+    from importlib_metadata import Distribution

 _BeforeTrace: TypeAlias = Callable[[str, Sequence[HookImpl], Mapping[str, Any]], None]
 _AfterTrace: TypeAlias = Callable[[_Result[Any], str, Sequence[HookImpl], Mapping[str, Any]], None]
 ```
then `mypy` doesn't like it, since the requirement is not available:
```sh
$ python tests/mypy_test.py --platform linux --python-version=3.7 stubs/pluggy
*** Testing Python 3.7 on linux
testing pluggy (5 files)... failure

stubs/pluggy/pluggy/_manager.pyi:24: error: Cannot find implementation or library stub for module named "importlib_metadata"  [import]
stubs/pluggy/pluggy/_manager.pyi:24: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports


--- exit status 1, 5 files checked ---
```

AlexWaygood: Hrm, I'll try to take a look today

AlexWaygood: > then `mypy` doesn't like it, since the requirement is not available:
> 
> ```shell
> $ python tests/mypy_test.py --platform linux --python-version=3.7 stubs/pluggy
> *** Testing Python 3.7 on linux
> Testing third-party packages...
> testing pluggy (5 files)... failure
> 
> stubs/pluggy/pluggy/_manager.pyi:24: error: Cannot find implementation or library stub for module named "importlib_metadata"  [import]
> stubs/pluggy/pluggy/_manager.pyi:24: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
> 
> 
> --- exit status 1, 5 files checked ---
> ```

Oh. You probably need to actually run the test _using_ Python 3.7 for the requirement to be installed by mypy_test.py. The --python-version 3.7 flag probably isn't enough.

(mypy_test.py these days dynamically creates a virtual environment on the fly for each package that has non-types dependencies, and uses pip to install all non-types dependencies into the venv. But pip is probably just ignoring the importlib-metadata dependency because you're python version is >=3.8? At a guess?)

hoefling: @AlexWaygood ahh, indeed! Sorry for the noise 🥲

AlexWaygood: Okay, looks like adding `importlib_metadata` as a dependency on 3.7 works with `mypy_test.py`, but _does_ break just about every other test we have. It's all probably _fixable_, but some of it would be a bit of a pain to fix, so maybe it would be better to just have `Distribution: TypeAlias = Any`, the same way you had it originally 🙃

Sorry for the churn!!

hoefling: @AlexWaygood yeah I mean since py3.7 is about to be dropped anyway, it doesn't matter this much anymore 😎

hoefling: @AlexWaygood looks like [this job](https://github.com/python/typeshed/actions/runs/5610198580/jobs/10264878450?pr=10473) failed now because of GH hiccups, not because of code changes.

AlexWaygood: > @AlexWaygood looks like [this job](https://github.com/python/typeshed/actions/runs/5610198580/jobs/10264878450?pr=10473) failed now because of GH hiccups, not because of code changes.

Yeah, it's been really flaky recently. Not sure why; I should probably open a bug report...

stubs/pluggy/pluggy/_tracing.pyi
```diff
@@ -0,0 +1,20 @@
...
+from typing_extensions import TypeAlias
+
+_Writer: TypeAlias = Callable[[str], object]
+_Processor: TypeAlias = Callable[[tuple[str, ...], tuple[Any, ...]], None]
+
+class TagTracer:
+    indent: int
+    def __init__(self) -> None: ...
+    def get(self, name: str) -> TagTracerSub: ...
+    def setwriter(self, writer: _Writer) -> None: ...
```
AlexWaygood, line 12: pytest does `setwriter(None)` here: https://github.com/pytest-dev/pytest/blob/1de0923e83e3258709559d86d142adaab65a42d8/src/_pytest/helpconfig.py#L126

And that makes sense to me, since `setwriter` just sets the private `_writer` variable: https://github.com/pytest-dev/pluggy/blob/e507275d33b03b7e7161b2d8c9b184939973bc91/src/pluggy/_tracing.py#L53

And the `_writer` variable is typed as `_Writer | None` in the pluggy source code: https://github.com/pytest-dev/pluggy/blob/e507275d33b03b7e7161b2d8c9b184939973bc91/src/pluggy/_tracing.py#L19

So I think the upstream annotation for this method is wrong, and it should be:

```suggestion
    def setwriter(self, writer: _Writer | None) -> None: ...
```

(This gets rid of another hit that mypy_primer surfaced)

stubs/pluggy/pluggy/_result.pyi
```diff
@@ -0,0 +1,21 @@
...
+from collections.abc import Callable
+from typing import Generic, TypeVar
+
+_T = TypeVar("_T")
+
+class HookCallError(Exception): ...
+
+class _Result(Generic[_T]):
+    _result: _T
+    _exception: BaseException
```
AlexWaygood, line 11: Hmm, do we need these private attributes? (Same point that I mentioned earlier about typeshed usually only providing stubs for public API)

```suggestion
```

stubs/pluggy/pluggy/_manager.pyi
```diff
@@ -0,0 +1,65 @@
...
+    def __init__(self, dist: Distribution) -> None: ...
+    @property
+    def project_name(self) -> str: ...
+    def __getattr__(self, attr: str, default: Incomplete | None = None) -> Any: ...
+    def __dir__(self) -> list[str]: ...
+
+class PluginManager:
+    project_name: Final[str]
+    trace: Final[TagTracerSub]
+    hook: Final[_HookRelay]
```
AlexWaygood, line 43: A few more "Final instance variables":

```suggestion
    project_name: str
    trace: TagTracerSub
    hook: _HookRelay
```

AlexWaygood (commented):
Nearly there, I think!

stubs/pluggy/pluggy/_hooks.pyi
AlexWaygood: > @AlexWaygood thank you for pointing me at PEP 591! Do you think it would be a good idea to add another rule to `flake8-pyi` to guard for it?

Not sure. I'd want to think about it a bit more. In general I feel like the concept of "Final instance variables in stub files" is a bit underspecced, so I'm not sure whether we'd want to enforce a rule or what that rule should enforce 😄

stubs/pluggy/pluggy/_tracing.pyi
hoefling: @AlexWaygood good catch, I took the liberty to propose this change into upstream as well (https://github.com/pytest-dev/pluggy/pull/423)

AlexWaygood: > @AlexWaygood good catch, I took the liberty to propose this change into upstream as well ([pytest-dev/pluggy#423](https://github.com/pytest-dev/pluggy/pull/423))

Nice! https://github.com/python/typeshed/pull/10473#discussion_r1268585883 could also probably be contributed upstream, if you feel like it!

AlexWaygood (approved):
LGTM. Thanks!

PR #10475, opened 2023-07-19
Update `python-crontab` to `3.0.*`

sobolevn:
Closes https://github.com/python/typeshed/pull/10460
Source: https://gitlab.com/doctormo/python-crontab/-/blob/master/crontab.py


stubs/python-crontab/crontab.pyi
```diff
@@ -32,7 +32,18 @@ CRON_COMMAND: str
...
-def open_pipe(cmd: str, *args: str, **flags: str) -> subprocess.Popen[Any]: ...
+class Process:
+    env: subprocess._ENV | None
+    args: tuple[str, ...]
+    has_run: bool
+    stdout: str | None
+    stderr: str | None
+    returncode: int | None
+    # `posix` and `env` are known special kwargs:
+    def __init__(self, *args: str, posix: bool = ..., env: subprocess._ENV | None = ..., **flags: object | None) -> None: ...
```
srittau, line 43: `cmd` is missing, `env` has a default, and `object | None` is redundant:

```suggestion
    def __init__(self, cmd: str, *args: str, posix: bool = ..., env: subprocess._ENV | None = None, **flags: object) -> None: ...
```

stubs/python-crontab/crontab.pyi
```diff
@@ -69,8 +80,7 @@ class CronTab:
     # Usually `kwargs` are just `now: datetime | None`, but technically this can
     # work for `CronItem` subclasses, which might define other kwargs.
     def run_pending(self, **kwargs: Any) -> Iterator[str]: ...
-    # There are two known kwargs and others are unused:
-    def run_scheduler(self, timeout: int = ..., *, warp: object = ..., cadence: int = ..., **kwargs: Unused) -> Iterator[str]: ...
+    def run_scheduler(self, timeout: int = ..., cadence: int = ..., warp: object = ...) -> Iterator[str]: ...
```
srittau, line 83:
```suggestion
    def run_scheduler(self, timeout: int = -1, cadence: int = 60, warp: bool = False) -> Iterator[str]: ...
```

srittau (changes requested):
Thanks, remarks below.

srittau (approved):
It seems the stubtest allowlist entry isn't needed anymore. LGTM, apart from that!

PR #10477, opened 2023-07-19
LoggerAdapter: the extra property is a simple dict

romuald:
Even if not properly documented, the extra property of LoggerAdapter objects is a simple dict and thus can be modified

This PR simply changes the type of the extra property from Mapping (immutable) to dict (mutable)


srittau (commented):
This is a bit tricky, since the docstring says: "Initialize the adapter with a logger and a dict-like object which provides contextual information." The only direct initialization in the stdlib I could find is actually in the same class, which indeed uses a `dict`.

That said, instead of `Mapping` we should at least use `MutableMapping` and probably change the second type arg to `Any` (i.e. `MutableMapping[str, Any]`). I'd like to hear more opinions, though.

AlexWaygood:
I lean towards leaving the stubs as they are here :/

As @srittau points out, the docstring is quite careful to say that `extra` should be a "dict-like object" rather than saying that it _has_ to be a dictionary. And in general, the `logging` docs emphasise in most places that duck typing is okay for parameters passed into `logging` class constructors. It also looks like the `logging` code at runtime is pretty careful never to assume that the `extra` mapping is a mutable mapping.

So I think it would be unnecessarily restrictive to mandate that a _mutable_ mapping has to be passed to `extra` parameter, and it would be unsafe to assume that the `extra` attribute is always a mutable mapping for any given `LoggerAdapater` instance.

The unfortunate thing is that 99/100 times it probably _will_ be a `dict`. But you can easily perform type narrowing with a simple `assert` statement, if you need to mutate the dictionary:

```py
def foo(x: LoggerAdapter) -> None:
    assert isinstance(x.extra, dict)
    x.extra["foo"] = "bar"
``` 

https://gist.github.com/mypy-play/6d34edfbc031c939918144d11d889555

AlexWaygood:
Closing as per my comments above -- but thanks for the PR @romuald!

PR #10481, opened 2023-07-19
Add a more precise signature for `AsyncMock.reset_mock()`

sobolevn:
Related https://github.com/python/typeshed/pull/10472


AlexWaygood (commented):
Maybe do the stdlib one at the same time?

https://github.com/python/typeshed/blob/9e7477d9866b74ce41bc3ef41c95f5f290c287ea/stdlib/unittest/mock.pyi#L384

AlexWaygood (approved):
Thanks!

PR #10484, opened 2023-07-20
Update `argparse` to 3.12

sobolevn:


AlexWaygood (commented):
We won't be able to get rid of the allowlist entry, but we _could_ add a py312 branch to the stub, since the `=None` default values for `type`, `choices` and `metavar` aren't really accurate anymore (they should probably be `=...` on py312):

https://github.com/python/typeshed/blob/4f56889f237da15bac1a570ad703a665bef983d9/stdlib/argparse.pyi#L339-L349

AlexWaygood:
Superseded by #10582

PR #10485, opened 2023-07-20
Update `caldav` to `1.3.*`

sobolevn:
Closes https://github.com/python/typeshed/pull/10482
Source: https://github.com/python-caldav/caldav


stubs/caldav/caldav/davclient.pyi
```diff
@@ -20,7 +20,7 @@ class DAVResponse:
     status: int
     headers: CaseInsensitiveDict[str]
     objects: dict[str, dict[str, str]]  # only defined after call to find_objects_and_props()
```
AlexWaygood, line 22:
```suggestion
    objects: dict[str, dict[str, str]]  # only defined after call to find_objects_and_props()
    huge_tree: bool
```

stubs/caldav/caldav/davclient.pyi
```diff
@@ -50,6 +50,7 @@ class DAVClient:
         ssl_verify_cert: bool | str = True,
         ssl_cert: str | tuple[str, str] | None = None,
         headers: dict[str, str] = {},
+        huge_tree: bool = False,
```
AlexWaygood, line 53: You need to add `huge_tree: bool` to the class body here as well

stubs/caldav/caldav/objects.pyi
AlexWaygood, line 1: A new method was added in this module that you haven't included: `get_relatives()` https://github.com/python-caldav/caldav/compare/v1.2.1...v1.3.5#diff-491a5dd4eece9b7ffe85041fb307fce0c648d69c707470fd697dbe2960a9ad7eR1848

sobolevn: Yes, because we have `ignore_missing_stub = true` set.
There are 45 exiting problems at the moment.

stubs/caldav/caldav/davclient.pyi
AlexWaygood: This is marked as resolved, but you haven't applied the change

sobolevn: Oh, it is because I was preparing https://github.com/python-caldav/caldav/pull/323 :)
I will change it now.

tobixen:
The 1.3 release seems to be cursed ... I suppose there will be a 1.3.7-release today as well

stubs/caldav/caldav/objects.pyi
```diff
@@ -221,6 +227,13 @@ class CalendarObjectResource(DAVObject):
     def add_organizer(self) -> None: ...
     def split_expanded(self) -> list[Self]: ...
     def expand_rrule(self, start: datetime.datetime, end: datetime.datetime) -> None: ...
+    def get_relatives(
+        self,
+        reltypes: Container[str] | None = None,
+        relfilter: Callable[[str], object] | None = None,
```
AlexWaygood, line 233: Looks like `relfilter` is expected to return a `bool`: https://github.com/python-caldav/caldav/compare/v1.2.1...v1.3.5#diff-491a5dd4eece9b7ffe85041fb307fce0c648d69c707470fd697dbe2960a9ad7eR1869-R1870

```suggestion
        relfilter: Callable[[str], bool] | None = None,
```

The fact that `rel` is explicitly `str`'d here also indicates to me that `rel` (the object being passed to `relfilter`) might not necessarily be a `str`: https://github.com/python-caldav/caldav/compare/v1.2.1...v1.3.5#diff-491a5dd4eece9b7ffe85041fb307fce0c648d69c707470fd697dbe2960a9ad7eR1874

_But_ I'm not sure what the type _should_ be

tobixen: bool is correct

tobixen: Sorry that this is lacking both documentation, tests and example code.

tobixen:
Now that the caldav library officially does not work with python2 anymore ... perhaps we could get type hints merged into the library itself?  I don't feel like prioritizing it myself, but I'd gladly accept merge/pull-requests.

stubs/caldav/caldav/objects.pyi
AlexWaygood: @tobixen, what's the expected type of the `rel` object that's _passed_ to the `relfilter` callable? Is it okay for the `relfilter` callable to assume that the object is going to be a `str`? Or should the parameter type be something more vague, like `object`?

AlexWaygood:
> Now that the caldav library officially does not work with python2 anymore ... perhaps we could get type hints merged into the library itself? I don't feel like prioritizing it myself, but I'd gladly accept merge/pull-requests.

That would be fantastic. The pytype typechecker has a `merge_pyi` tool that can help with automatically integrating stubs into runtime packages. (Not promising to take this on myself either, though 😄)

stubs/caldav/caldav/objects.pyi
tobixen: I regret creating this parameter at all ... I don't remember why I added it, it's not really needed (if one needs to do filtering, one can always do it after calling the function), there is no tests and no documentation.

Anyway, rel passed is a `icalendar.vText` attribute from the icalendar library.

AlexWaygood: Thanks so much for the help @tobixen <3

We can't have `icalendar` as a dependency for these stubs here in typeshed, unfortunately (as it's not `py.typed`), so the type hint we should use here for now should probably be:

```suggestion
        relfilter: Callable[[Any], bool] | None = None,
```

stubs/caldav/caldav/objects.pyi
```diff
@@ -1,14 +1,15 @@
 import datetime
 from _typeshed import Incomplete
-from collections.abc import Iterable, Iterator, Mapping, Sequence
+from collections import defaultdict
+from collections.abc import Callable, Container, Iterable, Iterator, Mapping, Sequence
 from typing import TypeVar, overload
```
AlexWaygood, line 5:
```suggestion
from typing import Any, TypeVar, overload
```

PR #10488, opened 2023-07-20
[stubsabot] Include repository in issue text

srittau:
Also, get GitHub repository for diff from METADATA.toml, instead of
scraping it from PyPI.

### DIFF
--- a/scripts/stubsabot.py
+++ b/scripts/stubsabot.py
@@ -262,6 +262,7 @@ async def get_github_repo_info(session: aiohttp.ClientSession, stub_info: StubIn
             tags: list[dict[str, Any]] = await response.json()
             assert isinstance(tags, list)
             return GithubInfo(repo_path=url_path, tags=tags)
+    return None
 
 
 class GithubDiffInfo(NamedTuple):
--- a/scripts/stubsabot.py
+++ b/scripts/stubsabot.py
@@ -248,20 +248,17 @@ async def get_github_repo_info(session: aiohttp.ClientSession, stub_info: StubIn
 
     Else, return None.
     """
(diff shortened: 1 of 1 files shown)


srittau (commented):
Add missing return

scripts/stubsabot.py
```diff
@@ -239,27 +241,27 @@ class GithubInfo:
...
+        return None
+    url_path = split_url.path.strip("/")
+    if len(Path(url_path).parts) != 2:
+        return None
+    github_tags_info_url = f"https://api.github.com/repos/{url_path}/tags"
+    async with session.get(github_tags_info_url, headers=get_github_api_headers()) as response:
+        if response.status == 200:
+            tags: list[dict[str, Any]] = await response.json()
+            assert isinstance(tags, list)
+            return GithubInfo(repo_path=url_path, tags=tags)
```
AlexWaygood, line 264: You could do this with a smaller diff relative to the existing function:

```suggestion
    if stub_info.upstream_repository:
        split_url = urllib.parse.urlsplit(stub_info.upstream_repository)
        if split_url.netloc == "github.com" and not split_url.query and not split_url.fragment:
            url_path = split_url.path.strip("/")
            if len(Path(url_path).parts) == 2:
                github_tags_info_url = f"https://api.github.com/repos/{url_path}/tags"
                async with session.get(github_tags_info_url, headers=get_github_api_headers()) as response:
                    if response.status == 200:
                        tags: list[dict[str, Any]] = await response.json()
                        assert isinstance(tags, list)
                        return GithubInfo(repo_path=url_path, tags=tags)
```

PR #10489, opened 2023-07-20
Add a few more upstream_repository URLs

Akuli:
Created with a slightly modified version of srittau's script from #10487:

<details>
<summary>script</summary>

```python
import asyncio
from pathlib import Path

import aiohttp

from scripts.stubsabot import fetch_pypi_info


async def run():
    async with aiohttp.ClientSession() as session:
        for p in Path("stubs").iterdir():
            if p.name == "gdb": continue
            if "upstream_repository" in (p / "METADATA.toml").read_text(): continue
            pypi_info = await fetch_pypi_info(p.name, session)
            print(p, pypi_info.info.get("project_urls"))
            the_url = None
            for url_name, url in (pypi_info.info.get("project_urls") or {}).items():
                url = url.replace("www.github.com", "github.com", 1)
                if url.startswith(("https://gitlab.com/", "https://github.com/", "https://bitbucket.org/")):
                    url = "/".join(url.split("/")[:5])  # truncate to https://site.com/user/repo
                    print("-->", url)
                    replace_metadata(p, url)


def replace_metadata(path: Path, repo_path: str) -> None:
    meta_path = path / "METADATA.toml"
    lines = meta_path.read_text().splitlines()
    url = repo_path
    new_lines = [lines[0], f'upstream_repository = "{url}"', *lines[1:]]
    meta_path.write_text("\n".join(new_lines) + "\n")


asyncio.run(run())
```

</details>

PR #10491, opened 2023-07-20
Auto-add upstream URLs in `create_baseline_stubs.py`

AlexWaygood:
Using a slightly-tweaked version of @Akuli's script from #10489 (which was itself a slightly tweaked version of @srittau's script from #10488!).

### DIFF
--- a/scripts/create_baseline_stubs.py
+++ b/scripts/create_baseline_stubs.py
@@ -20,8 +20,6 @@ import urllib.parse
 
 import aiohttp
 
-from stubsabot import fetch_pypi_info, get_github_api_headers
-
 if sys.version_info >= (3, 8):
     from importlib.metadata import distribution
 
@@ -76,18 +74,29 @@ def run_ruff(stub_dir: str) -> None:
     subprocess.run([sys.executable, "-m", "ruff", stub_dir])
 
 
+async def get_project_urls_from_pypi(project: str, session: aiohttp.ClientSession) -> list[str]:
+    pypi_root = f"https://pypi.org/pypi/{urllib.parse.quote(project)}"
+    async with session.get(f"{pypi_root}/json") as response:
+        response.raise_for_status()
+        j = await response.json()
+        if project_urls := j["info"].get("project_urls"):
+            assert isinstance(project_urls, dict)
+            return list(project_urls.values())
+        else:
+            return []
+
+
 async def get_upstream_repo_url(project: str) -> str | None:
     # aiohttp is overkill here, but it would also just be silly
     # to have both requests and aiohttp in our requirements-tests.txt file.
     async with aiohttp.ClientSession() as session:
-        pypi_info = await fetch_pypi_info(project, session)
-        for url in (pypi_info.info.get("project_urls") or {}).values():
+        for url in await get_project_urls_from_pypi(project, session):
             url = re.sub(r"^(https?://)?(www\.)", r"\1", url)
             netloc = urllib.parse.urlparse(url).netloc
             if netloc in {"gitlab.com", "github.com", "bitbucket.org", "foss.heptapod.net"}:
                 # truncate to https://site.com/user/repo
                 upstream_repo_url = "/".join(url.split("/")[:5])
-                async with session.get(upstream_repo_url, headers=get_github_api_headers()) as response:
+                async with session.get(upstream_repo_url) as response:
                     if response.status == 200:
                         return upstream_repo_url
     return None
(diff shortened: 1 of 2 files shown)


scripts/create_baseline_stubs.py
```diff
@@ -10,11 +10,16 @@
...
+import aiohttp
 import argparse
+import asyncio
 import os
 import re
 import subprocess
 import sys
+import urllib.parse
+
+from stubsabot import fetch_pypi_info, get_github_api_headers
```
AlexWaygood, line 22: Feels a bit weird importing from stubsabot here -- the other way of doing this would be to create a third module that they both import these functions from

AlexWaygood: After sleeping on it, I realised that importing from stubsabot here only actually saved us about 5 lines of code. It was unnecessary.

scripts/create_baseline_stubs.py
```diff
@@ -70,7 +74,36 @@ def run_ruff(stub_dir: str) -> None:
...
+    async with aiohttp.ClientSession() as session:
+        for url in await get_project_urls_from_pypi(project, session):
+            url = re.sub(r"^(https?://)?(www\.)", r"\1", url)
+            netloc = urllib.parse.urlparse(url).netloc
+            if netloc in {"gitlab.com", "github.com", "bitbucket.org", "foss.heptapod.net"}:
+                # truncate to https://site.com/user/repo
+                upstream_repo_url = "/".join(url.split("/")[:5])
+                async with session.get(upstream_repo_url) as response:
+                    if response.status == 200:
+                        return upstream_repo_url
```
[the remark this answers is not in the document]
AlexWaygood: Yes, it's an imprecise heuristic, but it's one that seems to work pretty well in practice -- we've been using it in stubsabot for a while, and I haven't noticed any incorrect results yet. This script is also only used for generating new stubs packages to contribute to typeshed, and those PRs tend to undergo careful manual review from us before being merged, so I think we'd probably spot it if the script added an incorrect URL here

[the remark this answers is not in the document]
AlexWaygood: I think you're right that the heuristic can be improved, though

scripts/create_baseline_stubs.py
```diff
@@ -79,9 +127,19 @@ def create_metadata(stub_dir: str, version: str) -> None:
     version = match.group(0)
     if os.path.exists(filename):
         return
+    metadata = f'version = "{version}.*"'
+    upstream_repo_url = asyncio.run(get_upstream_repo_url(project))
+    if upstream_repo_url is None:
+        warning = (
+            f"\nDid not succeed in finding a URL pointing to the upstream repo for {project!r}.\n"
```
Akuli, line 134: Slightly simpler:

```suggestion
            f"\nDid not succeed in finding a URL pointing to the source code repo for {project!r}.\n"
```

scripts/create_baseline_stubs.py
```diff
@@ -79,9 +127,19 @@ def create_metadata(stub_dir: str, version: str) -> None:
...
+            f"Please add it manually to the `stubs/{project}/METADATA.toml` file, if possible!\n"
```
Akuli, line 135: It could be hard to tell how exactly you're supposed to add it to the file. I think this is enough, if we assume the contributor looks at `CONTRIBUTING.md` or other `METADATA.toml` files:
```suggestion
            f"Please add it as `upstream_repository` to `stubs/{project}/METADATA.toml`, if possible!\n"
```

scripts/create_baseline_stubs.py
Akuli: Even simpler:
```suggestion
            f"\nDid not succeed in finding the source code repository of {project!r}.\n"
```

AlexWaygood:
```suggestion
            f"\nDid not succeed in finding a URL pointing to the source code for {project!r}.\n"
```

Akuli: What do you think of the "Even simpler" suggestion?

Akuli: Nevermind, I think it is clear enough as is.

AlexWaygood: Oh, I didn't see that one! I think I failed to refresh my browser.

Akuli (approved):

scripts/create_baseline_stubs.py
AlexWaygood: I pushed 00d2aa5, a slight variant

Issue #10494, opened 2023-07-21
Missing `url: str` in `stdlib.http.client.pyi`?


jamesbraza:
I have this code:

```python
from urllib import request

response: http.client.HTTPResponse = request.urlopen(request.Request("some_url", method="POST"))
response.url  # Warning on url
```

From the [`urllib.request` docs here](https://docs.python.org/3/library/urllib.request.html#module-urllib.response):
> The typical response object is a [urllib.response.addinfourl](https://docs.python.org/3/library/urllib.request.html#urllib.response.addinfourl) instance:

And `urllib.response.addinfourl` has:

> geturl()
>     _Deprecated since version 3.9:_ Deprecated in favor of [url](https://docs.python.org/3/library/urllib.request.html#urllib.response.addinfourl.url).

However, the `HTTPResponse` stub currently doesn't have `url: str` in it.  I think it should be added, would you agree?

---

I am willing to open a PR to fix this.

jamesbraza:
Alternately, should `http.client.HTTPResponse` inherit from (or mix in) `requests.response.addinfourl`?

sobolevn:
Yes, all documented attributes should be added :)

PR #10497, opened 2023-07-21
Added `url: str` to `http.client.HTTPResponse`

jamesbraza:
From https://docs.python.org/3/library/urllib.request.html#module-urllib.response:

> Functions defined by this module are used internally by the [urllib.request](https://docs.python.org/3/library/urllib.request.html#module-urllib.request) module. The typical response object is a [urllib.response.addinfourl](https://docs.python.org/3/library/urllib.request.html#urllib.response.addinfourl) instance:

The response of a [`urllib.request.urlopen`](https://docs.python.org/3/library/urllib.request.html#urllib.request.urlopen) for HTTP/HTTPS urls is a [`http.client.HTTPResponse`](https://docs.python.org/3/library/http.client.html#http.client.HTTPResponse).  Thus, I think `HTTPResponse` needs to conform to the interface of `addinfourl`.  Currently, `HTTPResponse` is missing the `url` field, which this PR adds.

Note that I placed `url: str` at the top of `HTTPResponse` to match `addinfourl`'s ordering.

Closes https://github.com/python/typeshed/issues/10494.

### DIFF
--- a/stdlib/http/client.pyi
+++ b/stdlib/http/client.pyi
@@ -103,7 +103,6 @@ class HTTPMessage(email.message.Message):
 def parse_headers(fp: io.BufferedIOBase, _class: Callable[[], email.message.Message] = ...) -> HTTPMessage: ...
 
 class HTTPResponse(io.BufferedIOBase, BinaryIO):  # type: ignore[misc]  # incompatible method definitions in the base classes
-    url: str
     msg: HTTPMessage
     headers: HTTPMessage
     version: int
@@ -137,6 +136,10 @@ class HTTPResponse(io.BufferedIOBase, BinaryIO):  # type: ignore[misc]  # incomp
     def geturl(self) -> str: ...
     def getcode(self) -> int: ...
     def begin(self) -> None: ...
+    # url is set on instances of the class in urllib.request.AbstractHTTPHandler.do_open
+    # to match urllib.response.addinfourl's interface.
+    # It's not set in HTTPResponse.__init__ or any other method on the class
+    url: str
 
 class HTTPConnection:
     auto_open: int  # undocumented
--- a/stdlib/http/client.pyi
+++ b/stdlib/http/client.pyi
@@ -103,6 +103,7 @@ class HTTPMessage(email.message.Message):
 def parse_headers(fp: io.BufferedIOBase, _class: Callable[[], email.message.Message] = ...) -> HTTPMessage: ...
 
 class HTTPResponse(io.BufferedIOBase, BinaryIO):  # type: ignore[misc]  # incompatible method definitions in the base classes
+    url: str
     msg: HTTPMessage
     headers: HTTPMessage
     version: int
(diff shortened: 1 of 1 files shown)


stdlib/http/client.pyi
```diff
@@ -103,6 +103,7 @@ class HTTPMessage(email.message.Message):
 def parse_headers(fp: io.BufferedIOBase, _class: Callable[[], email.message.Message] = ...) -> HTTPMessage: ...
 
 class HTTPResponse(io.BufferedIOBase, BinaryIO):  # type: ignore[misc]  # incompatible method definitions in the base classes
+    url: str
```
AlexWaygood, line 106: So, if I understand correctly, this attribute isn't set in the `__init__` method of this class, nor in any other method of this class; it's instead monkey-patched onto instances of this class by functions that live in `urllib.request` and/or `urllib.response`. Is that correct? If so, can we make it so that this is the last attribute annotation in the class body, and add a comment to that effect?

jamesbraza: Thanks for the response!  Just playing directly with `HTTPResponse`, I can see you're right.

```python
from http.client import HTTPResponse

response = HTTPResponse(some_sock)
response.url  # AttributeError
```

Note in the PR description there's justification for why the top line, though I now agree the bottom line makes more sense too (since `url` is not originally part of `HTTPResponse`, which is created by `urllib.request.HTTPSHandler`).

> it's instead monkey-patched onto instances of this class

This line in `urllib.request.AbstractHTTPHandler` adds the `url` attribute to the `HTTPResponse`: https://github.com/python/cpython/blob/main/Lib/urllib/request.py#L1334

What makes you say it's monkey patched?  To me, it seems like it's assigned.

---

Anyways, what do you think of the below comment?

```python
# NOTE: url is assigned by HTTP handler during do_open to match
# urllib.response.addinfourl's interface, not originally in HTTPResponse
url: str
```

AlexWaygood: > What makes you say it's monkey patched? To me, it seems like it's assigned.

You say tomato, I say tomato :P

"Monkey patching" is always just assignment in Python, right? Here I just used the term to indicate that an attribute that was never assigned to inside the class was being assigned to outside the class.

> Anyways, what do you think of the below comment?
> 
> ```python
> # NOTE: url is assigned by HTTP handler during do_open to match
> # urllib.response.addinfourl's interface, not originally in HTTPResponse
> url: str
> ```

How about:

```python
# url is set on instances of the class in urllib.request.AbstractHTTPHandler.do_open
# to match urllib.response.addinfourl's interface.
# It's not set in HTTPResponse.__init__ or any other method on the class
url: str
```

jamesbraza: Oh I actually had a misunderstanding of the term monkey patching.  I thought it only applied when using `patch` from `unittest.mock`, now I understand it's really a general term for dynamic attribute assignment.

Regardless, updated the comment to match this, thank you!

AlexWaygood (approved):
Thank you, looks great!

AlexWaygood:
(I realised I was slightly unclear about where exactly I wanted `url: str` to go; I've pushed to your branch to fix my nit there :)

PR #10502, opened 2023-07-24
Fix type of Future._exception

Dreamsorcerer:
Not even sure why this is `@property`, but it appears to default to None:
https://github.com/python/cpython/blob/4bbf071635504ee1c6d44f99c98e3fad191a3b13/Lib/asyncio/futures.py#L52
And even checks if it None:
https://github.com/python/cpython/blob/d93fe5afe2832a28b1e76e0a89d0a35c0d064bde/Lib/asyncio/futures.py#L202


hauntsaninja (approved):
Thanks!

PR #10504, opened 2023-07-25
Complete send2trash stubs

Akuli:


stubs/Send2Trash/METADATA.toml
```diff
@@ -1,6 +1,5 @@
 version = "1.8.*"
 upstream_repository = "https://github.com/arsenetar/send2trash"
-partial_stub = true
```
AlexWaygood, line 3: This field only changes whether `partial\n` is added to the `py.typed` file by the stub-uploader -- you'll want to tinker with the settings in the `tool.stubtest` table if you want to see what stuff is missing from the stubs

AlexWaygood:
I get this error running stubtest locally on Windows:

```
error: send2trash.compat.environb is not present at runtime
Stub: in file C:\Users\alexw\coding\typeshed\stubs\send2trash\send2trash\compat.pyi:7
Any
Runtime:
MISSING

Found 1 error (checked 6 modules)
```

Maybe we should just delete the `compat` submodule and allowlist it? It doesn't seem like it's meant to be imported by users, and I highly doubt any of our users are depending on it being in the stubs

stubs/Send2Trash/send2trash/__main__.pyi
```diff
@@ -0,0 +1,5 @@
+from collections.abc import Iterable
+
+from send2trash import send2trash as send2trash
```
AlexWaygood, line 3: We probably don't need the re-export here?

```suggestion
```

Akuli: stubgen "features" :)

PR #10505, opened 2023-07-25
[stubsabot] Add distributions argument

srittau:
This allows to specify which explicitly which distributions to update.
Useful for testing.


AlexWaygood (approved):
Nice, this is useful! I tried it out locally, and it worked as I would expect it to.

PR #10507, opened 2023-07-25
stubsabot: Add stubsabot label to PRs

srittau:


scripts/stubsabot.py
```diff
@@ -498,32 +500,50 @@ async def create_or_update_pull_request(*, title: str, body: str, branch_name: s
...
+        f"{TYPESHED_API_URL}/pulls",
         json={"title": title, "body": body, "head": f"{fork_owner}:{branch_name}", "base": "main"},
         headers=get_github_api_headers(),
     ) as response:
         resp_json = await response.json()
-        if response.status == 422 and any(
+        if response.status == 201:
+            pr_number = resp_json["number"]
+            assert isinstance(pr_number, int)
+        elif response.status == 422 and any(
```
AlexWaygood, line 511: Maybe we could use the `http.HTTPStatus` enum here. I'd fin this much more readable (I can't remember what 201 and 422 mean off the top of my head):

```suggestion
        if response.status == HTTPStatus.CREATED:
            pr_number = resp_json["number"]
            assert isinstance(pr_number, int)
        elif response.status == HTTPStatus.UNAUTHORIZED and any(
```

scripts/stubsabot.py
```diff
@@ -498,32 +500,50 @@ async def create_or_update_pull_request(*, title: str, body: str, branch_name: s
...
+    async with session.patch(
+        f"{TYPESHED_API_URL}/pulls/{pr_number}", json={"title": title, "body": body}, headers=get_github_api_headers()
+    ) as response:
+        response.raise_for_status()
+    return pr_number
+
+
+async def update_pull_request_label(*, pr_number: int, session: aiohttp.ClientSession) -> None:
+    async with session.post(
+        f"{TYPESHED_API_URL}/issues/{pr_number}/labels", json={"labels": [STUBSABOT_LABEL]}, headers=get_github_api_headers()
```
AlexWaygood, line 545:
```suggestion
        f"{TYPESHED_API_URL}/pulls/{pr_number}/labels", json={"labels": [STUBSABOT_LABEL]}, headers=get_github_api_headers()
```

srittau: That didn't work when I tried it. At least with GET.

srittau (commented):
😠😠😠😠

scripts/stubsabot.py
AlexWaygood: Oh, fair enough. https://github.com/python/typeshed/pull/10506 looked pretty convincing, so I guess this works as-is. Maybe worth adding a comment, though? It _looks_ incorrect, even if it isn't 😄

AlexWaygood (approved):

PR #10509, opened 2023-07-25
`nargs` could be `None` in `add_argument` method

Diogo-Rossi:
The argument `nargs` in stdlib `argparse.ArgumentParser.add_argument` could be `None`, according to the documentation
https://docs.python.org/3/library/argparse.html#nargs 

and also the source code https://github.com/python/cpython/blob/3.11/Lib/argparse.py 

The the absence of `None` cause false negative in code that is accepted

### DIFF
--- a/stdlib/argparse.pyi
+++ b/stdlib/argparse.pyi
@@ -85,7 +85,7 @@ class _ActionsContainer:
         self,
         *name_or_flags: str,
         action: _ActionStr | type[Action] = ...,
-        nargs: int | _NArgsStr | _SUPPRESS_T | None = ...,
+        nargs: int | _NArgsStr | _SUPPRESS_T | None = None,
         const: Any = ...,
         default: Any = ...,
         type: Callable[[str], _T] | FileType = ...,
--- a/stdlib/argparse.pyi
+++ b/stdlib/argparse.pyi
@@ -85,7 +85,7 @@ class _ActionsContainer:
         self,
         *name_or_flags: str,
         action: _ActionStr | type[Action] = ...,
-        nargs: int | _NArgsStr | _SUPPRESS_T = ...,
+        nargs: int | _NArgsStr | _SUPPRESS_T | None = ...,
         const: Any = ...,
         default: Any = ...,
         type: Callable[[str], _T] | FileType = ...,


AlexWaygood:
> The argument `nargs` in stdlib `argparse.ArgumentParser.add_argument` could be `None`, according to the documentation https://docs.python.org/3/library/argparse.html#nargs
> 
> and also the source code https://github.com/python/cpython/blob/3.11/Lib/argparse.py

Could you point to the specific bits in the docs or the source code that you think say that `nargs` could be `None` for the `add_argument` method specifically? To me, it looks like the docs say that `nargs` could be an int, or one of `"?"`, `"*"` or `"+"`. I don't see `None` mentioned anywhere :)

Diogo-Rossi:
> Could you point to the specific bits in the docs or the source code that you think say that `nargs` could be `None` for the `add_argument` method specifically? To me, it looks like the docs say that `nargs` could be an int, or one of `"?"`, `"*"` or `"+"`. I don't see `None` mentioned anywhere :)

Sure.


In the source code, it is on this snipet:

https://github.com/python/cpython/blob/1bc160947ea58fb3d11fa8667266bc80e9a14b7f/Lib/argparse.py#L2337-L2339
```python
...
# the default (None) is assumed to be a single argument
if nargs is None:
...
```
---
In the documentaion, it is implicit in this part

https://docs.python.org/3/library/argparse.html#nargs:~:text=Note%20that%20nargs%3D1%20produces%20a%20list%20of%20one%20item.%20This%20is%20different%20from%20the%20default%2C%20in%20which%20the%20item%20is%20produced%20by%20itself.

> [...]
> *Note that nargs=1 produces a list of one item. This is different from the default, in which the item is produced by itself.*
> [...]

Diogo-Rossi:
Actually, by reading the docs and the source, `None` seems to be the real default value of `nargs`.

May I add a commit setting the default to `None` instead of `...` ?

srittau:
Adding `None` to the types and as a default value sounds like the right choice to me, even if it's not explicitly documented. But I'll defer to @AlexWaygood's opinion.

Diogo-Rossi:
Ok, just added `None` as the default for `nargs`, like is in the source.

Fell free to comment and revert if it is not rigth.

srittau (approved):

AlexWaygood (approved):
Thanks!

PR #10513, opened 2023-07-28
Add more checks in `parse_metadata.py` for the `upstream_repository` field

AlexWaygood:
And remove duplicated checks from stubsabot

### DIFF
--- a/tests/parse_metadata.py
+++ b/tests/parse_metadata.py
@@ -6,6 +6,7 @@ from __future__ import annotations
 
 import os
 import re
+import urllib.parse
 from collections.abc import Mapping
 from dataclasses import dataclass
 from pathlib import Path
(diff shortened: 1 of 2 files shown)


tests/parse_metadata.py
```diff
@@ -199,6 +200,21 @@ def read_metadata(distribution: str) -> StubMetadata:
 
     upstream_repository: object = data.get("upstream_repository")
     assert isinstance(upstream_repository, (str, type(None)))
+    if isinstance(upstream_repository, str):
+        parsed_url = urllib.parse.urlsplit(upstream_repository)
+        assert parsed_url.scheme == "https", "URLs in the upstream_repository key should use https"
+        assert not parsed_url.netloc.startswith("www."), "`www.` should be removed from URLs in the upstream_repository key"
```
AlexWaygood, line 206:
```suggestion
        assert parsed_url.scheme == "https", "URLs in the upstream_repository field should use https"
        assert not parsed_url.netloc.startswith("www."), "`www.` should be removed from URLs in the upstream_repository field"
```

AlexWaygood:
Btw @srittau, I added a column for this data at https://alexwaygood.github.io/typeshed-stats/stats-csv/, and made the links clickable

PR #9768, opened 2023-02-20
`functools`: Add typing for decorated function arguments in cache API

asvetlov:
`functools.cache()` and `functools.lru_cache()` are affected.

PR #10523, opened 2023-07-31
`functools`: Add typing for decorated function arguments in cache API [take 2]

jakkdl:
Giving #9768 another attempt, adding [WMOkiishi](https://github.com/WMOkiishi)s suggested changes to see if that helps with mypy primer output.

Also rebased on top of `main/`


stdlib/functools.pyi
```diff
@@ -30,6 +30,8 @@ if sys.version_info >= (3, 9):
 
 _T = TypeVar("_T")
 _S = TypeVar("_S")
+_P = ParamSpec("_P")
```
Daverball, line 33:
```suggestion
_F = TypeVar("_F", bound=Callable[..., Any])
```

stdlib/functools.pyi
```diff
@@ -52,25 +54,35 @@ if sys.version_info >= (3, 9):
         typed: bool
 
 @final
-class _lru_cache_wrapper(Generic[_T]):
-    __wrapped__: Callable[..., _T]
-    def __call__(self, *args: Hashable, **kwargs: Hashable) -> _T: ...
+class _lru_cache_wrapper(Generic[_P, _T]):
+    __wrapped__: Callable[_P, _T]
+    def __call__(self, *args: _P.args, **kwargs: _P.kwargs) -> _T: ...
```
Daverball, line 59:
```suggestion
class _lru_cache_wrapper(_F):
    __wrapped__: _F
    __call__: _F
```

stdlib/functools.pyi
```diff
@@ -52,25 +54,35 @@ if sys.version_info >= (3, 9):
...
+    if sys.version_info >= (3, 8):
+        @overload
+        def __get__(self, __instance: None, __owner: type[_S] | None = ...) -> _lru_cache_wrapper[_P, _T]: ...
+        @overload
+        def __get__(self, __instance: _S, __owner: type[_S] | None = ...) -> Callable[Concatenate[_S, _P], _T]: ...
+    else:
+        @overload
+        def __get__(self, __instance: None, __owner: type[_S] | None) -> _lru_cache_wrapper[_P, _T]: ...
+        @overload
+        def __get__(self, __instance: _S, __owner: type[_S] | None) -> Callable[Concatenate[_S, _P], _T]: ...
```
Daverball, line 76:
```suggestion
    def __copy__(self) -> _lru_cache_wrapper[_F]: ...
    def __deepcopy__(self, __memo: Any) -> _lru_cache_wrapper[_F]: ...
    if sys.version_info >= (3, 8):
        def __get__(self, __instance: object | None, __owner: type[object] | None = ...) -> Self: ...
    else:
        def __get__(self, __instance: object | None, __owner: type[object] | None) -> Self: ...
```
I think this should always return itself, since you want the additional properties on the wrapper to be available regardless of whether you're accessing the wrapper from the instance or a class. It remaining callable with the same parameters is guaranteed by the `__call__` method.

stdlib/functools.pyi
```diff
@@ -52,25 +54,35 @@ if sys.version_info >= (3, 9):
...
 
 if sys.version_info >= (3, 8):
     @overload
-    def lru_cache(maxsize: int | None = 128, typed: bool = False) -> Callable[[Callable[..., _T]], _lru_cache_wrapper[_T]]: ...
+    def lru_cache(maxsize: int | None = 128, typed: bool = False) -> Callable[[Callable[_P, _T]], _lru_cache_wrapper[_P, _T]]: ...
```
Daverball, line 80:
```suggestion
    def lru_cache(maxsize: int | None = 128, typed: bool = False) -> Callable[_F], _lru_cache_wrapper[_F]]: ...
```

stdlib/functools.pyi
```diff
@@ -52,25 +54,35 @@ if sys.version_info >= (3, 9):
...
     @overload
-    def lru_cache(maxsize: Callable[..., _T], typed: bool = False) -> _lru_cache_wrapper[_T]: ...
+    def lru_cache(maxsize: Callable[_P, _T], typed: bool = False) -> _lru_cache_wrapper[_P, _T]: ...
```
Daverball, line 82:
```suggestion
    def lru_cache(maxsize: _F, typed: bool = False) -> _lru_cache_wrapper[_F]: ...
```

stdlib/functools.pyi
```diff
@@ -52,25 +54,35 @@ if sys.version_info >= (3, 9):
...
 
 else:
-    def lru_cache(maxsize: int | None = 128, typed: bool = False) -> Callable[[Callable[..., _T]], _lru_cache_wrapper[_T]]: ...
+    def lru_cache(maxsize: int | None = 128, typed: bool = False) -> Callable[[Callable[_P, _T]], _lru_cache_wrapper[_P, _T]]: ...
```
Daverball, line 85:
```suggestion
    def lru_cache(maxsize: int | None = 128, typed: bool = False) -> Callable[_F], _lru_cache_wrapper[_F]: ...
```

stdlib/functools.pyi
```diff
@@ -208,7 +220,7 @@ if sys.version_info >= (3, 8):
             def __class_getitem__(cls, item: Any) -> GenericAlias: ...
 
 if sys.version_info >= (3, 9):
-    def cache(__user_function: Callable[..., _T]) -> _lru_cache_wrapper[_T]: ...
+    def cache(__user_function: Callable[_P, _T]) -> _lru_cache_wrapper[_P, _T]: ...
```
Daverball, line 223:
```suggestion
    def cache(__user_function: _F) -> _lru_cache_wrapper[_F]: ...
```

Daverball (commented):
I think you might be overcomplicating things by trying to use `ParamSpec`, which I don't think is necessary here. It seems better to just forward the entire `Callable`, this way it doesn't matter what kind of a callable (method, classmethod, staticmethod, plain function, some arbitrary object with a __call__ method) it was.

Technically this could lead to some issues, since `_lru_cache_wrapper.__call__` should not actually return the original callable, but I think it's easier to make type checkers happy this way, since you don't have to worry about the `self` parameter and other subtleties.

Daverball:
Considering how hard you have to work with `ParamSpec` and `Concatenate` just to get to something that's close to being correct for some cases, but still not all of them, I'm inclined to believe that this would be better suited to an `Intersection` and would be yet another use case where `Intersection` would really shine and keep things relatively simple.

So in this case rather than trying to return a different type from `lru_cache` you just return an `Intersection` between the callable that's been passed in and a `Protocol` which contains the additional properties that were added to the function. i.e. something like this:
```python

class _CacheMethods(Protocol):
    def cache_info(self) -> _CacheInfo: ...
        def cache_parameters(self) -> _CacheParameters: ...

    @overload
    def lru_cache(maxsize: int | None = 128, typed: bool = False) -> Callable[[_F], Intersection[_F, _CacheMethods]]: ...
    @overload
    def lru_cache(maxsize: _F, typed: bool = False) -> Intersection[_F, _CacheMethods]: ...

else:
    def lru_cache(maxsize: int | None = 128, typed: bool = False) -> Callable[[_F], Intersection[_F, _CacheMethods]]: ...
```

jakkdl:
> Considering how hard you have to work with `ParamSpec` and `Concatenate` just to get to something that's close to being correct for some cases, but still not all of them, I'm inclined to believe that these just aren't the right tools for it, and it would be better suited to an `Intersection` and would be yet another use case where `Intersection` would really shine and keep things relatively simple by comparison.

Yeah after digging thoroughly into this I have more or less given up for the moment. It might be possible to improve the stubs a little bit merely using `Concatenate`, but in general the tools seems insufficient for the job. `Intersection` sounds kinda perfect, but with the PEP only being WIP I'm assuming it's not coming with 3.12

JelleZijlstra:
>  with the PEP only being WIP I'm assuming it's not coming with 3.12

Just FYI, assuming that PEP goes through, for typeshed usage the CPython release cycle doesn't really matter. The hypothetical timeline would go something like this:
- The PEP is created and gets a number
- `typing_extensions.Intersection` is added
- Type checkers add support for intersections
- The PEP is accepted
- Typeshed starts using `typing_extensions.Intersection`
- CPython implements the `&` operator for types (in 3.13 if this happens before ~May 2024, in 3.14 if after)

srittau:
@jakkdl Thanks for your work on this so far. Should we close the PR for the time being?

jakkdl:
> @jakkdl Thanks for your work on this so far. Should we close the PR for the time being?

yeah, might as well :)

tamird:
For the uninitiated: which PEP are you discussing above?

jakkdl:
> For the uninitiated: which PEP are you discussing above?

https://github.com/python/typing/issues/213#issuecomment-1648409245 iirc

Daverball:
> > For the uninitiated: which PEP are you discussing above?
> 
> [python/typing#213 (comment)](https://github.com/python/typing/issues/213#issuecomment-1648409245) iirc

Yes, exactly, the linked repository contains further information.

`OrderedIntersection` seems like it will get a PEP draft first however, courtesy of @mikeshardmind, read more about it here with the original pitch: https://github.com/CarliJoy/intersection_examples/issues/38

mikeshardmind:
👀 Let me add this issue to the list of motivating cases, thanks for the mention!

PR #10527, opened 2023-08-01
Add stubs for gevent

Daverball:
This is a pretty big package, because it re-implements a whole bunch of stdlib stuff so it can monkeypatch it to work cooperatively instead. There's also a lot of backwards compatibility cruft, as such the stubtest_allowlist is pretty massive.

There are still a couple of things to do so I'll put this into draft for now. For now I want to see if there are any complaints on the CI.


Daverball:
It looks like the internal pytype exception occur anywhere a a `Generic` class with a `ParamSpec` gets assigned a parameter list within the body of the class e.g. `Greenlet[[], None]` will cause this issue. Presumably because it's trying to cache bound types.

@rchen152 Could you take a look at this? Here's a minimal repro:
```python
from typing import Generic, TypeVar
from typing_extensions import ParamSpec

_P = ParamSpec('_P')
_T = TypeVar('_T')

class Test(Generic[_P, _T]):
    this_is_fine: Test[..., object]
    this_is_too: Test[_P, object]
    but_not_this: Test[[], object]
    or_this: Test[[object, object], object]

```

rchen152:
Thanks for the minimal repro! Looks like we're accidentally storing `[]` directly without processing it, hence the error when pytype tries to hash a tuple containing a list. This *looks* like it should be fairly easy to address; I'll try to get it fixed today and a fixed version of pytype out sometime next week.

Daverball:
I'm pretty happy with this. Tried the stubs out in one of my projects, really helped identify some subtle issues and potentially unsafe operations. I also found a minor issue with `gevent.spawn` and `gevent.spawn_later` since those are aliases for classmethods on `Greenlet`, which due to the order of the `__init__` overloads were bound to `Greenlet[[], None` rather than `Greenlet[_P, _T]`.

rchen152:
> Thanks for the minimal repro! Looks like we're accidentally storing `[]` directly without processing it, hence the error when pytype tries to hash a tuple containing a list. This _looks_ like it should be fairly easy to address; I'll try to get it fixed today and a fixed version of pytype out sometime next week.

Sorry this took a little longer than expected, but I just released pytype-2023.8.14, which contains a fix for this issue (https://github.com/google/pytype/commit/3e9a0138fc0ef504d9aea380b0211c345e98797c).

Daverball:
> > Thanks for the minimal repro! Looks like we're accidentally storing `[]` directly without processing it, hence the error when pytype tries to hash a tuple containing a list. This _looks_ like it should be fairly easy to address; I'll try to get it fixed today and a fixed version of pytype out sometime next week.
> 
> Sorry this took a little longer than expected, but I just released pytype-2023.8.14, which contains a fix for this issue ([google/pytype@3e9a013](https://github.com/google/pytype/commit/3e9a0138fc0ef504d9aea380b0211c345e98797c)).

No worries, I expect this pull request to sit around for another little while anyways, thanks for the fix!

srittau:
Seems like there are some problems with Python 3.12 now.

Daverball:
> Seems like there are some problems with Python 3.12 now.

Yes, it's because I punted on `gevent.ssl` (it just imports stdlib `ssl`) and `wrap_socket` was removed in 3.12. It should still be there in gevent though, since they implement their own.

pytype also really didn't like my workaround for the missing zope.interface stubs, so for now I gave up on staying compatible with the mypy plugin. It would be nice if there were soft dependencies in type checkers like you can get them in python using by catching `ImportError`. So type hints optionally can be more accurate without introducing a hard dependency on a third party stub that won't work with all the type checkers.

stubs/gevent/gevent/_abstract_linkable.pyi
```diff
@@ -0,0 +1,13 @@
...
+from gevent.hub import Hub
+
+class AbstractLinkable:
+    @property
+    def hub(self) -> Hub | None: ...
+    def __init__(self, hub: Hub | None = None) -> None: ...
+    def linkcount(self) -> int: ...
+    def rawlink(self, callback: Callable[[Self], object]) -> None: ...
+    def ready(self) -> bool: ...
+    def unlink(self, callback: Callable[[Self], object]) -> None: ...
```
JelleZijlstra, line 13:
```suggestion
    def rawlink(self, __callback: Callable[[Self], object]) -> None: ...
    def ready(self) -> bool: ...
    def unlink(self, __callback: Callable[[Self], object]) -> None: ...
```


```
In [26]: gevent._abstract_linkable.AbstractLinkable.unlink(1, callback=2)
Cell In[26], line 1
----> 1 gevent._abstract_linkable.AbstractLinkable.unlink(1, callback=2)

TypeError: unlink() takes no keyword arguments
```

stubs/gevent/gevent/_ffi/loop.pyi
```diff
@@ -0,0 +1,83 @@
...
+
+class AbstractLoop:
+    CALLBACK_CHECK_COUNT: int
+    error_handler: _ErrorHandler | None
+    starting_timer_may_update_loop_time: bool
+    # internal API, this __init__ will only be called from subclasses
+    def __init__(
+        self, ffi: object, lib: object, watchers: object, flags: int | None = ..., default: bool | None = ...
+    ) -> None: ...
+    def destroy(self) -> None: ...
```
JelleZijlstra, line 28:
```suggestion
    def destroy(self) -> bool | None: ...
```

stubs/gevent/gevent/_ffi/loop.pyi
```diff
@@ -0,0 +1,83 @@
...
+    def activecnt(self) -> int: ...
+    def io(self, fd: FileDescriptor, events: int, ref: bool = True, priority: int | None = None) -> _IoWatcher: ...
+    def closing_fd(self, fd: FileDescriptor) -> bool: ...
+    def timer(self, after: float, repeat: float = 0.0, ref: bool = True, priority: int | None = None) -> _TimerWatcher: ...
+    def signal(self, signum: int, ref: bool = True, priority: int | None = None) -> _Watcher: ...
+    def idle(self, ref: bool = True, priority: int | None = None) -> _Watcher: ...
+    def prepare(self, ref: bool = True, priority: int | None = None) -> _Watcher: ...
+    def check(self, ref: bool = True, priority: int | None = None) -> _Watcher: ...
+    if sys.platform != "win32":
+        def fork(self, ref: bool = True, priority: int | None = None) -> _Watcher: ...
```
JelleZijlstra, line 71: fork seems to be created unconditionally, only child and install_sigchld are in the if block at runtime.

stubs/gevent/gevent/_ffi/loop.pyi
```diff
@@ -0,0 +1,83 @@
...
+        def child(self, pid: int, trace: int = 0, ref: bool = True) -> _ChildWatcher: ...
+        def install_sigchld(self) -> None: ...
+
+    def async_(self, ref: bool = True, priority: int | None = None) -> _AsyncWatcher: ...
+    def stat(self, path: str, interval: float = 0.0, ref: bool = True, priority: bool | None = ...) -> _StatWatcher: ...
+    # These technically don't allow the functions arguments to be passed in as kwargs
```
JelleZijlstra, line 77: This is possible with TypeVarTuple, though mypy doesn't support that yet.

stubs/gevent/gevent/_ffi/loop.pyi
```diff
@@ -0,0 +1,83 @@
...
+    # but there's no way to express that yet with ParamSpec, however, we would still like
+    # to verify that the arguments match
+    def run_callback(self, func: Callable[_P, Any], *args: _P.args, **_: _P.kwargs) -> _Callback: ...
+    def run_callback_threadsafe(self, func: Callable[_P, Any], *args: _P.args, **_: _P.kwargs) -> _Callback: ...
```
JelleZijlstra, line 81:
```suggestion
    def run_callback(self, func: Callable[_P, object], *args: _P.args, **_: _P.kwargs) -> _Callback: ...
    def run_callback_threadsafe(self, func: Callable[_P, object], *args: _P.args, **_: _P.kwargs) -> _Callback: ...
```
Use `object` for ignored return types

stubs/gevent/gevent/_fileobjectcommon.pyi
```diff
@@ -0,0 +1,368 @@
...
+_IOT = TypeVar("_IOT", io.TextIOWrapper, io.FileIO, io.BufferedRandom, io.BufferedWriter, io.BufferedReader, IO[bytes], IO[Any])
+
+class cancel_wait_ex(IOError):
+    def __init__(self) -> None: ...
+
+class FileObjectClosed(IOError):
+    def __init__(self) -> None: ...
+
+class FlushingBufferedWriter(io.BufferedWriter):
+    def write(self, __b: ReadableBuffer) -> int: ...
```
JelleZijlstra, line 27: Can omit this method since it seems the same as the superclass.

stubs/gevent/gevent/_fileobjectcommon.pyi
```diff
@@ -0,0 +1,368 @@
...
+    ReadableBuffer,
+)
+from types import TracebackType
+from typing import IO, Any, AnyStr, ClassVar, Generic, TypeVar, overload
+from typing_extensions import Literal, Self
+
+from gevent.lock import DummySemaphore, Semaphore
+from gevent.threadpool import ThreadPool
+
+_IOT = TypeVar("_IOT", io.TextIOWrapper, io.FileIO, io.BufferedRandom, io.BufferedWriter, io.BufferedReader, IO[bytes], IO[Any])
```
JelleZijlstra, line 18: Constrained TypeVars are always a bit suspicious. They make sense if something supports a fixed set of classes. But this one includes the abstract `IO` class, so I don't think that applies here. Should we instead use something like `bound=IO[Any]`?

JelleZijlstra (commented):
I looked at the first few files, up to `_greenlet_primitives`. Overall this looks mostly good.

stubs/gevent/gevent/_ffi/loop.pyi
Daverball: True, however the docs state that `fork` isn't supported on any of the backends in win32, so if it's just going to throw an exception, there's not much of a point to include it in the stubs.

stubs/gevent/gevent/_fileobjectcommon.pyi
Daverball: Would that actually cover all the types? But I can certainly try it. Otherwise I suppose I can add whatever isn't covered into a union and make that the bound.

I think initially the type var only contained the specific classes from `io`, since I thought those were the only possible cases, but I had to add `IO[bytes]` and `IO[Any]` later on to cover a bunch of corner cases. But changing it to a type var with an upper bound instead of adding more types to the constrained type var, would probably have made more sense at the time.

Daverball: Looks like it's fine

stubs/gevent/gevent/_ffi/loop.pyi
Daverball: It's possible the docs are incorrect though, so I'd be open to lifting it outside the if statement.

stubs/gevent/gevent/_ffi/loop.pyi
Daverball: True, this also came up for `types-uWSGI`. For now a `ParamSpec` is probably still slightly better than just accepting any positional arguments.

stubs/gevent/gevent/fileobject.pyi
```diff
@@ -0,0 +1,156 @@
...
+            buffering: int | None = None,
+            closefd: bool | None = None,
+            atomic_write: bool = False,
+        ) -> None: ...
+
+    _FileObjectType: TypeAlias = type[FileObjectPosix[Any, Any] | FileObjectBlock[Any, Any] | FileObjectThread[Any, Any]]
+else:
+    _FileObjectType: TypeAlias = type[FileObjectBlock[Any, Any] | FileObjectThread[Any, Any]]
+
+FileObject: _FileObjectType = ...
```
srittau, line 156: Nit:

```suggestion
FileObject: _FileObjectType
```

srittau (approved):
I've browsed through the PR and nothing stands out as problematic. Considering the age and size of the PR I favor merging this. Any possible kinks can be ironed out later.

@JelleZijlstra had some remarks a few months back. Have these been answered?

Only une thing stood out: `__all__` is always added to the bottom of the file. Any particular reason for that? Commonly (at least here in typeshed) it's added to the top of the file.

Daverball:
> @JelleZijlstra had some remarks a few months back. Have these been answered?

I think the unresolved points haven't been addressed apart from the answer I have already given.

> Only une thing stood out: `__all__` is always added to the bottom of the file. Any particular reason for that? Commonly (at least here in typeshed) it's added to the top of the file.

No particular reason, other than gevent usually modifying `__all__` throughout the file, so it usually ends up further below in the source code, but I have no problem with moving them to the top if that is the convention in typeshed.

PR #10530, opened 2023-08-04
Add missing `TCP_USER_TIMEOUT` to `socket` module

sshane:
PR moved from https://github.com/python/mypy/pull/15811

Looks like this was backported from 3.7 to 3.6: https://bugs.python.org/issue26273 and https://github.com/python/cpython/blob/46366ca0486d07fe94c70d00771482c8ef1546fc/Doc/whatsnew/3.6.rst

### DIFF
--- a/stdlib/socket.pyi
+++ b/stdlib/socket.pyi
@@ -201,9 +201,10 @@ if sys.platform != "win32" and sys.platform != "darwin":
         TCP_LINGER2 as TCP_LINGER2,
         TCP_QUICKACK as TCP_QUICKACK,
         TCP_SYNCNT as TCP_SYNCNT,
-        TCP_USER_TIMEOUT as TCP_USER_TIMEOUT,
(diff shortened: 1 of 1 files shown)


stdlib/socket.pyi
```diff
@@ -203,6 +203,9 @@ if sys.platform != "win32" and sys.platform != "darwin":
         TCP_SYNCNT as TCP_SYNCNT,
         TCP_WINDOW_CLAMP as TCP_WINDOW_CLAMP,
     )
+
+    if sys.version_info >= (3, 6, 7):
```
AlexWaygood, line 207: We only support Python 3.7+ in typeshed, so the `sys.version_info` guard here is unnecessary :)

sshane:
done!

stdlib/socket.pyi
sshane: done!

AlexWaygood (approved):
Thanks!

PR #10531, opened 2023-08-04
fix(redis): improve typing of xadd/xdel

sileht:
typing copied from
https://redis-py.readthedocs.io/en/stable/commands.html#redis.commands.cluster.RedisClusterCommands.xadd
https://redis-py.readthedocs.io/en/stable/commands.html#redis.commands.cluster.RedisClusterCommands.xdel


sileht:
> Thank! Have you been able to take a look at the CI failures at all? :)

I fixed the issue.

stubs/redis/redis/commands/core.pyi
```diff
@@ -894,7 +894,7 @@ class StreamCommands:
         force=False,
         justid=False,
     ): ...
-    def xdel(self, name, *ids): ...
+    def xdel(self, name: _Key, *ids: str | int | bytes | memoryview): ...
```
AlexWaygood, line 897: Same issue here, please use `redis.typing.KeyT` instead of `redis.client._Key`

stubs/redis/redis/commands/core.pyi
```diff
@@ -928,8 +928,8 @@ class AsyncStreamCommands:
     async def xack(self, name, groupname, *ids): ...
     async def xadd(
         self,
-        name,
-        fields,
+        name: _Key,
```
AlexWaygood, line 931: Same issue here

stubs/redis/redis/commands/core.pyi
```diff
@@ -960,7 +960,7 @@ class AsyncStreamCommands:
         force=False,
         justid=False,
     ): ...
-    async def xdel(self, name, *ids): ...
+    async def xdel(self, name: _Key, *ids: str | int | bytes | memoryview): ...
```
AlexWaygood, line 963: Same issue here

AlexWaygood (changes requested):

AlexWaygood:
Hi @sileht, have you had a chance to look at my review comments at all? Shouldn't take long to fix :)

sileht:
> Hi @sileht, have you had a chance to look at my review comments at all? Shouldn't take long to fix :)

I made the update.

But something bothers me. Why does `_Key` even exists? It's used all over the files but I don't see `Union[bytes, str]` in redis-py documentation, only `Union[bytes, str, memoryview]`.

AlexWaygood (approved):
Thanks!

> But something bothers me. Why does `_Key` even exists? It's used all over the files but I don't see `Union[bytes, str]` in redis-py documentation, only `Union[bytes, str, memoryview]`.

Not sure. It's possible our annotations here predate the time when redis-py started added and documenting the expected types for various functions, in which case these types might be out of date. Feel free to file a follow-up PR if there are inaccuracies compared to what redis-py states in their docs/inline annotations :)

PR #10532, opened 2023-08-04
Fix `asyncio.Task.__init__` signature

sobolevn:
It was missing some params.

3.10: https://github.com/python/cpython/blob/3.10/Lib/asyncio/tasks.py#L93
3.11: https://github.com/python/cpython/blob/3.11/Lib/asyncio/tasks.py#L96
3.12: https://github.com/python/cpython/blob/3.12/Lib/asyncio/tasks.py#L101-L102


davidhalter:
@sobolevn Why did you use `name: str | None,` and not `name: str | None = ...`? It feels like this is wrong. This means that in Python 11+ the kw-only name argument would be mandatory. I assume that's wrong. Should I send a PR?

sobolevn:
@davidhalter yes, it is wrong. Please, send a PR with the fix :)

davidhalter:
Sent a pull request: https://github.com/python/typeshed/pull/10615

PR #10540, opened 2023-08-07
Filter(er) might return a LogRecord

kasium:
This feature is hidden in the docs: https://docs.python.org/3.13/library/logging.html.

> Changed in version 3.12: You can now return a [LogRecord](https://docs.python.org/3.13/library/logging.html#logging.LogRecord) instance from filters to replace the log record rather than modifying it in place. This allows filters attached to a [Handler](https://docs.python.org/3.13/library/logging.html#logging.Handler) to modify the log record before it is emitted, without having side effects on other handlers.


stdlib/logging/__init__.pyi
```diff
@@ -378,7 +378,7 @@ class Filter:
     name: str  # undocumented
     nlen: int  # undocumented
     def __init__(self, name: str = "") -> None: ...
-    def filter(self, record: LogRecord) -> bool: ...
+    def filter(self, record: LogRecord) -> bool | LogRecord: ...
```
AlexWaygood, line 381:
```suggestion
    if sys.version_info >= (3, 12):
        def filter(self, record: LogRecord) -> bool | LogRecord: ...
    else:
        def filter(self, record: LogRecord) -> bool: ...
```

AlexWaygood (changes requested):

AlexWaygood:
Should `logging.Filterer.filter` also be changed?

https://github.com/python/typeshed/blob/c326f2500b099903921af3b2c552c40bc3e0e5e0/stdlib/logging/__init__.pyi#L88

Your PR currently modifies `logging.Filter.filter()` and `logging.Manager.filter()`, but not `logging.Filterer.filter()`

kasium:
> Should `logging.Filterer.filter` also be changed?
> 
> https://github.com/python/typeshed/blob/c326f2500b099903921af3b2c552c40bc3e0e5e0/stdlib/logging/__init__.pyi#L88
> 
> Your PR currently modifies `logging.Filter.filter()` and `logging.Manager.filter()`, but not `logging.Filterer.filter()`

It does change logging.Filter and not logging:Manager

AlexWaygood:
> It does change logging.Filter and not logging:Manager

Oops, I misread the diff, apologies!

PR #10548, opened 2023-08-09
Update `turtle` to 3.12

sobolevn:
Source: https://github.com/python/cpython/blob/3.12/Lib/turtle.py


AlexWaygood (approved):

AlexWaygood:
Oh, stubtest is failing for the one in the global scope. Can't immediately see why, though.

sobolevn:
Because `turtle` uses some magic in `_make_global_funcs`, I am trying to figure out if this is correct.

sobolevn:
I find missing default and kw-only translated into positional-or-keyword as problems. I will report a bug to `cpython`.

AlexWaygood:
Why don't you revert the global-namespace `teleport` changes for now, until the CPython change is fixed? Then we can merge this now, and deal with the last remaining change later

sobolevn:
Sure!

AlexWaygood:
Thanks!

PR #10549, opened 2023-08-09
openpyxl: Class attributes set explicitly as another class'

Avasam:
Has overlap with #9511, so should reduce changes there.

I searched the openpyxl source code for instances of a class attribute being explicitly set as another class'.

In most cases, those will be descriptors, which pyright will have issues with re-assigning. Luckily, they'll also mostly refer to parent definitions. For typing purposes, this can be simplified to just not including them. I still left them in as comments to make it a bit more explicit.

Similarly, mypy has issues with re-assigning properties (see https://github.com/python/mypy/issues/13975) so I explicitly redefined those when they came from a non-parent class.


srittau (approved):
Looks reasonable, even if I didn't trace through some of the very dynamic code.

PR #10555, opened 2023-08-10
fnmatch: support PathLike

bluetech:
These are supported since Python 3.7 at least.


stdlib/fnmatch.pyi
```diff
@@ -1,9 +1,10 @@
...
 from typing import AnyStr
 
 __all__ = ["filter", "fnmatch", "fnmatchcase", "translate"]
 
-def fnmatch(name: AnyStr, pat: AnyStr) -> bool: ...
-def fnmatchcase(name: AnyStr, pat: AnyStr) -> bool: ...
-def filter(names: Iterable[AnyStr], pat: AnyStr) -> list[AnyStr]: ...
+def fnmatch(name: GenericPath, pat: GenericPath) -> bool: ...
+def fnmatchcase(name: GenericPath, pat: GenericPath) -> bool: ...
+def filter(names: Iterable[AnyStr], pat: GenericPath) -> list[AnyStr]: ...
```
bluetech, line 9: `names` has [an optimization](https://github.com/python/cpython/blob/eff2042fac6dc7d6e2f8aeec5a64a3a44a1a9648/Lib/fnmatch.py#L53-L54) which prevents `PathLike` from working on posix at least (probably from before PathLike was a thing), so I left it.

AlexWaygood (commented):
While it's true that these functions _do_ allow `PathLike` objects to be passed in, that feels like it's probably just an accident of the implementation that arises from the fact that they use `os.path` internally for normalising the inputs. It's not mentioned anywhere in the documentation that these functions support path-like objects, and nor is it tested anywhere. Moreover, the name of the module, `fnmatch`, is short for "filename-match", and a _name_ is nearly always a `str` or `bytes`.

I worry that by making this change, we'd be encouraging people to think that this is officially supported, whereas I think it's currently only accidentally supported, and could be accidentally broken at any time by CPython. The fact that `fnmatch.translate()` doesn't support path-like objects heightens my concern here.

bluetech:
@AlexWaygood What you say makes sense. I don't think it will ever be broken at this point, as that would be a breaking change, accidental or not. But I can see why typeshed wouldn't want to encode this.

AlexWaygood:
> I don't think it will ever be broken at this point, as that would be a breaking change, accidental or not.

Putting my CPython core dev hat on, I'm flattered by your estimation of our ability to spot accidental breaking changes when reviewing PRs at CPython ;)

_But_, since support for path-like objects is undocumented and untested, some on the core dev team might not even consider it a violation of our backwards-compatibility polcy to remove support for path-like objects without a deprecation period. There are differing opinions among the core dev team on that topic.

JelleZijlstra:
Might be good to open a CPython issue to clarify this. A possible outcome could be that we declare PathLike officially supported for fnmatch and add tests to that effect; alternatively we could make the documentation say explicitly that only strings are supported.

PR #10556, opened 2023-08-10
Update `sre_parse` to 3.12

sobolevn:
Source: https://github.com/python/cpython/blob/3.12/Lib/re/_parser.py


stdlib/sre_parse.pyi
```diff
@@ -87,14 +87,24 @@ class Tokenizer:
...
         def checkgroupname(self, name: str, offset: int, nested: int) -> None: ...
 
 def fix_flags(src: str | bytes, flags: int) -> int: ...
 
 _TemplateType: TypeAlias = tuple[list[tuple[int, int]], list[str | None]]
 _TemplateByteType: TypeAlias = tuple[list[tuple[int, int]], list[bytes | None]]
-if sys.version_info >= (3, 8):
+
+if sys.version_info >= (3, 12):
+    def parse(str: str, flags: int = 0, state: State | None = None) -> SubPattern: ...
```
AlexWaygood, line 101: It doesn't look like the definition of `parse` has changed in Python 3.12. Couldn't we do something like this?

```py
    @overload
    def parse_template(source: str, pattern: _Pattern[Any]) -> _TemplateType: ...
    @overload
    def parse_template(source: bytes, pattern: _Pattern[Any]) -> _TemplateByteType: ...

elif sys.version_info >= (3, 8):
    @overload
    def parse_template(source: str, state: _Pattern[Any]) -> _TemplateType: ...
    @overload
    def parse_template(source: bytes, state: _Pattern[Any]) -> _TemplateByteType: ...

else:
    @overload
    def parse_template(source: str, state: _Pattern[Any]) -> _TemplateType: ...
    @overload
    def parse_template(source: bytes, state: _Pattern[Any]) -> _TemplateByteType: ...

    def parse(str: str, flags: int = 0, state: State | None = None) -> SubPattern: ...

else:
    def parse(str: str, flags: int = 0, pattern: Pattern | None = None) -> SubPattern: ...
```

AlexWaygood (approved):
Thanks!

Issue #9710, opened 2023-02-11
PEP 692 (kwargs TypedDict)  tracker


hmc-cs-mdrissi:
Test example I'm using for support,

```python
from typing import TypedDict

from typing_extensions import Unpack


class Foo(TypedDict, total=False):
    a: int
    b: str


def f(**kwargs: Unpack[Foo]) -> None:
    ...


f(a=1, b="2")  # OK
f(a=1, b=2)  # Error: b has type str
f(a=1, b="2", c=3)  # Error: unexpected keyword argument "c"
```

- [x] mypy
- [x] pyright
- [ ] pytype google/pytype#1558
- [x] pyre
- [x] PyCharm (nice to have, but not required): Open [ticket](https://youtrack.jetbrains.com/issue/PY-55044).

pytype and pyre both produce 1 error on `f(**kwargs: Unpack[Foo]) `. Pytype produces `typing_extensions.Unpack not supported yet [not-supported-yet]` while pyre produces `pep_692.py:11:16 Undefined or invalid type [11]: Annotation Unpack is not defined as a type.`  

Unsure if typeshed usage for this should wait til pyre/pytype fully support pep 692 or if current status of them not checking lines reliant on 692 is enough (similar to Self type [status](https://github.com/python/typeshed/issues/6300)). pytype/pyre still detect errors normally for other parts of the file.

For mypy side we would also need to adjust line with [mypy flags](https://github.com/python/typeshed/blob/52cf716831ea11c08741db5f07ed740b3c9714c2/tests/mypy_test.py#L252) to include --enable-incomplete-feature=Unpack (maybe other incomplete features should be considered too).

Motivation for issue was yesterday I hit some lines in tensorflow stub where 692 would be helpful. Minor though and being explicit on arguments works fine (just adds a couple stubtest allowlist lines).

srittau:
Thanks for opening this! What would be the effect if someone uses a stub with mypy that uses this feature without `--enable-incomplete-feature`? Will it be treated as `Any` or will there be an error? In the latter case we can't use the feature, the former case would be acceptable in my opinion.

JelleZijlstra:
Since `--enable-incomplete-feature` is meant for experimental features that may not work properly yet, I think we shouldn't write typeshed stubs that rely on it. But on the mypy side, maybe we can move this aspect of `Unpack` out of "incomplete" status, since it's a lot simpler to implement than full PEP 646 support.

tmke8:
[...]

PR #10557, opened 2023-08-10
Add stubs for WTForms

Daverball:
This one's a bit of a doozy. I had already attempted to create stubs for this a few years back and gave up because they ended up being too annoying to use since the API isn't very friendly to static analysis.

But I've recently seen some examples in `types-PyYAML` that gave me renewed hope to attempt this and ended up with something I'm pretty happy with. It's not perfect, there are some things about the API that flat out don't work in type checkers:

`Field.__new__` will return by default an `UnboundField` which is not a subclass of `Field` and even if I were to make it a  faux-subclass and added overloads in `__new__` to make it work for the base class, I would then also need to add an overloaded `__new__` to every subclass and it would also break user-defined subclasses of `Field`. So that's not really sustainable. What I did instead was turn `Field` into a faux descriptor, that will return an `UnboundField` when accessed from the Form class and itself when it's accessed on the instance, that should cover most use-cases and it required only one additional hacky workaround for `FieldList` which has to accept `Field` in addition to `UnboundField`, if we don't want to force people to manually construct `UnboundField` instances for type checking.

Other than that I've added a few test cases for the more complex type protocols that can be passed into `Field`/`Form` such as `validators`, `widget` and `filters` since it was a little bit tricky to ensure there would be no false negatives while still retaining some amount of type safety.

I've integrated the stubs in two medium sized projects which both do some advanced things with the internals of WTForms in different ways and managed to catch a lot of the rough edges already.

A slight pain point right now is that form validation cannot result in any kind of type narrowing, so you will need to manually add some `assert form.field.data is not None` for fields with `InputRequired`/`DataRequired` validators. It shouldn't be too big of an issue though, since in a lot of cases you should just be using `form.populate_obj()` rather than accessing individual field data.

There also seems to be an issue with overriding `Field.widget` in subclasses, since it is both a `ClassVar` and instance variable. For some reason mypy will complain if subclasses annotate the field as `_Widget[Self]`, even though that matches the definition on the base class, so that seems like a false positive on mypy's end to me. This should only really be relevant though for user defined fields with a custom default widget that have subclasses themselves.

The docs for WTForms are pretty good and can be found here: https://wtforms.readthedocs.io/en/3.0.x/
Although it might still be quicker to just look at the source code: https://github.com/wtforms/wtforms/blob/3.0.1/src/wtforms/__init__.py


Daverball:
I tried out the stubs in a larger project and it was definitely a little bit more painful to use there, but this can mostly be attributed to the additional abstraction layer that project uses to dynamically create forms and the heavy use of `FormField`.

Other than that I found a small typing issue in `wtforms.form.BaseForm.errors` due to `FormField` violating the Liskov substitution principle. I considered changing it to `dict[str, Any]` to make it less annoying to use, but it doesn't actually come up that often. It's more common that errors will be inspected through individual fields.

There is still a type safety issue with `Field.errors` on the base class, due to the Liskov violation in `FormField` but it's probably still preferable over changing `Field.errors` to `Sequence[str] | dict[str, Sequence[str]`, since otherwise all custom Fields deriving from `Field` will need to override `errors` with the correct type, which is fine for the fields wtforms provides itself, but would be annoying for any user defined fields that derive from `Field`.

stubs/WTForms/wtforms/fields/choices.pyi
```diff
@@ -0,0 +1,78 @@
...
+        _form: BaseForm | None = None,
+        _prefix: str = "",
+        _translations: _SupportsGettextAndNgettext | None = None,
+        _meta: DefaultMeta | None = None,
+    ) -> None: ...
+    def iter_choices(self) -> Iterator[_FullChoice]: ...
+    def has_groups(self) -> bool: ...
+    def iter_groups(self) -> Iterator[_FullGroupedChoices]: ...
+    def __iter__(self) -> Iterator[_Option]: ...
+    _Option: type[_Option]
```
JelleZijlstra, line 44: Can you use a different name for the class? I'm not sure we can expect it to work reliably when you use the same name here both times.

stubs/WTForms/wtforms/fields/core.pyi
```diff
@@ -0,0 +1,133 @@
...
+    default: Any | None
+    description: str
+    render_kw: dict[str, Any]
+    filters: Sequence[_Filter]
+    flags: Flags
+    name: str
+    short_name: str
+    id: str
+    label: Label
+    # technically this can return UnboundField, but that is not allowed
```
JelleZijlstra, line 45: Would it be better to just omit this method? I'm not sure adding it gives us anything.

stubs/WTForms/wtforms/fields/core.pyi
```diff
@@ -0,0 +1,133 @@
...
+    # and we can't expect third parties to add a __new__ method to every field
+    # they define...
+    # This workaround only works for Form, not BaseForm, but we take what we can get
+    # BaseForm shouldn't really be used anyways
+    @overload
+    def __get__(self, obj: None, owner: type[object] | None = None) -> UnboundField[Self]: ...
+    @overload
+    def __get__(self, obj: object, owner: type[object] | None = None) -> Self: ...
+
+    # we put this at the bottom so it doesn't shadow builtins.type
```
JelleZijlstra, line 99: I don't think that makes sense in stubs

stubs/WTForms/wtforms/fields/datetime.pyi
```diff
@@ -0,0 +1,116 @@
...
+    ) -> None: ...
+
+class DateTimeLocalField(DateTimeField):
+    def __init__(
+        self,
+        label: str | None = None,
+        validators: tuple[_Validator[_FormT, Self], ...] | list[Any] | None = None,
+        format: str | list[str] = ...,
+        *,
+        filters: Sequence[_Filter] = (),
```
JelleZijlstra, line 105: This would be a really nice use case for PEP 692

stubs/WTForms/wtforms/fields/list.pyi
```diff
@@ -0,0 +1,44 @@
...
+        id: str | None = None,
+        widget: _Widget[FieldList[Any]] | None = None,
+        render_kw: dict[str, Any] | None = None,
+        name: str | None = None,
+        _form: BaseForm | None = None,
+        _prefix: str = "",
+        _translations: _SupportsGettextAndNgettext | None = None,
+        _meta: DefaultMeta | None = None,
+    ) -> None: ...
+    def append_entry(self, data: object = ...) -> _BoundFieldT: ...
```
JelleZijlstra, line 38:
```suggestion
    def append_entry(self, data: Any = ...) -> _BoundFieldT: ...
```
Doesn't seem like this can literally accept any object, so `Any` is more appropriate

stubs/WTForms/wtforms/fields/simple.pyi
```diff
@@ -0,0 +1,64 @@
...
+        render_kw: dict[str, Any] | None = None,
+        name: str | None = None,
+        _form: BaseForm | None = None,
+        _prefix: str = "",
+        _translations: _SupportsGettextAndNgettext | None = None,
+        _meta: DefaultMeta | None = None,
+    ) -> None: ...
+
+class StringField(Field):
+    data: str | None
```
JelleZijlstra, line 33: Would it make sense to make the base Field class generic over the data type?

stubs/WTForms/wtforms/i18n.pyi
```diff
@@ -0,0 +1,30 @@
...
+    def ugettext(self, __string: str) -> str: ...
+    def ungettext(self, __singular: str, __plural: str, __n: int) -> str: ...
+
+def messages_path() -> str: ...
+def get_builtin_gnu_translations(languages: Iterable[str] | None = None) -> GNUTranslations: ...
+@overload
+def get_translations(
+    languages: Iterable[str] | None = None, getter: Callable[[Iterable[str]], GNUTranslations] = ...
+) -> GNUTranslations: ...
+@overload
```
JelleZijlstra, line 17: This seems redundant with the next overload.

stubs/WTForms/wtforms/validators.pyi
```diff
@@ -0,0 +1,158 @@
...
+        self, min: float | Decimal | None = None, max: float | Decimal | None = None, message: str | None = None
+    ) -> None: ...
+    # any numeric field will work, for now we don't try to use a union
+    # to restrict to the defined numeric fields, since user-defined fields
+    # will likely not use a common base class, just like the existing
+    # numeric fields
+    def __call__(self, form: BaseForm, field: Field) -> None: ...
+
+class Optional:
+    string_check: Callable[[object], object]
```
JelleZijlstra, line 46: Seems like it's only expected to work with strs? And the return value is used as a boolean.
```suggestion
    string_check: Callable[[str], bool]
```

JelleZijlstra (commented):
Thanks, this looks like it was a heroic effort! I read through all the stubs and noted some minor issues.

stubs/WTForms/wtforms/fields/core.pyi
Daverball: `type` will be shadowed for the rest of the class body if i move it above so I either need to use `typing.Type` or define my own alias for the rest of the annotations in the class body that use `builtins.type`, so this seemed more pragmatic.

stubs/WTForms/wtforms/fields/core.pyi
Daverball: Considering the behavior of `__new__` might change in the future I wanted to leave a breadcrumb about the custom `__new__` so it can maybe be improved in the future. See https://github.com/python/mypy/pull/16020

stubs/WTForms/wtforms/fields/datetime.pyi
Daverball: Indeed, `Unpack` should not be used yet in typeshed though, or has this recently changed?

It may be worth to leave a comment though.

stubs/WTForms/wtforms/fields/simple.pyi
Daverball: I have considered it, and I think I even tried it out for a bit, but it ultimately seemed too complicated, for too little gain and would probably make matching `Widget` and `Validator` correctly even more of  a pain.

stubs/WTForms/wtforms/fields/choices.pyi
Daverball: I think it should be fine, since it's the final declaration in the class body. I would like to avoid changing the name, since you actually need to refer to this class sometimes, even though it's undocumented. The only reason it's even declared outside of the class body, is because pytype was crashing if I referred to it by `SelectFieldBase._Option`.

I could maybe try moving the declaration back into the class body, but put it as the very top and see if that solves the crash. The annotations within the class body could also get away with omitting the `SelectFieldBase.` prefix.

Daverball: Looks like moving the declaration to the top does not help pytype out either. Can you think of any other way to preserve the name, that's more robust?

stubs/WTForms/wtforms/i18n.pyi
Daverball: This allows for `get_translations(getter=...)` instead of `get_translations(None, ...)`. I would prefer not having to write two overloads either, but this is the only way to cover both call signatures.

stubs/WTForms/wtforms/fields/core.pyi
AlexWaygood: Mypy and pyright have different behaviour here (and pyright's behaviour arguably makes more sense, actually), so it's best not to make any assumptions on how this kind of thing will be interpreted by the type checker. See https://github.com/microsoft/pyright/issues/4856

stubs/WTForms/wtforms/fields/datetime.pyi
AlexWaygood: Can't be used yet, no, see #9710

stubs/WTForms/wtforms/fields/core.pyi
Daverball: That's fair. I will change it accordingly.

stubs/WTForms/wtforms/fields/choices.pyi
Daverball: I ended up defining a type alias for `type[_Option]` so it can be used for this one declaration, so that should be slightly more robust. Ideally `SelectFieldBase._Option` would be valid as a type, but I'm not sure how to achieve this in a way that will make pytype happy.

Daverball: I came up with a better solution of creating a type alias outside to refer to the class inside, now it doesn't matter if `_Option` refers to the class definition or the type alias inside the class body, the behavior should be the same and pytype is still happy with it too.

stubs/WTForms/wtforms/i18n.pyi
JelleZijlstra: Ah right, this is python/typing#1475. You have to really stare at these overloads sometimes.

JelleZijlstra:
The mypy-primer hits suggest the `| None` on the `data` fields will be annoying for real usage. Should we perhaps use `| Any` instead, or is there real unsafety in code like the pegen code flagged by mypy-primer? (https://github.com/we-like-parsers/pegen/blob/88e42cf776baa5c9fa1bf8e77993da1fbbf30409/src/pegen/web.py#L51 for reference)

Daverball:
> The mypy-primer hits suggest the `| None` on the `data` fields will be annoying for real usage. Should we perhaps use `| Any` instead, or is there real unsafety in code like the pegen code flagged by mypy-primer? (https://github.com/we-like-parsers/pegen/blob/88e42cf776baa5c9fa1bf8e77993da1fbbf30409/src/pegen/web.py#L51 for reference)

It can be a little bit annoying, but I ultimately decided that type safety is worth the little bit of extra hassle, especially considering it doesn't come up if you use `populate_obj`.

Most validators should guard against `None` anyways, since you can't rely on everyone putting the `Optional` validator first. So it only really comes up for small forms like login/search forms where you directly access the value of one or two fields, I don't think it's worth giving up type safety for that.

PR #10559, opened 2023-08-11
Bump mypy to 1.5.1

AlexWaygood:
As per the discussion in #10113, this means we stop testing with mypy/stubtest on Python 3.7, as mypy 1.5 no longer supports being run on Python 3.7. However, we're not removing _compatibility_ with Python 3.7 just yet -- we're not using PEP-570 syntax in stubs or getting rid of the Python 3.7 branches from stubs.


AlexWaygood:
Looks like something's gone wrong with the mypy release, which means that https://github.com/python/mypy/pull/15751 was accidentally left out of the release: https://github.com/python/mypy/issues/15588#issuecomment-1674554939

.vscode/settings.default.json
```diff
@@ -96,11 +96,10 @@
     "mypy-type-checker.importStrategy": "fromEnvironment",
     "mypy-type-checker.args": [
         "--custom-typeshed-dir=${workspaceFolder}",
-        "--python-version=3.7",
-        "--strict"
-        // Needed because a library stubbed in typeshed won't necessarily be installed inthe dev's environment
-        // Currentyl broken in dmypy: https://github.com/python/mypy/issues/10709
-        // "--ignore-missing-imports"
+        "--python-version=3.8",
```
srittau, line 99: Shouldn't we bump this to 3.11 while we're at it?

AlexWaygood: We promise in `tests/README.md` that our test suite can be run with Python 3.9+ (and some of our tests are run with Python 3.8 in CI, so that pip installs compatible versions of non-types dependencies in `mypy_test.py`). So... I guess we could debate whether we should be running mypy with `--python-version 3.8` or `--python-version 3.9` in CI. But 3.11 seems too high :)

srittau: Good point, I forgot that you can't check higher Python versions when using a lower Python at runtime. Next question: Should we even use `--python-version` in the config? It could be mildly surprising when a branch that is different from the installed Python version is checked in your editor.

AlexWaygood: But I don't have a strong opinion; I think trying to check typeshed using a mypy plugin in VSCode is probably always going to be a little broken :)

Avasam: It's the same reason we use the oldest supported version in `scripts/runtests.py`: older python versions tend to have less supported features. This was especially relevant for positional-only arguments using `/` on Python 3.7.

Given that:
1. In *theory* stubs shouldn't care about the Python version they're parse with.
2. I can see this giving some unexpected false positives in the IDE for TypeShed's own scripts

I think it makes sense to bump this to the minimal supported version for scripts. Maybe just add a comment to that effect.

AlexWaygood: Okay, cool, I'll bump to 3.9, since we only guarantee that the whole test suite can be run on Python 3.9+ (even though a couple of scripts will still need to be run on Python 3.8)

AlexWaygood:
```suggestion
        // We only guarantee all of our tests can be run if you're on Python 3.9 or higher
        "--python-version=3.9",
```

AlexWaygood:
All green now 🎉

hauntsaninja (approved):
Thank you!

PR #10563, opened 2023-08-11
Complete openpyxl title type annotations

Avasam:
Extracted and improved from #9511

- Completed and updated all title related type annotations I could find
- Use a `_Decodable` protocol instead of `bytes` for `_WorkbookChild.title`'s setter. The implementation explicitly checks for a `decode` attribute for non-string values instead of expecting a `bytes` (which relates to #9006)
- Didn't include what's already done in #10549


srittau (approved):
I did not compare to implementation, but looks plausible to me.

PR #10565, opened 2023-08-12
Improve `__(r)or__` signatures for TypedDict classes

AlexWaygood:
This isn't what was asked for in https://github.com/python/typing/issues/1445, but it seems worth doing anyway. I can't see a reason why it shouldn't be type-safe.


tests/stubtest_allowlists/py311.txt
```diff
@@ -137,6 +137,7 @@ typing\._SpecialForm.*
 typing._TypedDict.__delitem__
 typing._TypedDict.__ior__
 typing._TypedDict.__or__
+typing._TypedDict.__ror__
```
JelleZijlstra, line 140: We should probably just allowlist `typing\._TypedDict\..*` since the whole class is fictional

AlexWaygood: Yeah... there _is_ a `typing._TypedDict` class at runtime, but it has completely different semantics to what it's used for in the stubs: https://github.com/python/cpython/blob/2e1f688fe0f0a612e54c09f5a7027a834dd8b8d5/Lib/typing.py#L2989-L2990

JelleZijlstra (approved):

PR #10570, opened 2023-08-13
Make `openpyxl` stubtest-complete

Avasam:
Removes `ignore_missing_stub` and `partial_stub` from `openpyxl`'s metadata. There wasn't much missing, so this is simple enough.
Has some overlap with #9511 so will reduce changes there too.


stubs/openpyxl/openpyxl/pivot/table.pyi
```diff
@@ -953,3 +953,6 @@ class TableDefinition(Serialisable):
     def to_tree(self): ...
     @property
     def path(self): ...
+    def formatted_fields(self) -> dict[Incomplete, Incomplete]: ...
```
srittau, line 956:
```suggestion
    def formatted_fields(self) -> dict[Incomplete, list[Incomplete]]: ...
```

stubs/openpyxl/openpyxl/reader/strings.pyi
```diff
@@ -1 +1,6 @@
+from xml.etree.ElementTree import _FileRead
+
+from openpyxl.cell.rich_text import CellRichText
+
 def read_string_table(xml_source): ...
```
srittau, line 5:
```suggestion
def read_string_table(xml_source: _FileRead) -> list[str]: ...
```

srittau (approved):
Thanks, a few optional suggestions below.

PR #10571, opened 2023-08-13
Fill in all missing `upstream_repository`

Avasam:
I was navigating https://alexwaygood.github.io/typeshed-stats/stats-csv/ for unrelated reasons and noticed a lot of missing upstream source / url, and decided it'd be fun to go on a hunt.

The only actually closed-source project seems to be SAP's hdbcli.
Most of these were linked in their docs (although sometimes very obscure). Except for python-ini which has no documentation and no link on PyPI. But it's the same author so I assume it's the right repository.

I've updated the `upstream_repository` field validation to make the GDB change pass. I believe this respects the original intent of avoiding extra query params and fragments that may have accidentally been copied. And added more detailed error messages.

### DIFF
--- a/stubs/beautifulsoup4/METADATA.toml
+++ b/stubs/beautifulsoup4/METADATA.toml
@@ -1,5 +1,5 @@
 version = "4.12.*"
-upstream_repository = "https://git.launchpad.net/beautifulsoup"
+upstream_repository = "https://git.launchpad.net/beautifulsoup/tree"
 requires = ["types-html5lib"]
 partial_stub = true
 
--- a/stubs/beautifulsoup4/METADATA.toml
+++ b/stubs/beautifulsoup4/METADATA.toml
@@ -1,4 +1,5 @@
 version = "4.12.*"
+upstream_repository = "https://git.launchpad.net/beautifulsoup"
 requires = ["types-html5lib"]
 partial_stub = true
 
--- a/stubs/docutils/METADATA.toml
+++ b/stubs/docutils/METADATA.toml
@@ -1,4 +1,5 @@
 version = "0.20.*"
+upstream_repository = "https://sourceforge.net/p/docutils/code"
 partial_stub = true
 
 [tool.stubtest]
--- a/stubs/gdb/METADATA.toml
+++ b/stubs/gdb/METADATA.toml
@@ -1,4 +1,7 @@
 version = "12.1.*"
+# This is the official web portal for GDB,
+# see https://sourceware.org/gdb/current/ for other ways of obtaining the source code.
+upstream_repository = "https://sourceware.org/git/gitweb.cgi?p=binutils-gdb.git"
 extra_description = """\
   Type hints for GDB's \
   [Python API](https://sourceware.org/gdb/onlinedocs/gdb/Python-API.html). \
--- a/stubs/gdb/METADATA.toml
+++ b/stubs/gdb/METADATA.toml
@@ -1,7 +1,7 @@
 version = "12.1.*"
 # This is the official web portal for GDB,
 # see https://sourceware.org/gdb/current/ for other ways of obtaining the source code.
-upstream_repository = "https://sourceware.org/git/gitweb.cgi?p=binutils-gdb.git"
+upstream_repository = "https://sourceware.org/git/gitweb.cgi?p=binutils-gdb.git;a=tree"
 extra_description = """\
   Type hints for GDB's \
   [Python API](https://sourceware.org/gdb/onlinedocs/gdb/Python-API.html). \
--- a/stubs/pyaudio/METADATA.toml
+++ b/stubs/pyaudio/METADATA.toml
@@ -1,8 +1,7 @@
 version = "0.2.*"
 # There is no web portal for the source, this is the official source link:
 # https://people.csail.mit.edu/hubert/pyaudio/#sources
-# https://people.csail.mit.edu/hubert/git/pyaudio.git
-# upstream_repository =
+upstream_repository = "https://people.csail.mit.edu/hubert/git/pyaudio.git"
 
 [tool.stubtest]
 # linux and win32 are equivalent
--- a/stubs/pyaudio/METADATA.toml
+++ b/stubs/pyaudio/METADATA.toml
@@ -1,4 +1,7 @@
 version = "0.2.*"
+# There is no web portal for the source, this is the official source link:
+# https://people.csail.mit.edu/hubert/pyaudio/#sources
+upstream_repository = "https://people.csail.mit.edu/hubert/git/pyaudio.git"
 
 [tool.stubtest]
 # linux and win32 are equivalent
--- a/stubs/pyaudio/METADATA.toml
+++ b/stubs/pyaudio/METADATA.toml
@@ -1,6 +1,8 @@
 version = "0.2.*"
-# There is no web portal for the source, this is the official source link
-upstream_repository = "https://people.csail.mit.edu/hubert/pyaudio/#sources"
+# There is no web portal for the source, this is the official source link:
+# https://people.csail.mit.edu/hubert/pyaudio/#sources
+# https://people.csail.mit.edu/hubert/git/pyaudio.git
+# upstream_repository =
 
 [tool.stubtest]
 # linux and win32 are equivalent
--- a/stubs/pyaudio/METADATA.toml
+++ b/stubs/pyaudio/METADATA.toml
@@ -1,7 +1,6 @@
 version = "0.2.*"
 # There is no web portal for the source, this is the official source link:
-# https://people.csail.mit.edu/hubert/pyaudio/#sources
-upstream_repository = "https://people.csail.mit.edu/hubert/git/pyaudio.git"
+# upstream_repository = "https://people.csail.mit.edu/hubert/pyaudio/#sources"
 
 [tool.stubtest]
 # linux and win32 are equivalent
--- a/stubs/pyaudio/METADATA.toml
+++ b/stubs/pyaudio/METADATA.toml
@@ -1,7 +1,6 @@
 version = "0.2.*"
-# There is no web portal for the source, this is the official source link:
-# https://people.csail.mit.edu/hubert/pyaudio/#sources
-upstream_repository = "https://people.csail.mit.edu/hubert/git/pyaudio.git"
+# There is no web portal for the source, this is the official source link
+upstream_repository = "https://people.csail.mit.edu/hubert/pyaudio/#sources"
 
 [tool.stubtest]
 # linux and win32 are equivalent
--- a/tests/parse_metadata.py
+++ b/tests/parse_metadata.py
@@ -208,8 +208,8 @@ def read_metadata(distribution: str) -> StubMetadata:
         assert not parsed_url.netloc.startswith(
             "www."
         ), "`World Wide Web` subdomain (`www.`) should be removed from URLs in the upstream_repository field"
-        assert (
-            parsed_url.hostname not in QUERY_URL_ALLOWLIST and not parsed_url.query
+        assert parsed_url.hostname in QUERY_URL_ALLOWLIST or (
+            not parsed_url.query
         ), "Query params (`?`) should be removed from URLs in the upstream_repository field"
         assert not parsed_url.fragment, "Fragments (`#`) should be removed from URLs in the upstream_repository field"
         if parsed_url.netloc == "github.com":
--- a/tests/parse_metadata.py
+++ b/tests/parse_metadata.py
@@ -205,11 +205,13 @@ def read_metadata(distribution: str) -> StubMetadata:
     if isinstance(upstream_repository, str):
         parsed_url = urllib.parse.urlsplit(upstream_repository)
         assert parsed_url.scheme == "https", "URLs in the upstream_repository field should use https"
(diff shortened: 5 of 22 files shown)


AlexWaygood:
(It's possible that the validation we're doing in `parse_metadata.py` is overly strict at the moment)

stubs/pyaudio/METADATA.toml
```diff
@@ -1,7 +1,6 @@
 version = "0.2.*"
-# There is no web portal for the source, this is the official source link:
-# https://people.csail.mit.edu/hubert/pyaudio/#sources
-upstream_repository = "https://people.csail.mit.edu/hubert/git/pyaudio.git"
+# There is no web portal for the source, this is the official source link
+upstream_repository = "https://people.csail.mit.edu/hubert/pyaudio/#sources"
```
Avasam, line 3: @AlexWaygood This is my usage of a git URL or a fragment. I'll let you decide on whether this value or the restrictive parsing should change, depending on your needs. It's definitely an edge case.

tests/parse_metadata.py
```diff
@@ -203,9 +205,13 @@ def read_metadata(distribution: str) -> StubMetadata:
...
         parsed_url = urllib.parse.urlsplit(upstream_repository)
         assert parsed_url.scheme == "https", "URLs in the upstream_repository field should use https"
-        assert not parsed_url.netloc.startswith("www."), "`www.` should be removed from URLs in the upstream_repository field"
-        assert not parsed_url.query
-        assert not parsed_url.fragment
+        assert not parsed_url.netloc.startswith(
+            "www."
+        ), "`World Wide Web` subdomain (`www.`) should be removed from URLs in the upstream_repository field"
+        assert (
+            parsed_url.hostname not in QUERY_URL_ALLOWLIST and not parsed_url.query
```
AlexWaygood, line 212:
```suggestion
            parsed_url.hostname in QUERY_URL_ALLOWLIST or (not parsed_url.query)
```

stubs/pyaudio/METADATA.toml
Avasam: I've updated the `upstream_repository` field validation. I believe this respects the original intent of avoiding extra query params and fragments that may have accidentally been copied.

stubs/beautifulsoup4/METADATA.toml
```diff
@@ -1,4 +1,5 @@
 version = "4.12.*"
+upstream_repository = "https://git.launchpad.net/beautifulsoup"
```
AlexWaygood, line 2: Maybe this would be more useful? I generally don't care too much about the git log when I'm looking at an upstream repository to review a typeshed PR, I just wanna see the source 😆

```suggestion
upstream_repository = "https://git.launchpad.net/beautifulsoup/tree"
```

tests/parse_metadata.py
```diff
@@ -31,6 +31,8 @@
 
 
 _STUBTEST_PLATFORM_MAPPING: Final = {"linux": "apt_dependencies", "darwin": "brew_dependencies", "win32": "choco_dependencies"}
+QUERY_URL_ALLOWLIST = {"sourceware.org"}
+"""Some older websites have a bad pattern of using query params for navigation"""
```
Avasam, line 35: Feel free to wordsmith this.

stubs/gdb/METADATA.toml
```diff
@@ -1,4 +1,7 @@
 version = "12.1.*"
+# This is the official web portal for GDB,
+# see https://sourceware.org/gdb/current/ for other ways of obtaining the source code.
+upstream_repository = "https://sourceware.org/git/gitweb.cgi?p=binutils-gdb.git"
```
AlexWaygood, line 4: Idk how many people are really going to be interested in the source code for `gdb` (or if this will pass our validation checks), but:

```suggestion
upstream_repository = "https://sourceware.org/git/?p=binutils-gdb.git;a=tree"
```

stubs/pyaudio/METADATA.toml
```diff
@@ -1,4 +1,7 @@
 version = "0.2.*"
+# There is no web portal for the source, this is the official source link:
+# https://people.csail.mit.edu/hubert/pyaudio/#sources
+upstream_repository = "https://people.csail.mit.edu/hubert/git/pyaudio.git"
```
AlexWaygood, line 4: This is a bit weird; neither link would take us to a version of the source that you could view in a web browser. I think I'd prefer just doing this tbh:

```suggestion
# There is no web portal for the source, this is the official source link:
# https://people.csail.mit.edu/hubert/pyaudio/#sources
```

tests/parse_metadata.py
AlexWaygood, line 35:
```suggestion
# Some older websites have a bad pattern of using query params for navigation
QUERY_URL_ALLOWLIST = {"sourceware.org"}
```

tests/parse_metadata.py
```diff
@@ -203,9 +205,13 @@ def read_metadata(distribution: str) -> StubMetadata:
     if isinstance(upstream_repository, str):
         parsed_url = urllib.parse.urlsplit(upstream_repository)
         assert parsed_url.scheme == "https", "URLs in the upstream_repository field should use https"
```
AlexWaygood, line 207:
```suggestion
        assert parsed_url.scheme == "https", f"{distribution}: URLs in the upstream_repository field should use https"
```

tests/parse_metadata.py
```diff
@@ -203,9 +205,13 @@ def read_metadata(distribution: str) -> StubMetadata:
...
-        assert not parsed_url.netloc.startswith("www."), "`www.` should be removed from URLs in the upstream_repository field"
-        assert not parsed_url.query
-        assert not parsed_url.fragment
+        assert not parsed_url.netloc.startswith(
+            "www."
+        ), "`World Wide Web` subdomain (`www.`) should be removed from URLs in the upstream_repository field"
```
AlexWaygood, line 210:
```suggestion
        ), f"{distribution}: `World Wide Web` subdomain (`www.`) should be removed from URLs in the upstream_repository field"
```

tests/parse_metadata.py
```diff
@@ -203,9 +205,13 @@ def read_metadata(distribution: str) -> StubMetadata:
...
+        assert parsed_url.hostname in QUERY_URL_ALLOWLIST or (
+            not parsed_url.query
+        ), "Query params (`?`) should be removed from URLs in the upstream_repository field"
+        assert not parsed_url.fragment, "Fragments (`#`) should be removed from URLs in the upstream_repository field"
```
AlexWaygood, line 214:
```suggestion
        ), f"{distribution}: Query params (`?`) should be removed from URLs in the upstream_repository field"
        assert not parsed_url.fragment, f"{distribution}: Fragments (`#`) should be removed from URLs in the upstream_repository field"
```

AlexWaygood (commented):
Thanks!

stubs/docutils/METADATA.toml
```diff
@@ -1,4 +1,5 @@
 version = "0.20.*"
+upstream_repository = "https://sourceforge.net/p/docutils/code"
```
AlexWaygood, line 2: Oh, could you also add sourceforge to this set here?

https://github.com/python/typeshed/blob/df08fcec5f281ba01266869ec55368819aed0a77/scripts/create_baseline_stubs.py#L112

stubs/beautifulsoup4/METADATA.toml
Avasam: Agreed. For example docutils' sourceforge link goes directly to the code tab.
I'll update the launchpad link in pytz comment too

stubs/gdb/METADATA.toml
Avasam: It should, it's just part of the queryparam, will try

AlexWaygood: is it obvious i'm not a web developer

tests/parse_metadata.py
Avasam: I don't mind either way, but I liked the docstring showing up in IDEs ^^
image

AlexWaygood: To me it just looks pretty odd doing something I thought was Sphinx-specific when we don't autogenerate any docs from this file. I'd much prefer the comment here, I'm afraid :)

stubs/docutils/METADATA.toml
Avasam: I can. It seems the logic will truncate to `https://sourceforge.net/p/repo`, is that good here? I could special-case sourceforge to add back `/code`

AlexWaygood: Okay, since this is non-trivial in complexity, let's postpone it to a followup PR :)

tests/parse_metadata.py
```diff
@@ -202,10 +204,16 @@ def read_metadata(distribution: str) -> StubMetadata:
...
+        assert parsed_url.scheme == "https", f"{distribution}: URLs in the upstream_repository field should use https"
+        assert not parsed_url.netloc.startswith(
+            "www."
+        ), f"{distribution}: `World Wide Web` subdomain (`www.`) should be removed from URLs in the upstream_repository field"
+        assert parsed_url.hostname in QUERY_URL_ALLOWLIST or (
+            not parsed_url.query
+        ), f"{distribution}: Query params (`?`) should be removed from URLs in the upstream_repository field"
+        assert (
+            not parsed_url.fragment
+        ), f"{distribution}: Fragments (`#`) should be removed from URLs in the upstream_repository field"
```
AlexWaygood, line 216: I can't bear the way black formats these lines, I'm afraid ;)

```suggestion
        no_www_please = (
            f"{distribution}: `World Wide Web` subdomain (`www.`) should be removed from URLs in the upstream_repository field"
        )
        assert not parsed_url.netloc.startswith("www."), no_www_please
        no_query_params_please = (
            f"{distribution}: Query params (`?`) should be removed from URLs in the upstream_repository field"
        )
        assert parsed_url.hostname in QUERY_URL_ALLOWLIST or (not parsed_url.query), no_query_params_please
        no_fragments_please = f"{distribution}: Fragments (`#`) should be removed from URLs in the upstream_repository field"
        assert not parsed_url.fragment, no_fragments_please
```

tests/parse_metadata.py
```diff
@@ -31,6 +31,8 @@
 
 
 _STUBTEST_PLATFORM_MAPPING: Final = {"linux": "apt_dependencies", "darwin": "brew_dependencies", "win32": "choco_dependencies"}
+# Some older websites have a bad pattern of using query params for navigation.
+QUERY_URL_ALLOWLIST = {"sourceware.org"}
```
AlexWaygood, line 35: Actually, yes, could you make this private? Not a big deal, but it would be consistent with `_STUBTEST_PLATFORM_MAPPING` immediately above

tests/parse_metadata.py
Avasam: This is one of my big pet peaves with black too. I'll go with your changes.

tests/parse_metadata.py
Avasam:
```suggestion
_QUERY_URL_ALLOWLIST = {"sourceware.org"}
```

tests/parse_metadata.py
```diff
@@ -202,10 +204,16 @@ def read_metadata(distribution: str) -> StubMetadata:
...
         parsed_url = urllib.parse.urlsplit(upstream_repository)
-        assert parsed_url.scheme == "https", "URLs in the upstream_repository field should use https"
-        assert not parsed_url.netloc.startswith("www."), "`www.` should be removed from URLs in the upstream_repository field"
-        assert not parsed_url.query
-        assert not parsed_url.fragment
+        assert parsed_url.scheme == "https", f"{distribution}: URLs in the upstream_repository field should use https"
+        assert not parsed_url.netloc.startswith(
+            "www."
+        ), f"{distribution}: `World Wide Web` subdomain (`www.`) should be removed from URLs in the upstream_repository field"
+        assert parsed_url.hostname in QUERY_URL_ALLOWLIST or (
```
Avasam, line 211:
```suggestion
        assert parsed_url.hostname in _QUERY_URL_ALLOWLIST or (
```

tests/parse_metadata.py
```diff
@@ -202,10 +204,17 @@ def read_metadata(distribution: str) -> StubMetadata:
...
-        assert not parsed_url.fragment
+        assert parsed_url.scheme == "https", f"{distribution}: URLs in the upstream_repository field should use https"
+        no_www_please = (
+            f"{distribution}: `World Wide Web` subdomain (`www.`) should be removed from URLs in the upstream_repository field"
+        )
+        assert not parsed_url.netloc.startswith("www."), no_www_please
+        no_query_params_please = (
+            f"{distribution}: Query params (`?`) should be removed from URLs in the upstream_repository field"
+        )
+        assert parsed_url.hostname in QUERY_URL_ALLOWLIST or (not parsed_url.query), no_query_params_please
```
Avasam, line 215:
```suggestion
        assert parsed_url.hostname in _QUERY_URL_ALLOWLIST or (not parsed_url.query), no_query_params_please
```

AlexWaygood (approved):
Thanks!

srittau:
Thanks! These are actually the packages where the `upstream_repository` field is most useful and why I wanted it in the first place.

PR #10572, opened 2023-08-13
Configure Ruff for VSCode

Avasam:
I recently noticed typeshed started using Ruff. This will ensure that Ruff' autofixes are run correctly and automatically by the IDE VSCode. As well as automatically recommending the extension for users.

I've been using these configurations for a few days now without issues. They are also the same configuration I use in my own projects that use Ruff all the way (except for `"ruff.organizeImports": false`, which is specific to typeshed relying on isort for its import sorting)

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -3,6 +3,7 @@
 {
     "recommendations": [
         "bierner.github-markdown-preview",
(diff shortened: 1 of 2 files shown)


AlexWaygood (approved):
I assume this is all correct, since you're basically the maintainer for these files :)

PR #10574, opened 2023-08-13
pywin32: Fix PyNETRESOURCE descriptor members wrongly marked as readonly

### LINKED ISSUE #10272: pywin32: PyNETRESOURCE attributes incorrectly typed as read-only
Writable attributes of `PyNETRESOURCE` objects are incorrectly typed as read-only. For example, consider the following script:

```py
import win32wnet

net_resource = win32wnet.NETRESOURCE()
net_resource.lpRemoteName = r"\\10.0.0.1\share"
win32wnet.WNetAddConnection2(net_resource, Password="pwd", UserName="username")
```

The code works when executed (if actual values are subsituted for the remote name, username, and password), but mypy raises the error `Property "lpRemoteName" defined in "PyNETRESOURCE" is read-only  [misc]`.

pywin32's documentation mentions that `lpLocalName`, `lpRemoteName`, `lpComment`, and `lpProvider` are writable attributes (see [`netresource.htm`](https://github.com/mhammond/pywin32/blob/f6300f84dad8a165f45a65d7823a554fa3cc4a5a/win32/Demos/win32wnet/netresource.htm#L53-L60)), and their own tests include examples similar to the one above (see [`test_win32wnet.py`](https://github.com/mhammond/pywin32/blob/f6300f84dad8a165f45a65d7823a554fa3cc4a5a/win32/test/test_win32wnet.py#L124-L127) and [`testwnet.py`](https://github.com/mhammond/pywin32/blob/f6300f84dad8a165f45a65d7823a554fa3cc4a5a/win32/Demos/win32wnet/testwnet.py#L101-L105)).

Versions I am running:

* pywin32: 306
* types-pywin32: 306.0.0.2
* mypy: 1.3.0
* Python: 3.11.4

Avasam:
Fixes #10272

If you wanna test it for yourself:
```py
>>> from win32.win32wnet import *
>>> type(NETRESOURCEType.lpLocalName)
<class 'member_descriptor'>
>>> type(NETRESOURCEType.dwScope)
<class 'member_descriptor'>
>>> test = NETRESOURCEType()
>>> test.lpLocalName = ""
>>> test.lpLocalName
''
>>> test.lpLocalName = None
>>> test.lpLocalName
>>> test.lpLocalName = 0
TypeError: Objects of type 'int' can not be converted to Unicode.
>>> test.lpLocalName = b""
TypeError: Objects of type 'bytes' can not be converted to Unicode.
>>> test.dwScope = 3
>>> test.dwScope
3
>>> test.dwScope = True
>>> test.dwScope
1
>>> test.dwScope = None
TypeError: an integer is required (got type NoneType)
```

The changes in `stubs/pywin32/win32/win32wnet.pyi` are not necessary, but I noticed these incorrect parameters whilst trying to get a `PyNETRESOURCE` (until I realized I could just instantiate it from its alias). Figured may as well include it.


AlexWaygood:
Wrong PR title? :)

Avasam:
> Wrong PR title? :)

Yep, autogenerated by GitHub because I accidentally included a commit that shouldn't be there (then rebased before opening the actual PR).

stubs/pywin32/_win32typing.pyi
```diff
@@ -1175,20 +1175,24 @@ class PyMSG: ...
...
+    @dwScope.setter
+    def dwScope(self, value: int | bool) -> None: ...
     @property
     def dwType(self) -> int: ...
+    @dwType.setter
+    def dwType(self, value: int | bool) -> None: ...
     @property
     def dwDisplayType(self) -> int: ...
+    @dwDisplayType.setter
+    def dwDisplayType(self, value: int | bool) -> None: ...
```
AlexWaygood, line 1187: These unions are redundant, since `bool` is a subclass of `int` (_but_, I understand that this can be surprising, so I'm open to leaving this as it is if `bool` is treated differently to `int` in some way):

```suggestion
    def dwScope(self, value: int) -> None: ...
    @property
    def dwType(self) -> int: ...
    @dwType.setter
    def dwType(self, value: int) -> None: ...
    @property
    def dwDisplayType(self) -> int: ...
    @dwDisplayType.setter
    def dwDisplayType(self, value: int) -> None: ...
```

stubs/pywin32/_win32typing.pyi
```diff
@@ -1175,20 +1175,24 @@ class PyMSG: ...
...
-    @property
-    def lpComment(self): ...
-    @property
-    def lpLocalName(self): ...
-    @property
-    def lpProvider(self): ...
-    @property
-    def lpRemoteName(self): ...
+    @dwUsage.setter
+    def dwUsage(self, value: int | bool) -> None: ...
```
AlexWaygood, line 1191:
```suggestion
    def dwUsage(self, value: int) -> None: ...
```

stubs/pywin32/_win32typing.pyi
Avasam: TIL. I knew about `int`/`float` (ie: they're special cased by type checkers because of the [numeric tower]), but I never realized `bool` actually subclassed `int` 🤯 

```py
>>> bool.__bases__
(<class 'int'>,)
```

AlexWaygood: Yup! In early versions of Python there was no `bool` class, only `int`; having `bool` subclass `int` meant it was fully backwards-compatible

AlexWaygood (approved):
Thanks!

Avasam:
Hey @petamas ,
This is the reason:
```py
>>> from win32.Demos.win32wnet.winnetwk import *
>>> import win32wnet
>>> handle = win32wnet.WNetOpenEnum(RESOURCE_REMEMBERED, RESOURCETYPE_DISK, 0, None)
>>> handle
<PyNetEnumHANDLE:2352947023664>
>>> handle.handle
2352947023664
>>> win32wnet.WNetEnumResource(handle)
[]
>>> win32wnet.WNetEnumResource(handle.handle) 
TypeError: argument 1 must be PyHANDLE, not int
```

Can you provide a code example where `win32.win32wnet.WNetEnumResource` accepts an `int` ? (with the specific version of pywin32)

Looking at the source code I can see that it seems to try (and I quote) to "Support integer objects for b/w compat". But I couldn't replicate it in actuality: https://github.com/mhammond/pywin32/blob/25e01da89dbec0da9180e33f554f4841c2599551/win32/src/win32wnet/win32wnet.cpp#L305-L312
https://github.com/mhammond/pywin32/blob/25e01da89dbec0da9180e33f554f4841c2599551/win32/src/PyHANDLE.cpp#L29-L35

petamas:
@Avasam 

Thanks for the quick answer! You are absolutely right, it indeed cannot take an `int`. I got confused because `win32wnet.pyi` describes `WNetOpenEnum()`'s return type as `int`, but in reality, as your example shows, it is `PyHANDLE`, so the error lies within `WNetOpenEnum()`'s type hints, not `WNetEnumResource()`. Sorry for the confusion!

Issue #10480, opened 2023-07-19
_typeshed.Sentinel


srittau:
While preparing #10479 I noticed that there is a problem with stubtest and sentinels. Currently, we either annotate a sentinel default value in a hacky way or we need to add the function to the stubtest allowlist, masking other potential problems. We could just add a non-type alias `Sentinel = object` to `_typeshed` and use that as the default value and teach stubtest about it. Alternatively,  we could add a `_typeshed.Sentinel` type alias (probably to `Never`) and teach stubtest not to check default values when the argument includes that type. I prefer the first option, because I think this reads better:

```python
def foo(x: int | None = Sentinel)
def foo(x: int | None | Sentinel = ...)
```

[PEP 661](https://peps.python.org/pep-0661/) would alleviate the need for this in the long term, but non-PEP-661 sentinels will still be around "forever".

sobolevn:
One more use-case: `argparse.BooleanOptionalAction` with its `deprecated_default` sentinel: https://github.com/python/cpython/blob/832c37d42a395d4ea45994daffa5e41bd74ac1bb/Lib/argparse.py#L887-L898

srittau:
Of course the first form would not work without changes to type checkers, since I don't think we can define `Sentinel` in a way that it's accepted as default value. But those changes would basically be similar to what PEP 661 proposes anyway. So maybe the more pragmatic way is to define `Sentinel: TypeAlias = Never` and use it as a type annotation. This should work without any changes to type checkers, but we can change stubtest to ignore the implementation's default value in that case.

PR #10576, opened 2023-08-14
Add _typeshed.Sentinel

srittau:
Closes: #10480

### DIFF
--- a/stdlib/_typeshed/__init__.pyi
+++ b/stdlib/_typeshed/__init__.pyi
@@ -8,7 +8,7 @@ from dataclasses import Field
 from os import PathLike
 from types import FrameType, TracebackType
 from typing import Any, AnyStr, ClassVar, Generic, Protocol, TypeVar, overload
-from typing_extensions import Buffer, Final, Literal, LiteralString, TypeAlias, final
+from typing_extensions import Buffer, Final, Literal, LiteralString, Never, TypeAlias, final
 
 _KT = TypeVar("_KT")
 _KT_co = TypeVar("_KT_co", covariant=True)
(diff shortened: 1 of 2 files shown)


srittau:
This PR has a bit of a hen and egg problem: It doesn't really make sense to start using `Sentinel`, until stubtest handles it, because it introduces more problems than it solves. If this gets accepted, I'd prepare a PR for stubtest, which will hopefully make it into the next mypy version.

srittau:
I'll prepare a PR this week. But basically stubtest would just skip checking the default value if it recognizes `Sentinel` in the argument types. This allows us to use "moral" defaults like `None` in the argparse example linked in #10480: https://github.com/python/cpython/blob/832c37d42a395d4ea45994daffa5e41bd74ac1bb/Lib/argparse.py#L887-L898

AlexWaygood:
I don't _love_ the idea of special-casing something typeshed-specific in stubtest, but I could live with it if the implementation isn't too hacky. I agree there's a problem to be solved here. (But @hauntsaninja will obviously have the final word when it comes to stubtest changes.)

I really wish we could have PEP-661; it would solve all of this in a much more elegant way. But it honestly seems unlikely to happen any time soon, if ever.

hauntsaninja:
For cases where the sentinel is not ever meant to be user passed, I agree that `= _typeshed.sentinel_value` is the best spelling. I think this wouldn't require type checker changes if we do `_typeshed.sentinel_value: Any`. I think it might not even need a stubtest change:
```
λ cat z.py
def f(x: int = object()): ...

λ cat z.pyi
from typing import Any
_sentinel: Any
def f(x: int = _sentinel): ...

λ python -m mypy.stubtest z
Success: no issues found in 1 module
```

srittau:
Huh, that's actually much neater. I'll open a separate PR to try this.

srittau:
Closing in favor of #10582.

PR #10577, opened 2023-08-14
Don't pretend `EnumMeta` inherits from `ABCMeta`

AlexWaygood:
Fixes https://github.com/python/mypy/issues/15870


AlexWaygood:
I'm running the full mypy test suite against this PR just to make sure it doesn't break any mypy tests:
- https://github.com/AlexWaygood/mypy/pull/10

AlexWaygood:
Doesn't look like this causes any issues for mypy

hauntsaninja (approved):
I'm not sure I understand the history here, but if alex_primer says things are green, this looks great to me!

PR #10582, opened 2023-08-15
Add _typeshed.sentinel

srittau:


srittau:
Closes: #10480, alternative to #10576

stdlib/argparse.pyi
```diff
@@ -334,7 +335,21 @@ class Action(_AttributeHolder):
...
+        def __init__(
+            self,
+            option_strings: Sequence[str],
+            dest: str,
+            default: _T | str | None = None,
+            type: Callable[[str], _T] | FileType | None = sentinel,  # noqa: Y011
+            choices: Iterable[_T] | None = sentinel,  # noqa: Y011
+            required: bool = False,
+            help: str | None = None,
+            metavar: str | tuple[str, ...] | None = sentinel,  # noqa: Y011
```
AlexWaygood, line 349: I suppose you'll be wanting to add special-casing for this in flake8-pyi? :)

srittau: Indeed.

tests/stubtest_allowlists/py312.txt
```diff
@@ -1,5 +1,4 @@
 # Uncategorised, from Python 3.12
-argparse.BooleanOptionalAction.__init__
```
srittau, line 2: This is a clear gain, in my opinion. Previously, all stubtest checks for this method were disabled, now it just works.

AlexWaygood (approved):
This looks like a pretty nice solution, thanks. I confirmed locally that mypy still errors out if you do something like `BooleanOptionalAction(['foo'], 'bar', 'baz', type=12345)` (i.e., mypy just ignores the default value when deciding which values are legal to be passed to the call; it only looks at the type annotation).

I tested using `mypy -c "from argparse import BooleanOptionalAction; b = BooleanOptionalAction(['foo'], 'bar', 'baz', type=12345)" --custom-typeshed-dir .`

hauntsaninja (approved):

PR #10583, opened 2023-08-15
Bump jsonschema to 4.19.*

srittau:
Closes: #10417


srittau:
Depends on typeshed-internal/stub_uploader#100.

srittau:
I *assume* (not investigated yet) that referencing  doesn't declare a dependency on python >= 3.8.

AlexWaygood:
Maybe we should print this to the terminal even if the `--verbose` flag isn't specified for `mypy_test.py`. It's pretty hard to figure out why things are failing in cases like this without it:

https://github.com/python/typeshed/blob/54193d5ddd4d01158173952d4ac73a5f76c05b71/tests/mypy_test.py#L299-L302

AlexWaygood:
Okay, so it's installing `referencing==0.8.9` in CI on Python 3.7 (_not_ the latest version of `referencing` -- the latest version is `referencing==0.30.2`, and the latest version does indeed declare that it only supports Python 3.8+).

AlexWaygood:
> I guess we need a new METADATA.toml field `minimum_python_version` or similar.

We should be able to merge https://github.com/python/typeshed/pull/10559 as soon as the PyPI admins let mypy publish a 1.5 patch release fixing the stubtest regression. (Or we could just allowlist the stubtest regression for now, and unallowlist it as soon as mypy 1.5.1 comes out.) Once https://github.com/python/typeshed/pull/10559 is merged, we stop running `mypy_test.py` on Python 3.7, so the problem sort-of becomes moot.

We're only in this situation in the first place because the `referencing` didn't dot all the I's in its packaging metadata for its early releases, so it feels _unlikely_ it'll come up _too_ much.

AlexWaygood:
We unfortunately can't pin to any version of `referencing` earlier than v0.8.9 (even if they're compatible with Python 3.7, which I haven't checked), since v0.8.9 is the first version of `referencing` that has a `py.typed` marker.

PR #10589, opened 2023-08-16
Changes waitid to return waitid_result | None

### LINKED ISSUE #10564: os.waitid can return None
Specifically, it can return None when its options argument contains os.WNOHANG.
>If [WNOHANG](https://docs.python.org/3/library/os.html#os.WNOHANG) is specified and there are no matching children in the requested state, None is returned.

I'm not sure there's any way to express that with overrides, given that it depends on the value of a specific argument. I'm also not sure of the ergonomic effect of making it return `os.waitid_result | None`.

gotyaoi:
Fixes #10564


AlexWaygood (approved):
Thanks!

PR #10590, opened 2023-08-16
SSL `wrap_socket` and `wrap_bio` accept `bytes` as well

CoolCat467:
SSLContext `wrap_socket` and `wrap_bio` functions accept `bytes` as well as strings.


AlexWaygood (approved):
Thanks!

Issue #10567, opened 2023-08-13
Using `ctypes.CField` for annotations of fields within `ctypes.Structure`-like


junkmd:
Currently, attributes of `ctypes.Structure`, such as `ctypes.wintypes.POINT`, are annotated with names and types that [specified in `_fields_`](https://github.com/python/cpython/blob/ee40b3e20d9b8d62a9b36b777dff42db1e9049d5/Lib/ctypes/wintypes.py#L115-L118).

https://github.com/python/typeshed/blob/df08fcec5f281ba01266869ec55368819aed0a77/stdlib/ctypes/wintypes.pyi#L126-L128

The problem with this approach is that at runtime, the attributes return types like `int` or `bytes`, while type checkers interpret them as returning `c_int` or `c_char`.

Below is the simplest example of a structure created for explanation purposes.
At runtime, the types of these fields are `ctypes.CField`. In the example below, `x` is a data descriptor that returns `int` as the getter and accepts both `c_int` and `int` as the setter.

```py
>>> import ctypes
>>>      
>>> class Foo(ctypes.Structure):
...     pass
... 
>>> Foo._fields_ = [('x', ctypes.c_int)]
>>> Foo.x
<Field type=c_long, ofs=0, size=4>
>>> type(Foo.x) 
<class '_ctypes.CField'>
>>> foo = Foo()
>>> foo
<__main__.Foo object at 0x0000023CDB80B8C0>
>>> foo.x
0
>>> foo.x = 3
>>> foo.x
3
>>> foo.x = ctypes.c_int(2)
>>> foo.x
2
```

Having stubs with types that deviate from the runtime is not an ideal situation.
Therefore, I propose modifying `CField` as shown below for use in annotating fields.

```py
_T = TypeVar("_T")
_CT = TypeVar("_CT", bound=_CData)


class CField(Generic[_T, _CT]):
    offset: int
    size: int
    @overload
    def __get__(self, instance: None, owner: type[Any]) -> Self: ...
    @overload
    def __get__(self, instance: Any, owner: type[Any] | None) -> _T: ...
    def __set__(self, instance: Any, value: _T | _CT) -> None: ...


class Foo(Structure):
    x: ClassVar[CField[int, c_int]]


# Foo._fields_ = [('x', c_int)]  # required in runtime


a = Foo.x  # CField[int, c_int]
foo = Foo()
b = foo.x  # int
foo.x = 3  # OK
foo.x = c_int(2)  # OK
[...]

PR #10595, opened 2023-08-17
`_CField` as a generic descriptor protocol

junkmd:
Please see #10567

### DIFF
--- a/stdlib/_ctypes.pyi
+++ b/stdlib/_ctypes.pyi
@@ -122,15 +122,23 @@ class CFuncPtr(_PointerLike, _CData):
 
     def __call__(self, *args: Any, **kwargs: Any) -> Any: ...
 
-class _CField:
+_GetT = TypeVar("_GetT")
+_SetT = TypeVar("_SetT")
+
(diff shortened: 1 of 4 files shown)


junkmd:
Should I consider adding workarounds for the error that raises when combining `stubs/pyserial/serial/win32.pyi` with Pyright?

For example, using something like `_CField[Any, Any, Any]`.

junkmd:
> Should I consider adding workarounds for the error that raises when combining `stubs/pyserial/serial/win32.pyi` with Pyright?
> 
> For example, using something like `_CField[Any, Any, Any]`.

I added workarounds.

srittau:
Thanks! I assume many of the `_CField[Any, Any, Any]` annotations could be improved in the future? In that case, please use `Incomplete` (from `_typeshed`) instead of `Any` to mark that fact.

junkmd:
@srittau 

It's good catch. Thank you.

> please use `Incomplete` (from `_typeshed`) instead of `Any` to mark

I did that.

srittau (approved):
My ctypes knowledge is limited, but this looks like an improvement to me and primer output also looks good.

PR #10597, opened 2023-08-20
fix(openpyxl): the return type of Workbook.active should be WorkSheet

iujakchu:
```python
import sys
import openpyxl
def get_prompts(path):
    wb = openpyxl.load_workbook(path)
    sheet = wb.active
    print(type(sheet))
    if sheet is None:
        sys.exit(-1)
    prompts = [cell.value for cell in sheet["C"] if cell.value] #__getitem__ method not defined on type "_WorkbookChild"
    wb.close()
    return prompts

prompts = get_prompts("txt/txt.xlsx")
print(prompts)
```
the printing shows me that the type of sheet variable is :
`<class 'openpyxl.worksheet.worksheet.Worksheet'>`
If I understand correctly,the active function should return a `WorkSheet` or `None` type.
This pr fixes it.

There is also another setter active function,I have no idea whether should I change it btw.


srittau:
Unfortunately, changing the return type isn't possible. Previously discussed in #9836 and #10306.

PR #10598, opened 2023-08-21
Add `usersettings` stub files

aitorres:
`usersettings` ([github](https://github.com/glvnst/usersettings), [pypi](https://pypi.org/project/usersettings/)) is a small Python library to streamline "portable local settings storage" on Python applications. The project itself was last updated in 2020 and, while working on a wide array of Python versions, is unlikely to get type annotations on the code itself. I'm adding their stubs to contribute to any developer that is using this library (like me!) and would like complete type checking support.

This is my first contribution to `typeshed`, I think I covered everything as it is a small library but I'm happy to make any suggested changes in case there's room for improvement and thank you so much!


stubs/usersettings/usersettings.pyi
```diff
@@ -0,0 +1,14 @@
+from typing import Any, TypeVar
+
+_S = TypeVar("_S")
+
+class Settings(dict[str, object]):
+    app_id: str | None
+    settings_directory: str | None
+    settings_file: str | None
+    def __init__(self, app_id: str) -> None: ...
+    def add_setting(self, setting_name: str, setting_type: type[_S] = ..., default: _S | None = ...) -> None: ...
```
srittau, line 10:
```suggestion
    def add_setting(self, setting_name: str, setting_type: type[_S] = ..., default: _S | None = None) -> None: ...
```

stubs/usersettings/usersettings.pyi
```diff
@@ -0,0 +1,14 @@
+from typing import Any, TypeVar
+
+_S = TypeVar("_S")
+
+class Settings(dict[str, object]):
```
srittau, line 5: I recommend to use `Any` instead of object here to match the return type of `__getattr__`. Otherwise code like the following will need an explicit cast or isinstance check:

```python
settings = Settings("")
settings.add_setting("x", int)
4 + settings.x  # type checks, as x is Any
4 + setting["x"]  # fails, as "x" is object
```

stubs/usersettings/usersettings.pyi
```diff
@@ -0,0 +1,14 @@
...
+    app_id: str | None
+    settings_directory: str | None
+    settings_file: str | None
+    def __init__(self, app_id: str) -> None: ...
+    def add_setting(self, setting_name: str, setting_type: type[_S] = ..., default: _S | None = ...) -> None: ...
+    def load_settings(self) -> None: ...
+    def save_settings(self) -> None: ...
+    def __getattr__(self, setting_name: str) -> Any: ...
+    def __setattr__(self, setting_name: str, setting_value: object) -> None: ...
```
srittau, line 14: More of an academic issue:

```suggestion
    def __setattr__(self, setting_name: str, setting_value: Any) -> None: ...
```

While at runtime every object is accepted, semantically it shouldn't be. E.g.:

```python
settings = Settings("")
settings.add_setting("x", int)
settings["x"] = ""  # works at runtime, but semantically wrong
```

Ideally, the type checker would complain about this, but there is no way we can express this. `Any` is a marker for that, while `object` really means "We accept everything here."

srittau (changes requested):
Thanks! Looks good overall, but a few remarks below.

aitorres:
Thank you so much for your comments, and for taking the time to explain them! 😄 

Pushed out a new commit to address these comments

srittau (approved):

Issue #10599, opened 2023-08-21
scripts/runtests.py cannot run tests/mypy_test.py


Screwtapello:
Steps to reproduce:

1. clone a fresh copy of the repo

        git clone https://github.com/python/typeshed.git
        cd typeshed

1. Set up a local development enviroment as described in `CONTRIBUTING.md`:

        python3 -m venv .venv
        .venv/bin/pip install -U pip
        .venv/bin/pip install -r requirements-tests.txt

1. Run the local tests as described in `tests/README.md`:

        .venv/bin/python3 scripts/runtests.py stdlib/xml

Expected results:

* All tests pass or are skipped

Actual results:

mypy tests fail:

```
Running mypy for Python 3.7...
usage: mypy_test.py [-h] [-x [EXCLUDE ...]] [-v] [-p [{3.12,3.11,3.10,3.9,3.8} ...]]
                    [--platform [{linux,win32,darwin} ...]]
                    [filter ...]
mypy_test.py: error: argument -p/--python-version: invalid choice: '3.7' (choose from '3.12', '3.11', '3.10', '3.9', '3.8')
```

It looks like #10559 removed "3.7" from the `tests/mypy_test.py` wrapper, so it would reject requests to test types in Python 3.7 style, but did not update `scripts/runtests.py` to stop asking for Python 3.7 style.

Adding "3.7" back to the list of acceptable Python versions in `tests/mypy_test.py` gets things working nicely for me; alternatively, updating `scripts/runtests.py` to ask for Python 3.8 support works too.

srittau:
Thanks for the report! Python 3.7 is only semi-supported in typeshed, since mypy dropped support. Even if we only officially drop support for 3.7 in January 2024 (#10113),  I will send a PR to use 3.8 in runtests to prevent any potential problems.

PR #10600, opened 2023-08-21
Use Python 3.8 in runtests.py

srittau:
We only support running mypy using 3.8+ up, since mypy 1.5 dropped support for running with Python 3.7.

Closes: #10599

### DIFF
--- a/scripts/runtests.py
+++ b/scripts/runtests.py
@@ -26,7 +26,7 @@ _NPX_ERROR_MESSAGE = colored("\nSkipping Pyright tests: npx is not installed or 
 _SUCCESS = colored("Success", "green")
 _SKIPPED = colored("Skipped", "yellow")
 _FAILED = colored("Failed", "red")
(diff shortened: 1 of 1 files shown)


scripts/runtests.py
```diff
@@ -28,7 +28,7 @@ def colored(text: str, color: str | None = None, **kwargs: Any) -> str:  # type:
 _FAILED = colored("Failed", "red")
 # We're using the oldest supported version because it's the most likely to produce errors
```
AlexWaygood, line 29: Very minor, but we could maybe tweak the comment slightly:

```suggestion
# We're using the oldest fully supported version because it's the most likely to produce errors
```

AlexWaygood (approved):
(I haven't manually verified that this fixes the issue, but I assume it does)

srittau:
I verified it manually. My tests still fail, but that's because of some stubtest inconsistencies vis-à-vis installed libraries.

PR #10602, opened 2023-08-21
openpyxl: Typed `from_tree` methods

Avasam:
Extracted from #9511 for ease of review and to reduce changes there. Imma need a proper review on this one.

Also seems like a good example use case for Protocol intersection. So I'll link https://github.com/python/typing/issues/213 here.


srittau (approved):
Thanks and sorry for the long review.

PR #10605, opened 2023-08-22
Minor improvements to `stubtest_third_party.py`

AlexWaygood:
**(1) Give a nicer error message if you try to run stubtest on non-existent stubs.**

Before:

```pytb
typeshed>python tests/stubtest_third_party.py fpdf
  File "C:\Users\alexw\coding\typeshed\tests\stubtest_third_party.py", line 263, in <module>
    main()
  File "C:\Users\alexw\coding\typeshed\tests\stubtest_third_party.py", line 256, in main
    if not run_stubtest(dist, verbose=args.verbose, specified_platforms_only=args.specified_platforms_only):
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\alexw\coding\typeshed\tests\stubtest_third_party.py", line 21, in run_stubtest
    metadata = read_metadata(dist_name)
               ^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\alexw\coding\typeshed\tests\parse_metadata.py", line 172, in read_metadata
    with Path("stubs", distribution, "METADATA.toml").open("rb") as f:
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\alexw\AppData\Local\Programs\Python\Python311\Lib\pathlib.py", line 1044, in open
    return io.open(self, mode, buffering, encoding, errors, newline)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
FileNotFoundError: [Errno 2] No such file or directory: 'stubs\\fpdf\\METADATA.toml'
```

With this PR:

```pytb
typeshed>python tests/stubtest_third_party.py fpdf
usage: stubtest_third_party.py [-h] [-v] [--num-shards NUM_SHARDS] [--shard-index SHARD_INDEX] [--specified-platforms-only] [DISTRIBUTION ...]
stubtest_third_party.py: error: Typeshed has no stubs for 'fpdf'!
```

**(2) Print the Python version to the terminal if there's an error, as well as the output of `pip freeze`.**

You can find out pretty easily from looking at the GitHub Actions workflow (or from the output of `actions/setup-python` in CI) which version of Python the job is running on. It's still easy to forget, though, that the CI might be running a different version of Python to the one you're running locally, and this change makes things more convenient by putting all the information you need to debug all together in one place.

### DIFF
--- a/tests/stubtest_third_party.py
+++ b/tests/stubtest_third_party.py
@@ -16,9 +16,14 @@ from parse_metadata import get_recursive_requirements, read_metadata
 from utils import colored, get_mypy_req, make_venv, print_error, print_success_msg
 
 
-def run_stubtest(dist: Path, *, verbose: bool = False, specified_platforms_only: bool = False) -> bool:
+def run_stubtest(
+    dist: Path, *, parser: argparse.ArgumentParser, verbose: bool = False, specified_platforms_only: bool = False
+) -> bool:
     dist_name = dist.name
-    metadata = read_metadata(dist_name)
+    try:
+        metadata = read_metadata(dist_name)
+    except ValueError as e:
+        parser.error(str(e))
     print(f"{dist_name}... ", end="")
 
     stubtest_settings = metadata.stubtest_settings
@@ -109,6 +114,10 @@ def run_stubtest(dist: Path, *, verbose: bool = False, specified_platforms_only:
             print_error("fail")
             print_commands(dist, pip_cmd, stubtest_cmd, mypypath)
             print_command_output(e)
+
+            print("Python version: ", file=sys.stderr)
+            ret = subprocess.run([sys.executable, "-VV"], capture_output=True)
+            print_command_output(ret)
 
             print("Ran with the following environment:", file=sys.stderr)
             ret = subprocess.run([pip_exe, "freeze", "--all"], capture_output=True)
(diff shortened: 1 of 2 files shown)


tests/stubtest_third_party.py
```diff
@@ -16,9 +16,14 @@
...
 
-def run_stubtest(dist: Path, *, verbose: bool = False, specified_platforms_only: bool = False) -> bool:
+def run_stubtest(
+    dist: Path, *, parser: argparse.ArgumentParser, verbose: bool = False, specified_platforms_only: bool = False
+) -> bool:
     dist_name = dist.name
-    metadata = read_metadata(dist_name)
+    try:
+        metadata = read_metadata(dist_name)
+    except ValueError as e:
```
srittau, line 25: Maybe it's better to use a custom exception class here. `ValueError` is a pretty common exception, so there is a risk that an error gets swallowed, especially when we expand `read_metadata()` at some point.

srittau (approved):
One optional suggestion, but LGTM.

tests/stubtest_third_party.py
AlexWaygood: Yeah, I wondered about this. I'll add a `NoSuchStubError` class.

PR #10606, opened 2023-08-23
stdlib/xml/sax: Add type annotations

Screwtapello:
This doesn't add complete type annotations, but it improves things for the most commonly used objects in the SAX API.


Screwtapello:
It looks like a bunch of CI tests failed because third-party-library stubs that subclass `xml.sax.handler.ContentHandler` and override `startElement()` don't actually use the correct annotations for the `attrs` parameter.

- the docs describe an interface called "Attributes", which is superset of the `Mapping[str, str]` interface, but no such class exists in the code (probably because the code pre-dates ABCs)
- the implementation always uses an instance of `AttributesImpl`, which implements the "Attributes" interface, but of course doesn't actually declare it anywhere (or mention `Mapping`).
- My type annotation says the `attrs` parameter will always be of type `AttributesImpl` (which is true) but of course that conflicts with everyone that just read the documentation and said "sure, `Mapping[str, str]` is close enough"

I'm not sure what clever type machinery exists to solve this (type aliases? ABCs? Protocols?) but I'm happy to try things if someone can provide guidance.

JelleZijlstra (approved):
Thanks, I think your solution makes sense.

PR #10610, opened 2023-08-24
Consistently use `Generic` as the last base class

sobolevn:
This is more like an experiment motivated by https://github.com/python/cpython/issues/106102

I want to see if it can change anything.


AlexWaygood:
It's nice to be consistent about this as a matter of style, whether or not it has any semantic difference. We could maybe add a rule to flake8-pyi.

AlexWaygood (approved):

stubs/tqdm/tqdm/tk.pyi
```diff
@@ -8,7 +8,7 @@ __all__ = ["tqdm_tk", "ttkrange", "tqdm", "trange"]
 
 _T = TypeVar("_T")
 
-class tqdm_tk(Generic[_T], std_tqdm[_T]):
+class tqdm_tk(std_tqdm[_T], Generic[_T]):
```
JelleZijlstra, line 11: Technically the Generic base is redundant here, as there is another base that uses the TypeVar. That's another thing where maybe we could be more consistent.

JelleZijlstra: (Though obviously no need to change that in this PR.)

sobolevn:
Can we merge this? :)

sobolevn:
Thanks everyone! 👏

AlexWaygood:
> Can we merge this? :)

You're a maintainer! 😄

PR #10614, opened 2023-08-25
Workalendar stubs implementation

treasuryspring:


stubs/workalendar/workalendar/core.pyi
```diff
@@ -0,0 +1,206 @@
...
+SUN: Incomplete
+ISO_MON: Incomplete
+ISO_TUE: Incomplete
+ISO_WED: Incomplete
+ISO_THU: Incomplete
+ISO_FRI: Incomplete
+ISO_SAT: Incomplete
+ISO_SUN: Incomplete
+
+class classproperty:
```
Akuli, line 19: Presumably this should be removed. In the stubs we could just use `ClassVar` for anything accessible through a class.

Akuli (commented):
This looks reasonable for a first version of the stubs, and can be improved later as needed :)

stubs/workalendar/workalendar/usa/wyoming.pyi
```diff
@@ -0,0 +1,4 @@
+from .core import UnitedStates
+
+class Wyoming(UnitedStates):
+    martin_luther_king_label: str
```
Akuli, line 4: Presumably most attributes should be e.g. `ClassVar[str]` instead of `str`, since the library does this:

```
@iso_register('US-WY')
class Wyoming(UnitedStates):
    """Wyoming"""
    martin_luther_king_label = "Martin Luther King, Jr. / Wyoming Equality Day"
```

This applies to a lot of other attributes, so I wouldn't change them all by hand. For example, you could use a script, a `sed` command, or a fancy editor.

treasuryspring: I pushed an improved version just now. thanks for reviewing

stubs/workalendar/workalendar/core.pyi
```diff
@@ -0,0 +1,207 @@
...
+    getter: Incomplete
+    __doc__: Incomplete
+    def __init__(self, getter) -> None: ...
+    def __get__(self, instance, owner): ...
```
Akuli, line 24:
```suggestion
```

Akuli (approved):
Didn't check everything in detail, but can be improved later if needed :)

PR #10615, opened 2023-08-25
The name param in asyncio.Task.__init__ is optional

davidhalter:
See also comment on https://github.com/python/typeshed/pull/10532#issuecomment-1693670617 how this was introduced.


sobolevn (approved):

sobolevn:
It is interesting that `stubtest` didn't catch this.

JelleZijlstra:
Probably because of this:

```
In [7]: inspect.signature(asyncio.Task.__init__)
Out[7]: <Signature (self, /, *args, **kwargs)>
```

PR #10622, opened 2023-08-27
Update `ibm-db` to `3.2.*`

sobolevn:
Closes https://github.com/python/typeshed/pull/10616

I've submitted https://github.com/ibmdb/python-ibmdb/pull/884 to fix the bug with the constant name.


AlexWaygood:
Thanks!

PR #10623, opened 2023-08-28
Update `genericpath` to 3.12

sobolevn:
Source: https://github.com/python/cpython/blob/3.12/Lib/genericpath.py#L51-L57


stdlib/genericpath.pyi
```diff
@@ -36,6 +39,9 @@ def getsize(filename: FileDescriptorOrPath) -> int: ...
 def isfile(path: FileDescriptorOrPath) -> bool: ...
 def isdir(s: FileDescriptorOrPath) -> bool: ...
 
+if sys.version_info >= (3, 12):
+    def islink(path: FileDescriptorOrPath) -> bool: ...
```
srittau, line 43: `os.lstat(path)` is annotated with `StrOrBytesPath`, not `FileDescriptorPath`. Why the difference?

sobolevn: My bad, I looked at `stat()`, not `lstat()` 👍 
Good catch!

srittau (approved):

PR #10626, opened 2023-08-29
`asyncio.wait()` no longer allows awaitables in 3.11

altendky:
https://docs.python.org/3.10/library/asyncio-task.html#asyncio.wait
> Run awaitable objects in the aws iterable concurrently and block until the condition specified by `return_when`.

https://docs.python.org/3.11/library/asyncio-task.html#asyncio.wait
> Run Future and Task instances in the aws iterable concurrently and block until the condition specified by `return_when`.

https://github.com/python/cpython/blob/v3.11.0/Lib/asyncio/tasks.py#L414-L415
```python
    if any(coroutines.iscoroutine(f) for f in fs):
        raise TypeError("Passing coroutines is forbidden, use tasks explicitly.")
```


AlexWaygood (approved):
Thanks!

PR #10612, opened 2023-08-24
Add _loop, _tasks, and _on_task_done to the stub for asyncio.TaskGroup

josephcourtney:
This adds annotations for asyncio.TaskGroup.{_loop, _tasks, and _on_task_done(...)}. 

I think I did everything right but I am not super comfortable with github yet.

fixes #10611

PR #10628, opened 2023-08-29
runtests.py: note that stdlib stubtest might fail

AlexWaygood:
This is an attempt to clear up some of the confusion in https://github.com/python/typeshed/pull/10612#discussion_r1306440772 (cc. @josephcourtney). `tests/README.md` already has a note to this effect in the section on `stubtest_stdlib.py`, so this just adds a note to the `scripts/runtests.py` docs

### DIFF
--- a/scripts/runtests.py
+++ b/scripts/runtests.py
@@ -55,7 +55,8 @@ def main() -> None:
         action="store_true",
         help=(
             "Run stubtest for the selected package(s). Running stubtest may download and execute arbitrary code from PyPI: "
-            "only use this option if you trust the package you are testing."
+            "only use this option if you trust the package you are testing. Running stubtest on the stdlib may fail "
+            "unless you're using exactly the same version of Python that we use in CI for stubtest."
         ),
     )
     parser.add_argument("path", help="Path of the stub to test in format <folder>/<stub>, from the root of the project.")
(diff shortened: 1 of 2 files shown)


scripts/runtests.py
```diff
@@ -55,7 +55,8 @@ def main() -> None:
         action="store_true",
         help=(
             "Run stubtest for the selected package(s). Running stubtest may download and execute arbitrary code from PyPI: "
-            "only use this option if you trust the package you are testing."
+            "only use this option if you trust the package you are testing. Running stubtest on the stdlib may fail "
+            "unless you're using exactly the same version of Python that we use in CI for stubtest."
```
AlexWaygood, line 59:
```suggestion
            "if you're not using exactly the same version of Python that we use in CI for stubtest."
```

srittau:
I'm not sure we should put that in the option help string. There are other reasons why it might fail, for example different installed libraries. (Especially Tcl/Tk and SSL.) All this is a bit too much for that string.

AlexWaygood:
> I'm not sure we should put that in the option help string. There are other reasons why it might fail, for example different installed libraries. (Especially Tcl/Tk and SSL.) All this is a bit too much for that string.

Yeah. Maybe it would be better to add a short note to `tests/README.md`. So far I've been resistant to documenting `runtests.py` too much in `tests/README.md`, as I don't like the idea of duplicating docs between the `--help` for the script and the separate MarkDown file. But maybe it's time to bite that bullet.

srittau (approved):

PR #10629, opened 2023-08-29
create_baseline_stubs.py: Improve pyright config file editing

Akuli:
There was an indexing bug that added a trailing comma to the wrong line. It happened every time someone added stubs for a new package whose name is alphabetically at the end of the list, and [it looked like this](https://github.com/python/typeshed/pull/9874#discussion_r1134087111). We don't really need indexing spaghetti to deal with commas, since pyright's "JSON" config file supports trailing commas anyway (even though plain old JSON does not).

Tested locally by deleting stubs and adding them back.

### DIFF
--- a/scripts/create_baseline_stubs.py
+++ b/scripts/create_baseline_stubs.py
@@ -161,8 +161,8 @@ def add_pyright_exclusion(stub_dir: str) -> None:
     start = i
 
     before_third_party_excludes = lines[:start]
-    third_party_excludes = lines[start : end]
-    after_third_party_excludes = lines[end :]
+    third_party_excludes = lines[start:end]
+    after_third_party_excludes = lines[end:]
 
     last_line = third_party_excludes[-1].rstrip()
     if not last_line.endswith(","):


AlexWaygood (commented):
This relies on us remembering to keep a trailing comma for the last element in the list, right? We could edit it manually (which we do from time to time) and forget to keep a trailing comma with the last element, and none of our tests would fail. Then if somebody came along and tried to add stubs for `zzzzz`, the script would be broken.

We could possibly add a test to `check_consistent.py` that asserts that the last element in the list has a trailing comma. But that seems like it could be even more complicated than what we have now

AlexWaygood:
> Even though I still think the code looks like spaghetti

I mean, I don't disagree :)

Akuli:
I think I managed to make it look less spaghetti-like, though it is longer than before. I again tested it locally. It also sorts the list every time, and turns out there were a few places where it wasn't sorted correctly. The only downside I can see is that the code is a bit longer than before.

If we agree that this is a good idea, let's change the title of the PR before we merge.

scripts/create_baseline_stubs.py
```diff
@@ -152,22 +152,38 @@ def add_pyright_exclusion(stub_dir: str) -> None:
...
     while not lines[i].strip().startswith("]"):
         i += 1
-    # Must use forward slash in the .json file
-    line_to_add = f'        "{stub_dir}",'.replace("\\", "/")
-    initial = i - 1
-    while lines[i].lower() > line_to_add.lower():
+    end = i
+
+    # We assume that all third-party excludes must be at the end of the list.
+    # This helps with skipping special entries, such as "stubs/**/@tests/test_cases".
```
AlexWaygood, line 158: Ideally we'd assert this somehow in `check_consistent.py`, but I can't think of a simple way of doing so, and a complex way of doing so would probably be overkill for this use case. If this assumption becomes invalidated at some point, we can probably just deal with that when the problem arises.

AlexWaygood (approved):
This is a big readability improvement. Thanks!

PR #10630, opened 2023-08-29
Fix psycopg2

hamdanal:
Unblocks https://github.com/sbdchd/django-types/issues/74

I did some careful reading of the docs and the C code and some testing in the repl. This hopefully yields a much better quality stubs for psycopg2.


stubs/psycopg2/psycopg2/extras.pyi
```diff
@@ -16,6 +16,9 @@ from psycopg2._psycopg import (
     ReplicationConnection as _replicationConnection,
     ReplicationCursor as _replicationCursor,
     ReplicationMessage as ReplicationMessage,
+    connection as _connection,
+    cursor as _cursor,
+    quote_ident as quote_ident,
```
hamdanal, line 21: These are moved here because some cyclic imports were making pytype angry

stubs/psycopg2/psycopg2/_psycopg.pyi
```diff
@@ -1,109 +1,147 @@
...
-from typing_extensions import Literal, Self, TypeAlias
+from typing import Any, NoReturn, Protocol, SupportsInt, TypeVar, overload, type_check_only
+from typing_extensions import Literal, Self, SupportsIndex, TypeAlias
 
-import psycopg2
-import psycopg2.extensions
 from psycopg2.sql import Composable
 
 _Vars: TypeAlias = Sequence[Any] | Mapping[str, Any] | None
+_AcceptedByInt: TypeAlias = str | ReadableBuffer | SupportsInt | SupportsIndex | SupportsTrunc
```
hamdanal, line 11: I saw this alias used in many places in typeshed. Would it make sense to define it as a private attribute in `_typeshed`?

stubs/psycopg2/psycopg2/_psycopg.pyi
```diff
@@ -478,24 +530,50 @@ class lobject:
...
-def DateFromTicks(ticks): ...
-def IntervalFromPy(*args, **kwargs): ...
-def Time(hour, minutes, seconds, tzinfo=...): ...
-def TimeFromPy(*args, **kwargs): ...
-def TimeFromTicks(ticks): ...
-def Timestamp(year, month, day, hour, minutes, seconds, tzinfo=...): ...
-def TimestampFromPy(*args, **kwargs): ...
-def TimestampFromTicks(ticks): ...
+@type_check_only
+class _datetime:
```
hamdanal, line 534: This and other classes that implement the `ISQLQuote` protocol could be made generic and their read-only `adapted` attribute would have the type variable as a type, shall someone ask for this later, but for now this is good enough.

stubs/psycopg2/psycopg2/__init__.pyi
```diff
@@ -27,19 +27,28 @@ from psycopg2._psycopg import (
...
+    **kwargs: Any,
+) -> _T_conn: ...
 @overload
 def connect(
-    dsn: str | None = None, *, connection_factory: Callable[..., _T_conn], cursor_factory: None = None, **kwargs: Any
+    dsn: str | None = None,
+    *,
+    connection_factory: Callable[..., _T_conn],
+    cursor_factory: Callable[..., cursor] | None = None,
+    **kwargs: Any,
```
srittau, line 51: These two overloads look identical to me now.

stubs/psycopg2/psycopg2/_psycopg.pyi
srittau: Sure, a PR would be welcome, although the psycopg2 can only sensibly use it after most type checkers release a version that includes the new alias.

stubs/psycopg2/psycopg2/_psycopg.pyi
```diff
@@ -1,109 +1,147 @@
...
-TIMEARRAY: Any
-UNICODE: Any
-UNICODEARRAY: Any
-UNKNOWN: Any
-adapters: dict[Any, Any]
+
+class _ISQLQuoteProto(Protocol):
+    # Objects conforming this protocol should implement a getquoted() and optionally a prepare() method.
+    # The real ISQLQuote class is implemented below with more stuff.
+    def getquoted(self) -> bytes: ...
```
srittau, line 72: Nit: Could you add the commented out signature of `prepare()` as well to serve as documentation or for whenever we get optional protocol members? (python/typing#601)

stubs/psycopg2/psycopg2/_psycopg.pyi
srittau: Maybe add this as a `# TODO` comment?

stubs/psycopg2/psycopg2/_range.pyi
```diff
@@ -1,62 +1,72 @@
...
+
+from psycopg2._psycopg import connection, cursor
 
 class Range:
     def __init__(
         self, lower: Incomplete | None = None, upper: Incomplete | None = None, bounds: str = "[)", empty: bool = False
     ) -> None: ...
     @property
-    def lower(self): ...
+    def lower(self) -> Incomplete: ...
```
srittau, line 12: Nit:

```suggestion
    def lower(self): ...
```

stubs/psycopg2/psycopg2/_range.pyi
```diff
@@ -1,62 +1,72 @@
...
     @property
-    def upper(self): ...
+    def upper(self) -> Incomplete: ...
```
srittau, line 14:
```suggestion
    def upper(self): ...
```

srittau (changes requested):
Thanks and I'm sorry for the late review. Looks good to me, a few nits below.

stubs/psycopg2/psycopg2/__init__.pyi
hamdanal: Mmm. I was deleting the first overload then remembered why I kept both.
In the first, the function can be called with positional arguments while in the second only `dsn` can be positional (because we need to give it a default value and can't give a default to `connection_factory`) so this test
`assert_type(psycopg2.connect("test-conn", MyConnection), MyConnection)` fails. I added this test case so this doesn't regress.

hamdanal:
Running stubtest for psycopg2 locally fails on my machine because it tries to build it from source and apparently it is missing some dependencies. psycopg2 is special because the compiled wheels exist on PyPI as the `psycopg2-binary` package. I did this locally to get around this issue:

```diff
diff --git a/tests/stubtest_third_party.py b/tests/stubtest_third_party.py
index af393c35f..65241e7be 100755
--- a/tests/stubtest_third_party.py
+++ b/tests/stubtest_third_party.py
@@ -45,7 +45,10 @@ def run_stubtest(
             print_error("fail")
             raise
         dist_extras = ", ".join(stubtest_settings.extras)
-        dist_req = f"{dist_name}[{dist_extras}]=={metadata.version}"
+        pypi_dist_name = dist_name
+        if dist_name == "psycopg2":
+            pypi_dist_name = "psycopg2-binary"  # psycopg2 is named psycopg2-binary on PyPI
+        dist_req = f"{pypi_dist_name}[{dist_extras}]=={metadata.version}"

         # If tool.stubtest.stubtest_requirements exists, run "pip install" on it.
         if stubtest_settings.stubtest_requirements:
``` 
I don't know how many packages you have that are in the same situation but installing from wheels would make CI faster and cause less problems while developing. It might not be worth the effort though, I don't know.

stubs/psycopg2/psycopg2/__init__.pyi
srittau: Good point, I missed that.

srittau (approved):

stubs/psycopg2/psycopg2/sql.pyi
```diff
@@ -1,50 +1,50 @@
...
 class SQL(Composable):
-    def __init__(self, string) -> None: ...
+    def __init__(self, string: str) -> None: ...
     @property
     def string(self) -> str: ...
-    def as_string(self, context) -> str: ...
-    def format(self, *args, **kwargs) -> Composed: ...
-    def join(self, seq) -> Composed: ...
+    def format(self, *args: Composable, **kwargs: Composable) -> Composed: ...
+    def join(self, seq: SupportsIter[Composable]) -> Composed: ...
```
andersk, line 30: This needs to be `Iterable[Composable]`, not `SupportsIter[Composable]`. The difference is that

* `seq: Iterable[Composable]` means `iter(seq): Iterator[Composable]` (hence `next(iter(seq)): Composable`), while
* `seq: SupportsIter[Composable]` means `iter(seq): Composable`, which is not an iterator.

PR #10635, opened 2023-08-30
Update `configparser` to 3.12

sobolevn:
Source commit https://github.com/python/cpython/commit/296e4efebbd1865153ed3be24887a2af3898a0c4


stdlib/configparser.pyi
```diff
@@ -5,31 +5,53 @@ from re import Pattern
...
-    "ExtendedInterpolation",
-    "LegacyInterpolation",
-    "SectionProxy",
-    "ConverterMapping",
-    "DEFAULTSECT",
-    "MAX_INTERPOLATION_DEPTH",
-]
-
-if sys.version_info < (3, 12):
-    __all__ += ["SafeConfigParser"]
```
srittau, line 32: What was wrong with this syntax? This looks far more readable to me.

sobolevn: `__all__` was changed in 3.12 from `list` to `tuple`, I reflected that

AlexWaygood: Yeah, unfortunately stubtest will (reasonably, if you squint at it the right way) complain if `__all__` is a tuple at runtime but a list in the stub

srittau: Ah, I missed that.

srittau (approved):

PR #10636, opened 2023-08-30
Add `durations` to `unittest` in 3.12

sobolevn:
Source commit: https://github.com/python/cpython/commit/6883007a86bdf0d7cf4560b949fd5e577dab1013


stdlib/unittest/result.pyi
```diff
@@ -1,9 +1,12 @@
+import sys
 import unittest.case
 from _typeshed import OptExcInfo
 from collections.abc import Callable
 from typing import Any, TextIO, TypeVar
+from typing_extensions import TypeAlias
 
 _F = TypeVar("_F", bound=Callable[..., Any])
+_DurationsType: TypeAlias = list[tuple[unittest.case.TestCase, float]]
```
srittau, line 9: At least `collectedDurations` (ugh, camel case...) uses `list[tuple[str, float]]`:

https://github.com/python/cpython/blob/24b9bdd6eaf0b04667b6cd4c8154f7c7c10c2398/Lib/unittest/result.py#L170

srittau: Although the docs are wrong: https://docs.python.org/3.12/library/unittest.html?highlight=collecteddurations#unittest.TestResult.collectedDurations

But even the tests use `str`: https://github.com/python/cpython/blob/24b9bdd6eaf0b04667b6cd4c8154f7c7c10c2398/Lib/test/test_unittest/test_runner.py#L1402

I will send a PR to CPython.

sobolevn: Yeap, I used docs for this type. Let's wait for your PR to land first.

srittau (approved):
Thanks!

stdlib/unittest/result.pyi
srittau: Looks like a deliberate change to fix a circular reference, they just forgot to update the docs: https://github.com/python/cpython/pull/106888

stdlib/unittest/runner.pyi
```diff
@@ -14,23 +15,55 @@ class TextTestResult(unittest.result.TestResult):
...
-        warnings: type[Warning] | None = None,
-        *,
-        tb_locals: bool = False,
-    ) -> None: ...
+    stream: TextIO
+    descriptions: bool
+    verbosity: int
+    failfast: bool
+    buffer: bool
+    warnings: type[Warning] | None
```
AlexWaygood, line 36: `self.warnings` is passed as the first argument of `warnings.simplefilter` here:

https://github.com/python/cpython/blob/6c484c39beeb66d40ef0a73cc4f1e900ea498cfa/Lib/unittest/runner.py#L232-L234

That implies that `warnings` should be `warnings._ActionKind | None`:

https://github.com/python/typeshed/blob/8888ac0d6dcb2d20b1621550cb64b69eba289f11/stdlib/warnings.pyi#L38

(The same goes for the annotation of the parameter in the `__init__` method -- the current type looks incorrect to me)

sobolevn: I went with `warnings: str | None`, since it is has this type in `unittest/main.pyi` as well.

AlexWaygood: OK

AlexWaygood (approved):
Thanks!

PR #10638, opened 2023-08-30
Add stubs for libsass

Daverball:
This is a small single file module  which for the most part just contains the one function `compile` in the common use-case, but I stubbed out the rest of the public API and the utility package `sassutils` too, since there really wasn't all that much.

I used the docs as a first reference, since they contain some type hints and then verified it with the actual implementation.

The docs are available here: https://sass.github.io/libsass-python/index.html
Source code here: https://github.com/sass/libsass-python/blob/0.22.0/sass.py


srittau (approved):
A cursory inspection looks good!

PR #10642, opened 2023-08-31
Add `isjunction` and `splitroot` functions

sobolevn:
Source of `splitroot`:
- https://github.com/python/cpython/blob/059bd4d299b384d2b434ccb106f91cb4bb03bbb1/Lib/ntpath.py#L179-L228
- https://github.com/python/cpython/blob/059bd4d299b384d2b434ccb106f91cb4bb03bbb1/Lib/posixpath.py#L138-L164

Source of `isjunction`:
- https://github.com/python/cpython/blob/059bd4d299b384d2b434ccb106f91cb4bb03bbb1/Lib/ntpath.py#L282-L294
- https://github.com/python/cpython/blob/059bd4d299b384d2b434ccb106f91cb4bb03bbb1/Lib/posixpath.py#L192-L196


AlexWaygood (approved):
Thanks!

PR #10645, opened 2023-09-01
sqlite3: add Connection arg to cursorClass

ehandal:
Cursor instantiation expects a Connection object as an argument, which needs to be reflected in cursorClass factory argument of the Connection.cursor method.


srittau (approved):
Thanks!

PR #10646, opened 2023-09-01
Update `pathlib` to 3.12

sobolevn:
Source: https://github.com/python/cpython/blame/3.12/Lib/pathlib.py


stdlib/pathlib.pyi
```diff
@@ -38,7 +39,11 @@ class PurePath(PathLike[str]):
     def suffixes(self) -> list[str]: ...
     @property
     def stem(self) -> str: ...
-    def __new__(cls, *args: StrPath) -> Self: ...
+    if sys.version_info >= (3, 12):
+        def __new__(cls, *args: StrPath, **kwargs: Unused) -> Self: ...
```
sobolevn, line 43: Technically `*args` are also unused in `__new__`, but I fear that using `Unused` here might break `*args: StrPath` in `__init__`. So, I went with the safest bet.

stdlib/pathlib.pyi
```diff
@@ -53,15 +59,21 @@ class PurePath(PathLike[str]):
     def as_uri(self) -> str: ...
     def is_absolute(self) -> bool: ...
     def is_reserved(self) -> bool: ...
-    if sys.version_info >= (3, 9):
+    if sys.version_info >= (3, 12):
+        def is_relative_to(self, other: StrPath, /, *_deprecated: StrPath) -> bool: ...
```
srittau, line 63: We can't use PEP 570 syntax yet, even if it didn't break our tests, since we still support Python 3.7.

```suggestion
        def is_relative_to(self, other: StrPath, *__deprecated: StrPath) -> bool: ...
```

(Might need a stubtest allowlist entry.)

stdlib/pathlib.pyi
```diff
@@ -53,15 +59,21 @@ class PurePath(PathLike[str]):
...
         def is_relative_to(self, *other: StrPath) -> bool: ...
 
     if sys.version_info >= (3, 12):
         def match(self, path_pattern: str, *, case_sensitive: bool | None = None) -> bool: ...
     else:
         def match(self, path_pattern: str) -> bool: ...
 
-    def relative_to(self, *other: StrPath) -> Self: ...
+    if sys.version_info >= (3, 12):
+        def relative_to(self, other: StrPath, /, *_deprecated: StrPath, walk_up: bool = False) -> Self: ...
```
srittau, line 73:
```suggestion
        def relative_to(self, other: StrPath, *__deprecated: StrPath, walk_up: bool = False) -> Self: ...
```

srittau (changes requested):

stdlib/pathlib.pyi
AlexWaygood: The argument before the slash needs to have the `__` prefix, not the one after

srittau: I don't know what you mean. 😇

sobolevn:
Sunset of 3.7 is very awaited! 🌆

srittau (approved):

PR #10647, opened 2023-09-01
Update `zipfile` to 3.12

sobolevn:
Source commit: https://github.com/python/cpython/commit/36854bbb240e417c0df6f0014924fcc899388186#diff-d93d29530934ea6d5aef4319d8abf55f36c61626f8229b87d71588b69ed18e23


stdlib/zipfile.pyi
```diff
@@ -274,6 +274,12 @@ if sys.version_info >= (3, 8):
             def joinpath(self, *other: StrPath) -> Path: ...
         else:
             def joinpath(self, add: StrPath) -> Path: ...  # undocumented
+        if sys.version_info >= (3, 12):
+            def glob(self, pattern: str) -> map[Self]: ...
+            def rglob(self, pattern: str) -> map[Self]: ...
+            def is_symlink(self) -> Literal[False]: ...
+            def relative_to(self, other: Self, *extra: StrPath) -> str: ...
+            def match(self, path_pattern: str) -> bool: ...
```
AlexWaygood, line 282: There are also new `__eq__` and `__hash__` methods. `__eq__` methods are important for mypy's `--strict-equality` check; pyright uses `__hash__` methods to determine whether objects can be stored as dictionary keys or not.

```suggestion
            def match(self, path_pattern: str) -> bool: ...
            def __eq__(self, other: object) -> bool: ...
            def __hash__(self) -> int: ...
```

stdlib/zipfile.pyi
```diff
@@ -274,6 +274,12 @@ if sys.version_info >= (3, 8):
             def joinpath(self, *other: StrPath) -> Path: ...
         else:
             def joinpath(self, add: StrPath) -> Path: ...  # undocumented
+        if sys.version_info >= (3, 12):
+            def glob(self, pattern: str) -> map[Self]: ...
+            def rglob(self, pattern: str) -> map[Self]: ...
```
AlexWaygood, line 279: These look to me like they return generator expressions, not instances of `map`: https://github.com/python/cpython/commit/36854bbb240e417c0df6f0014924fcc899388186#diff-06a1fd516348dcbb784169d31d76c5171caedf093fb423ce2d08a4442452d013R359-R371

```suggestion
            def glob(self, pattern: str) -> Iterator[Self]: ...
            def rglob(self, pattern: str) -> Iterator[Self]: ...
```

Also, it looks like our annotation for `iterdir()` in this file is wrong: it should also return `Iterator[Self]`, but currently returns `Iterator[Path]`. Would you like to also correct that in this PR?

stdlib/zipfile.pyi
```diff
@@ -274,6 +274,12 @@ if sys.version_info >= (3, 8):
...
+            def is_symlink(self) -> Literal[False]: ...
+            def relative_to(self, other: Self, *extra: StrPath) -> str: ...
```
AlexWaygood, line 281: Does `other` _have_ to be an instance of exactly the same class as whatever `self` is? It looks to me like any `zipfile.Path` object will do:

```suggestion
            def relative_to(self, other: Path, *extra: StrPath) -> str: ...
```

stdlib/zipfile.pyi
sobolevn: This was the source of my `map` proposal. I still think that current runtime behaves incorrectly, but I don't like to persuade people, it literally drains life out of me :)

AlexWaygood: > but I don't like to persuade people, it literally drains life out of me :)

I can empathise with that. I'm sorry that I disagreed with your proposal this time :)

sobolevn: No worries, I find motivation in my own projects :)

AlexWaygood: One of the reasons I created typeshed-stats was so I could have a project nobody could argue with me about 😄

AlexWaygood (approved):
Thanks!

PR #10648, opened 2023-09-01
Add a case where zip() can be called with no arguments

plokmijnuhby:


Daverball:
Could you explain what your use-case for this is? Either way binding to `Never`/`NoReturn` seems wrong since both calling `zip()` and iterating over the result of that results in no runtime errors since you always will end up with zero iterations, so there's no opportunity for anything to go wrong, you can unpack into any shape and it will work because it will never need to unpack anything.

So returning `zip[Any]` or `zip[tuple[Any, ...]` seems more appropriate to me since it's type safe to use it that way, unless I'm missing something.

Daverball:
I've created a little mypy-play to see what a generic iterator that's bound to `NoReturn` would behave like and it doesn't seem like you get any real benefits out of it, in fact it makes it impossible for mypy to infer the type of the loop variable and since it's not bound to `Any` it will complain about it. Beyond that the only minor benefit you seem to get is that code below the explicit call to `__next__()` will be considered unreachable, but how often are you really gonna do that instead of using the `next` builtin function, which doesn't detect that case.

https://mypy-play.net/?mypy=latest&python=3.11&flags=warn-unreachable&gist=e81329055ba383508ecba7dc1146330d

plokmijnuhby:
Ah, then maybe I'm wrong about the correct way of doing this. I think I misunderstood how mypy works a little bit.

plokmijnuhby:
> `python/pyspark/pandas/typedef/typehints.py:779: error: Unused "type: ignore[has-type, misc]" comment  [unused-ignore]`

I found the code in question, and it's equivalent to this:
```
def func(params:zip):
    if isinstance(params, zip):
        tuple(x for x,y in params)  # previously this caused errors
```
The isinstance check is necessary here and seemingly causes mypy to get really confused about the type of params. This is probably an unrelated bug in mypy.

Daverball:
That's not a bug, because `zip` is generic an isinstance check is not enough to type narrow to a specific bound type. mypy will default to the first overload when that happens, previously this contained an unbound type var and would result in checking against `zip[tuple[_T1]`, which cannot be unpacked into two elements and neither of those two elements can be inferred due to it, so mypy will also complain that those two variables have no types.

So in some sense the new behavior can be seen as an improvement, because it will be more lenient in type narrowing contexts where the type vars cannot be bound. On the other hand it could  be seen as worse, since it will hide some type safety issues, but since there is no good way to make it type safe, without doing an expensive shape check for every iteration, that's probably fine.

plokmijnuhby:
Ah I see. It seems like how it is now is the most sensible way to do it then.
This is going beyond the scope of this PR, but I don't really understand why  `isinstance` checks on a variable trigger a stricter check than just annotating the same variable with a type. But still, if that's how mypy does it, that's how mypy does it.

Daverball:
Yeah it is a somewhat weird behavior that can lead to some issues, the type narrowing is perhaps a bit too aggressive for `isinstance` on generics with overloaded `__init__`/`__new__`. Although once type variable defaults are a thing that's maybe less of an issue. Cases such as these are rather rare, so it doesn't come up that often.

Once `TypeVarTuple` is more of a first class citizen and comes with `Map` to map the tuple elements classes like `zip` and `map` can be annotated more accurately.

PR #10649, opened 2023-09-02
Update `importlib.resources` to 3.12

sobolevn:
Source commit: https://github.com/python/cpython/commit/447d061bc7b978afedd3b0148715d2153ac726c5#diff-9181c6cb24e7bc2b5ff7451ec45cdd6223c021cac852b7c369e80ee7344de628


sobolevn:
Ok, since `stubtest` is not satisfied when I define signature "as it really is", I am going with "how it should be".

hauntsaninja (approved):

PR #10650, opened 2023-09-02
Update `unittest.mock` to 3.12

sobolevn:
Source: https://github.com/python/cpython/blob/4dc07d2bde2e219f3b6995891b8cda3e8094804a/Lib/unittest/mock.py#L420

### DIFF
--- a/stdlib/unittest/mock.pyi
+++ b/stdlib/unittest/mock.pyi
@@ -106,7 +106,24 @@ class Base:
 # We subclass with "Any" because mocks are explicitly designed to stand in for other types,
 # something that can't be expressed with our static type system.
 class NonCallableMock(Base, Any):
-    def __new__(__cls, *args: Any, **kw: Any) -> Self: ...
+    if sys.version_info >= (3, 12):
+        def __new__(
+            cls,
+            spec: list[str] | object | type[object] | None = None,
+            wraps: Any | None = None,
+            name: str | None = None,
+            spec_set: list[str] | object | type[object] | None = None,
+            parent: NonCallableMock | None = None,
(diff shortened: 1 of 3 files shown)


stdlib/unittest/mock.pyi
```diff
@@ -106,7 +106,25 @@ class Base:
 # We subclass with "Any" because mocks are explicitly designed to stand in for other types,
 # something that can't be expressed with our static type system.
 class NonCallableMock(Base, Any):
-    def __new__(__cls, *args: Any, **kw: Any) -> Self: ...
+    if sys.version_info >= (3, 12):
+        def __new__(
+            cls,
+            spec: list[str] | object | type[object] | None = None,
+            wraps: Any | None = None,
```
AlexWaygood, line 113: We use `Incomplete` for our stubs for the backport package, but we should probably have the same annotation in both places:

https://github.com/python/typeshed/blob/454438e4db1e259900f3f1aa602a39ef3ac23af2/stubs/mock/mock/mock.pyi#L92

Which is correct here: `Any` or `Incomplete`?

sobolevn: It should be `Any`: https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock

(You can grep CPython's tests)

AlexWaygood: Yes, `Any` looks correct here. Would you like to change the annotations in the third-party backport to `Any`? It would be worth doing, but it doesn't feel like it deserves its own PR -- we might as well do it here :)

AlexWaygood (approved):
Thanks!

PR #10652, opened 2023-09-02
Experiment with zip typing

### LINKED ISSUE #10651: zip vararg
mypy 0.950 (compiled: yes) and Python 3.10.4:

```python
z1 = list(zip((1, 10), (2, 20), (3, 30)))  # correct: list[Tuple[int, int, int]]
z2 = list(zip(*[(1, 10), (2, 20), (3, 30)]))  # list[tuple[Any, ...]] but I expect list[tuple[int, int, int]]
```
Any chance to get mypy infer the type of `zip(*l)`?

Thanks for reading!

For the story it comes from here https://github.com/joelgrus/data-science-from-scratch/blob/d5d0f117f41b3ccab3b07f1ee1fa21cfcf69afa1/first-edition/code-python3/machine_learning.py#L18 !

hauntsaninja:
It's been a couple years, maybe I can change zip without it being disastrous. This change makes zip type safe, which is probably undesirable. Would also fix #10651. Note this isn't a join vs union thing, both unions and joins are often undesirable for `zip(*x)` use cases.

### DIFF
--- a/stdlib/builtins.pyi
+++ b/stdlib/builtins.pyi
@@ -1848,17 +1848,7 @@ class zip(Iterator[_T_co], Generic[_T_co]):
             strict: bool = ...,
         ) -> zip[tuple[_T1, _T2, _T3, _T4, _T5]]: ...
         @overload
-        def __new__(
-            cls,
-            __iter1: Iterable[Any],
-            __iter2: Iterable[Any],
-            __iter3: Iterable[Any],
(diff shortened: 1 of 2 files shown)


srittau:
Looking at the primer output, this looks to be too disruptive for the moment, although there are also some "# type: ignore" removals in there that would merit closer inspection.

PR #10654, opened 2023-09-03
Update `importlib.metadata` to 3.12

sobolevn:
Source: https://github.com/python/cpython/blob/3.12/Lib/importlib/metadata/__init__.py


AlexWaygood (approved):
Thanks!

PR #10658, opened 2023-09-04
Add stubs for requests-oauthlib

Daverball:
These are mostly still baseline stubs since the same is true for `types-oauthlib`, however some simple things have been filled in/copied over from `types-requests`.

Source can be found at: https://github.com/requests/requests-oauthlib/blob/v1.3.1/requests_oauthlib/__init__.py
Docs here: https://requests-oauthlib.readthedocs.io/en/v1.3.1/

### DIFF
--- a/pyrightconfig.stricter.json
+++ b/pyrightconfig.stricter.json
@@ -68,6 +68,7 @@
         "stubs/qrcode",
         "stubs/redis",
         "stubs/requests",
+        "stubs/requests-oauthlib",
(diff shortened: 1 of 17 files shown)


Daverball:
Looks like GitHub Actions is having some issues today...

srittau:
> Looks like GitHub Actions is having some issues today...

When isn't it? I'll restart CI.

srittau (approved):
LGTM from a cursory inspection.

PR #10660, opened 2023-09-04
Add a lint forbidding PEP-570 syntax

AlexWaygood:
We don't want to start using PEP-570 syntax in typeshed yet, because we don't want to unnecessarily break compatibility with Python 3.7 (#10113) -- mypy users will not be able to check their code using `--python-version 3.7` after we start using PEP-570 syntax in typeshed. However, now that we no longer run mypy or stubtest in CI using Python 3.7, none of our CI fails if somebody proposes using PEP-570 syntax in a PR (e.g. https://github.com/python/typeshed/pull/10646#discussion_r1312919259 wouldn't have been caught if it hadn't been for @srittau's sharp eye!).

This PR proposes adding a simple lint to `check_new_syntax.py` so that we don't accidentally start using PEP-570 syntax in typeshed before we formally drop support for Python 3.7.

### DIFF
--- a/tests/check_new_syntax.py
+++ b/tests/check_new_syntax.py
@@ -27,6 +27,9 @@ def check_new_syntax(tree: ast.AST, path: Path, stub: str) -> list[str]:
             self.generic_visit(node)
 
     class PEP570Finder(ast.NodeVisitor):
+        def __init__(self) -> None:
+            self.lineno: int | None = None
+        
         def _visit_function(self, node: ast.FunctionDef | ast.AsyncFunctionDef) -> None:
             self.lineno = node.lineno
             self.generic_visit(node)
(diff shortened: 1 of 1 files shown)


tests/check_new_syntax.py
```diff
@@ -26,7 +26,23 @@ def visit_If(self, node: ast.If) -> None:
                 )
             self.generic_visit(node)
 
+    class PEP570Finder(ast.NodeVisitor):
+        def _visit_function(self, node: ast.FunctionDef | ast.AsyncFunctionDef) -> None:
+            self.lineno = node.lineno
+            self.generic_visit(node)
+
+        visit_FunctionDef = visit_AsyncFunctionDef = _visit_function
```
AlexWaygood, line 34: This dance is necessary because `ast.arguments` instances don't have `lineno` attributes

hauntsaninja (approved):
Lgtm! My two thoughts were 1) stubs don't use lambdas 2) check_old_syntax.py

AlexWaygood:
> 2) check_old_syntax.py

`ast_lints_that_dont_belong_in_flake8_pyi.py`

PR #10662, opened 2023-09-04
Remove pathlib.PurePath.__eq__

### LINKED ISSUE #10661: `comparison-overlap` not triggered for pathlib.Path
`--strict-comparison` seems to assume that `pathlib.Path` objects can be equal to anything.

**To Reproduce** ([playground](https://mypy-play.net/?mypy=latest&python=3.11&flags=strict-equality&gist=f87d542ff65df4ef507e73244fe1e538))

```python
# run with --strict-equality
from pathlib import Path

class A: pass

'x' == 1  # comparison-overlap: this is always false
A() == 1  # comparison-overlap: this is always false
Path('x') == 1  # fine for mypy
```

**Expected Behavior**
Both lines should trigger 'non-overlapping equality check'

**Actual Behavior**
Only `'x' == 1` triggers 'non-overlapping equality check'

hauntsaninja:
Fixes #10661


hauntsaninja:
I think the primer hits here are real issues

test_cases/stdlib/check_pathlib.py
```diff
@@ -0,0 +1,7 @@
+# pyright: strict
```
AlexWaygood, line 1: Which pyright error codes are missing from https://github.com/python/typeshed/blob/main/pyrightconfig.testcases.json? It looks to me like the pyright issues on your first commit were just because you had the type: ignore on the wrong line; not sure we need this: https://github.com/python/typeshed/actions/runs/6078220264/job/16489123042 

```suggestion
```

test_cases/stdlib/check_pathlib.py
```diff
@@ -0,0 +1,7 @@
...
+from __future__ import annotations
+
+from pathlib import Path
+
+if Path("asdf") == "asdf":  # type: ignore
```
AlexWaygood, line 6:
```suggestion
# make sure this causes mypy to emit a --strict-equality error,
# and pyright to report a reportUnnecessaryComparison error
if Path("asdf") == "asdf":  # type: ignore
```

test_cases/stdlib/check_pathlib.py
hauntsaninja: Ah sorry, I just copied the file I was using to confirm that pyright used the same heuristic into the test case dir.

AlexWaygood (approved):
It still feels like we're sorta working around shortcomings in the heuristics mypy/pyright use here. (As you said in the issue, this kind of thing is going to affect anybody who writes a `__eq__` method in a `py.typed` runtime library, and `__eq__` methods are pretty common!)

But I agree that this is a useful error for mypy/pyright to emit, so I'm happy for this change to be made on the basis that practicality beats purity :-)

hauntsaninja:
Sounds good, I'll leave open for a bit in case anyone else chimes in!

Akuli (approved):

AlexWaygood:
Here is a small regression that this PR causes, FYI: currently mypy will not complain about this code (on my Windows machine, the expression evaluates to `True`):

```py
from pathlib import *
PureWindowsPath("foo/bar") == Path("foo/bar")
```

But with this PR, it will:

```console
$ mypy -c "from pathlib import *; PureWindowsPath('foo/bar') == Path('foo/bar')" --strict-equality --custom-typeshed-dir .
<string>:1: error: Non-overlapping equality check (left operand type: "PureWindowsPath", right operand type: "Path")  [comparison-overlap]
```

hauntsaninja:
@AlexWaygood oh, interesting example! Definitely a false positive in terms of what the type checker thinks it's doing, but that code is pretty suspect and I'm happy to have it trigger a warning of some kind — depending on your platform, it could be true that those could never be equal!

So I think I'm still fine with the PR (though I'll certainly add test case and comments). We can experiment with making `Path.__new__` more accurate in a follow up. I'm trying to teach mypy about more kinds of `__new__` over here: https://github.com/python/mypy/pull/16020

AlexWaygood:
This change caught a real bug in stubtest (that was introduced by, er, me)! https://github.com/python/mypy/actions/runs/6202155467/job/16840319289?pr=16121

PR #10663, opened 2023-09-05
Fix typing of FreeTypeFont.getlength

mobyw:
`FreeTypeFont.getlength` in Pillow is returned as float

Reference: https://github.com/python-pillow/Pillow/blob/main/src/PIL/ImageFont.py#L263


srittau (approved):
Thanks!

PR #10665, opened 2023-09-05
Update `importlib.metadata` to 3.12

sobolevn:
Source: https://github.com/python/cpython/blob/cd2ef21b076b494224985e266c5f5f8b37c66618/Lib/importlib/metadata/_meta.py


stdlib/importlib/metadata/_meta.pyi
```diff
@@ -8,15 +9,34 @@ class PackageMetadata(Protocol):
...
+        @overload
+        def get(self, name: str, failobj: None = None) -> str | None: ...
+        @overload
+        def get(self, name: str, failobj: _T) -> _T | str: ...
+        @overload
+        def get_all(self, name: str, failobj: None = None) -> list[Any] | None: ...
+        @overload
+        def get_all(self, name: str, failobj: _T) -> list[Any] | _T: ...
+    else:
+        def get_all(self, name: str, failobj: _T = ...) -> list[Any] | _T: ...
```
AlexWaygood, line 24: The TypeVar could go unsolved here if no value is passed for `failobj`, which would lead the return type being inferred as `list[Any] | Any` (I know this is a problem in the existing annotations we have here, but that doesn't mean that we shouldn't fix it ;)

It's likely that we should just make it an overloaded function on Python <3.12 as well. I don't think the contract for `get_all()` changed in Python 3.12, it's just that the `importlib_metadata` authors learned a few new `typing` tricks and figured out how to improve their annotations

stdlib/importlib/metadata/_meta.pyi
```diff
@@ -8,15 +9,34 @@ class PackageMetadata(Protocol):
...
-    #  Strictly speaking `__div__` was defined in 3.10, not __truediv__,
-    # but it should have always been `__truediv__`.
-    def __truediv__(self) -> SimplePath: ...
+if sys.version_info >= (3, 12):
+    class SimplePath(Protocol[_T]):
+        def joinpath(self) -> _T: ...
+        @property
+        def parent(self) -> _T: ...
+        def read_text(self) -> str: ...
+        def __truediv__(self, other: _T | str) -> _T: ...
```
AlexWaygood, line 32: Hmm, this has a bad annotation in the CPython/`importlib_metadata` source code. If other is a `str`, what's the inferred return type meant to be? We should probably leave this as it is for now, but maybe we should open an issue over at `importlib_metadata` asking for clarification

sobolevn: Fix: https://github.com/python/importlib_metadata/pull/469

AlexWaygood: Thank you!

AlexWaygood (approved):
Thanks!

PR #10667, opened 2023-09-06
Fix stubsabot handling of post releases

### LINKED ISSUE #10666: Stubsabot failed on Wed Sep 06 2023
Stubsabot runs are listed here: https://github.com/python/typeshed/actions/workflows/stubsabot.yml

hauntsaninja:
Fixes #10666

If our specifier was an exact match, we now exact match everything


srittau (approved):

AlexWaygood:
Hmm, looks like this is now causing stubsabot to propose PRs that break the stub uploader:
- #10673

PR #10668, opened 2023-09-06
chore: remove tree-sitter (March 6, 2024)

amaanq:
I added more complete stub types [here](https://github.com/tree-sitter/py-tree-sitter/blob/master/tree_sitter/binding.pyi) so I think this is no longer needed, my original plan was to update them here but I noticed in the docs that types are to be added if a project doesn't provide them already


JelleZijlstra:
Looks like we'll also have to update our stubs for `tree-sitter-languages`, which currently depend on the `tree-sitter` stubs. We'll also need a change like https://github.com/typeshed-internal/stub_uploader/pull/99/files.

amaanq:
Gotcha, just added the py.typed file then. I completely missed that section in the contributing file, my bad :)

I'll edit the date to specify removal for March 6, 2024

AlexWaygood:
> I'll edit the date to specify removal for March 6, 2024

Strictly speaking, our six-month clock for removal hasn't yet started FYI, since there hasn't yet been a _release_ that includes a `py.typed` file. Without a release that includes a `py.typed` file, the type annotations included with `tree-sitter` still aren't available for your end users, unfortunately:

```console
(.venv) $ pip install tree-sitter
Collecting tree-sitter
  Obtaining dependency information for tree-sitter from https://files.pythonhosted.org/packages/79/d1/9a07660d21dfbbdd2961979ae7477f4a18a9b6b8a26b2772972513e41e46/tree_sitter-0.20.2-cp311-cp311-win_amd64.whl.metadata
  Downloading tree_sitter-0.20.2-cp311-cp311-win_amd64.whl.metadata (7.4 kB)
Downloading tree_sitter-0.20.2-cp311-cp311-win_amd64.whl (105 kB)
   ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 105.2/105.2 kB 1.5 MB/s eta 0:00:00
Installing collected packages: tree-sitter
Successfully installed tree-sitter-0.20.2

(.venv) $ mypy -c "import tree_sitter"
<string>:1: error: Library stubs not installed for "tree_sitter"  [import]
<string>:1: note: Hint: "python3 -m pip install types-tree-sitter"
<string>:1: note: (or run "mypy --install-types" to install all missing stub packages)
<string>:1: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
```

AlexWaygood:
> Strictly speaking, our six-month clock for removal hasn't yet started FYI, since there hasn't yet been a _release_ that includes a `py.typed` file.

Gonna close this for now, since we don't yet know when we'll be able to remove these stubs, and this already has merge conflicts. It'll also be a pretty easy PR to recreate — PR welcome when the time comes! :)

PR #10669, opened 2023-09-06
Update `asyncio.tasks` for py312

AlexWaygood:
Runtime implementation added in https://github.com/python/cpython/commit/a474e04388c2ef6aca75c26cb70a1b6200235feb

### DIFF
--- a/stdlib/asyncio/tasks.pyi
+++ b/stdlib/asyncio/tasks.pyi
@@ -2,7 +2,7 @@ import concurrent.futures
 import sys
 from collections.abc import Awaitable, Coroutine, Generator, Iterable, Iterator
 from types import FrameType
-from typing import Any, Generic, TextIO, TypeVar, overload
+from typing import Any, Generic, Protocol, TextIO, TypeVar, overload
 from typing_extensions import Literal, TypeAlias
 
 from . import _CoroutineLike
@@ -14,7 +14,7 @@ if sys.version_info >= (3, 9):
 if sys.version_info >= (3, 11):
     from contextvars import Context
 
-__all__ = (
+__all__: tuple[str, ...] = (
     "Task",
     "create_task",
     "FIRST_COMPLETED",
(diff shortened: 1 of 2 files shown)


stdlib/asyncio/tasks.pyi
```diff
@@ -356,5 +359,40 @@ else:
...
+
+    class _CustomTaskConstructor(Protocol[_TaskT_co]):
+        def __call__(
+            self,
+            __coro: _TaskCompatibleCoro[Any],
+            loop: AbstractEventLoop,
+            name: str | None,
+            context: Context | None,
+            eager_start: bool,
+        ) -> _TaskT_co: ...
```
AlexWaygood, line 374: I think the idea might be that it's permissible for the custom task constructor to return anything that _duck-types_ as an `asyncio.Task` -- it doesn't necessarily _have_ to be a subclass of `asyncio.Task`. That would be even more complicated than what we have currently, however... I'd vote for keeping it simple until we get a complaint that we're not allowing somebody's duck-type-compatible third-party `Task` implementation

stdlib/asyncio/tasks.pyi
```diff
@@ -356,5 +381,40 @@ else:
...
 def _leave_task(loop: AbstractEventLoop, task: Task[Any]) -> None: ...
+
+if sys.version_info >= (3, 12):
+    _TaskT_co = TypeVar("_TaskT_co", bound=Task[Any], covariant=True)
+
+    class _CustomTaskConstructor(Protocol[_TaskT_co]):
+        def __call__(
+            self,
+            __coro: _TaskCompatibleCoro[Any],
+            loop: AbstractEventLoop,
```
srittau, line 392: To make it a bit more lenient:

```suggestion
            __coro: _TaskCompatibleCoro[Any],
            *,
            loop: AbstractEventLoop,
```

srittau (approved):
One nit, but looks good.

stdlib/asyncio/tasks.pyi
AlexWaygood: nice catch!

PR #10670, opened 2023-09-06
Update `enum` for py312

AlexWaygood:


stdlib/enum.pyi
```diff
@@ -148,6 +150,18 @@ class EnumMeta(type):
...
+    # Overload 3 (py312+ only): Value lookup on an already existing enum class (complex case)
+    #
+    # >>> class Foo(enum.Enum):
+    # ...     X = 1, 2, 3
+    # >>> Foo(1, 2, 3)
+    # <Foo.X: (1, 2, 3)>
+    #
+    if sys.version_info >= (3, 12):
+        @overload
+        def __call__(cls: type[_EnumMemberT], value: Any, *values: Any) -> _EnumMemberT: ...
```
AlexWaygood, line 163: I added a third overload here rather than modifying the first overload, as if this were the first overload, there'd likely be cases where type checkers _should_ be selecting the functional-API overload but would go for this one instead. It's a very lenient signature.

sobolevn (approved):
Consider adding tests for this new overload.

test_cases/stdlib/check_enum.py
```diff
@@ -2,8 +2,29 @@
...
+assert_type(C, Type[C])
+assert_type(D, Type[D])
+
+
+class EnumOfTuples(enum.Enum):
+    X = 1, 2, 3
+    Y = 4, 5, 6
+
+
+assert_type(EnumOfTuples((1, 2, 3)), EnumOfTuples)
```
AlexWaygood, line 25: @sobolevn, I tried adding this test:

```suggestion
assert_type(EnumOfTuples((1, 2, 3)), EnumOfTuples)

if sys.version_info >= (3, 12):
    assert_type(EnumOfTuples(1, 2, 3), EnumOfTuples)
```

Unfortunately, the test failed :(

```
test_cases\stdlib\check_enum.py:28: error:
Too many arguments for "EnumOfTuples"  [call-arg]
        assert_type(EnumOfTuples(1, 2, 3), EnumOfTuples)
                    ^~~~~~~~~~~~~~~~~~~~~


Test completed with errors
```

Do you think there's some special-casing in mypy that will also need to be updated?

AlexWaygood: The tests I've added show that there is no regression w.r.t. existing functionality, anyway

sobolevn: There are two places where `enum.Enum` is special-cased:
1. During `semanal`: https://github.com/python/mypy/blob/9a35360739ced871feb6331a14a7bbacce00c7dc/mypy/semanal.py#L3095
2. In plugins: https://github.com/python/mypy/blob/9a35360739ced871feb6331a14a7bbacce00c7dc/mypy/plugins/default.py#L112-L123

Looks like it is related to `1.`

AlexWaygood: The test continues to fail if I apply this diff, combining overloads 1 and 3:

<details>
<summary>Diff</summary>

```diff
diff --git a/stdlib/enum.pyi b/stdlib/enum.pyi
index 10ea19257..157d09c84 100644
--- a/stdlib/enum.pyi
+++ b/stdlib/enum.pyi
@@ -120,9 +120,13 @@ class EnumMeta(type):
     def __bool__(self) -> Literal[True]: ...
     def __dir__(self) -> list[str]: ...

-    # Overload 1: Value lookup on an already existing enum class (simple case)
-    @overload
-    def __call__(cls: type[_EnumMemberT], value: Any, names: None = None) -> _EnumMemberT: ...
+    # Overload 1: Value lookup on an already existing enum class
+    if sys.version_info >= (3, 12):
+        @overload
+        def __call__(cls: type[_EnumMemberT], value: Any, names: Any | None = None, *values: Any) -> _EnumMemberT: ...
+    else:
+        @overload
+        def __call__(cls: type[_EnumMemberT], value: Any, names: None = None) -> _EnumMemberT: ...

     # Overload 2: Functional API for constructing new enum classes.
     if sys.version_info >= (3, 11):
@@ -151,17 +155,6 @@ class EnumMeta(type):
             start: int = 1,
         ) -> type[Enum]: ...

-    # Overload 3 (py312+ only): Value lookup on an already existing enum class (complex case)
-    #
-    # >>> class Foo(enum.Enum):
-    # ...     X = 1, 2, 3
-    # >>> Foo(1, 2, 3)
-    # <Foo.X: (1, 2, 3)>
-    #
-    if sys.version_info >= (3, 12):
-        @overload
-        def __call__(cls: type[_EnumMemberT], value: Any, *values: Any) -> _EnumMemberT: ...
```

</details>

So I'm not sure this is fixable in typeshed :(

sobolevn: Let's report it to mypy 👍

AlexWaygood: Yup, for now I've added a TODO 👍

PR #10672, opened 2023-09-06
First stab at tz aware datetimes using PEP696

Gobot1234:
DON'T MERGE THIS


AlexWaygood:
Thanks! I'm going to close this for now just to keep the number of open PRs here vaguely under control. Let's look at this again once mypy has better support for PEP-696 and (🤞) the PEP's been accepted by the Steering Council

PR #10679, opened 2023-09-07
Improve the accuracy of `(default)dict.__(r)or__`

### LINKED ISSUE #10678: False negatives with `dict.__(r)or__`
I have code equivalent to the following:

```python
from requests.structures import CaseInsensitiveDict

CaseInsensitiveDict() | dict()
```

`mypy` does not error.

My environment has:

* `mypy==1.5.1`
* `types-requests==2.31.0.2`
* Python 3.11.5

At runtime I get:

```
TypeError: unsupported operand type(s) for |: 'CaseInsensitiveDict' and 'dict'
```

I would expect an `[operator]` error. For example, if I swap the `CaseInsensitiveDict()` for `1`, I get:

```
error: Unsupported operand types for | ("int" and "dict[str, str]")  [operator]
```

I am happy to contribute, but I'm not sure what change is needed. I imagine a change is needed in https://github.com/python/typeshed/blob/main/stubs/requests/requests/structures.pyi.

AlexWaygood:
Fixes #10678


AlexWaygood:
Oh, one mypy_primer error isn't actually too bad at all. I vote for doing this, in that case!

Akuli:
Maybe add a test case? We can't use mypy_primer to ensure we don't break this again, since the improvement doesn't show up in the diff and open-source projects are unlikely to contain typing errors anyway.

srittau (approved):

AlexWaygood:
Sure, I'll add a test

srittau:
And the one primer hit actually show a genuine problem as using an arbitrary mapping wouldn't work either. This is actually something I wasn't aware of, so +10 for changing these annotations.

srittau:
One thing I'd try is using a small custom mapping-like type in the tests instead of `os.environ`. The latter has a lot going on, with generics and python-specific branches, and I wouldn't be surprised if a typechecker would special-case it due to its importance.

AlexWaygood:
Okay, for now I've just commented out the tests that were failing on pyright and left TODO comments. There's definitely a behaviour difference from mypy there -- to me it looks like a bug, but not sure. I'll file an issue with pyright later.

srittau (approved):

Akuli (approved):
At first glance the tests seemed like too much copy/pasta, but I think it's fine because dict and defaultdict behave differently.

AlexWaygood:
> At first glance the tests seemed like too much copy/pasta, but I think it's fine because dict and defaultdict behave differently.

Yeah, `dict.__or__` always returns `dict`, whereas `defaultdict.__or__` returns `Self`. I considered putting all the tests into a `check_pep584.py` file, but thought it would be too confusing to have some of the tests for `dict` in a file that had a name that didn't start with `check_dict`.

PR #10680, opened 2023-09-07
re: group methods return AnyStr or None

ehandal:
The `group`, `groups`, `groupdict`, and `__getitem__` methods default to None if a group doesn't participate in a match.


AlexWaygood:
Thanks for the PR! Unfortunately, this has been tried many times before, and as you can see, it would be far too disruptive for us to be able to consider merging this.

We could probably document this better. Might you be interested in changing your PR so that it just adds an explanatory comment about why we have the annotations the way they currently are?

PR #10685, opened 2023-09-08
Update `poplib` and `smtplib` for py312

AlexWaygood:
The runtime changes were made in https://github.com/python/cpython/commit/ef0e72b31d22f780d3a165d7d0471806061fe380


stdlib/poplib.pyi
```diff
@@ -51,14 +52,19 @@ class POP3:
...
+            self,
+            host: str,
+            port: int = 995,
+            keyfile: str | None = None,
+            certfile: str | None = None,
+            timeout: float = ...,
+            context: ssl.SSLContext | None = None,
+        ) -> None: ...
     # "context" is actually the last argument, but that breaks LSP and it doesn't really matter because all the arguments are ignored
     def stls(self, context: Any = None, keyfile: Any = None, certfile: Any = None) -> NoReturn: ...
```
srittau, line 70: The arguments were removed here as well.

srittau (approved):
I spotted one more instance, but LGTM apart from that.

stdlib/poplib.pyi
AlexWaygood: Ah, good spot. Stubtest didn't complain because the method's already allowlisted:

https://github.com/python/typeshed/blob/53144ca57202a7bc5b28fd8e48e6bf62ad1b8a5b/tests/stubtest_allowlists/py3_common.txt#L111

srittau (approved):

PR #10686, opened 2023-09-08
Fix `importlib.util.source_hash` return annotation

cdce8p:
> Return the hash of source_bytes as bytes. [...]
https://docs.python.org/3/library/importlib.html#importlib.util.source_hash

```py
>>> from pathlib import Path
>>> from importlib.util import source_hash
>>> 
>>> p = Path("README.md")
>>> print(source_hash(p.read_bytes()))
b'\x14\x9a~P\x1b\x89\x9f$'
```

Originally added in #6557


hauntsaninja (approved):
Thank you!

PR #10690, opened 2023-09-10
Fix stubtest complaints for various `.get()` methods

AlexWaygood:
`key` and `default` can be passed as keyword arguments for all of these methods


AlexWaygood:
Looks like a lot of the `dict.get()`-related errors in https://github.com/python/typeshed/pull/10690#issuecomment-1712852968 are actually "desirable" to some extent; mypy seems to be flagging a lot of issues in user code there. But the proposed changes to `dict.get()` are disruptive enough that they deserve to be considered in their own PR, so I've backed them out for now.

stdlib/collections/__init__.pyi
```diff
@@ -96,6 +96,11 @@ class UserDict(MutableMapping[_KT, _VT], Generic[_KT, _VT]):
         def __ior__(self, other: SupportsKeysAndGetItem[_KT, _VT]) -> Self: ...
         @overload
         def __ior__(self, other: Iterable[tuple[_KT, _VT]]) -> Self: ...
+    if sys.version_info >= (3, 12):
+        @overload
+        def get(self, key: _KT, default: None = None) -> _VT | None: ...
+        @overload
+        def get(self, key: _KT, default: _VT | _T) -> _VT | _T: ...
```
Daverball, line 103: Isn't the `_VT` in the default parameter redundant and something that was previously causing bugs? I seem to recall some issue and PR regarding that, because when the default is of `_VT`, then `_T` remains unbound and results in an `Any` in the return value union.

stdlib/weakref.pyi
```diff
@@ -65,6 +65,10 @@ class WeakValueDictionary(MutableMapping[_KT, _VT]):
     def copy(self) -> WeakValueDictionary[_KT, _VT]: ...
     __copy__ = copy
     def __deepcopy__(self, memo: Any) -> Self: ...
+    @overload
+    def get(self, key: _KT, default: None = None) -> _VT | None: ...
+    @overload
+    def get(self, key: _KT, default: _KT | _T) -> _VT | _T: ...
```
Daverball, line 71: Shouldn't this be `_VT | _T` too? Assuming it shouldn't actually be just `_T` as pointed out in the previous comment.

AlexWaygood: whoooops, good catch!

```suggestion
    def get(self, key: _KT, default: _VT | _T) -> _VT | _T: ...
```

stdlib/collections/__init__.pyi
AlexWaygood: Hmmmmm. Yes, I think you're right. Let me play around a little.

AlexWaygood:
Thanks for the great review @Daverball!

stdlib/collections/__init__.pyi
```diff
@@ -96,6 +96,11 @@ class UserDict(MutableMapping[_KT, _VT], Generic[_KT, _VT]):
         def __ior__(self, other: SupportsKeysAndGetItem[_KT, _VT]) -> Self: ...
         @overload
         def __ior__(self, other: Iterable[tuple[_KT, _VT]]) -> Self: ...
+    if sys.version_info >= (3, 12):
```
JelleZijlstra, line 99: I guess this is from python/cpython#17910

JelleZijlstra (approved):

PR #10691, opened 2023-09-10
Fix various stubtest complaints on Python 3.12

AlexWaygood:
This fixes all remaining new-in-py312 stubtest complaints that don't already have PRs open for them:
- `collections.UserDict.get` is tackled as part of #10690
- The `enum`-related allowlist entries are tackled in #10670


stdlib/http/client.pyi
```diff
@@ -169,6 +169,9 @@ class HTTPConnection:
     ) -> None: ...
     def getresponse(self) -> HTTPResponse: ...
     def set_debuglevel(self, level: int) -> None: ...
+    if sys.version_info >= (3, 12):
+        def get_proxy_response_headers(self) -> HTTPMessage | None: ...
```
JelleZijlstra, line 173: The docstring says "returns a dictionary" but that appears to be a lie, this type is right

JelleZijlstra (approved):

Issue #10592, opened 2023-08-16
Remove redis (not before February 2025)


srittau:
----

~~Deferred: As of redis-py 5.0.3 the upstream annotations are lacking. We should not remove these stubs, until the upstream annotations reach parity with the stubs.~~

srittau:
Redis 5.0.0 added a py.typed file. Still, I wonder whether we should remove our stubs at this point, as the redis code is basically untyped, even quite fundamental classes and methods.

Avasam:
@AlexWaygood typeshed recently supports creating partial stubs on purpose. Exactly for cases like this. It's something that is likely to happen eventually in pywin32 as well.

We can mark the stub as partial, but NOT stubtest incomplete. Then ignore known typed modules (with a comment) in stubtest.

Type tests will of course need the base library to be installed (can't remember if they already do that. Pyright?)
Idk if it'll need a stubsabot update to prevent re-openning an issue.

AlexWaygood:
Oh, so if we mark the stubs as `partial`, type checkers will only look at the annotations for the modules we include in typeshed, and fall back to inline types for the modules we don't include? (I'm not too familiar with the details of PEP-561!)

meshantz:
Since there are so many upstream issues yet to be resolve, can we remove the recommendation to remove `types-redis` mentioned on the package page https://pypi.org/project/types-redis/:

> Note: The redis package includes type annotations or type stubs since version 5.0.0. Please uninstall the types-redis package if you use this or a newer version.

Or if you don't want it removed, at least add some caveats?

srittau:
@meshantz Thanks for the suggestion. We now removed the wording and instead suggest du leave the package installed for now.

Avasam:
Note that https://github.com/typeshed-internal/stub_uploader/issues/108 will probably need to be fixed to properly support partial type stubs for Redis.

Avasam:
So, we're in march. Do we know which way we wanna go? I haven't checked the state of typing in redis in a while, but assuming it's still incomplete compared to typeshed, do we wanna keep going for a gradual transition?
https://github.com/redis/redis-py/issues/2730 is still an issue when it comes to improving the typeshed redis stubs.

sobolevn:
It is incomplete, many types are just `Any`s, I vote for improving `redis` first and removing `types-redis` after that.

srittau:
I'm marking this issue as deferred for now.

srittau:
[...]

PR #10694, opened 2023-09-11
[redis] Don't recommend uninstalling types-redis

srittau:
Cf. #10592

### DIFF
--- a/stubs/redis/METADATA.toml
+++ b/stubs/redis/METADATA.toml
@@ -3,7 +3,10 @@ upstream_repository = "https://github.com/redis/redis-py"
 # Requires a version of cryptography with a `py.typed` file
 requires = ["cryptography>=35.0.0", "types-pyOpenSSL"]
 partial_stub = true
-obsolete_since = "5.0.0" # Released on 2023-08-15
+# While upstream added a py.typed file, the upstream annotations are rather
+# incomplete. By using "obsolete_since" we recommend deinstalling this
+# package, which isn't the best idea at the moment.
+# obsolete_since = "5.0.0" # Released on 2023-08-15
 
 [tool.stubtest]
 ignore_missing_stub = true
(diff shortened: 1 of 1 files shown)


AlexWaygood (commented):
Could we add something in the `extra_description` field noting that upstream does have a py.typed, but that these stubs might be useful anyway?

stubs/redis/METADATA.toml
```diff
@@ -3,7 +3,15 @@ upstream_repository = "https://github.com/redis/redis-py"
...
 partial_stub = true
-obsolete_since = "5.0.0" # Released on 2023-08-15
+# While upstream added a py.typed file, the upstream annotations are rather
+# incomplete. By using "obsolete_since" we recommend deinstalling this
+# package, which isn't the best idea at the moment.
+# obsolete_since = "5.0.0" # Released on 2023-08-15
+extra_description = """\
+    Redis-py 5.0.0 added a py.typed file, but the annotations are fairly \
+    incomplete, which is why the typeshed maintainers recommend to continue \
+    using `types-redis` for the time being.\
```
AlexWaygood, line 13: It's hard to get the wording right here without making it sound like we're making a dig at the Redis maintainers, but maybe this is slightly less opinionated language:

```suggestion
    Note: Redis-py 5.0.0 added a py.typed file, but the inline annotations \
    are incomplete. Continuing to use `types-redis` for the time being may \
    lead to superior results.\
```

AlexWaygood (approved):
Thanks! Stubsabot will file a PR trying to mark these stubs as obsolete again, but we can mark it as "DO-NOT-MERGE".

We may possibly want to add some kind of `stubsabot_ignore` field to `METADATA.toml` files, for repos where we just want stubsabot to go away for now. Right now we've got `redis` and `tree-sitter`; previously we had `pyvmomi` (see discussion in #10149)

Issue #10145, opened 2023-05-05
`restval`/`restkey` from `csv.DictReader` can be `Any`


Viicos:
This is how `restval` and `restkey` are currently defined:

https://github.com/python/typeshed/blob/348f46452f8a864c5e0f279398fad0a943f67ab1/stdlib/csv.pyi#L69-L80

But looking at the [code](https://github.com/python/cpython/blob/1c420e138fd828895b6bd3c44ef99156e8796095/Lib/csv.py#L127-L131) and [docs](https://docs.python.org/3/library/csv.html#csv.DictReader), they basically can take any value (or at least any hashable value for `restkey`). Is it required for these two parameters to be `str | None`?

ljluestc:
import ast
import collections
import os

def find_constants(directory):
    constants = collections.defaultdict(list)

    for root, dirs, files in os.walk(directory):
        for file in files:
            if file.endswith(".py"):
                filepath = os.path.join(root, file)
                with open(filepath, "r") as f:
                    code = f.read()
                    tree = ast.parse(code)

                for node in ast.walk(tree):
                    if isinstance(node, ast.Assign) and len(node.targets) == 1 and isinstance(node.targets[0], ast.Name) and isinstance(node.value, ast.Constant):
                        constant_name = node.targets[0].id
                        constant_value = node.value.value
                        constants[constant_name].append((constant_value, filepath))

    return constants

def main():
    directory = "/path/to/your/codebase"
    constants = find_constants(directory)

    for name, occurrences in constants.items():
        if len(occurrences) > 1:
            print(f"Constant '{name}' is defined in multiple modules:")
            for value, filepath in occurrences:
                print(f"  Value: {value}, Defined in: {filepath}")
            print("Consider re-exporting or reusing this constant.")
            print()

if __name__ == "__main__":
    main()

srittau:
Considering that `DictReader` is already generic over the fieldnames, it would make sense to me to use `_T` for `restkey` as well. PR welcome!

PR #10696, opened 2023-09-11
Update `restval`/`restkey` types in `csv.DictReader`

Viicos:
Fixes https://github.com/python/typeshed/issues/10145

I'm not sure about the `Any` part, but as iterating over the `DictReader` gives a `dict[_T | Any, str | Any]`, figured I would use the same types


srittau (approved):
Thanks!

Issue #10700, opened 2023-09-11
Declare `http.cookies.SimpleCookie` as non generic(?)


flaeppe:
I'm a bit confused as to why `http.cookies.SimpleCookie` is declared as a generic class and not like below?

```python
class SimpleCookie(BaseCookie[str]):
    ...
```

The [documentation](https://docs.python.org/3/library/http.cookies.html#http.cookies.SimpleCookie) on `SimpleCookie` only talks about strings:

> class http.cookies.SimpleCookie([input])[¶](https://docs.python.org/3/library/http.cookies.html#http.cookies.SimpleCookie)
This class derives from [BaseCookie](https://docs.python.org/3/library/http.cookies.html#http.cookies.BaseCookie) and overrides value_decode() and value_encode(). SimpleCookie supports strings as cookie values. When setting the value, SimpleCookie calls the builtin [str()](https://docs.python.org/3/library/stdtypes.html#str) to convert the value to a string. Values received from HTTP are kept as strings.

And from what I'm seeing, implementations `value_decode` and `value_encode` for `SimpleCookie` only produces strings (but perhaps I'm missing something)?

Ref: https://github.com/python/cpython/blob/1ee50e2a78f644d81d341a08562073ad169d8cc7/Lib/http/cookies.py#L607-L612

PR #248, opened 2016-06-03
Improve http

tharvik:
Add types to `http`.

PR #10701, opened 2023-09-11
Make `http.cookies.SimpleCookie` non-generic

flaeppe:
Fixes: #10700


srittau (approved):
Thanks! While the primer hits indicate that this change is disruptive, looking at the hits in depth indicates that the code in question can be simplified significantly. Currently, using `SimpleCookie` requires an explicit type annotation to specify the generic type (which can only be `str`), so I think the benefits outweighs the disruption significantly.

I'll leave it open for now if other maintainers want to chime in.

AlexWaygood:
I'm not too familiar with this module, so this may be a dumb question, but should even `BaseCookie` be generic? I can't find any uses on `grep.app` where `BaseCookie` is subscripted with anything other than `str`, and the docs don't shine any light on a situation where you might have a `Morsel[bytes]` or whatever as the value in a `BaseCookie` dictionary.

```pycon
>>> from http.cookies import BaseCookie
>>> b = BaseCookie()
>>> b['foo'] = b'bar'
>>> b['foo'
<Morsel: foo=b'bar'>
>>> _.coded_value
"b'bar'"
>>> type(b['foo'].coded_value)
<class 'str'>
```

JelleZijlstra:
Chances are it's a Python 2 remnant: maybe in Python 2 you could have both `str` and `unicode` cookies.

AlexWaygood:
Looks like both classes were originally made generic in:
- #248

But there's no explanation in that PR thread as to why

flaeppe:
[Documentation for `BaseCookie.value_encode`](https://docs.python.org/3/library/http.cookies.html#http.cookies.BaseCookie.value_encode) says that `coded_value` will always "be converted to a string". Where it is set as `Morsel.coded_value`. So I suppose it should be valid to declare `BaseCookie[int]` and then implement a `value_decode` that parses an `int` from a `str`?

> I'm not too familiar with this module, so this may be a dumb question, but should even `BaseCookie` be generic? I can't find any uses on [grep.app](https://grep.app/search?current=2&q=BaseCookie%5B&case=true&words=true&filter%5Blang%5D%5B0%5D=Python) where `BaseCookie` is subscripted with anything other than `str`, and the docs don't shine any light on a situation where you might have a `Morsel[bytes]` or whatever as the value in a `BaseCookie` dictionary.
> 
> ```
> >>> from http.cookies import BaseCookie
> >>> b = BaseCookie()
> >>> b['foo'] = b'bar'
> >>> b['foo'
> <Morsel: foo=b'bar'>
> >>> _.coded_value
> "b'bar'"
> >>> type(b['foo'].coded_value)
> <class 'str'>
> ```

I think an additional thing to check up on here is this

```
>>> from http.cookies import BaseCookie
>>> class C(BaseCookie):
...     def value_encode(self, value):
...         return value, value
...
>>> c = C()
>>> c["foo"] = b"bar"
>>> c["foo"]
<Morsel: foo=b'bar'>
>>> c["foo"].coded_value
b'bar'
>>> type(c["foo"].coded_value)
<class 'bytes'>
>>> c["foo"].output()
"Set-Cookie: foo=b'bar'"
```

`Morsel.coded_value` is always stringified before it's sent. At least that's what I interpret from "`coded_value` will always be converted to a string".

Funny thing is though, just because you've declared `BaseCookie[int]` it won't make `.coded_value` return `int`. And the only control over that is to enforce usage of `Morsel[int]` as set value.

Consider this program, which passes, but `coded_value` is `<class 'str'>`

```python
from http.cookies import BaseCookie, Morsel
class C(BaseCookie[int]):
    ...

c = C()
c["foo"] = "1"
coded_value = c["foo"].coded_value
reveal_type(coded_value)  # Revealed type is "builtins.int"
```

But if you'd instead use below, types align.

```python
m = Morsel[int]()
m.set("foo", "1", 1)
c["foo"] = m
```

AlexWaygood:
OK cool, sounds like `BaseCookie` should remain generic, then. Thanks for the research and the thorough explanation @flaeppe!!

AlexWaygood (approved):
LGTM

Issue #10699, opened 2023-09-11
`value_encode` and `value_decode` of `http.cookies.BaseCookie` not returning tuples


flaeppe:
[Documentation](https://docs.python.org/3/library/http.cookies.html#cookie-objects) states

> Return a tuple (real_value, coded_value) [...]

For both `BaseCookie.value_decode` and `BaseCookie.value_encode`, although typeshed doesn't align:

https://github.com/python/typeshed/blob/32535ee8952dffecbe7a9fd04e5ed8ba0f1c8da2/stdlib/http/cookies.pyi#L50-L53

sobolevn:
@flaeppe please, feel free to send a PR :)

flaeppe:
Yes, already done: #10702

PR #10702, opened 2023-09-11
Make `value_decode` and `value_encode` of `BaseCookie` return 2-tuple

flaeppe:
Fixes: #10699


stdlib/http/cookies.pyi
```diff
@@ -49,8 +49,8 @@ class Morsel(dict[str, Any], Generic[_T]):
 
 class BaseCookie(dict[str, Morsel[_T]], Generic[_T]):
     def __init__(self, input: _DataType | None = None) -> None: ...
-    def value_decode(self, val: str) -> _T: ...
-    def value_encode(self, val: _T) -> str: ...
+    def value_decode(self, val: str) -> tuple[_T, str]: ...
+    def value_encode(self, val: _T) -> tuple[str, _T]: ...
```
flaeppe, line 53: I based the argument positioning by looking at `SimpleCookie`'s implementation:

https://github.com/python/cpython/blob/1ee50e2a78f644d81d341a08562073ad169d8cc7/Lib/http/cookies.py#L607-L612

I think that's what it should look like? Or is it incorrect? I'm getting a bit confused by [the documentation](https://docs.python.org/3/library/http.cookies.html#http.cookies.BaseCookie.value_decode)..

Perhaps I'm just mixing things up regarding en/decode:ing

srittau, line 53: The documentation says:

> BaseCookie.value_encode(val)
> Return a tuple (real_value, coded_value). val can be any type, but coded_value will always be converted to a string. [...]

So I think this should be:

```suggestion
    def value_encode(self, val: _T) -> tuple[_T, str]: ...
```

srittau (approved):
Thanks!

Issue #5835, opened 2021-08-02
Distribute stdlib types as types-stdlib


srittau:
This has been briefly discussed before, but I think it would be a good idea if we started distributing the stdlib types in a separate package `types-stdlib`, similar to third-party packages:

* Type checkers could optionally depend on (a specific version of) that package, instead of shipping their own copy, reducing coupling and arguably improving maintainability.
* Third-party stubs could require a minimum version of the `types-stdlib` package, for example if they are relying on `_typeshed` features or for cases like #5762. Type checkers that opt to distribute a copy of the stdlib, could warn if incompatible stubs are used.

Technical considerations:

* The versioning scheme would be similar to third-party stubs, e.g. `types-stdlib 3.10.X`, where `X` is incremented automatically. We should probably increment the version number as soon as the first beta is released and we are reasonably sure we have added all new/changed API.
* The stubs would be installed into a special directory, for example `stdlib`, or even a non-importable name like `types-stdlib` (so we'd have stubs `stdlib/sys.pyi` etc.) so that they are only used by type checkers that support this package and don't interfere.

gruebel:
> The versioning scheme would be similar to third-party stubs, e.g. types-stdlib 3.10.X, where X is incremented automatically. We should probably increment the version number as soon as the first beta is released and we are reasonably sure we have added all new/changed API.

Wouldn't it be better to go with a pre release suffix first, then start with `0`, when also Python `3.10.0` is released and after that do the auto increment?
Or stay close with the Python version and increment the suffix part `3.10.0-123`?

But I definitely like the idea 👍

srittau:
We don't do that with third-party stubs either, since the third element has no influence on the API. Also, 3.10.0, for example, would not be a pre-release from typeshed's view. 3.10.x stubs are also viable for Python 3.9 and before, but additionally support Python 3.10.

Akuli:
Idea: Rename `stdlib` to `stubs/stdlib` so that it is just like any other package, and also not confusingly outside `stubs/` even though it contains stubs.

Cheaterman:
This is an excellent idea! I was hitting #7085 / #7086 today, and I'd have loved to just upgrade types-stdlib to see it fixed :-)

JelleZijlstra:
I would find this very useful, and pyanalyze would definitely use it. Currently I have to manually make a new release of https://github.com/JelleZijlstra/typeshed_client when I need it.

[...]

PR #10706, opened 2023-09-12
openpyxl: Type usages of `PIL` and `zipfile`

Avasam:

Typed all usages of non-type, non excel, libraries external to openpyxl: `PIL` and `zipfile`.
Completed annotations of affected methods and classes. (unless they got complicated or uncertain)

### DIFF
--- a/stubs/openpyxl/openpyxl/packaging/relationship.pyi
+++ b/stubs/openpyxl/openpyxl/packaging/relationship.pyi
@@ -1,10 +1,13 @@
 from _typeshed import Incomplete, Unused
 from collections.abc import Generator
-from typing import ClassVar, overload
+from typing import ClassVar, TypeVar, overload
 from typing_extensions import Literal
+from zipfile import ZipFile
 
 from openpyxl.descriptors.base import Alias, String
 from openpyxl.descriptors.serialisable import Serialisable
+
+_SerialisableT = TypeVar("_SerialisableT", bound=Serialisable)
 
 class Relationship(Serialisable):
     tagname: ClassVar[str]
@@ -37,5 +40,12 @@ class RelationshipList(Serialisable):
     def to_tree(self): ...
 
 def get_rels_path(path): ...
-def get_dependents(archive, filename): ...
-def get_rel(archive, deps, id: Incomplete | None = None, cls: Incomplete | None = None): ...
+def get_dependents(archive: ZipFile, filename: str) -> RelationshipList: ...
+@overload
+def get_rel(
+    archive: ZipFile, deps: RelationshipList, id: str, cls: type[_SerialisableT]
+) -> _SerialisableT: ...  # incomplete: this could be restricted further from "Serialisable"
+@overload
+def get_rel(
+    archive: ZipFile, deps: RelationshipList, id: str | None = None, *, cls: type[_SerialisableT]
+) -> _SerialisableT: ...  # incomplete: this could be restricted further from "Serialisable"
--- a/stubs/openpyxl/openpyxl/reader/excel.pyi
+++ b/stubs/openpyxl/openpyxl/reader/excel.pyi
@@ -1,4 +1,5 @@
-from _typeshed import Incomplete, StrPath, SupportsRead
+from _typeshed import Incomplete, StrPath
+from typing import IO
 from typing_extensions import Final, Literal, TypeAlias
 from zipfile import ZipFile
 
@@ -26,7 +27,7 @@ class ExcelReader:
 
     def __init__(
         self,
-        fn: SupportsRead[bytes] | str,
+        fn: StrPath | IO[bytes],
         read_only: bool = False,
         keep_vba: bool = False,
         data_only: bool = False,
@@ -44,7 +45,7 @@ class ExcelReader:
     def read(self) -> None: ...
 
 def load_workbook(
-    filename: SupportsRead[bytes] | StrPath,
+    filename: StrPath | IO[bytes],
     read_only: bool = False,
     keep_vba: bool = False,
     data_only: bool = False,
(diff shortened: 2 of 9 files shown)


stubs/openpyxl/openpyxl/reader/excel.pyi
```diff
@@ -44,7 +45,7 @@ class ExcelReader:
     def read(self) -> None: ...
 
 def load_workbook(
-    filename: SupportsRead[bytes] | StrPath,
+    filename: StrPath | IO[bytes],
```
Avasam, line 48: https://github.com/python/typeshed/pull/9511/files#r1113278179 Avasam 
> This, and `ExcelReader.__init__` is ultimately passed to `ZipFile` in `_validate_archive`.
> 
> Even in read mode it still needs a `name` variable, `seek` and `tell` methods.

AlexWaygood
> I'd be okay with switching both of the conflicting annotations to `IO[bytes]`. We could probably come up with a more precise protocol, but for now consistency with the stdlib `zipfile` stub is probably the most important thing

stubs/openpyxl/openpyxl/packaging/relationship.pyi
```diff
@@ -37,5 +40,12 @@ class RelationshipList(Serialisable):
...
-def get_rel(archive, deps, id: Incomplete | None = None, cls: Incomplete | None = None): ...
+def get_dependents(archive: ZipFile, filename: str) -> RelationshipList: ...
+@overload
+def get_rel(
+    archive: ZipFile, deps: RelationshipList, id: str, cls: type[_SerialisableT]
+) -> _SerialisableT: ...  # incomplete: this could be restricted further from "Serialisable"
+@overload
+def get_rel(
+    archive: ZipFile, deps: RelationshipList, id: str | None = None, *, cls: type[_SerialisableT]
+) -> _SerialisableT: ...  # incomplete: this could be restricted further from "Serialisable"
```
Avasam, line 51: `cls` is always used at runtime: https://foss.heptapod.net/openpyxl/openpyxl/-/blob/branch/3.1/openpyxl/packaging/relationship.py#L143-150

About the `# incomplete`: Only `CacheDefinition`, `RecordList` and `TableDefinition` seem to define `rel_type`. (they also define is as ClassVars, which matches expected behaviour). But I can't find *any* class defining `deps`. The attribute seems to only be assigned in `get_rel` and used in `WorkbookParser.pivot_caches`. So I'm not exactly sure about which `Serialisable` child I should restrict to, if any.

Avasam: The more I look into it, the more I realize `obj.deps` is for internal used only and short-lived. Udated commit should be accurate.

stubs/openpyxl/openpyxl/reader/excel.pyi
Avasam: This change is responsible for the `mypy_primer` diff.
Affected code in Pandas: https://github.com/pandas-dev/pandas/blob/705d4312cf1d94ef2497bcd8091e0eabd1085f4a/pandas/io/excel/_openpyxl.py#L566

JelleZijlstra: Given the primer hit I think we should bite the bullet, introduce a protocol in zipfile, and use an equivalent protocol here.

stubs/openpyxl/openpyxl/packaging/relationship.pyi
```diff
@@ -37,5 +44,20 @@ class RelationshipList(Serialisable):
...
+@overload
+def get_rel(
+    archive: ZipFile, deps: RelationshipList, id: None = None, *, cls: type[_SerialisableRelTypeT]
+) -> _SerialisableRelTypeT | None: ...
+@overload
+def get_rel(
+    archive: ZipFile, deps: RelationshipList, id: None, cls: type[_SerialisableRelTypeT]
+) -> _SerialisableRelTypeT | None: ...
+@overload
+def get_rel(archive: ZipFile, deps: RelationshipList, id: str, *, cls: type[_SerialisableT]) -> _SerialisableT: ...
```
JelleZijlstra, line 61: Isn't this redundant with the next overload?

Avasam: I wanna say you're right? Probably a leftover from before I realized `cls` wasn't optional.

stubs/openpyxl/openpyxl/reader/excel.pyi
Avasam: I don't mind holding off on this PR until we have a proper Protocol defined for zipfile.

I'll link #5835 as this is another use-case for it. Otherwise we'll have to duplicate the protocol in openpyxl for a while.

JelleZijlstra: I don't think #5835 is happening anytime soon so for now we'll just have to duplicate it.

stubs/openpyxl/openpyxl/packaging/relationship.pyi
Avasam: I wanna say you're right? Probably a leftover from before I realized `cls` wasn't optional.

Avasam: Not sure how doable it is, but I suggested catching this at mypy: https://github.com/python/mypy/issues/16164

Avasam:
I abstracted the protocol issue away into an alias that clearly references #10880 and reduces false-positives. This should allow this PR to be merged without waiting for protocol and mypy updates.

srittau (approved):
Thanks! LGTM for now.

PR #10710, opened 2023-09-13
Update functools.pyi

rubikproxy:
The added feature, the memoize decorator, provides a convenient way to cache and reuse the results of a function, thereby improving performance by avoiding redundant computations. It works by storing previously computed results in a cache and returning them when the same inputs are provided to the function. This can be particularly useful for functions with expensive computations or I/O operations, as it reduces the need to recompute results for the same inputs, resulting in faster execution and improved efficiency.

### DIFF
--- a/stdlib/functools.pyi
+++ b/stdlib/functools.pyi
@@ -1,3 +1,4 @@
+from functools import update_wrapper, wraps
 import sys
 import types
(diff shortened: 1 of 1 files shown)


hauntsaninja:
Please read the README to understand what this project is about.

Issue #10657, opened 2023-09-03
Rename `codecs._Encoder`, `codecs._Decoder`, `.encode` and `.decode` `input` argument


denballakh:
`input` argument of `.encode` and `.decode` methods shadows built-in function `input`, which might trigger linter checks and make code editor highlight this variable as built-in.

This shows that nobody is calling these methods using `input` keyword argument:
https://grep.app/search?q=decode%28input%3D&filter[lang][0]=Python
https://grep.app/search?q=encode%28input%3D&filter[lang][0]=Python

srittau:
I haven't investigates this, but we should probably just turn them into positional-only arguments.

sobolevn:
Yes, they are pos-only in CPython:

```python
>>> import codecs
>>> codecs.getencoder('utf8')
<built-in function utf_8_encode>
```

```c
_codecs.utf_8_encode
    str: unicode
    errors: str(accept={str, NoneType}) = None
    /
```

Also, docs do not specify the signature https://docs.python.org/3/library/codecs.html#codecs.getencoder

So, pos-only is our best bet.

PR #10713, opened 2023-09-15
Make all protocols' param names pos-only in codecs.pyi

sobolevn:
Closes https://github.com/python/typeshed/issues/10657


srittau:
Hmm, I'm reconsidering this in light of the primer hits and the fact that [the docs](https://docs.python.org/3/library/codecs.html) state things like this:

> class codecs.IncrementalEncoder(errors='strict')
> Constructor for an [IncrementalEncoder](https://docs.python.org/3/library/codecs.html#codecs.IncrementalEncoder) instance.
> 
> All incremental encoders must provide this constructor interface. They are free to add additional keyword arguments, but only the ones defined here are used by the Python codec registry.

> **encode**
> **decode**
> The stateless encoding and decoding functions. These must be functions or methods which have the same interface as the [encode()](https://docs.python.org/3/library/codecs.html#codecs.Codec.encode) and [decode()](https://docs.python.org/3/library/codecs.html#codecs.Codec.decode) methods of Codec instances (see [Codec Interface](https://docs.python.org/3/library/codecs.html#codec-objects)). The functions or methods are expected to work in a stateless mode.
> **incrementalencoder**
> **incrementaldecoder**
> Incremental encoder and decoder classes or factory functions. These have to provide the interface defined by the base classes [IncrementalEncoder](https://docs.python.org/3/library/codecs.html#codecs.IncrementalEncoder) and [IncrementalDecoder](https://docs.python.org/3/library/codecs.html#codecs.IncrementalDecoder), respectively. Incremental codecs can maintain state.

srittau:
At least for the incremental coders. Since normal codecs are seemingly implemented with positional-only args in CPython, using pos-only protocols seems correct and able to catch errors.

srittau:
Suggestion: Change the normal codecs, but leave the incremental coders for now and wait what for the discussion in python/cpython#109444.

srittau (approved):

Issue #9537, opened 2023-01-14
A race condition in `mypy_test.py`


AlexWaygood:
#9408 (my PR) appears to have introduced a race condition in `mypy_test.py`.

## Evidence of the race

We've had several "flukey" errors in CI over the last few days, all with the same traceback:
- On Python 3.7: https://github.com/python/typeshed/actions/runs/3909559123/jobs/6682289730
- On Python 3.8:
  - https://github.com/python/typeshed/actions/runs/3904904286/jobs/6671274990
  - https://github.com/python/typeshed/actions/runs/3895029167/jobs/6649818621
- On Python 3.9: https://github.com/python/typeshed/actions/runs/3916447325/jobs/6695558555

<details>
<summary>The traceback is the same every time:</summary>

```pytb
Testing third-party packages...
Traceback (most recent call last):
  File "./tests/mypy_test.py", line 557, in <module>
    main()
  File "./tests/mypy_test.py", line 544, in main
    code, files_checked_this_version = test_typeshed(code, args=config, tempdir=td_path)
  File "./tests/mypy_test.py", line 525, in test_typeshed
    code, third_party_files_checked = test_third_party_stubs(code, args, tempdir)
  File "./tests/mypy_test.py", line 500, in test_third_party_stubs
    setup_virtual_environments(distributions_to_check, args, tempdir)
  File "./tests/mypy_test.py", line 444, in setup_virtual_environments
    requirements_set, venv_info = venv_info_future.result()
  File "/opt/hostedtoolcache/Python/3.7.15/x64/lib/python3.7/concurrent/futures/_base.py", line 428, in result
    return self.__get_result()
  File "/opt/hostedtoolcache/Python/3.7.15/x64/lib/python3.7/concurrent/futures/_base.py", line 384, in __get_result
    raise self._exception
  File "/opt/hostedtoolcache/Python/3.7.15/x64/lib/python3.7/concurrent/futures/thread.py", line 57, in run
    result = self.fn(*self.args, **self.kwargs)
  File "./tests/mypy_test.py", line 384, in setup_venv_for_external_requirements_set
    return requirements_set, make_venv(venv_dir)
  File "/home/runner/work/typeshed/typeshed/tests/utils.py", line 150, in make_venv
    venv.create(venv_dir, with_pip=True, clear=True)
  File "/opt/hostedtoolcache/Python/3.7.15/x64/lib/python3.7/venv/__init__.py", line 390, in create
    builder.create(env_dir)
[...]

PR #10714, opened 2023-09-15
Speedup `regr_test.py` by running test cases concurrently

AlexWaygood:
As we add more and more test cases, `regr_test.py` is getting kinda slow. Other than mypy_primer, it's now the slowest CI job we have by a long way (and we don't run mypy_primer on _all_ PRs -- for example, it's skipped on this PR!).

The reason for the slowness is that we now have regression tests for 11 stubs packages. We run all regression tests on Python 3.8-12 inclusive, and we run them all on linux, darwin and win32. That adds up for a total of 165 subprocesses that are created for each run of the test when it's run with `--all` (the flag we use in CI).

At some point we may want to consider sharding this test between GitHub Actions workers, similar to the way we run `mypy_test.py` and pyright in CI. (We can also possibly reconsider whether we _need_ to, e.g., run all tests on darwin, linux and Windows). For now, though, we can speed things up a _lot_ just by running the subprocess concurrently using a `ProcessPoolExecutor`. This cuts the execution time in CI roughly in half, from around 5-6 minutes to around 2-3 minutes.

(N.B.: A `ProcessPoolExecutor` feels like a slightly blunt instrument here with a lot of overhead; I'm sure there are more efficient ways of spawning subprocesses concurrently. I got this to work reasonably quickly, however, and when I tried different approaches I quickly ran into race conditions. I think this is ~good enough for now.)

### DIFF
--- a/tests/regr_test.py
+++ b/tests/regr_test.py
@@ -33,7 +33,6 @@ from utils import (
 )
 
 ReturnCode: TypeAlias = int
-PrintQueue: TypeAlias = queue.Queue[str]
 
 TEST_CASES = "test_cases"
 VENV_DIR = ".venv"
--- a/tests/regr_test.py
+++ b/tests/regr_test.py
@@ -4,12 +4,18 @@
 from __future__ import annotations
 
 import argparse
+import concurrent.futures
+import multiprocessing
 import os
+import queue
 import re
 import shutil
 import subprocess
 import sys
 import tempfile
+import threading
+from contextlib import ExitStack, suppress
+from dataclasses import dataclass
 from enum import IntEnum
 from itertools import product
 from pathlib import Path
@@ -24,11 +30,11 @@ from utils import (
     get_mypy_req,
     make_venv,
     print_error,
-    print_success_msg,
     testcase_dir_from_package_name,
 )
 
 ReturnCode: TypeAlias = int
+PrintQueue: TypeAlias = queue.Queue[str]
 
 TEST_CASES = "test_cases"
 VENV_DIR = ".venv"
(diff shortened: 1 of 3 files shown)


AlexWaygood:
Example of a successful CI run: https://github.com/python/typeshed/actions/runs/6199279959/job/16831483816

Example of a failing CI run: https://github.com/python/typeshed/actions/runs/6199348575/job/16831709368?pr=10714

AlexWaygood:
Here's an alternative approach that uses `asyncio` to create the concurrent subprocesses rather than a `ProcessPoolExecutor`: https://github.com/python/typeshed/compare/main...AlexWaygood:typeshed:optimise-regrtest-async?expand=1.

It seems to work fine, without race conditions, and is a similar approach to what `mypy_primer` does. But it ends up being more code, and doesn't seem any faster in practice. So I'm inclined to stick with the current approach.

JelleZijlstra:
Any specific reason to use processes rather than threads here? Seems like the workers are mostly waiting on external processes, so the GIL shouldn't be much of an issue. I tend to avoid multiprocessing as it's more prone to failing in weird ways than threads are.

AlexWaygood:
I initially tried a ThreadPoolExecutor, but quickly ran into weird race conditions where some of the mypy subprocesses couldn't find various stdlib modules. But I'll take another look and see if I can make threads work here. I agree they make more sense.

AlexWaygood:
I couldn't repro the race conditions anymore when I tried to, so I switched to a `ThreadPoolExecutor`. I'm a _tiny_ bit wary of this after our experiences in #9537, but threads do feel like a better fit here, so let's see how it goes.

JelleZijlstra (approved):

AlexWaygood:
Thanks @JelleZijlstra!

tests/regr_test.py
```diff
@@ -202,39 +203,103 @@ def run_testcases(
...
+@dataclass(frozen=True)
+class Result:
+    code: int
+    command_run: str
+    stderr: str
+    stdout: str
+    test_case_dir: Path
+    tempdir: Path
+
+    def print_description(self, *, verbosity: Verbosity) -> None:
```
Avasam, line 226: @AlexWaygood After this refactor, the `verbosity` argument of `Result.print_description` has been left unused. Was that intentional ?

AlexWaygood: Nope!

PR #10717, opened 2023-09-16
Update protobufs to support new generated messages

euroelessar:
Example of the generated protobuf file which would benefit from the change: https://github.com/nipunn1313/mypy-protobuf/blob/7f4a558c00faf8fac0cd6d7a6d1332d1643cc08c/mypy_protobuf/extensions_pb2.py#L19-L22


hauntsaninja (approved):
Looks good!

PR #10719, opened 2023-09-16
removed type_check_only marker from EllipsisType in python >=3.10

### LINKED ISSUE #10718: EllipsisType marked as `type_check_only` even for python ≥3.10
Typeshed defines: https://github.com/python/typeshed/blob/d88205caeeb9fba0e8c7cc021a05207e886ac706/stdlib/types.pyi#L630

and 

https://github.com/python/typeshed/blob/d88205caeeb9fba0e8c7cc021a05207e886ac706/stdlib/builtins.pyi#L1908-L1912

this leads to a problem in pyright, which raises a false positive for this code:

```python
from types import EllipsisType

x = ...
match x:
    case EllipsisType():  # ✘ error: "ellipsis" is marked as @type_check_only
        print(0)
```

cf. https://github.com/microsoft/pyright/issues/5954

I think there needs to be a case-distinction, since 3.10 added `types.EllipsisType`.

```python
if sys.version_info() < (3,10):
    # Actually the type of Ellipsis is <type 'ellipsis'>, but since it's
    # not exposed anywhere under that name, we make it private here.
    @final
    @type_check_only
    class ellipsis: ...
else:
    # 3.10 added types.EllipsisType
    @final
    class ellipsis: ...
```

randolf-scholz:
Fixes #10718

### DIFF
--- a/stdlib/builtins.pyi
+++ b/stdlib/builtins.pyi
@@ -1905,15 +1905,16 @@ def __import__(
 ) -> types.ModuleType: ...
 def __build_class__(__func: Callable[[], _Cell | Any], __name: str, *bases: Any, metaclass: Any = ..., **kwds: Any) -> Any: ...
 
-if sys.version_info < (3, 10):
+if sys.version_info >= (3, 10):
+    # In Python 3.10, EllipsisType is exposed publicly in the types module.
+    @final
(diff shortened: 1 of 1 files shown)


srittau (approved):
Thanks!

PR #10721, opened 2023-09-17
Add seaborn stubs

hamdanal:
Requires
- [x] https://github.com/typeshed-internal/stub_uploader/pull/102
- [x] https://github.com/typeshed-internal/stub_uploader/pull/103
- [x] https://github.com/python/typeshed/pull/10724


AlexWaygood:
For the `pytype_test.py` failures, we'll probably need to make some adjustments to this function:

https://github.com/python/typeshed/blob/cecab9c6eafb41e747ac1d8e1bf68a46cf52848d/tests/pytype_test.py#L152

If need be, we could probably just add some hacky special-casing for pandas-stubs to the function, e.g.

```python
if req_name == "pandas-stubs":
    return ["pandas"]
```

AlexWaygood:
The `mypy_test.py` failures are harder to fix :(

We may need to consider adding a `requires-python` field to our `METADATA.toml` files. If we did that, it would require also making some changes to the `stub_uploader` so that it includes the `requires-python` data in the `setup.py` files it generates

hamdanal:
> The `mypy_test.py` failures are harder to fix :(
> 
> We may need to consider adding a `requires-python` field to our `METADATA.toml` files, so we could declare that a stubs package only works with e.g. Python 3.9+. If we did that, it would enable us to skip running mypy with Python 3.8 for Seaborn in CI. But it would require also making some changes to the `stub_uploader` so that it includes the `requires-python` data in the `setup.py` files it generates.

Thanks for looking into this. Should we open an issue about this to see what other maintainers say?

AlexWaygood:
> > The `mypy_test.py` failures are harder to fix :(
> > We may need to consider adding a `requires-python` field to our `METADATA.toml` files, so we could declare that a stubs package only works with e.g. Python 3.9+. If we did that, it would enable us to skip running mypy with Python 3.8 for Seaborn in CI. But it would require also making some changes to the `stub_uploader` so that it includes the `requires-python` data in the `setup.py` files it generates.
> 
> Thanks for looking into this. Should we open an issue about this to see what other maintainers say?

Yes, that would be great!

AlexWaygood:
I've triggered a rerun of the stub-uploader tests now https://github.com/typeshed-internal/stub_uploader/pull/102 has been merged

AlexWaygood:
All green now 🥳 though I think we should also wait for https://github.com/typeshed-internal/stub_uploader/pull/103 before we can consider merging this

hamdanal:
I'll mark the PR as "Ready for review" because it is ready for review :P.
Maybe we can add a https://github.com/python/typeshed/labels/DO-NOT-MERGE or https://github.com/python/typeshed/labels/deferred label?

AlexWaygood:
https://github.com/typeshed-internal/stub_uploader/pull/103 was merged 🎉

JelleZijlstra:
It seems like seaborn has a lot of inline types, and these stubs closely track them. Have you considered asking whether they would ship a py.typed? I don't see any discussion of that in their issue tracker.

stubs/seaborn/seaborn/_core/scales.pyi
```diff
@@ -0,0 +1,100 @@
...
+from matplotlib.units import ConversionInterface
+from numpy.typing import ArrayLike, NDArray
+from pandas import Series
+from seaborn._core.typing import Default
+
+TransFuncs: TypeAlias = tuple[Callable[[ArrayLike], ArrayLike], Callable[[ArrayLike], ArrayLike]]
+Pipeline: TypeAlias = Sequence[Callable[[Any], Any] | None]
+
+class Scale:
+    values: tuple[Incomplete] | str | list[Incomplete] | dict[Incomplete, Incomplete] | None
```
JelleZijlstra, line 19:
```suggestion
    values: tuple[Incomplete, ...] | str | list[Incomplete] | dict[Incomplete, Incomplete] | None
```
Surely it doesn't accept only one-tuples.

stubs/seaborn/seaborn/_core/scales.pyi
```diff
@@ -0,0 +1,100 @@
...
+    def __post_init__(self) -> None: ...
+    def tick(self) -> Self: ...
+    def label(self) -> Self: ...
+    def __call__(self, data: Series[Any]) -> ArrayLike: ...
+
+@dataclass
+class Boolean(Scale):
+    values: tuple[Incomplete] | list[Incomplete] | dict[Incomplete, Incomplete] | None = None
+    def tick(self, locator: Locator | None = None) -> Self: ...  # type: ignore[override]
```
JelleZijlstra, line 28: hm, I don't see why this override is incompatible, and when I tried a variant of this code in mypy I didn't get any errors.

stubs/seaborn/seaborn/_core/scales.pyi
```diff
@@ -0,0 +1,100 @@
...
+        formatter: Formatter | None = None,
+        *,
+        like: str | Callable[[float], str] | None = None,
+        base: int | None | Default = ...,
+        unit: str | None = None,
+    ) -> Self: ...
+
+@dataclass
+class Temporal(ContinuousBase):
+    trans: ClassVar[Incomplete]  # not sure it is a classvar but the runtime has no annotation so it is not a dataclass field
```
JelleZijlstra, line 75: Seems like it's just unused? Might be better to omit it.

stubs/seaborn/seaborn/_core/typing.pyi
```diff
@@ -0,0 +1,24 @@
...
+VariableSpec: TypeAlias = ColumnName | Vector | None
+VariableSpecList: TypeAlias = list[VariableSpec] | Index[Any] | None
+DataSource: TypeAlias = DataFrame | Mapping[Hashable, Vector] | None
+OrderSpec: TypeAlias = Iterable[str] | None
+NormSpec: TypeAlias = tuple[float | None, float | None] | Normalize | None
+PaletteSpec: TypeAlias = str | list[Incomplete] | dict[Incomplete, Incomplete] | Colormap | None
+DiscreteValueSpec: TypeAlias = dict[Incomplete, Incomplete] | list[Incomplete] | None
+ContinuousValueSpec: TypeAlias = tuple[float, float] | list[float] | dict[Any, float] | None
+
+class Default: ...
```
JelleZijlstra, line 22: Any reason not to include Deprecated?

JelleZijlstra (commented):
Looks at the `_core` files. Overall this is mostly close to the types in the runtime library, though the runtime occasionally misses some annotations and your stubs do a better job there.

hamdanal:
> It seems like seaborn has a lot of inline types, and these stubs closely track them. Have you considered asking whether they would ship a py.typed? I don't see any discussion of that in their issue tracker.

seaborn has two APIs, the traditional _functional API_ which is totally untyped and the new _objects API_ which is mostly typed. The objects API is still under development and is incomplete so the functional API is still widely used and I don't see this changing soon. The author of seaborn has indicated before (I'll try to find the discussion) that they are opposed to adding type annotations to the functional API because its functions signatures are already large, hence this PR.

stubs/seaborn/seaborn/_core/typing.pyi
hamdanal: Looks like there is a new release where this was added :)
I'll update the PR.

stubs/seaborn/seaborn/_core/scales.pyi
hamdanal: It is used in the base class. It is also defined as a dataclass field in `Continuous` which is a sibling class to `Temporal`. So I'd say it must exist but it is weirdly defined.

stubs/seaborn/seaborn/axisgrid.pyi
```diff
@@ -0,0 +1,243 @@
...
+    diag_kind: Literal["auto", "hist", "kde"] | None = "auto",
+    markers: Incomplete | None = None,
+    height: float = 2.5,
+    aspect: float = 1,
+    corner: bool = False,
+    dropna: bool = False,
+    plot_kws: dict[str, Any] | None = None,
+    diag_kws: dict[str, Any] | None = None,
+    grid_kws: dict[str, Any] | None = None,
+    size: float | None = None,
```
JelleZijlstra, line 220: This param is deprecated, maybe add a comment

stubs/seaborn/seaborn/axisgrid.pyi
```diff
@@ -0,0 +1,243 @@
...
+) -> PairGrid: ...
+def jointplot(
+    data: Incomplete | None = None,
+    *,
+    x: Incomplete | None = None,
+    y: Incomplete | None = None,
+    hue: Incomplete | None = None,
+    kind: Literal["scatter", "kde", "hist", "hex", "reg", "resid"] = "scatter",
```
JelleZijlstra, line 228: The code does a bunch of `startswith` calls, so not sure we can use `Literal`.

stubs/seaborn/seaborn/categorical.pyi
```diff
@@ -0,0 +1,291 @@
...
+    hue_norm: Incomplete | None = None,
+    log_scale: _LogScale | None = None,
+    native_scale: bool = False,
+    formatter: Callable[..., str] | None = None,
+    legend: _Legend = "auto",
+    scale: Incomplete = ...,  # deprecated
+    box_kws: dict[str, Any] | None = None,
+    flier_kws: dict[str, Any] | None = None,
+    line_kws: dict[str, Any] | None = None,
+    ax: Axes | None = None,
```
JelleZijlstra, line 113: Also takes `**kwargs`, which get passed to `plot_kws`.
```suggestion
    ax: Axes | None = None,
    **kwargs: Any,
```

JelleZijlstra (commented):
Thanks for clarifying, indeed the files outside `_core` are mostly not annotated. I reviewed up to `cm.pyi` now, mostly by spot-checking and reading along with the implementation.

stubs/seaborn/seaborn/axisgrid.pyi
```diff
@@ -0,0 +1,243 @@
...
+    hue_names: list[str]
+    hue_vals: Series[Incomplete]
+    hue_kws: dict[str, Any]
+    palette: _RGBColorPalette
+    def __init__(
+        self,
+        data: DataFrame,
+        *,
+        hue: str | None = None,
+        vars: Iterable[str] | None = None,
```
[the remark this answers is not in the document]
hamdanal: Mmm your are right. It is however common to pass in numpy arrays or pandas index or series or what not. These cannot be typed as `Sequence[str]`. Either we keep `Iterable` and accept some false negatives (I expect them to be rare anyway) or we'll have to use a union of these like `Sequence[str] | NDArray[Any] | pd.Index | pd.Series[str]` (hopefully I didn't forget another common use).
I prefer the `Iterable` annotation but not enough to argue about it. Let me know which option you prefer.

stubs/seaborn/seaborn/axisgrid.pyi
```diff
@@ -0,0 +1,243 @@
...
+        x_vars: Iterable[str] | None = None,
```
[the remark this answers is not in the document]
hamdanal: These are documented as "lists of variable names" but I annotated them as iterables because they are casted to list at runtime. The `isscalar` check seems to allow `str` to be accepted but this is already covered by `Iterable[str]`. I think no change is needed here?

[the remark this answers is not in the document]
JelleZijlstra: Doesn't feel great to rely on the fact that `str` is an `Iterable[str]`, as usually that isn't what you want. As I recall pytype generally disallows str for Iterable[str]. So let's be explicit:
```suggestion
        x_vars: Iterable[str] | str | None = None,
```

stubs/seaborn/seaborn/axisgrid.pyi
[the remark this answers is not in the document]
JelleZijlstra: Let's just keep this as is then.

stubs/seaborn/seaborn/external/appdirs.pyi
```diff
@@ -0,0 +1,9 @@
+from typing_extensions import Literal
```
JelleZijlstra, line 1: This directory seems to contain vendored external libraries. Are they intended to be used by end users? If not, it may be better to omit them from the stubs.

stubs/seaborn/seaborn/external/husl.pyi
```diff
@@ -0,0 +1,42 @@
...
+def hex_to_husl(hex: str) -> list[float]: ...
+def huslp_to_rgb(h: float, s: float, l: float) -> list[float]: ...
+def huslp_to_hex(h: float, s: float, l: float) -> str: ...
+def rgb_to_huslp(r: float, g: float, b: float) -> list[float]: ...
+def hex_to_huslp(hex: str) -> list[float]: ...
+def lch_to_rgb(l: float, c: float, h: float) -> list[float]: ...
+def rgb_to_lch(r: float, g: float, b: float) -> list[float]: ...
+def max_chroma(L: float, H: float) -> float: ...
+def max_chroma_pastel(L: float) -> float: ...
+def dot_product(a: float, b: float) -> float: ...
```
JelleZijlstra, line 25: This seems to take two iterables of floats.

stubs/seaborn/seaborn/palettes.pyi
```diff
@@ -0,0 +1,149 @@
...
+]
+
+_T = TypeVar("_T")
+
+SEABORN_PALETTES: dict[str, list[str]]
+MPL_QUAL_PALS: dict[str, int]
+QUAL_PALETTE_SIZES: dict[str, int]
+QUAL_PALETTES: list[str]
+
+class _ColorPalette(list[_T]):
```
JelleZijlstra, line 30: Seems like the members get passed to `mpl.colors.rgb2hex`, which takes some sort of color objects, so I don't think this should be generic over an unbounded TypeVar.

stubs/seaborn/seaborn/palettes.pyi
```diff
@@ -0,0 +1,149 @@
...
+    "deep", "deep6", "muted", "muted6", "pastel", "pastel6", "bright", "bright6", "dark", "dark6", "colorblind", "colorblind6"
+]
+
+@overload
+def color_palette(  # type: ignore[misc]
+    palette: _SeabornPaletteName | None = None, n_colors: int | None = None, desat: float | None = None, *, as_cmap: Literal[True]
+) -> list[str]: ...  # this might be a bug in seaborn because we expect the return type to be a Colormap instance
+@overload
+def color_palette(
+    palette: str | Sequence[ColorType], n_colors: int | None = None, desat: float | None = None, *, as_cmap: Literal[True]
```
JelleZijlstra, line 46: Seems like if `desat` is given (and `as_cmap` is true) it returns a list, not a ColorMap.

stubs/seaborn/seaborn/palettes.pyi
```diff
@@ -0,0 +1,149 @@
...
+@overload
+def husl_palette(
+    n_colors: int = 6, h: float = 0.01, s: float = 0.9, l: float = 0.65, *, as_cmap: Literal[True]
+) -> ListedColormap: ...
+@overload
+def husl_palette(
+    n_colors: int = 6, h: float = 0.01, s: float = 0.9, l: float = 0.65, as_cmap: Literal[False] = False
+) -> _RGBColorPalette: ...
+@overload
+def mpl_palette(name: str, n_colors: int = 6, *, as_cmap: Literal[True]) -> LinearSegmentedColormap: ...
```
JelleZijlstra, line 72: Technically this and similar overloads on nearby functions are incomplete, as they don't cover calls like `mpl_palette("x", 1, True)`. You'd need another overload to cover that case.

JelleZijlstra (commented):
Got to the end this time! A few more comments.

stubs/seaborn/seaborn/external/appdirs.pyi
hamdanal: I see at least `seaborn.external.kde.gaussian_kde` class used as a type annotation elsewhere in the code. Should I only keep this class and delete other types in the `external` sub package? It is unlikely that their presence or absence will cause any trouble.

stubs/seaborn/seaborn/palettes.pyi
hamdanal: `desat` and `as_cmap` cannot be used together, you get `TypeError: 'ListedColormap' object is not iterable` at runtime if you pass both.

stubs/seaborn/seaborn/palettes.pyi
hamdanal: You are right. I deliberately left out the third overload here because I think in this case it is OK for the stub to be a little bit stricter than runtime and to expect a keyword argument especially that a usage like `mpl_palette("x", 1, True)` feels cryptic. But if you prefer I can add the missing overload.

stubs/seaborn/seaborn/external/appdirs.pyi
JelleZijlstra: Let's just keep it.

stubs/seaborn/seaborn/palettes.pyi
JelleZijlstra: OK, let's just keep it as is as we can always change it later. Agree these really should be keyword-only.

JelleZijlstra:
Thanks for your work @hamdanal!

Issue #10722, opened 2023-09-17
Stubs that require a version of python higher than the minimum supported by typeshed


hamdanal:
In #10721, the seaborn stubs depend on matplotlib that only included a `py.typed` marker file in version 3.8.
matplotlib version 3.8 requires python `>=3.9` which leads to the mypy tests failing on python 3.8.

Alex Waygood suggests adding a `requires-python` field to `METADATA.toml` and teaching stub-uploader to include it in the generated `setup.py` https://github.com/python/typeshed/pull/10721#issuecomment-1722454399.

Please let us know what you think about this idea or if you have another solution.

AlexWaygood:
I think this would be a good idea.

We actually already have a need for this: typeshed claims to support Python 3.7+, but the latest version of `types-jsonschema` will only work with mypy on Python 3.8+. The only reason why this isn't causing our CI to fail is that we no longer run mypy in CI on Python 3.7, because the latest version of mypy only supports Python 3.8+. But that isn't particularly principled; we should signal  to our users that the stubs package we're shipping only works with Python 3.8+, by including that information in the generated `setup.py` file.

AlexWaygood:
Oh, actually, the types-jsonschema PR I was thinking of hasn't been merged yet... the package _will_ require Python 3.8+ if #10583 is merged :-)

AlexWaygood:
The METADATA.toml field should be optional — the default assumption should be that most stubs packages can be used with all versions of Python that typeshed supports.

We should add some validation to `parse_metadata.py` to ensure that we don't have any `requires-python` fields that specify a minimum Python version less than the oldest version of Python supported by typeshed.

I'm not sure what the stub uploader should do if it encounters a stubs package that doesn't have a `requires-python` field in `METADATA.toml`. Should it add `requires-python` to the generated `setup.py` file anyway, and set it to the lowest version of Python supported by typeshed? Or should it just not add anything to the `setup.py` file?

AlexWaygood:
> I'm not sure what the stub uploader should do if it encounters a stubs package that doesn't have a `requires-python` field in `METADATA.toml`. Should it add `requires-python` to the generated `setup.py` file anyway, and set it to the lowest version of Python supported by typeshed? Or should it just not add anything to the `setup.py` file?

[...]

PR #10724, opened 2023-09-17
Add optional requires_python to third party stubs metadata

hamdanal:
The typeshed part of #10722

This PR implements the following:
- parse an optional `requires_python` field in the third party stubs
- the field must be a valid Python version specifier that specifies the minimum Python version supported by the stub
- the version must be at least the minimum version supported by typeshed (currently 3.7)
- if the field exists, mypy, regression, and stubtest tests are skipped on Python versions that do not satisfy the requirement

stub-uploader still needs to learn about this field and include it in the `setup.py` file of the corresponding package with a fallback to the minimal version supported by typeshed.


AlexWaygood:
`regr_test.py` should also probably be updated so that it doesn't run the test cases for a package that has `requires_python >= 3.9` and `--python-version 3.8` is passed to the script.

It wouldn't be as easy to adjust pyright/pytype, but I also don't think it'll be as important to fix those tests in a "principled" way; we can probably leave them as they are for now

AlexWaygood:
The changes to `regr_test.py` can probably be deferred to a followup PR, though; we don't need to do it all together

hamdanal:
> It wouldn't be as easy to adjust pyright/pytype, but I also don't think it'll be as important to "fix" those tests in a "principled" way; we can probably leave them as they are for now

Okay. I'll leave them for later

> The changes to `regr_test.py` can probably be deferred to a followup PR, though; we don't need to do it all together

Let me see if it is simple enough to do here, if not I'll leave it as well

tests/parse_metadata.py
```diff
@@ -240,6 +246,16 @@ def read_metadata(distribution: str) -> StubMetadata:
...
+    requires_python_str: object = data.get("requires_python")
+    if requires_python_str is None:
+        requires_python = None
+    else:
+        assert isinstance(requires_python_str, str)
+        requires_python = Specifier(requires_python_str)
+        # Check minimum Python version is not less than the oldest version of Python supported by typeshed
+        assert Specifier(f">={OLDEST_SUPPORTED_PYTHON}").contains(
+            requires_python.version
+        ), f"'requires_python' contains versions lower than the oldest supported Python {OLDEST_SUPPORTED_PYTHON}"
```
hamdanal, line 259: Should we add a check that `requires-python` only specifies the minimum version? Something like
```python
assert requires_python.operator == ">="
```

AlexWaygood: Sounds reasonable to me

tests/regr_test.py
```diff
@@ -254,6 +254,12 @@ def main() -> ReturnCode:
 
     code = 0
     for testcase_dir in testcase_directories:
+        assert isinstance(testcase_dir, PackageInfo)
+        if not testcase_dir.is_stdlib:
+            metadata = read_metadata(testcase_dir.name)
+            if metadata.requires_python and not metadata.requires_python.contains(PYTHON_VERSION):
```
AlexWaygood, line 260: Same comment here as in `mypy_test.py`: I think you should be looking at the `versions_to_test` local variable here rather than `PYTHON_VERSION`.

AlexWaygood (commented):
Lots of good stuff here, but needs a few changes -- see my comments below:

tests/parse_metadata.py
```diff
@@ -30,6 +32,8 @@
     "read_stubtest_settings",
 ]
 
+OLDEST_SUPPORTED_PYTHON: Final = "3.7"
```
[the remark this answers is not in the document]
hamdanal: TBH I haven't looked at the stub-uploader code yet. But I like the idea of defining the constant in pyproject.toml, I'll update the PR.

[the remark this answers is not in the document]
Avasam: For later, https://github.com/python/typeshed/blob/main/scripts/runtests.py#L31 could also use the specified stub's oldest python version instead of the common oldest supported overall.

tests/regr_test.py
```diff
@@ -254,6 +254,12 @@ def main() -> ReturnCode:
 
     code = 0
     for testcase_dir in testcase_directories:
+        assert isinstance(testcase_dir, PackageInfo)
```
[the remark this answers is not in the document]
hamdanal: No but it is on my list of things to investigate when I have time. It could be a bug in mypy.

tests/parse_metadata.py
```diff
@@ -240,6 +246,17 @@ def read_metadata(distribution: str) -> StubMetadata:
...
+    requires_python_str: object = data.get("requires_python")
+    if requires_python_str is None:
+        requires_python = None
+    else:
+        assert isinstance(requires_python_str, str)
+        requires_python = Specifier(requires_python_str)
+        # Check minimum Python version is not less than the oldest version of Python supported by typeshed
+        assert Specifier(f">={OLDEST_SUPPORTED_PYTHON}").contains(
+            requires_python.version
+        ), f"'requires_python' contains versions lower than the oldest supported Python {OLDEST_SUPPORTED_PYTHON}"
```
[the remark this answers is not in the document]
hamdanal: Interesting idea. This only moves the error one year. If we keep the current implementation, we will get an error when typeshed drops **the version** specified. If we implement your suggestion, we get an error when typeshed drops all versions **below** the version specified. I am +0 on the change, let me know what you prefer.

tests/mypy_test.py
```diff
@@ -480,6 +482,11 @@ def test_third_party_stubs(code: int, args: TestConfig, tempdir: Path) -> TestRe
         if spec_matches_path(gitignore_spec, distribution_path):
             continue
 
+        dist_metadata = read_metadata(distribution)
+        if dist_metadata.requires_python and not dist_metadata.requires_python.contains(PYTHON_VERSION):
```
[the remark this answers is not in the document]
hamdanal: Now I think we should look at both `args.version` and `parse_metadata.PYTHON_VERSION`. We want to skip the test either when we are running against an unsupported version, or because mypy and the package and its dependencies need to be installed on an supported version.

tests/stubtest_third_party.py
```diff
@@ -37,6 +37,12 @@ def run_stubtest(
             return True
         print(colored(f"Note: {dist_name} is not currently tested on {sys.platform} in typeshed's CI.", "yellow"))
 
+    if metadata.requires_python:
+        if not metadata.requires_python.contains(PYTHON_VERSION):
+            print(colored(f"skipping (requires Python {metadata.requires_python})", "yellow"))
+            return True
+        print(colored(f"Note: {dist_name} requires Python {metadata.requires_python}.", "yellow"))
```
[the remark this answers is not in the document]
hamdanal: Yes. I just copied the logic of the platform check two lines above. But yeah it is not important I can delete it

tests/mypy_test.py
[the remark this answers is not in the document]
AlexWaygood: Hmm, good point. Your logic makes sense to me. Would be nice, if possible, to tailor the message printed to the terminal to explain exactly why we're not checking these stubs in this specific situation, though

tests/parse_metadata.py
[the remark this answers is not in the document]
AlexWaygood: Cool, let's go with my suggestion 👍 I feel like it's in keeping with the ridiculous pedantry of some of the checks we already have:

https://github.com/python/typeshed/blob/53422f8918debdaf60e1605d0e2e3ce77a810afd/tests/parse_metadata.py#L197

tests/mypy_test.py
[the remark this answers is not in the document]
hamdanal: BTW in your example, you used `requires_python = ">=3.8"` and `python tests/mypy_test.py -p3.8 stubs/requests`. I expect it not to skip the test in this case. Maybe you meant `python tests/mypy_test.py -p3.7 stubs/requests`?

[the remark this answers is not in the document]
AlexWaygood: > Maybe you meant `python tests/mypy_test.py -p3.7 stubs/requests`

yes :)

tests/parse_metadata.py
```diff
@@ -28,8 +30,13 @@
     "read_dependencies",
     "read_metadata",
     "read_stubtest_settings",
+    "OLDEST_SUPPORTED_PYTHON",
+    "PYTHON_VERSION",
 ]
 
+with open("pyproject.toml", "rb") as config:
+    OLDEST_SUPPORTED_PYTHON: Final[str] = tomli.load(config)["tool"]["typeshed"]["oldest_supported_python"]
```
hamdanal, line 38: `OLDEST_SUPPORTED_PYTHON` is only used in the `read_metadata` function. Should I move it there?

AlexWaygood: That feels like it would be inefficient (we'd have to reload it from the `pyproject.toml` file every time `read_metadata` was called with a different argument). But I think I'd prefer to do something like this -- not massively keen on files being opened and read from just as a side effect of importing a module:

```py
@cache
def _get_oldest_supported_python() -> str:
    with open("pyproject.toml", "rb") as config:
        val = tomli.load(config)["tool"]["typeshed"]["oldest_supported_python"]
    assert type(val) is str
    return val
```

And then in `read_metadata()` you can just do

```py
oldest_supported_python = _get_oldest_supported_python()
```

AlexWaygood: We could possibly make `_get_oldest_supported_python()` public at a later date if we find it's useful from other files in our `tests/` directory, like @Avasam suggests

tests/parse_metadata.py
```diff
@@ -30,6 +32,8 @@
     "read_stubtest_settings",
 ]
 
+OLDEST_SUPPORTED_PYTHON: Final = "3.7"
+PYTHON_VERSION: Final = f"{sys.version_info.major}.{sys.version_info.minor}"
```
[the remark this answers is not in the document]
AlexWaygood: Hmm... on second thoughts... maybe this constant really belongs in `utils.py`, rather than `parse_metadata.py`? This file already depends on `utils.py`, so I don't think importing it from `utils.py` would be a problem

tests/parse_metadata.py
```diff
@@ -240,6 +254,20 @@ def read_metadata(distribution: str) -> StubMetadata:
     assert type(uploaded_to_pypi) is bool
     partial_stub: object = data.get("partial_stub", True)
     assert type(partial_stub) is bool
+    requires_python_str: object = data.get("requires_python")
+    oldest_supported_python = _get_oldest_supported_python()
+    oldest_supported_python_specifier = Specifier(f">={oldest_supported_python}")
+    if requires_python_str is None:
+        requires_python = oldest_supported_python_specifier
+    else:
+        assert isinstance(requires_python_str, str)
```
AlexWaygood, line 260: I prefer to be extremely strict about the validation we do in this file: I don't know how a `str` subclass could turn up here, so something pretty scary is happening if it does. I'd like to know about that if it happens! (It won't. But the point stands.)

```suggestion
        assert type(requires_python_str) is str
```

tests/parse_metadata.py
```diff
@@ -240,6 +251,20 @@ def read_metadata(distribution: str) -> StubMetadata:
...
+        requires_python = Specifier(requires_python_str)
+        assert requires_python != oldest_supported_python_specifier, f'requires_python="{requires_python}" is redundant'
+        # Check minimum Python version is not less than the oldest version of Python supported by typeshed
+        assert oldest_supported_python_specifier.contains(
+            requires_python.version
+        ), f"'requires_python' contains versions lower than the oldest supported Python ({oldest_supported_python})"
```
AlexWaygood, line 266:
```suggestion
        ), f"'requires_python' contains versions lower than typeshed's oldest supported Python ({oldest_supported_python})"
```

AlexWaygood:
With this PR checked out, I tried applying this diff:

```diff
diff --git a/stubs/requests/METADATA.toml b/stubs/requests/METADATA.toml
index 39371b90e..677799c7d 100644
--- a/stubs/requests/METADATA.toml
+++ b/stubs/requests/METADATA.toml
@@ -1,6 +1,7 @@
 version = "2.31.*"
 upstream_repository = "https://github.com/psf/requests"
 requires = ["types-urllib3"]
+requires_python = ">=3.9"

 [tool.stubtest]
 extras = ["socks"]
```

I then ran this command:

```
python tests/mypy_test.py stubs/requests
```

Output:

```pytb
*** Testing Python 3.12 on win32
testing requests (17 files)... success

testing requests (17 files)... success

*** Testing Python 3.10 on win32
testing requests (17 files)... success

*** Testing Python 3.9 on win32
testing requests (17 files)... success

*** Testing Python 3.8 on win32
skipping 'requests' for target Python 3.8 (requires Python >=3.9)
  File "C:\Users\alexw\coding\typeshed\tests\mypy_test.py", line 573, in <module>
    main()
  File "C:\Users\alexw\coding\typeshed\tests\mypy_test.py", line 556, in main
    code, files_checked_this_version, packages_skipped_this_version = test_typeshed(code, args=config, tempdir=td_path)
                                                                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\alexw\coding\typeshed\tests\mypy_test.py", line 535, in test_typeshed
    code, third_party_files_checked, third_party_packages_skipped = test_third_party_stubs(code, args, tempdir)
                                                                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\alexw\coding\typeshed\tests\mypy_test.py", line 511, in test_third_party_stubs
    assert _DISTRIBUTION_TO_VENV_MAPPING.keys() == distributions_to_check.keys()
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError
```

tests/mypy_test.py
```diff
@@ -531,19 +548,24 @@ def main() -> None:
...
-            code, files_checked_this_version = test_typeshed(code, args=config, tempdir=td_path)
+            code, files_checked_this_version, packages_skipped_this_version = test_typeshed(code, args=config, tempdir=td_path)
             total_files_checked += files_checked_this_version
+            total_packages_skipped += packages_skipped_this_version
     if code:
         print_error(f"--- exit status {code}, {total_files_checked} files checked ---")
         sys.exit(code)
-    if not total_files_checked:
+    if total_packages_skipped:
+        print(colored(f"--- {total_packages_skipped} packages skipped ---", "yellow"))
```
AlexWaygood, line 563: Grammar nit:

```suggestion
        plural = "" if total_packages_skipped == 1 else "s"
        print(colored(f"--- {total_packages_skipped} package{plural} skipped ---", "yellow"))
```

Could possibly do similar for the `total_files_checked` print() immediately below, as well :)

tests/mypy_test.py
```diff
@@ -480,6 +483,18 @@ def test_third_party_stubs(code: int, args: TestConfig, tempdir: Path) -> TestRe
         if spec_matches_path(gitignore_spec, distribution_path):
             continue
 
+        metadata = read_metadata(distribution)
+        if not metadata.requires_python.contains(PYTHON_VERSION):
+            msg = f"skipping {distribution!r} on Python {PYTHON_VERSION} (requires Python {metadata.requires_python})"
```
AlexWaygood, line 488: Maybe make it slightly clearer that the issue is the Python version the test is being _run_ with, not the Python version passed on the command line:

```suggestion
            msg = f"skipping {distribution!r} (requires Python {metadata.requires_python}; test is being run using Python {PYTHON_VERSION})"
```

tests/regr_test.py
```diff
@@ -250,13 +251,27 @@ def main() -> ReturnCode:
...
+        versions_to_test = args.versions_to_test or [PYTHON_VERSION]
 
     code = 0
     for testcase_dir in testcase_directories:
+        assert isinstance(testcase_dir, PackageInfo)
+        metadata = None
+        if not testcase_dir.is_stdlib:
+            metadata = read_metadata(testcase_dir.name)
+            if not metadata.requires_python.contains(PYTHON_VERSION):
+                msg = f"skipping {testcase_dir.name!r} on Python {PYTHON_VERSION} (requires Python {metadata.requires_python})"
```
AlexWaygood, line 263:
```suggestion
                msg = f"skipping {testcase_dir.name!r} (requires Python {metadata.requires_python}; test is being run using Python {PYTHON_VERSION})"
```

AlexWaygood:
Running this:

```
python tests/regr_test.py requests -p3.8
```

Results in this:

```
skipping 'requests' for target Python 3.8 (requires Python >=3.9)

Test completed successfully!
```

Could we have it print something like "All tests were skipped!" to the terminal, if all tests were skipped?

tests/mypy_test.py
hamdanal: and above :)

tests/mypy_test.py
```diff
@@ -493,34 +511,48 @@ def test_third_party_stubs(code: int, args: TestConfig, tempdir: Path) -> TestRe
...
+    }
+    if distributions_without_venv:
+        setup_virtual_environments(distributions_without_venv, args, tempdir)
+
+    # Check that there is a venv for every distribution we're testing.
+    # Some venvs may exist from previous runs but are skipped in this run.
+    assert _DISTRIBUTION_TO_VENV_MAPPING.keys() >= distributions_to_check.keys()
+
+    for distribution in distributions_to_check:
+        venv_info = _DISTRIBUTION_TO_VENV_MAPPING[distribution]
```
hamdanal, line 528: OK so this should work now in all cases (hopefully). I tested with your requests example and all these invocations work
- `python tests/mypy_test.py stubs/requests`
- `python tests/mypy_test.py stubs/requests*`
- `python tests/mypy_test.py stubs/requests -p 3.8 3.9 3.10 3.11`
- `python tests/mypy_test.py stubs/requests -p 3.11 3.8 3.9 3.10`  # this also crashed before, order is important
- `python tests/mypy_test.py stubs/requests* -p 3.8 3.9 3.10 3.11`  # multiple packages together
- `python tests/mypy_test.py stubs/requests* -p 3.11 3.8 3.9 3.10`
Also the for loop now correctly skips distributions that should not run. It must loop over `distributions_to_check`, not venvs

tests/regr_test.py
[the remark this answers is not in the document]
hamdanal: So this turned out to be a mypy bug with `NamedTuple | Any` that's already fixed on master https://mypy-play.net/?mypy=master&python=3.11&gist=70ab043c7071876a2f03091afa14d3c8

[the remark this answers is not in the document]
AlexWaygood: Very bizarre, but cool that it's already fixed! In _that_ case, maybe a better fix for us here would be to `type: ignore` the mypy complaints away? Then we'll get an "Unused type: ignore" message from mypy when we next bump our mypy pin.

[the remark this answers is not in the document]
hamdanal: Bizarre error indeed. I added the type ignores

tests/regr_test.py
```diff
@@ -250,19 +251,40 @@ def main() -> ReturnCode:
...
     else:
         platforms_to_test = args.platforms_to_test or [sys.platform]
-        versions_to_test = args.versions_to_test or [f"3.{sys.version_info[1]}"]
+        versions_to_test = args.versions_to_test or [PYTHON_VERSION]
 
     code = 0
+    completed_tests = 0
     for testcase_dir in testcase_directories:
+        metadata = None
+        if not testcase_dir.is_stdlib:  # type: ignore[misc]
```
AlexWaygood, line 260:
```suggestion
        if not testcase_dir.is_stdlib:  # type: ignore[misc]  # mypy bug, already fixed on master
```

tests/regr_test.py
```diff
@@ -250,19 +251,40 @@ def main() -> ReturnCode:
...
+                msg = (
+                    f"skipping {testcase_dir.name!r} (requires Python {metadata.requires_python}; "
+                    f"test is being run using Python {PYTHON_VERSION})"
+                )
+                print(colored(msg, "yellow"))
+                continue
         with tempfile.TemporaryDirectory() as td:
             tempdir = Path(td)
             for platform, version in product(platforms_to_test, versions_to_test):
+                if not testcase_dir.is_stdlib:  # type: ignore[misc]
```
AlexWaygood, line 272:
```suggestion
                if not testcase_dir.is_stdlib:  # type: ignore[misc]  # mypy bug, already fixed on master
```

AlexWaygood:
Oh, shoot, I just caused you merge conflicts with #10714 -- sorry :(

hamdanal:
> Oh, shoot, I just caused you merge conflicts with #10714 -- sorry :(
> 
> I was planning on merging this first and then fixing the conflicts in my PR, but I forgot :(

No problem, I fixed them.

I removed the `product` loop and used separate for loops. This allows us to minimize the number of version checks we do and the number of messages we print for skipped tests. Now we print one "skipped" message per version only.

hamdanal:
@AlexWaygood WDYT about this patch?
```diff
diff --git a/tests/parse_metadata.py b/tests/parse_metadata.py
index dfb266ad0..9977db339 100644
--- a/tests/parse_metadata.py
+++ b/tests/parse_metadata.py
@@ -7,7 +7,7 @@ from __future__ import annotations
 import os
 import re
 import urllib.parse
-from collections.abc import Mapping
+from collections.abc import Callable, Mapping
 from dataclasses import dataclass
 from pathlib import Path
 from typing import NamedTuple
@@ -18,7 +18,7 @@ from packaging.requirements import Requirement
 from packaging.specifiers import Specifier
 from packaging.version import Version
 
-from utils import cache
+from utils import PYTHON_VERSION, cache, colored
 
 __all__ = [
     "NoSuchStubError",
@@ -129,6 +129,7 @@ class StubMetadata:
     Don't construct instances directly; use the `read_metadata` function.
     """
 
+    name: str
     version: str
     requires: Annotated[list[str], "The raw requirements as listed in METADATA.toml"]
     extra_description: str | None
@@ -141,6 +142,21 @@ class StubMetadata:
     stubtest_settings: StubtestSettings
     requires_python: Annotated[Specifier, "Versions of Python supported by the stub package"]
 
+    def skip_test(self, version: str | None = None, printer: Callable[[str], object] = print) -> bool:
+        """Return True if the stub should be skipped.
+
+        If version is None, the runtime Python version is used.
+        """
+        if self.requires_python.contains(version if version is not None else PYTHON_VERSION):
+            return False
+        msg = f"skipping {self.name!r} "
+        if version is not None:
+            msg += f"for target Python {version} (requires Python {self.requires_python})"
+        else:
+            msg += f"(requires Python {self.requires_python}; test is being run using Python {PYTHON_VERSION})"
+        printer(colored(msg, "yellow"))
+        return True
+
 
 _KNOWN_METADATA_FIELDS: Final = frozenset(
     {
@@ -277,6 +293,7 @@ def read_metadata(distribution: str) -> StubMetadata:
             assert key in tk, f"Unrecognised {tool} key {key!r} for {distribution!r}"
 
     return StubMetadata(
+        name=distribution,
         version=version,
         requires=requires,
         extra_description=extra_description,
```
since we are doing this check in a lot of places. Then we can simply do:
```python
if metadata.skip_test():
    continue
```
to check against runtime Python version or
```python
if metadata.skip_test(version):
    continue
```
to check against target version from the command line

AlexWaygood:
> @AlexWaygood WDYT about this patch?

Hmm, not sure -- I think I prefer it as you currently have it. That patch leads to more shared code, it's true, but it feels strange to me to be printing stuff to the terminal from `parse_metadata.py`. The idea of that module is that it contains common logic to help abstract parsing the `METADATA.toml` files and provide validation for security purposes. I feel like those functions should be ignorant of how they're being used and of things like the specific Python version they're being run on.

I wouldn't be opposed to adding a little helper method, though, e.g.

```py
class StubMetadata:
    def should_skip_tests(py_version: str) -> bool:
        return self.requires_python.contains(version)
```

tests/mypy_test.py
```diff
@@ -531,19 +563,28 @@ def main() -> None:
...
             total_files_checked += files_checked_this_version
+            total_packages_skipped += packages_skipped_this_version
     if code:
-        print_error(f"--- exit status {code}, {total_files_checked} files checked ---")
+        plural = "" if total_files_checked == 1 else "s"
+        print_error(f"--- exit status {code}, {total_files_checked} file{plural} checked ---")
         sys.exit(code)
-    if not total_files_checked:
+    if total_packages_skipped:
+        {1: ""}.get(total_packages_skipped, "s")
```
AlexWaygood, line 579: What does this line do?

tests/regr_test.py
AlexWaygood, line 1: Could we maybe do something like this? I'd prefer to keep the block of code inside the `ThreadPoolExecutor()` context manager fairly small, so that it's really easy to see exactly which bit of the script needs the multithreading. Filtering which test cases need to be run can be done before we enter that block.

```diff
diff --git a/tests/regr_test.py b/tests/regr_test.py
index 3fb7a8656..e47e597b4 100755
--- a/tests/regr_test.py
+++ b/tests/regr_test.py
@@ -13,9 +13,11 @@ import subprocess
 import sys
 import tempfile
 import threading
+from collections.abc import Callable
 from contextlib import ExitStack, suppress
 from dataclasses import dataclass
 from enum import IntEnum
+from functools import partial
 from pathlib import Path
 from typing_extensions import TypeAlias

@@ -275,6 +277,34 @@ def concurrently_run_testcases(
         package_info: Path(stack.enter_context(tempfile.TemporaryDirectory())) for package_info in testcase_directories
     }

+    to_do: list[Callable[[], Result]] = []
+    for testcase_dir, tempdir in packageinfo_to_tempdir.items():
+        metadata = None
+        if not testcase_dir.is_stdlib:  # type: ignore[misc]  # mypy bug, already fixed on master
+            metadata = read_metadata(testcase_dir.name)
+            if not metadata.requires_python.contains(PYTHON_VERSION):
+                msg = (
+                    f"skipping {testcase_dir.name!r} (requires Python {metadata.requires_python}; "
+                    f"test is being run using Python {PYTHON_VERSION})"
+                )
+                print(colored(msg, "yellow"))
+            if not metadata.requires_python.contains(PYTHON_VERSION):
+                msg = (
+                    f"skipping {testcase_dir.name!r} (requires Python {metadata.requires_python}; "
+                    f"test is being run using Python {PYTHON_VERSION})"
+                )
+                print(colored(msg, "yellow"))
+                continue
+        for version in versions_to_test:
+            if not testcase_dir.is_stdlib:  # type: ignore[misc]  # mypy bug, already fixed on master
+                assert metadata is not None
+                if not metadata.requires_python.contains(version):
+                    msg = (
+                        f"skipping {testcase_dir.name!r} for target Python {version} (requires Python {metadata.requires_python})"
+                    )
+                    print(colored(msg, "yellow"))
+                    continue
+            to_do.extend(
+                [
+                    partial(test_testcase_directory, testcase_dir, version, platform, verbosity=verbosity, tempdir=tempdir)
+                    for platform in platforms_to_test
+                ]
+            )
+
     event = threading.Event()
     printer_thread = threading.Thread(target=print_queued_messages, args=(event,))
     printer_thread.start()
@@ -289,31 +319,7 @@ def concurrently_run_testcases(
         ]
         concurrent.futures.wait(testcase_futures)

-        mypy_futures: list[concurrent.futures.Future[Result]] = []
-        for testcase_dir, tempdir in packageinfo_to_tempdir.items():
-            metadata = None
-            if not testcase_dir.is_stdlib:  # type: ignore[misc]  # mypy bug, already fixed on master
-                metadata = read_metadata(testcase_dir.name)
-                if not metadata.requires_python.contains(PYTHON_VERSION):
-                    msg = (
-                        f"skipping {testcase_dir.name!r} (requires Python {metadata.requires_python}; "
-                        f"test is being run using Python {PYTHON_VERSION})"
-                    )
-                    _PRINT_QUEUE.put(colored(msg, "yellow"))
-                    continue
-            for version in versions_to_test:
-                if not testcase_dir.is_stdlib:  # type: ignore[misc]  # mypy bug, already fixed on master
-                    assert metadata is not None
-                    if not metadata.requires_python.contains(version):
-                        msg = f"skipping {testcase_dir.name!r} for target Python {version} (requires Python {metadata.requires_python})"
-                        _PRINT_QUEUE.put(colored(msg, "yellow"))
-                        continue
-                for platform in platforms_to_test:
-                    mypy_futures.append(
-                        executor.submit(
-                            test_testcase_directory, testcase_dir, version, platform, verbosity=verbosity, tempdir=tempdir
-                        )
-                    )
+        mypy_futures = [executor.submit(task) for task in to_do]
         results = [future.result() for future in mypy_futures]
```

tests/regr_test.py
```diff
@@ -315,15 +334,20 @@ def main() -> ReturnCode:
...
+
     results: list[Result] | None = None
 
     with ExitStack() as stack:
         results = concurrently_run_testcases(stack, testcase_directories, verbosity, platforms_to_test, versions_to_test)
 
     assert results is not None
+    if not results:
+        print(colored("All tests were skipped!", "yellow"))
+        return 0
```
AlexWaygood, line 349: Hmm. I think we should return a nonzero exit code if _all_ tests were skipped. (This also applies to `mypy_test.py` and `stubtest_third_party.py`.) It's probably not what the user was expecting when they ran the script. If you're just running it from the command line, you'll notice that they've all been skipped due to the verbiage printed to the terminal, so it's okay. But if this is being run in CI, we could have all tests being skipped (due to something going wrong somewhere, idk), and the job would be marked as a success.

hamdanal:
It is strange hence my question :D
In this case let's keep things as they are, `self.requires_python.contains(version)` is expressive enough IMO.

tests/mypy_test.py
hamdanal: 😵‍💫

tests/regr_test.py
hamdanal: I made this change with a slight modification to make the code a little bit more concise

tests/regr_test.py
```diff
@@ -274,6 +276,29 @@ def concurrently_run_testcases(
...
+                assert requires_python is not None
+                if not requires_python.contains(version):
+                    msg = f"skipping {pkg!r} for target Python {version} (requires Python {requires_python})"
+                    print(colored(msg, "yellow"))
+                    continue
+            to_do.extend(
+                [
+                    partial(test_testcase_directory, testcase_dir, version, platform, verbosity=verbosity, tempdir=tempdir)
+                    for platform in platforms_to_test
+                ]
```
AlexWaygood, line 300: Looks like the list comprehension is actually unnecessary here; we can just make do with a generator expression. (My bad.)

```suggestion
                partial(test_testcase_directory, testcase_dir, version, platform, verbosity=verbosity, tempdir=tempdir)
                for platform in platforms_to_test
```

tests/regr_test.py
```diff
@@ -315,15 +335,20 @@ def main() -> ReturnCode:
         platforms_to_test, versions_to_test = SUPPORTED_PLATFORMS, SUPPORTED_VERSIONS
     else:
         platforms_to_test = args.platforms_to_test or [sys.platform]
-        versions_to_test = args.versions_to_test or [f"3.{sys.version_info[1]}"]
+        versions_to_test = args.versions_to_test or [PYTHON_VERSION]
 
     code = 0
+
```
AlexWaygood, line 341: spurious edit ;)

```suggestion
```

hamdanal: actually the line `code = 0` is not needed :)

AlexWaygood: Kill it with fire! It must have been accidentally left in after my big refactor in #10714 :D

hamdanal, line 341:
```suggestion
```

tests/mypy_test.py
```diff
@@ -493,34 +511,48 @@ def test_third_party_stubs(code: int, args: TestConfig, tempdir: Path) -> TestRe
     if not _DISTRIBUTION_TO_VENV_MAPPING:
         setup_virtual_environments(distributions_to_check, args, tempdir)
 
-    assert _DISTRIBUTION_TO_VENV_MAPPING.keys() == distributions_to_check.keys()
-
-    for distribution, venv_info in _DISTRIBUTION_TO_VENV_MAPPING.items():
+    # Some distributions may have been skipped earlier and therefore don't have a venv.
```
AlexWaygood, line 514: The comment immediately above here is no longer accurate:

https://github.com/python/typeshed/blob/9ae52399ac9bd6142bd1c345581ba80ebbd2f350/tests/mypy_test.py#L508-L510

I think we can simplify this logic slightly? Something like this:

```diff
diff --git a/tests/mypy_test.py b/tests/mypy_test.py
index 602b747fc..0ff0cd66c 100755
--- a/tests/mypy_test.py
+++ b/tests/mypy_test.py
@@ -395,6 +395,9 @@ def install_requirements_for_venv(venv_info: VenvInfo, args: TestConfig, externa

 def setup_virtual_environments(distributions: dict[str, PackageDependencies], args: TestConfig, tempdir: Path) -> None:     """Logic necessary for testing stubs with non-types dependencies in isolated environments."""
+    if not distributions:
+        return  # hooray! Nothing to do
+
     # STAGE 1: Determine which (if any) stubs packages require virtual environments.
     # Group stubs packages according to their external-requirements sets

@@ -505,20 +508,17 @@ def test_third_party_stubs(code: int, args: TestConfig, tempdir: Path) -> TestRe
         ):
             distributions_to_check[distribution] = get_recursive_requirements(distribution)

-    # If it's the first time test_third_party_stubs() has been called during this session,
-    # setup the necessary virtual environments for testing the third-party stubs.
-    # It should only be necessary to call setup_virtual_environments() once per session.
-    if not _DISTRIBUTION_TO_VENV_MAPPING:
-        setup_virtual_environments(distributions_to_check, args, tempdir)
-
-    # Some distributions may have been skipped earlier and therefore don't have a venv.
+    # Setup the necessary virtual environments for testing the third-party stubs.
+    # Note that some stubs may not be tested on all Python versions
+    # (due to version incompatibilities),
+    # so we can't guarantee that setup_virtual_environments()
+    # will only be called once per session.
     distributions_without_venv = {
         distribution: requirements
         for distribution, requirements in distributions_to_check.items()
         if distribution not in _DISTRIBUTION_TO_VENV_MAPPING
     }
-    if distributions_without_venv:
-        setup_virtual_environments(distributions_without_venv, args, tempdir)
+    setup_virtual_environments(distributions_without_venv, args, tempdir)
```

AlexWaygood (approved):
Other than https://github.com/python/typeshed/pull/10724#discussion_r1335206682, this now looks great. Thanks for persevering with this, there was a lot to think through here!

tests/regr_test.py
```diff
@@ -274,6 +276,27 @@ def concurrently_run_testcases(
...
+            if not testcase_dir.is_stdlib:  # type: ignore[misc]  # mypy bug, already fixed on master
+                assert requires_python is not None
+                if not requires_python.contains(version):
+                    msg = f"skipping {pkg!r} for target Python {version} (requires Python {requires_python})"
+                    print(colored(msg, "yellow"))
+                    continue
+            to_do.extend(
+                partial(test_testcase_directory, testcase_dir, version, platform, verbosity=verbosity, tempdir=tempdir)
+                for platform in platforms_to_test
+            )
```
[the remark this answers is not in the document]
hamdanal: Good idea, this avoids creating all the threads.
From a quick look, it looks like the two new lines should be dedented a level? I think they are still inside the outer loop. I cannot check right now but please feel free to commit the suggestion.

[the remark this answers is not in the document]
AlexWaygood: Ah, good catch!

AlexWaygood:
@srittau, did you want to take a quick look, or are you okay with me merging this?

srittau:
I haven't had time to look at it, but I don't need to. Please merge if you are confident.

AlexWaygood:
Thanks again @hamdanal! :D

hamdanal:
Thank you Alex, as always, you've been very helpful :)

Avasam:
Thanks! My only issue, is that it's not documented in https://github.com/python/typeshed/blob/main/CONTRIBUTING.md#the-metadatatoml-file !

AlexWaygood:
> Thanks! My only issue, is that it's not documented in https://github.com/python/typeshed/blob/main/CONTRIBUTING.md#the-metadatatoml-file !

A very good point, that I completely missed!

AlexWaygood:
> Thanks! My only issue, is that it's not documented in https://github.com/python/typeshed/blob/main/CONTRIBUTING.md#the-metadatatoml-file !

Added it to the to-do list in https://github.com/python/typeshed/issues/10722#issuecomment-1732290791

PR #10725, opened 2023-09-18
oauthlib: Update types for RequestValidator to match implementation

Mark-Hetherington:
Type hints updated based on docstrings from implementation. 

Resolves typing issues that have arisen in a project using a RequestValidator implementation.


srittau (approved):
Thanks! The annotations look sensible to me.

Issue #10727, opened 2023-09-18
`tuple` methods should use `Self`?


randolf-scholz:
In the definition of `tuple`

https://github.com/python/typeshed/blob/8597724ac113d187d2ebc9f58cef3b410b1d5dae/stdlib/builtins.pyi#L956-L980

It seems in many methods, **in the arbitrary size case**, one could/should substitute `tuple[_T_co, ...]` with `Self`, in particular `__add__(self, __value: tuple[_T_co, ...])`, `__mul__`, `__rmul__` and `__getitem__(self, __key: slice)`. I.e. when `tuple[_T_co, ...]` appears as a return value.

How can one do a case distinction between the arbitrary size case and the fixed size case?

randolf-scholz:
Ad least for `list` it seems that the override should be `def __getitem__(self, __s: slice) -> Self` instead of https://github.com/python/typeshed/blob/8597724ac113d187d2ebc9f58cef3b410b1d5dae/stdlib/builtins.pyi#L1037-L1038. However, both inherit this method from `Sequence`.

Thinking about it, it seems that even for `Sequence` we should have

```python
def __getitem__(self, __s: slice) -> Self
```

and `tuple` is rather the odd one out.

AlexWaygood:
> Thinking about it, it seems that even for `Sequence` we should have
> 
> ```python
> def __getitem__(self, __s: slice) -> Self
> ```

Be careful about that one. Changing the signature of `Sequence.__getitem__` means you're changing the definition of what it means in Python for something to be considered a Sequence. I'm not sure it's true that a class _has_ to return `Self` from `__getitem__` in order for it to be considered a `Sequence` — I think we can be more lenient there

randolf-scholz:
As per my other comment, one possible solution would be to introduce a `FixedSizeSequence` that captures the nature of `tuple[T1, ..., Tn]` and then do a case distinction when calling `tuple[...]`

- `tuple[T, ...]` -> `Sequence[T]`
- `tuple[T1, ...., Tn]` -> `FixedSizeSquence[Ts]`, where `Ts` is a `TypeVarTuple`.

Or, and I don't know if this is possible, do a case distinction in `Sequence` where `Sequence[T]` gives one set of overloads, but `Sequence[Ts]` for a `TypeVarTuple` gives another set of overloads.

randolf-scholz:
> I'm not sure I fully understand this proposal, since Sequence is generic over a single TypeVar rather than a TypeVarTuple

[...]

PR #10728, opened 2023-09-18
changed tuple methods to use self

randolf-scholz:
Testing the ramifications of naively implementing #10727.


stdlib/builtins.pyi
```diff
@@ -960,7 +960,7 @@ class tuple(Sequence[_T_co], Generic[_T_co]):
     @overload
     def __getitem__(self, __key: SupportsIndex) -> _T_co: ...
     @overload
-    def __getitem__(self, __key: slice) -> tuple[_T_co, ...]: ...
+    def __getitem__(self, __key: slice) -> Self: ...
```
AlexWaygood, line 963: That would imply that this should hold true, which I don't think it's accurate:

```python
def func(t: tuple[int, int, int]):
    x = t[:2]
    reveal_type(x)  # tuple[int, int, int], according to this PR?
```

randolf-scholz: Ah, I guess one needs the `TypeVarTuple` stuff. The issue is that for instance `torch.Size` subclasses `tuple`, but adding two `torch.Size` instances gives a `torch.Size` and not a `tuple`.

randolf-scholz: I guess in some sense the issue is really the special casing here, we have two things: `tuple` with a finite set of types and `tuple` with arbitrary size. The `Self` only applies directly to the arbitrary size case. Is there any good way to do a case-distinction for this?

AlexWaygood: Yeah, the issue is partly the extent to which `tuple` is heavily special-cased, partly the fact we can't use `TypeVarTuple` yet in typeshed, and partly the fact that we'd probably also need higher-kinded types to truly express in the stubs what you're going for here :-)

I don't think there's any way in today's type system to express what you're trying to express here, unfortunately :-(

Pytorch could maybe do something like this, as a workaround:

```py
class Size:
    if TYPE_CHECKING:
        def __getitem__(self, __key: slice) -> Self: ...
```

Sorry that I can't offer anything better :/

randolf-scholz: Even for the base class `Sequence` it seems logical that we should have `def __getitem__(self, __s: slice) -> Self`. Tuple with a finite set of types is really quite different to this semantically.

randolf-scholz: Just spitballing here, but it seems like there should be `FiniteSequence(Ts)` and `tuple[T1, ..., Tn]` should be a subtype of `FiniteSequence` whereas `tuple[T, ...]` should be a subtype of `Sequence[T]`.

AlexWaygood: > Even for the base class `Sequence` it seems logical that we should have `def __getitem__(self, __s: slice) -> Self`.

See my comments here on that one: https://github.com/python/typeshed/issues/10727#issuecomment-1723396017

Issue #8832, opened 2022-10-03
Consider adding `__slots__` to stubs?


sobolevn:
Since mypy now supports `__slots__` checking, we can add `__slots__` to stubs.
This will allow us to correctly raise `Something is not in __slots__` when assigning extra attributes.

Right now there are no `__slots__` defined: https://github.com/python/typeshed/search?q=__slots__

If this is something we are going for, there are several things to consider:
1. Adding this feature to `stubtest`. It can check that `__slots__` contents do match. Probably behind a feature flag. We can enable it for `stdlib` only
2. Should we allow skipping `_protected` attributes? For example, if some class is defined as
   ```python
   class A:
      __slots__ = ('_prop',)
   ```

   Should we allow this stub?

   ```python
   class A:
       __slots__ = ()
   ```

AlexWaygood:
> 1. Adding this feature to `stubtest`. It can check that `__slots__` contents do match. Probably behind a feature flag. We can enable it for `stdlib` only

It's currently included in stubtest's list of always-ignored dunders inside classes: https://github.com/python/mypy/blob/dc5c299aa190949f2300b163ccc10257a779006d/mypy/stubtest.py#L1280-L1281.

hauntsaninja:
Technically it's not always ignored, it's "ignorable". Aka if the runtime has `__slots__`, stubtest won't currently enforce that the stub has `__slots__`. But if the stub does have `__slots__`, stubtest should make sure they match the runtime.

JelleZijlstra:
I agree that adding `__slots__` is useful in stubs.

For your question 2, I think it's best to add protected attributes. That way, we can just have the slots definition match exactly between runtime and the stub, which is easier to test and review.

sobolevn:
Ok, I will start with `stubtest` PR! 👍

Akuli:
> to correctly raise Something is not in __slots__ when assigning extra attributes.

Can you clarify this? Type checkers complain about assigning non-existent attributes anyway, unless it's a subclass and then the `__slots__` of the parent class doesn't matter anyway at runtime (unless the subclass defines its own `__slots__`, in that case the fields that are actually accessible are `parent class slots + child class slots`).

AlexWaygood:
As @Akuli says, I don't think adding `__slots__` to the stub helps much with whether or not type checkers will validate attribute assignments.

Where it _might_ help is in code like this:

```pycon
>>> class A:
...     __slots__ = ["foo"]
... 
...     
[...]

PR #10730, opened 2023-09-19
define slots on abc.ABC

SebbyLaw:
Add `__slots__` to the declaration of `abc.ABC` to properly reflect its implementation.

This allows Pyright to properly check ABCs, see https://github.com/microsoft/pyright/issues/5976 for more details


AlexWaygood (approved):
This is our first ever addition of `__slots__` to a class in typeshed! We previously discussed adding `__slots__` to classes in #8832, however, and it doesn't look like anybody was hugely opposed to adding `__slots__` to stubs. And this fixes a pyright false positive, so let's go for it.

PR #10731, opened 2023-09-19
[jsonschema] Add registry arg to Validator init

HansAarneLiblik:
Added `registry` argument to Validators `__init__`

Source: https://python-jsonschema.readthedocs.io/en/stable/api/jsonschema/protocols/#jsonschema.protocols.Validator


AlexWaygood:
Looks like this was all taken care of in #10583, but thanks for the PR!

PR #10738, opened 2023-09-20
typing.pyi: IO.name may be bytes

### LINKED ISSUE #10737: stdlib/typing.pyi IO.name declared `str` but it should be `str | bytes`
```python
Python 3.11.5 (main, Aug 25 2023, 07:43:52) [GCC 12.2.0] on linux
>>> open(b'/dev/null')
<_io.TextIOWrapper name=b'/dev/null' mode='r' encoding='UTF-8'>
```

File IO classes can have `bytes` names, not just `str`.  The name field at construction time and thus name property on IO should thus be `str | bytes`.

JelleZijlstra:
Fixes #10737 

Worried this may be too disruptive, but let's see what mypy-primer says.


hauntsaninja:
We could `str | Any` it and then make it Generic if we get PEP 696?

JelleZijlstra:
`| Any` works pretty well here.

The pandas code that's still flagged by mypy-primer (https://github.com/pandas-dev/pandas/blob/b303665df0337448abbc6e3107be1f0ff7c98fb5/pandas/io/parquet.py#L210) does `isinstance(path_or_handle.name, (str, bytes))`, so we're producing a result that's more in line with expectations.

hauntsaninja (approved):

srittau (approved):

PR #10740, opened 2023-09-20
[requests] Improve exception class constructors

intgr:
Somewhat related to #8989.

Relevant code: https://requests.readthedocs.io/en/latest/_modules/requests/exceptions/#RequestException


stubs/requests/requests/exceptions.pyi
```diff
@@ -1,18 +1,21 @@
...
+
 from urllib3.exceptions import HTTPError as BaseHTTPError
 
 from .models import Request, Response
 
 class RequestException(OSError):
     response: Response | None
     request: Request | None
-    def __init__(self, *args, **kwargs) -> None: ...
+    def __init__(self, *args: Any, request: Request | None = ..., response: Response | None = ...) -> None: ...
```
JelleZijlstra, line 10:
```suggestion
    def __init__(self, *args: object, request: Request | None = ..., response: Response | None = ...) -> None: ...
```
Consistent with how we annotate `BaseException.__init__`.

intgr: Good point, changed.

JelleZijlstra:
Should we make the arguments also optional on HTTPError?

srittau:
> Should we make the arguments also optional on HTTPError?

Why add `HTTPError.__init__()` at all? It's just inherited in the implementation.

intgr:
> Why add HTTPError.__init__() at all? It's just inherited in the implementation.

#8989 created an assumption that `request/response` attrs are always present and not `None` in `HTTPError`. That was done specifically to reduce regressions in mypy_primer (because users had already made this assumption): https://github.com/python/typeshed/pull/8989#issuecomment-1292118549

```python
class HTTPError(RequestException):
    request: Request    # <-- can't be None
    response: Response  # <-- can't be None
```

That assumption is true for errors originating from `requests` library proper, but as can be seen here, does not hold for third party code raising these exceptions.

Relaxing `HTTPError.__init__` isn't correct, but maybe it's a useful tradeoff between introducing regressions and accuracy. I'll leave it up to you, maybe you have a better gut feeling for making such tradeoffs.

srittau:
It's a tradeoff. The primer output is a bit concerning, although most of the problems seem to be in test code. I'm fine with this change, despite the primer output. Otherwise we should revert the `| None` annotations in `HTTPError`, which is probably worse.

Issue #10349, opened 2023-06-23
`redis.asyncio.Redis.from_url` missing overloads


elder-frog:
`redis.asyncio.Redis.from_url` missing overloads

```python
@classmethod
def from_url(cls, url: str, **kwargs) -> Redis[Any]: ...
```

https://github.com/python/typeshed/blob/4c1e94d8e0f212f3b611d06698427d77e696e636/stubs/redis/redis/asyncio/client.pyi#L34C11-L34C11

PR #10742, opened 2023-09-21
Add overloads to redis.asyncio.Redis

aminalaee:
Closes https://github.com/python/typeshed/issues/10349

Add overloads to `redis.asyncio.Redis` `from_url` and `__init__` based on `decode_responses` to `Redis[str]` and `Redis[bytes]`


aminalaee:
Does it make sense to add arguments from `redis.Client.from_url` to `redis.asyncio.Client.from_url`?
https://github.com/python/typeshed/blob/53422f8918debdaf60e1605d0e2e3ce77a810afd/stubs/redis/redis/client.pyi#L86-L90

There are two recent arguments added to the `redis.asyncio.Client.from_url` https://github.com/redis/redis-py/blob/1b370da71709ad31ec947bf14af4b5831d57b9cb/redis/asyncio/client.py#L116, not sure if we want to add types for them too.

JelleZijlstra:
> There are two recent arguments added to the `redis.asyncio.Client.from_url` https://github.com/redis/redis-py/blob/1b370da71709ad31ec947bf14af4b5831d57b9cb/redis/asyncio/client.py#L116, not sure if we want to add types for them too.

We should add them once there is a release of redis-py that includes them.

JelleZijlstra:
Thanks!

PR #10743, opened 2023-09-21
Add stubs for python-decouple

markis:
This is a small single file module which for the most has a main `config` function and a few supporting classes.

I used the docs as a first reference, since they contain some type hints and then verified it with the actual implementation.

The doc is available here: https://github.com/HBNetwork/python-decouple/blob/master/README.rst
Source code here: hhttps://github.com/HBNetwork/python-decouple/blob/master/decouple.py


stubs/python-decouple/decouple.pyi
```diff
@@ -0,0 +1,105 @@
...
+@overload
+def Csv() -> _CsvType[str, list[str]]: ...
+@overload
+def Csv(cast: Callable[..., _T], delimter: str = ..., strip: str = ...) -> _CsvType[_T, list[_T]]: ...
+@overload
+def Csv(cast: Callable[..., _T], delimiter: str, strip: str, post_process: Callable[..., _TCsv]) -> _CsvType[_T, _TCsv]: ...
+@overload
+def Csv(
+    *, cast: Callable[..., _T], post_process: Callable[..., _TCsv], delimiter: str = ..., strip: str = ...
+) -> _CsvType[_T, _TCsv]: ...
```
markis, line 93: This is likely the most complex use case in this module. The Csv class exhibits default behavior that can be modified based on parameters passed into its `__init__` method. By default, `cast` is set to `str`, and `post_process` is set to `list`. Therefore, if these parameters are explicitly defined, the types provided should be used. However, if they are not specified, the return type of the `__call__` method should default to `list[str]`.

markis: If there is a better way to express this, I am open to suggestions.

JelleZijlstra: I haven't looked deeply, but it sounds like you simply need a generic `__init__` method.

stubs/python-decouple/decouple.pyi
```diff
@@ -0,0 +1,98 @@
...
+
+class Csv(Generic[_T, _TCsv]):
+    cast: Callable[..., _T]
+    delimiter: str
+    strip: str
+    post_process: Callable[..., _TCsv]
+    def __call__(self, value: str) -> _TCsv: ...
+    def __init__(
+        self, cast: Callable[..., _T] = str, delimiter: str = ..., strip: str = ..., post_process: Callable[..., _TCsv] = list[_T]
+    ) -> None: ...
```
Daverball, line 86:
```suggestion
    @overload
    def __init__(
        self: Csv[str, list[str]], cast: Callable[[str], str] = str, delimiter: str = ..., strip: str = ..., post_process: Callable[[Iterable[str]], list[str]] = list
    ) -> None: ...
    @overload
    def __init__(
        self: Csv[_T, list[_T]], cast: Callable[[str], _T], delimiter: str = ..., strip: str = ..., post_process: Callable[[Iterable[_T]], list[_T]] = list
    ) -> None: ...
    ) -> None: ...
    @overload
    def __init__(
        self: Csv[str, _TCsv], cast: Callable[[str], _T] = str, delimiter: str = ..., strip: str = ..., *, post_process: Callable[Iterable[str], _TCsv]
    ) -> None: ...
    @overload
    def __init__(
        self: Csv[_T, _TCsv], cast: Callable[[str], _T], delimiter: str, strip: str, post_process: Callable[[Iterable[_T]], _TCsv]
```
I think this should be something along these lines. The arguments of `cast`/`post_process` are very much defined by the implementation and since they are optional parameters you will need to provide overloads, because otherwise _T/_TCsv will not get bound.

It's also worth noting that you will need to import `Iterable` to make this suggestion work.

stubs/python-decouple/decouple.pyi
```diff
@@ -0,0 +1,98 @@
+from _typeshed import Incomplete
+from collections import OrderedDict
+from collections.abc import Callable, Sequence
+from configparser import ConfigParser
+from typing import Any, Generic, TextIO, TypeVar, overload
+
+_T = TypeVar("_T")
+_TCsv = TypeVar("_TCsv", bound=Sequence[Any])
```
Daverball, line 8: I think this upper bound is too strict, I would change this to `Collection[Any]` at the very least, but technically this can be anything and the docs also state as much (you could e.g. imagine someone calculating an integer hash from the individual values by providing `post_process=lambda x: hash(tuple(x))`) so I'm not sure if it's worth providing an upper bound.

stubs/python-decouple/decouple.pyi
```diff
@@ -0,0 +1,98 @@
...
+    flat: Sequence[_T]
+    cast: Callable[..., _T]
+    choices: Sequence[_T]
+    def __init__(
+        self,
+        flat: Sequence[_T] | None = None,
+        cast: Callable[..., _T] = ...,
+        choices: Sequence[_T | tuple[Incomplete, _T]] | None = None,
+    ) -> None: ...
+    def __call__(self, value: str) -> _T: ...
```
Daverball, line 98:
```suggestion
    cast: Callable[..., _T]
    choices: Sequence[_T]
    @overload
    def __init__(
        self: Choices[str],
        flat: Sequence[str] | None = None,
        cast: Callable[[str], str] = ...,
        choices: Sequence[tuple[str, str]] | None = None,
    ) -> None: ...
    @overload
    def __init__(
        self: Choices[_T],
        flat: Sequence[_T],
        cast: Callable[[str], _T] = ...,
        choices: Sequence[tuple[_T, str]] | None = None,
    ) -> None: ...
    @overload
    def __init__(
        self: Choices[_T],
        flat: None,
        cast: Callable[[str], _T],
        choices: Sequence[tuple[_T, str]] | None = None,
    ) -> None: ...
    @overload
    def __init__(
        self: Choices[_T],
        flat: None = None,
        cast: Callable[[str], _T] = ...,
        *,
        choices: Sequence[tuple[_T, str]],
    ) -> None: ...
    def __call__(self, value: str) -> _T: ...
```
Same deal here, this takes some overloads to bind correctly and the parameters  of cast are given, and choices does need to be a sequence of 2 value tuples, the second value of which is always a string.

Technically you could make this generic in the input value type as well, but since it is meant for dealing with environment variables, it probably makes sense to stick with `str` as all the examples have as well.

stubs/python-decouple/decouple.pyi
markis: Yep, that's a great point!

stubs/python-decouple/decouple.pyi
markis: This is a great suggestion!  I didn't think about setting the type on `self` to influence the generic type.

I updated the PR with your suggestions, everything looks great, but there might be a weird regression with cast by itself. For some reason, mypy always thinks the return is a `str`.

Example:
```python
from decouple import Choices, Csv
from typing_extensions import reveal_type

reveal_type(Choices(flat=[1, 2])(""))
reveal_type(Choices(choices=[(1, "one"), (2, "two")])(""))
reveal_type(Choices(cast=int, flat=[1, 2])(""))
reveal_type(Choices(cast=int, choices=[(1, "one"), (2, "two")])(""))
reveal_type(Choices(flat=[1, 2], choices=[(1, "one"), (2, "two")])(""))
reveal_type(Choices(cast=int, flat=[1, 2], choices=[(1, "one"), (2, "two")])(""))
reveal_type(Choices(cast=int)(""))
```

Output:
```
test.py:4: note: Revealed type is "builtins.int"
test.py:5: note: Revealed type is "builtins.int"
test.py:6: note: Revealed type is "builtins.int"
test.py:7: note: Revealed type is "builtins.int"
test.py:8: note: Revealed type is "builtins.int"
test.py:9: note: Revealed type is "builtins.int"
test.py:10: note: Revealed type is "builtins.str"
test.py:10: error: Argument "cast" to "Choices" has incompatible type "type[int]"; expected "Callable[[str], str]"  [arg-type]
```

markis: Would you know how to address the `reveal_type(Choices(cast=int)(""))` case?

markis: I am sorry, I messed this up. Let me delete my comments and update it.

markis: Thank you for this suggestion! I did not realize you could suggest the types of `_T` and `_Csv` through the `self` argument.  Really interesting stuff.

After I implemented your suggestions, I am getting some regressions on my code while using this type file.  Would you know how to fix it?  Also, I am wondering if the `Iterable` is having some  unwanted side effects.  Also, the errors I am seeing are coming from mypy, I haven't checked the other type checkers.

```python
from decouple import Choices, Csv
from typing_extensions import reveal_type

# No error and correct type
reveal_type(Csv()(""))  # note: Revealed type is "builtins.list[builtins.str]"

# No error and correct type
reveal_type(Csv(cast=int)(""))  # note: Revealed type is "builtins.list[builtins.int]"

# error: Argument "post_process" to "Csv" has incompatible type "type[set[Any]]"; expected "Callable[[Iterable[str]], list[str]]"  [arg-type]
reveal_type(Csv(post_process=set)(""))  # note: Revealed type is "builtins.list[builtins.str]"

# error: Argument 1 to "Csv" has incompatible type "type[int]"; expected "Callable[[str], str]"  [arg-type]
# error: Argument 4 to "Csv" has incompatible type "type[tuple[Any, ...]]"; expected "Callable[[Iterable[str]], list[str]]"  [arg-type]#
reveal_type(Csv(int, ",", " ", tuple)(""))

# error: Argument "cast" to "Csv" has incompatible type "type[int]"; expected "Callable[[str], str]"  [arg-type]
# error: Argument "post_process" to "Csv" has incompatible type "type[tuple[Any, ...]]"; expected "Callable[[Iterable[str]], list[str]]"  [arg-type]
reveal_type(Csv(cast=int, post_process=tuple)(""))

# error: Argument "cast" to "Csv" has incompatible type "type[int]"; expected "Callable[[str], str]"  [arg-type]
# error: Argument "post_process" to "Csv" has incompatible type "type[tuple[Any, ...]]"; expected "Callable[[Iterable[str]], list[str]]"  [arg-type]
reveal_type(Csv(cast=int, delimiter=",", post_process=tuple)(""))

```

stubs/python-decouple/decouple.pyi
markis: This looks great!  Thanks again for the suggestion.

I have one issue, it seems like when `cast` is used by itself mypy seems to get confused.

```python
from decouple import Choices, Csv
from typing_extensions import reveal_type

# error: Argument "cast" to "Choices" has incompatible type "type[int]"; expected "Callable[[str], str]"  [arg-type]
reveal_type(Choices(cast=int)(""))

reveal_type(Choices(flat=[1, 2])("")) # note: Revealed type is "builtins.int"
reveal_type(Choices(choices=[(1, "one"), (2, "two")])("")) #note: Revealed type is "builtins.int"
reveal_type(Choices(cast=int, flat=[1, 2])("")) #note: Revealed type is "builtins.int"
reveal_type(Choices(cast=int, choices=[(1, "one"), (2, "two")])("")) #note: Revealed type is "builtins.int"
reveal_type(Choices(flat=[1, 2], choices=[(1, "one"), (2, "two")])("")) #note: Revealed type is "builtins.int"
reveal_type(Choices(cast=int, flat=[1, 2], choices=[(1, "one"), (2, "two")])("")) #note: Revealed type is "builtins.int"
```

Daverball: This needs one additional overload to cover that case:
```python
    @overload
    def __init__(
        self: Choices[_T],
        flat: Sequence[_T] | None = None,
        *,
        cast: Callable[[str], _T],
        choices: Sequence[tuple[_T, str]] | None = None,
    ) -> None: ...
```

Daverball: generic overload matching gets complicated really quickly if you want to cover every possible case...

stubs/python-decouple/decouple.pyi
```diff
@@ -0,0 +1,137 @@
...
+    @overload
+    def __call__(self, option: str, default: str, cast: Callable[..., _T]) -> _T: ...
+
+config: AutoConfig
+
+class Csv(Generic[_T, _TCsv]):
+    cast: Callable[..., _T]
+    delimiter: str
+    strip: str
+    post_process: Callable[..., _TCsv]
```
Daverball, line 82:
```suggestion
    cast: Callable[[str], _T]
    delimiter: str
    strip: str
    post_process: Callable[[Iterable[_T], _TCsv]
```
I forgot to include these in my suggestion

stubs/python-decouple/decouple.pyi
markis: > generic overload matching gets complicated really quickly if you want to cover every possible case on arguments other than the first one or two...

Yeah, I know what you mean. 

> I wish overloads would let you violate the requirement of positional optional parameters having to come after all the required ones, it would get rid of a lot of redundant overloads, that have to currently be used in order to express both the positional and keyword version of the overload.
> 
> But that would of course require some changes to the language, since the parser would have to ignore these violations specifically only on overloads, the alternative would be to have some typing specific sentinel value that means the parameter is required in order to match that overload, but the parser can still treat it like an optional one. But I am getting off course here.

TBH, this is how I thought it would work but the more I dug into making this type module, the more I realized that wasn't the case.  But, now I know and thanks for the background!

stubs/python-decouple/decouple.pyi
markis: Updated. Thanks again for your help!

stubs/python-decouple/decouple.pyi
```diff
@@ -0,0 +1,158 @@
...
+        self: Csv[_T, _TCsv],
+        cast: Callable[[str], _T],
+        delimiter: str = ",",
+        strip: str = ...,
+        *,
+        post_process: Callable[[Iterable[Any]], _TCsv],
+    ) -> None: ...
+    @overload
+    def __init__(
+        self: Csv[_T, _TCsv], cast: Callable[[str], _T], delimiter: str, strip: str, post_process: Callable[[Iterable[_T]], _TCsv]
```
Daverball, line 124:
```suggestion
        post_process: Callable[[Iterable[Any]], list[_T]] = ...,
    ) -> None: ...
    @overload
    def __init__(
        self: Csv[str, _TCsv],
        cast: Callable[[str], str] = ...,
        delimiter: str = ...,
        strip: str = ...,
        *,
        post_process: Callable[[Iterable[Any]], _TCsv],
    ) -> None: ...
    @overload
    def __init__(
        self: Csv[_T, _TCsv],
        cast: Callable[[str], _T],
        delimiter: str = ",",
        strip: str = ...,
        *,
        post_process: Callable[[Iterable[Any]], _TCsv],
    ) -> None: ...
    @overload
    def __init__(
        self: Csv[_T, _TCsv], cast: Callable[[str], _T], delimiter: str, strip: str, post_process: Callable[[Iterable[Any]], _TCsv]
```
You missed replacing a couple of `Iterable[_T]` with `Iterable[Any]`

markis: Thanks, updated!

markis:
@Daverball @JelleZijlstra, everything is updated.  Could you give it a review?

Daverball (approved):
LGTM

Only thing I might consider is adding a couple of regression tests for the complex overloads on `Csv` and `Choices` to make sure they keep working correctly for a bunch of common cases.

stubs/python-decouple/decouple.pyi
```diff
@@ -0,0 +1,162 @@
...
+    @overload
+    def get(self, option: str, *, cast: Callable[[str], _T]) -> _T: ...
+    @overload
+    def get(self, option: str, default: str | _T, cast: Callable[[str], _T]) -> _T: ...
+    @overload
+    def __call__(self, option: str, default: str = ...) -> str: ...
+    @overload
+    def __call__(self, option: str, *, cast: Callable[[str], _T]) -> _T: ...
+    @overload
+    def __call__(self, option: str, default: str | _T, cast: Callable[[str], _T]) -> _T: ...
```
Daverball, line 37:
```suggestion
    @overload
    def get(self, option: str, default: str = ...) -> str: ...
    @overload
    def get(self, option: str, default: _T) -> str | _T: ...
    @overload
    def get(self, option: str, *, cast: Callable[[str], _T]) -> _T: ...
    @overload
    def get(self, option: str, default: str | _T, cast: Callable[[str], _T]) -> _T: ...
    @overload
    def __call__(self, option: str, default: str = ...) -> str: ...
    @overload
    def __call__(self, option: str, default: _T) -> str | _T: ...
    @overload
    def __call__(self, option: str, *, cast: Callable[[str], _T]) -> _T: ...
    @overload
    def __call__(self, option: str, default: str | _T, cast: Callable[[str], _T]) -> _T: ...
```
Looking at the examples, `bool` is a special case that is treated differently and results in a cast that accepts both `str` and `bool`, thus allowing `default` to be `bool` as well, but in general `default` probably can't always pass through. So it might get a little complicated, but what is definitely true however, is, that without a `cast`, the default will just pass through unchanged, so the overloads without a `cast` should look different.

I am still not a big fan of `default: str | _T`, since that technically means cast should be `Callable[[str | _T], _T]` but that's too strict if `default` is actually just a `str`. So you would need yet another overload, and it would probably end up resulting in more type errors for stuff that should be legal.

But this is kind of just an inherent problem of this API, it would make a lot more sense for `default` to just return immediately, without going through `cast`.

markis: Same, not a fan of `default: str | _T` and you are correct about `bool` being a special case.

Another case that prompted the change was also `int`, for example `config("EMAIL_PORT", default=25, cast=int)`.  Which is in the docs, so I wanted to make sure it's a use case that's supported.

markis: However, I should add that I incorporated your suggested changes in the PR.

stubs/python-decouple/@tests/stubtest_allowlist.txt
```diff
@@ -0,0 +1,2 @@
+# Runtime has a default of it's internal Undefined class, but ultimately it is a string
```
JelleZijlstra, line 1:
```suggestion
# Runtime has a default of its internal Undefined class, but ultimately it is a string
```

stubs/python-decouple/decouple.pyi
```diff
@@ -0,0 +1,168 @@
...
+
+_T = TypeVar("_T")
+_TCsv = TypeVar("_TCsv")
+
+PYVERSION: Incomplete  # undocumented
+DEFAULT_ENCODING: str  # undocumented
+TRUE_VALUES: set[str]  # undocumented
+FALSE_VALUES: set[str]  # undocumented
+
+def read_config(parser: ConfigParser, file: TextIO) -> None: ...  # undocumented
```
JelleZijlstra, line 15:
```suggestion
def read_config(parser: ConfigParser, file: Iterable[str]) -> None: ...  # undocumented
```
Matching https://github.com/python/typeshed/blob/96eaa767ddfb5b206e0a57a5663daf07212b2675/stdlib/configparser.pyi#L150 which this calls. We generally try to avoid the "IO" classes as they are too restrictive.

stubs/python-decouple/decouple.pyi
```diff
@@ -0,0 +1,168 @@
...
+    def __call__(self, option: str, default: str | _T, cast: Callable[[str], _T]) -> _T: ...
+
+class _Repository(Protocol):  # undocumented
+    def __contains__(self, key: str) -> bool: ...
+    def __getitem__(self, key: str) -> str: ...
+
+class RepositoryEmpty:  # undocumented
+    def __init__(self, source: str = ..., encoding: str = ...) -> None: ...
+    def __contains__(self, key: str) -> bool: ...
+    def __getitem__(self, key: str) -> str: ...
```
JelleZijlstra, line 50: This actually returns None. That's inconsistent with the protocol though, and maybe reflecting it accurately would make the types too complicated. Seems worth a comment, though.

stubs/python-decouple/decouple.pyi
```diff
@@ -0,0 +1,168 @@
...
+    @overload
+    def __call__(self, option: str, default: str | _T, cast: Callable[[str], _T]) -> _T: ...
+
+config: AutoConfig
+
+class Csv(Generic[_T, _TCsv]):
+    cast: Callable[[str], _T]
+    delimiter: str
+    strip: str
+    post_process: Callable[[Iterable[Any]], _TCsv]
```
JelleZijlstra, line 92: Shouldn't the Any be _T as the post-processor gets passed whatever came out of `cast`?

stubs/python-decouple/decouple.pyi
```diff
@@ -0,0 +1,168 @@
...
+    def __call__(self, value: str) -> _TCsv: ...
```
JelleZijlstra, line 93: The implementation also supports passing None. In that case, `post_process` is called with no arguments. Typing that accurately would probably require making a callback protocol for post_process.

stubs/python-decouple/decouple.pyi
```diff
@@ -0,0 +1,168 @@
...
+        cast: Callable[[str], _T],
+        delimiter: str,
+        strip: str,
+        post_process: Callable[[Iterable[Any]], _TCsv],
+    ) -> None: ...
+
+class Choices(Generic[_T]):
+    flat: Sequence[_T]
+    cast: Callable[[str], _T]
+    choices: Sequence[_T]
```
JelleZijlstra, line 140: Shouldn't this be `Sequence[tuple[_T, object]]`?

Also in the overloads below I think the second tuple member in `choices` should be `object`, as the implementation ignores it.

stubs/python-decouple/decouple.pyi
Daverball: Yes, that was my first suggestion, and we tried that, however mypy will then reject generic classes as a `post_process` since it can't seem to solve for both type vars simultaneously i.e. it will reject `post_process=set` but it would accept `post_process=set[str]`.

I agree that this should work, but it's also a very challenging scenario for an overload solver, so I am not surprised mypy can't do it yet.

stubs/python-decouple/decouple.pyi
Daverball: Good catch, I glossed over that part in my review. I am however a little concerned how well a callback protocol will behave, considering mypy would match `Csv(post_process=set)` to `Csv[Any, builtins.list[Any]]` when I tried to replace it with a callback protocol. Changing the order of overloads would so the first one is fully generic would probably solve that, but you would still end up with unsolved type vars. Although I am not sure I have tried a callback protocol with `Any` for the first argument, instead of `_T`, so maybe that will behave a bit better.

stubs/python-decouple/decouple.pyi
Daverball: Indeed, I had already suggested the same change for `choices` but I guess it fell victim to manual copy pasta.

While I do agree that the second member in the tuple is ignored and could be anything, I don't think it's a good change.
It expects that when `cast` is used on the second element in the tuple, for it to return the first element, so while the implementation is bogus and backwards and doesn't verify that, the stubs should probably reflect the intent and the intent was for the second element to be the string representation of the choice.

stubs/python-decouple/@tests/stubtest_allowlist.txt
markis: Thanks, updated!

stubs/python-decouple/decouple.pyi
markis: Nice! Updated.

stubs/python-decouple/decouple.pyi
markis: Yes, this is a complicated situation. I added a comment.

stubs/python-decouple/decouple.pyi
markis: Should I add a comment that it should be `_T` but mypy wasn't able to support it.  So in the future, it's more obvious about the decision?

stubs/python-decouple/decouple.pyi
```diff
@@ -0,0 +1,181 @@
...
+    @overload
+    def __call__(self, option: str, default: _T) -> str | _T: ...
+    @overload
+    def __call__(self, option: str, *, cast: Callable[[str], _T]) -> _T: ...
+    @overload
+    def __call__(self, option: str, default: str | _T, cast: Callable[[str], _T]) -> _T: ...
+
+config: AutoConfig
+
+class _PostProcess(Protocol[_T_co, _TCsv_co]):  # undocumented
```
srittau, line 91: mypy wants `_T_co` to be contravariant, pyright wants it to be covariant (which makes sense, since it's used as the type var to `Iterable`, which is covariant). Not sure what to do here.

But then ...

stubs/python-decouple/decouple.pyi
```diff
@@ -0,0 +1,181 @@
...
+    @overload
+    def __call__(self) -> _TCsv_co: ...
+    @overload
+    def __call__(self, /, __value: Iterable[_T_co]) -> _TCsv_co: ...
+    @overload
+    def __call__(self, /, __value: Iterable[Any]) -> _TCsv_co: ...
```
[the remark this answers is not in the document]
JelleZijlstra: I just came to the same conclusion.

srittau:
@JelleZijlstra Since you've already fixed some problem, do you want to close the remaining problems (`__init__` -> `__new__`)? Otherwise I would close this PR.

srittau:
Thanks for contributing! I'm closing this PR for now, because it still fails some tests after three months of inactivity. If you are still interested, please feel free to open a new PR (or ping us to reopen this one).

PR #10745, opened 2023-09-21
fix openpyxl worksheet columns to return `tuple[Cell, ...]`

### LINKED ISSUE #10744: openpyxl Worksheet columns is incorrectly typed
Property `columns` should return a generator of `tuple[Cell]` instead of a single `Cell`.

This is a very similar issue to #9904, except for `column`.

ostr00000:
Fix for incorrect return type for property `columns` in `Worksheet` in `openpyxl`.

closes #10744


srittau:
Thanks for your contribution, @ostr00000 and sorry that it took so long. I've now merged @Avasam's big PR, which includes your change.

Issue #10611, opened 2023-08-24
Should annotations for private attributes and methods be included to facilitate type checking of subclasses?


josephcourtney:
I wrote a subclass of asyncio.TaskGroup that allows the user to add a predefined task to the group (source included for reference). However, mypy complains that _loop, _on_task_done, and _tasks since they are not included in the TaskGroup stub. Strangely, however, it does not complain about calling the method _check_accepting_tasks.

I see that private methods and attributes are typically not included in stubs, but this makes their access in subclasses produce errors in type checkers. Is there a good alternative to including them in a stub that I am not familiar with?

I'd be happy to add the missing annotations to the stub.

```python
# In compliance with the Python Software Foundation License Version 2, the following is reproduced:
#
# Copyright (c) 2001, 2002, 2003, 2004, 2005, 2006, 2007, 2008, 2009, 2010,
# 2011, 2012, 2013, 2014, 2015, 2016, 2017, 2018, 2019, 2020, 2021, 2022, 2023 Python Software Foundation;
# All Rights Reserved
#
# parts of the definition of FriendlyTaskGroup are derived from the definition of TaskGroup in
# cpython/Lib/asyncio/taskgroups.py, Python version ~3.12.0rc1, (git short hash c163d7f)


import contextvars
from asyncio import Task, TaskGroup
from collections.abc import Coroutine


class FriendlyTaskGroup(TaskGroup):
    """FriendlyTaskGroup."""

    def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)

    @property
    def loop(self):
        return self._loop

    def _check_accepting_tasks(self, raise_exception=True):
        if not self._entered:
            if raise_exception:
                msg = f"TaskGroup {self!r} has not been entered"
                raise RuntimeError(msg)
            return False

        if self._exiting and not self._tasks:
            if raise_exception:
                msg = f"TaskGroup {self!r} is finished"
                raise RuntimeError(msg)
            return False
        if self._aborting:
            if raise_exception:
                msg = f"TaskGroup {self!r} is shutting down"
                raise RuntimeError(msg)
            return False

        return True

    def add_task(self, task: Task) -> Task:
        """Add a task in this task group.

        Parameters
        ----------
        task : asyncio.Task
            The task to add to this group.

        Returns
        ----------
        task : asyncio.Task
            The same task provided as input.
        """
[...]

PR #10747, opened 2023-09-21
feat: Add _thread attribute for logging.handlers.QueueListener

tomasfarias:
This seems like a small change so I've skipped opening an issue.

The only way of knowing whether a `logging.handlers.QueueListener` has already been started is by checking that the `_thread` attribute has been set to a non-`None` value. This attribute is not included in type stubs, so running `mypy` on a code like this will raise a `"QueueListener" has no attribute "_thread"` error, even though it's perfectly valid:

```python
import logging.handlers
import queue

LOG_QUEUE: queue.Queue = queue.Queue(-1)

QUEUE_LISTENER: logging.handlers.QueueListener = logging.handlers.QueueListener(LOG_QUEUE, logging.StreamHandler())


def check():
    assert hasattr(QUEUE_LISTENER, "_thread")
    
    if QUEUE_LISTENER._thread is None:  # error: "QueueListener" has no attribute "_thread"
        QUEUE_LISTENER.start()
```

Delaying the start of a `logging.handlers.QueueListener` can be a useful pattern if sharing a logger in multiple modules.

I've read the contributing docs and it seems that adding private attributes is discouraged. 

However, [this comment](https://github.com/python/typeshed/issues/10611#issuecomment-1691722625) from last month indicates that the policy is to accept changes if requested. So, this PR requests the addition of `_thread` to the type stub of `logging.handlers.QueueListener` as it's the only way of knowing we have started to listen on a queue (or at least the most direct way).

Hope this can be accepted and thank you for your work in maintaining typeshed!

### DIFF
--- a/stdlib/logging/handlers.pyi
+++ b/stdlib/logging/handlers.pyi
@@ -2,14 +2,13 @@ import datetime
 import http.client
 import ssl
 import sys
+from _typeshed import ReadableBuffer, StrPath
 from collections.abc import Callable
 from logging import FileHandler, Handler, LogRecord
 from re import Pattern
 from socket import SocketKind, socket
 from threading import Thread
 from typing import Any, ClassVar, Protocol, TypeVar
-
-from _typeshed import ReadableBuffer, StrPath
 
 _T = TypeVar("_T")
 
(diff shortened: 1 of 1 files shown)


Akuli (approved):

Akuli:
Thanks! I checked the source code, and I don't see a better way find out whether `start()` has been called.

PR #10748, opened 2023-09-22
Add os.EX_OK for Python 3.11

aminalaee:
closes https://github.com/python/typeshed/issues/10242

PR #10749, opened 2023-09-22
Update os on Windows, Python3.12+

aminalaee:
Following the comment https://github.com/python/typeshed/pull/10748#discussion_r1334279493


stdlib/os/__init__.pyi
```diff
@@ -1044,3 +1050,8 @@ if sys.version_info >= (3, 9):
 
     if sys.platform == "linux":
         def pidfd_open(pid: int, flags: int = ...) -> int: ...
+
+if sys.version_info >= (3, 12) and sys.platform == "win32":
+    def listdrives() -> list[str]: ...
+    def listmounts(str) -> list[str]: ...
```
AlexWaygood, line 1056:
```suggestion
    def listmounts(volume: str) -> list[str]: ...
```

stdlib/os/__init__.pyi
```diff
@@ -349,13 +349,17 @@ class stat_result(structseq[float], tuple[int, int, int, int, int, int, int, flo
...
             # On other Unix systems (such as FreeBSD), the following attributes may be
             # available (but may be only filled out if root tries to use them):
             @property
             def st_gen(self) -> int: ...  # file generation number
-            @property
-            def st_birthtime(self) -> int: ...  # time of file creation
+
+    if (sys.version_info >= (3, 12) and sys.platform == "win32") or (sys.platform != "linux" and sys.platform != "win32"):
+        @property
+        def st_birthtime(self) -> int: ...  # time of file creation in seconds
```
aminalaee, line 359: I think this can be improved, open to suggestions:
- Either repeat the definition of`st_birthtime` for win32 and non-linux platforms
- Or add some nested condition to for this

stdlib/os/__init__.pyi
```diff
@@ -349,13 +349,17 @@ class stat_result(structseq[float], tuple[int, int, int, int, int, int, int, flo
         @property
         def st_rdev(self) -> int: ...  # type of device if an inode device
         if sys.platform != "linux":
-            # These properties are available on MacOS, but not on Windows or Ubuntu.
             # On other Unix systems (such as FreeBSD), the following attributes may be
```
aminalaee, line 352: And same comment applies to st_birthtime since it can be available in FreeBSD.

tests/stubtest_allowlists/win32-py312.txt
```diff
@@ -5,11 +5,3 @@ asyncio.windows_events.IocpProactor.finish_socket_func
 asyncio.windows_events.IocpProactor.recvfrom_into
 msvcrt.GetErrorMode
 ntpath.isdir
-os.get_blocking
-os.listdrives
-os.listmounts
-os.listvolumes
-os.path.isdir
```
AlexWaygood, line 12: I'd just add this allowlist entry back for now to fix the stubtest error; it's different enough to the others that it's probably good to consider it in its own PR

stdlib/os/__init__.pyi
```diff
@@ -349,13 +354,12 @@ class stat_result(structseq[float], tuple[int, int, int, int, int, int, int, flo
...
```
AlexWaygood, line 357:
```suggestion
            # These properties are available on MacOS, but not Ubuntu.
            # On other Unix systems (such as FreeBSD), the following attributes may be
```

stdlib/os/__init__.pyi
```diff
@@ -341,6 +341,11 @@ class stat_result(structseq[float], tuple[int, int, int, int, int, int, int, flo
         if sys.version_info >= (3, 8):
             @property
             def st_reparse_tag(self) -> int: ...
+        if sys.version_info >= (3, 12):
+            @property
+            def st_birthtime(self) -> int: ...  # time of file creation in seconds
```
AlexWaygood, line 346: Hmm, not on my machine! ;)

```pycon
Python 3.13.0a0 (heads/main:d4cea794a7, Sep 21 2023, 13:24:35) [MSC v.1932 64 bit (AMD64)] on win32
>>> import os
>>> type(os.fstat(0))
<class 'os.stat_result'>
>>> type(os.fstat(0).st_birthtime)
<class 'float'>
```

```suggestion
            def st_birthtime(self) -> float: ...  # time of file creation in seconds
```

AlexWaygood: I don't have a macOS machine to test with, but it's possible that the existing definition here is also incorrect on this:

https://github.com/python/typeshed/blob/6dfa2852300ed610b93dbf1455fe628d394a1207/stdlib/os/__init__.pyi#L357-L358

aminalaee: I can test with a macOS machine.

aminalaee:
```shell
Python 3.10.10 (main, Mar 28 2023, 07:32:15) [Clang 14.0.0 (clang-1400.0.29.202)] on darwin
>>> import os
>>> type(os.fstat(0).st_birthtime)
<class 'float'>
````

Yeah looks like both definitions are wrong :D nice catch

AlexWaygood (approved):
Thanks!

PR #10750, opened 2023-09-22
[PyYAML] type the whole Emitter class

pcorpet:


stubs/PyYAML/yaml/emitter.pyi
```diff
@@ -26,94 +29,100 @@ class ScalarAnalysis:
...
-    allow_unicode: Any
-    best_indent: Any
-    best_width: Any
-    best_line_break: Any
-    tag_prefixes: Any
-    prepared_anchor: Any
-    prepared_tag: Any
-    analysis: Any
-    style: Any
+    encoding: str
```
JelleZijlstra, line 34:
```suggestion
    encoding: str | None
```
`__init__` sets it to None

JelleZijlstra (commented):
Thanks, just noticed one issue

stubs/PyYAML/yaml/emitter.pyi
pcorpet: Indeed. Done.

PR #6812, opened 2022-01-04
Make `os.path.join()` first parameter pos-only

AlexWaygood:
I was recently surprised to find that the mypy test suite [doesn't pass mypy](https://github.com/python/mypy/pull/11895) when run on Windows. This is because `posixpath.join()` and `ntpath.join()` have different names for the first parameter.

Given that both `posixpath` and `ntpath` are undocumented modules -- which only exist to provide platform-specific implementations for the `os.path` module -- I think it makes sense to make the first parameter positional-only for both functions. The whole point of the `os.path` module is that you can get path-related functions that will reliably work on all systems. That guarantee goes away if you use a keyword argument for the first parameter of `os.path.join()`.

PR #10751, opened 2023-09-22
Update os.path.isdir on Windows Python3.12+

aminalaee:
Following https://github.com/python/typeshed/pull/10748#discussion_r1334279493
cc @AlexWaygood 

Ref: https://github.com/python/cpython/blame/b28ffaa193efc66f46ab90d383279174a11a11d7/Modules/clinic/posixmodule.c.h#L1989

Seems like `os.path.isdir` has different signature `os.path.isdir(path)` compared to `os.path.isdir(s)` in netpath.py, not sure if this was intentional in CPython.

### DIFF
--- a/stdlib/ntpath.pyi
+++ b/stdlib/ntpath.pyi
@@ -1,5 +1,5 @@
 import sys
-from _typeshed import BytesPath, StrPath
+from _typeshed import BytesPath, FileDescriptorOrPath, StrPath
 from genericpath import (
     commonprefix as commonprefix,
     exists as exists,
@@ -7,7 +7,6 @@ from genericpath import (
     getctime as getctime,
     getmtime as getmtime,
     getsize as getsize,
-    isdir as isdir,
     isfile as isfile,
     samefile as samefile,
     sameopenfile as sameopenfile,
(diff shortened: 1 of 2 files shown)


JelleZijlstra:
This feels unfortunate and maybe should be fixed in CPython instead. Or perhaps the parameter should be made positional-only.

AlexWaygood:
> This feels unfortunate and maybe should be fixed in CPython instead. Or perhaps the parameter should be made positional-only.

Yeah, I think one of the two would be the better fix. Kinda the point of `os.path` is that the module is meant to be an abstraction that's fully cross-platform. So it makes things somewhat awkward when a function in `os.path` has a different signature on Windows than on other platforms. See #6812, where we had a similar issue, and tackled it by pretending the parameter was positional-only.

Here I think the best solution might be to raise this with CPython. Since the ntpath implementer is new, it seems unlikely to me that they intended to give it a different parameter name to the posix implementation.

aminalaee:
Yeah I think if this was not intentional in CPython, there's still time to change it. I will open an issue now.

JelleZijlstra:
The fix was merged into CPython and is planned to go into 3.12 eventually, so I think we shouldn't change anything here. Thanks for uncovering this issue!

PR #10753, opened 2023-09-23
Fix `contextlib.asynccontextmanager` to work with coroutine functions

sobolevn:
Right now a code from the docs: https://docs.python.org/3/library/contextlib.html#contextlib.asynccontextmanager does not type-check.

Example: https://mypy-play.net/?mypy=latest&python=3.11&gist=f20e230fa7b1af15064b2828dfd7dfd7

We are also required to have `AsyncGenerator` annotation here, because `.athrow()` method is used in this function: https://github.com/python/cpython/blob/e8be0c9c5a7c2327b3dd64009f45ee0682322dcb/Lib/contextlib.py#L226 (since 3.8: https://github.com/python/cpython/blob/3.8/Lib/contextlib.py#L189)

Runtime:

```python
>>> from typing import AsyncGenerator
>>> from contextlib import asynccontextmanager
>>> import asyncio
>>> @asynccontextmanager
... async def async_context() -> AsyncGenerator[str, None]:
...     yield 'example'
... 
>>> async def main():
...     async with async_context() as ac:
...         print(ac)
... 
>>> asyncio.run(main())
example
```

So, I went with the fix that also has some tests.

### DIFF
--- a/test_cases/stdlib/check_contextlib.py
+++ b/test_cases/stdlib/check_contextlib.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
-from contextlib import ExitStack
+from contextlib import ExitStack, asynccontextmanager
+from typing import AsyncIterator, Awaitable
 from typing_extensions import assert_type
 
 
@@ -18,3 +19,17 @@ with stack as cm:
     assert_type(cm, ExitStack)
 with thing as cm2:
     assert_type(cm2, Thing)
+
+
+@asynccontextmanager
+async def async_context() -> AsyncGenerator[str, None]:
+    yield 'example'
+
+
+async def async_gen() -> AsyncGenerator[str, None]:
+    yield 'async gen'
+
+
+@asynccontextmanager
+def async_cm_func() -> AsyncGenerator[str, None]:
+    return async_gen()
--- a/test_cases/stdlib/check_contextlib.py
+++ b/test_cases/stdlib/check_contextlib.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 from contextlib import ExitStack, asynccontextmanager
-from typing import AsyncIterator, Awaitable
+from typing import AsyncGenerator
 from typing_extensions import assert_type
 
 
--- a/stdlib/contextlib.pyi
+++ b/stdlib/contextlib.pyi
@@ -2,7 +2,7 @@ import abc
 import sys
 from _typeshed import FileDescriptorOrPath, Unused
 from abc import abstractmethod
-from collections.abc import AsyncGenerator, Awaitable, Callable, Coroutine, Generator, Iterator
+from collections.abc import AsyncGenerator, AsyncIterator, Awaitable, Callable, Coroutine, Generator, Iterator
 from types import TracebackType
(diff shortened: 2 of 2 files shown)


sobolevn:
Most errors are about `AsyncIterator` -> `AsyncGenerator` change. Examples: 
- https://github.com/psycopg/psycopg/blob/0b8379bc93fbd4a86abe10c0d8d648cfa964eb08/psycopg/psycopg/cursor_async.py#L222-L229
- https://github.com/psycopg/psycopg/blob/0b8379bc93fbd4a86abe10c0d8d648cfa964eb08/psycopg/psycopg/connection_async.py#L295-L300
- https://github.com/nolar/kopf/blob/538df59b88d1aab7b985d703483497f73c6c4783/kopf/_core/actions/execution.py#L190-L192

AlexWaygood:
> Most errors are about `AsyncIterator` -> `AsyncGenerator` change.

Could you defer that to a separate PR, so we can weigh it up on its own?

sobolevn:
Technically, this line is not quite correct as well: https://github.com/python/typeshed/blob/e1b60060ec5715ac34c459d1b116fe65610e8ba6/stdlib/contextlib.pyi#L76

Because CPython calls `self.gen.throw()` here: https://github.com/python/cpython/blob/e8be0c9c5a7c2327b3dd64009f45ee0682322dcb/Lib/contextlib.py#L159

AlexWaygood:
> Technically, this line is not quite correct as well:

Indeed. Discussed _extensively_ in #7430 (and, before that, #2773) :)

test_cases/stdlib/check_contextlib.py
```diff
@@ -18,3 +19,17 @@ class Thing(ExitStack):
...
+    yield "example"
+
+
+async def async_gen() -> AsyncGenerator[str, None]:
+    yield "async gen"
+
+
+@asynccontextmanager
+def async_cm_func() -> AsyncGenerator[str, None]:
+    return async_gen()
```
AlexWaygood, line 35: All these test cases appear to pass on `main`

AlexWaygood:
This fails mypy currently:

```py
from contextlib import asynccontextmanager
from typing import AsyncIterator, AsyncGenerator

@asynccontextmanager
async def db_connection() -> AsyncGenerator[None, None]: ...
```

But this doesn't:

```py
from contextlib import asynccontextmanager
from typing import AsyncIterator, AsyncGenerator

@asynccontextmanager
async def db_connection() -> AsyncGenerator[None, None]:
    yield None
```

I think this is because of the rules laid out in https://mypy.readthedocs.io/en/latest/more_types.html#asynchronous-iterators.

test_cases/stdlib/check_contextlib.py
sobolevn: Yes, mypy already knows how to solve this on `main` 👍

AlexWaygood: > Yes, mypy already knows how to solve this on `main` 👍

Then I'm confused. What's the problem we're trying to fix, if these tests all already pass?

sobolevn: Yeap, since it was fixed on `main`, we can drop this fix from `typeshed`

PR #10754, opened 2023-09-23
Revert "Do a dry run of stubsabot on PRs touching `scripts/` or `tests/`"

AlexWaygood:
Reverts python/typeshed#10508. It looks like doing a dry run of stubsabot on every PR touching `tests/` or `scripts/` is causing us to exceed our rate limit for GitHub's API: https://github.com/python/typeshed/actions/runs/6285081126/job/17067070901?pr=10724

### DIFF
--- a/.github/workflows/typecheck_typeshed_code.yml
+++ b/.github/workflows/typecheck_typeshed_code.yml
@@ -1,5 +1,4 @@
-# This workflow is for testing typeshed's scripts and tests themselves
-name: Meta-tests
+name: Typecheck-typeshed-code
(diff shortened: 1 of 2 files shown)


AlexWaygood:
It's possible https://github.com/python/typeshed/actions/runs/6285081126/job/17067070901?pr=10724 was a flake due to lots of commits being pushed to #10724, all of which ended up triggering this workflow.

AlexWaygood:
I'll close this for now, but we can still revert the PR if we see this happen again

PR #10756, opened 2023-09-24
Make syslog `setlogmask`, `LOG_MASK` and `LOG_UPTO` positional-only

aminalaee:
Runtime argument has changed for the functions in Python3.12+.
Not sure if it makes sense to follow https://github.com/python/typeshed/pull/6812  and make it positional-only here or have different signatures.


AlexWaygood (approved):
Thanks! Weird that stubtest wasn't emitting an error on Python <3.12, but it does indeed seem like these have always been positional-only

PR #10757, opened 2023-09-24
Add `os.unshare` clone flags for Linux, Python 3.12+

aminalaee:
Ref: https://docs.python.org/3.12/library/os.html#os-unshare-clone-flags


stdlib/os/__init__.pyi
```diff
@@ -1056,3 +1056,19 @@ if sys.version_info >= (3, 12) and sys.platform == "win32":
...
+
+if sys.version_info >= (3, 12) and sys.platform == "linux":
+    CLONE_FILES: int
+    CLONE_FS: int
+    CLONE_NEWCGROUP: int
+    CLONE_NEWIPC: int
+    CLONE_NEWNET: int
+    CLONE_NEWNS: int
+    CLONE_NEWPID: int
+    CLONE_NEWTIME: int
```
aminalaee, line 1068: Strange that this is not available in runtime, though testing locally I can confirm it works:

```shell
Python 3.13.0a0 (heads/main-dirty:e8be0c9c5a, Sep 23 2023, 14:07:37) [GCC 12.3.0] on linux
>>> import os, posix
>>> os.CLONE_NEWTIME
128
>>> posix.CLONE_NEWTIME
128
```

AlexWaygood: In this kind of situation, the best thing to do is just to add an allowlist entry, with a comment explaining that it's available on some linux builds but not on others. It's fine to have allowlist entries if there's a good reason to have them 👍

aminalaee: I tried to take a look, the `CLONE_NEWTIME ` seems to be added in Kernel 5.6 but current Github image is already on Kernel 6.
Another strange thing is that `CLONE_NEWTIME ` was not in the allowlist originally, now I can add it, not sure if that is relevant.

AlexWaygood (approved):
Thanks!

PR #10758, opened 2023-09-24
Add definition for visit_MatchSingleton to ast.NodeVisitor

plokmijnuhby:
It appears this method was somehow missed off.


AlexWaygood (approved):
thanks!

PR #10761, opened 2023-09-24
Change types in "builtins" and "types" modules to have __new__ instead of __init__, where necessary

plokmijnuhby:
This means that calls to eg `range.__new__(range, 0)` will be checked correctly. The most common case for this is if a subclass of range makes its own `__new__` method, and calls `super().__new__` (although some of the classes affected by this PR cannot be subclassed).

I have deliberately ignored the reversed class, because it is the subject of its own ongoing PR, #10655.


srittau (approved):
Thanks!

Issue #10764, opened 2023-09-25
Typing of requests.HTTPError possibly too strict


moltob:
Recent contributions https://github.com/python/typeshed/pull/8989 and https://github.com/python/typeshed/pull/10740 improved the typing of `requests.HTTPError`, so now we have:

https://github.com/python/typeshed/blob/5b8193b9fd437cca0ad00609c367f5188798c870/stubs/requests/requests/exceptions.pyi#L13-L16

This is causing issues in our code as it appears to be incompatible to the way requests itself is using the type:

Now the `request` argument is mandatory. However, requests itself does not do it this way, e.g. see the implementation of [`Response.raise_for_status`](https://github.com/psf/requests/blob/881281250f74549f560408e5546d95a8cd73ce28/src/requests/models.py#L1018-L1019):

![image](https://github.com/python/typeshed/assets/6240478/39b5dbb7-16e8-49dd-8c7a-ead5bc0ac040)

Secondly, the type is now restricted to be `Request`.  In particular this prevents assigning a `PreparedRequest` (which is not derived from `Request`). However, this is exactly what requests is doing in [`RequestException.__init__`](https://github.com/psf/requests/blob/881281250f74549f560408e5546d95a8cd73ce28/src/requests/exceptions.py#L21-L23), in addition to explicitly allowing _not_ to pass a `request` argument:

![image](https://github.com/python/typeshed/assets/6240478/98a4a6c6-0521-48cb-a22d-ba5d0d5da790)

Please reconsider the strict typing.

@JelleZijlstra @srittau

akx:
It looks like the type of `request` in the exception should be `Request | PreparedRequest`, since `response.request` is typed to be a `PreparedRequest`; trying to manually do `HTTPError(response=resp, request=resp.request)` now raises 

> Argument "request" to "HTTPError" has incompatible type "PreparedRequest"; expected "Request"

srittau:
Thanks to @akx's PR, the `request` type is now relaxed. But per the example given I think the assumptions in #8989 and #10740 are wrong and we need to allow `None` values (and defaults) again for `request` and `response`.

hsheth2:
I opened https://github.com/python/typeshed/pull/10776 to fix this - @srittau let me know what you think.

srittau:
[...]

PR #10766, opened 2023-09-25
[requests] Loosen HTTPError's `Response` type

akx:
`requests` itself allows a `None` type for `Response`; it will internally attempt to read it from the `request` passed.

There is thus no strict guarantee that a `HTTPError` will have a truthy `request`.

Fixes https://github.com/python/typeshed/issues/10764
Follows up on https://github.com/python/typeshed/pull/10740

See https://github.com/psf/requests/blob/881281250f74549f560408e5546d95a8cd73ce28/src/requests/exceptions.py#L17-L24 for the relevant inference code.


akx:
Ah, actually, well, the ecosystem test speaks for itself – this is probably not a good idea.

moltob:
And this was also the wrong thing you loosened up. :-) The `request` is the issue, not the `response`.

PR #10767, opened 2023-09-25
[requests] Allow PreparedRequest for RequestException(request=...)

akx:
Since [`requests` can and will internally read a `request` off the `response` object passed in to `RequestException`](https://github.com/psf/requests/blob/881281250f74549f560408e5546d95a8cd73ce28/src/requests/exceptions.py#L17-L24), and that `request`'s type is `PreparedRequest`, not `Request`, this makes the types for the exception reflect that.

This allows type-safely constructing an exception with `HTTPError(response=resp, request=resp.request)`.

Refs https://github.com/python/typeshed/issues/10764#issuecomment-1733135906
Follows up on https://github.com/python/typeshed/pull/10740


srittau (approved):
Thanks!

osherdp:
@akx @srittau, any idea why it failed to detect it for ``HTTPError``?

For the following code:
```
import requests
response = requests.get("https://google.com")
raise requests.exceptions.HTTPError("message", response=response)
```

I'm getting:
```
script.py:4: error: Missing named argument "request" for "HTTPError"  [call-arg]
```

srittau:
@osherdp Please see #10764.

PR #10771, opened 2023-09-25
Add `os.splice` and flags on Linux for Python 3.10+

aminalaee:
Ref: https://docs.python.org/3/library/os.html#os.splice


stdlib/os/__init__.pyi
```diff
@@ -1062,9 +1062,15 @@ if sys.version_info >= (3, 10) and sys.platform == "linux":
...
     EFD_SEMAPHORE: int
+    SPLICE_F_MORE: int
+    SPLICE_F_MOVE: int
+    SPLICE_F_NONBLOCK: int
     def eventfd(initval: int, flags: int = 524288) -> FileDescriptor: ...
     def eventfd_read(fd: FileDescriptor) -> int: ...
     def eventfd_write(fd: FileDescriptor, value: int) -> None: ...
+    def splice(
+        src: FileDescriptor, dst: FileDescriptor, count: int, offset_src: int | None = ..., offset_dst: int | None = ...
+    ) -> int: ...
```
srittau, line 1073: At least in Python 3.11 there seems to be a `flags` argument:

```
Signature: os.splice(src, dst, count, offset_src=None, offset_dst=None, flags=0)
Docstring:
Transfer count bytes from one pipe to a descriptor or vice versa.

  src
    Source file descriptor.
  dst
    Destination file descriptor.
  count
    Number of bytes to copy.
  offset_src
    Starting offset in src.
  offset_dst
    Starting offset in dst.
  flags
    Flags to modify the semantics of the call.

If offset_src is None, then src is read from the current position;
respectively for offset_dst. The offset associated to the file
descriptor that refers to a pipe must be None.
```

aminalaee: yeah it's not documented.

aminalaee: I can open an issue in CPython for this to see if it's a mistake.

srittau: I think this is mainly a documentation oversight. After all, the flags are also defined. We should include the argument in any case, but I think a CPython issue is appreciated.

srittau (approved):

Issue #10772, opened 2023-09-25
types-jsonschema 4.19.0.0 cannot be used with mypy on Python 3.7


GavBurke:
There's not much info I can give in this "bug report", so hopefully others seen the same issue and can chime in, but after our nightly pipeline upgrade it's types-jsonschema from 4.17.0.10 to 4.19.0.0 on Saturday night, our pipeline now breaks with mypy reporting `/home/gavin/.local/lib/python3.7/site-packages/referencing/_core.py:188: error: invalid syntax  [syntax]
Found 1 error in 1 file (errors prevented further checking)`

Downgrading types-jsonschema to 4.17.0.10 then fixes this issue

AlexWaygood:
`types-jsonschema 4.19.0.0` depends on the `referencing` library following #10583. The `referencing` library is incompatible with Python 3.7, as it uses PEP-570 syntax (but the library didn't state as such in its `pyproject.toml` or `setup.py` file in its early versions, meaning that pip will happily install it on Python 3.7, unfortunately).

AlexWaygood:
Once https://github.com/typeshed-internal/stub_uploader/pull/103 is merged, we will be able to push a release of `types-jsonschema` to PyPI that properly declares that it can only be used with Python 3.8+. Not sure what's to be done with the already-released version of `types-jsonschema`, though.

AlexWaygood:
> We could yank release 14.9.0.0.

Yeah. Suggested plan of action:
1. Revert https://github.com/python/typeshed/pull/10583 temporarily
2. Yang release 4.19.0.0 from PyPI
3. Merge https://github.com/typeshed-internal/stub_uploader/pull/103
4. Revert the revert of https://github.com/python/typeshed/pull/10583, but add `requires_python = ">= 3.8"` to the `METADATA.toml` file

srittau:
This seems to have worked: https://pypi.org/project/types-jsonschema/#files

@ilevkivskyi Could you yank types-jsonschema 4.19.0.0 and 4.19.0.1?

GavBurke:
> This seems to have worked: https://pypi.org/project/types-jsonschema/#files
> 
> @ilevkivskyi Could you yank types-jsonschema 4.19.0.0 and 4.19.0.1?

Any timeline on the yanking of the 4.19.x python 3.7 packages??

ilevkivskyi:
> Could you yank types-jsonschema 4.19.0.0 and 4.19.0.1?

Done.

[...]

PR #10773, opened 2023-09-25
Revert "Bump jsonschema to 4.19.*"

AlexWaygood:
Reverts python/typeshed#10583. See #10772: the PR makes `types-jsonschema` incompatible with Python 3.7, since `types-jsonschema` now depends on the `referencing` library, which requires Python 3.8+ (but _didn't_ state as much in its `setup.py`/`pyproject.toml` in early versions released to PyPI).

We should be able to revert the revert once https://github.com/typeshed-internal/stub_uploader/pull/103 is merged, which will allow us to properly declare that the stubs package requires Python 3.8+.


srittau:
I don't think this is a good idea. I'm not sure the stub uploader will handle this case gracefully.

AlexWaygood:
> I don't think this is a good idea. I'm not sure the stub uploader will handle this case gracefully.

Ah. Well, you know better than me when it comes to that.

What will the stub-uploader do if we yank v4.19.0.0 without doing this revert? Will it simply stop pushing new versions to PyPI for the time being, or will it push v4.19.0.0 again?

srittau:
I'll just merge the changes to the stub uploader and kick off a new release of types-jsonschema. Then we should be safe.

AlexWaygood:
Okay, great!

PR #10776, opened 2023-09-25
[requests] loosen HTTPError constructor

hsheth2:
Follow up on https://github.com/python/typeshed/pull/10766, should help with https://github.com/python/typeshed/issues/10764


stubs/requests/requests/exceptions.pyi
```diff
@@ -16,7 +16,7 @@ class JSONDecodeError(InvalidJSONError): ...
 class HTTPError(RequestException):
     request: Request | PreparedRequest
```
srittau, line 17: We also need to add `None` to `request`.

srittau (changes requested):
Thanks, we need one more change.

stubs/requests/requests/exceptions.pyi
hsheth2, line 17:
```suggestion
    request: Request | PreparedRequest | None
```

srittau (approved):
Thanks!

PR #10780, opened 2023-09-26
fix(redis): make xadd fields use SupportsItems

sileht:
Using an InvariantMapping is wrong, the type of each values of the dict
can be different.

This change uses the SupportsItems Protocol instead to match as much as possible the
code.


sileht:
> Why?

You're too fast :) I didn't finish to write the description correctly. 

That said, I just saw a discussion to drop types-redis, so not sure you continue to accept change to this typing.

(btw, redis 5.x builtin typing is far to be usable due to https://github.com/redis/redis-py/issues/2933)

srittau:
While I'm fine with replacing `Mapping` with an appropriate covariant protocol (we have a few of those available in `_typeshed`),we not should just revert this commit and replace the existing annotations with no annotations.

sileht:
I propose another solution.

srittau:
Hmm, redis-py seems to explicitly check for a dict:

https://github.com/redis/redis-py/blob/cc4bc1a544d1030aec1696baef2861064fa8dd1c/redis/commands/core.py#L3547

@sileht Can you give an example where type checkers have problems with the current annotations?

AlexWaygood:
> Hmm, redis-py seems to explicitly check for a dict:

FWIW, it does look to me like the current annotations we have are the worst of both worlds. We don't get nice covariant types, since `Mapping` is invariant in its key types -- but we're also not particularly accurate, since `redis-py` does actually require a `dict` (for some reason) rather than just any old `Mapping`. So I think we should either change the stubs to be more accurate (`Mapping` -> `dict`), or change them to be more user-friendly (`Mapping` -> `SupportsItems`, or some other protocol). The status quo doesn't seem great.

I'd also be curious for @sileht's use case where the current annotations are causing a problem, though; it's hard to solve a problem if we don't know exactly what the problem is.

sileht:
Currently, you can't write thing like:

```
class RedisStreamData(typing.TypedDict):
    foo: str
    bar: bytes

redis.xadd("stream", RedisStreamData({"foo": "bar", "bar": b"foo"}))
```

srittau (commented):
Considering @sileht's valid use case and the fact that we prefer false negatives over false positives, I think this change is the best we can do.

@sileht Could you add a comment before the `fields` argument, something like "Only accepts dict objects, but for variance reasons we use a looser annotation." (Maybe you can come up with a better wording.)

stubs/redis/redis/commands/core.pyi
```diff
@@ -863,7 +863,8 @@ class StreamCommands:
     def xadd(
         self,
         name: KeyT,
-        fields: Mapping[bytes | memoryview | str | float, bytes | memoryview | str | float],
+        # Only accepts dict objects, but for variance reasons we use a looser annotation
+        fields: SupportsItems[bytes | memoryview | str | float, bytes | memoryview | str | float],
```
AlexWaygood, line 867: Your use case will still not pass type-checking, even with this change, as `TypedDict`s have `dict_items[str, object]` as a return type their `.items()` method, which is not a subtype of `AbstractSet[[tuple[bytes | memoryview | str | float, bytes | memoryview | str | float]]`.

AlexWaygood: If you need to be able to pass a `TypedDict` object for the `fields` parameter, you'll have to use `Any` as the second argument:

```suggestion
        fields: SupportsItems[bytes | memoryview | str | float, Any],
```

AlexWaygood: I discovered that the fix didn't work by adding this test case:

```py
# This was added at file location stubs/redis/@tests/test_cases/check_xadd.py

import redis

class RedisStreamData(TypedDict):
    foo: str
    bar: bytes


def check_xadd(r: redis.Redis[str]) -> None:
    r.xadd("stream", RedisStreamData({"foo": "bar", "bar": b"foo"}))
```

And then running

```
python tests/regr_test.py redis -p3.8 --verbosity VERBOSE
```

The test passes fails with your current PR, but passes if you change the `fields` annotations to  `SupportsItems[bytes | memoryview | str | float, Any]`

Daverball: Could we maybe do a little bit better by accepting `SupportsItems[bytes | memoryview | str | float, bytes | memoryview | str | float] | TypedDict`?

There are probably no valid use-cases besides `TypedDict`, where we can't restrict the value type.

Daverball: Actually I think the better solution might be `SupportsItems[bytes | memoryview | str | float, bytes | memoryview | str | float] | Mapping[str, bytes | memoryview | str | float]` since the value type in Mapping should be covariant, it's only really the key type that's tripping us up here.

AlexWaygood: @Daverball, `Mapping[str, bytes | memoryview | str | float]` is a subtype of `SupportsItems[bytes | memoryview | str | float, bytes | memoryview | str | float]`, so the second item in your union there is redundant. But you have exactly the same issue when it comes to using a `TypedDict` here: it is impossible to construct a `TypedDict` that is a subtype of `Mapping[str, bytes | memoryview | str | float]`.

Daverball: For some reason my assumption was that for `Mapping` the plugin would try to match according to the `TypedDict` members. But I guess I never actually try to mix `TypedDict` with a  regular `Mapping`.

But in that case the argument that we should change the annotation in order to allow `TypedDict` doesn't really hold water to me, since the choice of `object` rather than `Any` seems intentional me, in order to avoid confusion when mixing `Mapping` with `TypedDict`, which shouldn't really happen, either you use a `TypedDict` all the way through or you stick with a `Mapping`.

AlexWaygood: > But in that case the argument that we should change the annotation in order to allow `TypedDict` doesn't really hold water to me, since the choice of `object` rather than `Any` seems intentional me, in order to avoid confusion when mixing `Mapping` with `TypedDict`, which shouldn't really happen, either you use a `TypedDict` all the way through or you stick with a `Mapping`.

Yeah, maybe. The main point I'm making is that the PR as it currently stands does not fix the problem it's attempting to fix. That seems like a pretty relevant thing to consider ;) Especially considering that this PR makes the types arguably less accurate, given that at runtime `redis-py` really does need a `dict` instance, not any old `Mapping` or `SupportsItems` object.

AlexWaygood: Put another way: if this PR isn't actually fixing any real-world use case, then we shouldn't make any change here

Daverball: Yes, I totally agree, although the fact that the value type on `Mapping` isn't covariant is still a valid problem and would break code where the `dict` isn't passed in directly as a literal.

I would harbor a guess that the reason they chose `Mapping` was pragmatism, so they didn't have to write a union of the 16+ or so permutations of all the possible dicts that can be passed in, but forgot, that the key type on `Mapping` isn't covariant either.

AlexWaygood: > although the fact that the value type on `Mapping` isn't covariant

The value type on Mapping _is_ covariant, it's the key type that isn't :-)

https://github.com/python/typeshed/blob/4dad5311d56e0b90ab82756a8c2a9ce645a788b1/stdlib/typing.pyi#L633

The issue about `TypedDict`s all being `Mapping[str, object]`s is a distinct issue to `Mapping` being invariant in its key type

> ...is still a valid problem and would break code where the `dict` isn't passed in directly as a literal.

Maybe, but it becomes a purely theoretical problem :-)

Daverball: > The value type on Mapping _is_ covariant, it's the key type that isn't :-)

Yes, you're right, that was a typo/brain fart, thanks for catching that. :)
In the second paragraph I'm correctly saying that the key type isn't covariant on `Mapping`.

> Maybe, but it becomes a purely theoretical problem :-)

I mean I wouldn't really call it purely theoretical, even if the most common way to use this function, is with dict literals inside the function call, it still is a practical problem with the annotation, if you only wanted to support literals, we could just go back to using `dict` and take the better accuracy as a win.

sileht: So `dict[Any, Any]` is the best trade-off here?

Daverball: It's certainly one of the options, what's considered the best is a matter of personal preference to some degree and how the library is most commonly used if you want to take a more objective stance.

I would personally changing the annotation in one of three possible ways:
1. `dict[bytes | memoryview | str | float, bytes | memoryview | str | float]`
    This is fully type safe and works with dict-literals, which is the most common use-case, but it isn't covariant, so if an intermediary variable is used, it will not work unless it uses this same exact type.
4. `SupportsItems[bytes | memoryview | str | float, bytes | memoryview | str | float]`
    This is covariant and still fully type safe for all the individual keys and values, since the implementation never actually modifies either of them, and thus a little bit more user-friendly. But it doesn't properly reject mappings that aren't a subclass of `dict`.
6. `dict[Any, Any]`
    This properly rejects non-dict but provides no type safety for the keys and values, but is also more user-friendly.

JelleZijlstra: I feel the best option is `SupportsItems[bytes | memoryview | str | float, Any]`. That gives us type safety for the key and still accepts TypedDicts (which are compatible with `Mapping[str, object]`).

JelleZijlstra: I feel the best option is `SupportsItems[bytes | memoryview | str | float, Any]`. That gives us type safety for the key and still accepts TypedDicts (which are compatible with `Mapping[str, object]`).

Daverball: > I feel the best option is `SupportsItems[bytes | memoryview | str | float, Any]`. That gives us type safety for the key and still accepts TypedDicts (which are compatible with `Mapping[str, object]`).

It doesn't feel right to me to change the stubs specifically to allow `TypedDict` inputs, when we don't seem to do that anywhere else. If we wanted people to be able to use `TypedDict` interchangeably with `Mapping` arguments, shouldn't we change `TypedDict` instead and change the value type from `object` to `Any`? It seems backwards to me. Especially considering it's actually used interchangeably with `dict` here, and not just `Mapping`, due to the `isinstance` check at runtime.

Either seems like a bad idea though, because it's easy for people to get tricked into thinking they get type-safety on the values, since it's not very well documented that type checkers forego inferring the value type if you don't access a specific key with a string literal on a `TypedDict`.

P.S. I don't really understand the performance arguments type checkers make to justify this behavior. Wouldn't type checkers just be able to build a value type union while they're parsing/inferring the `TypedDict` the first time and progressively build the key/value type mapping, generate the signatures for `items()` etc. statically from that type union and store it on the resulting type? Even with subclassing, they could just retrieve the value type union from the parent class, once that's finished parsing/inferring. It doesn't seem very expensive to do this once per `TypedDict` or subclass thereof, not more expensive than figuring out the value type from the key type on item access anyhow

stubs/redis/redis/commands/core.pyi
```diff
@@ -928,7 +929,8 @@ class AsyncStreamCommands:
     async def xadd(
         self,
         name: KeyT,
-        fields: Mapping[bytes | memoryview | str | float, bytes | memoryview | str | float],
+        # Only accepts dict objects, but for variance reasons we use a looser annotation
+        fields: SupportsItems[bytes | memoryview | str | float, bytes | memoryview | str | float],
```
JelleZijlstra, line 933:
```suggestion
        fields: SupportsItems[bytes | memoryview | str | float, Any],
```
As discussed in the long thread

AlexWaygood (approved):

JelleZijlstra:
Pyright is now unhappy because xadd doesn't have a return type. If I'm reading https://redis.io/commands/xadd/ right, it should return a str most of the time, but possibly None if `nomkstream` is True.

JelleZijlstra (approved):
That's the easier solution :)

AlexWaygood:
> Pyright is now unhappy because xadd doesn't have a return type. If I'm reading [redis.io/commands/xadd](https://redis.io/commands/xadd/) right, it should return a str most of the time, but possibly None if `nomkstream` is True.

Oh I just ignored it for now. But if you have a better fix, feel free to push it :)

PR #10782, opened 2023-09-26
Skip all parts of the daily test on forks of typeshed

### LINKED ISSUE #10781: Daily tests are running on forks
I just want to let you know that automated daily tests are running on forks of typeshed. I didn't know this until I got a notification about a failing test here https://github.com/hamdanal/typeshed/actions/runs/6306583853. Maybe automated tests should not run on forks? Only tests triggered by changes/commits/PRs would be relevant to forks (I think!).

AlexWaygood:
We skip 2/3 jobs in the daily test when running on forks:

https://github.com/python/typeshed/blob/b23042e854e79b54c4d2c609cf1e12a2357335a9/.github/workflows/daily.yml#L33

https://github.com/python/typeshed/blob/b23042e854e79b54c4d2c609cf1e12a2357335a9/.github/workflows/daily.yml#L59

But we missed this condition for the stub-uploader job

Fixes #10781

### DIFF
--- a/.github/workflows/daily.yml
+++ b/.github/workflows/daily.yml
@@ -98,8 +98,10 @@ jobs:
           fi
 
           $PYTHON_EXECUTABLE tests/stubtest_third_party.py --specified-platforms-only --num-shards 4 --shard-index ${{ matrix.shard-index }}
+
   stub-uploader:
     name: Run the stub_uploader tests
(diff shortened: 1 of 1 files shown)


hamdanal (approved):
LGTM.
As you pointed out, same thing is done in other jobs so this should work.
(Thanks for fixing the issue faster than it took me to file it :D)

PR #10784, opened 2023-09-27
Update `tty` to python 3.12

sobolevn:
Source: https://github.com/python/cpython/blob/3.12/Lib/tty.py

### DIFF
--- a/stdlib/tty.pyi
+++ b/stdlib/tty.pyi
@@ -1,5 +1,5 @@
 import sys
-from typing import IO
+from typing import IO, Any
 from typing_extensions import TypeAlias
 
 if sys.platform != "win32":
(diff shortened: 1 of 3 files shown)


stdlib/tty.pyi
```diff
@@ -17,3 +17,12 @@ if sys.platform != "win32":
     CC: int
     def setraw(fd: _FD, when: int = 2) -> None: ...
     def setcbreak(fd: _FD, when: int = 2) -> None: ...
+
+    if sys.version_info >= (3, 12):
+        __all__ += ["cfmakeraw", "cfmakecbreak"]
```
srittau, line 22: Nit: My quick random sampling shows me that we usually move these `__all__ +=` constructs to the top of the file, even if it means a duplicated version/platform check. Personally, I find this better too, since you can easier spot the contents of `__all__` for all branches.

srittau (approved):
LGTM, but one optional nit.

PR #10785, opened 2023-09-27
[tty] py312: Fix return types of set(raw|cbreak)

srittau:
Also add `termios._AttrReturn` type alias to be used in tty

### DIFF
--- a/stdlib/termios.pyi
+++ b/stdlib/termios.pyi
@@ -3,10 +3,12 @@ from _typeshed import FileDescriptorLike
 from typing import Any
 from typing_extensions import TypeAlias
 
-if sys.platform != "win32":
-    # Must be a list of length 7, containing 6 ints and a list of NCCS 1-character bytes or ints.
-    _Attr: TypeAlias = list[int | list[bytes | int]]
+# Must be a list of length 7, containing 6 ints and a list of NCCS 1-character bytes or ints.
+_Attr: TypeAlias = list[int | list[bytes | int]]
+# Same as _Attr for return types; we use Any to avoid a union.
+_AttrReturn: TypeAlias = list[Any]
 
+if sys.platform != "win32":
     B0: int
     B1000000: int
     B110: int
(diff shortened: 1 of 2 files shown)


stdlib/termios.pyi
```diff
@@ -3,10 +3,12 @@ from _typeshed import FileDescriptorLike
 from typing import Any
 from typing_extensions import TypeAlias
 
-if sys.platform != "win32":
-    # Must be a list of length 7, containing 6 ints and a list of NCCS 1-character bytes or ints.
-    _Attr: TypeAlias = list[int | list[bytes | int]]
+# Must be a list of length 7, containing 6 ints and a list of NCCS 1-character bytes or ints.
+_Attr: TypeAlias = list[int | list[bytes | int]]
```
AlexWaygood, line 7: To avoid issues with invariance...

```suggestion
_Attr: TypeAlias = list[int] | list[list[bytes]] | list[int | list[bytes]] | list[list[int]] | list[int | list[int]] | list[list[int | bytes]] | list[int | list[int | bytes]]
```

stdlib/tty.pyi
```diff
@@ -1,10 +1,16 @@
 import sys
+from termios import _AttrReturn
 from typing import IO
 from typing_extensions import TypeAlias
 
 if sys.platform != "win32":
     __all__ = ["setraw", "setcbreak"]
 
+    if sys.version_info >= (3, 12):
+        _SetReturn: TypeAlias = _AttrReturn
```
sobolevn, line 10: `Set` here sounds like `set()`. Maybe some other name?

stdlib/termios.pyi
srittau: Good catch, although `list[int]` and `list[list[...]]` are not valid as this is a fixed-size list with specific types for specific positions.

srittau:
@sobolevn I've also replaced `_Mode` with `termio._Attr` as I believe these represent the same thing?

stdlib/tty.pyi
```diff
@@ -1,12 +1,17 @@
 import sys
-from typing import IO, Any
+from termios import _Attr, _AttrReturn
```
AlexWaygood, line 2: These names are a bit ambiguous when imported — maybe alias them?

```suggestion
from termios import _Attr as _TermiosAttr, _AttrReturn as _TermiosAttrReturn
```

JelleZijlstra:
Submitted python/cpython#101028 to document some of these changes better

PR #10787, opened 2023-09-28
Type all `@property` in openpyxl

### LINKED ISSUE #10744: openpyxl Worksheet columns is incorrectly typed

Avasam:
- Overlaps #9511 to reduce changes there.
- Closes #10744 & closes #10745 (superseeds it)
- Typed all `@property` in openpyxl. And some related functions used by said properties (may as well do it now instead of having to redo the type analysis later).
- Validated that no `@property` or `@abstractproperty` is missing from source.
- Excluded worksheet-related properties in Workbook class. Those come with their own set of issues and difficulties.
- Changed a `bytes` to `ReadableBuffer` (relates to #9006)


srittau (approved):
I didn't compare this to the implementation, but this all looks reasonable to me.

PR #10790, opened 2023-09-28
Limit `os.sendfile` arguments based on platform

aminalaee:
According to the docs https://docs.python.org/3.8/library/os.html#os.sendfile and the [manpage](https://man7.org/linux/man-pages/man2/sendfile.2.html#:~:text=sendfile()%20copies%20data%20between,to%20and%20from%20user%20space). the `headers`, `trailers` and `flags` are not available on Linux.


srittau (approved):
Thanks!

PR #10791, opened 2023-09-28
Make `os.setresuid` and `os.setresgid` positional-only

aminalaee:
https://docs.python.org/3.12/library/os.html#os.setregid

I was curious if this was positional-only on Python 3.8 and seems like it is even though not documented:

```shell
Python 3.9.13+ (heads/3.9:66f4593280, Sep 28 2023, 10:59:39) 
[GCC 12.3.0] on linux
>>> import os, inspect
>>> inspect.signature(os.setresuid)
<Signature (ruid, euid, suid, /)>

Python 3.8.13+ (heads/3.8:51f1ae5ceb, Sep 28 2023, 11:00:22) 
[GCC 12.3.0] on linux
>>> import os, inspect
>>> inspect.signature(os.setresuid)
<Signature (ruid, euid, suid, /)>
```

https://docs.python.org/3.10/library/os.html#os.setresuid


AlexWaygood (approved):
Nice, thank you! For Reasons, we only check whether or not arguments are positional-only on Python 3.10+, which is why we don't have the allowlist entries on Python 3.8 or 3.9 :)

PR #10792, opened 2023-09-28
Add `MAP_STACK` to `mmap` module

sobolevn:
Source:

```c
#ifdef MAP_STACK
    // Mostly a no-op on Linux and NetBSD, but useful on OpenBSD
    // for stack usage (even on x86 arch)
    ADD_INT_MACRO(module, MAP_STACK);
#endif
```

Docs:

```rst
    .. versionadded:: 3.11
       Added :data:`MAP_STACK` constant.
```

### DIFF
--- a/stdlib/mmap.pyi
+++ b/stdlib/mmap.pyi
@@ -16,8 +16,8 @@ if sys.platform == "linux":
     MAP_EXECUTABLE: int
     if sys.version_info >= (3, 10):
         MAP_POPULATE: int
-    if sys.version_info >= (3, 11):
-        MAP_STACK: int
+if sys.version_info >= (3, 11) and sys.platform != "win32" and sys.platform != "darwin":
+    MAP_STACK: int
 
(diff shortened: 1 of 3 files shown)


stdlib/mmap.pyi
```diff
@@ -16,6 +16,8 @@ if sys.platform == "linux":
     MAP_EXECUTABLE: int
     if sys.version_info >= (3, 10):
         MAP_POPULATE: int
+    if sys.version_info >= (3, 11):
+        MAP_STACK: int
```
AlexWaygood, line 20: The comment in the CPython source code seems to indicate that these also exist on some more exotic platforms -- we should probably do our best to account for those, even if we don't test for them in CI

```suggestion
if sys.version_info >= (3, 11) and sys.platform != "win32" and sys.platform != "darwin":
    MAP_STACK: int
```

AlexWaygood (approved):
Thanks!

PR #4727, opened 2020-10-28
Fix annotations for resource.prlimit()

cptpcrd:
The current annotations specify the `limits` argument as an `Optional[Tuple[int, int]]`; this is incorrect. The real `resource.prlimit()` function does not accept `None` for the `limits` argument; it is only possible to avoid specifying new resource limits by omitting `limits` completely.

PR #10794, opened 2023-09-28
Update `resource.prlimit` for Python 3.12+

aminalaee:
https://docs.python.org/3.12/library/resource.html#resource.prlimit


stdlib/resource.pyi
```diff
@@ -86,8 +85,5 @@ if sys.platform != "win32":
     def getrusage(__who: int) -> struct_rusage: ...
     def setrlimit(__resource: int, __limits: tuple[int, int]) -> None: ...
     if sys.platform == "linux":
-        @overload
-        def prlimit(pid: int, resource: int, limits: tuple[int, int]) -> tuple[int, int]: ...
-        @overload
-        def prlimit(pid: int, resource: int) -> tuple[int, int]: ...
+        def prlimit(__pid: int, __resource: int, __limits: tuple[int, int] | None = ...) -> tuple[int, int]: ...
```
AlexWaygood, line 88: According to the contributor in #4727, it's not possible to pass `None` to this function

```suggestion
        def prlimit(__pid: int, __resource: int, __limits: tuple[int, int] = ...) -> tuple[int, int]: ...
```

aminalaee: I tried that before, but fails:
```
error: resource.prlimit is inconsistent, runtime argument "limits" has a default value of type None, which is incompatible with stub argument type tuple[builtins.int, builtins.int]
```

Maybe it's changed in 3.12, I can check that.

AlexWaygood: Stubtest isn't always 100% correct on what's allowed at runtime. If necessary, we should keep the allowlist entry and add a comment saying that stubtest thinks `None` can be passed, but it's not actually allowed at runtime.

But yeah, I don't have a linux machine to test on, so please experiment and report back the results!! :D

aminalaee:
```shell
Python 3.11.4 (main, Jun  7 2023, 12:45:48) [GCC 11.3.0] on linux
>>> import resource
>>> resource.prlimit(1, 2, None)
TypeError: 'NoneType' object is not iterable

Python 3.12.0rc3+ (heads/3.12:538f505a37, Sep 28 2023, 14:02:27) [GCC 12.3.0] on linux
>>> import resource 
>>> resource.prlimit(1, 2, None)
PermissionError: [Errno 1] Operation not permitted
```

AlexWaygood: Okay, great, so it changed in 3.12. Probably needs to be this, then?

```suggestion
        if sys.version_info >= (3, 12):
            def prlimit(__pid: int, __resource: int, __limits: tuple[int, int] | None = ...) -> tuple[int, int]: ...
        else:
            def prlimit(__pid: int, __resource: int, __limits: tuple[int, int] = ...) -> tuple[int, int]: ...
```

AlexWaygood (approved):
Thanks! :D

PR #10797, opened 2023-09-28
Add type stubs for `pyjks`

intgr:
Adds type stubs for the `pyjks` Python package, for handling various Java and BouncyCastle keystore files.

* Documentation: https://pyjks.readthedocs.io/en/latest/
* GitHub: https://github.com/kurtbrose/pyjks

The `jks.bks` submodule is currently untyped, I can work on that after this initial PR is merged.

The `pyjks` upstream is fairly sta(b)le. Last commit was 3 years ago, issues and pull requests aren't being actively handled. But it still works for my purposes and for many others. So there shouldn't be much churn here.

Just in case I asked if upstream is interested in embedded type hints, but received no reply:

* https://github.com/kurtbrose/pyjks/issues/77


stubs/mock/mock/mock.pyi
```diff
@@ -7,9 +7,31 @@ from typing_extensions import Literal, ParamSpec, Self
 
 _F = TypeVar("_F", bound=Callable[..., Any])
 _AF = TypeVar("_AF", bound=Callable[..., Coroutine[Any, Any, Any]])
+class _Call(tuple[Any, ...]):
+    name: Any
+    parent: Any
+    from_kall: Any
```
intgr, line 13: TODO revert

stubs/pyjks/@tests/stubtest_allowlist.txt
```diff
@@ -0,0 +1,22 @@
...
+jks.SecretKeyEntry.key_size
+jks.__all__
+# Attributes implemented by __getattr__
+jks.jks.PrivateKeyEntry.algorithm_oid
+jks.jks.PrivateKeyEntry.pkey
+jks.jks.PrivateKeyEntry.pkey_pkcs8
+jks.jks.SecretKeyEntry.algorithm
+jks.jks.SecretKeyEntry.key
+jks.jks.SecretKeyEntry.key_size
+jks.PrivateKeyEntry.algorithm_oid
```
AlexWaygood, line 15: isn't this the same as the allowlist entry on line 1?

intgr: Ah yes, my bad, I got confused because my git push failed.

stubs/pyjks/jks/__init__.pyi
```diff
@@ -0,0 +1,52 @@
+from .bks import *
+from .jks import *
+from .jks import __version__, __version_info__
+from .util import *
+
+# pyjks exports lots of junk such as jks.jks.SIGNATURE_WHITENING, jks.util.b8 etc.
+# We don't mark those as re-exported as those don't seem like intended part of the public API.
```
intgr, line 7: Does this make sense? (in which case these need allowlist entries)

Or should I just blindly re-export everything that `pyjks` does and not worry about it?

stubs/pyjks/@tests/stubtest_allowlist.txt
AlexWaygood: oops, sorry 😶

stubs/pyjks/jks/__init__.pyi
AlexWaygood: It's not the approach we usually take, since this implies that somebody can do something like this, which will obviously fail at runtime:

```pycon
>>> import jks
>>> for attr in jks.__all__:
...     do_thing_with_attr(attr)
```

The approach we usually take is to explicitly re-export the things that are meant to be re-exported (`from jks import foo as foo, bar as bar`, etc.) and, if necessary, allowlist all the things that are just incidentally imported and aren't really meant to be publicly reexported.

intgr:
This should be ready for review and maybe merging.

Thanks a lot for your help so far @AlexWaygood 🥇

stubs/pyjks/@tests/stubtest_allowlist.txt
```diff
@@ -0,0 +1,58 @@
...
+jks.PrivateKeyEntry.pkey
+jks.PrivateKeyEntry.pkey_pkcs8
+jks.SecretKeyEntry.algorithm
+jks.SecretKeyEntry.key
+jks.SecretKeyEntry.key_size
+# Implicit imports, not re-exported
```
AlexWaygood, line 14: just to make it a little more readable:

```suggestion

# Implicit imports, not re-exported
```

stubs/pyjks/@tests/stubtest_allowlist.txt
```diff
@@ -0,0 +1,58 @@
...
+jks.jks.py23basestring
+jks.sun_crypto.DSA_OID
+jks.sun_crypto.DSA_WITH_SHA1_OID
+jks.sun_crypto.RSA_ENCRYPTION_OID
+jks.sun_crypto.b1
+jks.sun_crypto.b2
+jks.sun_crypto.b4
+jks.sun_crypto.b8
+jks.sun_crypto.py23basestring
+# Unintended reexports due to `from .utils import *` imports at runtime
```
AlexWaygood, line 57: ditto:

```suggestion

# Unintended reexports due to `from .utils import *` imports at runtime
```

stubs/pyjks/jks/jks.pyi
```diff
@@ -0,0 +1,107 @@
...
+from typing import Any, NoReturn
+from typing_extensions import Literal, Self, TypeAlias
+
+from .util import AbstractKeystore, AbstractKeystoreEntry
+
+__version_info__: tuple[int, int, int] | tuple[int, int, int, str]
+__version__: str
+MAGIC_NUMBER_JKS: bytes
+MAGIC_NUMBER_JCEKS: bytes
+SIGNATURE_WHITENING: bytes
```
AlexWaygood, line 12:
```suggestion
from typing_extensions import Final, Literal, Self, TypeAlias

from .util import AbstractKeystore, AbstractKeystoreEntry

__version_info__: tuple[int, int, int] | tuple[int, int, int, str]
__version__: str
MAGIC_NUMBER_JKS: Final[bytes]
MAGIC_NUMBER_JCEKS: Final[bytes]
SIGNATURE_WHITENING: Final[bytes]
```

stubs/pyjks/jks/jks.pyi
```diff
@@ -0,0 +1,107 @@
...
+    def __init__(
+        self,
+        *,
+        type: str = ...,
+        cert: bytes = ...,
+        store_type: _JksType = ...,
+        alias: str = ...,
+        timestamp: int = ...,
+        **kwargs: Any,
+    ) -> None: ...
```
AlexWaygood, line 31: This is pretty unsound -- we should either use `Literal` types in both the `__init__` signature and the annotations for the `__init__` parameters, or we should use `str` in both the `__init__` signature and the annotations for the `__init__` parameters. (The former is more precise but also more annoying for users of the stubs ;)

stubs/pyjks/jks/rfc7292.pyi
```diff
@@ -0,0 +1,28 @@
+from _typeshed import Incomplete
+from hashlib import _Hash
+from typing_extensions import Literal, TypeAlias
+
+from pyasn1.type import univ
+
+PBE_WITH_SHA1_AND_TRIPLE_DES_CBC_OID: tuple[int, ...]
```
AlexWaygood, line 7:
```suggestion
from typing_extensions import Final, Literal, TypeAlias

from pyasn1.type import univ

PBE_WITH_SHA1_AND_TRIPLE_DES_CBC_OID: Final[tuple[int, ...]]
```

stubs/pyjks/jks/sun_crypto.pyi
```diff
@@ -0,0 +1,6 @@
+SUN_JKS_ALGO_ID: tuple[int, ...]
+SUN_JCE_ALGO_ID: tuple[int, ...]
```
AlexWaygood, line 2:
```suggestion
from typing_extensions import Final

SUN_JKS_ALGO_ID: Final[tuple[int, ...]]
SUN_JCE_ALGO_ID: Final[tuple[int, ...]]
```

stubs/pyjks/jks/util.pyi
```diff
@@ -0,0 +1,66 @@
...
+from .jks import PrivateKeyEntry
+
+b8: Struct
+b4: Struct
+b2: Struct
+b1: Struct
+py23basestring: tuple[type[str], type[str]]
+RSA_ENCRYPTION_OID: tuple[int, ...]
+DSA_OID: tuple[int, ...]
+DSA_WITH_SHA1_OID: tuple[int, ...]
```
AlexWaygood, line 17:
```suggestion
from typing_extensions import Final, Literal, Self, TypeAlias

from .bks import BksKeyEntry
from .jks import PrivateKeyEntry

b8: Struct
b4: Struct
b2: Struct
b1: Struct
py23basestring: tuple[type[str], type[str]]
RSA_ENCRYPTION_OID: Final[tuple[int, ...]]
DSA_OID: Final[tuple[int, ...]]
DSA_WITH_SHA1_OID: Final[tuple[int, ...]]
```

AlexWaygood (commented):
Looks pretty good, a few small comments from a quick skim through:

stubs/pyjks/jks/jks.pyi
intgr: You mean the `type` argument? Yeah, that was an oversight, changing to `_CertType`.

The `__init__` constructors in this library are wonky and not meant for public usage. Not sure if we can signal this in some way from type hints.

AlexWaygood: > You mean the `type` argument?

Yup, that's what I meant :)



> The `__init__` constructors in this library are wonky and not meant for public usage. Not sure if we can signal this in some way from type hints.

Hmm, no, no way of signalling that (outside of putting a comment in the source code of the stubs for nobody to ever read 🙃)

stubs/pyjks/jks/jks.pyi
```diff
@@ -0,0 +1,107 @@
+from _typeshed import Incomplete
+from collections.abc import Iterable
+from typing import Any, NoReturn
+from typing_extensions import Final, Literal, Self, TypeAlias
+
+from .util import AbstractKeystore, AbstractKeystoreEntry
+
+__version_info__: tuple[int, int, int] | tuple[int, int, int, str]
+__version__: str
```
intgr, line 9: Would make sense to add `Final[]` to these too, right? And for the `Literal[]` contstants?

AlexWaygood: > Would make sense to add `Final[]` to these too, right?

Sure

> And for the Literal[] contstants?

I have a preference for `FOO: Final = 42` over `Foo: Final[Literal[42]]` (we allow both)

intgr: Ah cool, changed

intgr: FYI `FOO: Final = 42` isn't supported by the PyCharm type system. But to be fair, that's the least of its problems 😆

image

AlexWaygood: uhhh yeah don't use PyCharm's type checker :)

stubs/pyjks/jks/util.pyi
```diff
@@ -0,0 +1,66 @@
...
+class UnexpectedAlgorithmException(KeystoreException): ...
+class UnexpectedKeyEncodingException(KeystoreException): ...
+class UnsupportedKeystoreTypeException(KeystoreException): ...
+class UnsupportedKeystoreEntryTypeException(KeystoreException): ...
+class UnsupportedKeyFormatException(KeystoreException): ...
+
+class AbstractKeystore:
+    store_type: _KeystoreType
+    entries: Mapping[str, AbstractKeystoreEntry]
+    def __init__(self, store_type: _KeystoreType, entries: dict[str, AbstractKeystoreEntry]) -> None: ...
```
AlexWaygood, line 42: I think this might be back to front: it looks to me like you can provide any mapping to `__init__`, but it will always be converted to a `dict` before being set as an instance attribute: https://github.com/kurtbrose/pyjks/blob/0a046e54337a0c271ef5f7fdd9aa79c5ad350486/jks/util.py#L78-L80

```suggestion
    entries: dict[str, AbstractKeystoreEntry]
    def __init__(self, store_type: _KeystoreType, entries: Mapping[str, AbstractKeystoreEntry]) -> None: ...
```

You could probably be even more permissive and use `_typeshed.SupportsKeysAndGetItem`, since that's all that `builtins.dict` needs for its constructor:

https://github.com/python/typeshed/blob/16b98846017c07ebbca9c75c099cb4b6c6b11ad8/stdlib/builtins.pyi#L1063-L1071

```suggestion
    entries: dict[str, AbstractKeystoreEntry]
    def __init__(self, store_type: _KeystoreType, entries: SupportsKeysAndGetItem[str, AbstractKeystoreEntry]) -> None: ...
```

stubs/pyjks/jks/util.pyi
```diff
@@ -0,0 +1,66 @@
...
+    @classmethod
+    def load(cls, filename: str | bytes | Path, store_password: str | None, try_decrypt_keys: bool = True) -> Self: ...
+    def save(self, filename: str | bytes | Path, store_password: str) -> None: ...
```
AlexWaygood, line 45: I think you want `_typeshed.FileDescriptorOrPath` here, same as we use for the relevant overloads for `builtins.open`:

https://github.com/python/typeshed/blob/16b98846017c07ebbca9c75c099cb4b6c6b11ad8/stdlib/builtins.pyi#L1601-L1611

```suggestion
    def load(cls, filename: FileDescriptorOrPath, store_password: str | None, try_decrypt_keys: bool = True) -> Self: ...
    def save(self, filename: FileDescriptorOrPath, store_password: str) -> None: ...
```

stubs/pyjks/jks/util.pyi
```diff
@@ -0,0 +1,66 @@
...
+    def __init__(self, *, store_type: _KeystoreType = ..., alias: str = ..., timestamp: int = ..., **kwargs: Any) -> None: ...
+    @classmethod
+    def new(cls, alias: str) -> Self: ...
+    def is_decrypted(self) -> bool: ...
+    def decrypt(self, key_password: str) -> None: ...
+    def encrypt(self, key_password: str) -> None: ...
+
+def as_hex(ba: bytes | bytearray) -> str: ...
+def as_pem(der_bytes: bytes, type: _PemType) -> str: ...
+def bitstring_to_bytes(bitstr: Iterable[Literal[0, 1]]) -> bytes: ...
```
AlexWaygood, line 60: This annotation might be pretty annoying in practice. I'm not sure it's ever going to be possible to narrow the type of something to `Iterable[Literal[0, 1]]` without using a `cast()`. Could consider using `Iterable[int]` instead:

```suggestion
def bitstring_to_bytes(bitstr: Iterable[int]) -> bytes: ...
```

AlexWaygood (commented):
A few more comments; otherwise this looks good to go!

stubs/pyjks/jks/util.pyi
intgr: The reason for hinting `AbstractKeystore.entries: Mapping` is because a the subclass `KeyStore` overrides `entries` and mypy complained about mutability.

But I'm fine with using `type: ignore[override]` to suppress that.

AlexWaygood: Aha, got it. I'm that case, I'm fine with what using `Mapping` for the `entries` attribute (but please add an explanatory comment). I'd also be happy with changing it to `dict` and adding the `type: ignore`; will defer to you on which one you'd prefer!

intgr: I don't have strong feelings, but using `dict` seems more consistent.

AlexWaygood (approved):
LGTM, thanks!

intgr:
Cool! Should be ready from my perspective.

PR #10798, opened 2023-09-28
Fix typings for unittest.TestCase.assertLessEqual

ImogenBits:
Currently `unittest.TestCase.assertLessEqual` uses `SupportsDunderLT` and `SupportsDunderGT` to bound its arguments, when it should be using `SupportsDunderLE` and `SupportsDunderGE` since it's using `<=` in [its implementation](https://github.com/python/cpython/blob/f580edcc6a4c528020afe46c753db713474acad6/Lib/unittest/case.py#L1263), not `<`. This leads to false positives when making tests for classes that only implement the lax comparisons and not the strict ones.

This PR just fixes these two bounds to use the correct protocols.


srittau (approved):
Thanks!

PR #10800, opened 2023-09-29
Bump pyinstaller to 6.0.*

Avasam:
Closes #10752


stubs/pyinstaller/PyInstaller/building/api.pyi
```diff
@@ -22,9 +21,9 @@ if sys.platform == "darwin":
     _CodesignIdentityParam: TypeAlias = str | None
 else:
     _TargetArch: TypeAlias = None
-    _SuportedTargetArchParam: TypeAlias = object
+    _SuportedTargetArchParam: TypeAlias = Unused
```
JelleZijlstra, line 24: Supported?

stubs/pyinstaller/PyInstaller/utils/win32/winmanifest.pyi
```diff
@@ -1,127 +0,0 @@
-from _typeshed import Incomplete, ReadableBuffer, Unused
```
JelleZijlstra, line 1: Seems like this file still exists, why was it deleted? The contents do look very different from what's currently in the stub.

JelleZijlstra (approved):

stubs/pyinstaller/PyInstaller/utils/win32/winmanifest.pyi
Avasam: Valid question, this is not immediately obvious:

This module is not part of the public API and was only partially included for the `Manifest` type.
The `Manifest` class no longer exists (it is now a bytes-string)

See the `PyInstaller.depend.utils` entry in `stubtest_allowlist.txt` commented with `# Most modules are not meant to be used, yet are not marked as private`

PR #10801, opened 2023-09-29
Bump pyasn1 to 0.5.*

Avasam:
Closes #10068 
- Read through the code source changes
- Added missing `__all__` (which 0.5.0 updated a lot)
- Exposed `TAG_MAP`, `TYPE_MAP`, `tagMap` and `typeMap`
  - I justify those being part of the public API (or at least publicly used) through https://github.com/pyasn1/pyasn1/issues/9 and https://pyasn1.readthedocs.io/en/latest/changelog.html?highlight=Expose#revision-0-1-4
- Generated `streaming.pyi` using stubgen
- Fixed all stubtest issues
- Ran stubdefaulter to cover new defaults

0.5.0 also contains a ton of changes to return types. Luckily for me, they're still missing in the stubs, so I didn't have to deal with that.


srittau (approved):
Thanks!

I did not compare to the implementation, but the changes look reasonable (and most would be caught by stubtest).

PR #10802, opened 2023-09-29
Add `x_exports` to `xxlimited`

sobolevn:
Tests:

```python
        xxo = self.module.Xxo()
        self.assertEqual(xxo.x_exports, 0)
        b1 = memoryview(xxo)
        self.assertEqual(xxo.x_exports, 1)
        b2 = memoryview(xxo)
        self.assertEqual(xxo.x_exports, 2)
```


srittau (approved):
Thanks!

PR #10805, opened 2023-09-29
All locale functions should exist on macos

sobolevn:
I've seen the comment about this error being specific to GitHub Actions, but since it now not 3.10.5, but 3.10.13 we should try to remove these exceptions.

Local check:

```python
>>> import locale
>>> locale.gettext('a')
'a'
>>> locale.bindtextdomain
<built-in function bindtextdomain>
>>> locale.bind_textdomain_codeset
<built-in function bind_textdomain_codeset>
>>> locale.dcgettext
<built-in function dcgettext>
>>> locale.dgettext
<built-in function dgettext>
>>> locale.gettext
<built-in function gettext>
>>> locale.textdomain
<built-in function textdomain>
```

### DIFF
--- a/tests/stubtest_allowlists/darwin-py310.txt
+++ b/tests/stubtest_allowlists/darwin-py310.txt
@@ -1,13 +1,1 @@
 _?curses.color_pair
-
-# Github Actions on macOS with Python 3.10.5 claims these are missing, but they do exist locally
-(locale.bind_textdomain_codeset)?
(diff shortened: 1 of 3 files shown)


sobolevn:
3.11 and 3.12 are still affected. Maybe this is a `setup-python` bug? 🤔

AlexWaygood (approved):
Let's give this a go! 🤞

Issue #10095, opened 2023-04-28
Remove stubs for dj-database-url (not until October 2023)


AlexWaygood:
Marked as obsolete in https://github.com/python/typeshed/pull/10094

PR #10806, opened 2023-09-29
Remove `dj-database-url` stubs

sobolevn:
This is the last day of September and it is time to remove these stubs today or tomorrow.
Closes https://github.com/python/typeshed/issues/10095


hauntsaninja (approved):
Thanks!

PR #10807, opened 2023-09-30
`dbm.gnu` does exist on macos

sobolevn:
At least locally:

```
~/Desktop/cpython  main ✔                                                                 
» ~/.pyenv/versions/3.9.17/bin/python
Python 3.9.17 (main, Jul 12 2023, 22:57:20) 
[Clang 14.0.3 (clang-1403.0.22.14.1)] on darwin
>>> import dbm.gnu
>>> 
KeyboardInterrupt
>>> 
                                                                                           
~/Desktop/cpython  main ✔                                                                 
» python                             
Python 3.11.5 (main, Sep  5 2023, 10:34:31) [Clang 14.0.3 (clang-1403.0.22.14.1)] on darwin
>>> import dbm.gnu
>>> 
```

Let's try it in CI.

### DIFF
--- a/.github/workflows/stubtest_stdlib.yml
+++ b/.github/workflows/stubtest_stdlib.yml
@@ -31,7 +31,7 @@ jobs:
     strategy:
       matrix:
         # tkinter doesn't import on macOS 12
-        os: ["ubuntu-latest", "windows-latest", "macos-11"]
+        os: ["ubuntu-latest", "windows-latest", "macos-latest"]
         python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
       fail-fast: false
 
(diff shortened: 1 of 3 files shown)


AlexWaygood:
Seems like it still doesn't exist in CI, for whatever reason :)

sobolevn:
@AlexWaygood please, take a look at my `Temp` job output, it looks like a bug in CPython's build.

AlexWaygood:
You could try bumping to `macos-12` and see if that helps (IIRC there were loads of `tkinter` issues last time we tried that, but maybe things have changed)

AlexWaygood:
> @AlexWaygood please, take a look at my `Temp` job output, it looks like a bug in CPython's build.

Seems like that's probably an issue that needs to be raised with `actions/setup-python` rather than something we can fix here at typeshed :/

PR #10809, opened 2023-09-30
Fix `posixsubprocess` module for python3.12

sobolevn:


AlexWaygood (approved):
Stubtest can now validate the signature following https://github.com/python/cpython/commit/dfc5c41632c5c9b54edd43d11dd492a52ed5d372

PR #10810, opened 2023-09-30
Make pytype respect required python versions of stubs

hamdanal:
This is the pytype part of #10722

### DIFF
--- a/tests/pytype_test.py
+++ b/tests/pytype_test.py
@@ -19,6 +19,7 @@ import importlib.metadata
 import inspect
 import os
 import sys
+import textwrap
 import traceback
 from collections import defaultdict
 from collections.abc import Iterable, Sequence
@@ -50,7 +51,7 @@ def main() -> None:
     typeshed_location = args.typeshed_location or os.getcwd()
     old_typeshed_home = os.environ.get(TYPESHED_HOME)
     os.environ[TYPESHED_HOME] = typeshed_location
-    print("Testing files with pytype...")
+    print(f"Testing files with pytype on Python {PYTHON_VERSION}...")
     with chdir(typeshed_location):
         check_subdirs_discoverable(TYPESHED_SUBDIRS)
         files_to_test = determine_files_to_test(paths=args.files or TYPESHED_SUBDIRS)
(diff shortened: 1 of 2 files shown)


tests/pytype_test.py
```diff
@@ -133,17 +133,20 @@ def classify_files(paths: Sequence[str]) -> tuple[list[str], defaultdict[str, li
     """Classify files into stdlib and stubs by distribution."""
     stdlib: list[str] = []
     stubs: defaultdict[str, list[str]] = defaultdict(list)
-    stubs_location = Path("stubs").resolve()
+    stubs_path = Path("stubs")
+    stubs_absolute_path = Path(stubs_path).resolve()
```
JelleZijlstra, line 137:
```suggestion
    stubs_absolute_path = stubs_path.resolve()
```
It's already a Path

AlexWaygood:
I just posted some explanation in https://github.com/python/typeshed/issues/10722#issuecomment-1742150405 of the context/history around why we pytype and pyright are run so differently from mypy in CI.

The result of all that is that `pytype_test.py` now only works if you've already done something like what we do in CI to install all the non-types dependencies for all our stubs packages:

https://github.com/python/typeshed/blob/3eb9ff7f65acede795dc8be615e6f443d1f81578/.github/workflows/tests.yml#L71-L80

(This probably isn't an ideal state of affairs; we should probably alter `pytype_test.py` so that it includes this logic and installs all these packages into a temporary venv, rather than having to run two scripts sequentially in order for the second one to work.)

As a result, I think this PR as it currently stands is insufficient, since `tests/get_external_stub_requirements.py` is probably still going to include `matplotlib>=3.8` once #10721 is merged, and `matplotlib>=3.8` can't be installed with Python <3.9. (`tests/get_external_stub_requirements.py` already requires Python 3.9+ to even run, though; we require Python 3.9+ for all our tests except mypy_test.py, which needs to be compatible with Python 3.8.)

hamdanal:
Thanks for the context. I’ll close this now.

AlexWaygood:
No worries. Sorry I wasn't clearer earlier about the "philosophical difference" between this script and the mypy scripts :(

hamdanal:
Don’t worry about it. I was happy it worked locally that I forgot to look how it works in CI. I will look into it again in the future when I get some time and if no one else beats me to it.

PR #10811, opened 2023-09-30
Fix `curses.color_pair` signature

sobolevn:
In 3.7 it is pos-or-keyword: https://github.com/python/cpython/blob/0f56adb8d74e703974811620559d96c999334547/Modules/_cursesmodule.c#L2210-L2219

In 3.8+ (not from 3.8.0) it is pos-only: https://github.com/python/cpython/blob/edf701c4d83f399bff8612c34fef95da7663945e/Modules/_cursesmodule.c#L2611-L2632


tests/stubtest_allowlists/darwin-py310.txt
AlexWaygood, line 1: Why not delete this file entirely? `stubtest_stdlib.py` explicitly checks for its existence:

https://github.com/python/typeshed/blob/16b98846017c07ebbca9c75c099cb4b6c6b11ad8/tests/stubtest_stdlib.py#L38-L41

And we already don't have a `darwin-py38.txt` allowlist or a `darwin-py39.txt` allowlist

stdlib/_curses.pyi
```diff
@@ -278,9 +278,9 @@ if sys.platform != "win32":
     def color_content(__color_number: int) -> tuple[int, int, int]: ...
     # Changed in Python 3.8.8 and 3.9.2
     if sys.version_info >= (3, 8):
-        def color_pair(pair_number: int) -> int: ...
+        def color_pair(__pair_number: int) -> int: ...
     else:
-        def color_pair(__color_number: int) -> int: ...
+        def color_pair(pair_number: int) -> int: ...
```
AlexWaygood, line 283: I don't have a linux or mac to test on (and Windows doesn't have `curses`), but it looks to me like this was positional-only on 3.7 as well: https://tio.run/##K6gsycjPM/7/PzO3IL@oRKG4spiroCgzr0RDCcjUK0stKs7Mz1PILFbSUUAS0OSCqk8uLSpOLeaCUHrJ@Tn5RfEFiZlFGiAiPq80Nym1yNbESPP/fwA

AlexWaygood (approved):

PR #10812, opened 2023-09-30
Remove stubs for `urllib3`

### LINKED ISSUE #6898: Remove urllib3 types (not before October 2023)
#6858 added stubs for urllib3. It might have been a mistake, because urllib3 seems to have its own type annotations and even a `py.typed` marker, making the stubs unnecessary.

### LINKED ISSUE #10142: [requests] Change urllib3 dependency
As of requests 2.30, the urllib3 dependency is pinned to `>=1.21.1,<3`, so it works with urllib3 v2, which ships its own type stubs. We should change the dependency from types-urllib3 to `urllib3>=2,<3`.

AlexWaygood:
Fixes #6898.
Fixes #10142.

Note that this has a significant impact on the number of stubs packages with non-types dependencies. Both `types-requests` and `types-influxdb-client` now depend on `urllib3>=2` instead of `types-urllib3`. That in turn means that `types-caldav`, `types-slumber` and `types-requests-oauthlib` all depend indirectly on `urllib3`, since all three stubs packages depend on `types-requests`. All of that might mean that `mypy_test.py` and `regr_test.py` slow down a bit. If so, I have some ideas for mitigations I could work on in future PRs for speeding them up a bit.


AlexWaygood:
N.B. I checked stubtest for `types-caldav`, `types-slumber` and `types-requests-oauthlib` all locally; everything was green

AlexWaygood:
> All of that might mean that `mypy_test.py` and `regr_test.py` slow down a bit.

Looks like `regr_test.py` in particular gets a bit slower, but nothing too catastrophic :)

JelleZijlstra (approved):

PR #10813, opened 2023-09-30
fix `__match_args__` attribute name

bzoracler:
```pycon
>>> import ast
>>> m = ast.parse("type ListOrSet[T] = list[T] | set[T]")
>>> print(m.body[0].__match_args__)
('name', 'type_params', 'value')
```


JelleZijlstra:
Thanks, good catch. I wonder if something in CI should validate that the `__slots__` match existing attribute names.

PR #10815, opened 2023-10-01
[pyjks] Fill in jks.bks and other modules, improved constructors

intgr:
Additionally:
* Fixed type of `PrivateKeyEntry.__init__(encrypted=)` argument
* Improved type of `SecretKeyEntry.__init__(sealed_obj=)` argument
* Removed defaults from constructor args that shouldn't be None (alias, timestamp, etc).
* Added `| None` to constructor arguments that may be None
* Replaced `= ...` with `= None` where appropriate
* Added `@overload` constructors for encrypted and unencrypted usages
* Added helpful comments to `__init__` constructors


intgr:
Ah thanks again, I forgot the `*` in args list.

AlexWaygood:
We should improve the stubtest error message there — that kind of thing comes up a lot, and stubtest could be clearer about what's gone wrong in this case!

AlexWaygood:
> We should improve the stubtest error message there — that kind of thing comes up a lot, and stubtest could be clearer about what's gone wrong in this case!

Filed https://github.com/python/mypy/pull/16210 to improve this!

stubs/pyjks/jks/bks.pyi
```diff
@@ -1,69 +1,125 @@
...
+    cert_chain: list[tuple[_CertType, bytes]]
+    def __init__(
+        self,
+        *,
+        cert_chain: list[tuple[_CertType, bytes]] = ...,
+        encrypted: bytes | None = ...,
+        store_type: _BksType = ...,
+        alias: str = ...,
+        timestamp: int = ...,
+        **kwargs: Any,
```
srittau, line 34: I think `_typeshed.Unused` is a better fit here (and in similar places).

stubs/pyjks/jks/bks.pyi
```diff
@@ -1,69 +1,125 @@
...
+    store_type: _BksType | None
+    cert_chain: list[tuple[_CertType, bytes]]
+    def __init__(
+        self,
+        *,
+        cert_chain: list[tuple[_CertType, bytes]] = ...,
+        encrypted: bytes | None = ...,
+        store_type: _BksType = ...,
+        alias: str = ...,
+        timestamp: int = ...,
```
srittau, line 33: As `kwargs.get("...")` is used:

```suggestion
        encrypted: bytes | None = None,
        store_type: _BksType = None,
        alias: str | None = None,
        timestamp: int | None = None,
```

The annotations for `AbstractKeystoreEntry` seem to be wrong, too.

Same for sub-classes.

stubs/pyjks/jks/jks.pyi
```diff
@@ -47,7 +47,7 @@ class PrivateKeyEntry(AbstractKeystoreEntry):
         self,
         *,
         cert_chain: list[tuple[_CertType, bytes]] = ...,
-        encrypted: bool = ...,
+        encrypted: bytes | None = ...,
```
srittau, line 50:
```suggestion
        encrypted: bytes | None = None,
```

(Same for some of the other defaults. Their types are also missing `None`, see above. But changing this in this PR is optional.)

srittau (changes requested):
Thanks! Notes below.

stubs/pyjks/jks/bks.pyi
intgr: This is just sloppy API design in `pyjks`. The `__init__` constructor isn't really meant for user code (it may be useful for mocking in tests, but that's about it). Some constructable classes actually have a `new()` classmethod for this purpose, with a more reasonable interface.

Attributes like `alias`, `timestamp`, `store_type` are never `None` in real use cases, when reading or manipulating keystores.

_I could_ make them None'able in `__init__`, but changing attributes to be None'able would IMO make the stubs annoying to use and reduce its usefulness.

intgr: So if passing `None` to these args causes unsound results, maybe instead I should remove defaults of those args? Even though the upstream code allows omitting them.

srittau: Yes, I think removing the defaults sounds like the best idea. This will need stubtest allow entries, though.

intgr:
Pushed lots of changes:
* Improved type of `SecretKeyEntry.__init__(sealed_obj=)` argument
* Removed defaults from constructor args that shouldn't be None (alias, timestamp, etc).
* Added `| None` to constructor arguments that may be None
* Added `@overload` constructors for encrypted and unencrypted usages
* Replaced `= ...` with `= None` where appropriate

I could split these back out and open separate PR if you'd prefer.

stubs/pyjks/jks/bks.pyi
intgr: OK, did that. Turns out stubtest allowlists didn't need to change.

srittau (approved):
Thanks!

PR #10816, opened 2023-10-01
Make collections.abcs more consistent with runtime implementation

Gobot1234:
Interested to see how much breakage primer reports.

refs: python/typing#1480


Gobot1234:
Looking over these issues, all the aiohttp ones look like a mypy bug where it thinks that any annotated attribute is abstract the others hopefully might be easy enough fixes if you rename typing.Generator strings to types.GeneratorType, I'll see if a mypy PR can get rid of most of them. If that isn't possible would it be possible for mypy to selectively remove this commit?

stdlib/types.pyi
```diff
@@ -358,6 +358,12 @@ _ReturnT_co = TypeVar("_ReturnT_co", covariant=True)
 
 @final
 class GeneratorType(Generator[_YieldT_co, _SendT_contra, _ReturnT_co]):
```
srittau, line 360: I'd be interested what happens if we just change types.pyi for now, but remove the `typing` base classes to better match runtime.

stdlib/typing.pyi
```diff
@@ -419,32 +419,16 @@ class Generator(Iterator[_YieldT_co], Generic[_YieldT_co, _SendT_contra, _Return
     @abstractmethod
     def throw(self, __typ: BaseException, __val: None = None, __tb: TracebackType | None = None) -> _YieldT_co: ...
     def close(self) -> None: ...
-    def __iter__(self) -> Generator[_YieldT_co, _SendT_contra, _ReturnT_co]: ...
```
Gobot1234, line 422: I'm unsure as to whether or not this should be -> Self the runtime has it as that in Iterator

stdlib/types.pyi
Gobot1234: I'm not entirely sure what you're proposing, wouldn't that just crash mypy because typing.Generator wouldn't exist?

srittau: I meant just using `class GeneratorType:` here, not deleting `typing.Generator`.

Gobot1234: typing.Generator hasn't been deleted it's just a protocol now

stdlib/typing.pyi
Gobot1234: Upon further consideration, not implementing it as returning Self is probably a bug so I'll change it in (Async)Iterator even

AlexWaygood: `typing.Generator.__iter__` shouldn't be annotated as returning `Self` for the same reason that `typing.Iterator.__iter__` shouldn't be annotated as returning `Self`. 99% of all iterators have `__iter__` methods that return `Self`, and 99% of generators are the same. But if we hinted them as such in the stub, we'd be changing the very _definition_ of Iterator/Generator such that the remaining 1% of iterators and generators would no longer be considered iterators or generators. And that wouldn't be accurate: it's not part of the contract that iterators have to return `self` from their `__iter__` methods at runtime in order for them to be considered iterators, they just have to return an iterable object.

With `types.GeneratorType`, it's a different issue, because that class doesn't define an interface in the same way. We can just reflect what happens at runtime there.

JelleZijlstra:
This change feels correct, let's see if we can get type checkers to agree.

stdlib/typing.pyi
JelleZijlstra: I suspect removing this method is causing the mypy-primer fallout. Let's add it back.

stdlib/typing.pyi
```diff
@@ -463,15 +463,6 @@ class Generator(Iterator[_YieldT_co], Generic[_YieldT_co, _SendT_contra, _Return
     @abstractmethod
     def throw(self, typ: BaseException, val: None = None, tb: TracebackType | None = None, /) -> _YieldT_co: ...
     def close(self) -> None: ...
```
JelleZijlstra, line 465:
```suggestion
    def close(self) -> None: ...
    def __iter__(self) -> Generator[_YieldT_co, _SendT_contra, _ReturnT_co]: ...
```

JelleZijlstra:
cc @A5rocks regarding the primer changes in trio. I think they are technically correct; trio uses the Coroutine ABC, which does not guarantee the existence of attributes like `cr_frame`. However, I'm not sure this makes the user experience better in practice.

A5rocks:
> cc @A5rocks regarding the primer changes in trio. I think they are technically correct; trio uses the Coroutine ABC, which does not guarantee the existence of attributes like `cr_frame`. However, I'm not sure this makes the user experience better in practice.

I agree with these changes being technically correct, because `asyncio.iscoroutine` exists. (I'm not too sure about if it didn't exist, but oh well. I see something about a `types.coroutine` decorator but I can't get a `collections.abc.Coroutine` isinstance check to pass.) Here's another type that's incorrect, that I just spotted:

```
>>> import asyncio
>>> @asyncio.coroutine
... def blah():
...   yield "hello"
...
>>> asyncio.iscoroutine(blah())
True
>>> blah().cr_origin  # this is defined on CoroutineType
AttributeError: 'generator' object has no attribute 'cr_origin'
```

Anyways, I think it's better to be more correct. We have added asserts to check `cr_frame` in the past (because it's Optional) and this is just more along that line. Though take what I say with a grain of salt because I don't really know the internal details.

Issue #8217, opened 2022-06-30
`__eq__`'s type hint for primitives accept `object` while in many cases it returns `NotImplemented`


mmohaveri:
As can be seen in [builtins.pyi](https://github.com/python/typeshed/blob/master/stdlib/builtins.pyi), all primitives are defining their `__eq__` and `__ne__` functions in the following manner:
```python
    def __eq__(self, __x: object) -> bool: ...
    def __ne__(self, __x: object) -> bool: ...
```

This will cause incorrect type errors in cases where the second operand has overloaded `__eq__` or `__ne__` in a way that it does not return `bool`. Consider the following example:

```python
T = TypeVar("T", float, bool)

class Matrix(Generic[T]):
  def __init__(self, data: List[List[T]]):
    self.data = data

  def __eq__(self, other: "Matrix[T] | float | bool") -> Matrix[bool]:
    if not isinstance(other, Matrix):
      return Matrix([[elem == other for elem in row] for row in self.data])
    return Matrix((elem == other_elem for elem, other_elem in zip(row1, row2)] for row1, row2 in zip(self.data, other.data)]

  def __ne__(self, other: "Matrix[T] | float | bool") -> Matrix[bool]: ...
```

Here we're trying to create a class that mimics numpy's `NDArray` in a very limited way.

Now if I try to check the equality of a `Matrix` with a `float`, the return type will be inferred as `Matrix`, since LHS of the operation has a matching `__eq__` method which returns matrix (and the returned type in runtime is `Matrix` as well)

But if I try to check the equality of a `float` with a `Matrix`, the returned type will be inferred as `bool` because LHS of the operation has a compatible `__eq__` (`def __eq__(self, __x: object) -> bool`, object and `Matrix` ar compatible) while in the runtime because float's __eq__ will return a `NotImplemented`, interpreter will try to run the `__ne__` function of the RHS which returns a `Matrix`.

This will create situations where the inferred type of a simple equality check operation does not match the runtime's type, all because 
primitives stubs has been defined in a way that is not representative of their runtime behavior.

srittau:
This is indeed problematic and the natural solution would be an overloads returning `NotImplemented`. In the case of `float` (where `FloatLike` would have to replaced with something appropriate):

```python
@overload
def __eq__(self, other: FloatLike) -> bool: ...
@overload
def __eq__(self, other: object) -> NotImplemented: ...
```

[...]

PR #10819, opened 2023-10-01
Remove redundant `NoReturn` union in openpyxl

Avasam:
From https://github.com/python/typeshed/issues/8217#issuecomment-1742239764

Unless you think there's documentation/semantic value in keeping the union. kindof like how sometimes we have redundant unions with `None` or `Any`. In which case I'll update the comment to make that obvious. Although pyright/Pylance seem to completely strip it out.
image

Not going with `def tagname(self) -> NoReturn: ...` for the reason listed in the comment.


srittau (approved):
`NoReturn` doesn't make much sense to me in an union. Many methods can raise an exception without us annotating that.

PR #10820, opened 2023-10-02
Remove `DefaultSelector.fileno` from allowlists

sobolevn:


stdlib/selectors.pyi
```diff
@@ -66,8 +66,12 @@ class KqueueSelector(BaseSelector):
     def select(self, timeout: float | None = None) -> list[tuple[SelectorKey, _EventMask]]: ...
     def get_map(self) -> Mapping[FileDescriptorLike, SelectorKey]: ...
 
+# Not a real class in runtime, it is just a conditional alias to other real selectors.
+# But, since it depends on the platform, we cannot express this in stubs.
```
AlexWaygood, line 70: Nice, I think this is the right call -- just a small suggestion about the wording here:

```suggestion
# Not a real class at runtime, it is just a conditional alias to other real selectors.
# The runtime logic is more fine-grained than a `sys.platform` check;
# not really expressible in the stubs
```

AlexWaygood (approved):
Thanks!

PR #10822, opened 2023-10-02
Fix Windows-specific allowlist entries in `_msi`, `msilib`, `selectors` and `subprocess`

AlexWaygood:


srittau (approved):
Looks reasonable.

PR #10823, opened 2023-10-02
`select.POLLRDHUP` does not exist on macos

sobolevn:
```
» ./python.exe
Python 3.13.0a0 (heads/main:98c0c1de18e, Sep 28 2023, 09:21:43) [Clang 14.0.3 (clang-1403.0.22.14.1)] on darwin
>>> import select
>>> select.POLLRDHUP
AttributeError: module 'select' has no attribute 'POLLRDHUP'. Did you mean: 'POLLHUP'?
```

And stable Python:

```
Python 3.11.5 (main, Aug 24 2023, 15:09:45) [Clang 14.0.3 (clang-1403.0.22.14.1)] on darwin
>>> import select
>>> select.POLLRDHUP
AttributeError: module 'select' has no attribute 'POLLRDHUP'. Did you mean: 'POLLHUP'?
```

See https://manpages.courier-mta.org/htmlman2/poll.2.html

### DIFF
--- a/stdlib/select.pyi
+++ b/stdlib/select.pyi
@@ -15,7 +15,8 @@ if sys.platform != "win32":
     POLLOUT: int
     POLLPRI: int
     POLLRDBAND: int
-    POLLRDHUP: int
+    if sys.platform != "darwin":
+        POLLRDHUP: int
     POLLRDNORM: int
     POLLWRBAND: int
     POLLWRNORM: int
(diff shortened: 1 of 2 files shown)


stdlib/select.pyi
```diff
@@ -15,7 +15,8 @@ if sys.platform != "win32":
     POLLOUT: int
     POLLPRI: int
     POLLRDBAND: int
-    POLLRDHUP: int
+    if sys.platform != "darwin":
```
sobolevn, line 18: Or maybe it should `== "linux"`?

sobolevn: A bit sad that we don't have some other unix, for example, bsd based.

AlexWaygood (approved):

PR #10824, opened 2023-10-02
Fix allowlist entries in `asyncio.windows_events`

AlexWaygood:
And fix the return value of `IocpProactor.recvfrom_into()`, which I think I screwed up slightly in #10817.

I _think_ these are basically correct, but asyncio isn't really my area of expertise...


stdlib/asyncio/windows_events.pyi
```diff
@@ -48,6 +48,12 @@ if sys.platform == "win32":
         def select(self, timeout: int | None = None) -> list[futures.Future[Any]]: ...
         def recv(self, conn: socket.socket, nbytes: int, flags: int = 0) -> futures.Future[bytes]: ...
         def recv_into(self, conn: socket.socket, buf: WriteableBuffer, flags: int = 0) -> futures.Future[Any]: ...
+        def recvfrom(
+            self, conn: socket.socket, nbytes: int, flags: int = 0
+        ) -> futures.Future[tuple[int, socket._RetAddress]]: ...
```
JelleZijlstra, line 53:
```suggestion
        ) -> futures.Future[tuple[bytes, socket._RetAddress]]: ...
```
There is an error case returning bytes in the Python code, and if I'm reading the C code in `_overlapped` correctly, this hits the `TYPE_READ_FROM` case in `_overlapped_Overlapped_getresult_impl`, which returns a bytes in the first tuple member.

JelleZijlstra (changes requested):

stdlib/asyncio/windows_events.pyi
AlexWaygood: Good catch.

AlexWaygood: That matches the other `recvfrom` signatures in typeshed as well. I think I misread some things somewhere.

PR #10831, opened 2023-10-03
Fix allowlists on linux and macos for `posix` module

sobolevn:
Mac demo:

```
>>> import posix
>>> posix.NGROUPS_MAX
16
>>> posix.EX_NOTFOUND
AttributeError: module 'posix' has no attribute 'EX_NOTFOUND'
>>> 
```


sobolevn:
`EX_NOTFOUND` exists on some Unix systems, but not all of them. For example, our CI does not have it: https://bugs.python.org/issue9933

AlexWaygood (approved):
Thanks!

Issue #10834, opened 2023-10-04
Incorrect types filters in `bleach.sanitizer.Cleaner`


fsan:
Hello

I have read the README.md for typeshed and it suggests to open a github issue for discussing potential incorrect stubs. I hope this is the right place to post this question.

I am having problems with `types-bleach` and I believe it might be because the stubs are incorrect.
I simplified the minimal example of `bleach.sanitizer.Cleaner` to open this ticket and I was able to reproduce it using `types-bleach="*"` and `bleach~=6.0` in my dependencies.

Here is the code that mypy is rejecting:
```python3
from bleach.html5lib_shim import Filter
from bleach.sanitizer import Cleaner, BleachSanitizerFilter

class MyFilter(Filter):
        ...

if __name__ == "__main__":
    cleaner: Cleaner = Cleaner(filters=[MyFilter])
```

When running: `mypy main.py`:
```
main.py:10: error: List item 0 has incompatible type "type[MyFilter]"; expected "_Filter"  [list-item]
main.py:10: note: Following member(s) of "MyFilter" have conflicts:
main.py:10: note:     Expected:
main.py:10: note:         def __call__(self, *, source: BleachSanitizerFilter) -> Any
main.py:10: note:     Got:
main.py:10: note:         def MyFilter(self) -> MyFilter
Found 1 error in 1 file (checked 1 source file)
```

I do not believe this is a liskov substitution problem because, as you can see in the documentation for Mozilla bleach [here](https://github.com/mozilla/bleach/blob/main/docs/clean.rst#html5lib-filters-filters), `Cleaner.__init__` expects a `Type[Filter]` as list element and not a `Filter`.

If I understood correctly the the problem, `bleach-stubs.sanitizer` should have a different typing.

Current types:
```python3
class Cleaner:
    # ...
    filters: Iterable[_Filter]
     #...
    def __init__(
        self,
        # ...
        filters: Iterable[_Filter] | None = None,
        # ...
    ) -> None: ...
    # ...
```

My suggested typing

```python3
class Cleaner:
    # ...
    filters: Iterable[Type[Filter]]
     #...
    def __init__(
        self,
        # ...
        filters: Iterable[Type[Filter]] | None = None,
        # ...
    ) -> None: ...
    # ...
```

I would like to know if the analysis above makes sense or if I am missing something.
Thank you 🙇

srittau:
[...]

PR #10835, opened 2023-10-04
[bleach] Add html5lib_shim.Filter__init__/__iter__

srittau:
The missing __init__ method caused `Filter` to be incompatible with the `_Filter` protocol, preventing valid initialization of `Cleaner` instances.

Closes: #10834


AlexWaygood (approved):
Nice!

Issue #10837, opened 2023-10-05
Communicating our versioning policy better


srittau:
#10825 made it clear that we need to communicate our versioning policy better. The first step is to add a note about it to package descriptions as suggested by me in typeshed-internal/stub_uploader#104. But I think we could also communicate the "specialness" of the typeshed part of the version number better. A few ideas:

* Don't reset the version number when the upstream version changes.
* Start with a high number, e.g. 1000 to make it stand out, e.g. 1.2.3.1023
* Use the current date (and limit to one upload per day): 1.2.3.231005
* Bump the version significantly on major changes, e.g. 1.2.3.1023 -> 1.2.3.2000
* *Edit:* Use epochs, e.g. 45!1.2.3

Personally, I think options 1 and 2 combined would be a good solution, although I find option 3 strangely appealing as well.

srittau:
I'd like to bump this again. I still think we should at least bump the minimum stub version up to 1000 to signify that it's "special", although I really like the idea to use the current date. A version like 1.2.3.231012 looks weird, but it really stands out. Any opinions?

Akuli:
I like the date versioning (aka [calver](https://calver.org/)) idea, and I think it would be a good fit for typeshed.

Maybe we should use full years in the version number, as in `1.2.3.20231012`? The meaning is more obvious, and I don't really see any downsides in this.

AlexWaygood:
Thanks for bumping @srittau! Agree that our current policy looks too much like we're just making a "minor version bump", which isn't really how we do things here.

I basically agree with @Akuli -- I also quite like the date versioning idea!

srittau:
We also don't need to limit the uploads. Just use the best free date if the current date is already in use. I don't think it's a problem if a package is a day in the future from time to time.

Akuli:
IMO it would be better to limit the uploads. Suppose a contributor is doing large manual changes into a specific package, and makes 5 pull requests with 3 hours between each PR (total time spent would be 12 hours, so not unreasonable). If we merge the PRs quickly, this would get stub-uploader 4 days ahead in one day.

AlexWaygood:
I agree with @Akuli. I also think limiting uploads of any one stubs package to once per day is fine. That's already a much more frequent release schedule than you'll get with basically any other open-source package.

srittau:
typeshed-internal/stub_uploader#114

Issue #10825, opened 2023-10-02
`types-requests>=2.31.0.7` can't be installed alongside `urllib3<2`


jessemyers-lettuce:
This [PR](https://github.com/python/typeshed/pull/10812) creates a breaking change because it adds:

```
urllib3>=2
```

As of this writing, `requests` itself has this dependency declaration:

```
urllib3>=1.21.1,<3
```

This means that there are scenarios where `requests` can be installed, but `types-requests` cannot.

In my case, I am using `boto3` and `botocore` for AWS interactions and it defines its own `urllib3` dependency range that produces this incompatiblity:

```
urllib3>=1.25.4,<1.27
```

Please consider aligning the `types-requests` version ranges to the same bounds that `requests` itself uses.

AlexWaygood:
If you have a requirements file that also includes a package that pins `urllib3>=1.25.4,<1.27`, pip (or any other standards-based installer) should be able to determine that the maximum version of `types-requests` that can be installed in `types-requests==2.31.0.6`, since the latest version depends on `urllib3>=2`. If that doesn't work for some reason, would pinning `types-requests` to `<2.31.0.7` be an option for you?

AlexWaygood:
`urllib3<2` will unfortunately not work for our `types-requests` stubs, as they have to have a version of `urllib3` with typing information.

jessemyers-lettuce:
@AlexWaygood My goal is always to use the most recent version of every dependency that I can without breaking my systems. I use an automated process (`renovate`) to continuously integrate the latest.

So yes, I am pinning a specific version.

But... it seems to me that you've taken a step to optimize your development process -- I don't know what these stubs do so I can't comment on the value added -- and this comes at the expense of the users of your library.

Put another way: it seems obvious to me that the latest version of `types-requests` should be 100% compatible with the latest version of `requests` **OR** `types-requests` should have an upper bound version for `requests` itself (which is probably not something you want).

AlexWaygood:
> But... it seems to me that you've taken a step to optimize your development process -- I don't know what these stubs do so I can't comment on the value added -- and this comes at the expense of the users of your library.

To be clear, without this change, it was painful for our users to use `types-requests` alongside the latest version of `urllib3` (see #10786). This change resolved that problem. It's hard for me to see a way that we could solve that problem and your problem simultaneously here.

jessemyers-lettuce:
It seems like there are two obvious alternatives:

[...]

PR #10840, opened 2023-10-05
Add a section "Stub package versioning"

srittau:


srittau:
Any suggestions welcome!

This will of course need an update if the discussion in #10837 concludes that we update our versioning scheme.

AlexWaygood:
Could we also explicitly recommend against using mypy's `--install-types` option (and _especially_ not `--install-types --non-interactive`)? The behaviour of `--install-types` https://github.com/python/typeshed/issues/10825#issuecomment-1747127732 is... quite bad.

README.md
```diff
@@ -46,6 +46,50 @@ and are automatically released (multiple times a day, when needed) by
 Type checkers should be able to use these stub packages when installed. For more
 details, see the documentation for your type checker.
 
+### Stub package versioning
```
AlexWaygood, line 49:
```suggestion
### Package versioning for third-party stubs
```

README.md
```diff
@@ -46,6 +46,50 @@ and are automatically released (multiple times a day, when needed) by
...
+
+Version numbers of third-party stub packages consist of at least four elements,
+where the last element represents the stub version. The other elements
+are the version of the package being stubbed.
```
AlexWaygood, line 53:
```suggestion
Version numbers of third-party stub packages consist of at least four parts.
All parts of the stub version, except for the last part, correspond to the
version of the runtime package being stubbed. For example, if the `types-foo`
package has version `1.1.2.0`, this signifies that the package contains stubs
for the `foo==1.1.2`. The final element of the version indicates that this is
the first revision of the stubs for `foo`: if an update to the stubs was
pushed the following day (but the stubs were still aiming to provide
annotations for `foo==1.1.2`), then the version of `types-foo` would
increment to `1.1.2.1`.
```

README.md
```diff
@@ -46,6 +46,50 @@ and are automatically released (multiple times a day, when needed) by
...
+
+While we try to minimize the number of breaking changes, due to the nature
+of stubs, every version can introduce changes that make your code fail
+to type check.
```
AlexWaygood, line 57:
```suggestion
At typeshed, we try to keep breaking changes to a minimum. However, due to the
nature of stubs, any version bump can introduce changes that might make your
code fail to type check.
```

README.md
```diff
@@ -46,6 +46,50 @@ and are automatically released (multiple times a day, when needed) by
...
+
+There are several strategies to pin the version of the stubs you are using,
+all with their own tradeoffs:
```
AlexWaygood, line 60:
```suggestion
There are several strategies available for specifying the version of a stubs
package you're using, each with its own tradeoffs
```

README.md
```diff
@@ -46,6 +46,50 @@ and are automatically released (multiple times a day, when needed) by
...
+
+* Use the same pin that you use for the package being stubbed. For example,
+  if you use `requests>=2.30.0,<2.32`, you can use
+  `types-requests>=2.30.0,<2.32`. This ensures that the stubs are compatible
+  with the package you are using, but it carries a small risk of breaking
+  type checking due to changes in the stubs. Also, stubs often lag behind
```
AlexWaygood, line 66:
```suggestion
  type checking due to changes in the stubs.
  
  Another risk of this strategy is that stubs often lag behind
```

srittau:
Isn't `--install-types` something that mypy should document better? Maybe even issue a warning when it's used? Otherwise, any wording suggestions welcome!

README.md
```diff
@@ -46,6 +46,50 @@ and are automatically released (multiple times a day, when needed) by
...
+  to a certain minimum version, because it fixes a critical bug, but you
+  could be unable to update the stubs, since an update has not been released.
+* When the package being stubbed is using [Semantic Versioning](https://semver.org/lang/de/),
+  you could use the same major version for the stubs. For example, if you use
+  `requests>=2.30.0,<2.33`, you can use `types-requests>=2.29.3.4,<3`. This
+  means that the stubs are not as tighly coupled to the package being stubbed,
+  and you will gain potential improvements in the stubs, but it carries a
+  risk, since the stubs might provide a different API surface than the
+  implementation. It also works around the problem that stubs often lag behind
+  the package being stubbed.
```
AlexWaygood, line 77: I'm not sure I fully understand this strategy :/

AlexWaygood:
> Isn't `--install-types` something that mypy should document better? Maybe even issue a warning when it's used? Otherwise, any wording suggestions welcome!

Yes. It's unfortunately very widely used, however, so I think there'd be an outcry if mypy even tried to deprecate it :/

README.md
```diff
@@ -46,6 +46,50 @@ and are automatically released (multiple times a day, when needed) by
...
+* Pin the stubs to a known good version and update manually. For example, if
+  you use `types-requests==2.31.0.1`, you can be sure that upgrading
+  dependencies will not break type checking. However, you will miss out on
+  improvements in the stubs that could potentially improve type checking.
```
AlexWaygood, line 81: Well, you'll only "miss out" for as long as you fail to upgrade. This depends on how often you bump your dependencies, right?

README.md
```diff
@@ -46,6 +46,50 @@ and are automatically released (multiple times a day, when needed) by
...
+  It also has the risk that the stubs you are using are not compatible with
+  the package being stubbed.
+* Don't pin the stubs. This is the most flexible option, but it carries the
+  risk that the stubs become incompatible with the package being stubbed.
+  For example, when a new major version of the package is released and the
+  stubs are updated before you update the package being stubbed.
```
AlexWaygood, line 87:
```suggestion
  For example, if a new major version of the package is released, there's a
  chance the stubs might be updated to reflect the new version of the runtime
  package before you update the package being stubbed.
```

README.md
```diff
@@ -46,6 +46,50 @@ and are automatically released (multiple times a day, when needed) by
...
+
+You can also use the different strategies as needed. For example, you could
+default match the pinning, but pin the stubs to a known good version when
+a problem arises that can't easily be fixed.
```
AlexWaygood, line 91: If we used a numbered list rather than a bulleted list for the strategies, we could say something like "you could default to strategy (1), but fall back to strategy (2) when a problem arises that can't easily be fixed". That might be slightly more elegant.

AlexWaygood (commented):
Overall, this looks great! A few wording suggestions:

srittau (commented):
Reword version explanation, mostly using @AlexWaygood's wording

README.md
srittau: I used a slightly modified version of your wording.

srittau (commented):
(VSCode #$%$&*$E@#)

README.md
srittau: I removed it, I didn't really understand it myself.

srittau (commented):
(VSCode #$%$&*$E@#)

README.md
```diff
@@ -46,6 +46,52 @@ and are automatically released (multiple times a day, when needed) by
 Type checkers should be able to use these stub packages when installed. For more
 details, see the documentation for your type checker.
 
+### Package versioning for third-party stubs
+
+Version numbers of third-party stub packages consist of at least four parts.
+All parts of the stub version, except for the last part, correspond to the
+version of the runtime package being stubbed. For example, if the `types-foo`
+package has version `1.1.2.0`, this signifies that the package contains stubs
+for the `foo==1.1.2`. In this example, the final element of the version
```
AlexWaygood, line 55:
```suggestion
for `foo==1.1.2`. In this example, the final element of the version
```

README.md
```diff
@@ -46,6 +46,52 @@ and are automatically released (multiple times a day, when needed) by
...
+indicates that this is the first revision of the stubs for `foo`. If an update
+to the stubs was pushed (but the stubs were still aiming to provide
```
AlexWaygood, line 57: The English subjunctive ;)

```suggestion
to the stubs were pushed (but the stubs were still aiming to provide
```

README.md
```diff
@@ -46,6 +46,52 @@ and are automatically released (multiple times a day, when needed) by
...
+
+  Another risk of this strategy is that stubs often lag behind
+  the package being stubbed. You might want to force the package being stubbed
+  to a certain minimum version, because it fixes a critical bug, but you
+  could be unable to update the stubs, since an update has not been released.
+2. Pin the stubs to a known good version and update manually. For example, if
+  you use `types-requests==2.31.0.1`, you can be sure that upgrading
+  dependencies will not break type checking. However, you will miss out on
+  improvements in the stubs that could potentially improve type checking,
+  unless you manually update the stubs.
```
AlexWaygood, line 82:
```suggestion
  until you manually update the pin.
```

Also not sure about the word "manually" for this strategy, since many people use dependabot, renovate, or similar tools these days

README.md
```diff
@@ -46,6 +46,52 @@ and are automatically released (multiple times a day, when needed) by
...
+code fail to type check.
+
+There are several strategies available for specifying the version of a stubs
+package you're using, each with its own tradeoffs
+
+1. Use the same pin that you use for the package being stubbed. For example,
+  if you use `requests>=2.30.0,<2.32`, you can use
+  `types-requests>=2.30.0,<2.32`. This ensures that the stubs are compatible
+  with the package you are using, but it carries a small risk of breaking
+  type checking due to changes in the stubs.
```
AlexWaygood, line 72: The indentation looks weird here, now that it's switched to a numbered list -- I think it all needs to be indented 3 spaces rather than two, now

README.md
```diff
@@ -46,6 +46,52 @@ and are automatically released (multiple times a day, when needed) by
...
+  could be unable to update the stubs, since an update has not been released.
+2. Pin the stubs to a known good version and update manually. For example, if
+  you use `types-requests==2.31.0.1`, you can be sure that upgrading
+  dependencies will not break type checking. However, you will miss out on
+  improvements in the stubs that could potentially improve type checking,
+  unless you manually update the stubs.
+  It also has the risk that the stubs you are using are not compatible with
+  the package being stubbed.
+3. Don't pin the stubs. This is the most flexible option, but it carries the
+  risk that the stubs become incompatible with the package being stubbed.
```
AlexWaygood, line 86:
```suggestion
3. Don't pin the stubs. This is the option that demands the least work from
   you when it comes to updating version pins, and has the advantage that you
   will automatically benefit from improved stubs whenever a new version of the
   stubs package is released. However, it carries the risk that the stubs
   become incompatible with the package being stubbed.

```

README.md
```diff
@@ -46,6 +46,52 @@ and are automatically released (multiple times a day, when needed) by
...
+2. Pin the stubs to a known good version and update manually. For example, if
+   you use `types-requests==2.31.0.1`, you can be sure that upgrading
+   dependencies will not break type checking. However, you will miss out on
+   improvements in the stubs that could potentially improve type checking,
+   until you update the pin.
+   It also has the risk that the stubs you are using are not compatible with
+   the package being stubbed.
+3. Don't pin the stubs. This is the most flexible option, but it carries the
+   risk that the stubs become incompatible with the package being stubbed.
+   For example, if a new major version of the package is released, there's a
```
AlexWaygood, line 87:
```suggestion
3. Don't pin the stubs. This is the option that demands the least work from
   you when it comes to updating version pins, and has the advantage that you
   will automatically benefit from improved stubs whenever a new version of the
   stubs package is released. However, it carries the risk that the stubs
   become incompatible with the package being stubbed.

   For example, if a new major version of the package is released, there's a
```

README.md
```diff
@@ -46,6 +46,52 @@ and are automatically released (multiple times a day, when needed) by
...
+   the package being stubbed. You might want to force the package being stubbed
+   to a certain minimum version, because it fixes a critical bug, but you
+   could be unable to update the stubs, since an update has not been released.
+2. Pin the stubs to a known good version and update manually. For example, if
+   you use `types-requests==2.31.0.1`, you can be sure that upgrading
+   dependencies will not break type checking. However, you will miss out on
+   improvements in the stubs that could potentially improve type checking,
+   until you update the pin.
+   It also has the risk that the stubs you are using are not compatible with
+   the package being stubbed.
```
AlexWaygood, line 84:
```suggestion
2. Pin the stubs to a known good version and update the pin from time to time
   (either manually, or using a tool such as dependabot or renovate).

   For example, if you use `types-requests==2.31.0.1`, you can have confidence
   that upgrading dependencies will not break type checking. However, you will
   miss out on improvements in the stubs that could potentially improve type
   checking until you update the pin. This strategy also has the risk that the
   stubs you are using might become incompatible with the package being stubbed.
```

README.md
```diff
@@ -46,6 +46,52 @@ and are automatically released (multiple times a day, when needed) by
...
+to the stubs were pushed (but the stubs were still aiming to provide
+annotations for `foo==1.1.2`), then the version of `types-foo` would
+increment to `1.1.2.1`.
+
+At typeshed, we try to keep breaking changes to a minimum. However, due to the
+nature of stubs, any version bump can introduce changes that might make your
+code fail to type check.
+
+There are several strategies available for specifying the version of a stubs
+package you're using, each with its own tradeoffs
```
JelleZijlstra, line 66:
```suggestion
package you're using, each with its own tradeoffs:
```

AlexWaygood, line 66:
```suggestion
package you're using, each with its own tradeoffs:
```

README.md
```diff
@@ -46,6 +46,52 @@ and are automatically released (multiple times a day, when needed) by
...
+   It also has the risk that the stubs you are using are not compatible with
+   the package being stubbed.
+3. Don't pin the stubs. This is the most flexible option, but it carries the
+   risk that the stubs become incompatible with the package being stubbed.
+   For example, if a new major version of the package is released, there's a
+   chance the stubs might be updated to reflect the new version of the runtime
+   package before you update the package being stubbed.
+
+You can also use the different strategies as needed. For example, you could
+default to strategy (1), but fall back to strategy (2) when
```
AlexWaygood, line 92:
```suggestion
You can also switch between the different strategies as needed. For example, you
could default to strategy (1), but fall back to strategy (2) when
```

README.md
AlexWaygood: duplicate of https://github.com/python/typeshed/pull/10840#discussion_r1347533070 ;)

JelleZijlstra: I spent too much time convincing myself that the "at least four parts" is right :) (Seems like pywin32 itself uses a two-part versioning scheme, but we add enough .0s to indeed make it four-part.)

AlexWaygood (approved):
This is great. Thanks!

README.md
```diff
@@ -46,6 +46,57 @@ and are automatically released (multiple times a day, when needed) by
...
+
+Version numbers of third-party stub packages consist of at least four parts.
+All parts of the stub version, except for the last part, correspond to the
+version of the runtime package being stubbed. For example, if the `types-foo`
+package has version `1.1.2.0`, this signifies that the package contains stubs
+for `foo==1.1.2`. In this example, the final element of the version
+indicates that this is the first revision of the stubs for `foo`. If an update
+to the stubs were pushed (but the stubs were still aiming to provide
+annotations for `foo==1.1.2`), then the version of `types-foo` would
+increment to `1.1.2.1`.
```
AlexWaygood, line 59: What happens if `foo` at runtime has version `1.1.1`, and typeshed has `version = 1.1.*` in `stubs/foo/METADATA.toml`? Will the version for `types-foo` be `1.1.0.whatever`, or `1.1.1.whatever`?

AlexWaygood: Hmm, looks like the version numbers can get out of sync in cases like that:
- `Deprecated` has version `2.1.14` at runtime: https://pypi.org/project/Deprecated/
- https://github.com/python/typeshed/blob/54a8fd74a785ff7916bade3699a2be4542f1b1c0/stubs/Deprecated/METADATA.toml#L1
- `types-Deprecated` is currently on version `2.1.9.3`: https://pypi.org/project/types-Deprecated/.

AlexWaygood: Maybe something like this?

```suggestion
version of the runtime package being stubbed. For example, if the `types-foo`
package has version `1.1.2.0`, this signifies that the package contains stubs
for `foo>=1.1.2`, but which have not been tested against `foo`>=1.2`.
In this example, the final element of the version indicates that this is the
first revision of the stubs for `foo>=1.1.2`. If an update to the stubs were
pushed (but the stubs were still aiming to provide annotations for
`foo>=1.1.2`), then the version of `types-foo` would increment to `1.1.2.1`.
```

README.md
```diff
@@ -46,6 +46,57 @@ and are automatically released (multiple times a day, when needed) by
...
+1. Use the same pin that you use for the package being stubbed. For example,
+   if you use `requests>=2.30.0,<2.32`, you can use
+   `types-requests>=2.30.0,<2.32`. This ensures that the stubs are compatible
+   with the package you are using, but it carries a small risk of breaking
+   type checking due to changes in the stubs.
+
+   Another risk of this strategy is that stubs often lag behind
+   the package being stubbed. You might want to force the package being stubbed
+   to a certain minimum version, because it fixes a critical bug, but you
+   could be unable to update the stubs, since an update has not been released.
```
hauntsaninja, line 77:
```suggestion
   to a certain minimum version because it fixes a critical bug, but if
   correspondingly updated stubs have not been released, your type
   checking results may not be fully accurate.
```
nit: in practice, slightly stale stubs still usually work; I don't want to make it sound like you won't be able to get work done.

hauntsaninja (commented):
(A little late to this, so I haven't read the existing resolved comments)

README.md
```diff
@@ -46,6 +46,57 @@ and are automatically released (multiple times a day, when needed) by
...
+
+Version numbers of third-party stub packages consist of at least four parts.
+All parts of the stub version, except for the last part, correspond to the
+version of the runtime package being stubbed. For example, if the `types-foo`
+package has version `1.1.2.0`, this signifies that the package contains stubs
+for `foo>=1.1.2`, but which have only been tested against `foo<1.2`.
+In this example, the final element of the version indicates that this is the
+first revision of the stubs for `foo>=1.1.2`. If an update to the stubs were
+pushed (but the stubs were still aiming to provide annotations for
+`foo>=1.1.2`), then the version of `types-foo` would increment to `1.1.2.1`.
```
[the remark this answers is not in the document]
AlexWaygood: (Edited your suggestion to fix a minor typo)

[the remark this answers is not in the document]
AlexWaygood: I do still find the current situation with types-Deprecated that I described in https://github.com/python/typeshed/pull/10840#discussion_r1348644670  slightly hard to square with any explanation of our versioning scheme tbh. @hauntsaninja, any idea what's going on there?

README.md
hauntsaninja: There is some legacy there because of interaction with previous versioning. stub_uploader sees a) the specifier `"1.2.*"`, b) the fact that the last thing we uploaded was `1.2.9.3` (and at time of version swap last thing we uploaded was `1.2.9`) and just bumps.

The next time stubsabot bumps the version specifier, everything will become good. Specifier will be `1.3.*`, `Deprecated` will be `1.3.x` and we will be `1.3.0.y`.

Note the cause the complexity here is that we don't maintain multiple versions of stubs, so we do want to have each version of stub be workable for a specifier, but we can't put `*` into versions, so (assuming no legacy) we put `0` if possible

README.md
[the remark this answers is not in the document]
hauntsaninja: Yeah, it's little confusing, commented on why it does what it does

JelleZijlstra (approved):

srittau:
I think I integrated all comments, but I wouldn't be surprised if there's still a lot of fine-tuning possible in future PRs.

PR #10843, opened 2023-10-05
Add datadog stubs

drichardson:


drichardson:
oops, created in wrong repo, apologies 😅

AlexWaygood:
Looks to me like datadog has a py.typed file FYI, so adding stubs shouldn't really be necessary, I don't think :)

drichardson:
> It looks to me like datadog has a py.typed file FYI, so adding stubs shouldn't really be necessary, I don't think :)

Oh interesting. I see that marker exists, but they don't have type hints anywhere.

drichardson:
Ah actually they are just using the older style comment type hints.

drichardson:
OK so the guidance here would  be to try to get the functions missing type hints appropriately annotated in the datadog repo, correct?

AlexWaygood:
> OK so the guidance here would be to try to get the functions missing type hints appropriately annotated in the datadog repo, correct?

Yup!

Issue #10844, opened 2023-10-06
`PyInstaller.building.api.EXE` is missing `hide_console`


bersbersbers:
Compare https://github.com/pyinstaller/pyinstaller/pull/7735 and

> (Windows) Add an option to hide or minimize the console window in console-enabled applications, but only if the program’s process owns the console window (i.e., the program was not launched from an existing console window). ([#7729](https://github.com/pyinstaller/pyinstaller/issues/7729))

from https://pyinstaller.org/en/v6.0.0/CHANGES.html#id1

srittau:
Indeed. This should probably be added for all platforms, because it exists on all platforms and just issues a warning on non-Windows platforms. PR welcome!

This should have been caught by our tests, but wasn't, since we have an overzealous stubtest allowlist entry: `PyInstaller\.building\.\w+?`.

PR #10845, opened 2023-10-06
stubtest: List ignored `building` sub-modules explicitly

srittau:
This prevents stubtest from ignoring errors in modules that are stubbed.

### DIFF
--- a/stubs/pyinstaller/@tests/stubtest_allowlist.txt
+++ b/stubs/pyinstaller/@tests/stubtest_allowlist.txt
@@ -9,20 +9,20 @@ PyInstaller.__main__.run_makespec
 PyInstaller.utils.hooks.conda.lib_dir
 
 # A mix of modules meant to be private, and shallow incomplete type references for other modules
-PyInstaller.building.build_main.*
+PyInstaller\.building\.build_main\..*
 PyInstaller.building.datastruct.unique_name
-PyInstaller.building.icon.*
-PyInstaller.building.makespec.*
-PyInstaller.building.osx.*
-PyInstaller.building.splash_templates.*
-PyInstaller.building.templates.*
-PyInstaller.building.utils.*
-PyInstaller.depend.analysis.*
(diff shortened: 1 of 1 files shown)


srittau:
Cf #10844

stubs/pyinstaller/@tests/stubtest_allowlist.txt
```diff
@@ -9,9 +9,14 @@ PyInstaller.__main__.run_makespec
...
 # A mix of modules meant to be private, and shallow incomplete type references for other modules
-PyInstaller\.building\.\w+?
 PyInstaller.building.build_main.*
 PyInstaller.building.datastruct.unique_name
+PyInstaller.building.icon.*
+PyInstaller.building.makespec.*
+PyInstaller.building.osx.*
+PyInstaller.building.splash_templates.*
+PyInstaller.building.templates.*
+PyInstaller.building.utils.*
```
AlexWaygood, line 19:
```suggestion
PyInstaller.building.icon
PyInstaller.building.makespec
PyInstaller.building.osx
PyInstaller.building.splash_templates
PyInstaller.building.templates
PyInstaller.building.utils
```

Stubtest still passes with this change (I tested locally). If the module itself is missing, it only results in one stubtest error, so there's no need for regexes here. Avoiding regexes means that we'll get a helpful "unused allowlist" error if we ever add stubs for one of these modules, and we won't necessarily be accidentally allowlisting anything _in_ the module if we ever add stubs for that module.

We should probably get rid of the `.*`s in some of the other entries in this allowlist, too.

srittau: I cleaned up the list further. The section titled "A mix of modules meant to be private, and shallow incomplete type references for other modules" is a bit suspect to me, too, but I don't want to delve into that now.

AlexWaygood (approved):
Nice, thanks

PR #10849, opened 2023-10-06
Fix `ssl.SSLSocket` methods on windows

sobolevn:
Here's what is going on.

1. These three methods always fail with `NotImplementedError` in CPython: https://github.com/python/cpython/blob/201dc11aeb4699de3c5ebaea9798796c30087bcc/Lib/ssl.py#L1302-L1308 and  https://github.com/python/cpython/blob/201dc11aeb4699de3c5ebaea9798796c30087bcc/Lib/ssl.py#L1228-L1232 
2. But, they are conditionally defined on `socket.socket` on all patforms, except Windows: https://github.com/python/cpython/blob/201dc11aeb4699de3c5ebaea9798796c30087bcc/Modules/socketmodule.c#L5192-L5199
3. So, stubtest fails on Win (this is why we have these entries in allowlist), because it sees these 3 methods in runtime, but not in stubs

I think that the current case is suboptimal, because with the proposed solution we can detect `SSLObject.sendmsg` usage and provide at least some error checking to users.


AlexWaygood (commented):
Maybe the `*args` and `**kwargs` should be annotated with `Never` rather than `Unused`? You're probably doing something wrong if you're passing arguments to these methods; it would be good for the type checker to emit an error in that case.

AlexWaygood (approved):

PR #10854, opened 2023-10-07
Use Python 3.11 for the stub_uploader tests in the daily test

### LINKED ISSUE #10853: Daily tests failed on Sat Oct 07 2023

AlexWaygood:
Hopefully fixes #10853


AlexWaygood:
This is an easy fix, and hopefully an uncontroversial change to do anyway.

Maybe the test in stub_uploader could also be deleted, or skipped on Python <3.11. But that can be considered separately.

PR #10855, opened 2023-10-07
Remove allowlist entries for `_ctypes` on win

sobolevn:
Source: https://github.com/python/cpython/blob/64f158e7b09e67d0bf5c8603ff88c86ed4e8f8fd/Modules/_ctypes/callproc.c#L1361-L1433


stdlib/_ctypes.pyi
```diff
@@ -44,6 +44,8 @@ if sys.platform == "win32":
     def FormatError(code: int = ...) -> str: ...
     def get_last_error() -> int: ...
     def set_last_error(value: int) -> int: ...
+    def LoadLibrary(__name: str, __load_flags: int) -> int: ...
```
srittau, line 47:
```suggestion
    def LoadLibrary(__name: str, __load_flags: int = 0) -> int: ...
```

srittau (changes requested):
Thanks! I believe `load_flags` has a default value, though:

https://github.com/python/cpython/blob/64f158e7b09e67d0bf5c8603ff88c86ed4e8f8fd/Modules/_ctypes/callproc.c#L1375C10-L1375C26

stdlib/_ctypes.pyi
AlexWaygood: Can confirm from local testing on a Windows machine.

sobolevn: Yes, it does have `int load_flags = 0;`

AlexWaygood (approved):
lgtm

srittau (approved):

PR #10856, opened 2023-10-07
Improved the README.md readablity.

TYehan:
Overall, these improvements aim to provide a better reading experience for users, making it easier for them to access information and understand the content's key points.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,8 +6,8 @@
 
 ## About
 
-Typeshed is a repository containing external type annotations for the Python standard library, Python builtins, and third-party packages contributed by individuals external to those projects. 
-  
+Typeshed is a repository containing external type annotations for the Python standard library, Python builtins, and third-party packages contributed by individuals external to those projects.
+
 > **These data are used for purposes such as static analysis, type checking, and type inference.**
 
 For information on how to use `typeshed`, read below.  Information for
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -24,51 +20,45 @@ Typeshed fully supports Python versions 3.8 and up. Support for Python 3.7
...
 
 Type checkers should be able to use these stub packages when installed. For more
 details, see the documentation for your type checker.
 
+---
 ### The `_typeshed` package
 
 typeshed includes a package `_typeshed` as part of the standard library.
-This package and its submodules contains utility types, but is not
+This package and its submodules contain utility types but are not
```
AlexWaygood, line 46: You're correct that the grammar isn't spot-on here, but I don't think it's a major issue, really. And if anything, your PR makes it worse. If we're being picky about this, it should probably be something like

```md
This package and its submodules contain utility types, but the package is not
```

AlexWaygood (commented):
Thanks, but almost all of these changes offer no significant improvement to the existing text.

In fact, many of these changes are harmful: it's much better for accessibility, and much easier to edit, if each line is wrapped to c. 80 characters. Your PR creates several extremely long lines, and I'm not sure why.

I'm closing this PR, I'm afraid, as I don't think it's solving a real problem.

PR #10857, opened 2023-10-07
Fix platform availability of some Unix constants

AlexWaygood:


stdlib/os/__init__.pyi
```diff
@@ -70,9 +70,20 @@ if sys.platform != "win32":
...
+        # sys.platform could have one of many values on a FreeBSD Python build,
+        # so the sys-module docs recommend doing `if sys.platform.startswith('freebsd')`
+        # to detect FreeBSD builds. Unfortunately that would be too dynamic
+        # for type checkers, however.
+        SF_NODISKIO: int
+        SF_MNOWAIT: int
+        SF_SYNC: int
+
+        if sys.version_info >= (3, 11):
+            SF_NOCACHE: int
```
AlexWaygood, line 86: `os` docs:

- https://docs.python.org/3/library/os.html#os.SF_NODISKIO
- https://docs.python.org/3/library/os.html#os.SF_NOCACHE

Source code: https://github.com/python/cpython/blob/de2a4036cbfd5e41a5bdd2b81122b7765729af83/Modules/posixmodule.c#L16562-L16575

stdlib/select.pyi
```diff
@@ -137,7 +137,6 @@ if sys.platform == "linux":
     EPOLLRDNORM: int
     EPOLLWRBAND: int
     EPOLLWRNORM: int
-    EPOLL_RDHUP: int
```
AlexWaygood, line 140: I don't think this ever existed at runtime. I think it was just a figment of our imagination. Note the name is very similar to `select.EPOLLRDHUP`, which _does_ exist at runtime (and which we already have elsewhere in the stub).

stdlib/_curses.pyi
```diff
@@ -61,7 +61,8 @@ if sys.platform != "win32":
     A_DIM: int
     A_HORIZONTAL: int
     A_INVIS: int
-    A_ITALIC: int
+    if sys.platform != "darwin":
+        A_ITALIC: int
```
AlexWaygood, line 65: Docs here: https://docs.python.org/3/library/curses.html#curses.A_ITALIC

Immediately above this table, the docs state:

> Some constants are available to specify character cell attributes. The exact constants available are system dependent.

JelleZijlstra: I tried and it indeed exists on Linux but not macOS

stdlib/os/__init__.pyi
```diff
@@ -70,9 +70,20 @@ if sys.platform != "win32":
         POSIX_FADV_WILLNEED: int
         POSIX_FADV_DONTNEED: int
 
-    SF_NODISKIO: int
-    SF_MNOWAIT: int
-    SF_SYNC: int
+    if sys.platform != "linux" and sys.platform != "darwin":
+        # In the os-module docs, these are marked as being available
+        # on "Unix, not Emscripten, not WASI."
```
JelleZijlstra, line 75: So maybe we should change the os docs?

Can confirm no os.SF_* contants on Linux or macOS.

stdlib/signal.pyi
```diff
@@ -47,6 +45,9 @@ class Signals(IntEnum):
         SIGWINCH: int
         SIGXCPU: int
         SIGXFSZ: int
+        if sys.platform != "linux":
```
JelleZijlstra, line 48: For me these indeed exist on macOS but not Linux.

JelleZijlstra (approved):
Thanks!

stdlib/os/__init__.pyi
AlexWaygood: > So maybe we should change the os docs?

Yeah, I think so

PR #10859, opened 2023-10-08
Add `_locale` module

sobolevn:


stdlib/_locale.pyi
```diff
@@ -0,0 +1,96 @@
...
+if sys.version_info >= (3, 11):
+    def getencoding() -> str: ...
+
+def strcoll(__os1: str, __os2: str) -> int: ...
+def strxfrm(__string: str) -> str: ...
+
+# native gettext functions
+# https://docs.python.org/3/library/locale.html#access-to-message-catalogs
+# https://github.com/python/cpython/blob/f4c03484da59049eb62a9bf7777b963e2267d187/Modules/_localemodule.c#L626
+if sys.platform == "linux" or sys.platform == "darwin":
```
AlexWaygood, line 25: Are you sure these only exist on linux and darwin, and not on any other unix systems? This is a bit of a smell to me; at a guess, I'd think this would probably be more accurate:

```suggestion
if sys.platform != "win32":
```

stdlib/locale.pyi
```diff
@@ -34,97 +32,102 @@ if sys.version_info >= (3, 11):
...
+    localeconv as localeconv,
+    setlocale as setlocale,
+    strcoll as strcoll,
+    strxfrm as strxfrm,
+)
+
+if sys.version_info >= (3, 11):
+    from _locale import getencoding as getencoding
+
+# However, some parts of `_local` module is platform-specific:
```
AlexWaygood, line 52:
```suggestion
# Some parts of the `_locale` module are platform-specific:
```

stdlib/locale.pyi
```diff
@@ -34,97 +32,102 @@ if sys.version_info >= (3, 11):
...
+if sys.platform == "linux" or sys.platform == "darwin":
```
AlexWaygood, line 53: Same comment as above: are you sure these only exist on linux/darwin, and not on any other unix systems?

stdlib/locale.pyi
```diff
@@ -34,97 +32,102 @@ if sys.version_info >= (3, 11):
...
+        PM_STR as PM_STR,
+        RADIXCHAR as RADIXCHAR,
+        T_FMT as T_FMT,
+        T_FMT_AMPM as T_FMT_AMPM,
+        THOUSEP as THOUSEP,
+        YESEXPR as YESEXPR,
+        nl_langinfo as nl_langinfo,
+    )
+
+    __all__ += ["LC_MESSAGES"]
```
AlexWaygood, line 114: Could you move this up to just below line 33, so that all the `__all__` definitions are in the same place in the file? Linking to https://github.com/python/cpython/blob/d96a8cdfe201546f6d077fa93953b695c4ea8c59/Lib/locale.py#L1726-L1731 in a comment might also be good, as well

tests/stubtest_allowlists/darwin-py311.txt
```diff
@@ -1,7 +1,8 @@
...
-(locale.dcgettext)?
-(locale.dgettext)?
-(locale.gettext)?
-(locale.textdomain)?
+(_locale.bind_textdomain_codeset)?
+(_locale.bindtextdomain)?
+(_locale.dcgettext)?
+(_locale.dgettext)?
+(_locale.gettext)?
+(_locale.textdomain)?
```
AlexWaygood, line 8: It looks like this PR changes things so that these are no longer reexported from `locale.pyi`. Is that deliberate?

If they were being reexported from `locale.pyi`, the allowlist entries would need to be something like

```suggestion
(_?locale.bind_textdomain_codeset)?
(_?locale.bindtextdomain)?
(_?locale.dcgettext)?
(_?locale.dgettext)?
(_?locale.gettext)?
(_?locale.textdomain)?
```

AlexWaygood (commented):
Nice! A few comments:

stdlib/_locale.pyi
sobolevn: We need freebsd runner to test this :)
I am fine with both for now.

sobolevn:
Programming is broken: 
Снимок экрана 2023-10-08 в 13 39 15

AlexWaygood (approved):
Fab, thank you! This was possibly the hardest stdlib allowlist entry to solve other than this one :D

https://github.com/python/typeshed/blob/64d6d965685bc175606a93b9c6f90de224e97fe0/tests/stubtest_allowlists/linux.txt#L1

PR #10866, opened 2023-10-11
Fix mypy_test if mypy returns negative exit code

srittau:


tests/mypy_test.py
```diff
@@ -11,6 +11,7 @@
 import sys
 import tempfile
 import time
+from calendar import c
```
AlexWaygood, line 14: ?

tests/mypy_test.py
```diff
@@ -573,8 +575,8 @@ def main() -> None:
             total_packages_skipped += packages_skipped_this_version
     if code:
         plural = "" if total_files_checked == 1 else "s"
-        print_error(f"--- exit status {code}, {total_files_checked} file{plural} checked ---")
-        sys.exit(code)
+        print_error(f"--- mypy exit status {code}, {total_files_checked} file{plural} checked ---")
+        sys.exit(1)
```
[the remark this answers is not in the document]
srittau: I don't think it makes much sense to propagate just one exit code to the end, since mypy might fail multiple times. I'll just add an enum to signal the "worst" exit code.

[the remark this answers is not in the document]
AlexWaygood: Yeah, agreed that the existing behaviour didn't really make much sense! I like the change you made to print the returncode for each mypy failure.

AlexWaygood (commented):
Things I like here:
- The existing behaviour of propagating the "maximum" returncode didn't make much sense; this PR fixes it
- I love the new `MypyResult` enum

I'm not so sure about the changes to the `TestResults` class, though. Something that struck me as I was reviewing #10724 was that it's a bit weird that we use the same `TestResults` class to represent "the results of running mypy on one stubs package" and "the results of running `mypy_test.py` on the whole of typeshed". Those are conceptually different tasks, and it's already a bit of a leaky abstraction on `main` -- having a `packages_skipped` attribute on a `TestResults` instance representing the results of running mypy on one stubs package doesn't really make much sense. I think we should probably separate out the two concepts into two different "result" classes, but this PR feels like it's going in the opposite direction a little bit

srittau:
We could let `test_stdlib()` and `test_third_party_distribution()` just return mypy's return code and files tested, but all other functions are already aggregating results.

tests/mypy_test.py
```diff
@@ -361,14 +371,36 @@ def test_stdlib(code: int, args: TestConfig) -> TestResults:
...
+
+    def skip_package(self) -> TestSummary:
+        return TestSummary(self.mypy_result, self.files_checked, self.packages_skipped + 1)
+
+    def combine(self, other: TestSummary) -> TestSummary:
+        result = self.mypy_result if self.mypy_result.value > other.mypy_result.value else other.mypy_result
+        checked = self.files_checked + other.files_checked
+        skipped = self.packages_skipped + other.packages_skipped
+        errors = self.packages_with_errors + other.packages_with_errors
+        return TestSummary(result, checked, skipped, errors)
```
AlexWaygood, line 403: Maybe the dataclass should be non-frozen? Then these methods could update the dataclass in-place, and return `None`? No strong opinion, though

AlexWaygood (approved):
Thanks, this looks great! Just one more optional nit:

tests/mypy_test.py
srittau: I am trying to avoid having to pass the previous summaries into the various functions so they are a bit more "standalone". Not particular necessary in our code, but I found this a bit cleaner.

srittau: Although you are right, for the register/skip method this makes sense.

AlexWaygood: Give me a moment, I'll come up with a suggestion

srittau: I was mistaken anyway. Unfreezing it has no effect on passing it to functions.

AlexWaygood (approved):
Looks great. Thanks!

PR #10867, opened 2023-10-11
Bump WTForms to 3.1.*

sobolevn:
CC @Daverball


Daverball (commented):
Looks good as far as the things stubtest could detect.

However there was also another change that wasn't detected by stubtest to the `choices` parameter and `iter_choices` in `SelectFieldBase`.

`choices` now supports an optional third tuple element to pass `render_kw` to the `SelectFieldBase._Option` field and `iter_choices` as a result now yields this as a fourth tuple element.

Relevant changes are here: https://github.com/wtforms/wtforms/compare/3.0.1...3.1.0#diff-946c78ea141ed1ff5a8c866568adb3e658363228c09b2e009bbc3468c85dbbd3

The parameter change is not as urgent, since passing the choices the old way is still supported, but `iter_choices` should probably be changed now. I can do a follow-up PR for the parameter, unless you don't mind changing it yourself, although I would probably add some regression tests to make sure both shapes are accepted properly.

Daverball:
In short, I'd change `_FullChoice: TypeAlias = tuple[Any, str, bool]  # value, label, selected` to `_FullChoice: TypeAlias = tuple[Any, str, bool, dict[str, Any]  # value, label, selected, render_kw`

For `_Choice` you could try `tuple[Any, str] | tuple[Any, str, dict[str, Any]` although I'd be concerned about it rejected mixed choices.

sobolevn:
> However there was also another change that wasn't detected by stubtest to the choices parameter and iter_choices in SelectFieldBase.

> The parameter change is not as urgent, since passing the choices the old way is still supported

Let's do it in an another PR then :)

Daverball:
> > However there was also another change that wasn't detected by stubtest to the choices parameter and iter_choices in SelectFieldBase.
> 
> > The parameter change is not as urgent, since passing the choices the old way is still supported
> 
> Let's do it in an another PR then :)

Sounds good to me, `_GroupedChoices` is probably already problematic anyways due to the use of `dict`. I can't recall if the implementation actually checks for `dict`. But it's certainly annoying with the lack of covariance. So adding some tests there is a good idea either way, I will do that in a follow-up PR.

Just changing `_FullChoice` should be good enough for now, that should result in appropriate errors in people's code when they rely on `iter_choices`.

JelleZijlstra:
Fixed by #10868, thanks @sobolevn!

PR #10875, opened 2023-10-13
[requests] Sync HttpError stub to implementation

ubersan:
The [implementation](https://github.com/psf/requests/blob/main/src/requests/exceptions.py#L45) has no restrictions on providing a response. Is there a reason why the stub requires it?


srittau:
Cf #10764

srittau:
This PR is problematic, as the primer output shows. A lot of code assumes that `HTTPError.response` is not `None`, which is true for the only place in requests that raises an `HTTPError`:

https://github.com/psf/requests/blob/b55bb15c35132f021fc0997c6a72730abe2216f5/src/requests/models.py#L1019

On the hand, some code outside requests that constructs `HTTPError`s assumes that the `response` argument is optional. Even requests has a test that asserts that:

https://github.com/psf/requests/blob/b55bb15c35132f021fc0997c6a72730abe2216f5/tests/test_requests.py#L1542

Especially considering the test in requests, I propose to accept this PR, even if it means that some code needs asserts or type ignore. It's also the safer alternative.

srittau (approved):
LGTM, see my last comment, but I would like another maintainer to chime in.

AlexWaygood:
There's no way of making everybody happy here, but this approach does have the advantage that it is, well, correct.

ubersan:
thanks for the comments, I agree it's not super nice to break quite some apps, but at least we're starting to being aligned again with the implementation going forward

intgr:
FWIW I think this is a step backwards.

srittau:
@intgr In the this is an issue to be taken up with the requests project. It would make sense to me that `HTTPError` requires the response to be set, but this isn't the case in the implementation, and in fact there is an explicit test in requests that `HTTPError` can be used without a response set.

bluetech:
This change has caused quite a lot of busy work for us as well. I think that in practice, it would have been better to leave `response` not optional. `requests` always passes a response to `HTTPError`, and "HTTP error" itself sort of implies that a response be present (otherwise it's not really an HTTP error...). I think that outweighs type unsafety from some 3rd party packages not passing a response. But I understand @srittau's rationale and I wouldn't change it back at this point.

asottile-sentry:
is there any way to change this so that by default it has `Response` -- but when constructed directly it does not?  this change is a big step backwards otherwise

Akuli:
I think this should be reverted.

As already mentioned, the only place in requests that raises `HTTPError` is:

```python3
        if http_error_msg:
            raise HTTPError(http_error_msg, response=self)
```

So practically, when you do `except HTTPError`, the response will always be set. I think handling an exception with try/except is by far the most important use case for exception type stubs in a library. So making it "just work" should be our main goal.

IMO the unit test that initializes response to `None` doesn't mean anything. When you write unit tests, you tend to write unit tests for everything, including that dumb feature that seemed like a good idea at the time but was not needed after all. To me, the test would be more meaningful if it was more integration test style than unit test style (for example, if the test actually made requests throw `HTTPError` with no response set).

Is there really a practical problem that this PR solves, and is it really worth solving? If there is, and I'm just unaware of it, I'd suggest using the Any trick (now documented in CONTRIBUTING.md).

srittau:
The practical problem is that people initialize `HTTPError` in their own code.

Akuli:
> is there any way to change this so that by default it has Response -- but when constructed directly it does not?

Not really, but we did the next best thing in #11207: you can now assume it has a `Response` or check whether it's missing without any type checking errors. A new version of `types-requests` with the fix will be released automatically in about 12 hours.

Issue #10876, opened 2023-10-13
`OAuth2Session.authorization_url` is incorrectly typed


MattExact:
PR #10658 recently added support for `requests-oauthlib`
However, there is an incorrect type for `OAuth2Session.authorization_url`

https://github.com/python/typeshed/blob/cc38a755878f31932a245152963ea52ef864a3cf/stubs/requests-oauthlib/requests_oauthlib/oauth2_session.pyi#L74

[If you look at the source code for the method](https://github.com/requests/requests-oauthlib/blob/3a2a852e33c691c7e793300ce366a01b6e4b3848/requests_oauthlib/oauth2_session.py#L170-L189), it returns a `tuple[str, str]`, not a `str`.

PR #10878, opened 2023-10-13
Fix the type for `OAuth2Session.authorization_url`

MattExact:
Fixes https://github.com/python/typeshed/issues/10876

[See the source for the method here](https://github.com/requests/requests-oauthlib/blob/3a2a852e33c691c7e793300ce366a01b6e4b3848/requests_oauthlib/oauth2_session.py#L170-L189)


AlexWaygood (approved):
Thanks!

PR #10879, opened 2023-10-13
Complete type stubs for xml.dom.minidom (#6886)

### LINKED ISSUE #6886: xml typing incomplete

zackw:
This patch completes all the type stubs for xml.dom.minidom.  I did not look at the rest of the xml.dom directory except as required to make the typing of minidom complete, so it probably doesn’t completely resolve #6886.

xml.dom.minidom is very poorly documented—its authors seem to have assumed that one could refer to the W3C DOM spec.  But at the same time the implementation diverges significantly from the spec, so I do not have complete confidence in all the type annotations.  Still, it should be an improvement on what’s there now.

The following commands now pass with no errors when executed from the top of the typeshed tree:

```sh
mypy --custom-typeshed-dir . --strict --warn-incomplete-stub \
     -m xml.dom.minidom
stubtest --custom-typeshed-dir . xml.dom.minidom
```

**However**, attempting to do the same for xml.dom.minicompat produces the following error:

```sh
mypy --custom-typeshed-dir . --strict --warn-incomplete-stub \
     -m xml.dom.minicompat
stdlib/xml/dom/minicompat.pyi:11: error: All bases of a protocol must be protocols  [misc]
```

I don’t know how to fix this, because in the implementation, EmptyNodeList is not a subclass of NodeList, therefore if I don’t make NodeList a protocol, then xml.dom.minidom.Childless fails stubtest. This patch probably should not land until this is resolved.  Any advice you could give would be welcome.


zackw:
It looks like all the failures are due to the xml.dom.minicompat issue I mentioned up top.

For clarity, this is one of those places where the implementation diverges from the spec.  All implementations of the `Node` abstract interface are required to have a `childNodes` property that implements the `NodeList` interface. xml.dom.minicompat defines _two_ concrete implementations of that interface: `NodeList`, which is a subclass of `list`, and `EmptyNodeList`, which is a subclass of `tuple`.  Neither is a subclass of the other. xml.dom.minidom uses xml.dom.minicompat.NodeList for `childNodes` for some types that implement Node, and xml.dom.minicompat.EmptyNodeList for other types that implement Node.

It ought to be perfectly fine for _users_ of xml.dom.minidom to treat the type of `childNodes` as always being `NodeList`, no matter what.  That's how the spec designers intended it to be used, and I can get xml.dom.minidom to typecheck correctly with

```
from xml.dom.minicompat import NodeList
class Node:
    childNodes: NodeList
```

if I make xml.dom.minicompat.NodeList be a Protocol in the type stubs. But `tuple`, `list`, and `Sequence` are _not_ Protocols, and so that same change breaks the typechecking of minicompat itself.  I can't change the relationship between EmptyNodeList and NodeList unilaterally from the stubs -- that would require a change to the actual stdlib. I could put minicompat back the way it was and have Node be typed like so instead

```
from xml.dom.minicompat import EmptyNodeList, NodeList
class Node:
    childNodes: NodeList | EmptyNodeList
```

but then every user of the childNodes property would have to know about EmptyNodeList, which the authors of the stdlib intended to be a pure implementation detail. The entire minicompat module is undocumented in the stdlib. So I don't like that option either.

Again, any advice you could give would be welcome.

stdlib/xml/dom/minicompat.pyi
```diff
@@ -1,14 +1,14 @@
...
 
 __all__ = ["NodeList", "EmptyNodeList", "StringTypes", "defproperty"]
 
-_T = TypeVar("_T")
+_T = TypeVar("_T", covariant=True)
 
 StringTypes: tuple[type[str]]
 
-class NodeList(list[_T]):
+class NodeList(Sequence[_T], Protocol):
```
hamdanal, line 11: This doesn't look right. At runtime `NodeList` inherits from `list` and it is indeed a list and its append method is used elsewhere in the code. Why was it changed?

stdlib/xml/dom/minidom.pyi
```diff
@@ -95,74 +101,73 @@ class Node(xml.dom.Node):
         ) -> bytes: ...
 
     def hasChildNodes(self) -> bool: ...
-    def insertBefore(self, newChild, refChild): ...
+    def insertBefore(self, newChild: _N, refChild: Node) -> _N: ...
```
hamdanal, line 104: There is a `if refChild is None` check at runtime
```suggestion
    def insertBefore(self, newChild: _N, refChild: Node | None) -> _N: ...
```

stdlib/xml/dom/minidom.pyi
```diff
@@ -95,74 +101,73 @@ class Node(xml.dom.Node):
...
-    def getInterface(self, feature): ...
-    def getUserData(self, key): ...
-    def setUserData(self, key, data, handler): ...
-    childNodes: Incomplete
+    def cloneNode(self, deep: bool) -> Self: ...
+    def isSupported(self, feature: str, version: str | None) -> bool: ...
+    def isSameNode(self, other: Node) -> bool: ...
+    def getInterface(self, feature: str) -> Self | None: ...
+    def getUserData(self, key: str) -> object: ...
+    def setUserData(self, key: str, data: object, handler: UserDataHandler) -> object | None: ...
```
hamdanal, line 114: Returning `object` would require every use of the return value to be guarded by `isinstance` checks. `Any` is preferred for return values.
```suggestion
    def getUserData(self, key: str) -> Any: ...
    def setUserData(self, key: str, data: object, handler: UserDataHandler) -> Any: ...
```

stdlib/xml/dom/minidom.pyi
```diff
@@ -13,15 +15,19 @@ def parse(
     file: str | SupportsRead[ReadableBuffer | str], parser: XMLReader | None = None, bufsize: int | None = None
 ) -> Document: ...
 def parseString(string: str | ReadableBuffer, parser: XMLReader | None = None) -> Document: ...
-def getDOMImplementation(features=None) -> DOMImplementation | None: ...
+def getDOMImplementation(features: str | Iterable[tuple[str, str | None]] | None = None) -> DOMImplementation | None: ...
 
 class Node(xml.dom.Node):
+    parentNode: Node | None
+    childNodes: NodeList
```
hamdanal, line 22: To silence pyright. Need to import `Incomplete` from `_typeshed`.
```suggestion
    childNodes: NodeList[Incomplete]
```

(same for the `NodeList` annotations below)

stdlib/xml/dom/minidom.pyi
```diff
@@ -172,77 +177,72 @@ class NamedNodeMap:
...
-    attributes: Incomplete
-    childNodes: Incomplete
-    firstChild: Incomplete
-    lastChild: Incomplete
-    def appendChild(self, node) -> NoReturn: ...
-    def hasChildNodes(self) -> bool: ...
-    def insertBefore(self, newChild, refChild) -> NoReturn: ...
-    def removeChild(self, oldChild) -> NoReturn: ...
+    attributes: None
+    childNodes: NodeList
```
hamdanal, line 221: This is an `EmptyNodeList`. If the type checker complains you can add a `type: ignore` comment
```suggestion
    childNodes: EmptyNodeList
```

stdlib/xml/dom/minidom.pyi
```diff
@@ -254,11 +254,10 @@ class CharacterData(Childless, Node):
...
     nodeType: int
     nodeName: str
-    attributes: Incomplete
-    data: Incomplete
-    def splitText(self, offset): ...
+
+    def splitText(self, offset: int) -> Self: ...
     def writexml(self, writer: SupportsWrite[str], indent: str = "", addindent: str = "", newl: str = "") -> None: ...
-    def replaceWholeText(self, content): ...
+    def replaceWholeText(self, content: str) -> Self: ...
```
hamdanal, line 260: This returns `None` if `content` is falsy.

stdlib/xml/dom/minidom.pyi
```diff
@@ -276,56 +275,55 @@ class CDATASection(Text):
...
     def __len__(self) -> int: ...
-    def getNamedItem(self, name): ...
-    def getNamedItemNS(self, namespaceURI: str, localName): ...
-    def __getitem__(self, name_or_tuple): ...
-    def item(self, index): ...
-    def removeNamedItem(self, name) -> None: ...
-    def removeNamedItemNS(self, namespaceURI: str, localName) -> None: ...
-    def setNamedItem(self, node) -> None: ...
-    def setNamedItemNS(self, node) -> None: ...
+    def getNamedItem(self, name: str) -> Node: ...
```
hamdanal, line 280: This returns `None` if the node is not found in the sequence
```suggestion
    def getNamedItem(self, name: str) -> Node | None: ...
```
(same below)

stdlib/xml/dom/minidom.pyi
```diff
@@ -276,56 +275,55 @@ class CDATASection(Text):
...
+    def getNamedItemNS(self, namespaceURI: str, localName: str) -> Node: ...
+    def __getitem__(self, name_or_tuple: tuple[str, str] | str) -> Node: ...
+    def item(self, index: int) -> Node: ...
```
hamdanal, line 283: This explicitly returns `None`. It is similar in behavior to `dict.get()`
```suggestion
    def item(self, index: int) -> Node | None: ...
```

hamdanal (commented):
_Note I am not a maintainer of typeshed, just a regular contributor_

This is not a full review, just a few things I noticed after a quick spot check

srittau:
Thanks for contributing! I'm closing this PR for now, because it still fails some tests and has unresolved review feedback after three months of inactivity. If you are still interested, please feel free to open a new PR (or ping us to reopen this one).

JelleZijlstra:
Also thanks to @hamdanal for the useful review!

PR #10881, opened 2023-10-13
Allow specifying python version in runtests script

hamdanal:
This makes the script usable for packages that do not support the minimum python version hardcoded in the script.

Another part of #10722


srittau (approved):
Thanks!

PR #10882, opened 2023-10-13
Typed openpyxl colors descriptors

Avasam:
Typed openpyxl's Color related descriptors.


srittau (approved):
Thanks! Looks reasonable.

PR #10883, opened 2023-10-13
openpyxl: type `__iter__` dunders

Avasam:
Typed all `__iter__` dunders in openpyxl


stubs/openpyxl/openpyxl/worksheet/dimensions.pyi
```diff
@@ -42,7 +42,7 @@ class Dimension(Strict, StyleableObject):
         visible: Unused = True,
         style: Incomplete | None = None,
     ) -> None: ...
-    def __iter__(self) -> Iterator[tuple[str, str]]: ...
+    def __iter__(self) -> Generator[tuple[str, str], None, None]: ...
```
AlexWaygood, line 45: In general, I prefer to annotate `__iter__` methods as returning `Iterator` rather than returning `Generator`. While it's true that this one, and many of the other methods you're annotating in this PR, return a generator _now_, that's almost certainly an implementation detail that the author of the runtime package doesn't consider part of the public API. I'm sure they wouldn't blink twice before switching to an implementation where the `__iter__` method returned some other kind of iterator (if there was a reason to do so), and at that point, our stubs would be incorrect :)

Put another way: are people really meant to be `.send`ing or `.throw`ing into the object returned from this `__iter__` method? Or can we just go with the simpler annotation here?

AlexWaygood (commented):
Thanks! These annotations all look correct, but I _would_ prefer to generally return `Iterator` rather than `Generator` from `__iter__` methods, so that we don't start representing implementation details in the stubs as though they're public API: see below

stubs/openpyxl/openpyxl/worksheet/dimensions.pyi
Avasam: I see. I can update the __iter__ annotations to use iterators. No problem.

Avasam: FWIW, could/should this be a "best practices" rule for Flake8-PYI whenever we see `def __iter__(self) -> Generator[X, Any | None, None]` ?

AlexWaygood: Probably, yeah!

AlexWaygood (approved):
Thanks!

PR #10890, opened 2023-10-15
Add `sys.monitoring` from Python 3.12

lancelote:
Type annotations for `sys.monitoring` introduced by https://github.com/python/cpython/issues/103082 ([PEP 669](https://peps.python.org/pep-0669/)) in Python 3.12.

I am not entirely sure my approach is correct as `sys.monitoring` is a namespace. I refactored `sys` module into a package, `_monitoring.pyi` is then imported inside `sys/__init__.py`. Will appreciate any feedback on it 🙇


AlexWaygood:
Thanks! I think this is probably the right approach. We already do something similar with `os.path`: `os` isn't actually a package at runtime, but is in typeshed

AlexWaygood:
Ugh. This whole "`sys.monitoring` is a module, but not one you can import directly, `sys` isn't actually a package" thing is a pain. It looks like stubtest isn't checking any of the attributes on `sys.monitoring` in this PR, likely because of the fact that `sys.monitoring` _is_ actually an instance of `types.ModuleType`:

```pycon
>>> import sys
>>> sys.monitoring
<module 'sys.monitoring'>
>>> import types
>>> type(sys.monitoring) is types.ModuleType
True
```

Stubtest special-cases `ModuleType` instances in the global namespace specifically here, to avoid checking modules twice, which I _think_ is causing it to just avoid checking anything to do with `sys.monitoring`: https://github.com/python/mypy/blob/ff9deb3001d9c7cc84a1e2fed9125bf456b1d68b/mypy/stubtest.py#L358-L359

We may have to add some special-casing for `sys.monitoring` specifically to stubtest. For now, I'll just give this a manual review, though :)

stdlib/sys/_monitoring.pyi
```diff
@@ -0,0 +1,44 @@
...
+from typing import Any
+
+DEBUGGER_ID: int
+COVERAGE_ID: int
+PROFILER_ID: int
+OPTIMIZER_ID: int
+
+def use_tool_id(id: int, name: str) -> None: ...
+def free_tool_id(id: int) -> None: ...
+def get_tool(id: int) -> str | None: ...
```
AlexWaygood, line 12:
```pycon
>>> sys.monitoring.use_tool_id(id=1, name="foo")
TypeError: sys.monitoring.use_tool_id() takes no keyword arguments
>>> sys.monitoring.free_tool_id(id=42)
TypeError: sys.monitoring.free_tool_id() takes no keyword arguments
>>> sys.monitoring.get_tool(id=42)
TypeError: sys.monitoring.get_tool() takes no keyword arguments
```

```suggestion
def use_tool_id(__id: int, __name: str) -> None: ...
def free_tool_id(__id: int) -> None: ...
def get_tool(__id: int) -> str | None: ...
```

stdlib/sys/_monitoring.pyi
```diff
@@ -0,0 +1,44 @@
...
+
+class events(SimpleNamespace):
```
AlexWaygood, line 14: At runtime, this doesn't actually inherit from `SimpleNamespace`; it's actually an _instance_ of `SimpleNamespace`. There's no good way to annotate instances of `SimpleNamespace` currently (and probably never will be). I think we should do something like what we already do with `sys.implementation`, which is a very similar situation:

https://github.com/python/typeshed/blob/df05ff2f52a1c0aaff1196661c251875da0c9eb2/stdlib/sys.pyi#L190-L196

```suggestion
events: _events

class _events:
```

stdlib/sys/_monitoring.pyi
```diff
@@ -0,0 +1,44 @@
...
+    PY_UNWIND: int
+    PY_YIELD: int
+    RAISE: int
+    RERAISE: int
+    STOP_ITERATION: int
+
+def get_events(tool_id: int) -> int: ...
+def set_events(tool_id: int, event_set: int) -> None: ...
+def get_local_events(tool_id: int, code: CodeType) -> int: ...
+def set_local_events(tool_id: int, code: CodeType, event_set: int) -> int: ...
```
AlexWaygood, line 37:
```suggestion
def get_events(__tool_id: int) -> int: ...
def set_events(__tool_id: int, __event_set: int) -> None: ...
def get_local_events(__tool_id: int, __code: CodeType) -> int: ...
def set_local_events(__tool_id: int, __code: CodeType, __event_set: int) -> int: ...
```

stdlib/sys/_monitoring.pyi
```diff
@@ -0,0 +1,44 @@
...
+def restart_events() -> None: ...  # undocumented
+def _all_events() -> dict[str, int]: ...  # undocumented
+
+DISABLE: object
+MISSING: object
+
+def register_callback(tool_id: int, event: int, func: Callable[..., Any] | None) -> Callable[..., Any] | None: ...
```
AlexWaygood, line 44:
```suggestion
def register_callback(__tool_id: int, __event: int, __func: Callable[..., Any] | None) -> Callable[..., Any] | None: ...
```

stdlib/sys/_monitoring.pyi
```diff
@@ -0,0 +1,46 @@
...
+    RAISE: int
+    RERAISE: int
+    STOP_ITERATION: int
+
+def get_events(__tool_id: int) -> int: ...
+def set_events(__tool_id: int, __event_set: int) -> None: ...
+def get_local_events(__tool_id: int, __code: CodeType) -> int: ...
+def set_local_events(__tool_id: int, __code: CodeType, __event_set: int) -> int: ...
+def restart_events() -> None: ...  # undocumented
+def _all_events() -> dict[str, int]: ...  # undocumented
```
AlexWaygood, line 41: Do you need this one in your own code? We generally try to avoid stubbing private, undocumented functions unless there's a good reason to include them

```suggestion
```

stdlib/sys/_monitoring.pyi
```diff
@@ -0,0 +1,46 @@
...
+from typing import Any
+
+DEBUGGER_ID: int
+COVERAGE_ID: int
+PROFILER_ID: int
+OPTIMIZER_ID: int
+
+def use_tool_id(__id: int, __name: str) -> None: ...
+def free_tool_id(__id: int) -> None: ...
+def get_tool(__id: int) -> str | None: ...
```
AlexWaygood, line 12: Looks like the argument names are slightly different at runtime:

```suggestion
def use_tool_id(__tool_id: int, __name: str) -> None: ...
def free_tool_id(__tool_id: int) -> None: ...
def get_tool(__tool_id: int) -> str | None: ...
```

stdlib/sys/_monitoring.pyi
lancelote: No, I don't use it.

AlexWaygood (approved):
Thanks! I'll push a change to fixup a few comments, and then merge :D

lancelote:
@AlexWaygood Thank you for your time reviewing the PR 🙇

nedbat:
I would like to use these stubs, but they are not in 1.7.1.  When might they be available in a mypy release, or how can I use the stub files locally in my project?

AlexWaygood:
Hi @nedbat! These stubs have already been added to the mypy `master` branch: https://github.com/python/mypy/blob/master/mypy/typeshed/stdlib/sys/_monitoring.pyi. They should be included in the next minor version of mypy (mypy 1.8). Unfortunately I don't think there's a timeline for when mypy 1.8 will be released, and I doubt it will be released in the next few weeks, since mypy 1.7.1 was only released a few days ago. I would assume mypy 1.8 will be released in the next 1-2 months, however.

In the meantime, if you want to use the stub files locally in your project, you can clone typeshed, and use mypy's `--custom-typeshed-dir` command-line setting to point mypy to your local typeshed clone.

nedbat:
> Unfortunately I don't think there's a timeline for when mypy 1.8 will be released, and I doubt it will be released in the next few weeks, since mypy 1.7.1 was only released a few days ago.

Is a 1.7.2 unheard-of? :hope:

Locally cloning typeshed is a possibility, but I would need to do that for CI also, and it starts to get involved...

AlexWaygood:
> Is a 1.7.2 unheard-of? :hope:

You could try asking at https://github.com/python/mypy/issues/16341 if you could have a mypy patch release cherry-picking the stubs for `sys.monitoring`. But the mypy release process is unfortunately complex from what I understand, so the answer might be "no" (or "no response") :-(

Anyway, that's not really something we have control over here at typeshed -- you'd have to make the request over at mypy :-)

AlexWaygood:
> Is a 1.7.2 unheard-of? :hope:

Tagging @hauntsaninja, though, as a mypy/typeshed maintainer who has cut mypy patch releases in the past :-)

JelleZijlstra:
We also don't generally update typeshed in mypy bugfix releases: those releases are meant to fix regressions and crashes, not to introduce new updates or fix already-existing bugs.

nedbat:
Maybe this deserves a completely different thread, but: what is the policy about updating stdlib stubs? The [pull request that added sys.monitoring](https://github.com/python/cpython/pull/103083) was merged in April. Should we expect updated stubs as part of stdlib work?

AlexWaygood:
> Maybe this deserves a completely different thread, but: what is the policy about updating stdlib stubs? The [pull request that added sys.monitoring](https://github.com/python/cpython/pull/103083) was merged in April. Should we expect updated stubs as part of stdlib work?

In general, we start working on reflecting additions to the stdlib as soon as the beta period starts; most new stdlib modules are generally added in typeshed well before they're available at runtime in a non-beta release.

For this module in particular, we didn't notice it was missing until after 3.12.0 was released, because our tests didn't pick up that it was missing. Our tests missed that it was missing because of the unique way that `sys.monitoring` is implemented, as I discussed in https://github.com/python/typeshed/pull/10890#issuecomment-1763522569.

AlexWaygood:
See also:

- https://github.com/python/mypy/pull/16269

hauntsaninja:
My total guess is it would be a few months before 1.8, so if someone cherry picked just this change to the mypy 1.7 branch, I'd be willing to cut a 1.7.2.

AlexWaygood:
> My total guess is it would be a few months before 1.8, so if someone cherry picked just this change to the mypy 1.7 branch, I'd be willing to cut a 1.7.2.

- https://github.com/python/mypy/pull/16589

AlexWaygood:
@nedbat, mypy 1.8 has now been released, which includes these stubs :)

PR #10898, opened 2023-10-16
Type `pywin32`'s `win32.win32print.StartDocPrinter` method

### LINKED ISSUE #10891: pywin32.win32.win32print.StartDocPrinter wrong positional arguments order
Greetings!

I would like to report a simple issue:

The order of positional arguments to [pywin32.win32.win32print.StartDocPrinter](https://github.com/python/typeshed/blob/main/stubs/pywin32/win32/win32print.pyi#L16) should be `(hprinter, level, _tuple)` as indicated in [pywin32 docs](https://mhammond.github.io/pywin32/win32print__StartDocPrinter_meth.html) instead of `(hprinter, _tuple, level)`.

This is causing false positives during the type checking.

Avasam:
Closes #10891

Names and types come from https://mhammond.github.io/pywin32/win32print__StartDocPrinter_meth.html, https://github.com/mhammond/pywin32/blob/main/win32/src/win32print/win32print.cpp#L763 and local testing


stubs/pywin32/win32/win32print.pyi
```diff
@@ -1,4 +1,6 @@
 from _typeshed import Incomplete
+from collections.abc import Sequence
+from typing import Literal
```
Avasam, line 3:
```suggestion
from typing_extensions import Literal
```

stubs/pywin32/win32/win32print.pyi
```diff
@@ -13,7 +15,7 @@ def GetDefaultPrinter() -> str: ...
 def GetDefaultPrinterW() -> str: ...
 def SetDefaultPrinter(printer: str): ...
 def SetDefaultPrinterW(Printer: str): ...
-def StartDocPrinter(hprinter: _win32typing.PyPrinterHANDLE, _tuple, level: int = ...): ...
+def StartDocPrinter(__hprinter: _win32typing.PyPrinterHANDLE | int, __level: Literal[1], __tuple: Sequence[str]): ...
```
JelleZijlstra, line 18:
```suggestion
def StartDocPrinter(__hprinter: _win32typing.PyPrinterHANDLE | int, __level: Literal[1], __tuple: tuple[str, str, str]): ...
```
It's documented as a tuple, though from looking at the CPython code you are probably right that other sequences would work at runtime.

Avasam: I wouldn't rely too heavily on the documentation pages https://github.com/mhammond/pywin32#docs
> The docs are a long and sad story, but there's now an online version of the helpfile that ships with the installers. Lots of that is very old, but some is auto-generated and current. Would love help untangling the docs!

That said, the validation error message for the 3rd param mentions a "sequence" and I did validate that non-list non-tuple sequences work (like a string `"123"`)

JelleZijlstra: I think it's better to use `tuple[str, str, str]` here. The C argument parsing used accepts other objects recognized as sequences in C (which isn't the same concept as `Sequence` in Python), but tuples make the most sense for heterogeneous, fixed-length sequences. It's similar to how we type parameters as `bool` even if the C code will accept anything that is convertible to a bool.

JelleZijlstra (approved):

PR #10908, opened 2023-10-18
gh-10906: make Any a class on 3.11+

### LINKED ISSUE #10906: typing.Any should be a class on python 3.11+
in 3.11 Any was changed from a _SpecialForm object into a class: https://github.com/python/cpython/commit/5a4973e29f2f5c4ee8c086f40325786c62381540#diff-ddb987fca5f5df0c9a2f5521ed687919d70bb3d64eaeb8021f98833a2a716887R442

graingert:
Fixes #10906


stdlib/typing.pyi
```diff
@@ -135,7 +135,12 @@ AsyncContextManager = AbstractAsyncContextManager
 # This itself is only available during type checking
 def type_check_only(func_or_cls: _F) -> _F: ...
 
-Any = object()
+if sys.version_info >= (3, 11):
+    class Any:
```
Avasam, line 139: This ought to fix your stubtest metaclass issue:
```suggestion
    class _AnyMeta(type): ...
    class Any(metaclass=_AnyMeta):
```

Avasam:
The last stubtest issue would be resolved by adding `Any.__new__`'s definition. However I'm not entirely certain how to type it. I'm guessing `def __new__(cls, *args: object, **kargs: object) -> Self: ...`. But I'm curious if we can somehow represent that `TypeError` statically.

Anyway mypy seems happy from the mypy_primer results. "new" ibis issues are mostly rewords of a symbol name conflict.
pyright however, really isn't happy about this change, good luck with that! ^^"

srittau:
The primer hits are interesting. This seems to fix some type annotations where `Any` is used as a type, like in hydra_zen:

https://github.com/mit-ll-responsible-ai/hydra-zen/blob/main/src/hydra_zen/structured_configs/_utils.py

```python
def sanitized_type(...) -> type:
    ...
    return Any
    ...

```

Or in pydantic (https://github.com/pydantic/pydantic/blob/main/pydantic/_internal/_generate_schema.py):

```python
def _generate_parameter_schema(self, ..., annotation: type[Any], ...): ...

def _callable_schema(...):
    ...
    if ...:
        annotation = Any
    else:
        annotation = type_hints[name]
    ....
    self._generate_parameter_schema(..., annotation, ...)
```

JelleZijlstra:
The `type` return annotation in hydra-zen is simply incorrect though: they also have a branch that returns a Union (https://github.com/mit-ll-responsible-ai/hydra-zen/blob/cfa9f37175eeab39f2a1f52754c58391947c0ccb/src/hydra_zen/structured_configs/_utils.py#L337), which isn't a `type`.

hauntsaninja (changes requested):
I don't think we should do this. Any is still conceptually very much a special form.

In the past, we've been quite happy to not try to reflect implementation when it comes to special objects in typing.py and I think that has been a good policy (here's a mypy issue from yesterday as a reminder of why this is fraught https://github.com/python/mypy/issues/16297 ). If we do want to make a change like this, we should clarify exactly what behaviour we expect from type checkers, when targeting which versions of Python. If we're unable to decide, status quo should win since it is significantly easier to reason about.

PR #10909, opened 2023-10-18
Check  `*_pb2.pyi` files again

Avasam:
I was looking at Ruff vs Flake8 differences in typeshed and noticed this.

Start checking `*_pb2.pyi` files again as the issue seems to be resolved.
Fixes all `NQA102 "# noqa: F821" has no matching violations` that were ignored (we don't have a generation script for `stubs/s2clientprotocol` so that was done manually)


AlexWaygood:
All these `*_pb2.pyi` files are autogenerated with mypy-protobuf -- did you use the script [here](https://github.com/python/typeshed/blob/main/scripts/generate_proto_stubs.sh) to regenerate them, or did you edit them manually? :)

If you used the latest version of mypy-protobuf to regenerate these files, we should probably update the pinned version here, so these changes aren't overridden the next time the files are regenerated:

https://github.com/python/typeshed/blob/608eca5234d92ef62a308365fed49b14aa4db748/scripts/generate_proto_stubs.sh#L16

Avasam:
The script ends at the followign in git bash:
```bash
+ python3 -m venv venv
+ source venv/bin/activate
./scripts/generate_proto_stubs.sh: line 47: venv/bin/activate: No such file or directory
```
And immediately fails with the following in WSL:
```sh
/usr/bin/env: ‘bash\r’: No such file or directory
```

So I can't really validate locally

AlexWaygood:
(Yeah, I've never run the script locally either)

Avasam:
This should be better. I could auto-remove redundant F821 with Ruff once https://github.com/astral-sh/ruff/issues/3011 is fixed

.flake8
```diff
@@ -37,9 +37,8 @@ per-file-ignores =
   #     F811 redefinition of unused '...'
   stdlib/typing.pyi: B, E701, E741, F401, F403, F405, F811
   # Generated protobuf files include docstrings
-  # *_pb2.pyi: B, E701, E741, F401, F403, F405, Y021, Y026, Y053, Y054
+  # NQA102: Flake8-PYI monkey-patches F821 in a way it no longer triggers on class def
```
AlexWaygood, line 40: We should add a comment here saying that NQA102 should be removed from this list when we upgrade to `mypy-protobuf==3.5.0` and regenerate the stubs.

Although, since `mypy-protobuf==3.5.0` is already out, maybe we can just ask one of the maintainers on a Unix machine to regenerate the mypy-protobuf stubs for us using `mypy-protobuf==3.5.0`? Then we don't have to have any TODO comments anymore...

@JelleZijlstra / @srittau / @hauntsaninja? :)

I believe all that's required is to bump the pinned version here and then re-run the script:

https://github.com/python/typeshed/blob/608eca5234d92ef62a308365fed49b14aa4db748/scripts/generate_proto_stubs.sh#L16

JelleZijlstra: on it

JelleZijlstra: I ran it with 3.5.0 and didn't get any changes in the generated files relative to master. Are you sure mypy-protobuf 3.5.0 fixes the lint issue? I don't see anything in the above discussion that suggests that's the case.

JelleZijlstra: Also, the script didn't quite work for me on both macOS and Linux, probably due to minor differences in how commands like `sed` work. It's hard to write portable shell scripts.

AlexWaygood: > I ran it with 3.5.0 and didn't get any changes in the generated files relative to master. Are you sure mypy-protobuf 3.5.0 fixes the lint issue?

Huh. I'm not 100% sure. I assumed it was fixed in 3.5.0 due to https://github.com/nipunn1313/mypy-protobuf/pull/525#issuecomment-1656478258, which I think predates the 3.5.0 release

Avasam: Whether it does fix it or not, do we want to bump anyway? Although it might be good to be able to run the script first ^^"

JelleZijlstra: I can submit a PR with a version that works for me on Linux, but not sure it will be any more portable than what's currently there.

JelleZijlstra: Wait I was wrong, I forgot to actually bump the version before I started running it on Linux. In fact the new version does remove the noqa comments.

Avasam:
While we're at it:
Our Black configs "force-exclude" `.*_pb2.pyi` files with the following comment:
> # Exclude protobuf files because they have long line lengths
> # Ideally, we could configure Black to allow longer line lengths
> # for just these files, but doesn't seem possible yet.

But isn't that exactly what Black is for?
Did it use to make Black crash?
Is it because they're generated, so we don't want to bother formatting line-length? But then, shouldn't we not care about formatting them at all? Why is line-length specifically of concern here?

The protobuf stubs generation script even tries to run black (which currently ends up doing nothing), so I'm a bit confused.

.flake8
Avasam: With #10914 completed, I'm marking this conversation as resolved.

JelleZijlstra (approved):

AlexWaygood:
Thanks!

PR #10910, opened 2023-10-18
Use Ruff for `from __future__ import annotations` checks

Avasam:
Suggestion to use Ruff for faster and smarter `from __future__ import annotations` checks


.github/workflows/tests.yml
```diff
@@ -45,8 +45,17 @@ jobs:
...
       - run: ./tests/check_new_syntax.py
 
+  ruff:
+    name: Lint with Ruff
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: chartboost/ruff-action@v1
+        with:
+          version: "0.0.292" # must match .pre-commit-config.yaml and requirements-test.txt
```
srittau, line 55: Well, the version doesn't match. :)

Avasam: Thanks, I got distracted and forgot to check for other locations after merging

srittau (approved):
Thanks!

PR #10912, opened 2023-10-18
Replacing isort with Ruff

### LINKED ISSUE #10911: Replace isort with ruff?
I think @AlexWaygood suggested this originally and I pushed back against it. I've come around to the idea to using ruff for sorting imports, iff ruff has feature parity with and same sort order as isort (for our use case).

Avasam:
This PR explores using Ruff for linting and sorting imports.
This should lead to less configurations, one less dependency and faster imports auto-formatting
Ruff respects isort's action comments like `isort:skip_file`

There's some overlap with #10910 since both need to initially add the Ruff check in github actions.
Closes #10911


Avasam:
Waiting for #10915 and #10914 to land before fixing conflicts

pyproject.toml
```diff
@@ -7,15 +7,12 @@ skip_magic_trailing_comma = true
...
 
-[tool.isort]
-profile = "black"
-combine_as_imports = true
-line_length = 130
-skip = [".git", ".github", ".venv"]
-extra_standard_library = [
-    "typing_extensions",
+[tool.ruff.isort]
+combine-as-imports = true
```
intgr, line 11: I recommend also `split-on-trailing-comma = false`, for closer isort-like behavior.

Otherwise Ruff doesn't collapse multiline imports (even if they do fit on one line).

* See https://github.com/astral-sh/ruff/issues/6471

scripts/create_baseline_stubs.py
```diff
@@ -65,14 +65,9 @@ def run_black(stub_dir: str) -> None:
...
 
-def run_isort(stub_dir: str) -> None:
-    print(f"Running isort: isort {stub_dir}")
-    subprocess.run([sys.executable, "-m", "isort", stub_dir])
-
-
 def run_ruff(stub_dir: str) -> None:
     print(f"Running Ruff: ruff {stub_dir}")
-    subprocess.run([sys.executable, "-m", "ruff", stub_dir])
+    subprocess.run([sys.executable, "-m", "ruff", stub_dir, "--fix-only"])
```
Akuli, line 70: What does this do?

scripts/generate_proto_stubs.sh
```diff
@@ -73,7 +73,7 @@ PROTO_FILES=$(grep "GenProto.*google" $PYTHON_PROTOBUF_DIR/python/setup.py | \
 # shellcheck disable=SC2086
 protoc_install/bin/protoc --proto_path="$PYTHON_PROTOBUF_DIR/src" --mypy_out="relax_strict_optional_primitives:$REPO_ROOT/stubs/protobuf" $PROTO_FILES
 
-isort "$REPO_ROOT/stubs/protobuf"
+ruff "$REPO_ROOT/stubs/protobuf" --exit-non-zero-on-fix --fix-only
```
Akuli, line 76: I think `--exit-non-zero-on-fix` is backwards. If fixing happens, we want ruff to exit with success so that the script continues. It uses `set -e`, so nonzero exit code would stop the script here.

scripts/create_baseline_stubs.py
JelleZijlstra: Fix errors it can fix, and don't report on things it can't.

Ruff now recommends explicitly adding `check`:
```suggestion
    subprocess.run([sys.executable, "-m", "ruff", "check", stub_dir, "--fix-only"])
```

scripts/generate_proto_stubs.sh
Avasam: oops, you're right.
I'm also wondering whether `--fix-only` is wanted here. If you're gonna get errors later, may as well report them now? As long as it doesn't block script execution.

scripts/create_baseline_stubs.py
Avasam: > Ruff now recommends explicitly adding check

Right because of `ruff format` and to be explicit. I'll update other code locations too.

JelleZijlstra (approved):

PR #10915, opened 2023-10-19
mypy-protobuf v3.5: Fix and run protobuf generation scripts

Avasam:
@JelleZijlstra Suggested update to your PR, this should fix everything in both protobuf generation scripts (except `PYTHON_PROTOBUF_VERSION` being `4.21.8`)
- Bumped protobuf to 3.5 in both generation scripts
- Fixed the `METADATA.toml` update
- Run formatters and update metadata.toml in tensorflow protobuff's script
- Fixed `sync_tensorflow_protobuf_stubs.sh` to work when running from any location
- Make it easier to rerun `sync_tensorflow_protobuf_stubs.sh` when it failed midway (the repo clone is extremely slow, even with the optimsation I added)
- order entries alphabetically
- Added `tensorflow/tsl/protobuf/*.proto` as it is a dependency of generated modules
- Removed `tensorflow/core/protobuf/coordination_service_pb2.pyi` from the suppressions as it is no longer a generated module


scripts/sync_tensorflow_protobuf_stubs.sh
```diff
@@ -3,63 +3,70 @@ set -euxo pipefail
...
 
 pip install mypy-protobuf=="$MYPY_PROTOBUF_VERSION"
 
-mkdir repository
+cd "$(dirname "$0")" > /dev/null
+cd ../stubs/tensorflow
+mkdir -p repository
 pushd repository &> /dev/null
-    git clone https://github.com/tensorflow/tensorflow.git
+    # If the script fails halfway, it's nice to be able to re-run it immediatly
```
JelleZijlstra, line 24:
```suggestion
    # If the script fails halfway, it's nice to be able to re-run it immediately
```

JelleZijlstra (commented):
Thanks!

PR #10919, opened 2023-10-20
[django-filter] Absorb django-filter-stubs into typeshed

### LINKED ISSUE #10918: Proposal: absorb `django-filter-stubs` into typeshed
The `django-filter-stubs` project has has signalled that they want to be merged into `typeshed`, project author @DavisRayM agreeing with it: https://github.com/DavisRayM/django-filter-stubs/issues/13

I'm interested in implementing this.

I propose we do this in a way that preserves original commits and authorship information. I can do some git magic to "transplant" upstream commits to this repo, in the correct directory structure, but skipping changes/commits that don't touch `.pyi` files. And then you can merge that without squashing.

But if you want to squash merge and aren't interested in preserving history, then let me know and I won't bother with that.

intgr:
As described in #10918, I have added stubs from `django-filter-stubs` project. Upstream issue: https://github.com/DavisRayM/django-filter-stubs/issues/13

**Please don't use squash merge** to merge this (unless there's explicit decision to reconsider #10918). I have preserved original commit authors. There really weren't many changes in stubs files.

* Closes #10918
* Closes https://github.com/DavisRayM/django-filter-stubs/issues/13


AlexWaygood:
pre-commit.ci autofix

intgr:
This should be more or less ready for initial merge. If we can get this merged, I can look into the stubtest allowlist issues separately. Unless you think I should fix them immediately here?

Some questions:

* The stubs depend on `django-stubs` and `djangorestframework-stubs`. If we continue with this merge, these need to be added to `stub_uploader` allowlist. Nikita Sobolev can vouch for the fact that these dependencies aren't evil :smiling_imp:
* `stubtest` gets exception `django.core.exceptions.ImproperlyConfigured` from poking around in some `django-filter` internals. I could just allowlist these.

  But if we could somehow set `DJANGO_SETTINGS_MODULE` env variable for the stubtest run (to some empty Python module), then we wouldn't need to suppress them. Is there any precedent to this?
* Upstream used `Any` for incomplete types. Should I convert all of these to `_typeshed.Incomplete` or not bother with this?

stubs/django-filter/django_filters/fields.pyi
```diff
@@ -0,0 +1,93 @@
+from typing import Any
+
+from django import forms
+
+from .conf import settings as settings
+from .constants import EMPTY_VALUES as EMPTY_VALUES
```
intgr, line 6: There are lots of re-exports that shouldn't be re-exported. Should I clean these up before initial merge, or in follow-up?

stubs/django-filter/django_filters/fields.pyi
```diff
@@ -0,0 +1,93 @@
...
+class DateTimeRangeField(RangeField):
+    widget: Any = ...
+    def __init__(self, *args: Any, **kwargs: Any) -> None: ...
+
+class IsoDateTimeRangeField(RangeField):
+    widget: Any = ...
+    def __init__(self, *args: Any, **kwargs: Any) -> None: ...
+
+class TimeRangeField(RangeField):
+    widget: Any = ...
```
intgr, line 35: Should I clean up attribute assignments?

```suggestion
    widget: Any
```

stubs/django-filter/django_filters/rest_framework/backends.pyi
```diff
@@ -0,0 +1,24 @@
+from typing import Any
+
+from django.utils.deprecation import RenameMethodsBase
+
+from .. import compat as compat, utils as utils
+from . import filters as filters, filterset as filterset
```
intgr, line 6: Should I convert relative imports to absolute? Or I could do that in follow-up as well.

intgr (commented):
Some further questions. Personally I'd prefer to get this merged first and then do further improvements in follow-up PRs.

stubs/django-filter/django_filters/rest_framework/backends.pyi
AlexWaygood: We're pretty inconsistent about this across typeshed at the moment, so either is fine. Ideally the files in the `stubs/django-filter/` directory would pick a style and stick to it, but even there, tbh I don't really care that much about it

AlexWaygood:
(Writing from my phone, I have limited access to a laptop with Internet until Tuesday)

In general, I think it's fine to tackle issues with these stubs in followup PRs, now you've got CI passing (thanks!). Most things in `stubs/peewee/` are still unannotated; we still have many allowlist entries for `types-redis` — `django-filter` will hardly be unique in its stubs not being quite up to the standard of, say, our stdlib stubs.

For your question (1) — I can't remember off the top of my head what the exact rules are, but unfortunately adding `django-stubs` and `djangorestframework-stubs` to the stub_uploader allowlist might not be sufficient to get CI passing here. I think the stub_uploader might also check that all external dependencies listed in the `requires` field are actually dependencies of the runtime `django-filter` package, as an additional security check. I might be misremembering, though.

For your question (2), I think we're reasonably open to adding special-casing for specific stubs packages if it means we can run stubtest on them in CI. If you look at `tests/stubtest_third_party.py`, we already have some special-casing hardcoded into the script so that we can run stubtest on `types-uwisgi` in CI

intgr:
> For your question (1) [...] unfortunately adding `django-stubs` and `djangorestframework-stubs` to the stub_uploader allowlist might not be sufficient to get CI passinghere. I think the stub_uploader might also check that all external dependencies listed in the requires field are actually dependencies of the runtime django-filter package

Yep, that's also mentioned in https://github.com/typeshed-internal/stub_uploader/issues/90#issue-1587077892

Ideally these wouldn't be hard dependencies of the pacakge, but would only be installed in CI when testing. But there seems to be no such option currently?

Users of `django` who want to use `types-django-filter` are likely to have Django's stubs installed anyway. And if they don't, then `types-django-filter` is just a little less useful, but still better than nothing.

There are two different stubs packages for Django: `django-stubs` and `django-types` and we shouldn't force that choice.

stubs/django-filter/django_filters/rest_framework/backends.pyi
Avasam: I *personally* just use whatever I copy-pasted from the source, or my editor auto-imported. Sometimes I prefer using relative imports on purpose when my editor (ie: VSCode with Pylance) can't pickup a newly added module to typeshed stubs.

As @AlexWaygood said, we're pretty inconsistent about it.

intgr:
I think anything else we can fix/improve afterwards, but the `stub_uploader` failure is a blocker: https://github.com/python/typeshed/pull/10919#issuecomment-1774068623

One idea would be to create a new `METADATA.toml` option e.g. `test_requires = [...]` that's only used in CI and not subject to the strict checks as `requires = [...]` is.

AlexWaygood:
> One idea would be to create a new `METADATA.toml` option e.g. `test_requires = [...]` that's only used in CI and not subject to the strict checks as `requires = [...]` is. What do you think?

I'd prefer to avoid doing that, if at all possible. In 99% of cases in typeshed, if an external package is needed to make our tests pass, it's also needed for the stubs to make sense when they're installed by a user. I worry that by adding this option to our `METADATA.toml` files, we'd encourage people to add "test-only" dependencies which would mean our stubs would start behaving very differently under the conditions of our tests than they actually do when installed by users. We want to avoid that -- we want our tests to represent as closely as possible how our stubs packages behave when they're installed by users.

I believe that if a stubs package from typeshed imports a type that mypy can't resolve (because a dependency is missing), the imported type will just get silently resolved as `Any` -- mypy won't complain about it! So the user will get no warning that the stubs they installed aren't working as intended.

For the specific case of `types-django-filter`, maybe it's extremely unlikely that people will have `types-django-filter` installed without having `django-stubs` or `django-types` installed. But... that still feels like a risk I'm not willing to take -- people do crazy things on the internet :)

I'm going to take a look now and see if there are any other ways of solving this problem.

stubs/django-filter/django_filters/fields.pyi
AlexWaygood: follow-up is fine imo

stubs/django-filter/django_filters/fields.pyi
AlexWaygood: Fine to do this in a follow-up imo

intgr:
A few months have gone past. Has anything changed in terms of introducing a dependency on `django-stubs`? Or maybe some fresh ideas to solve this blocker?

There are a few other Django-related or djangorestframework-related projects I might want to provide stubs for, but no point in undertaking that work until this blocker has been cleared. And maintaining a 3rd party stubs project outside of `typeshed` is a bigger commitment than I want to undertake.

srittau:
@intgr Sorry for the long process. I've added the Django stubs to the stub uploader allowlist. I see this as uncritical. There are some type errors now, though.

intgr:
Note that stub_uploader tests are giving the following error:

> E stub_uploader.metadata.InvalidRequires: Expected dependency django-stubs to be listed in django-filter's requires_dist or the sdist's *.egg-info/requires.txt

This blocker was described in https://github.com/typeshed-internal/stub_uploader/pull/111#issuecomment-1774072444 / https://github.com/typeshed-internal/stub_uploader/issues/90#issue-1587077892

Do you have an idea what to do about this error? I suspect django-filter upstream will not want to add django-stubs to its dependencies.

Ganji00:
@intgr maybe we can try try to make a PR with django-stubs as a dependency to django-filter?

JelleZijlstra:
Sorry for the inaction here. I feel we should be able to address this with changes in stub-uploader. (Maybe a rule that if you depend on X at runtime, the stubs may depends on X-stubs, for some trusted values of X?)

hamdanal:
> Sorry for the inaction here. I feel we should be able to address this with changes in stub-uploader. (Maybe a rule that if you depend on X at runtime, the stubs may depends on X-stubs, for some trusted values of X?)

This has been addressed in https://github.com/typeshed-internal/stub_uploader/pull/152. Perhaps someone can merge main to trigger ci to see if it works.

stubs/django-filter/METADATA.toml
```diff
@@ -0,0 +1,3 @@
+version = "23.3.*"
+upstream_repository = "https://github.com/carltongibson/django-filter"
+requires = ["django-stubs", "djangorestframework-stubs"]
```
hamdanal, line 3: You'd need to remove this to pass the stub_uploader CI. It doesn't look like it is used anywhere in the stubs, not sure why it was included.
```suggestion
requires = ["django-stubs"]
```

intgr: It was used in `__init__.pyi`: `from . import rest_framework as rest_framework`. But indeed this is totally unnecessary and is the only usage, removed.

By the way, thanks a lot for improving stub-uploader, and for letting me know about the changes!

intgr:
It's a mystery to me, why Pyright fails on this file (only with older Pythons)... I didn't even touch it :smile:

>  /home/runner/work/typeshed/typeshed/stubs/setuptools/**setuptools/compat/py310.pyi**:9:81 - error: Unnecessary "# pyright: ignore" rule: "reportMissingImports" (reportUnnecessaryTypeIgnoreComment)

stubs/django-filter/METADATA.toml
hamdanal: > It was used in `__init__.pyi`: `from . import rest_framework as rest_framework`. But indeed this is totally unnecessary and is the only usage, removed.

This looks like a locally defined module named `rest_framework`, not the djangorestframework package. I don't think it should be removed.

> By the way, thanks a lot for improving stub-uploader, and for letting me know about the changes!

Thank you and thanks for working on django stubs!

intgr: > This looks like a locally defined module named rest_framework

Ah right you are!

> I don't think it should be removed.

I don't think it does anything. It's re-exporting the module `django_filter.rest_framework` that is already available under the same module path? :smile:

hamdanal: > I don't think it does anything. It's re-exporting the module `django_filter.rest_framework` in `django_filter/__init__.py` that is already available under the exact same module path? 😄

If you remove it then this will not work (at least with pyright):
```python
import django_filter
django_filter.rest_framework  # <- this will be a type error and would not have autocompletion in VS Code
```

AlexWaygood:
> It's a mystery to me, why Pyright fails on this file (only with older Pythons)... I didn't even touch it 😄

https://github.com/python/typeshed/pull/13101#issuecomment-2498572819

AlexWaygood:
See also my suggested solution for the pyright issue: https://github.com/python/typeshed/pull/13101#issuecomment-2498615639

(But it should be done in a separate PR to this one)

hamdanal:
> See also my suggested solution for the pyright issue: [#13101 (comment)](https://github.com/python/typeshed/pull/13101#issuecomment-2498615639)
> 
> (But it should be done in a separate PR to this one)

#13280

intgr:
Nice, this now passes CI.

As a reminder, **Please don't use squash merge** to merge this (unless there's explicit decision to reconsider #10918). I have preserved original commit authors in the commits here.

stubs/django-filter/METADATA.toml
hamdanal: @intgr I still think the the removed `from . import rest_framework as rest_framework` must be restored. It is deliberately imported at runtime for the specific reason that would let pyright accept the code snippet I mentioned above, i.e. to avoid having to type `import django_fiters.rest_framework`.

See the comment here:
https://github.com/carltongibson/django-filter/blob/2494df96c6387a9fa411fcb00b696b15dfd9216b/django_filters/__init__.py#L7-L10

intgr: I think it's silly and nobody should use it like that. The documentation and tests don't use it that way either.

But if you feel strongly about this, fair enough.

stubs/django-filter/django_filters/__init__.pyi
```diff
@@ -0,0 +1,9 @@
+from typing import Any
+
+from . import rest_framework as rest_framework
+from .filters import *
+from .filterset import FilterSet as FilterSet
+
+def parse_version(version: Any): ...
```
JelleZijlstra, line 7: We now believe Any should be used only in places where the type system can't express the precise type. As a result, we'd want a comment with every Any explaining why it's there.

If this is Any merely because you haven't figured out the precise type, you can leave the parameter unannotated, or use `_typeshed.Incomplete` in cases where that's not possible (e.g., for the global below).

huynguyengl99:
Hi @intgr, it looks like this PR has been inactive for a while. Are you running into any issues that are blocking progress? I’d be happy to help or continue the work to get it merged — I can also fork or copy your branch if needed.

intgr:
Oh, I had forgotten about this, thanks for the reminder.

Looks like it just needs minor tweaks to solve JelleZijlstra's comment. I'll update it and get it ready for review in a few days.

huynguyengl99:
Hi @intgr! 👋

 I'm working on another PR for this package (#14540) to improve some missing types and add comments explaining Any usage, plus running stubtest with django-stubs.

 Would love to see if we can combine our PRs together and create the best possible type stubs for django-filter? Happy to collaborate and merge our work! 🤝

 Let me know what you think!

intgr:
@huynguyengl99 I had a look at your PR. Your type hints are a lot more thorough already.

So let's get your #14540 merged first and we will be in a better state. Then I will go over the diff & see if there is anything left to improve.

intgr:
* Closing, deferring to https://github.com/python/typeshed/pull/14540

huynguyengl99:
Thanks a lot @intgr ! 😊 Looking forward to getting #14540 merged.

PR #10920, opened 2023-10-22
Fix return type of `inspect.getblock`

Viicos:
I don't think there's any case where it returns something else than `list[str]`


JelleZijlstra:
It returns a slice from the input sequence: https://github.com/python/cpython/blob/88bac5d5044e577825db1f9367af908dc9a3ad82/Lib/inspect.py#L1271. Slicing a non-list sequence (such as a tuple) generally gives you a non-list sequence, so I think the current annotation is correct.

Akuli:
Maybe we could distinguish sliceable and non-sliceable sequences? I'm not sure if it's worth the complexity it would introduce.

```
>>> import collections
>>> d = collections.deque([1,2,3])
>>> d[1:]
TypeError: sequence index must be integer, not 'slice'
```

JelleZijlstra:
@Akuli that's a good observation: `deque` is a Sequence and our stubs for Sequence indicate support for slicing, but `deque` doesn't support slicing. We have some type ignores as a result. Possibly we should change something to fix this discrepancy. Any solution might be unduly disruptive though, and I don't see much relationship with this PR.

AlexWaygood:
I think `deque` is an example of "practicality beats purity": it's a sequence in every respect other than the fact it doesn't support slicing, so I think it makes sense to consider a sequence even though, strictly speaking, it actually isn't (according to the definitions we have in `typing.pyi`).

I agree with @JelleZijlstra that this isn't really relevant to this PR though. Possibly we could do something like this, relying on the fact that slicing a sequence `s` nearly always produces an object with the same type as `s`:

```py
from typing import TypeVar

_StrSequenceT = TypeVar("_StrSequenceT", bound=Sequence[str])

def getblock(lines: _StrSequenceT) -> _StrSequenceT: ...
```

Viicos:
> Slicing a non-list sequence (such as a tuple) generally gives you a non-list sequence, so I think the current annotation is correct.

Sorry for not taking a deeper look at it at first. I figured having `Sequence` as a return type isn't really common.

Using the `TypeVar` seems to be a good option, it would avoid having to cast to a specific type the return value.

JelleZijlstra:
This solution isn't strictly correct as Sequence doesn't guarantee that slicing returns the same type. Maybe that's fine, but I kind of prefer the existing simpler annotation.

AlexWaygood:
One more option: we could just special-case the two most common Sequence types:

```py
@overload
def getblock(lines: list[str]) -> list[str]: ...
@overload
def getblock(lines: tuple[str, ...]) -> tuple[str, ...]: ...
@overload
def getblock(lines: Sequence[str]) -> Sequence[str]: ...
```

Viicos:
> Actually, it looks like we can do the "principled" thing here with a protocol, and have the best of all worlds:

I had a thought about using a `Protocol` as well, I'll add a commit to see how it goes

Akuli:
A fancy protocol solution wouldn't be a very nice IDE experience. If you see `_IterableAndSliceable[_T]` in your IDE autocompletions, you won't immediately know what it's supposed to mean. IMO this is fine for complex functions, but seems like a sign of over-engineering for something as simple as this.

I still think the overload solution is the best way to go. It returns a good type in pretty much all practical situations, but it's also simple and readable.

AlexWaygood:
In the past I feel like we've generally prioritised type-checking over other users of typeshed, even if we've been trying to pay more attention to those other users recently. So I mildly prefer the solution using the protocol, since it's a general solution that feels more principled. I'm also not really sure it's typeshed's job to make sure that IDEs give user-friendly autocompletions.

Having said that, I agree that such a complex solution using protocols does feel like a bit like overkill here, and the solution using overloads certainly isn't _incorrect_. So I'm okay with either.

stdlib/inspect.pyi
```diff
@@ -294,6 +294,12 @@ _SourceObjectType: TypeAlias = (
 
 def findsource(object: _SourceObjectType) -> tuple[list[str], int]: ...
 def getabsfile(object: _SourceObjectType, _filename: str | None = None) -> str: ...
+@overload
```
AlexWaygood, line 297:
```suggestion
# Special-case the two most common input types here
# to avoid the annoyingly vague `Sequence[str]` return type
```

AlexWaygood (approved):
thanks!

PR #10926, opened 2023-10-25
Allow `str` for `tkinter.Scrollbar.set()`

### LINKED ISSUE #10071: tkinter scrollbar set accept also str not only float
Tkinter scrollbar set() accepts also str not only float:

    def set(self, first: float, last: float) -> None: ...

should be:

    def set(self, first: float|str, last: float|str) -> None: ...

Paulie-Aditya:
Fixes #10071


stdlib/tkinter/__init__.pyi
```diff
@@ -2888,7 +2888,7 @@ class Scrollbar(Widget):
     def fraction(self, x: int, y: int) -> float: ...
     def identify(self, x: int, y: int) -> Literal["arrow1", "arrow2", "slider", "trough1", "trough2", ""]: ...
     def get(self) -> tuple[float, float, float, float] | tuple[float, float]: ...
-    def set(self, first: float, last: float) -> None: ...
+    def set(self, float(first), float(last)) -> None: ...
```
AlexWaygood, line 2891: Hmm, this isn't valid Python syntax :)

Are you sure you fully understand the issue you're trying to fix?

Paulie-Aditya: yeah just a sec fixing it now

AlexWaygood: No, that wouldn't work. Functions in stub files shouldn't have any bodies.

If you're new to the idea of what a stub file is, please see https://peps.python.org/pep-0484/#stub-files or https://typing.readthedocs.io/en/latest/source/stubs.html#type-stubs for an explanation :)

Akuli (approved):
Thanks!

PR #10927, opened 2023-10-26
Finishes stubs for `webob.cookies` and improves `samesite` type

Daverball:
I noticed that the `samesite` parameter for cookies only accepted the documented lowercase spelling of the values, but it probably should also accept the official title case spelling from the specification. The implementation accepts either spelling too.

Given this opportunity I also finished the stubs for the cookies module.

### DIFF
--- a/stubs/WebOb/webob/cookies.pyi
+++ b/stubs/WebOb/webob/cookies.pyi
@@ -1,13 +1,24 @@
-from _typeshed import Incomplete
+from _typeshed import sentinel
 from _typeshed.wsgi import WSGIEnvironment
-from collections.abc import ItemsView, Iterator, KeysView, MutableMapping, ValuesView
+from collections.abc import Callable, Collection, ItemsView, Iterator, KeysView, MutableMapping, ValuesView
 from datetime import date, datetime, timedelta
-from typing import TypeVar, overload
-from typing_extensions import Literal
+from hashlib import _Hash
+from typing import Any, Protocol, TypeVar, overload
+from typing_extensions import Literal, TypeAlias
 
 from webob.descriptors import _AsymmetricProperty
+from webob.request import Request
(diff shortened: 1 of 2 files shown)


stubs/WebOb/webob/cookies.pyi
```diff
@@ -74,75 +85,98 @@ def make_cookie(
...
     comment: str | None = None,
-    samesite: Literal["strict", "lax", "none"] | None = None,
+    samesite: _SameSitePolicy | None = None,
 ) -> str: ...
 
 class JSONSerializer:
-    def dumps(self, appstruct): ...
-    def loads(self, bstruct): ...
+    def dumps(self, appstruct: Any) -> bytes: ...
+    def loads(self, bstruct: bytes) -> Any: ...
```
srittau, line 93: As this is sent through `text_()`:

```suggestion
    def loads(self, bstruct: bytes | str) -> Any: ...
```

stubs/WebOb/webob/cookies.pyi
```diff
@@ -74,75 +85,98 @@ def make_cookie(
...
-        max_age: Incomplete | None = None,
-        httponly: Incomplete | None = None,
-        samesite: Incomplete | None = None,
+        max_age: int | timedelta | None = None,
+        httponly: bool | None = None,
+        samesite: _SameSitePolicy | None = None,
         path: str = "/",
-        domains: Incomplete | None = None,
-        serializer: Incomplete | None = None,
+        # even thoug the docs claim any iterable is fine, that is
```
srittau, line 131: Nit:
```suggestion
        # even though the docs claim any iterable is fine, that is
```

srittau (changes requested):
Thanks, just one thing I noticed below.

srittau (approved):

PR #10928, opened 2023-10-26
Add stubs for `importlib.(resources.)readers`

AlexWaygood:
The annotations are incomplete. The module is somewhat poorly documented, and I wanted to defer anything too complex to future PRs, so that this was easier to review


AlexWaygood:
I'm not sure why https://github.com/python/typeshed/pull/10928/commits/8232196dbb82b5f104538f2e149d5eee40a1584e was needed. I thought both `mypy_test.py` and pyright had special handling for our `stdlib/VERSIONS` file, so that we didn't need to do things like this, but both were failing without that change. It's also confusing to me that stubtest passed, even though `mypy_test.py` didn't.

stdlib/importlib/readers.pyi
```diff
@@ -0,0 +1,60 @@
...
+        def resource_path(self, resource: StrPath) -> str: ...
+        def files(self) -> pathlib.Path: ...
+
+    class ZipReader(abc.TraversableResources):
+        prefix: str
+        archive: Incomplete
+        def __init__(self, loader, module: str) -> None: ...
+        def open_resource(self, resource: str) -> BufferedReader: ...
+        def is_resource(self, path: StrPath) -> bool: ...
+        def files(self): ...
```
AlexWaygood, line 38: This obviously returns a `zipfile.Path` at runtime. Unfortunately, if you add that annotation right now, mypy complains:

```
stdlib\importlib\readers.pyi:39: error: Return type "Path" of "files" incompatible with return type "Traversable" in supertype "TraversableResources"  [override]
```

This points to an error either in our zipfile stubs or in our `importlib.(resources.)abc` stubs. Type checkers should probably understand `zipfile.Path` as being a subtype of `Traversable`; I believe that's the intention of the authors of these modules.

Anyway, that looks non-trivial to fix, so I'd rather defer it to another PR for now.

stdlib/importlib/readers.pyi
```diff
@@ -0,0 +1,60 @@
...
+
+if sys.version_info >= (3, 11):
+    import importlib.resources.abc as abc
+else:
+    import importlib.abc as abc
+
+if sys.version_info >= (3, 10):
+    if sys.version_info >= (3, 11):
+        __all__ = ["FileReader", "ZipReader", "MultiplexedPath", "NamespaceReader"]
+    else:
```
srittau, line 21: Nit:

```suggestion
    if sys.version_info < (3, 11):
```

While `else` is of course correct here, a separate `if` statement makes it clearer that these are unrelated version differences.

stdlib/importlib/readers.pyi
```diff
@@ -0,0 +1,60 @@
...
+        prefix: str
+        archive: Incomplete
+        def __init__(self, loader, module: str) -> None: ...
+        def open_resource(self, resource: str) -> BufferedReader: ...
+        def is_resource(self, path: StrPath) -> bool: ...
+        def files(self): ...
+
+    class MultiplexedPath(abc.Traversable):
+        def __init__(self, *paths: abc.Traversable) -> None: ...
+        def iterdir(self) -> Iterator[Self | abc.Traversable]: ...
```
srittau, line 42:
```suggestion
        def iterdir(self) -> Iterator[abc.Traversable]: ...
```

In 3.10, this only ever returns a `Path` iterator, so maybe just using `Traversable` is the most compatible for now.

stdlib/importlib/readers.pyi
```diff
@@ -0,0 +1,60 @@
...
+        def read_bytes(self) -> NoReturn: ...
+        def read_text(self, *args: Never, **kwargs: Never) -> NoReturn: ...  # type: ignore[override]
+        def is_dir(self) -> Literal[True]: ...
+        def is_file(self) -> Literal[False]: ...
+        if sys.version_info >= (3, 12):
+            def joinpath(self, *descendants: str) -> abc.Traversable: ...
+        else:
+            def joinpath(self, child: str) -> abc.Traversable: ...  # type: ignore[override]
+            __truediv__ = joinpath
+        def open(self, *args: Never, **kwargs: Never) -> NoReturn: ...  # type: ignore[override]
```
srittau, line 52:
```suggestion
        def open(self, *args: Unused, **kwargs: Unused) -> NoReturn: ...  # type: ignore[override]
```

`Never` would disallow arguably allowed calls like `.open("r")` (even if those always raise an exception, of course).

srittau (changes requested):
Thanks! A few things I noticed below.

stdlib/importlib/readers.pyi
AlexWaygood: ah yes, `Self` is of course a subtype of `Traversable` anyhow!

stdlib/importlib/readers.pyi
AlexWaygood: Yes, that was a deliberate choice... I thought maybe it would be useful for type checkers to warn you if you provided arguments to this function. Since it always fails at runtime, it seems like you're probably making a mistake if you're passing arguments to it?

srittau: But that should already be caught by the `NoReturn`.While this is a far-fetched example (using a non-existent typing feature), this disallows things like this:

```python
trav_open: SignatureOf[Traversible.open] = MultiplexedPath.open
```

I wonder why no type checker complains about using `Never`, since this breaks `Traversible`'s contract. 

That said, I don't feel strongly about this.

AlexWaygood: > But that should already be caught by the `NoReturn`

Only if you have mypy's `--warn-unreachable` setting turned on (or the equivalent setting in other type checkers), and most people don't, unfortunately. Whereas annotating the arguments with `Never` means you'll get an error from the type checker even if you only have the default settings enabled.

I'll concede that this is a bit of a hack, and that we're really working around the lack of something like https://github.com/python/typing/issues/1043 -- but I think it's useful for the time being

PR #10931, opened 2023-10-26
Add stubs for `importlib.(resources.)simple`

AlexWaygood:
And do some minor cleanup of `importlib.(resources.)readers` following https://github.com/python/typeshed/commit/1c184fea33d50ba430410cf1b25d4aea2df2981e


stdlib/importlib/readers.pyi
```diff
@@ -4,6 +4,7 @@
```
JelleZijlstra, line 4: Not a matter for this PR, but at runtime (at least on main) this file re-exports code from `importlib.resources.readers`, so ideally we should do the same in our stub.

stdlib/importlib/resources/simple.pyi
```diff
@@ -0,0 +1,48 @@
...
+        def joinpath(self, name: Never) -> NoReturn: ...  # type: ignore[override]
+
+    class ResourceContainer(Traversable, metaclass=abc.ABCMeta):
+        reader: SimpleReader
+        def __init__(self, reader: SimpleReader) -> None: ...
+        def is_dir(self) -> Literal[True]: ...
+        def is_file(self) -> Literal[False]: ...
+        def iterdir(self) -> Iterator[ResourceHandle | ResourceContainer]: ...
+        def open(self, *args: Never, **kwargs: Never) -> NoReturn: ...  # type: ignore[override]
+        def joinpath(self, *descendants: str) -> Traversable: ...
```
JelleZijlstra, line 45: This doesn't exist at runtime on main.

stdlib/importlib/resources/simple.pyi
```diff
@@ -0,0 +1,48 @@
...
+
+    class TraversableReader(TraversableResources, SimpleReader, metaclass=abc.ABCMeta):
```
JelleZijlstra, line 47: Why do we need the metaclass? Doesn't seem to exist at runtime.

AlexWaygood: It inherits from a class that has abstract methods, but does not implement those abstract methods, meaning it is still abstract. Because it's really easy to do that accidentally in a stub, mypy has an undocumented feature that says you have to explicitly redeclare the metaclass if that's really what you want: https://github.com/python/mypy/blob/9011ca8b4dedc0e7177737b5265f69694afa91b5/mypy/semanal_classprop.py#L94

stdlib/importlib/readers.pyi
AlexWaygood: Yeah, see the comments on lines 1-3 of this file :)

stdlib/importlib/resources/simple.pyi
AlexWaygood: Thanks, fixed in https://github.com/python/typeshed/pull/10931/commits/d9f598aaac8436d4065e5a8cf447749f4c2762c9

AlexWaygood:
Aha, I see pre-commit is now doing the same thing to me here that it's doing to @hamdanal in #10721 🙃

See https://github.com/python/typeshed/pull/10939#pullrequestreview-1702764114 for explanation...

stdlib/importlib/resources/simple.pyi
JelleZijlstra: Oh yes, I implemented that feature :)

PR #10932, opened 2023-10-26
[fpdf2] add compatibility with fpdf2 2.7.6

mschoettle:
Release Notes: https://github.com/py-pdf/fpdf2/blob/master/CHANGELOG.md#276---2023-10-11

Changes `txt` to `text` and brings all the other stubs up to date. 

~~I am wondering whether it would be better to add `text` as an additional argument instead?~~

~~This PR currently only covers what's listed in the "Deprecated" section of the release notes. It does not cover the new method `text_columns` (https://py-pdf.github.io/fpdf2/TextColumns.html). Let me know if this should be added as well at the same time.~~

Closes #10609


mschoettle:
The pipeline answered the above question for me :)

AlexWaygood:
For entirely new modules which are missing from the stub, mypy's stubgen tool can be pretty useful for generating baseline stubs

mschoettle:
I completed bringing the stubs up to date with v2.7.6. I used mypy's stubgen tool and compared the two versions. There's one place where I replaced `Number` first with `int | float` and then `float` due to a violation. I left a TODO there.

stubs/fpdf2/fpdf/encryption.pyi
```diff
@@ -64,29 +64,43 @@ class StandardSecurityHandler:
     o: str
     k: str
     u: str
+    # The following fields are only defined after a call to generate_user_password_rev6().
+    ue: Incomplete
+    # The following fields are only defined after a call to generate_owner_password_rev6().
+    oe: Incomplete
+    # The following fields are only defined after a call to generate_perms_rev6().
+    perms_rev6: Incomplete
```
srittau, line 72:
```suggestion
    # The following field is only defined after a call to generate_user_password_rev6().
    ue: Incomplete
    # The following field is only defined after a call to generate_owner_password_rev6().
    oe: Incomplete
    # The following field is only defined after a call to generate_perms_rev6().
    perms_rev6: Incomplete
```

stubs/fpdf2/fpdf/encryption.pyi
```diff
@@ -64,29 +64,43 @@ class StandardSecurityHandler:
...
+        encryption_method: EncryptionMethod = ...,
         encrypt_metadata: bool = False,
     ) -> None: ...
-    def generate_passwords(self, file_id) -> None: ...
+    def generate_passwords(self, file_id: str) -> None: ...
     def get_encryption_obj(self) -> EncryptionDictionary: ...
-    def encrypt(self, text: str | bytes | bytearray, obj_id) -> bytes: ...
-    def encrypt_string(self, string, obj_id): ...
-    def encrypt_stream(self, stream, obj_id): ...
+    def encrypt(self, text: str | bytes | bytearray, obj_id: int) -> str | bytes: ...
```
srittau, line 85: This should use an overload, otherwise users have to use an `isinstance` check when providing either a `str` or a `bytes` (but not both):

```suggestion
    @overload
    def encrypt(self, text: bytes | bytearray, obj_id: int) -> bytes: ...
    @overload
    def encrypt(self, text: str, obj_id: int) -> str: ...
```

(`overload`  needs to be added to the imports.)

stubs/fpdf2/fpdf/fonts.pyi
```diff
@@ -24,5 +26,81 @@ class FontFace:
...
+    font_b_box: Incomplete
+    italic_angle: Incomplete
+    stem_v: Incomplete
+    missing_width: Incomplete
+    font_name: Incomplete
+    def __init__(self, ascent, descent, cap_height, flags, font_b_box, italic_angle, stem_v, missing_width) -> None: ...
+
+class Glyph:
+    glyph_id: int
+    unicode: tuple[Incomplete]
```
srittau, line 76: `tuple[Incomplete]` means "a tuple with one element".

```suggestion
    unicode: tuple[Incomplete, ...]
```

stubs/fpdf2/fpdf/line_break.pyi
```diff
@@ -57,12 +58,28 @@ class Fragment:
...
+        word_spacing: Incomplete,
+        adjust_x: Incomplete,
+        adjust_y: Incomplete,
+        h: Incomplete,
+    ): ...
+    def render_pdf_text_ttf(self, frag_ws: Incomplete, word_spacing: Incomplete): ...
+    def render_with_text_shaping(
+        self, pos_x: Incomplete, pos_y: Incomplete, h: Incomplete, word_spacing: Incomplete, text_shaping_parms: Incomplete
+    ): ...
+    def render_pdf_text_core(self, frag_ws: Incomplete, current_ws: Incomplete): ...
```
srittau, line 74: Nit: The `Incomplete` is redundant here:

```suggestion
    def render_pdf_text(
        self,
        frag_ws,
        current_ws,
        word_spacing,
        adjust_x,
        adjust_y,
        h,
    ): ...
    def render_pdf_text_ttf(self, frag_ws, word_spacing): ...
    def render_with_text_shaping(self, pos_x, pos_y, h, word_spacing, text_shaping_parms): ...
    def render_pdf_text_core(self, frag_ws, current_ws): ...
```

stubs/fpdf2/fpdf/util.pyi
```diff
@@ -1,13 +1,24 @@
...
-from typing import Any, AnyStr
+from typing import Any, AnyStr, NamedTuple
 from typing_extensions import Final, Literal, TypeAlias
 
 _Unit: TypeAlias = Literal["pt", "mm", "cm", "in"]
 
 PIL_MEM_BLOCK_SIZE_IN_MIB: Final = 16
 
+class Padding(NamedTuple):
+    # TODO: replaced Number with float
```
srittau, line 10:
```suggestion
    # TODO: replace Number with float
```

stubs/fpdf2/fpdf/util.pyi
```diff
@@ -1,13 +1,24 @@
...
+    top: float = ...
+    right: float = ...
+    bottom: float = ...
+    left: float = ...
```
srittau, line 14: Nit: `= ...` is redundant:

```suggestion
    top: float
    right: float
    bottom: float
    left: float
```

srittau (changes requested):
Thanks! A few things I noticed below.

stubs/fpdf2/fpdf/util.pyi
mschoettle: What I meant here is that `stubgen` created this with `Number` and I replaced it with `float` (`int | float` caused a violation that suggests to only use `float`). 

I can revert to `Number` or remove the TODO.

mschoettle: Removing the default `= ...` caused the following error:

```
error: fpdf.util.Padding.__new__ is inconsistent, runtime argument "top" has a default value but stub argument does not
Stub: in file /home/runner/work/typeshed/typeshed/stubs/fpdf2/fpdf/util.pyi:9
def [_NT <: tuple[builtins.float, builtins.float, builtins.float, builtins.float]] (_cls: Type[_NT`-1], top: builtins.float, right: builtins.float, bottom: builtins.float, left: builtins.float) -> _NT`-1
Runtime:
def (_cls, top: numbers.Number = 0, right: numbers.Number = 0, bottom: numbers.Number = 0, left: numbers.Number = 0)
```

So I switched back to `Number`.

mschoettle: Hm, I can't get it to work. What's the correct way to reflect the implementation?

I've tried:

```py
top: Number
top: Number = ...
top: Number = 0
```

srittau: @mschoettle The annotations in the upstream source code using `Number` are a bit weird. Let me experiment for a bit.

srittau (approved):
@mschoettle Just using `float` seems to work fine. I suggest we don't use `Number`.

mschoettle:
@srittau Thanks a lot for fixing the weird `Number` issue!

PR #10934, opened 2023-10-26
Simplify Traversable.open() signature

### LINKED ISSUE #10930: `zipfile.Path` is not understood by type checkers as a subtype of `importlib.abc.Traversable`
[The docs](https://docs.python.org/3/library/zipfile.html) state that it's meant to be a subtype of that protocol:

> Class that implements a subset of the interface provided by [pathlib.Path](https://docs.python.org/3/library/pathlib.html#pathlib.Path), including the full [importlib.resources.abc.Traversable](https://docs.python.org/3/library/importlib.resources.abc.html#importlib.resources.abc.Traversable) interface.

But type checkers don't see it that way: https://mypy-play.net/?mypy=latest&python=3.11&gist=fe6a59cc57d5a1f7fa081445800b083c

It's not immediately clear to me how to fix this.

srittau:
This is necessary so that implentors can reasonably implement this method. For example `zipfile.Path.open()` (which is considered a `Traversable`) only supports this subset.

Closes #10930


srittau:
Closes: #10930

stdlib/importlib/abc.pyi
```diff
@@ -140,66 +132,20 @@ if sys.version_info >= (3, 9):
...
-            errors: None = None,
-            newline: None = None,
-        ) -> BufferedReader: ...
-        # Buffering cannot be determined: fall back to BinaryIO
+            mode: Literal["r", "w"] = "r",
+            *,
+            encoding: str | None = None,
+            errors: str | None = None,
+            newline: str | None = None,
+        ) -> IO[str]: ...
```
JelleZijlstra, line 145: I wonder if we can make the return type a Protocol with a few common methods like `read`. To get the exact signature, could look at how importlib uses Traversable internally.

hauntsaninja (commented):
Thanks for working on this! Worth adding a test that a pathlib.Pathlib is an importlib.abc.Traversable?

stdlib/importlib/abc.pyi
srittau: This was my long term plan (to get rid of `IO`). But I don't want to look into this at this moment. Partly to keep this change small, partly because of laziness.

srittau:
After one or two short revisions, this is now working.

AlexWaygood (approved):
Nice, thank you!

PR #10935, opened 2023-10-26
Improve regression test documentation

srittau:
* Add instructions on how to run the tests to the `test-cases` directory.
* Add instruction on how to test stdlib to the `regr_test.py --help` output.

### DIFF
--- a/test_cases/README.md
+++ b/test_cases/README.md
@@ -32,6 +32,13 @@ stubs where decisions have been taken that might be slightly unusual. These
 test cases serve a different purpose: to check that type checkers do not emit
 false-positive errors for idiomatic usage of these classes.
 
+## Running the tests
+
+To run the tests in this directory, run `python tests/typecheck_typeshed.py stdlib`
+from the root of the typeshed repository. This assumes that the development
+environment has been set up as described in the [CONTRIBUTING.md](../CONTRIBUTING.md)
+document.
+
 ### How the tests work
 
 The code in this directory is not intended to be directly executed. Instead,
--- a/test_cases/README.md
+++ b/test_cases/README.md
@@ -3,7 +3,7 @@
 This directory contains code samples that act as a regression test for
 typeshed's stdlib stubs.
 
-**This directory should _only_ contain test cases for functions and classes which
+**This directory should *only* contain test cases for functions and classes which
 are known to have caused problems in the past, where the stubs are difficult to
 get right.** 100% test coverage for typeshed is neither necessary nor
 desirable, as it would lead to code duplication. Moreover, typeshed has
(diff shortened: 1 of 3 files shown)


test_cases/README.md
```diff
@@ -32,6 +32,13 @@ stubs where decisions have been taken that might be slightly unusual. These
 test cases serve a different purpose: to check that type checkers do not emit
 false-positive errors for idiomatic usage of these classes.
 
+## Running the tests
+
+To run the tests in this directory, run `python tests/typecheck_typeshed.py stdlib`
```
AlexWaygood, line 37:
```suggestion
To run the tests in this directory, run `python tests/regr_test.py stdlib`
```

srittau: I don't even ... I blame Copilot. 😇

hauntsaninja (approved):

test_cases/README.md
```diff
@@ -32,6 +32,13 @@ stubs where decisions have been taken that might be slightly unusual. These
 test cases serve a different purpose: to check that type checkers do not emit
 false-positive errors for idiomatic usage of these classes.
 
+## Running the tests
+
+To run the tests in this directory, run `python tests/regr_test.py stdlib`
+from the root of the typeshed repository. This assumes that the development
+environment has been set up as described in the [CONTRIBUTING.md](../CONTRIBUTING.md)
+document.
```
AlexWaygood, line 40: These instructions are mypy-specific, but we also run pyright on these test cases in CI. Should we also add some instructions for pyright users?

I think the command you'd need to run would be `pyright --project pyrightconfig.testcases.json test_cases`, but I don't actually have pyright installed locally, so I can't verify that.

srittau: Let's go with these instructions for now. It would be useful to add a small stub script in tests, though, that call pyright with the correct paramters.

AlexWaygood: How about just this for now, then?

```suggestion
To verify the test cases in this directory with mypy, run `python tests/regr_test.py stdlib`
from the root of the typeshed repository. This assumes that the development
environment has been set up as described in the [CONTRIBUTING.md](../CONTRIBUTING.md)
document.
```

My hesitation with the current wording is that part of the point of these test cases is that they're supposed to be typechecker-agnostic, but the script is mypy-specific. (In theory we could start running pytype/pyre on these test cases at some point, as well as mypy and pyright.)

AlexWaygood (approved):
Thanks!

AlexWaygood:
One other thing we could do might be this, to make it easier to navigate between the different pieces of our docs for this script:

```diff
diff --git a/tests/README.md b/tests/README.md
index ccef075e2..47437c599 100644
--- a/tests/README.md
+++ b/tests/README.md
@@ -104,7 +104,8 @@ the stubs in typeshed (including the standard library).
 ## regr\_test.py

 This test runs mypy against the test cases for typeshed's stdlib and third-party
-stubs. See the README in the `test_cases` directory for more information about what
+stubs. See [the README in the `test_cases` directory](../test_cases/README.md)
+for more information about what
 these test cases are for and how they work. Run `python tests/regr_test.py --help`
 for information on the various configuration options.
```

PR #10937, opened 2023-10-27
Add `**kwargs` for logging.Logger methods

forana:
These are included in python's docs: https://docs.python.org/3/library/logging.html

### DIFF
--- a/stdlib/logging/__init__.pyi
+++ b/stdlib/logging/__init__.pyi
@@ -209,7 +209,6 @@ class Logger(Filterer):
             extra: Mapping[str, object] | None = None,
             stack_info: bool = False,
             stacklevel: int = 1,
-            **kwargs: object,
         ) -> None: ...  # undocumented
     else:
         def debug(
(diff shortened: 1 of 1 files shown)


thomasmccabe (approved):

AlexWaygood:
This would imply that `debug()`, etc. all take arbitrary keyword arguments. That doesn't seem to be true:

```pycon
Python 3.11.5 (tags/v3.11.5:cce6ba9, Aug 24 2023, 14:38:34) [MSC v.1936 64 bit (AMD64)] on win32
>>> import logging
>>> log = logging.Logger('foo')
>>> log.debug('debugging', foooooooooooooooooo=42)
  File "C:\Users\alexw\AppData\Local\Programs\Python\Python311\Lib\logging\__init__.py", line 1477, in debug
    self._log(DEBUG, msg, args, **kwargs)
TypeError: Logger._log() got an unexpected keyword argument 'foooooooooooooooooo'
```

forana:
@AlexWaygood oh, wonderful - you are indeed correct: https://github.com/python/cpython/blob/main/Lib/logging/__init__.py#L1501

The functions do declare `**kwargs`, but will error out if arguments that aren't valid for `_log` are passed. _Wonderful._

I'll close this out.

PR #10939, opened 2023-10-28
Enable ruff's isort rules on files generated using mypy-protobuf

hamdanal:
Originally opened to stop pre-commit-ci fixing other unrelated PRs due to https://results.pre-commit.ci/run/github/31696383/1698462880.Wm4kIn7fRrmhF3zdHHVXvg of https://github.com/python/typeshed/pull/10912.

The issue was then fixed in #10940 (thanks Alex). This is now a PR to enable running ruff's isort rules on *_pb2.pyi files!


AlexWaygood (approved):
Thanks! The cause here is a merge race between a PR that updated our generated protobuf and tensorflow stubs, and a PR that switched us from using isort to ruff for import sorting. 

On the one hand, all these files have `isort:skip_file` in their module docstrings. It seems obvious why ruff isn't respecting _that_ directive in the same way isort used to!

On the other hand, I don't see a compelling reason to exclude our generated stubs from import sorting, given that pre-commit should apply the fixes automatically when we next update our generated stubs. So I say we go ahead with this!

hamdanal:
You are right, I didn't see the isort skip until you mentioned it. I did some tests and it appears ruff only respects this line if it is a comment, and not in the docstring:

This doesn't work
```python
"""
isort:skip_file
"""
```

nor this
```python
"""
# isort:skip_file
"""
```

but this works
```python
"""
...
"""
# isort:skip_file
```

But as you said, there is no compelling reason to skip sorting imports in these files

AlexWaygood:
I think I'll create a stopgap PR to skip running ruff's isort rules on `*_pb2.pyi` files, since that preserves the status quo and will stop pre-commit-ci pushing unwelcome "autofixes" to every PR that's filed at typeshed. But let's keep this open since, as I say, there doesn't appear (to me) to be a compelling reason not to sort imports in these files.

AlexWaygood:
> I think I'll create a stopgap PR to skip running ruff's isort rules on `*_pb2.pyi` files, since that preserves the status quo and will stop pre-commit-ci pushing unwelcome "autofixes" to every PR that's filed at typeshed. But let's keep this open since, as I say, there doesn't appear (to me) to be a compelling reason not to sort imports in these files.

Done in #10940. This is now a PR to enable running ruff's isort rules on `*_pb2.pyi` files!

Avasam:
> On the one hand, all these files have `isort:skip_file` in their module docstrings. It seems obvious why ruff isn't respecting _that_ directive in the same way isort used to!

According to its doc, Ruff should respect the comment: https://docs.astral.sh/ruff/linter/#action-comments
So either it's a bug in Ruff, or the documentation is incorrect. Either way, worth raising upstream imo.

PR #10940, opened 2023-10-28
Skip ruff's isort rules on files generated with mypy-protobuf

AlexWaygood:
This takes us back to how things worked prior to https://github.com/python/typeshed/commit/49ba409da8906d7950dfedc7b34c3dea67bb0db9. See https://github.com/python/typeshed/pull/10939#pullrequestreview-1702764114 for an explanation of what's going on here.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -7,6 +7,48 @@ skip_magic_trailing_comma = true
 # for just these files, but doesn't seem possible yet.
 force-exclude = ".*_pb2.pyi"
 
+[tool.ruff]
+line-length = 130
+# Oldest supported Python version
+target-version = "py37"
+fix = true
+exclude = [
+    # virtual environment
+    ".env",
+    ".venv",
+    "env",
+    # cache directories, etc.:
+    ".git",
+    ".mypy_cache",
+    ".pytype",
+    ".venv",
+    "env",
+]
+select = [
+    "FA", # flake8-future-annotations
+    "I", # isort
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -7,6 +7,48 @@ skip_magic_trailing_comma = true
 # for just these files, but doesn't seem possible yet.
 force-exclude = ".*_pb2.pyi"
 
+[tool.ruff]
```
AlexWaygood, line 10: Moved this table above the `tool.ruff.isort` table, as I found the current order somewhat confusing

pyproject.toml
```diff
@@ -7,6 +7,48 @@ skip_magic_trailing_comma = true
...
+    "UP013", # Class-based syntax for TypedDicts
+    "UP014", # Class-based syntax for NamedTuples
+    "UP019", # Use str over typing.Text
+    "UP035", # import from typing, not typing_extensions, wherever possible
+    "UP039", # don't use parens after a class definition with no bases
+]
+
+[tool.ruff.per-file-ignores]
+"*_pb2.pyi" = ["I"]  # Skip isort rules for files generated with mypy-protobuf
+
```
AlexWaygood, line 51: This is the only bit of new config in this PR

Avasam:
According to its doc, Ruff should respect the comment: https://docs.astral.sh/ruff/linter/#action-comments
So either it's a bug in Ruff, or the documentation is incorrect. Either way, worth raising upstream imo.

PR #10942, opened 2023-10-28
fix nullability of scrypt params

### LINKED ISSUE #10941: Wrong argument nullability in hashlib.scrypt
The `hashlib.scrypt` function parameter type are wrong. The type hints indicate that `n`, `r` and `p` are optional:
https://github.com/python/typeshed/blob/5643362a1ea3187e3068a9107558e83a80afb334/stdlib/hashlib.pyi#L119-L122

but they are not:
https://docs.python.org/3/library/hashlib.html#hashlib.scrypt

Trolldemorted:
Fixes #10941

### DIFF
--- a/stdlib/hashlib.pyi
+++ b/stdlib/hashlib.pyi
@@ -113,14 +113,7 @@ shake_128 = _VarLenHash
 shake_256 = _VarLenHash
 
 def scrypt(
-    password: ReadableBuffer,
-    *,
-    salt: ReadableBuffer,
-    n: int,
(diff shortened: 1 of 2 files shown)


srittau:
Thanks! `hashlib.scrypt` needs to be added to `tests/stubtest_allowlists/py3_common.txt` (with an appropriate comment).

Trolldemorted:
@srittau like this?

srittau (approved):
Thanks!

PR #10944, opened 2023-10-28
Add stubs for portpicker

JelleZijlstra:


AlexWaygood:
weird that the upstream repo has `from __future__ import annotations` at the top but doesn't actually have any annotations

stubs/portpicker/portpicker.pyi
```diff
@@ -0,0 +1,18 @@
+from typing_extensions import TypeAlias
+
```
AlexWaygood, line 2:
```suggestion

__all__ = ('bind', 'is_port_free', 'pick_unused_port', 'return_port',
           'add_reserved_port', 'get_port_from_port_server')

```

stubs/portpicker/portpicker.pyi
```diff
@@ -0,0 +1,18 @@
...
+_Port: TypeAlias = int
+
+class NoFreePortFoundError(Exception): ...
+
+def add_reserved_port(port: _Port) -> None: ...
+def return_port(port: _Port) -> None: ...
+def bind(port: _Port, socket_type: int, socket_proto: int) -> _Port | None: ...
```
AlexWaygood, line 9: We could possibly be slightly stricter here, and say that `socket_type` has to be an instance of `socket.SocketKind`? The runtime accepts an `int`, but `socket.SocketKind` is more precise -- any member of that enum is guaranteed to work for this function, whereas it's not the case that any `int` is guaranteed to work

AlexWaygood (commented):
Incomplete review, I had to go halfway through looking

stubs/portpicker/portpicker.pyi
JelleZijlstra: Good catch. Cross-referencing python/mypy#10321: stubgen should do this for us.

stubs/portpicker/portpicker.pyi
JelleZijlstra: Good point, changing.

stubs/portpicker/portpicker.pyi
```diff
@@ -0,0 +1,21 @@
...
+__all__ = ("bind", "is_port_free", "pick_unused_port", "return_port", "add_reserved_port", "get_port_from_port_server")
+
+class NoFreePortFoundError(Exception): ...
+
+def add_reserved_port(port: _Port) -> None: ...
+def return_port(port: _Port) -> None: ...
+def bind(port: _Port, socket_type: socket.SocketKind, socket_proto: int) -> _Port | None: ...
+def is_port_free(port: _Port) -> bool: ...
+def pick_unused_port(pid: int | None = ..., portserver_address: str | tuple[str, _Port] | None = ...) -> _Port: ...
+def get_port_from_port_server(portserver_address: str | tuple[str, _Port], pid: int | None = ...) -> _Port | None: ...
```
AlexWaygood, line 15:
```suggestion
def pick_unused_port(pid: int | None = None, portserver_address: str | tuple[str, _Port] | None = None) -> _Port: ...
def get_port_from_port_server(portserver_address: str | tuple[str, _Port], pid: int | None = ...) -> _Port | None: ...
```

stubs/portpicker/portpicker.pyi
```diff
@@ -0,0 +1,21 @@
...
+__all__ = ("bind", "is_port_free", "pick_unused_port", "return_port", "add_reserved_port", "get_port_from_port_server")
+
+class NoFreePortFoundError(Exception): ...
+
+def add_reserved_port(port: _Port) -> None: ...
+def return_port(port: _Port) -> None: ...
+def bind(port: _Port, socket_type: socket.SocketKind, socket_proto: int) -> _Port | None: ...
+def is_port_free(port: _Port) -> bool: ...
+def pick_unused_port(pid: int | None = None, portserver_address: str | tuple[str, _Port] | None = None) -> _Port: ...
+def get_port_from_port_server(portserver_address: str | tuple[str, _Port], pid: int | None = None) -> _Port | None: ...
```
AlexWaygood, line 15: Are you sure it's okay to pass `tuple`s for `portserver_address`? I skimmed through the source code, and it looks to me like it only accepts a `str` for that argument. That seems to be borne out by my REPL experiments, too:

```pycon
Python 3.11.5 (tags/v3.11.5:cce6ba9, Aug 24 2023, 14:38:34) [MSC v.1936 64 bit (AMD64)] on win32
>>> import portpicker
>>> portpicker.pick_unused_port(portserver_address=('foo', 'bar'))
  File "C:\Users\alexw\coding\typeshed\.venv\Lib\site-packages\portpicker.py", line 223, in pick_unused_port
    return _pick_unused_port(pid, portserver_address)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\alexw\coding\typeshed\.venv\Lib\site-packages\portpicker.py", line 247, in _pick_unused_port
    port = get_port_from_port_server(portserver_address, pid=pid)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\alexw\coding\typeshed\.venv\Lib\site-packages\portpicker.py", line 435, in get_port_from_port_server
    buf = _windows_get_port_from_port_server(portserver_address, pid)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\alexw\coding\typeshed\.venv\Lib\site-packages\portpicker.py", line 389, in _windows_get_port_from_port_server
    handle = _winapi.CreateFile(
             ^^^^^^^^^^^^^^^^^^^
TypeError: CreateFile() argument 1 must be str, not tuple
>>> portpicker.get_port_from_port_server(portserver_address=('foo', 'bar'))
  File "C:\Users\alexw\coding\typeshed\.venv\Lib\site-packages\portpicker.py", line 435, in get_port_from_port_server
    buf = _windows_get_port_from_port_server(portserver_address, pid)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\alexw\coding\typeshed\.venv\Lib\site-packages\portpicker.py", line 389, in _windows_get_port_from_port_server
    handle = _winapi.CreateFile(
             ^^^^^^^^^^^^^^^^^^^
TypeError: CreateFile() argument 1 must be str, not tuple
```

JelleZijlstra: I believe it accepts a tuple on OSes that are not Windows and that don't support Unix sockets. That's probably a very small set. I can just narrow it to str.

AlexWaygood (approved):
Thanks!

PR #10946, opened 2023-10-29
Add types to some `tkinter.Text` methods.

TeamSpen210:
Some of the methods for `tkinter.Text` were not typed, this adds definitions to the `image_*()` and `window_*()` methods. `Text.count()` still remains, but [that method](https://tcl.tk/man/tcl8.6/TkCmd/text.htm#M96) is really complex, not sure exactly how it works.


stdlib/tkinter/__init__.pyi
```diff
@@ -3064,11 +3064,33 @@ class Text(Widget, XView, YView):
...
+    @overload
+    def image_configure(self, index: _TextIndex, cnf: str) -> tuple[str, str, str, str, str | int]: ...
+    @overload
+    def image_configure(
+        self,
+        index: _TextIndex,
+        cnf: dict[str, Any] | None = {},
+        *,
+        align: Literal["top", "center", "bottom", "baseline"] = ...,
+        image: _Image = ...,
```
Akuli, line 3077: These should be `_ImageSpec` instead of `_Image`.

stdlib/tkinter/__init__.pyi
```diff
@@ -3064,11 +3064,33 @@ class Text(Widget, XView, YView):
     def edit_separator(self) -> None: ...  # actually returns empty string
     def edit_undo(self) -> None: ...  # actually returns empty string
     def get(self, index1: _TextIndex, index2: _TextIndex | None = None) -> str: ...
-    # TODO: image_* methods
-    def image_cget(self, index, option): ...
-    def image_configure(self, index, cnf: Incomplete | None = None, **kw): ...
-    def image_create(self, index, cnf={}, **kw): ...
-    def image_names(self): ...
+    def image_cget(self, index: _TextIndex, option: str) -> str | int: ...  # image returned as string, others return str, int.
```
Akuli, line 3067: This is annoying to use, because you need something like:

```
align = text.image_cget(image_name, 'align')
assert isinstance(align, str)
# do something with align
```

Basically, when a user gets a type `str | int`, type checkers do not let the user just assume it is a string. They think you need to somehow handle both cases.

It would be better to do:

```suggestion
    def image_cget(self, index: _TextIndex, option: str) -> Any: ...  # image returned as string, others return str, int.
```

Or maybe:

```suggestion
    @overload
    def image_cget(self, index: _TextIndex, option: Literal["align", "image", "name"]) -> str: ...
    @overload
    def image_cget(self, index: _TextIndex, option: Literal["padx", "pady"]) -> int: ...
    @overload
    def image_cget(self, index: _TextIndex, option: str) -> Any: ...
```

stdlib/tkinter/__init__.pyi
```diff
@@ -3166,12 +3188,36 @@ class Text(Widget, XView, YView):
     def tag_ranges(self, tagName: str) -> tuple[_tkinter.Tcl_Obj, ...]: ...
     # tag_remove and tag_delete are different
     def tag_remove(self, tagName: str, index1: _TextIndex, index2: _TextIndex | None = None) -> None: ...
-    # TODO: window_* methods
-    def window_cget(self, index, option): ...
-    def window_configure(self, index, cnf: Incomplete | None = None, **kw): ...
+    def window_cget(self, index: _TextIndex, option: str) -> str | int: ...  # window returned as str, others are str/int.
```
Akuli, line 3191: This needs overloads or `Any` similarly to `image_cget()`.

stdlib/tkinter/__init__.pyi
```diff
@@ -3166,12 +3188,36 @@ class Text(Widget, XView, YView):
...
+        index: _TextIndex,
+        cnf: dict[str, Any] | None = {},
+        *,
+        align: Literal["top", "center", "bottom", "baseline"] = ...,
+        create: str = ...,
+        padx: _ScreenUnits = ...,
+        pady: _ScreenUnits = ...,
+        stretch: bool = ...,
+        window: Misc = ...,
+    ) -> None: ...  # actually returns empty string
```
Akuli, line 3219: On my system it returns None:

```suggestion
    ) -> None: ...
```

Akuli (commented):
Thanks for working on this! Tkinter is tricky :)

Akuli:
For what it's worth, I'm not sure how `.count()` works either. Tcl's underlying count command is somewhat reasonable, but tkinter's way of wrapping it into a Python method is just weird and surprising.

Last time I needed to use `.count()`, I did this:

```python
def count(widget: tkinter.Text, start: str, end: str, *, option: str = "-chars") -> int:
    # tkinter's .count() method is weird, returns tuples and Nones weirdly
    return widget.tk.call(widget, "count", option, start, end)
```

stdlib/tkinter/__init__.pyi
```diff
@@ -2891,6 +2891,7 @@ class Scrollbar(Widget):
     def set(self, first: float | str, last: float | str) -> None: ...
 
 _TextIndex: TypeAlias = _tkinter.Tcl_Obj | str | float | Misc
+_TextChildAlign: TypeAlias = Literal["top", "center", "bottom", "baseline"]
```
Akuli, line 2894: I would actually prefer copy/pasting `Literal["top", "center", "bottom", "baseline"]` into all places where it is used. Type aliases make stubs more confusing, and also provide a worse experience for IDE users: you see `_TextChildAlign` instead of the list of possible values.

stdlib/tkinter/__init__.pyi
```diff
@@ -3166,12 +3196,45 @@ class Text(Widget, XView, YView):
     def tag_ranges(self, tagName: str) -> tuple[_tkinter.Tcl_Obj, ...]: ...
     # tag_remove and tag_delete are different
     def tag_remove(self, tagName: str, index1: _TextIndex, index2: _TextIndex | None = None) -> None: ...
-    # TODO: window_* methods
-    def window_cget(self, index, option): ...
-    def window_configure(self, index, cnf: Incomplete | None = None, **kw): ...
+    @overload
+    def window_cget(self, index: _TextIndex, option: Literal["padx", "pady"]) -> int: ...
+    @overload
+    def window_cget(self, index: _TextIndex, option: Literal["stretch"]) -> Literal[0, 1]: ...
```
Akuli, line 3202: This is technically wrong, but it will be more convenient and practical to just use `bool`.

```suggestion
    def window_cget(self, index: _TextIndex, option: Literal["stretch"]) -> bool: ...  # actually returns Literal[0, 1]
```

For example, you might want to copy a stretch value from one embedded window to another, and you would do it like this:

```python
text.window_configure(index1, stretch=text.window_cget(index2, 'stretch'))
```

stdlib/tkinter/__init__.pyi
```diff
@@ -3166,12 +3196,45 @@ class Text(Widget, XView, YView):
...
+    @overload
+    def window_cget(self, index: _TextIndex, option: Literal["align"]) -> _TextChildAlign: ...
+    @overload  # window is set to a widget, but read as the string name.
```
Akuli, line 3205: We should also support setting it to a string, so that you can set any option to whatever `window_cget()` returns.

Pretty much all arguments can be given as strings in tkinter, but we type things with `... | str` only when passing in a string makes sense in practice. I think that is the case here.

stdlib/tkinter/__init__.pyi
TeamSpen210: An alternative solution is to make `window_configure` accept `bool | Literal[0, 1]`, that way this would still work while accurately describing the runtime behaviour. If it was annotated as returning bool I could imagine users doing `blah is True`. Though this would be probably inconsistent with the rest of the module..

Akuli: Indeed, we could change the `bool` handling in a separate PR.

Akuli (approved):
Thanks!

PR #10949, opened 2023-10-30
Fix incorrect typehint for callbacks in multiprocessing.Pool

Serious-senpai:
The typehints for callback function in `pool.Pool` methods are incorrect

### DIFF
--- a/stdlib/multiprocessing/pool.pyi
+++ b/stdlib/multiprocessing/pool.pyi
@@ -82,7 +82,7 @@ class Pool:
         func: Callable[..., _T],
         args: Iterable[Any] = (),
         kwds: Mapping[str, Any] = {},
-        callback: Callable[[Iterable[_T]], object] | None = None,
+        callback: Callable[[_T], object] | None = None,
         error_callback: Callable[[BaseException], object] | None = None,
     ) -> AsyncResult[_T]: ...
     def map(self, func: Callable[[_S], _T], iterable: Iterable[_S], chunksize: int | None = None) -> list[_T]: ...
--- a/stdlib/multiprocessing/pool.pyi
+++ b/stdlib/multiprocessing/pool.pyi
@@ -82,7 +82,7 @@ class Pool:
         func: Callable[..., _T],
         args: Iterable[Any] = (),
         kwds: Mapping[str, Any] = {},
-        callback: Callable[[_T], object] | None = None,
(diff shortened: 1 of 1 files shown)


stdlib/multiprocessing/pool.pyi
```diff
@@ -82,7 +82,7 @@ class Pool:
         func: Callable[..., _T],
         args: Iterable[Any] = (),
         kwds: Mapping[str, Any] = {},
-        callback: Callable[[_T], object] | None = None,
+        callback: Callable[[Iterable[_T]], object] | None = None,
```
srittau, line 85: I don't think this is correct. `apply_async`s callback gets passed a single result.

stdlib/multiprocessing/pool.pyi
```diff
@@ -91,7 +91,7 @@ class Pool:
         func: Callable[[_S], _T],
         iterable: Iterable[_S],
         chunksize: int | None = None,
-        callback: Callable[[_T], object] | None = None,
+        callback: Callable[[Iterable[_T]], object] | None = None,
```
srittau, line 94: In fact, a list is provided:

```suggestion
        callback: Callable[[list[_T]], object] | None = None,
```

stdlib/multiprocessing/pool.pyi
```diff
@@ -102,7 +102,7 @@ class Pool:
         func: Callable[..., _T],
         iterable: Iterable[Iterable[Any]],
         chunksize: int | None = None,
-        callback: Callable[[_T], object] | None = None,
+        callback: Callable[[Iterable[_T]], object] | None = None,
```
srittau, line 105:
```suggestion
        callback: Callable[[list[_T]], object] | None = None,
```

srittau (changes requested):
Thanks, remarks below!

stdlib/multiprocessing/pool.pyi
Serious-senpai: Sorry, this was my mistake.

stdlib/multiprocessing/pool.pyi
Serious-senpai: Should we use `list` or `typing.List`?

AlexWaygood: `list`

srittau (approved):

PR #10954, opened 2023-10-31
Delete `jsonschema._keywords.additionalItems`

### LINKED ISSUE #10953: Daily tests failed on Tue Oct 31 2023

AlexWaygood:
Fixes #10953


JelleZijlstra:
I prefer #10955 which also moves the method into another file.

AlexWaygood:
> I prefer #10955 which also moves the method into another file.

Yes, that's better :)

PR #10962, opened 2023-11-01
openpyxl TypeGuards and comparison dunders

Avasam:
Extracted from, and improved over, #9511

Typed all TypeGuards I could find (`def is*`, `def has*`, and related methods). As well as comparison dunders (`__contains__`, `__eq__`, `__nq__`, `__lt__`, `__gt__`, `__le__`, `__ge__`)


stubs/openpyxl/openpyxl/styles/numbers.pyi
```diff
@@ -51,12 +51,18 @@ LOCALE_GROUP: Final = r"\[(?!hh?\]|mm?\]|ss?\])[^\]]*\]"
 STRIP_RE: Final[Pattern[str]]
 TIMEDELTA_RE: Final[Pattern[str]]
 
-def is_date_format(fmt) -> bool: ...
-def is_timedelta_format(fmt): ...
-def is_datetime(fmt): ...
-def is_builtin(fmt): ...
+def is_date_format(fmt: str | None) -> TypeGuard[str]: ...
+def is_timedelta_format(fmt: str | None) -> TypeGuard[str]: ...
```
srittau, line 55: I'm unsure whether we can really use type guard here, as these do more than just check the type. Maybe it would make more sense to use overloads?

```suggestion
@overload
def is_date_format(fmt: None) -> None: ...
@overload
def is_date_format(fmt: str) -> bool: ...
@overload
def is_timedelta_format(fmt: None) -> None: ...
@overload
def is_timedelta_format(fmt: str) -> bool: ...
```

Cc @erictraut as resident type guard expert.

Implementation for reference: https://foss.heptapod.net/openpyxl/openpyxl/-/blob/branch/3.1/openpyxl/styles/numbers.py

stubs/openpyxl/openpyxl/styles/numbers.pyi
```diff
@@ -51,12 +51,18 @@ LOCALE_GROUP: Final = r"\[(?!hh?\]|mm?\]|ss?\])[^\]]*\]"
...
+def is_datetime(fmt: None) -> None: ...
+@overload
+def is_datetime(fmt: str) -> Literal["datetime", "date", "time"] | None: ...
+def is_builtin(fmt: str | None) -> TypeGuard[str]: ...
 def builtin_format_code(index): ...
-def builtin_format_id(fmt): ...
+@overload
+def builtin_format_id(fmt: None) -> None: ...
+@overload
+def builtin_format_id(fmt: str) -> str | None: ...
```
srittau, line 65:
```suggestion
def builtin_format_id(fmt: str) -> int | None: ...
```

stubs/openpyxl/openpyxl/worksheet/cell_range.pyi
```diff
@@ -18,7 +18,7 @@ class CellRange(Serialisable):
     @overload
     def __init__(
         self,
-        range_string: str | None,
+        range_string: str,
```
srittau, line 21: `None` is check explicitly *and* is the default: https://foss.heptapod.net/openpyxl/openpyxl/-/blob/branch/3.1/openpyxl/worksheet/cell_range.py#L48

```suggestion
        range_string: str | None = None,
```

stubs/openpyxl/openpyxl/worksheet/datavalidation.pyi
```diff
@@ -81,7 +84,7 @@ class DataValidation(Serialisable):
         allow_blank: Incomplete | None = False,
     ) -> None: ...
     def add(self, cell) -> None: ...
-    def __contains__(self, cell): ...
+    def __contains__(self, cell: _HasCoordinate) -> bool: ...
```
srittau, line 87: This doesn't seem right to me, as a `hasattr` check explicitly checks for the the existence of `coordinate`, but continues whether it exists or not. I don't know what the correct annotation should be though. Maybe just `_HasCoordinate | Incomplete` for now?

srittau (changes requested):
Thanks for slugging through this! Notes below.

stubs/openpyxl/openpyxl/worksheet/cell_range.pyi
Avasam: That's what the overload below is for. If `range_string` is `None`, you HAVE TO specify min/max col/row. If it's a string, then the min/max col/row params are unused.

stubs/openpyxl/openpyxl/worksheet/datavalidation.pyi
Avasam: `_HasCoordinate | CellRange`, thanks for the heads-up. I went too fast here and forgot to compelte the annotation after making the protocol.

Avasam: `_HasCoordinate | CellRange | str`, thanks for the heads-up. I went too fast here and forgot to compelte the annotation after making the protocol.

stubs/openpyxl/openpyxl/styles/numbers.pyi
Avasam: oops, thanks

stubs/openpyxl/openpyxl/styles/numbers.pyi
Avasam: My understanding of `TypeGuard` is that if the function returns `True` then the parameter is necessarily of the guarded type. Which is the case for both here: `fmt=None` will imediatly return `False`. `fmt=""` may or may not return `True`.
I don't see anything in https://peps.python.org/pep-0647 suggesting that returning `False` has to narrow.

That being said, mypy seems to agree here, but pyright/pylance seems to ignore the typeguard, so I'll wait for an answer:
image
image
image

stubs/openpyxl/openpyxl/worksheet/cell_range.pyi
srittau: For some reason I totally missed the overload ...

stubs/openpyxl/openpyxl/styles/numbers.pyi
erictraut: This looks to me like a fine use of `TypeGuard`. The overload approach is correct from a type perspective, but it won't result in type narrowing when the function is used in a conditional statement, so `TypeGuard` is probably more appropriate here than an overload.

Both pyright and mypy honor the type guard in this test code.

```python
from typing import TypeGuard


def is_date_format(fmt: None | str) -> TypeGuard[str]:
    ...

def func(test: str | None):
    if is_date_format(test):
        reveal_type(test)  # str
    else:
        reveal_type(test)  # str | None
```

srittau (approved):

PR #10968, opened 2023-11-02
Update xml's `Callable[..., Element]` types (`tag` param in `__init__`)

Avasam:
The `tag` param in `__init__` is typed as `str | Callable[..., Element]`, which really doesn't make sense to me looking at usages, and there's no comment explaining why. It really seems like the `tag` parameter of `Element.__init__` should not be `Callable`.

Sidenote, even though `Element.itertext` does [check for possible `None` `self.tag`](https://github.com/python/cpython/blob/3.12/Lib/xml/etree/ElementTree.py#L402), the [docstring](https://github.com/python/cpython/blob/3.12/Lib/xml/etree/ElementTree.py#L183) and [documentation](https://docs.python.org/3/library/xml.etree.elementtree.html#xml.etree.ElementTree.Element) do suggest it should only be a `str`. So I left them as `str` and not `str | None`.

Relevant for #10967

---

Using re-assignement for `PI` (as implementation does) just leads to better method signatures.


stubs/openpyxl/openpyxl/xml/_functions_overloads.pyi
```diff
@@ -16,8 +16,10 @@ _T_co = TypeVar("_T_co", covariant=True)
 
 # Comment from openpyxl.cell.rich_text.py
 # Usually an Element() from either lxml or xml.etree (has a 'tag' element)
+# lxml.etree._Element
+# xml.etree.Element
 class _HasTag(Protocol):
-    tag: Any  # AnyOf[str, None, Callable[..., AnyOf[str, None]]]
```
Avasam, line 20: This strange Union was caused by `Element`'s typing leading me to believe it could be `Callable` and the runtime implementation leading me to believe it could be `None` (see PR description).
Looking at xml.etree._Element typing from lxml-stubs as well as implementation of openpyxl looks like this is never assumed to be `None` neither.

PR #10969, opened 2023-11-03
Improve hdbcli types

kasium:
* parameters argument of execute methods is optional and allows none values
* paramstyle is a str and a tuple of str (hdbcli does not strictly follow PEP 249)
* Add missing functions to ResultRow


stubs/hdbcli/hdbcli/dbapi.pyi
```diff
@@ -4,13 +4,13 @@ from collections.abc import Sequence
...
 from typing import Any, overload
-from typing_extensions import Literal, TypeAlias
+from typing_extensions import Literal, Self, TypeAlias
 
 from .resultrow import ResultRow
 
 apilevel: str
 threadsafety: int
-paramstyle: tuple[str, ...]
+paramstyle: str | tuple[str, ...]  # hdbcli defines it as a tuple which does not follow PEP 249
```
srittau, line 13: I like the comment, but when it's defined as a tuple, why the `str` alternative?

kasium: Valid point. I thought that there is a use case setting a value, but there is not

kasium:
@srittau can you please consider merging it? Thanks a lot!

srittau:
@kasium See my last comment: Could you re-add the comment about paramstyle being non-standard?

srittau (approved):

PR #10971, opened 2023-11-03
Complete Flask-Migrate and mark as Strict

Avasam:
Thanks to typeshed-stats, I noticed Flask-Migrate had very little unannotated methods, so I decided to take a swing at it.

See comments below for more details


stubs/Flask-Migrate/flask_migrate/__init__.pyi
```diff
@@ -9,20 +12,33 @@ from flask_sqlalchemy import SQLAlchemy
...
     template_directory: str | None
-    def __init__(self, *args, **kwargs) -> None: ...
+    # Same as alembic.config.Config + template_directory kwarg
+    def __init__(
+        self,
+        file_: StrPath | None = None,
+        ini_section: str = "alembic",
+        output_buffer: TextIO | None = None,
+        stdout: TextIO = sys.stdout,
+        cmd_opts: Namespace | None = None,
```
Avasam, line 29: `TextIO` and `Namespace` come directly from `alembic`'s own inline types, but I have some concern about using those.
Namely if `TextIO` should be replaced by some buffer protocol.

stubs/Flask-Migrate/flask_migrate/__init__.pyi
```diff
@@ -9,20 +12,33 @@ from flask_sqlalchemy import SQLAlchemy
...
+        config_args: SupportsKeysAndGetItem[str, _AlembicConfigValue] | Iterable[tuple[str, _AlembicConfigValue]] = ...,
```
Avasam, line 30: `config_args` and `attributes` are used to create and update a `dict` respectively. I don't remember if it's better to type this as a `Mapping` like `alembic` did for `config_args` or maybe make a `ConvertibleToDict` type like in #10707

stubs/Flask-Migrate/flask_migrate/__init__.pyi
srittau: I haven't looked at it, but generally for I/O types in argument positions, protocols are preferred.

Avasam: I dig down a bit more.
`stdout` goes to https://github.com/sqlalchemy/alembic/blob/main/alembic/util/messaging.py#L47
`output_buffer` goes to https://github.com/sqlalchemy/alembic/blob/main/alembic/runtime/migration.py#L173

stubs/Flask-Migrate/flask_migrate/__init__.pyi
```diff
@@ -1,28 +1,53 @@
...
 _T = TypeVar("_T")
+_T_contra = TypeVar("_T_contra", contravariant=True)
 _P = ParamSpec("_P")
 _ConfigureCallback: TypeAlias = Callable[[Config], Config]
+_AlembicConfigValue: TypeAlias = Any
 
 alembic_version: tuple[int, int, int]
 log: Logger
 
+class _SupportsWriteAndFlush(SupportsWrite[_T_contra], SupportsFlush, Protocol): ...
```
srittau, line 23: `SupportsFlush` is new and not yet distributed with at least mypy. It's safer to do this for now:

```suggestion
# TODO: Use _typeshed.SupportsFlush when it's available in type checkers.
class _SupportsWriteAndFlush(SupportsWrite[_T_contra], Protocol):
    def flush(self) -> object: ...
```

srittau (changes requested):
Thanks, one suggestion for type checker compatibility.

srittau (approved):

PR #10976, opened 2023-11-04
Support for pexpect.spawn(..., logfile=sys.stdout)

PokkaKiyo:
We can set the logfile for pexpect's spawn using sys.stdout, but the type hint shows an error.

### DIFF
--- a/stubs/pexpect/pexpect/fdpexpect.pyi
+++ b/stubs/pexpect/pexpect/fdpexpect.pyi
@@ -1,6 +1,6 @@
 from _typeshed import Incomplete
 
-from .spawnbase import SpawnBase
+from .spawnbase import SpawnBase, _Logfile
 
 class fdspawn(SpawnBase):
     args: Incomplete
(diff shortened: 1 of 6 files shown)


srittau (commented):
Thanks! Looking at the source code, it might be better to just define a custom protocol, which we can use for both `SpawnBase` and `spawn`:

```python
class Logfile(Protocol):
    def write(self, __s: str) -> object:...
    def flush(self) -> object: ...
```

Source:

https://github.com/pexpect/pexpect/blob/c53497663c1a25603ccf8ba7daca03b319387344/pexpect/spawnbase.py#L128-L130

AlexWaygood:
> Thanks! Looking at the source code, it might be better to just define a custom protocol, which we can use for both `SpawnBase` and `spawn`:
> 
> ```python
> class _Logfile(Protocol):
>     def write(self, __s: str) -> object:...
>     def flush(self) -> object: ...
> ```

We already define a protocol like that elsewhere in typeshed. Maybe we should add it to `stdlib/_typeshed/__init__.pyi`?:

https://github.com/python/typeshed/blob/6c5d1d3ae89453523fdcca939be00cf91cb42ef9/stdlib/builtins.pyi#L1640-L1641

srittau:
> We already define a protocol like that elsewhere in typeshed. Maybe we should add it to stdlib/_typeshed/__init__.pyi?

While I would be ok with that, I am not a huge fan of adding all sometimes-used permutations of I/O methods to `_typeshed`. I'd rather just switch to intersection types (python/typing#213) at some point in the far future. That said, moving the protocol to `_typeshed` could help use find places where we could use intersection types in the future.

(Also, `flush()` should return `object`, not `None`.)

srittau (approved):
Thanks!

PR #10977, opened 2023-11-04
Add pymacaroons stubs

lengau:
https://github.com/ecordell/pymacaroons


stubs/pymacaroons/pymacaroons/macaroon.pyi
```diff
@@ -0,0 +1,49 @@
...
+    @classmethod
+    def deserialize(
+        cls, serialized: str | bytes | bytearray, serializer: BaseSerializer | JsonSerializer | None = None
+    ) -> Macaroon: ...
+    @property
+    def location(self) -> str: ...
+    @property
+    def version(self) -> int: ...
+    @property
+    def identifier(self) -> str | bytes: ...
```
lengau, line 33: I'd like advice on what to do with this. A v1 macaroon gets this as `str`, a v2 as `bytes` (set by the `version` parameter to `__init__`).

stubs/pymacaroons/@tests/test_cases/check_macaroon.py
```diff
@@ -0,0 +1,38 @@
...
+]
+
+for macaroon in macaroons:
+    assert_type(macaroon.serialize(), str)
+    assert_type(macaroon.location, str)
+    assert_type(macaroon.version, int)
+    assert_type(macaroon.identifier, Union[str, bytes])
+    assert_type(macaroon.identifier_bytes, bytes)
+    assert_type(macaroon.signature, str)
+    assert_type(macaroon.copy(), Macaroon)
```
srittau, line 38: Unfortunately, the loop won't really work as a loop, since the tests are not run, but only checked statically. But since `__init__` has no overloads at the moment, we don't really need to check more than one object `Macaroon` object. If we add an `@overload` (see below), we could statically assign two `Macaroon` objects, one with `MACAROON_V1` and one with `MACAROON_V2` and just duplicate the assert_types.

stubs/pymacaroons/@tests/test_cases/check_serialization.py
```diff
@@ -0,0 +1,13 @@
...
+from pymacaroons import Macaroon
+from pymacaroons.serializers import BinarySerializer, JsonSerializer
+
+macaroons = [Macaroon.deserialize(""), Macaroon.deserialize("", JsonSerializer()), Macaroon.deserialize("", BinarySerializer())]
+
+for macaroon in macaroons:
+    assert_type(macaroon.serialize(), str)
+    assert_type(macaroon.serialize(None), str)
+    assert_type(macaroon.serialize(JsonSerializer()), str)
+    assert_type(macaroon.serialize(BinarySerializer()), str)
```
srittau, line 13: Same issue here.

stubs/pymacaroons/pymacaroons/macaroon.pyi
srittau: Maybe it would make sense to make the `Macaroon` object generic over the identifier type and add overloads to `__init__`:

```python
class Macaroon(Generic[AnyStr]):
    @overload
    def __init__(self: Self[str], ..., version: Literal[1] = 1) -> None: ...
    @overload
    def __init__(self: Self[bytes], ..., version: Literal[2]) -> None: ...  # all extra arguments without default
    @overload
    def __init__(self: Self[bytes], ..., *, version: Literal[2]) -> None: ...  # all extra arguments with default
    ...
    @property
    def identifier(self) -> AnyStr: ...
```

stubs/pymacaroons/pymacaroons/macaroon.pyi
```diff
@@ -0,0 +1,49 @@
+from collections.abc import Collection
+
+from pymacaroons import Caveat
+from pymacaroons.binders import HashSignaturesBinder
+from pymacaroons.caveat_delegates import FirstPartyCaveatDelegate, ThirdPartyCaveatDelegate
+from pymacaroons.serializers import BaseSerializer, JsonSerializer
+
+MACAROON_V1: int
+MACAROON_V2: int
```
srittau, line 9: These should be final, otherwise  the overload trick doesn't work with these constants:

```suggestion
MACAROON_V1: Final = 1
MACAROON_V2: Final = 2
```

srittau (commented):
Not a full review, just some remarks about the `identifier` problem and the tests.

srittau:
Thanks for contributing! I'm closing this PR for now, because it still fails some tests and has unresolved review feedbackafter three months of inactivity. If you are still interested, please feel free to open a new PR (or ping us to reopen this one).

PR #10978, opened 2023-11-05
http.HTTPConnection.putheader tweak

Tinche:
`http.HTTPConnection.putheader` supports bytes and strings as header names and values.

Here's a test demonstrating this: https://github.com/python/cpython/blob/main/Lib/test/test_httplib.py#L233


Tinche:
This work was actually motivated by some work I was trying to do in urllib3, so I'll probably be fixing this breakage over there if/when this gets merged.

PR #10979, opened 2023-11-05
Fix the hint for XMLGenerator.__init__().

### LINKED ISSUE #10550: xml.sax.saxutils.XMLGenerator can take SupportsWrite[bytes]
Currently `XMLGenerator.__init__` is typed as taking `SupportsWrite[str]` for the code path in `xml.sax.saxutils._gettextwriter()` that wraps the file-like object into `io.TextIOWrapper` which, I think, can take binary file-like objects.

Please let me know if I'm wrong.

wRAR:
Fixes #10550.

I've checked that `TextIOWrapper` (in addition to docs saying it wraps a binary stream) indeed doesn't work with a str-based stream so keeping `SupportsWrite[str]` indeed is wrong.


srittau (approved):
Thanks!

PR #10983, opened 2023-11-06
Introduce _typeshed.SupportsFlush

srittau:
`flush()` is a standard I/O method. This allows us to construct I/O protocols using composition. (I.e. using base classes at the moment and hopefully using the `&` operator at some point in the future.)

### DIFF
--- a/stdlib/_typeshed/__init__.pyi
+++ b/stdlib/_typeshed/__init__.pyi
@@ -236,6 +236,10 @@ class SupportsNoArgReadline(Protocol[_T_co]):
 class SupportsWrite(Protocol[_T_contra]):
     def write(self, __s: _T_contra) -> object: ...
 
+# stable
+class SupportsFlush(Protocol):
+    def flush(self) -> object: ...
+
 # Unfortunately PEP 688 does not allow us to distinguish read-only
 # from writable buffers. We use these aliases for readability for now.
 # Perhaps a future extension of the buffer protocol will allow us to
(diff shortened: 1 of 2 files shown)


stdlib/_typeshed/__init__.pyi
```diff
@@ -236,6 +236,10 @@ class SupportsNoArgReadline(Protocol[_T_co]):
 class SupportsWrite(Protocol[_T_contra]):
     def write(self, __s: _T_contra) -> object: ...
 
+# stable
+class SupportsFlush(Protocol):
+    def flush(self) -> object: ...
```
AlexWaygood, line 241: I actually feel like this might be a situation where we could be opinionated and say, "Actually, file-like objects are doing the wrong thing if they return a non-`None` object from their `flush()` method; that's pretty weird". But I don't feel strongly about it!

AlexWaygood (approved):

stdlib/_typeshed/__init__.pyi
srittau: For consistency alone, I would use `object` here. Also, it catches the (admittedly very theoretical) problem, where someone calling `flush()` tries to use the return value.

AlexWaygood: Sure, I'm fine with keeping this as `-> object`

srittau: (We could also start using `Unused` instead of `object` for protocol return types to signal intent better, but this should be done in a separate PR.)

PR #10984, opened 2023-11-06
asyncio: remove overly specific protocols

JelleZijlstra:
The _warn parameter to these methods is just there to work around
some finalization issues and should never be used by users. In
addition, these protocols are out of date (the "stacklevel" argument
is not used by current CPython main). I don't think we gain anything
by trying to maintain these protocol definitions.


srittau:
Maybe we can just remove `__del__`? Since the argument is for internal use only anyway, I don't think there is much value in keeping the methods. (Would need an allowlist entry, though, probably.)

AlexWaygood:
I think stubtest actually already doesn't care whether `__del__` exists in a stub or not

srittau:
After a quick grep: We have 16 instances of `__del__` in stdlib, 8 of which are in asyncio. I'd say, let's get rid of them and see whether stubtest complains.

AlexWaygood:
> After a quick grep: We have 16 instances of `__del__` in stdlib, 8 of which are in asyncio. I'd say, let's get rid of them and see whether stubtest complains.

Sounds good! I'm pretty confident stubtest won't complain, because of https://github.com/python/mypy/blob/544e6ce119ec6bdd5eabab53a433264c98dc7d9c/mypy/stubtest.py#L1377

JelleZijlstra:
One reason to keep it: https://docs.python.org/3/reference/datamodel.html#object.__del__ says 

> If a base class has a [__del__()](https://docs.python.org/3/reference/datamodel.html#object.__del__) method, the derived class’s [__del__()](https://docs.python.org/3/reference/datamodel.html#object.__del__) method, if any, must explicitly call it to ensure proper deletion of the base class part of the instance.

A type checker or other kind of linter might want to enforce this rule, but to do so, it needs to know whether `__del__` exists in the base class. So I do think there is value in having these methods in typeshed.

JelleZijlstra:
And relatedly, removing `__del__` methods from typeshed would mean that a derived class that does `super().__del__()` would get a false positive from a type checker, because there is no `object.__del__`.

AlexWaygood:
> One reason to keep it: https://docs.python.org/3/reference/datamodel.html#object.__del__ says
> 
> > If a base class has a [**del**()](https://docs.python.org/3/reference/datamodel.html#object.__del__) method, the derived class’s [**del**()](https://docs.python.org/3/reference/datamodel.html#object.__del__) method, if any, must explicitly call it to ensure proper deletion of the base class part of the instance.
> 
> A type checker or other kind of linter might want to enforce this rule, but to do so, it needs to know whether `__del__` exists in the base class. So I do think there is value in having these methods in typeshed.

In that case, we might want to remove it from stubtest's `IGNORABLE_CLASS_DUNDERS` frozenset

JelleZijlstra:
> In that case, we might want to remove it from stubtest's IGNORABLE_CLASS_DUNDERS frozenset

Agree. For this PR though, I think I'd prefer to just pretend the `_warn` argument doesn't exist, since no users should be calling it and it exists only to early-bind the warn function. (The problem it guards against is that when `__del__` is called, the module globals may already have been cleared.)

srittau (approved):

tests/stubtest_allowlists/py3_common.txt
```diff
@@ -374,6 +374,10 @@ threading.Condition.release
 multiprocessing.dummy.Condition.acquire
 multiprocessing.dummy.Condition.release
 
+# Some of these take a "_warn" parameter at runtime
+# for technical reasons; no need to expose it in the stub.
+asyncio\..*\.__del__
```
AlexWaygood, line 379: I'd actually prefer it if we tediously enumerated all the methods that need to be allowlisted here, since using wildcards in allowlist entries like this often comes back to bite us later

PR #10988, opened 2023-11-07
Bump ruff

sobolevn:
Btw, since `ruff@1.0.2` there's a builtin autoformatter instead of `black`. 

It produces a code that is a bit different to one that `black` produces. I don't think that we need to change anything, just spreading the news.


srittau (changes requested):
`.github/workflows/tests.yml` must also be changed.

srittau (approved):

PR #10989, opened 2023-11-07
Fix return type of dbm.whichdb

tmke8:
It is `None` if the given file doesn't exist: https://docs.python.org/3/library/dbm.html#dbm.whichdb


srittau (approved):
Thanks!

Issue #10980, opened 2023-11-06
Mypy complains about the `all_equal` recipe from the itertools docs


AlexWaygood:
The following function is given as [a recipe in the `itertools` docs](https://docs.python.org/3/library/itertools.html#itertools-recipes), indicating that it's an idiomatic usage of `groupby`:

```py
from itertools import groupby

def all_equal(iterable):
    g = groupby(iterable)
    return next(g, True) and not next(g, False)
```

I think the correct way of adding type annotations to this function would be as follows, since it will work on arbitrary iterables:

```py
from collections.abc import Iterable
from itertools import groupby

def all_equal(iterable: Iterable[object]) -> bool:
    g = groupby(iterable)
    return next(g, True) and not next(g, False)
```

Unfortunately, however, mypy complains about this function:

```
error: Argument 1 to "next" has incompatible type "groupby[object, object]"; expected "SupportsNext[bool]"  [arg-type]
```

(Mypy gives a similar error if I use `Iterable[Any]` instead of `Iterable[object]` for the argument annotation.)

Perhaps we should consider copy-and-pasting all the itertools recipes into our `test_cases` directory. They're all meant to be idiomatic uses of itertools, so if any of them fail to type check, there's probably a problem somewhere.

AlexWaygood:
(I encountered this in https://github.com/PyCQA/flake8-pyi/pull/432)

AlexWaygood:
Hmm, I initially assumed this was probably an issue with typeshed's stubs -- either the stub for `itertools.groupby`, or the stub for `builtins.next`. Looking closer, I'm now not so sure. Pyright seems okay with this code; this may just be a problem with mypy's ability to solve the type variables here.

srittau:
> Perhaps we should consider copy-and-pasting all the itertools recipes into our test_cases directory. They're all meant to be idiomatic uses of itertools, so if any of them fail to type check, there's probably a problem somewhere.

Sounds like a good idea.

To the problem at hand: This sounds like a mypy bug to me. This is how the `next()` overload that should be used is defined:

```python
@overload
def next(__i: SupportsNext[_T], __default: _VT) -> _T | _VT: ...
```

[...]

PR #10992, opened 2023-11-08
Add several itertools recipes to the `test_cases` directory

AlexWaygood:
As discussed in #10980. Several of the recipes are deliberately omitted for now; not all are easy to add to our `test_cases` directory, for various reasons. I defer the ones where certain decisions might need debate to future PRs; this PR attempts to add only the recipes which can be added to our `test_cases` directory relatively easily.


AlexWaygood:
The recipes themselves are copied-and-pasted from the itertools docs, docstrings and comments included. All I've done to them is add type annotations where necessary, in order to get our tests to pass.

AlexWaygood:
In total, this adds 24/33 recipes (25 if you count the one I "added" that's commented out, because it really needs PEP-646).

AlexWaygood:
> In total, this adds 24/33 recipes (25 if you count the one I "added" that's commented out, because it really needs PEP-646).

Now it _is_ 25/33, since the one that needs PEP-646 has been uncommented :)

test_cases/stdlib/itertools/check_itertools_recipes.py
```diff
@@ -0,0 +1,368 @@
...
+        return starmap(func, repeat(args))
+    return starmap(func, repeat(args, times))
+
+
+def flatten(list_of_lists: Iterable[Iterable[_T]]) -> chain[_T]:
+    "Flatten one level of nesting"
+    return chain.from_iterable(list_of_lists)
+
+
+def ncycles(iterable: Iterable[_T], n: int) -> Iterator[_T]:
```
srittau, line 57:
```suggestion
def ncycles(iterable: Iterable[_T], n: int) -> chain[_T]:
```

test_cases/stdlib/itertools/check_itertools_recipes.py
```diff
@@ -0,0 +1,368 @@
...
+            if k not in seen:
+                seen.add(k)
+                yield element
+        # For use cases that allow the last matching element to be returned,
+        # a faster but non-lazy solution is:
+        #      t1, t2 = tee(iterable)
+        #      yield from dict(zip(map(key, t1), t2)).values()
+
+
+def powerset(iterable: Iterable[_T]) -> Iterator[tuple[_T, ...]]:
```
srittau, line 272:
```suggestion
def powerset(iterable: Iterable[_T]) -> chain[tuple[_T, ...]]:
```

test_cases/stdlib/itertools/check_itertools_recipes.py
```diff
@@ -0,0 +1,368 @@
...
+    "powerset([1,2,3]) --> () (1,) (2,) (3,) (1,2) (1,3) (2,3) (1,2,3)"
+    s = list(iterable)
+    return chain.from_iterable(combinations(s, r) for r in range(len(s) + 1))
+
+
+def polynomial_derivative(coefficients: Sequence[int]) -> list[int]:
```
srittau, line 278: Also works with `float` and other types that can be multiplied with `int`s (decimal?). It could be worthwhile to use a type var here, but that's optional.

srittau (changes requested):

test_cases/stdlib/itertools/check_itertools_recipes.py
JelleZijlstra: I think an Iterator return is actually better here; there's nothing useful you can do with a chain object you can't do with any Iterator, and it's basically an implementation detail exactly what kind of iterator you're returning.

AlexWaygood: I think there's exactly one useful thing that you can do with a `chain` that you can't do with another other `Iterator`: you can access the `from_iterable` classmethod.

Accessing an alternative constructor like that from an instance is pretty unusual, though, and I'm inclined to agree that the precise kind of iterator being returned here is probably an implementation detail. I think there's another place in this file where I already have `-> chain`; I'm actually inclined to change _that_ one so that it's `-> Iterator`

srittau: > I think an Iterator return is actually better here; there's nothing useful you can do with a chain object you can't do with any Iterator, and it's basically an implementation detail exactly what kind of iterator you're returning.

In that case, we should change the annotations that are returning `chain` to `Iterator` as well.

srittau (approved):

AlexWaygood:
Our first use of PEP-646 in typeshed! Even if it is only in a test case

PR #10996, opened 2023-11-08
stdlib: Fix os.startfile

Jakob-Stadler:
Attempted fix for #10991


JelleZijlstra (approved):
Looks consistent with the source code in CPython.

hauntsaninja (approved):
Thank you!

PR #10998, opened 2023-11-09
Add types for `pygments.lex`.

jonathanslenders:
This adds type annotations to the `pygments.lex` helper.

The return type is the same as what we use in `pygments.Lexer.get_tokens`.


srittau (approved):
Thanks!

PR #10994, opened 2023-11-08
ast: mark ast.Num etc. as deprecated

JelleZijlstra:
PEP 702 was just accepted. As the decorator should simply be ignored
by type checkers that do not support it yet, I don't see anything
blocking us from using it now.

This is the first simple-looking deprecation in
https://docs.python.org/3.12/whatsnew/3.12.html#deprecated
I'll see what CI has to say here before looking at others.

PR #11001, opened 2023-11-09
Mark deprecated methods as `deprecated` in `wave` module

sobolevn:


stdlib/wave.pyi
```diff
@@ -37,7 +37,9 @@ class Wave_read:
     def getcomptype(self) -> str: ...
     def getcompname(self) -> str: ...
     def getparams(self) -> _wave_params: ...
+    @deprecated("Removal scheduled for Python 3.15")
```
AlexWaygood, line 40: Would it be worth mentioning the version it's deprecated in, as well? Python 3.15 will be a long way off for people still using Python 3.8

```suggestion
    @deprecated("Deprecated in Python 3.13; removal scheduled for Python 3.15")
```

JelleZijlstra: I didn't do this in #10994, but i wonder if we should apply the decorator conditionally (under `if sys.version_info >= (3, 13)` in this case). Should users who are say on 3.9 see this warning? On the one hand, it's useful for users to know that something they're using will go away; on the other hand, if they're on an old version, the replacement may not be usable yet, so the warning is not actionable.

AlexWaygood: Yes, I was writing up this issue as you were posting this comment :)

- #11002

PR #11003, opened 2023-11-09
Review `SupportsInt` and `SupportsFloat` usages in 3rd-party stubs

Avasam:
Mypy 1.7 has been released since #10707 .
I have reviewed all usages of  `SupportsInt` and `SupportsFloat` that didn't use the full Union.
Direct replacements have been extracted to #11022

Which also makes me realize two things:
1. A bunch of `Supports*` in stdlib could also use `ConvertibleTo*` instead.
2. `SupportsInt` isn't used anywhere in 3rd-party stubs anymore. But `CONTIBUTING.md`'s guide section about Protocols providing callable methods explicitly mentions it. Maybe it'd be a good idea to mention the `Convertible` aliases also/instead ?


stubs/PyAutoGUI/pyautogui/__init__.pyi
```diff
@@ -19,7 +20,10 @@ from pyscreeze import (
 
 _P = ParamSpec("_P")
 _R = TypeVar("_R")
-_NormalizeableXArg: TypeAlias = str | SupportsInt | Sequence[SupportsInt]
+# Explicitely mentionning str despite being in the ConvertibleToInt Alias because it has a different meaning (filename on screen)
```
AlexWaygood, line 23:
```suggestion
# Explicitly mentioning str despite being in the ConvertibleToInt Alias because it has a different meaning (filename on screen)
```

AlexWaygood:
> Mypy 1.6 (and 1.6.1) has been released since #10707 .

We still can't go ahead with this just yet, unfortunately. If you look at mypy's vendored stubs in `mypy/typeshed/stdlib/_typeshed/__init__.py`, it doesn't include these aliases in the `1.6.1` tag (which is the latest release). We'll need to wait for mypy 1.7, which should include these aliases in mypy's vendored version of typeshed.

Avasam:
> > Mypy 1.6 (and 1.6.1) has been released since #10707 .
> 
> We still can't go ahead with this just yet, unfortunately. If you look at mypy's vendored stubs in `mypy/typeshed/stdlib/_typeshed/__init__.py`, it doesn't include these aliases in the `1.6.1` tag (which is the latest release). We'll need to wait for mypy 1.7, which should include these aliases in mypy's vendored version of typeshed.

Oh my bad, I thought the stubs were updated as well. draft + DO NOT MERGE it is in the mean time. It'll probably accumulate a couple merge conflicts but that's fine they'll be easy to fix.

Akuli:
I don't think we should apply `ConvertibleToInt` blindly with anything converts things to `int`. Because `ConvertibleToInt` includes `str`, using it means that users won't get an error if they accidentally pass a string instead of a number.

AlexWaygood:
Mypy 1.7 is out now, but I'd still like to wait another week or so before we consider this, to give people time to upgrade to the latest mypy -- being overly hasty with things like this has caused issues in the past. Agree with @Akuli that it might also be good to be opinionated in some places, and continue using `SupportsInt` -- we should think about this carefully

Avasam:
There's a couple places where it's a simple existing union replacement. Splitting it up might make it easier to review intent of other `SupportsInt` changes?

AlexWaygood:
> There's a couple places where it's a simple existing union replacement. Splitting it up might make it easier to review intent of other `SupportsInt` changes?

that'd be great, yeah :)

stubs/PyAutoGUI/pyautogui/__init__.pyi
```diff
@@ -70,10 +74,10 @@ def size() -> Size: ...
 
 resolution = size
 
-def onScreen(x: _NormalizeableXArg | None, y: SupportsInt | None = None) -> bool: ...
+def onScreen(x: _NormalizeableXArg | None, y: ConvertibleToInt | None = None) -> bool: ...
```
AlexWaygood, line 77: Are there any examples in the documentation where `y` is called with a `str` argument? Or, is there any particular reason that you know of to pass a `str` to the `y` parameter?

It's true that the function at runtime calls `int()` on the argument passed to `y` before doing anything with it, so if you pass `"10"` to the `y` parameter, it _will work_. But what's the reason for doing so? `SupportsInt` isn't a perfect type hint here, but it ensures that the type checker will still emit an error if a user accidentally tries to pass `"foo"` to the `y` parameter -- and with this change, it wouldn't.

_In general_, typeshed tries to put more weight on avoiding false positives than on avoiding false negatives. But in a situation like this, where (as far as I can see) there's no real use case for passing a `str`, I think we can and should be more opinionated than the function at runtime.

stubs/keyboard/keyboard/mouse.pyi
```diff
@@ -42,7 +43,7 @@ def click(button: _MouseButton = "left") -> None: ...
 def double_click(button: _MouseButton = "left") -> None: ...
 def right_click() -> None: ...
 def wheel(delta: int = 1) -> None: ...
-def move(x: SupportsInt, y: SupportsInt, absolute: bool = True, duration: float = 0) -> None: ...
+def move(x: ConvertibleToInt, y: ConvertibleToInt, absolute: bool = True, duration: float = 0) -> None: ...
```
AlexWaygood, line 46: Same question here: what's the reason why a user might want to pass a `str` to this function?

stubs/netaddr/netaddr/eui/__init__.pyi
```diff
@@ -40,7 +41,7 @@ class EUI(BaseIdentifier):
     @property
     def value(self) -> int: ...
     @value.setter
-    def value(self, value: str | SupportsInt | SupportsIndex) -> None: ...
+    def value(self, value: ConvertibleToInt) -> None: ...
```
AlexWaygood, line 44: This seems like a good change, since the `.value` property is explicitly there in order to transform `str`, `bytes` objects etc. into `int` before they're stored on the `EUI` instance

stubs/netaddr/netaddr/ip/__init__.pyi
```diff
@@ -71,9 +71,9 @@ class IPAddress(BaseIP):
     def ipv4(self) -> Self: ...
     def ipv6(self, ipv4_compatible: bool = False) -> Self: ...
     def format(self, dialect: type[ipv6_verbose] | None = None) -> str: ...
-    def __or__(self, other: str | SupportsInt | SupportsIndex) -> Self: ...
-    def __and__(self, other: str | SupportsInt | SupportsIndex) -> Self: ...
-    def __xor__(self, other: str | SupportsInt | SupportsIndex) -> Self: ...
+    def __or__(self, other: ConvertibleToInt) -> Self: ...
+    def __and__(self, other: ConvertibleToInt) -> Self: ...
+    def __xor__(self, other: ConvertibleToInt) -> Self: ...
```
AlexWaygood, line 76: I would actually be inclined to make these stricter, rather than less strict. The [docstrings](https://github.com/netaddr/netaddr/blob/e212d8e5e67b1f1e09b4fc7833331e320cea622f/netaddr/ip/__init__.py#L610-L635) say:

```
:param other: An `IPAddress` object (or other int-like object).
```

To me, a `str` doesn't count as an int-like object, and I can't see a great reason to pass a `str` in here.

```suggestion
    def __or__(self, other: SupportsInt | SupportsIndex) -> Self: ...
    def __and__(self, other: SupportsInt | SupportsIndex) -> Self: ...
    def __xor__(self, other: SupportsInt | SupportsIndex) -> Self: ...
```

stubs/netaddr/netaddr/ip/__init__.pyi
```diff
@@ -148,14 +148,14 @@ class IPRange(BaseIP, IPListMixin):
     def cidrs(self) -> list[IPNetwork]: ...
 
 def iter_unique_ips(*args: IPRange | _IPNetworkAddr) -> Iterator[IPAddress]: ...
-def cidr_abbrev_to_verbose(abbrev_cidr: str | SupportsInt | SupportsIndex) -> str: ...
+def cidr_abbrev_to_verbose(abbrev_cidr: ConvertibleToInt) -> str: ...
```
AlexWaygood, line 151: This change seems reasonable to me; the function at runtime is deliberately very permissive in what it accepts

stubs/netaddr/netaddr/ip/__init__.pyi
```diff
@@ -148,14 +148,14 @@ class IPRange(BaseIP, IPListMixin):
...
 def cidr_merge(ip_addrs: Iterable[IPRange | _IPNetworkAddr]) -> list[IPNetwork]: ...
 def cidr_exclude(target: _IPNetworkAddr, exclude: _IPNetworkAddr) -> list[IPNetwork]: ...
 def cidr_partition(
     target: _IPNetworkAddr, exclude: _IPNetworkAddr
 ) -> tuple[list[IPNetwork], list[IPNetwork], list[IPNetwork]]: ...
 def spanning_cidr(ip_addrs: Iterable[_IPNetworkAddr]) -> IPNetwork: ...
-def iter_iprange(start: _IPAddressAddr, end: _IPAddressAddr, step: SupportsInt | SupportsIndex = 1) -> Iterator[IPAddress]: ...
+def iter_iprange(start: _IPAddressAddr, end: _IPAddressAddr, step: ConvertibleToInt = 1) -> Iterator[IPAddress]: ...
```
AlexWaygood, line 158: I don't see any reason to make this change: I think `SupportsInt | SupportsIndex` expresses the idea of "int-like objects" better than `ConvertibleToInt` -- and I don't think it makes sense to pass something that isn't an "int-like object" to the `step` parameter here

stubs/python-dateutil/dateutil/relativedelta.pyi
```diff
@@ -87,11 +88,11 @@ class relativedelta:
...
-    def __mul__(self, other: SupportsFloat) -> Self: ...
-    def __rmul__(self, other: SupportsFloat) -> Self: ...
+    def __mul__(self, other: ConvertibleToFloat) -> Self: ...
+    def __rmul__(self, other: ConvertibleToFloat) -> Self: ...
     def __eq__(self, other: object) -> bool: ...
     def __ne__(self, other: object) -> bool: ...
-    def __div__(self, other: SupportsFloat) -> Self: ...
-    def __truediv__(self, other: SupportsFloat) -> Self: ...
+    def __div__(self, other: ConvertibleToFloat) -> Self: ...
+    def __truediv__(self, other: ConvertibleToFloat) -> Self: ...
```
AlexWaygood, line 96: While technically it would _work_, I don't think there would ever be a good reason to try to multiply or divide a `relativedelta` object by a string `"10"` (for example), would there? I think the current annotation is better here.

stubs/PyAutoGUI/pyautogui/__init__.pyi
Avasam: I don't think so. I'll revert here. How about the normalizable X arg (see `_NormalizeableXArg` aslias)? It can be a `str`, but `str` there means "filename on screen"

stubs/keyboard/keyboard/mouse.pyi
Avasam: Unless some API give screen coordinates or relative positions in `str`. I'm not aware of any screen/window API that do that.

stubs/PyAutoGUI/pyautogui/__init__.pyi
AlexWaygood: Yeah, I'm fine with that one!

AlexWaygood (approved):
Thanks!

PR #11005, opened 2023-11-10
CONTRIBUTING: Add @deprecated, other updates

### LINKED ISSUE #11002: How should we add `@deprecated` to objects in the standard library?
If an object is deprecated in Python 3.12, to be removed in Python 3.14, should we decorate it like this? (1)

```py
@deprecated("Will be removed in Python 3.14")
class Foo: ...
```

Like this? (2)

```py
@deprecated("Deprecated in Python 3.12; will be removed in Python 3.14")
class Foo: ...
```

Or like this? (3)

```py
    @deprecated("Will be removed in Python 3.14")
    class Foo: ...
else:
    class Foo: ...
```

For somebody who uses Python 3.8, it might be surprising if a type checker started emitting warnings about an object being deprecated if the runtime only emits a `DeprecationWarning` on Python 3.12+. For that user, Python 3.14 will be a long way away; they might not even be _able_ to switch to the newer way of doing things yet (perhaps the "replacement idiom" only exists on Python 3.9+). For us in typeshed, however, (3) is obviously the most tedious option for us to maintain, as it will mean more branches.

Thoughts?

### LINKED ISSUE #11004: How to use _typeshed.wsgi with typing.get_type_hints
Hello there,
is there a way to get to work combination of `_typeshed.wsgi` and `typing.get_type_hints` in runtime?
In flask codebase there is something like
```py

import typing as t

if t.TYPE_CHECKING:  # pragma: no cover
    from _typeshed.wsgi import WSGIApplication  # noqa: F401
    from werkzeug.datastructures import Headers  # noqa: F401
    from werkzeug.sansio.response import Response  # noqa: F401

# The possible types that are directly convertible or are a Response object.
ResponseValue = t.Union[
    "Response",
    str,
    bytes,
    t.List[t.Any],
    # Only dict is actually accepted, but Mapping allows for TypedDict.
    t.Mapping[str, t.Any],
    t.Iterator[str],
    t.Iterator[bytes],
]

# the possible types for an individual HTTP header
# This should be a Union, but mypy doesn't pass unless it's a TypeVar.
HeaderValue = t.Union[str, t.List[str], t.Tuple[str, ...]]

# the possible types for HTTP headers
HeadersValue = t.Union[
    "Headers",
    t.Mapping[str, HeaderValue],
    t.Sequence[t.Tuple[str, HeaderValue]],
]

# The possible types returned by a route function.
ResponseReturnValue = t.Union[
    ResponseValue,
    t.Tuple[ResponseValue, HeadersValue],
    t.Tuple[ResponseValue, int],
    t.Tuple[ResponseValue, int, HeadersValue],
    "WSGIApplication",
]
```
https://github.com/pallets/flask/blob/3.0.0/src/flask/typing.py
and in my application I use something like
```py
from flask.typing import ResponseReturnValue

@inject
def my_endpoint() -> ResponseReturnValue:
    return "asdf"

...
def inject(method):
    ...
    arg_types = typing.get_type_hints(method)
    # get args from DI container and call actual function with injected args
```
which ends with error like
```
NameError: name 'Response' is not defined
```
Here is complete playground for my issue https://www.online-python.com/AsZX867MEO

I can probably use my own type for that - that's why I ask here - but not sure how to use typeshed in runtime.
Then I can probably send PR directly to flask with fix for such edge cases (as nobody probably encountered this based on how long this is in flask already).

Thanks for you tips and help in advice.

JelleZijlstra:
Fixes #11002.

I started out just documenting conventions with `@deprecated`, but
found some other things in need of updates.

- PEP 585 and 612 are now fully supported
- Add several other PEPs that we fully support
- Use updated links to PEPs
- Split off a few sections from the long "Conventions" section
- PEP 688 was accepted

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -285,21 +285,21 @@ Accepted features that *cannot* yet be used in typeshed include:
   use argument names prefixed with `__` instead
 
 The following features are partially supported:
-- [PEP 585](https://www.python.org/dev/peps/pep-0585/) (builtin
-  generics): see [#4820](https://github.com/python/typeshed/issues/4820),
-  mostly supported but bugs remain for a few specific cases
-- [PEP 612](https://www.python.org/dev/peps/pep-0612/) (ParamSpec):
-  see [#4827](https://github.com/python/typeshed/issues/4827),
-  supported in some contexts but requires `# type: ignore` comments
 
 Supported features include:
-- [PEP 544](https://www.python.org/dev/peps/pep-0544/) (Protocol)
-- [PEP 586](https://www.python.org/dev/peps/pep-0586/) (Literal)
-- [PEP 591](https://www.python.org/dev/peps/pep-0591/) (Final/@final)
-- [PEP 589](https://www.python.org/dev/peps/pep-0589/) (TypedDict)
-- [PEP 604](https://www.python.org/dev/peps/pep-0604/) (`Foo | Bar` union syntax)
-- [PEP 647](https://www.python.org/dev/peps/pep-0647/) (TypeGuard):
+- [PEP 544](https://peps.python.org/pep-0544/) (Protocol)
+- [PEP 585](https://peps.python.org/pep-0585/) (builtin generics)
+- [PEP 586](https://peps.python.org/pep-0586/) (Literal)
+- [PEP 591](https://peps.python.org/pep-0591/) (Final/@final)
+- [PEP 589](https://peps.python.org/pep-0589/) (TypedDict)
+- [PEP 604](https://peps.python.org/pep-0604/) (`Foo | Bar` union syntax)
+- [PEP 612](https://peps.python.org/pep-0612/) (ParamSpec)
+- [PEP 647](https://peps.python.org/pep-0647/) (TypeGuard):
   see [#5406](https://github.com/python/typeshed/issues/5406)
+- [PEP 655](https://peps.python.org/pep-0655/) (`Required` and `NotRequired`)
+- [PEP 673](https://peps.python.org/pep-0673/) (`Self`)
+- [PEP 675](https://peps.python.org/pep-0675/) (`LiteralString`)
+- [PEP 702](https://peps.python.org/pep-0702/) (`@deprecated()`)
 
 Features from the `typing` module that are not present in all
 supported Python 3 versions must be imported from `typing_extensions`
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -512,6 +520,22 @@ into any of those categories, use your best judgement.
...
 
+### `@deprecated`
+
+Typeshed uses the `@typing_extensions.deprecated` decorator
+(`@warnings.deprecated` since Python 3.13) to mark deprecated
+functionality; see [PEP 702](https://peps.python.org/pep-0702/).
+
+A few guidelines for how to use it:
+
+* In the standard library, apply the decorator only in Python versions
```
hauntsaninja, line 531: Commented in the issue, but I think it would be nice to be more aggressive in cases where an appropriate replacement already exists

srittau (changes requested):
The header "The following features are partially supported:" could be removed, since the list was removed.

JelleZijlstra:
@srittau I'm actually going to move PEP 702 into that section as there are some bugs that mean we can't use it in certain contexts yet.

CONTRIBUTING.md
JelleZijlstra: I agree now, reworded the section to say we should apply it to versions where a good replacement exists.

srittau (approved):

AlexWaygood (approved):
This is great, thanks!

CONTRIBUTING.md
```diff
@@ -285,21 +285,25 @@ Accepted features that *cannot* yet be used in typeshed include:
...
 The following features are partially supported:
-- [PEP 585](https://www.python.org/dev/peps/pep-0585/) (builtin
-  generics): see [#4820](https://github.com/python/typeshed/issues/4820),
-  mostly supported but bugs remain for a few specific cases
-- [PEP 612](https://www.python.org/dev/peps/pep-0612/) (ParamSpec):
-  see [#4827](https://github.com/python/typeshed/issues/4827),
-  supported in some contexts but requires `# type: ignore` comments
+- [PEP 702](https://peps.python.org/pep-0702/) (`@deprecated()`)
+  - For now, cannot be used in combination with other decorators
+    (e.g., `@overload` and `@property`) due to bugs in
```
JelleZijlstra, line 290:
```suggestion
    (e.g., `@overload`, `@classmethod`, and `@property`) due to bugs in
```

AlexWaygood:
I'm guessing pytype just doesn't like have two decorators stacked anywhere in a stub. I'm guessing it also crashes if you try to deprecate an abstractmethod

JelleZijlstra:
Probably, I guess it's not worth listing exactly which decorators are problematic.

PR #11009, opened 2023-11-10
stdlib: more deprecations

JelleZijlstra:
I went through the list in https://docs.python.org/3.12/whatsnew/3.12.html#deprecated and added `@deprecated` in all cases where it seemed practical. I hope this will give more useful input for #11002.

I'll comment below on the individual deprecations.


stdlib/argparse.pyi
```diff
@@ -337,11 +337,23 @@ class Action(_AttributeHolder):
...
             option_strings: Sequence[str],
             dest: str,
-            default: _T | str | None = None,
+            default: bool | None = None,
+            *,
+            required: bool = False,
+            help: str | None = None,
+        ) -> None: ...
+        @overload
+        @deprecated("The `type`, `choices`, and `metavar` parameters are ignored and will be removed in Python 3.14.")
```
JelleZijlstra, line 351: These apparently never did anything useful, so it seems fair to emit the warning on all versions.

stdlib/argparse.pyi
```diff
@@ -337,11 +337,23 @@ class Action(_AttributeHolder):
 
 if sys.version_info >= (3, 12):
     class BooleanOptionalAction(Action):
+        @overload
         def __init__(
             self,
             option_strings: Sequence[str],
             dest: str,
-            default: _T | str | None = None,
+            default: bool | None = None,
```
JelleZijlstra, line 345: It appears that a non-bool default doesn't really make sense on this action.

stdlib/asyncio/events.pyi
```diff
@@ -599,8 +599,10 @@ class AbstractEventLoopPolicy:
     def new_event_loop(self) -> AbstractEventLoop: ...
     # Child processes handling (Unix only).
     @abstractmethod
+    @deprecated("Deprecated as of Python 3.12; will be removed in Python 3.14")
```
JelleZijlstra, line 602: Wasn't sure about these child watchers as it's not clear to me what if anything they should be replaced with. Possibly we should emit the warning on 3.12+ only.

stdlib/builtins.pyi
```diff
@@ -937,6 +938,8 @@ class bool(int):
     @overload
     def __rxor__(self, __value: int) -> int: ...
     def __getnewargs__(self) -> tuple[int]: ...
+    @deprecated("Will throw an error in Python 3.14. Use `not` for logical negation of bools instead.")
```
JelleZijlstra, line 941: This is `~True`. It always behaved confusingly, so it seems fine to warn about it on all versions.

stdlib/datetime.pyi
```diff
@@ -265,6 +265,7 @@ class datetime(date):
         def fromtimestamp(cls, __timestamp: float, tz: _TzInfo | None = ...) -> Self: ...
 
     @classmethod
+    @deprecated("Use timezone-aware objects to represent datetimes in UTC; e.g. by calling .fromtimestamp(datetime.UTC)")
```
JelleZijlstra, line 268: This and utcnow() were always mildly discouraged and the replacement has always been available, so chose to deprecate across versions.

stdlib/email/utils.pyi
```diff
@@ -54,6 +54,10 @@ def formatdate(timeval: float | None = None, localtime: bool = False, usegmt: bo
 def format_datetime(dt: datetime.datetime, usegmt: bool = False) -> str: ...
 
 if sys.version_info >= (3, 12):
+    @overload
+    def localtime(dt: datetime.datetime | None = None) -> datetime.datetime: ...
+    @overload
+    @deprecated("The `isdst` parameter does nothing and will be removed in Python 3.14.")
```
JelleZijlstra, line 60: The `isdst` parameter did *something* before 3.12; didn't study the function enough to determine whether it did anything useful.

stdlib/os/__init__.pyi
```diff
@@ -346,8 +346,13 @@ class stat_result(structseq[float], tuple[int, int, int, int, int, int, int, flo
     @property
     def st_mtime(self) -> float: ...  # time of most recent content modification,
     # platform dependent (time of most recent metadata change on Unix, or the time of creation on Windows)
-    @property
-    def st_ctime(self) -> float: ...
+    if sys.version_info >= (3, 12) and sys.platform == "win32":
```
JelleZijlstra, line 349: The replacement `st_birthtime` is only available on 3.12+, so it doesn't make sense to warn on earlier versions.

stdlib/pkgutil.pyi
```diff
@@ -35,8 +36,10 @@ if sys.version_info < (3, 12):
     class ImpLoader:
         def __init__(self, fullname: str, file: IO[str], filename: str, etc: tuple[str, str, int]) -> None: ...
 
+@deprecated("Use importlib.util.find_spec() instead. Will be removed in Python 3.14.")
```
JelleZijlstra, line 39: `find_spec()` is apparently available on all supported versions.

stdlib/shutil.pyi
```diff
@@ -95,21 +95,28 @@ class _RmtreeType(Protocol):
...
         @overload
+        def __call__(
+            self,
+            path: StrOrBytesPath,
+            ignore_errors: bool = False,
+            *,
+            dir_fd: int | None = None,
+        ) -> None: ...
+        @overload
+        @deprecated("The `onerror` parameter is deprecated and will be removed in Python 3.14. Use `onexc` instead.")
```
JelleZijlstra, line 100: Since the new parameter is only available on 3.12+, applied the warning on 3.12+ only.

stdlib/tarfile.pyi
```diff
@@ -292,22 +292,45 @@ class TarFile:
     def list(self, verbose: bool = True, *, members: _list[TarInfo] | None = None) -> None: ...
     def next(self) -> TarInfo | None: ...
     if sys.version_info >= (3, 8):
+        @overload
+        @deprecated("Extracting tar archives without specifying `filter` is deprecated until Python 3.14, when 'data' filter will become the default.")
```
JelleZijlstra: This was backported all the way back, so seems fair to also emit the warning.

stdlib/types.pyi
```diff
@@ -144,8 +144,13 @@ class CodeType:
     def co_name(self) -> str: ...
     @property
     def co_firstlineno(self) -> int: ...
-    @property
-    def co_lnotab(self) -> bytes: ...
+    if sys.version_info >= (3, 10):
+        @property
+        @deprecated("Will be removed in Python 3.14. Use the co_lines() method instead.")
```
JelleZijlstra, line 149: co_lines() is only in 3.10+.

stdlib/xml/etree/ElementTree.pyi
```diff
@@ -124,6 +124,8 @@ class Element:
     def __setitem__(self, __key: SupportsIndex, __value: Element) -> None: ...
     @overload
     def __setitem__(self, __key: slice, __value: Iterable[Element]) -> None: ...
+    @deprecated("Testing an element's truth value is deprecated.")
```
JelleZijlstra, line 127: This always behaved confusingly (and raised a FutureWarning under some circumstances), so seems fine to always warn.

JelleZijlstra:
Seems like I managed to crash stubtest, I'll have to debug

hauntsaninja:
stubtest doesn't have great decorator support, it just hardcodes a few decorators

AlexWaygood:
> Seems like I managed to crash stubtest, I'll have to debug

Same traceback as https://github.com/python/mypy/issues/14950

AlexWaygood:
Looks like deprecating a property also breaks pytype :/

stdlib/asyncio/events.pyi
AlexWaygood: I haven't looked at this in detail, but in situations like this where we're not sure, I feel like we should probably err on the side of caution and only emit warnings on py312+

stdlib/datetime.pyi
```diff
@@ -278,6 +279,7 @@ class datetime(date):
         def now(cls, tz: _TzInfo) -> datetime: ...
 
     @classmethod
+    @deprecated("Use timezone-aware objects to represent datetimes in UTC; e.g. by calling .now(datetime.UTC)")
```
AlexWaygood, line 282: Not sure about this one. It may have always been discouraged to use this method, but I think the deprecation has been pretty disruptive — see https://discuss.python.org/t/deprecating-utcnow-and-utcfromtimestamp/26221/12?u=alexwaygood. For Python 3.8 users, the method is still going to be around for a while; do we need to bother them with this warning just yet? (On the other hand: if the method is widely used, perhaps it's all the more important to make sure the word gets out that it's deprecated...)

AlexWaygood (commented):
The substantive changes look pretty good to me; just have some qualms about `datetime` and `asyncio`

stdlib/datetime.pyi
hauntsaninja: I think it's worth it. Work had a bug just this week because someone was using these.

(Also I think it will be around for like eight years before CPython removes it, I'd definitely oppose removing it in 3.14)

stdlib/asyncio/events.pyi
JelleZijlstra: Changed this to warn only on 3.12+

AlexWaygood (approved):
LGTM. The ones that combine `@overload` with `@deprecated` will probably have to wait until mypy 1.8 is released, though, due to the stubtest crashes

JelleZijlstra:
Ah right, forgot we already fixed that. There is also at least one other pytype crash. I might open another PR with the simpler ones and keep this one open to merge later.

AlexWaygood:
This is no longer deferred; stubtest no longer crashes on this with the latest mypy. There are still a few "regular" stubtest failures, however

Issue #8708, opened 2022-09-08
PEP 646 (Variadic Generics) tracker


srittau:
* [x] mypy (https://github.com/python/mypy/issues/12280)
* [ ] pytype (https://github.com/google/pytype/issues/1525)
* [x] pyright
* [x] pyre
* [ ] PyCharm (nice to have, but not required)

AlexWaygood:
I've ticked off pyright; I believe it's supported PEP 646 for a while now.

AlexWaygood:
PEP 646 lists Pyre as [providing a reference implementation of the PEP](https://peps.python.org/pep-0646/#reference-implementation), so they probably also have pretty good support.

AlexWaygood:
PEP 646 support would be useful for (feel free to edit this):

- `asyncio.events.Handle`
- `asyncio.events.TimerHandle`
- `asyncio.events.AbstractEventLoop.call_at`
- `asyncio.events.AbstractEventLoop.call_later`
- `asyncio.events.AbstractEventLoop.call_soon`
- `asyncio.base_events.BaseEventLoop.call_at`
- `asyncio.base_events.BaseEventLoop.call_later`
- `asyncio.base_events.BaseEventLoop.call_soon`
- (Maybe): `functools.partial` (#8703)
- Various functions in the `keyboard` third-party stubs:
  - https://github.com/python/typeshed/blob/8a8db9c6d4e39b4c530b6b6ce03bb6869c394419/stubs/keyboard/keyboard/mouse.pyi#L47-L56
  - https://github.com/python/typeshed/blob/8a8db9c6d4e39b4c530b6b6ce03bb6869c394419/stubs/keyboard/keyboard/__init__.pyi#L37
  - https://github.com/python/typeshed/blob/8a8db9c6d4e39b4c530b6b6ce03bb6869c394419/stubs/keyboard/keyboard/__init__.pyi#L58-L65

Note that we won't be able to use PEP 646 as it currently stands for `builtins.map`, `builtins.zip` or `asyncio.gather`: https://github.com/python/cpython/pull/32103#discussion_r836917997

srittau:
This is now used in several places.

PR #11010, opened 2023-11-10
Bump mypy to 1.7.0

AlexWaygood:
This mypy release means that mypy now supports PEP-646 and PEP-692 (`TypeVarTuple` and `Unpack`) by default 🥳

X-ref:
- #8708
- #9710

### DIFF
--- a/test_cases/stdlib/check_dataclasses.py
+++ b/test_cases/stdlib/check_dataclasses.py
@@ -87,9 +87,4 @@ def check_other_isdataclass_overloads(x: type, y: object) -> None:
         assert_type(dc.fields(y), Tuple[dc.Field[Any], ...])
         assert_type(dc.asdict(y), Dict[str, Any])
         assert_type(dc.astuple(y), Tuple[Any, ...])
-
-        # No longer passes with mypy 1.5.0
-        # now that mypy gives a tailored signature for dataclasses.replace()
-        # (https://github.com/python/mypy/issues/15843):
-        #
-        # dc.replace(y)
+        dc.replace(y)
--- a/requirements-tests.txt
+++ b/requirements-tests.txt
@@ -6,7 +6,7 @@ flake8==6.1.0            # must match .pre-commit-config.yaml
 flake8-bugbear==23.9.16  # must match .pre-commit-config.yaml
 flake8-noqa==1.3.2       # must match .pre-commit-config.yaml
 flake8-pyi==23.11.0      # must match .pre-commit-config.yaml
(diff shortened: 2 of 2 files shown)


test_cases/stdlib/check_dataclasses.py
```diff
@@ -87,9 +87,4 @@ def check_other_isdataclass_overloads(x: type, y: object) -> None:
         assert_type(dc.fields(y), Tuple[dc.Field[Any], ...])
         assert_type(dc.asdict(y), Dict[str, Any])
         assert_type(dc.astuple(y), Tuple[Any, ...])
-
-        # No longer passes with mypy 1.5.0
-        # now that mypy gives a tailored signature for dataclasses.replace()
-        # (https://github.com/python/mypy/issues/15843):
-        #
-        # dc.replace(y)
+        dc.replace(y)
```
AlexWaygood, line 90: This passes mypy again, following https://github.com/python/mypy/pull/15962 (thanks @ikonst!)

JelleZijlstra (approved):

jessemyers-lettuce:
FYI, this release "regressed" `pydantic` usage. This snippet

```py
from pydantic import BaseModel

class Foo(BaseModel):
    pass
```

Now raises:

```
test.py:3: error: Explicit "Any" is not allowed  [misc]
    class Foo(BaseModel):
    ^~~~~~~~~~~~~~~~~~~~
test.py:3: error: Type of decorated function contains type "Any"
("Callable[[DefaultNamedArg(dict[str, Any] | None, '__pydantic_extra__'), DefaultNamedArg(set[str], '__pydantic_fields_set__'), DefaultNamedArg(dict[str, Any] | None, '__pydantic_private__')], None]")  [misc]
    class Foo(BaseModel):
    ^~~~~~~~~~~~~~~~~~~~
```

I'd guess that this behavior is correct... there is, in fact, an explicit `Any` in the `BaseModel` type definition. But it's going to be very annoying to have to explicitly ignore that everywhere.

AlexWaygood:
@jessemyers-lettuce are you sure you're commenting on the right issue? This change was a purely internal change, bumping the version of mypy we install for running typeshed's tests in our CI. Maybe you meant to comment on https://github.com/python/mypy/issues/16341?

JelleZijlstra:
The behavior you're describing should be fixed in either mypy or pydantic; typeshed has nothing to do with it.

jessemyers-lettuce:
@AlexWaygood, apologies, you're right. I misread my test report.

PR #11011, opened 2023-11-10
Use @deprecated in tkinter

Akuli:
`@deprecated` didn't exist when I stubbed these methods.


stdlib/tkinter/__init__.pyi
```diff
@@ -344,6 +347,7 @@ class Misc:
     def tk_focusNext(self) -> Misc | None: ...
     def tk_focusPrev(self) -> Misc | None: ...
     @overload
+    @deprecated(".after(timeout) without a callback function freezes the GUI app, use .after(timeout, func) instead")
```
Akuli, line 350: This would be a better use case for `@always_error` (or whatever the name would be) than `@deprecated`, but IMO `@deprecated` is much better than nothing here.

stdlib/tkinter/__init__.pyi
```diff
@@ -273,10 +273,13 @@ class Variable:
     def trace_add(self, mode: _TraceMode, callback: Callable[[str, str, str], object]) -> str: ...
     def trace_remove(self, mode: _TraceMode, cbname: str) -> None: ...
     def trace_info(self) -> list[tuple[tuple[_TraceMode, ...], str]]: ...
-    def trace_variable(self, mode, callback): ...  # deprecated
-    def trace_vdelete(self, mode, cbname) -> None: ...  # deprecated
-    def trace_vinfo(self): ...  # deprecated
-    trace = trace_variable  # deprecated
+    @deprecated("use trace_add() instead of trace_variable()")
```
JelleZijlstra, line 276: Is this scheduled for removal in a specific version? If so, let's add "will be removed in Python 3.14" (or whatever other version)

Akuli: There is no associated Python version. [Documentation for the Tcl commands](https://www.tcl.tk/man/tcl/TclCmd/trace.html) that these methods wrap are being deprecated doesn't specify a version either, it only says:

> These subcommands are deprecated and will likely be removed in a future version of Tcl.

JelleZijlstra: I see, makes sense.

JelleZijlstra:
I think stubtest crashes on deprecated overloads. I can look into fixing that but probably not until next week.

stdlib/tkinter/__init__.pyi
AlexWaygood: This is the only one that's crashing stubtest and pytype (I think) -- you could just defer this one for now

AlexWaygood (approved):
LGTM (though I haven't checked that these are in fact deprecated -- I trust you there ;)

Akuli:
I wasn't sure if `trace = trace_variable` makes `trace()` deprecated when `trace_variable()` is deprecated, so I added a `def trace(...)` instead. This also lets me specify a better error message that actually mentions `trace()`.

Deprecations don't seem to actually work yet? I get:

```
(env) akuli@Akuli-ThinkPad:~/typeshed$ pip install --upgrade mypy
Requirement already satisfied: mypy in ./env/lib/python3.11/site-packages (1.7.0)
Requirement already satisfied: typing-extensions>=4.1.0 in ./env/lib/python3.11/site-packages (from mypy) (4.7.1)
Requirement already satisfied: mypy-extensions>=1.0.0 in ./env/lib/python3.11/site-packages (from mypy) (1.0.0)

(env) akuli@Akuli-ThinkPad:~/typeshed$ cat a.py
import tkinter
tkinter.Variable().trace('w', print)
tkinter.Variable().trace_variable('w', print)

(env) akuli@Akuli-ThinkPad:~/typeshed$ mypy --custom-typeshed-dir . a.py
```

AlexWaygood:
Yes, mypy hasn't implemented `@deprecated` yet. Possibly only pyright has — not sure about pyre/pytype.

AlexWaygood:
Nice palindromic PR number FYI — glad it didn't go to stubsabot

PR #11012, opened 2023-11-10
Fix type hints for `unittest.util.unorderable_list_difference()`

RJPercival:
It claimed that the arguments were `Sequence` types (immutable), but the function removes elements from them, so they must be `MutableSequence` types. Specifically, it calls `pop()` on the first argument and `remove()` on the second argument.

See [the function implementation](https://github.com/python/cpython/blob/0b06d2482d77e02c5d40e221f6046c9c355458b2/Lib/unittest/util.py#L98-L113) for details.


AlexWaygood (approved):
Thanks! This will make the function less ergonomic for users (since `MutableSequence` is generic over an invariant TypeVar, whereas `Sequence` is generic over a covariant TypeVar). However, I think type safety trumps ergonomics in this case (especially since I can find hardly any actual uses of this function on grep.app)

PR #11013, opened 2023-11-10
tkinter: Require passing in a callback function to `.after()` 

Akuli:
Tkinter's `.after()` wraps Tcl's `after` command, which can be used in two ways: either as a blocking delay (like `time.sleep()` in Python), or to run a callback function in Tcl's event loop later (like `window.setTimeout` in javascript).

The blocking variant is useless in Python. It freezes your tkinter app and is equivalent to `time.sleep()` anyway.

I have seen other tkinter programmers using it twice, once by a beginner and once by an experienced tkinter user. In both cases, it was an accident and they forgot to pass in a callback function.


stdlib/tkinter/__init__.pyi
```diff
@@ -347,9 +347,8 @@ class Misc:
     def tk_focusFollowsMouse(self) -> None: ...
     def tk_focusNext(self) -> Misc | None: ...
     def tk_focusPrev(self) -> Misc | None: ...
-    @overload
-    def after(self, ms: int, func: None = None) -> None: ...
-    @overload
+    # .after() can be called without the "func" argument, but it is basically never what you want.
+    # It behaves like time.sleep() and freezes the GUI app.
     def after(self, ms: int | Literal["idle"], func: Callable[..., object], *args: Any) -> str: ...
```
AlexWaygood, line 352: Presumably the `args` are passed to `func`? If so, we might be able to use PEP-646 now that mypy 1.7 is released (you could give it a try!)

```suggestion
    def after(self, ms: int | Literal["idle"], func: Callable[Unpack[Ts], object], *args: Unpack[Ts]) -> str: ...
```

(With `_Ts = TypeVarTuple("_Ts")` somewhere higher up in this file, and both `TypeVarTuple`/`Unpack` imported from `typing_extensions`)

AlexWaygood: This can also be deferred to a future PR, though

Akuli: I tried it, and it worked (with a small tweak: `Callable[[Unpack[Ts]], object]`). I remember trying to use `ParamSpec` in the past, but it didn't work because you need to use both `*args` and `**kwargs` with it.

AlexWaygood: Yes, thanks! I forgot that you need to unpack it "inside of the parameter list". It's annoying that it works differently to `ParamSpec` in that way (though I understand the reason why)

AlexWaygood:
okay, looks like pytype crashes if we use `Unpack` :(

Akuli:
I'll make a separate PR of the `Unpack` stuff because it confuses pyright.

AlexWaygood (approved):

PR #11015, opened 2023-11-10
Use PEP-646 in stubs for `asyncio`

AlexWaygood:
As with #11014, I think this will crash pytype for now -- but I'm curious to see the mypy_primer output


stdlib/asyncio/events.pyi
```diff
@@ -67,33 +68,37 @@ class _TaskFactory(Protocol):
...
     def __hash__(self) -> int: ...
     def when(self) -> float: ...
-    def __lt__(self, other: TimerHandle) -> bool: ...
-    def __le__(self, other: TimerHandle) -> bool: ...
-    def __gt__(self, other: TimerHandle) -> bool: ...
-    def __ge__(self, other: TimerHandle) -> bool: ...
+    def __lt__(self, other: TimerHandle[Unpack[tuple[Any, ...]]]) -> bool: ...
+    def __le__(self, other: TimerHandle[Unpack[tuple[Any, ...]]]) -> bool: ...
+    def __gt__(self, other: TimerHandle[Unpack[tuple[Any, ...]]]) -> bool: ...
+    def __ge__(self, other: TimerHandle[Unpack[tuple[Any, ...]]]) -> bool: ...
```
AlexWaygood, line 101: not exactly the prettiest type hints in the world, but I think this is the correct way of doing this...

AlexWaygood:
## mypy_primer analysis

### tornado

They're subclassing `AbstractEventLoop`; they'll just have to update their annotations on their subclass, unfortunately.

### websockets

This is arguably a true positive: they're passing a function that expects an `asyncio.Task` as the first argument to `call_soon()`, then they're passing the result of `asyncio.current_task()` as the second object: https://github.com/python-websockets/websockets/blob/ec3bd2ab06278602c1d6018b476699e090036373/src/websockets/legacy/async_timeout.py#L237-L241

According to our stubs, `asyncio.current_task()` could return `None`, so mypy is right to complain here:

https://github.com/python/typeshed/blob/aabeacfdbd6e67092c70898a2b09a51f9cf28b45/stdlib/asyncio/tasks.pyi#L439

### black

I think their annotations are slightly incorrect: see https://github.com/psf/black/pull/4047

### discord

I think their annotations are definitely incorrect. The code in question is here: https://github.com/Rapptz/discord.py/blob/a5d03d4a1e6f4314518f6f55d8b7130ab3f4336e/discord/ext/tasks/__init__.py#L107-L119. `self.future` is annotated as being an `asyncio.Future[None]`, which implies that you should only ever call `self.future.set_result(None)` with `None`, due to our annotations for `asyncio.Future.set_result()`. But they're passing (indirectly, via `loop.call_later`) `self.future.set_result(True)` on lines 114 and 119.

srittau:
@AlexWaygood You marked this as ready for review, but it's still tagged as "deferred". Is the latter an oversight?

AlexWaygood:
> @AlexWaygood You marked this as ready for review, but it's still tagged as "deferred". Is the latter an oversight?

It can't be merged until pytype no longer crashes whenever we use `Unpack`, so in that sense it is indeed deferred. But I think the PR itself is in good condition, so there's nothing stopping you from _reviewing_ it if you'd like to :-)

stdlib/asyncio/events.pyi
```diff
@@ -83,8 +88,8 @@ class TimerHandle(Handle):
     def __init__(
         self,
         when: float,
-        callback: Callable[..., object],
-        args: Sequence[Any],
+        callback: Callable[[Unpack[_Ts]], object],
+        args: tuple[Unpack[_Ts]],
```
srittau, line 92: I'd prefer if we kept `Sequence` here, following the philosophy that false negatives are preferable to false positives. (As lists or other sequences are fine as `args` arguments.)

srittau (changes requested):
Thanks! I really like the additional type safety that `Unpack` brings us. One complaint, though.

stdlib/asyncio/events.pyi
JelleZijlstra: In that case, we can't get type safety though, because only tuples can be heterogeneous in the type system.

AlexWaygood: Unfortunately, I can find a few examples on grep.app of people passing non-tuple sequences to the `args` parameter. People doing things like that will be broken by this change, it's true:
- https://github.com/pulumi/pulumi/blob/273951bb17ef45efeeebfe25be174c6df10c1e08/sdk/python/lib/pulumi/runtime/sync_await.py#L80
- https://github.com/ask/mode/blob/a104009f0c96790b9f6140179b4968da07a38c81/mode/utils/loops.py#L61-L65

I don't see a particularly strong use case for passing a `list` or a `deque` or whatever to `args`, though -- I think it would be fairly trivial, if you _have_ got a non-tuple `Sequence`, to cast it to a `tuple` before passing it to the constructor of `Handle` or `TimerHandle`. And as Jelle says, if we want to use `Sequence` for `args` here, we essentially have to just leave the constructors for `Handle` and `TimerHandle` as they are; you can only use `Unpack` with a `tuple`. If I make this change:

```diff
--- a/stdlib/asyncio/events.pyi
+++ b/stdlib/asyncio/events.pyi
@@ -74,7 +74,7 @@ class Handle:
     def __init__(
         self,
         callback: Callable[[Unpack[_Ts]], object],
-        args: tuple[Unpack[_Ts]],
+        args: Sequence[Unpack[_Ts]],
         loop: AbstractEventLoop,
         context: Context | None = None,
     ) -> None: ...
@@ -89,7 +89,7 @@ class TimerHandle(Handle):
         self,
         when: float,
         callback: Callable[[Unpack[_Ts]], object],
-        args: tuple[Unpack[_Ts]],
+        args: Sequence[Unpack[_Ts]],
         loop: AbstractEventLoop,
         context: Context | None = None,
     ) -> None: ...
```

Then mypy correctly complains:

```
stdlib\asyncio\events.pyi:77: error: Unpack is only valid in a variadic position  [valid-type]
stdlib\asyncio\events.pyi:92: error: Unpack is only valid in a variadic position  [valid-type]
```

srittau: > In that case, we can't get type safety though, because only tuples can be heterogeneous in the type system.

Yes, that's unfortunate, but requiring `tuple` here is just incorrect. We've always maintained that we prefer false negatives over false positives, unless there's a strong case for doing so. In this case, I consider passing a `list` as completely reasonable.

AlexWaygood: I feel like the benefits in terms of type safety (which seem to me to be large here) outweigh the risk of false-positive errors in this case. However, since this is the only thing you object to in this PR, I'll remove it for now, so that it can be considered separately in its own PR :-)

srittau (approved):

AlexWaygood:
Oh, I thought #11120 would unblock this, but looks like I got mixed up :)

https://github.com/google/pytype/commit/bbf10c126405c5e1a5dcb91eb903a78fe8c84d5e didn't quite make it into pytype==2023.12.7 -- it'll be included in the next pytype release.

AlexWaygood:
And we're there!

PR #11018, opened 2023-11-11
stdlib: Add a _FilterProtocol type to logging's Filterer.addFilter() filter argument.

### LINKED ISSUE #5700: logging.Logger.addFilter

jenstroeger:
Closes #5700 

As discussed in the related issue, starting with comment https://github.com/python/typeshed/issues/5700#issuecomment-1806718877.

### DIFF
--- a/stdlib/logging/__init__.pyi
+++ b/stdlib/logging/__init__.pyi
@@ -7,7 +7,7 @@ from re import Pattern
 from string import Template
 from time import struct_time
 from types import FrameType, TracebackType
-from typing import Any, ClassVar, Generic, TextIO, TypeVar, overload
+from typing import Any, ClassVar, Generic, Protocol, TextIO, TypeVar, overload
 from typing_extensions import Literal, Self, TypeAlias
 
 if sys.version_info >= (3, 11):
@@ -63,10 +63,13 @@ if sys.version_info >= (3, 11):
 if sys.version_info >= (3, 12):
     __all__ += ["getHandlerByName", "getHandlerNames"]
 
+class _SupportsFilter(Protocol):
+    def filter(self, record: LogRecord) -> bool: ...
+
 _SysExcInfoType: TypeAlias = tuple[type[BaseException], BaseException, TracebackType | None] | tuple[None, None, None]
 _ExcInfoType: TypeAlias = None | bool | _SysExcInfoType | BaseException
 _ArgsType: TypeAlias = tuple[object, ...] | Mapping[str, object]
-_FilterType: TypeAlias = Filter | Callable[[LogRecord], bool]
+_FilterType: TypeAlias = Filter | Callable[[LogRecord], bool] | _SupportsFilter
 _Level: TypeAlias = int | str
 _FormatStyle: TypeAlias = Literal["%", "{", "$"]
 
--- a/stdlib/logging/__init__.pyi
+++ b/stdlib/logging/__init__.pyi
@@ -63,18 +63,22 @@ if sys.version_info >= (3, 11):
 if sys.version_info >= (3, 12):
     __all__ += ["getHandlerByName", "getHandlerNames"]
 
-class _SupportsFilter(Protocol):
-    if sys.version_info >= (3, 12):
-        def filter(self, record: LogRecord) -> bool | LogRecord: ...
-    else:
-        def filter(self, record: LogRecord) -> bool: ...
-
 _SysExcInfoType: TypeAlias = tuple[type[BaseException], BaseException, TracebackType | None] | tuple[None, None, None]
 _ExcInfoType: TypeAlias = None | bool | _SysExcInfoType | BaseException
 _ArgsType: TypeAlias = tuple[object, ...] | Mapping[str, object]
-_FilterType: TypeAlias = _SupportsFilter | Callable[[LogRecord], bool]
 _Level: TypeAlias = int | str
 _FormatStyle: TypeAlias = Literal["%", "{", "$"]
+
+if sys.version_info >= (3, 12):
+    class _SupportsFilter(Protocol):
+        def filter(self, record: LogRecord) -> bool | LogRecord: ...
+
+    _FilterType: TypeAlias = Filter | Callable[[LogRecord], bool | LogRecord] | _SupportsFilter
+else:
+    class _SupportsFilter(Protocol):
+        def filter(self, record: LogRecord) -> bool: ...
+
+    _FilterType: TypeAlias = Filter | Callable[[LogRecord], bool] | _SupportsFilter
 
 raiseExceptions: bool
 logThreads: bool
--- a/stdlib/logging/__init__.pyi
+++ b/stdlib/logging/__init__.pyi
@@ -64,12 +64,15 @@ if sys.version_info >= (3, 12):
     __all__ += ["getHandlerByName", "getHandlerNames"]
 
 class _SupportsFilter(Protocol):
-    def filter(self, record: LogRecord) -> bool: ...
+    if sys.version_info >= (3, 12):
+        def filter(self, record: LogRecord) -> bool | LogRecord: ...
+    else:
+        def filter(self, record: LogRecord) -> bool: ...
 
 _SysExcInfoType: TypeAlias = tuple[type[BaseException], BaseException, TracebackType | None] | tuple[None, None, None]
 _ExcInfoType: TypeAlias = None | bool | _SysExcInfoType | BaseException
 _ArgsType: TypeAlias = tuple[object, ...] | Mapping[str, object]
-_FilterType: TypeAlias = Filter | Callable[[LogRecord], bool] | _SupportsFilter
+_FilterType: TypeAlias = _SupportsFilter | Callable[[LogRecord], bool]
 _Level: TypeAlias = int | str
 _FormatStyle: TypeAlias = Literal["%", "{", "$"]
 
--- a/stdlib/logging/__init__.pyi
+++ b/stdlib/logging/__init__.pyi
@@ -71,12 +71,12 @@ _FormatStyle: TypeAlias = Literal["%", "{", "$"]
 
 if sys.version_info >= (3, 12):
     class _SupportsFilter(Protocol):
-        def filter(self, record: LogRecord) -> bool | LogRecord: ...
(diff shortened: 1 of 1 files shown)


stdlib/logging/__init__.pyi
```diff
@@ -63,10 +63,13 @@ if sys.version_info >= (3, 11):
...
     __all__ += ["getHandlerByName", "getHandlerNames"]
 
+class _FilterProtocol(Protocol):
+    def filter(self, record: LogRecord) -> bool: ...
+
 _SysExcInfoType: TypeAlias = tuple[type[BaseException], BaseException, TracebackType | None] | tuple[None, None, None]
 _ExcInfoType: TypeAlias = None | bool | _SysExcInfoType | BaseException
 _ArgsType: TypeAlias = tuple[object, ...] | Mapping[str, object]
-_FilterType: TypeAlias = Filter | Callable[[LogRecord], bool]
+_FilterType: TypeAlias = Filter | Callable[[LogRecord], bool] | _FilterProtocol
```
Akuli, line 72: Any `Filter` instance has a `.filter()` method, so you can do:

```suggestion
_FilterType: TypeAlias = Callable[[LogRecord], bool] | _FilterProtocol
```

This is IMO slightly simpler / easier to understand.

stdlib/logging/__init__.pyi
```diff
@@ -63,10 +63,13 @@ if sys.version_info >= (3, 11):
 if sys.version_info >= (3, 12):
     __all__ += ["getHandlerByName", "getHandlerNames"]
 
+class _FilterProtocol(Protocol):
```
Akuli, line 66: You could rename this to `_SupportsFilter`. We have various other protocols named `SupportsX` that we use when we need a method named `X`. I'm not sure which way it's more readable though, so I'll let you decide :)

Akuli (commented):
Looks good! A couple nits.

stdlib/logging/__init__.pyi
Akuli: Actually, the explicit `Filter` type is needed after #10540.

stdlib/logging/__init__.pyi
jenstroeger: Ah ok, I wasn’t sure of your naming conventions.

stdlib/logging/__init__.pyi
jenstroeger: Let me know how you’d like me to proceed here — keep or remove `Filter` 🤓

jenstroeger:
Also, I’m not sure about your merge habits (do you squash the PR or not) so I’m just going to keep amending to the original commit so there’s just one commit to merge. Keeps the history clean.

AlexWaygood:
> Also, I’m not sure about your merge habits (do you squash the PR or not) so I’m just going to keep amending to the original commit so there’s just one commit to merge. Keeps the history clean.

We always squash-merge at typeshed :) see our CONTRIBUTING guidelines: https://github.com/python/typeshed/blob/main/CONTRIBUTING.md#submitting-changes

stdlib/logging/__init__.pyi
Akuli: I'd say remove `Filter` (like you already did), but support returning a `LogRecord` object on 3.12+ just like in #10540.

jenstroeger: Commit https://github.com/python/typeshed/pull/11018/commits/55e37101cdefa8e7f69b53144745b3b3c0c16a1a.

But then, shouldn’t the `Callable` be adjusted also to return `bool | LogRecord` on Python 3.12+ ?

jenstroeger: > @Akuli I'd say remove Filter, […]

I guess that broke Sphinx below, although it _does_ inherit from `logging.Filter` [here](https://github.com/sphinx-doc/sphinx/blob/bb74aec2b6aa1179868d83134013450c9ff9d4d6/sphinx/util/logging.py#L473).

stdlib/logging/__init__.pyi
```diff
@@ -63,10 +63,16 @@ if sys.version_info >= (3, 11):
...
+    if sys.version_info >= (3, 12):
+        def filter(self, record: LogRecord) -> bool | LogRecord: ...
+    else:
+        def filter(self, record: LogRecord) -> bool: ...
+
 _SysExcInfoType: TypeAlias = tuple[type[BaseException], BaseException, TracebackType | None] | tuple[None, None, None]
 _ExcInfoType: TypeAlias = None | bool | _SysExcInfoType | BaseException
 _ArgsType: TypeAlias = tuple[object, ...] | Mapping[str, object]
-_FilterType: TypeAlias = Filter | Callable[[LogRecord], bool]
+_FilterType: TypeAlias = _SupportsFilter | Callable[[LogRecord], bool]
```
Akuli, line 75: - Turns out we should keep `Filter` here, based on the mypy_primer output. Sphinx uses `# type: ignore` when they define a custom filter with a weird `filter()` method. Including `Filter` here means that they don't get errors when adding the custom filter, only when defining it. Getting errors in only one place is better IMO.
- On 3.12+, we should probably support the `LogRecord` return type here too.

```suggestion
if sys.version_info >= (3, 12):
    _FilterType: TypeAlias = Filter | _SupportsFilter | Callable[[LogRecord], bool | LogRecord]
else:
    _FilterType: TypeAlias = Filter | _SupportsFilter | Callable[[LogRecord], bool]
```

Akuli (commented):
This is turning out trickier than expected :)

stdlib/logging/__init__.pyi
Akuli: Somehow I didn't see your last two comments, and commented the same exact things myself :)

stdlib/logging/__init__.pyi
jenstroeger: Commit https://github.com/python/typeshed/pull/11018/commits/eed6015703ce4369bc4d032c33fa768e53ce58f0 🤓

stdlib/logging/__init__.pyi
jenstroeger: > Somehow I didn't see your last two comments, and commented the same exact things myself :)

Great minds 👍🏼

jenstroeger:
I was thinking that making by `_FilterType` private we’ll have to the _replicate_ it wherever a logging filter can be passed, see for example `pytest` issue https://github.com/pytest-dev/pytest/issues/11610 🤔

Akuli:
Pytest and whoever else needs to annotate "just some filter" can use `logging._FilterType`, even if it is named with an underscore in front.

stdlib/logging/__init__.pyi
```diff
@@ -66,10 +66,20 @@ if sys.version_info >= (3, 12):
 _SysExcInfoType: TypeAlias = tuple[type[BaseException], BaseException, TracebackType | None] | tuple[None, None, None]
 _ExcInfoType: TypeAlias = None | bool | _SysExcInfoType | BaseException
 _ArgsType: TypeAlias = tuple[object, ...] | Mapping[str, object]
-_FilterType: TypeAlias = Filter | Callable[[LogRecord], bool]
 _Level: TypeAlias = int | str
 _FormatStyle: TypeAlias = Literal["%", "{", "$"]
 
+if sys.version_info >= (3, 12):
+    class _SupportsFilter(Protocol):
+        def filter(self, record: LogRecord) -> bool | LogRecord: ...
```
Akuli, line 74: Let's make the argument positional-only in the protocol, so if your class names it something else than `Record`, it will still match the protocol:

```suggestion
        def filter(self, __record: LogRecord) -> bool | LogRecord: ...
```

stdlib/logging/__init__.pyi
```diff
@@ -66,10 +66,20 @@ if sys.version_info >= (3, 12):
...
+
+    _FilterType: TypeAlias = Filter | Callable[[LogRecord], bool | LogRecord] | _SupportsFilter
+else:
+    class _SupportsFilter(Protocol):
+        def filter(self, record: LogRecord) -> bool: ...
```
Akuli, line 79: Same here:
```suggestion
        def filter(self, __record: LogRecord) -> bool: ...
```

Akuli (commented):
Thanks for working on this! One more nit.

stdlib/logging/__init__.pyi
jenstroeger: I made the change in commit https://github.com/python/typeshed/pull/11018/commits/fe68d1aa8399e68ffc87c35834b997a42fdc5a73, but I’m still a little unclear: is the dunder `__record` a convention that the name is now “flexible”?

It doesn’t make the `record` arg (not capitalized `Record`) positional-only, though, that would [require a `/`](https://peps.python.org/pep-0570/) as well, right?

stdlib/logging/__init__.pyi
jenstroeger: Commit https://github.com/python/typeshed/commit/fe68d1aa8399e68ffc87c35834b997a42fdc5a73.

stdlib/logging/__init__.pyi
Akuli: Type-checkers special-case `__dunder`-named arguments and treat them as positional-only. This is a bit surprising, and we will soon be able to switch to the new `/` syntax.

Akuli (approved):
Thanks!

stdlib/logging/__init__.pyi
AlexWaygood: This is documented in PEP 484: https://peps.python.org/pep-0484/#positional-only-arguments

(PEP 484 predates [PEP 570](https://peps.python.org/pep-0570/), which was only implemented in Python 3.8 -- typeshed still supports Python 3.7.)

PR #11019, opened 2023-11-11
Provide precise type hints for permutations

Serious-senpai:
Provide more precise type hints for `itertools.permutations` with `r` from 1 to 5 (similar to `itertools.combinations`)


stdlib/itertools.pyi
```diff
@@ -241,10 +241,21 @@ class product(Iterator[_T_co]):
     def __iter__(self) -> Self: ...
     def __next__(self) -> _T_co: ...
 
-class permutations(Iterator[tuple[_T, ...]], Generic[_T]):
-    def __init__(self, iterable: Iterable[_T], r: int | None = ...) -> None: ...
+class permutations(Iterator[_T_co]):
+    @overload
+    def __new__(cls, iterable: Iterable[_T], r: Literal[1]) -> permutations[tuple[_T]]: ...
```
AlexWaygood, line 246: Maybe we could omit this one? I can't think of a valid use case for passing in `1` as the second argument; not sure it really deserves its own overload

AlexWaygood (commented):
Looks great, thanks! Just one nit:

stdlib/itertools.pyi
Serious-senpai: I don't think this overload is necessary too, but there's no harm in adding it right?

Serious-senpai: After some consideration, I believe this overload is a good-for-nothing 😅, will remove it now.

AlexWaygood: Thanks! Yeah, the harm is small, but it does do some harm. It makes the stub more verbose — that's annoying for people reading through typeshed trying to understand the stubs, and it's also annoying for users of IDEs — when they hover over the function in VSCode, for example, the displayed signature will have one more overload.

Analysing functions with lots of overloads can also cause performance issues for type checkers, though with only 5 the impact should be pretty minimal.

All in all, these are very minor points, and I wouldn't have cared about them _if_ I could see a way in which this overload could have been useful — but I couldn't :)

AlexWaygood (approved):
Thanks!

PR #11022, opened 2023-11-12
Direct Union replacements for ConvertibleToInt/ConvertibleToFloat

Avasam:
Extracted from #11003 . Only does direct replacements of the full aliased union.


AlexWaygood (approved):
LGTM, thanks! I still think we should hold off at least a few days before merging, though -- there were a few regressions in mypy 1.7, so it's very possible that not everyone will be able to upgrade just yet.

Avasam:
> I still think we should hold off at least a few days before merging, though

Agreed. I mistook the typeshed sync to be in 1.6 otherwise I would've waited a bit more before opening #11003

AlexWaygood:
I think we're good to go on this now!

PR #11023, opened 2023-11-12
Bump pyinstaller to 6.2.*

Avasam:
Closes #11021 .
Just a false-positive from stubtest. `autocomplete` is for internal use for the CLI app. About as useful as exposing "logger"

### DIFF
--- a/stubs/pyinstaller/METADATA.toml
+++ b/stubs/pyinstaller/METADATA.toml
@@ -4,5 +4,4 @@ requires = ["types-setuptools"]
 requires_python = ">=3.8"
 
 [tool.stubtest]
-# Avoids stubtest-false-positives, doesn't affect type-hints since autocomplete is for internal use for the CLI app
 extras = ["completion"]
--- a/stubs/pyinstaller/METADATA.toml
+++ b/stubs/pyinstaller/METADATA.toml
@@ -1,4 +1,4 @@
-version = "6.1.*"
+version = "6.2.*"
 upstream_repository = "https://github.com/pyinstaller/pyinstaller"
 requires = ["types-setuptools"]
 requires_python = ">=3.8"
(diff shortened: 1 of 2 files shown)


stubs/pyinstaller/METADATA.toml
```diff
@@ -1,4 +1,4 @@
-version = "6.1.*"
+version = "6.2.*"
 upstream_repository = "https://github.com/pyinstaller/pyinstaller"
 requires = ["types-setuptools"]
 requires_python = ">=3.8"
```
AlexWaygood, line 4: If we did this, we wouldn't need the new allowlist entry (I think), since this would mean that `argcomplete` would be installed in CI, which would mean we'd never hit the `except ImportError` branch [here](https://github.com/pyinstaller/pyinstaller/blob/dd40fb2cd75b86b5bbe23843bafde09740f451af/PyInstaller/__main__.py#L27-L32), which would mean stubtest would be able to detect that it's just imported from another module

```suggestion
requires_python = ">=3.8"

[tool.stubtest]
extras = ["completion"]
```

Avasam: Makes sense, let's try that. I'll add a comment specifying why though, since the extra wouldn't affect the end-user's type hints

stubs/pyinstaller/METADATA.toml
```diff
@@ -1,4 +1,8 @@
...
+
+[tool.stubtest]
+# Avoids stubtest-false-positives, doesn't affect type-hints since autocomplete is for internal use for the CLI app
```
AlexWaygood, line 7: This field in the `tool.stubtest` table is just an internal-to-typeshed thing; I don't think the contents of the field could ever impact users in any way. So I think we could _probably_ do without this comment -- if it's removed, tests will start failing, which is probably all the documentation we need here :)

But if we _do_ want to keep it, 116 chars is too long for a comment ;)

```suggestion
# Avoids stubtest-false-positives.
# Doesn't affect type-hints since autocomplete is for internal use for the CLI app
```

AlexWaygood (approved):
thanks!

PR #11025, opened 2023-11-12
Bump PyScreeze to 0.1.30

Avasam:
Closes #11020


stubs/PyScreeze/pyscreeze/__init__.pyi
```diff
@@ -165,6 +168,7 @@ def locateOnWindow(
     step: int = 1,
     confidence: None = None,
 ) -> Box | None: ...
+def screenshotWindow(title: Unused) -> None: ...  # Not implemented yet.
```
Avasam, line 171: I *assume* this is gonna be a string. But it could also be a window handle depending on future implementation. Probably fine as-is since it's accurate, makes stubtest happy, and reduces chance of false-positives later.

stubs/PyScreeze/pyscreeze/__init__.pyi
```diff
@@ -173,7 +181,16 @@ def pixelMatchesColor(
...
+
+if sys.platform == "win32":
+    def screenshot(
+        imageFilename: StrOrBytesPath | None = None, region: tuple[int, int, int, int] | None = None, allScreens: bool = False
+    ) -> Image.Image: ...
+
+else:
+    def screenshot(
+        imageFilename: StrOrBytesPath | None = None, region: tuple[int, int, int, int] | None = None
+    ) -> Image.Image: ...
```
AlexWaygood, line 193: This looks correct, but I'm very confused about what's going on here... if I `pip install pyscreeze` into a virtual environment, I can see the `allScreens` argument is there in `site-packages/pyscreeze/__init__.py`. But I can't see it anywhere when I view https://github.com/asweigart/pyscreeze/blob/master/pyscreeze/__init__.py online... this isn't really a question for you, I'm just quite confused 😵‍💫

Avasam: I... what?? I'm guessing a local upload to PyPI that contained more changes than pushed to master? Good catch, I'll open an issue cuz wut.

stubs/PyScreeze/pyscreeze/__init__.pyi
```diff
@@ -11,10 +12,17 @@ _R = TypeVar("_R")
...
+PILLOW_VERSION: Final[tuple[int, int, int]]
 RUNNING_PYTHON_2: Final = False
-GRAYSCALE_DEFAULT: Final = False
-scrotExists: Final[bool]
+SCROT_EXISTS: Final[bool]
+GNOMESCREENSHOT_EXISTS: Final[bool]
+
+if sys.platform == "linux":
+    RUNNING_X11: Final[bool]
+    RUNNING_WAYLAND: Final[bool]
```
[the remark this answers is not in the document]
Avasam: `SCROT_EXISTS` and `GNOMESCREENSHOT_EXISTS` are nice to have since you can tell ahead of time using Pillow's own logic whether those functionality are A) available, and B) will be used by PyScreeze. Which is something you may also want to communicate to the user.
Similar for Wayland vs X11, although that one's a lot more simple.

These are things I'm interested in having ahead of time (without copying the logic) for AutoSplit on Linux (the port's waiting for more refactor, but I have a branch)

`PILLOW_VERSION` and `RUNNING_PYTHON_2` though, yeah maybe not. But also, it's there, and adding to the allowlist would have a similar maintenance cost (must update entry if name changes or is removed)

AlexWaygood (approved):
Thanks!

PR #11036, opened 2023-11-17
[jsonschema] Bump to 4.20.*

srittau:
Closes #11035


stubs/jsonschema/jsonschema/_legacy_keywords.pyi
```diff
@@ -15,3 +15,7 @@ def properties_draft3(validator, properties, instance, schema) -> None: ...
 def type_draft3(validator, types, instance, schema) -> None: ...
 def contains_draft6_draft7(validator, contains, instance, schema) -> None: ...
 def recursiveRef(validator, recursiveRef, instance, schema) -> None: ...
+def find_evaluated_item_indexes_by_schema(validator, instance, schema): ...
+def unevaluatedItems_draft2019(validator, unevaluatedItems, instance, schema): ...
+def find_evaluated_property_keys_by_schema(validator, instance, schema): ...
+def unevaluatedProperties_draft2019(validator, uP, instance, schema): ...
```
AlexWaygood, line 21:
```suggestion
def find_evaluated_item_indexes_by_schema(validator, instance, schema) -> list[int]: ...
def unevaluatedItems_draft2019(validator, unevaluatedItems, instance, schema) -> Iterator[ValidationError]: ...
def find_evaluated_property_keys_by_schema(validator, instance, schema) -> list[Incomplete]: ...
def unevaluatedProperties_draft2019(validator, uP, instance, schema) -> Iterator[ValidationError]: ...
```

stubs/jsonschema/jsonschema/exceptions.pyi
```diff
@@ -72,11 +72,12 @@ class FormatError(Exception):
     def __init__(self, message, cause: Incomplete | None = None) -> None: ...
 
 class ErrorTree:
-    errors: Incomplete
-    def __init__(self, errors=()) -> None: ...
-    def __contains__(self, index): ...
+    errors: MutableMapping[str, ValidationError]
+    def __init__(self, errors: Iterable[ValidationError] = ()) -> None: ...
+    def __contains__(self, index: str | int) -> bool: ...
```
AlexWaygood, line 77: I know this is just copied from the runtime package's source code, but this means that type checkers won't see this class as a subtype of `Container`, since `Container.__contains__` has `object` as the parameter annotation. So it's questionable whether the annotation in the runtime package is correct.

stubs/jsonschema/jsonschema/exceptions.pyi
```diff
@@ -72,11 +72,12 @@ class FormatError(Exception):
...
     def __getitem__(self, index): ...
-    def __setitem__(self, index, value) -> None: ...
+    @deprecated("ErrorTree.__setitem__ is deprecated without replacement.")
+    def __setitem__(self, index: str | int, value: ErrorTree) -> None: ...
     def __iter__(self): ...
```
AlexWaygood, line 81:
```suggestion
    def __iter__(self) -> Iterator[str]: ...
```

AlexWaygood (approved):
Looks good, some small suggestions below from skimming the runtime package's source code

stubs/jsonschema/jsonschema/exceptions.pyi
srittau: Looking at the source, `object` should work fine here.

PR #11040, opened 2023-11-18
stdlib: fix `typeobject.__base__` for class `object`

XuehaiPan:
For class `object`, its `__base__` attribute is `None`, and `__bases__` is an empty tuple `()`.

```python
>>> object.__base__ is None
True
>>> object.__bases__
()
```

For other type objects, the `__base__` attribute is a subclass of `type`, and `__bases__` is a non-empty tuple of type(s).


srittau (approved):
LGTM, especially since the tests pass, but since this is a change to such a fundamental class, I'd like second maintainer to chime in.

stdlib/builtins.pyi
```diff
@@ -161,8 +161,10 @@ class classmethod(Generic[_T, _P, _R_co]):
         def __wrapped__(self) -> Callable[Concatenate[type[_T], _P], _R_co]: ...
 
 class type:
+    # object.__base__ is None. Otherwise, it would be a type.
     @property
-    def __base__(self) -> type: ...
+    def __base__(self) -> type | None: ...
+    # object.__bases__ is (). Otherwise, it would be a non-empty tuple of type(s).
```
AlexWaygood, line 167: Not sure we need this comment? It doesn't contradict the below type annotation, and seems kinda obvious to me :)

```suggestion
```

AlexWaygood (commented):
Looks good to me too! Just one nit:

AlexWaygood (approved):
Thanks!

PR #11044, opened 2023-11-18
stdlib: Apply some simple deprecations

JelleZijlstra:
Split out from #11009: only deprecations that shouldn't crash pytype or stubtest.


srittau:
It seems `xml.etree.cElementTree.Element.__bool__` (i.e. the C version) was only added in Python 3.12, while `xml.etree.ElementTree.Element.__bool__` (the Python version) was always(tm) present. This confuses stubtest. Maybe split this out into a separate PR as well, as it seems to be a bit complicated?

srittau (approved):

PR #11046, opened 2023-11-19
Add `NamedTupleMeta` metaclass

sobolevn:
Refs https://github.com/python/mypy/issues/16521

Source:
- 3.8: https://github.com/python/cpython/blob/3.8/Lib/typing.py#L1613
- 3.9: https://github.com/python/cpython/blob/3.9/Lib/typing.py#L1864
- 3.10: https://github.com/python/cpython/blob/3.10/Lib/typing.py#L2265
- 3.11: https://github.com/python/cpython/blob/3.11/Lib/typing.py#L2885
- 3.12: https://github.com/python/cpython/blob/3.12/Lib/typing.py#L2739

### DIFF
--- a/stdlib/typing.pyi
+++ b/stdlib/typing.pyi
@@ -2,7 +2,7 @@ import collections  # Needed by aliases like DefaultDict, see mypy issue 2986
 import sys
 import typing_extensions
 from _collections_abc import dict_items, dict_keys, dict_values
-from _typeshed import IdentityFunction, Incomplete, ReadableBuffer, SupportsKeysAndGetItem, Self as _typeshed_Self
+from _typeshed import IdentityFunction, Incomplete, ReadableBuffer, Self as _typeshed_Self, SupportsKeysAndGetItem
 from abc import ABCMeta, abstractmethod
 from contextlib import AbstractAsyncContextManager, AbstractContextManager
 from re import Match as Match, Pattern as Pattern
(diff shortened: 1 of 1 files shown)


sobolevn:
Mypy does not like this change:

>  stdlib/typing.pyi:865: error: Metaclass conflict: the metaclass of a derived class must be a (non-strict) subclass of the metaclasses of all its bases

sobolevn:
I think the reason is that `tuple` inherits types that use `ABCMeta`, so, let's preted that `NamedTupleMeta` also has `ABCMeta`.

sobolevn:
Now pyright finds some cases where `NamedTuple` is used together with `Enum`:

```python
class _ASN1ObjectBase(NamedTuple):
    nid: int
    shortname: str
    longname: str
    oid: str

class _ASN1Object(_ASN1ObjectBase):
    def __new__(cls, oid: str) -> Self: ...
    @classmethod
    def fromnid(cls, nid: int) -> Self: ...
    @classmethod
    def fromname(cls, name: str) -> Self: ...

class Purpose(_ASN1Object, enum.Enum):
    SERVER_AUTH: _ASN1Object
    CLIENT_AUTH: _ASN1Object
```

Another option for us is to use `# type: ignore[misc]` on this definition. I will give this a try.

sobolevn:
:(

JelleZijlstra:
Sorry, seems like this didn't work out. The linked mypy bug should likely be fixed with special-casing in mypy itself.

PR #11048, opened 2023-11-20
Make `ExitStack` generic in return type of `__exit__`

Daverball:
See https://discuss.python.org/t/add-an-else-clause-to-with-statements-to-detect-early-termination/38031 for related discussion.


Daverball:
Quite a few mypy primer hits here. This change probably does not make sense until after PEP696 support.

srittau:
See #11422 for the type var generics feature tracker.

Daverball:
This is really encouraging, PEP-696 truly shines here. I'll check if there are any other places in typeshed that use `AbstractContextManager`/`AbstractAsyncContextManager` that should have their second parameter specified.

srittau:

Daverball:
One of the mypy_primer hits is irrelevant, since it is the same unrelated error with different types and the other one looks like an improvement to me.

There are a couple of caveats to this change however:
- Right now I'm explicitly specifying the second parameter in subclasses to `AbstractContextManager`/`AbstractAsyncContextManager` in the sdlib as long as they override `__exit__` with a different value than the default for clarity. Technically this shouldn't be necessary and we could keep the surface area of the change smaller. Or we could go the opposite way and make sure we always specify the second parameter everywhere in the stdlib. Either of those options would probably be more manageable for maintainers of type checkers that don't yet support PEP-696, compared to the mish mash we have now.
- `typing.ContextManager` and `typing.AsyncContextManager` are just aliases for the ABCs that changed, but this means the Python docs are no longer accurate. Technically it is a harmless mismatch, since you can omit the parameter, that is missing from the docs, but we should probably still consider updating the docs, if we accept this change.

stdlib/typing.pyi
```diff
@@ -129,8 +129,8 @@ if sys.version_info >= (3, 11):
 if sys.version_info >= (3, 12):
     __all__ += ["TypeAliasType", "override"]
 
-ContextManager = AbstractContextManager
-AsyncContextManager = AbstractAsyncContextManager
+ContextManager = AbstractContextManager[_T_co, bool | None]  # noqa: Y026
+AsyncContextManager = AbstractAsyncContextManager[_T_co, bool | None]  # noqa: Y026
```
srittau, line 133: I assume that this change is the reason that mypy primer stumbles when checking kopf:

https://github.com/nolar/kopf/blob/96e8cbcec16c78be177922f5d8c000fa4b4022c1/kopf/_core/engines/admission.py#L266

```python
    elif isinstance(settings.admission.server, AsyncContextManager):
```

Just adding generic parameters here seems wrong to me as this is a simple name aliasing, not a type alias. Why are those changes necessary?

Daverball: Refer to my previous comment: https://github.com/python/typeshed/pull/11048#issuecomment-2012965236

This is the third iteration of this change, both using a subclass or completely copying the protocol with just the one generic parameter seem to make stubtest really unhappy for some reason. It starts complaining that `ContextManager`/`AsyncContextManager` aren't types.

Leaving the old aliases intact is a possibility, however, we then end up with a false negative if people use the second parameter on `ContextManager`/`AsyncContextManager`, since that will result in a runtime error, due to restrictive implementation of `typing.Generic`, which checks the number of parameters passed in.

Daverball: I suppose the stubtest errors can actually be ignored. I think the reason why stubtest didn't complain before, is, because it was just a regular assignment. Technically the complaint is accurate `typing._SpecialGenericAlias` is not a type, I just assumed stubtest already special cases these aliases, but apparently it doesn't, because it wasn't necessary for the simple re-exports.

So I will revert to the previous solution, which was more sane.

Daverball:
The new primer hit seems fine to me. I think the only reason that previously didn't error is because `AbstractContextManager[Any]` (due to the missing type parameter) was in the MRO, which used to be an alias for `ContextManager`, so mypy took a shortcut and didn't attempt to structurally match, because the simpler subclassing check already told it everything is fine (due to `Any`). This could actually be considered a mypy bug. I'm not sure how I feel about subclasses that override all occurrences of a `TypeVar` that was previously bound to `Any` with a concrete type still behaving as if the type was `Any` in some scenarios...

srittau (approved):
LGTM, the new primer hits are promising. I'll leave it open a bit, in case another maintainer wants to chime in.

PR #11050, opened 2023-11-21
Improve a few dataclass reprs in stubsabot

AlexWaygood:
This would have made interactively debugging #11049 a bit easier -- some of these dataclass fields have huge reprs

### DIFF
--- a/scripts/stubsabot.py
+++ b/scripts/stubsabot.py
@@ -329,9 +329,6 @@ class DiffAnalysis:
     py_files: list[FileInfo]
     py_files_stubbed_in_typeshed: list[FileInfo]
 
-    def __repr__(self) -> str:
-        return f"<DiffAnalysis object at {id(self)}>"
-
     @property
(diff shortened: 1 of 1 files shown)


scripts/stubsabot.py
```diff
@@ -323,12 +323,15 @@ def _plural_s(num: int, /) -> str:
...
 
-@dataclass
+@dataclass(repr=False)
 class DiffAnalysis:
     MAXIMUM_NUMBER_OF_FILES_TO_LIST: ClassVar[int] = 7
     py_files: list[FileInfo]
     py_files_stubbed_in_typeshed: list[FileInfo]
 
+    def __repr__(self) -> str:
+        return f"<DiffAnalysis object at {id(self)}>"
```
JelleZijlstra, line 333: Isn't this just the default? I don't think you need both this override and setting `repr=False` in the decorator.

AlexWaygood: oh, good point

JelleZijlstra (approved):

PR #11056, opened 2023-11-22
Enable `redundant-self` mypy error code

sobolevn:
Docs: https://mypy.readthedocs.io/en/stable/error_code_list2.html#check-that-methods-do-not-have-redundant-self-annotations-redundant-self

Let's see if it works!


AlexWaygood (commented):
Want to enable it in `typecheck_typeshed.py` as well?

(I don't think we should add it to `regr_test.py`; it's not relevant to the things we're testing there.)

AlexWaygood (approved):

PR #11058, opened 2023-11-22
stdlib: HTTPSConnection(port=) is int|None

### LINKED ISSUE #11057: `HTTPSConnection.__init__(port=)` incorrectly annotated as `str | None` for 3.12
This was introduced in #10392.

As far as I can tell by reading the code, it's only ever valid to give an integer or `None`.  This matches the (correct) typeshed annotations for `HTTPConnection` and versions before Python 3.12.  The problem seems to be specific to `HTTPSConnection` and Python 3.12.

allisonkarlitskaya:
Fix an apparent typo in the type of this field.

Closes #11057

### DIFF
--- a/stdlib/http/client.pyi
+++ b/stdlib/http/client.pyi
@@ -187,7 +187,7 @@ class HTTPSConnection(HTTPConnection):
         def __init__(
             self,
             host: str,
-            port: str | None = None,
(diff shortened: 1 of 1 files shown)


AlexWaygood (approved):
Thanks!

PR #11061, opened 2023-11-23
stdlib: remove unused `type: ignore` comments

AlexWaygood:
And, where relevant, use the new `[overload-overlap]` error code instead of the `[misc]` error code. The `[overload-overlap]` code ignores a narrower set of errors from mypy.

I prepared this PR using the same technique I used in #9801.


srittau (approved):

srittau:
Nice. We're down to 121 misc ignores (according to `rgrep ignore\\[misc\\] stdlib/ stubs/ | wc -l`. I hope we will get this down to 0 at some point.

PR #11064, opened 2023-11-23
Fix typos in docs and comments

AlexWaygood:
Found with codespell


stdlib/importlib/abc.pyi
```diff
@@ -134,7 +134,7 @@ if sys.version_info >= (3, 9):
             def joinpath(self, __child: str) -> Traversable: ...
 
         # The documentation and runtime protocol allows *args, **kwargs arguments,
-        # but this would mean that all implementors would have to support them,
+        # but this would mean that all implementers would have to support them,
```
JelleZijlstra, line 137: More than you wanted to know on this one: https://internals.rust-lang.org/t/spelling-bikeshed-implementor-or-implementer/16926

AlexWaygood: correct, that is more than i wanted to know

Issue #8703, opened 2022-09-07
functools.partial should use PEP612 ParamSpec


benbariteau:
[Currently](https://github.com/python/typeshed/blob/e2ce7c6344dc9544b3555a4b81aff5bfadc1430c/stdlib/functools.pyi#L75), functools.partial's `__call__` method takes `*args: Any, **kwargs: Any` and returns `_T` (a TypeVar for the return type of the callable the partial is wrapping). This is decent, but matching the parameter typing would make this more useful in more situations.

I can sort of see how to deal with this, although I'll admit I'm unsure how to make this generically work. I imagine you could probably deal with some number of `Concatenate`s (maybe up to 5 or 10) to deal with most cases. It's possible fixing this requires another PEP to allow an arbitrary-length `Concatenate`-type pattern.

Related to https://github.com/python/mypy/issues/1484

AlexWaygood:
You can nearly get there with a combination of PEP 612 and PEP 646:

```python
from typing import Any, Callable, Concatenate, Generic, ParamSpec, TypeVar, TypeVarTuple, overload

_P1 = ParamSpec("_P1")
_P2 = ParamSpec("_P2")
_T = TypeVar("_T")
_R_co = TypeVar("_R_co", covariant=True)
_Ts = TypeVarTuple("_Ts")

class partial(Generic[_P1, _P2, _T, _R_co, *_Ts]):
    @overload
    def __new__(cls, __func: Callable[_P1, _R_co]) -> partial[_P1, _P1, Any, _R_co]: ...
    @overload
    def __new__(cls, __func: Callable[Concatenate[*_Ts, _P2], _R_co], *args: *_Ts) -> partial[Concatenate[*_Ts, _P2], _P2, Any, _R_co, *_Ts]: ...
    @overload
    def __new__(cls, __func: Callable[_P1, _R_co], *args: *_Ts, **kwargs: _T) -> partial[_P1, ..., _T, _R_co, *_Ts]: ...
    def __call__(self, *args: _P2.args, **kwargs: _P2.kwargs) -> _R_co: ...
    func: Callable[_P1, _R_co]
    args: tuple[*_Ts]
    keywords: dict[str, _T]
```

Mypy doesn't support PEP 646 yet, however. And even if it did, there would still be some corner cases that don't work (https://github.com/PyCQA/flake8-pyi/pull/150#issuecomment-1023575984)

benbariteau:
[...]

PR #11066, opened 2023-11-23
Update setuptools to v69

Avasam:
This is an attempt at closing #11049 without needing any stubtest update.

The main point is about removing `pkg_resources._vendor` and adding `packaging` to the requirements

Edit: Looks like either stubtest or stub_uploader is gonna need an update to keep the packaging subclassing.

---

Separatly, I'm not sure how to best handle `newer_pairwise_group`. typing it as a `def` stubtest says `error: setuptools.modified.newer_pairwise_group is not a function`. typing is as `partial[Incomplete]` looses parameters signature (ref #8703). And pyright seems to have some special handling of `functools.partial` seeing the following (which also validate correct parameter names and types in the cli):
image
I feel like I should go with an allowlist entry for mypy users.


stubs/setuptools/setuptools/dep_util.pyi
```diff
@@ -1 +1 @@
-def newer_pairwise_group(sources_groups, targets): ...
+from ._distutils._modified import newer_group as newer_group, newer_pairwise_group as newer_pairwise_group
```
Avasam, line 1: This entire module (`stubs/setuptools/setuptools/dep_util.pyi`) is deprecated in favor of `setuptools.modified`
I could keep the defs instead of importing from `_distutils` so I can mark with `@deprecated`

stubs/setuptools/setuptools/modified.pyi
```diff
@@ -0,0 +1,8 @@
+from ._distutils._modified import (
+    newer as newer,
+    newer_group as newer_group,
+    newer_pairwise as newer_pairwise,
+    newer_pairwise_group as newer_pairwise_group,
+)
```
Avasam, line 6: I could define the methods here instead of in `setuptools/_distutils/_modified` to avoid a adding a private module

JelleZijlstra:
I gather `newer_pairwise_group` is a `functools.partial` at runtime? I would agree with stubbing it as an equivalent `def` function, and allowlisting stubtest if it complains.

stubs/setuptools/setuptools/modified.pyi
AlexWaygood: Better to define them where they're defined at runtime imo, as you currently have it

Avasam:
> I gather `newer_pairwise_group` is a `functools.partial` at runtime? I would agree with stubbing it as an equivalent `def` function, and allowlisting stubtest if it complains.

Yep, if it makes it clearer, this is what it'd look like using `functool.partial` assignement in stub: https://github.com/python/typeshed/pull/11066/commits/5946e13ad44e668d28ac597f9092121b74835cb9#diff-c7f29e111ff63f317e294acc4ea34b5ee0181dfe30dfcbb54cb3fe0c7df53803R9

Avasam:
Superseeded by #11069

PR #11067, opened 2023-11-24
Added paramiko3.2 PKey and PublicBlob methods

dispread:
Addded missing paramiko methods from https://docs.paramiko.org/en/latest/api/keys.html#paramiko.pkey.PKey


AlexWaygood (approved):
Thanks!

PR #11068, opened 2023-11-24
Provide precise type hints for combinations_with_replacement

Serious-senpai:
Provide more precise type hints for `itertools.combinations_with_replacement` with `r` from 2 to 5, similar to https://github.com/python/typeshed/pull/11019


AlexWaygood (approved):
Thanks!

PR #11071, opened 2023-11-26
psycopg2: fix RangeCaster.__init__ annotation

exg:
The pgrange and pyrange arguments should be a string or a class object derived from RangeAdapter and Range, respectively. See

https://github.com/psycopg/psycopg2/blob/2.9.9/lib/_range.py#L313 https://github.com/psycopg/psycopg2/blob/2.9.9/lib/_range.py#L330

### DIFF
--- a/stubs/psycopg2/psycopg2/_range.pyi
+++ b/stubs/psycopg2/psycopg2/_range.pyi
@@ -51,7 +51,7 @@ class RangeCaster:
     typecaster: Any
     array_typecaster: Any
     def __init__(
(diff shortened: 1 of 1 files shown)


hamdanal (changes requested):
Nice catch.
The type of the `pyrange` parameter of the `register_range` function 20 lines above also needs to be updated.

exg:
> Nice catch. The type of the `pyrange` parameter of the `register_range` function 20 lines above also needs to be updated.

Fixed, thanks.

hamdanal (approved):

srittau (approved):
Thanks!

PR #11073, opened 2023-11-27
`socket.{if_nameindex,if_nametoindex,if_indextoname}` not exported on PyPy on Windows

A5rocks:
Here:

```
PS C:\Users\A5rocks\Downloads\pypy3.10-v7.3.13-win64\pypy3.10-v7.3.13-win64> ./pypy.exe
Python 3.10.13 (f1607341da97ff5a1e93430b6e8c4af0ad1aa019, Sep 28 2023, 05:42:24)
[PyPy 7.3.13 with MSC v.1929 64 bit (AMD64)] on win32
>>>> from socket import if_indextoname
ImportError: cannot import name 'if_indextoname' from 'socket' (C:\Users\A5rocks\Downloads\pypy3.10-v7.3.13-win64\pypy3.10-v7.3.13-win64\Lib\socket.py)
>>>> from socket import if_nameindex
ImportError: cannot import name 'if_nameindex' from 'socket' (C:\Users\A5rocks\Downloads\pypy3.10-v7.3.13-win64\pypy3.10-v7.3.13-win64\Lib\socket.py)
>>>> from socket import if_nametoindex
ImportError: cannot import name 'if_nametoindex' from 'socket' (C:\Users\A5rocks\Downloads\pypy3.10-v7.3.13-win64\pypy3.10-v7.3.13-win64\Lib\socket.py)
```

I'm going to raise a bug report over at PyPy but for now the stubs are incorrect.


A5rocks:
Actually, looks like none of these functions exist, ever: https://foss.heptapod.net/pypy/pypy/-/issues/3249

A5rocks:
Aww I actually assumed that `sys.implementation.name` was a valid thing to check against :(

Well, guess there's no way to resolve this then.

JelleZijlstra:
Yes, the type system has no way to distinguish between CPython and PyPy. We could add it, but I think this is the first time I've seen a need for it, so it may not be worth the trouble.

A5rocks:
Yeah I found https://github.com/python/typing/issues/1339 but 0 :+1: or issues linking it probably means not much use.

PR #11076, opened 2023-11-28
asyncio: create_subprocess_exec is the same as create_subprocess_shell, except where it isnt

pajod:
Duplicate for the _exec case what has already been duplicated for the _shell case in 38dfb57adf7404fdf7419c472c08e3f0670b72f3 and follow-ups.
Except for the `shell: bool` argument: that one is the whole point of having two; that one needs to be flipped.


AlexWaygood (approved):
Nice catch, thanks!

PR #11078, opened 2023-11-28
Remove hack obsoleted by PEP 655

srittau:


srittau:
Huh, it seems that while pytype supports `NotRequired`, it doesn't support `Required`. I'll invert the TypeDict and try again.

srittau:
google/pytype#1535

stdlib/logging/config.pyi
```diff
@@ -50,17 +50,15 @@ _FilterConfiguration: TypeAlias = _FilterConfigurationTypedDict | dict[str, Any]
...
-class _DictConfigArgs(_OptionalDictConfigArgs, TypedDict):
+class _DictConfigArgs(TypedDict):
     version: Literal[1]
+    formatters: NotRequired[dict[str, _FormatterConfiguration]]
+    filters: NotRequired[dict[str, _FilterConfiguration]]
+    handlers: NotRequired[dict[str, _HandlerConfiguration]]
+    loggers: NotRequired[dict[str, _LoggerConfiguration]]
+    root: NotRequired[_RootLoggerConfiguration]
+    incremental: NotRequired[bool]
+    disable_existing_loggers: NotRequired[bool]
```
AlexWaygood, line 61: Ugh... maybe we should just wait for pytype to fix the `Required` bug? I agree the existing hack is ugly, but this seems pretty ugly too :(

srittau: While this is also ugly, I still prefer it to the old hack, which is a bit hard to understand. That said, maybe the pytype problem gets fixed quickly.

rchen152: Yeah, this should be a quick fix in pytype =) I'll try to get a release out in the next day or two.

rchen152: Pytype release 2023.11.29 contains the fix.

AlexWaygood (approved):
Thanks!

PR #11083, opened 2023-11-29
Bump boltons to 23.1.*

Avasam:
Closes #10958


stubs/boltons/boltons/ecoutils.pyi
```diff
@@ -15,7 +15,6 @@ EXPAT_VERSION: str
 CPU_COUNT: int
 HAVE_THREADING: bool
 HAVE_IPV6: bool
-RLIMIT_NOFILE: int
```
Avasam, line 18: This was a non-win32-only import. So it was caught on my Windows machine. But it doesn't seem this is meant to be re-exported.

AlexWaygood (approved):
Thanks!

PR #11084, opened 2023-11-29
Bump pyright to 1.1.339

Avasam:
I'm mostly interested in [1.1.335](https://github.com/microsoft/pyright/releases/tag/1.1.335) as it fixes issues with the `runtests.py` script:
> Fixed recent regression in the CLI that caused the `--files` command-line parameters to be overridden by the "include" setting in the config file.

PR #11085, opened 2023-11-29
psycopg2: improve `cursor_factory` params

Avasam:
Change originating from: https://github.com/python/typeshed/pull/11084#issuecomment-1831274491

idk if specifying the possible kwarg is necessary here. If not, then the tests in `stubs/psycopg2/@tests/test_cases/check_extensions.py` are still accurate. @JelleZijlstra I believe you last looked at this in https://github.com/python/typeshed/pull/7964/files#diff-0f425c352331e97f72634f6e626545d46059d6744c3b7e6a439119482869c8eeR91

### DIFF
--- a/stubs/psycopg2/psycopg2/_psycopg.pyi
+++ b/stubs/psycopg2/psycopg2/_psycopg.pyi
@@ -89,7 +89,7 @@ class _SupportsReadAndReadlineAndWrite(_SupportsReadAndReadline, SupportsWrite[s
 class cursor:
     arraysize: int
     binary_types: Incomplete | None
-    connection: _Connection
+    connection: connection
     itersize: int
     row_factory: Incomplete | None
     scrollable: bool | None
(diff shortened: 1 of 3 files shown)


hamdanal (approved):
LGTM. This is where the factory gets called: https://github.com/psycopg/psycopg2/blob/5fb59cd6eebd96e2c8a69a3a9d606534eb95abf0/psycopg/connection_type.c#L107

stubs/psycopg2/psycopg2/_psycopg.pyi
```diff
@@ -89,7 +89,7 @@ class _SupportsReadAndReadlineAndWrite(_SupportsReadAndReadline, SupportsWrite[s
 class cursor:
     arraysize: int
     binary_types: Incomplete | None
-    connection: _Connection
+    connection: connection
```
JelleZijlstra, line 92: I am not sure this is safe in all type checkers, since some may understand this as a forward reference. I'd prefer to keep the alias.

Avasam: No problem! I've open https://github.com/python/typeshed/issues/11113 so we don't forget to validate.

PR #11086, opened 2023-11-29
Fixes to `typing` and `typing_extensions` stubs

AlexWaygood:
Some of these are necessary due to the new `typing_extensions` release. Some are related things I noticed along the way.


JelleZijlstra (approved):
Thanks!

PR #11090, opened 2023-11-30
Fix: `functools.reduce` does not accept keyword arguments

rayansostenes:
```
from functools import reduce

reduce(lambda a, b: a+b, 'ABC', initial='')
```

I get no type error for the above code, but at runtime I get
```
TypeError: reduce() takes no keyword arguments
```

This PR alters the type definition of `functools.reduce` to make all parameters positional-only.


AlexWaygood (approved):
Thanks!

JelleZijlstra:
Seems like stubtest didn't catch this because `functools.reduce` doesn't use Argument Clinic. A comment in the source code says:

```
// Not converted to argument clinic, because of `args` in-place modification.
// AC will affect performance.
```

Whether that's true I'm not sure.

AlexWaygood:
The performance of AC has improved in recent versions, I think, so it might be worth revisiting that comment to see if it still applies

PR #11091, opened 2023-11-30
Add some missing attribute annotations to `trace.pyi`

AlexWaygood:
This will be useful for https://github.com/python/cpython/issues/109413

### DIFF
--- a/stdlib/trace.pyi
+++ b/stdlib/trace.pyi
@@ -46,8 +46,8 @@ class Trace:
     donothing: int
     trace: int
     start_time: int | None
-    globaltrace: Callable[[types.FrameType, str, Any], TraceFunction]
-    localtrace: Callable[[types.FrameType, str, Any], TraceFunction]
+    globaltrace: TraceFunction
+    localtrace: TraceFunction
     def __init__(
         self,
         count: int = 1,
--- a/stdlib/trace.pyi
+++ b/stdlib/trace.pyi
@@ -1,7 +1,7 @@
 import sys
 import types
-from _typeshed import StrPath, TraceFunction
-from collections.abc import Callable, Mapping, Sequence
+from _typeshed import Incomplete, StrPath, TraceFunction
+from collections.abc import Callable, Iterable, Mapping, Sequence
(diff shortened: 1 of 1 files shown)


stdlib/trace.pyi
```diff
@@ -27,7 +33,21 @@ class CoverageResults:
...
+    inifile: StrPath | None
+    outfile: StrPath | None
+    ignore: _Ignore
+    counts: dict[str, int]
+    pathtobasename: dict[Incomplete, Incomplete]
+    donothing: int
+    trace: int
+    start_time: int | None
+    globaltrace: Callable[[types.FrameType, str, Any], TraceFunction]
+    localtrace: Callable[[types.FrameType, str, Any], TraceFunction]
```
Avasam, line 50: Should those be `TraceFunction` directly?
The main difference is the `TraceFunction | None` return type.

`globaltrace` is passed directly to `threading.settrace` and `sys.settrace` which are both typed to accept a `TraceFunction`

`self.globaltrace` is also assigned a bunch of different possible functions that always return `None` (like `self.globaltrace_trackcallers`, `globaltrace_countfuncs`, etc.)

Avasam (commented):
I have two questions, other than that, this looks good assuming `__init__` was already typed correctly\*.

1. (see in-code comment about `TraceFunction`)
2. \*`infile` and `outfile` look like they could work as file descriptors. The online documentation says "name of the file" specifically, but docstring only says "file" https://docs.python.org/3/library/trace.html#trace.Trace I know for the stdlib typeshed sometimes differs from "what's possible" to "do the correct thing", so I'll defer to your judgement here.

stdlib/trace.pyi
AlexWaygood: Good catch. Yep, I think these should both just be `TraceFunction` directly. They're both just used for tracing, so, even if we _could_ be more precise with `self.localstrace`, it's probably fine to simplify here (and the alias gives a better sense of the intended purpose, anyway!)

AlexWaygood:
Thanks for the review! :D

PR #11092, opened 2023-12-01
Various annotations extraction from #9511

Avasam:
Extracted from, and improved over #9511 to make these changes reviewable.  
The following from #9511 is not present here:
- Any change relating to cell values, `Cell`, `MergedCell`, `ReadOnlyWorksheet`, `_WorkbookChild` annotations and their subclasses that can't be explicitly known by looking at surface implementation (so new instances and `isinstance` checks)
- 2 specialized Element Protocols that will be needed in `stubs/openpyxl/openpyxl/worksheet/_reader.pyi`
- Descriptors in `stubs/openpyxl/openpyxl/styles/styleable.pyi`
- `to_tree` methods (#10967)


stubs/openpyxl/openpyxl/worksheet/_reader.pyi
```diff
@@ -46,15 +49,15 @@ class WorkSheetParser:
     min_row: Incomplete | None
     min_col: Incomplete | None
     epoch: datetime
-    source: Incomplete
-    shared_strings: Incomplete
+    source: _FileRead
+    shared_strings: SupportsGetItem[int, Incomplete]
```
Avasam, line 53: It doesn't feel right to have such a simple Protocol as an attribute, but openpyxl's own usage of it is extremely simple. I don't see the property being dynamically bound anywhere else. It's in a private module. And it's undocumented.

Avasam:
Ah fun /s, even if return types are accurate, it leads to errors if the value is fed back into a method that's still inaccurately only expects `Worksheet`. I'd rather update both at once to avoid introducing false-positives downstream. And updating those params is out of scope for this PR. So I'll revert some return type changes for now.

stubs/openpyxl/openpyxl/worksheet/worksheet.pyi
```diff
@@ -183,22 +186,43 @@ class Worksheet(_WorkbookChild):
...
-        range_string: str | None = None,
-        start_row: int | None = None,
-        start_column: int | None = None,
-        end_row: int | None = None,
-        end_column: int | None = None,
+        range_string: str,
+        start_row: Unused = None,
+        start_column: Unused = None,
+        end_row: Unused = None,
+        end_column: Unused = None,
```
srittau, line 206: I wonder whether it would be safer to use `: None = None` here. It seems providing `range_string` and the rows and cols - while working at runtime - could introduce subtle bugs and is probably not what anyone wants.

But we should at least use `ConvertibleToInt | None` instead of `Unused`, because providing anything else (while working) is surely a bug.

srittau (changes requested):
Thanks, just one remark below.

stubs/openpyxl/openpyxl/worksheet/worksheet.pyi
Avasam: Ok. I can see how passing both (which this allows) may lead to unexpected bug by not knowing which param is actually used. I'll move the concept that these params go unused in that overload as a comment instead of the parameter type.

srittau (approved):

PR #11093, opened 2023-12-01
stdlib: Annotate `type.__prepare__` to return a mutable object

slanzmich:
During class creation, the namespace returned by `__prepare__` is populated when the class body is executed. Therefore, it must be mutable.

Calling `type.__prepare__` returns a `dict` instance, so I annotated it as such. Please let me know if a different type (e.g. `MutableMapping`) is preferred.


srittau:
Thanks! Considering [PEP 3155](https://peps.python.org/pep-3115/#invoking-the-metaclass) says the following, I think `MutableMapping` would be a better fit, at least for now:

> __prepare__ returns a dictionary-like object which is used to store the class member definitions during evaluation of the class body. In other words, the class body is evaluated as a function block (just like it is now), except that the local variables dictionary is replaced by the dictionary returned from __prepare__. This dictionary object can be a regular dictionary or a custom mapping type.

> This dictionary-like object is not required to support the full dictionary interface. A dictionary which supports a limited set of dictionary operations will restrict what kinds of actions can occur during evaluation of the class body. A minimal implementation might only support adding and retrieving values from the dictionary - most class bodies will do no more than that during evaluation. For some classes, it may be desirable to support deletion as well. Many metaclasses will need to make a copy of this dictionary afterwards, so iteration or other means for reading out the dictionary contents may also be useful.

If my understanding of this is correct, in an ideal world, `type` would be generic over the namespace returned by `__prepare__` and we'd use a minimal protocol for it, but this would a) require type var defaults and b) possibly introduce further complications without much gain.

slanzmich:
Thanks for pointing to [PEP 3155](https://peps.python.org/pep-3115/#invoking-the-metaclass), I only had a look at the [data model docs](https://docs.python.org/3/reference/datamodel.html), where I only saw less detail.

I changed the annotation to `MutableMapping`.

What should I do with the [comment above](#issuecomment-1836006526) about breaking something with pydantic? Should I also open a PR there? Their annotation follows the incorrect one here.

srittau (approved):

Issue #2333, opened 2018-07-14
Should HTTPMessage return `Optional[str]` for `__getitem__`?


asottile:
```pycon
>>> from urllib.request import urlopen
>>> urlopen('https://example.com')
<http.client.HTTPResponse object at 0x7fc608f39fd0>
>>> resp = urlopen('https://example.com')
>>> resp.headers['foo']
>>> resp.headers['connection']
'close'
>>> type(resp.headers['connection'])
<class 'str'>
```

```python
import urllib.request
from typing import Optional

resp = urllib.request.urlopen('https://example.com')
x: Optional[str] = resp.headers['connection']
```

```console
$ mypy t.py
t.py:5: error: Incompatible types in assignment (expression has type "Union[str, Header, None]", variable has type "Optional[str]")
```

If this sounds sane, I'd like to add `__getitem__` [here](https://github.com/python/typeshed/blob/cc45366ca5841153f3924037cb26c6dcfe648621/stdlib/3/http/client.pyi#L80)?

JelleZijlstra:
Does it never return `Header`? I think we discussed this in some issue before.

asottile:
I can't seem to find a situation where it returns `Header`, it's possible I'm overlooking something.

OJFord:
> In a model generated from bytes, any header values that (in contravention of the RFCs) contain non-ASCII bytes will, when retrieved through this interface, be represented as [Header](https://docs.python.org/3.5/library/email.header.html#email.header.Header) objects with a charset of unknown-8bit.

from the last paragraph of intro, just above https://docs.python.org/3.5/library/email.message.html#email.message.Message.__len__

asottile:
Interesting!  Though it doesn't appear to be true for `HTTPResponse`:

```bash
 while true; do echo -e "HTTP/1.1 200 OK\r\nContent-Length: 0\r\nWat: $(python -c 'print("\u2603")')\r\n" | nc -l localhost 1500; done
```

```pycon
$ python3
Python 3.6.5 (default, Apr  1 2018, 05:46:30) 
[GCC 7.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from urllib.request import urlopen
>>> x = urlopen('http://localhost:1500')
>>> x.headers
<http.client.HTTPMessage object at 0x7f51af37ae80>
>>> x.headers['Wat']
'â\x98\x83'
>>> type(x.headers['Wat'])
<class 'str'>
```

[...]

Issue #11094, opened 2023-12-01
Documenting the Any trick


Akuli:
We have explained the Any trick to contributors and users several times, usually related to `re` stubs, but also a few times in other cases, like #8069 and https://github.com/python/typeshed/issues/10564#issuecomment-1675792312.

Maybe we should explain it in CONTRIBUTING.md (or some other file) and comment usages of the Any trick, something like this?

```python
def foo(
    bla bla bla: BlaBlaBla,
    bla bla bla: BlaBlaBla,
) -> Foo | Any: ...  # "the Any trick", see CONTRIBUTING.md
```

Akuli:
Related: #7870

srittau:
Unlike #7870, we could just introduce a type alias for `Any` and call it `UnsafeNone` or some other kind of name that includes `None`?

Akuli:
I have mixed thoughts about the `UnsafeNone` idea.

On the one hand, seeing `re.Match | Any` in IDE autocompletions is worse than `re.Match[str] | UnsafeNone` for someone new to typing (or new to python) who just wants to parse some data with a regex. `TheAnyTrick[Foo]` would be much worse though, because with `re.Match | WhatEver` you can just ignore the stuff at the end (as people naturally do). Also, we might be lucky enough to not get stuck on naming.

On the other hand, the return type may actually be shown as `re.Match[str] | _typeshed.UnsafeNone`, and now the thing you really want is less than half of the type shown. Also, including `None` in the name of a type that doesn't actually use `None` seems weird, especially if we need to use the `Any` trick so that the rarely returned type isn't `None`. For example, a function almost always returns `Foo`, but can also return a `LegacyFoo` if a user has enabled `LegacyFoo` support by passing in `support_legacy_foo=True` somewhere.

For naming the type alias, I think we should think about people new to Python who just want to parse something with a regex. Do we want their IDE to show `re.Match[str] | UnsafeNone` or `re.Match[str] | MaybeNone` or `re.Match[str] | SometimesNone` or something else?

srittau:
> Do we want their IDE to show re.Match[str] | UnsafeNone or re.Match[str] | MaybeNone or re.Match[str] | SometimesNone or something else?

I believe that showing them something other than `Any` is an improvement. The reason `Any` is used is just unclear. An alias is at least a marker that something is going on here. When we also add a comment to the alias in `_typeshed`, when they press the "go to source" on the alias, they can see what exactly the problem is.

[...]

PR #11095, opened 2023-12-01
`email.message.Message`: Improve payload methods and Explain why `__getitem__` isn't typed to return None

Avasam:
Addresses part of #2333

From https://github.com/python/typeshed/issues/2333#issuecomment-1836676333:
> Taking a look at this test run: https://github.com/Avasam/typeshed/pull/36/files#diff-ced02b01b42b244e9b7bc2e12ef7a215bde89a29072c99ade903f58fb83a3bc7R36
> 
> spack would have to change their assert to first store the value in a local variable then assert that variable: https://github.com/spack/spack/blob/develop/lib/spack/spack/oci/oci.py#L138-L142
> 
> twine implicitely uses `header_factory`. https://github.com/pypa/twine/blob/main/twine/commands/check.py#L73C40-L73C54  They could replace with
> ```python
>     _, header = msg.policy.header_store_parse("content-type", value)`
>     return msg.get_content_type(), header.params
> ```
> 
> urllib3 is a false-positive, and an interesting one: https://github.com/urllib3/urllib3/blob/main/src/urllib3/contrib/emscripten/fetch.py#L382C24-L382C24

I don't think the false-positives a None union here would introduce is something we want. Especially when `.get` exists for that purpose.
I'm sure the explanation in my comment could be made clearer. Feel free to improve it.


hauntsaninja (commented):
Should these be using unions with Any? E.g. we definitely want to handle the payload case, so `get_payload() -> _PayloadType | Any` seems better

Avasam:
> Should these be using unions with Any? E.g. we definitely want to handle the payload case, so `get_payload() -> _PayloadType | Any` seems better

Let's try it, I can always revert it. (also relates to #11094 )

Avasam:
Looks like a true positive given that pip is passing around the base `Message` type ?

stdlib/email/message.pyi
```diff
@@ -24,14 +25,18 @@ class Message:
     def set_unixfrom(self, unixfrom: str) -> None: ...
     def get_unixfrom(self) -> str | None: ...
     def attach(self, payload: Message) -> None: ...
-    def get_payload(self, i: int | None = None, decode: bool = False) -> Any: ...  # returns _PayloadType | None
+    def get_payload(
```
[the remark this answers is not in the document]
Avasam: I think I got it all figured out :) This is more than I originally planned for this PR and doesn't address the linked issue, but sure, why not.

hauntsaninja (approved):
Nice! (I think we could have gotten away with nominal typing for set_payload, but duck typing can't hurt!)

stdlib/email/message.pyi
```diff
@@ -23,16 +33,40 @@ class Message:
...
     def set_charset(self, charset: _CharsetType) -> None: ...
     def get_charset(self) -> _CharsetType: ...
     def __len__(self) -> int: ...
     def __contains__(self, name: str) -> bool: ...
     def __iter__(self) -> Iterator[str]: ...
-    def __getitem__(self, name: str) -> _HeaderType: ...
-    def __setitem__(self, name: str, val: _HeaderType) -> None: ...
+    # Same as `get` with `failobj=None`, but with the expectation that it won't return None in most scenarios
+    # This is important for protocols using __getitem__, like SupportsKeysAndGetItem
+    def __getitem__(self, name: str) -> _HeaderType | Any: ...  # AnyOf[_HeaderType, None]
```
AlexWaygood, line 68: `_HeaderType` is already just a type alias to `Any`, no?

Avasam: Yes, but it's more about the semantics of what to expect for subclasses. And the assumption that *eventually* something will come in the type system to improve the union.

With #11094 this would be something like `_HeaderType | MaybeNone`
Which would preferably evaluate to something akin to `AnyOf[str, Header, None]` if the type system allowed us.

`HttpMessage`'s overrides should just copy `Message` and replace `_HeaderType` with str. Which would be: `def __getitem__(self, name: str) -> str | MaybeNone: ...` ("Any trick" alias name pending).

@srittau mentioned wanting an actual `str | None` union there. But I am hesitant as it leads to false positives with protocols using `__getitem__` that should otherwise be safe (see the `urllib3` use above of `dict(...)` where the parameter is a `SupportsKeysAndGetItem`.

Usage of the Any trick in `HttpMessage.__getitem__` does seem to bypass that kind of false-positives (quick test over at https://github.com/Avasam/typeshed/pull/37/files#diff-ced02b01b42b244e9b7bc2e12ef7a215bde89a29072c99ade903f58fb83a3bc7R68 )

---

I might combine this PR and #11114 to centralize discussion, I thought the `str` overrides in `HttpMessage` would be a more contentious point and I took a habit of splitting PRs in smaller chunks ^^

AlexWaygood: Hmm. But this just seems confusing to me -- I worry that others will come along and ask the question I just asked. Maybe just making the comment a bit longer would be better here?

```suggestion
    # Morally, the return type should be `AnyOf[_HeaderType, None]`,
    # which we could spell as `_HeaderType | Any`,
    # *but* `_HeaderType` itself is currently an alias to `Any`...
    def __getitem__(self, name: str) -> _HeaderType: ...
```

---

> I took a habit of splitting PRs in smaller chunks ^^

Splitting PRs into smaller chunks is definitely Good!!

Avasam: Clear and explicit. I like your suggestion

Avasam:
@AlexWaygood  @hauntsaninja  Is this OK to merge?

AlexWaygood:
I haven't looked too deeply at this (and while I'm happy to, I probably won't get to it in the next few days, unfortunately), but @hauntsaninja approved it earlier, and I trust both his and your judgement :)

Avasam:
I feel confortable with this. It's #11114 I'm less certain about. If something comes up it can be updated there or in a new PR.

asottile-sentry:
fwiw the annotations to `get_payload()` broke this usage here -- unclear how to proceed since the usage itself is safe (it's always `str`) but the typing now complains:

https://github.com/getsentry/sentry/blob/58dda0436f67e728eda3e1caad9766b3d6188b7a/src/sentry/runner/commands/sendmail.py#L15

```
src/sentry/runner/commands/sendmail.py:15: error: Argument "message" to "send_mail" has incompatible type "Message | str | list[Message | str] | Any"; expected "str"  [arg-type]
```

srittau:
@asottile-sentry I had a similar problem. #11574 should fix this by changing the argument type back to `Any`.

Avasam:
I think it should be safe to coerce to a string?
`Message`'s own string representation is just the string message. And whilst static type-checking can't infer whether you have a multipart message, I assume you already expect to always have a single-part there.

---

Alternatively, the middle overload
```py
    @overload  # either
    def get_payload(self, i: None = None, decode: Literal[False] = False) -> _PayloadType | _MultipartPayloadType | Any: ...
```
could be kept as only returning `Any` (too complex, until we get something like #566 or try to encode the single/multi-part with Generic typing, probably after #11422)

PR #11098, opened 2023-12-04
PurePath is not a PathLike subclass

tungol:
It matches the protocol, but it's not an explicit subclass. As far as I can see, it seems okay to just remove that.


tungol:
oh, I think I misunderstood the situation.

PR #11099, opened 2023-12-04
fix(psycopg2): fix copy_expert signature

mehdigmira:
Current copy_expert signature forces file-like object to be both read and write.
However according to pyscopg2, it's either one or the other https://www.psycopg.org/docs/cursor.html#cursor.copy_expert because you can either issue a `COPY FROM` or `COPY TO` statement


srittau (approved):
Thanks! Note that this makes `copy_expert()` unsafe, since the wrong copy statement could be used with the wrong type of file-like object. That said, the current annotations can lead to false negatives, which is why I think we should apply this fix. Leaving this open for a third opinion, though.

PR #11100, opened 2023-12-04
update two classes to use StrEnum as their base class, when available.

tungol:

Creating StrEnum in _typeshed avoids needing to repeat the enum, nearly unchanged, in both branches of the version check. I hope I haven't missed something.


tungol:
I forgot to pass in the allowlist to stubtest when I found this issue, and if I had realized it was on the allowlist I probably would have left it alone. But now that we're here, I think this is still valid?

srittau (approved):
Thanks!

PR #11107, opened 2023-12-06
Stubtest fixes for the latest patch release of py311

### LINKED ISSUE #11106: Daily tests failed on Wed Dec 06 2023

AlexWaygood:
Fixes #11106

### DIFF
--- a/stdlib/io.pyi
+++ b/stdlib/io.pyi
@@ -33,6 +33,9 @@ __all__ = [
 if sys.version_info >= (3, 8):
     __all__ += ["open_code"]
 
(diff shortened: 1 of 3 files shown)


hauntsaninja (approved):
Thanks! I hope we don't have the thing where Github gives us inconsistent Python 3.11 versions for a bit

AlexWaygood:
> Thanks! I hope we don't have the thing where Github gives us inconsistent Python 3.11 versions for a bit

I think I fixed that in https://github.com/python/typeshed/pull/10641! But let's see 🤞

PR #11108, opened 2023-12-06
Fix `pywin32`'s `PyCBitmap.GetBitmapBits` method

Avasam:
Tested locally. Ref: https://github.com/mhammond/pywin32/blob/main/Pythonwin/win32bitmap.cpp#L461

Technically accepts any int. But that's residual from Python 2 / C implementation.


stubs/pywin32/_win32typing.pyi
```diff
@@ -4801,7 +4801,10 @@ class PyCBitmap:
         arg1: tuple[Incomplete, Incomplete, Incomplete, Incomplete],
     ) -> None: ...
     def GetInfo(self): ...
-    def GetBitmapBits(self, asString: int = ...) -> str: ...
+    @overload
+    def GetBitmapBits(self, __asString: Literal[False] = False) -> tuple[int, ...]: ...
+    @overload
+    def GetBitmapBits(self, __asString: Literal[True]) -> bytes: ...
```
AlexWaygood, line 4807: Worth having a fallback overload, to avoid false positives if the type checker is unable to infer a `Literal`?

```suggestion
    def GetBitmapBits(self, __asString: Literal[True]) -> bytes: ...
    @overload
    def GetBitmapBits(self, __asString: bool) -> Any: ...
```

Avasam: Ah, thought it was managed by checkers for some reason.

Although I'm tempted to prefer the real union here: a typical use-case would use a literal. And if you take an unknown boolean parameter, you should really support both return types. There's only two possibilities here, so either you use a literal for the 1 suported case, or you support both.

```py
import win32ui
bitmap = win32ui.CreateBitmap()
# ...
def foo(as_bytes) -> bytes:
  return bitmap.GetBitmapBits(as_bytes) # This doesn't make sense
  
def foo(as_bytes) -> bytes: # Unused param
  return bitmap.GetBitmapBits(True) # This is more sensible
```

AlexWaygood: I'm okay with going with a union return type for the fallback overload here. After all, it's the fallback, we hope the type checker generally doesn't hit it :)

> thought it was managed by checkers for some reason.

Type checkers do "`Union` math" for overloads in some situations, but generally don't when it comes to overloads that use `Literal` types. I think there's a good reason for that, but I can't give it to you off the top of my head I'm afraid :)

AlexWaygood (approved):
From my reading of the C code, this looks to be correct!

PR #11110, opened 2023-12-06
In email.feedparser, BytesFeedParser is a subclass of FeedParser

tungol:


stdlib/email/feedparser.pyi
```diff
@@ -15,10 +15,10 @@ class FeedParser(Generic[_MessageT]):
...
 
-class BytesFeedParser(Generic[_MessageT]):
+class BytesFeedParser(FeedParser[_MessageT]):
     @overload
     def __init__(self: BytesFeedParser[Message], _factory: None = None, *, policy: Policy = ...) -> None: ...
     @overload
     def __init__(self, _factory: Callable[[], _MessageT], *, policy: Policy = ...) -> None: ...
-    def feed(self, data: bytes | bytearray) -> None: ...
+    def feed(self, data: bytes | bytearray) -> None: ...  # type: ignore[override]
     def close(self) -> _MessageT: ...
```
tungol, line 24: I think we need to keep the redefinition of __init__ even though it's not actually redefined in the subclass in cpython, due to the typing on the self argument.

That said, the redefinition of close() here could be removed. it's identical to that of the superclass. I wasn't sure whether that meant that it should be removed or not. In the end it kept the diff very simple to leave it alone, so that's what I did.

srittau: We generally remove redundant definitions in sub-classes, especially if they aren't defined in the implementation.

tungol: Thanks - I removed it.

PR #11112, opened 2023-12-06
Fix `aiofiles.os.wrap` definition origin

Avasam:
Both pyright an pylance were reporting issues with this import locally. Decided to look into it and noticed that `wrap` method was defined in the wrong file. And is probably not meant to be re-exported from `aiofiles.ospath`. It's not even documented in https://github.com/Tinche/aiofiles#usage , which gives credence to this being an internal wrapper function.

pyright:
```
E:\Users\Avasam\Documents\Git\typeshed\stubs\aiofiles\aiofiles\os.pyi
  E:\Users\Avasam\Documents\Git\typeshed\stubs\aiofiles\aiofiles\os.pyi:9:29 - error: "wrap" is unknown import symbol (reportGeneralTypeIssues)
  E:\Users\Avasam\Documents\Git\typeshed\stubs\aiofiles\aiofiles\os.pyi:9:37 - error: Type of "wrap" is unknown (reportUnknownVariableType)
```

Pylance:
image


stubs/aiofiles/@tests/stubtest_allowlist.txt
```diff
@@ -93,3 +93,6 @@ aiofiles.tempfile.temptypes.AsyncTemporaryDirectory.cleanup
 # Metaclass differs:
 aiofiles.base.AiofilesContextManager
 aiofiles.tempfile.AiofilesContextManagerTempDir
+
+# Internal wrapper, not meant to be re-exported
+aiofiles.ospath.wrap
```
Avasam, line 98: https://github.com/python/typeshed/actions/runs/7121014427/job/19389429006?pr=11112#step:5:61

I'm surprised that stubtest thinks this needs to be a re-exported symbol. Usually it only does that with unused imports.

AlexWaygood: Probably aiofiles is doing some really dynamic thing that screws up subtest's heuristics for figuring out if something was actually defined in this module, or originated from another module

AlexWaygood: The heuristics are here: https://github.com/python/mypy/blob/1cdeecdcc11115a3d15610801fdf91b0ed6bbe0a/mypy/stubtest.py#L376

srittau:
I'm not sure what the problem is here, but `wrap` is defined in `ospath` in the implementation, so the current stubs seem correct to me:

https://github.com/Tinche/aiofiles/blob/213ee8b325e09bb6026c2495cf5810023f342f3f/src/aiofiles/ospath.py#L7

Avasam:
> I'm not sure what the problem is here, but `wrap` is defined in `ospath` in the implementation, so the current stubs seem correct to me:

Uh, ok, keep this PR on hold, I should investigate why that's not the case for me locally. Might explain my surprise about stubtest result.

Could be a version thing, I'll update my local version tomorow and check.

srittau:
It was apparently moved with aiofiles 32.2.

PR #11114, opened 2023-12-06
override `http.HTTPMessage` methods to only use `str` for the header type

### LINKED ISSUE #2333: Should HTTPMessage return `Optional[str]` for `__getitem__`?

Avasam:
**Assuming** `HTTPMessage` *should* only work in `str` headers, which I'm not certain about. But either way would close #2333 .

Signature of the overridden methods based on #11095 .
An alternative to overriding is to make `email.message.Message` generic. Which would cause close to 100 changes in stdlib. Unless we defer until PEP 696 support.


stdlib/http/client.pyi
```diff
@@ -99,6 +99,23 @@ responses: dict[int, str]
 
 class HTTPMessage(email.message.Message):
     def getallmatchingheaders(self, name: str) -> list[str]: ...  # undocumented
+    # override below all of Message's methods that use `_HeaderType` / `_HeaderTypeParam` with `str`
+    # avoiding having to make Message generic
```
srittau, line 103: I would express this differently. It's not so much about avoiding making `Message` generic, but about the fact that `HTTPMessage` breaks the LSP and only accepts `str`.

stdlib/http/client.pyi
```diff
@@ -99,6 +99,23 @@ responses: dict[int, str]
...
+    def __getitem__(self, name: str) -> str: ...  # AnyOf[str, None]
```
srittau, line 104: `str | None` should be correct here. A header is not guaranteed to exist, so not being prepared for a `None` value is a strong indication of a bug.

srittau (commented):
I think the PR looks good overall. Limiting `HTTPMessage` to just `str` seems ok, especially given the primer output.

stdlib/http/client.pyi
Avasam:
```suggestion
    # override below all of Message's methods that use `_HeaderType` / `_HeaderTypeParam` with `str`
    # `HTTPMessage` breaks the Liskov substitution principle by only intending for `str` headers
    # This is easier than making `Message` generic
```

stdlib/http/client.pyi
Avasam: That being said, the Any trick could be usable here (and in the referenced PR). I should do a primer run with it.

srittau: I'm not sure I agree with that reasoning. While `get()` exists, `__getattr__` returns it as well. This is arguably a misfeature in the implementation (as it should just raise), but still that's something that a user should check against, because it absolutely can return `None`, even in unexpected scenarios.

srittau: I'm still not happy about this. This should be an error in my opinion:

```python
def parse_headers(headers: SupportsGetItem[str]):
    header_value = headers["Content-Type"]
    full_header = f"Content-Type: {header_value}"

parse_headers(HTTPMessage(...))
```

Users of `HTTPMessage` must be equipped to get `None` values from `__getitem__`.

stdlib/http/client.pyi
```diff
@@ -99,6 +99,24 @@ responses: dict[int, str]
 
 class HTTPMessage(email.message.Message):
     def getallmatchingheaders(self, name: str) -> list[str]: ...  # undocumented
+    # override below all of Message's methods that use `_HeaderType` / `_HeaderTypeParam` with `str`
+    # `HTTPMessage` breaks the Liskov substitution principle by only intending for `str` headers
+    # This is easier than making `Message` generic
+    def __getitem__(self, name: str) -> str | Any: ...  # See comment on `Message.__getitem__`
```
Avasam, line 105:
```suggestion
    def __getitem__(self, name: str) -> str | None: ...
```

srittau (approved):

Issue #7770, opened 2022-05-02
`getroot()` method of `xml.etree.ElementTree` return type should be nullable


andrea-guarino-sonarsource:
Hello,
I'm one of the maintainers of [Sonar Python](https://github.com/SonarSource/sonar-python) analyzer and we use typeshed for our type inference engine. (btw thanks a lot for maintaining typeshed!)

While investigating a false positive reported by a user, I noticed that that `getroot()` method of `ElementTree`class has return type annotation `-> Element` (see [here](https://github.com/python/typeshed/blob/4bcb1a95abb5986e67149f16fa4929a43b23eab9/stdlib/xml/etree/ElementTree.pyi#L197)).

I think `getroot()` return type annotation should be`ElementTree | None` instead, because it can returns None when root element is not provided:

```python
import xml.etree.ElementTree as ET

tree = ET.ElementTree()
print(tree.getroot()) # prints None
```

In fact, e.g. 3.10 implementation of ElementTree is defined as follows:
```python
class ElementTree:
    def __init__(self, element=None, file=None):
        # assert element is None or iselement(element)
        self._root = element # first node
        if file:
            self.parse(file)

    def getroot(self):
        """Return root element of this tree."""
        return self._root
```

I'm happy to open a PR if you think the reported issue is valid.

Thanks,
Andrea

srittau:
The problem with adding `None` to the return type is that users that know that a root exists (for example when parsing an ElementTree) have to do an extra check, something we like to avoid in typeshed. E.g.:

```python
import xml.etree.ElementTree as ET

tree = ET.parse()
root = tree.getroot()  # "root" is never None
```

Our usual solution would be to return `Element | Any` to at least somewhat reflect the fact that `getroot()` can not only return `Element`, but something else as well.

Obligatory comment that python/typing#566 would help with that as well.

linzouzou:
if "return Element | Any" were used, it's kind of overlapped since "Any" includes "Element".

Akuli:
Indeed. You would expect `Element | Any` to be same as just `Any`, but it isn't.

For example, consider this: (mostly copy/paste from xml module docs)

```python
import xml.etree.ElementTree as ET
tree = ET.parse('country_data.xml')
root = tree.getroot()
print(root.tagg)  # this line is a typo
```

- `Any` means "please do not complain" to type checkers. If `root` has type `Any`, you will no error for this.
[...]

PR #11117, opened 2023-12-07
Document "the Any trick"

Avasam:
Addresses the documentation part of #11094 .

Copied almost directly from https://github.com/python/typeshed/issues/7770#issuecomment-1836035325 just to get the ball rolling. 
Wordsmith away! I don't feel strongly about any particular description, as long as it ends up documented and explained!

I'm not including an alias here because getting "the Any trick" documented _**at all**_ is the most important part we all agreed on. I also noticed that it's not always clear **why** an `Any` union is used. So an alias for those cases should definitely help, but will require some thinking over changes. When creating the alias, it can immediately be used in stdlib as well. And a small mention can be added to the this here documentation. Something like:
```md
### "The `Any` trick" and `MaybeNone`
...

To help with semantics and avoid confusion, we defined an alias `MaybeNone` to be used in the same way. Indicating that the developper can be ready to deal with a potentially `None` value, and let type-checkers know that type-guarding on `None` is not redundant.
```


AlexWaygood:
+1 for adding docs first — it doesn't preclude adding an alias at a later date, and it's easier to get consensus on!

CONTRIBUTING.md
```diff
@@ -570,6 +570,26 @@ It's a "to do" item and should be replaced if possible. `Any` is used when
...
+### "The `Any` trick"
+
+Consider this: (mostly copy/paste from xml module docs)
+
+```python
+import xml.etree.ElementTree as ET
+tree = ET.parse('country_data.xml')
+root = tree.getroot()
+print(root.tagg)  # this line is a typo
+```
```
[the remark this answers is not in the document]
Avasam: I'm actually having a hard time coming up with a simple small example with `re.match` where it makes sense to return `re.Match[T] | Any` instead of just `re.Match[str]` if the string is already validated.

Like:
```py
import re

PATTERN = re.compile("test")

def match_my_pattern(value: str) -> re.Match[str]:
    assert value.startswith("test")
    return re.match(PATTERN, value)
```

I can't find any example of `-> Match[T] | Any` in typeshed either.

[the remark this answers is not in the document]
JelleZijlstra: Ah right, we use the trick for `Match.group`: https://github.com/python/typeshed/blob/acfdebfeb83c110d8ffefcfa7cc68bf3ab94817f/stdlib/re.pyi#L77

Not as prominent as `re.match` itself, but might still be a better example than xml.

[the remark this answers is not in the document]
Avasam: I think the explanation is now a lot clearer as-is. The code example could probably be improved.

Akuli (commented):
A pile of small nits that should make this easier to understand.

CONTRIBUTING.md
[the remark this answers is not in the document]
Akuli: Maybe we should go with something really practical for the example. For example, suppose you have a legacy system that for historical reasons has two kinds of user IDs. Old IDs look like `"legacy_userid_123"` and new IDs look like `"456_akuli"`. The function below extracts the number from a new ID and returns `None` if you give it a legacy ID.

```python
def parse_name_from_new_id(user_id: str) -> int | None:
    match = re.fullmatch(r"\d+_(.*)", user_id)
    if match is None:
        return None
    return match.group(1).capitalise()
```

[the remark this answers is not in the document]
Akuli: Here's another example, mostly taken from [editorconfig parsing code in one of my projects](https://github.com/Akuli/porcupine/blob/main/porcupine/plugins/editorconfig.py#L147-L164). It finds the 1 and 10 from a string that looks like like `"foo{1..10}.txt"`.

```python
match = re.fullmatch(r"\{(-?[0-9]+)\.\.(-?[0-9]+)\}", glob)
if match is not None:
    min_value = int(match.group(1))
    max_value = int(match.group(2))
```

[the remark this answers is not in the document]
Avasam: I'm less fond of your 2nd example, because it doesn't show that the trick still catches invalid attributes (due to a type for example).

I'll update your first example with the more obvious `uper` typo.

CONTRIBUTING.md
```diff
@@ -570,6 +570,40 @@ It's a "to do" item and should be replaced if possible. `Any` is used when
...
+
+* `-> Any` would mean "please do not complain" to type checkers. If `last_group` has type `Any`, you will get no error for this.
+* `-> str` would mean "will always be a `str`", which is wrong, and would cause type checkers to emit errors for code like `if last_group is None`.
+* `-> str | None` means "you must check for None", which is correct but can get annoying for some common patterns. Checks like `assert last_group is not None` would need to be added into various places only to satisfy type checkers, even when it is impossible to actually get a `None` value (type checkers aren't smart enough to know this).
+* `-> str | Any` means "must be prepared to handle a `str`". You will get an error for `last_group.uper`, because it is not valid when `last_group` is a `str`. But type checkers are happy with `if last_group is None` checks, because we're saying it can also be something else than an `str`.
+
+In typeshed we unofficially call returning `Foo | Any` "the Any trick". We tend to use it whenever something can be `None`, but requiring users to check for `None` would be more painful than helpful.
```
AlexWaygood, line 605: These lines are very long -- could you wrap to ~80 characters per line?

Avasam: Oh right, I got line-wrapping enabled for markdown so I don't notice ^^

CONTRIBUTING.md
```diff
@@ -570,6 +570,58 @@ It's a "to do" item and should be replaced if possible. `Any` is used when
...
+    match = re.fullmatch(r"\d+_(.*)", user_id)
+    if match is None:
+        return None
+    name_group = match.group(1)
+    return name_group.uper()  # This line is a typo (`uper` --> `upper`)
+```
+
+Regexes are often used so that the regex matches any string, or the string
+has already been validated so that the regex will match. But type checkers
+don't know this, so complaining about the `None` would get very annoying.
```
Akuli, line 605: The example is about whether a given group exists, not about whether the regex matches. (I probably misunderstood this at some point.)

The group thing isn't as obvious, so maybe we should explain it, something like this:

```suggestion
The `.group()` method returns `None` when the given group was not a part of the match.
For example, with a regex like `r"\d+_(.*)|legacy_userid_\d+"`, we would get a match whose `.group(1)` is `None` for the user ID is `"legacy_userid_7"`.
But here the regex is written so that the group always exists, and `match.group(1)` cannot return `None`.
Match groups are almost always used in this way.
```

CONTRIBUTING.md
```diff
@@ -570,6 +570,58 @@ It's a "to do" item and should be replaced if possible. `Any` is used when
...
+
+* `-> Any` would mean "please do not complain" to type checkers.
```
Akuli, line 607: This probably needs something to bring the reader's attention back to the return type.

```suggestion
Let's now consider typeshed's `-> str | Any` annotation of the `.group()` method:

* `-> Any` would mean "please do not complain" to type checkers.
```

CONTRIBUTING.md
```diff
@@ -570,6 +570,58 @@ It's a "to do" item and should be replaced if possible. `Any` is used when
 it's not possible to accurately type an item using the current type system.
 It should be used sparingly.
 
+### "The `Any` trick"
+
+Consider the following signature of `re.Match[str].group`:
```
Akuli, line 575:
```suggestion
Consider the following (simplified) signature of `re.Match[str].group`:
```

CONTRIBUTING.md
```diff
@@ -570,6 +570,58 @@ It's a "to do" item and should be replaced if possible. `Any` is used when
...
+
+```python
+class Match:
+  def group(self, __group: str | int) -> str | Any: ...
```
Akuli, line 579: Tiny nit: 4-space indentation.
```suggestion
    def group(self, __group: str | int) -> str | Any: ...
```

Akuli (commented):
This is pretty good. A few more nits.

CONTRIBUTING.md
Akuli: Not sure if this is a good idea. I'll let others decide.

Avasam: I was thinking about it too. It's a partial class signature. And only for one relevant overload of a method.

CONTRIBUTING.md
```diff
@@ -570,6 +570,61 @@ It's a "to do" item and should be replaced if possible. `Any` is used when
...
+* `-> str | Any` means "must be prepared to handle a `str`". You will get an
+  error for `name_group.uper`, because it is not valid when `name_group` is a
+  `str`. But type checkers are happy with `if name_group is None` checks,
+  because we're saying it can also be something else than an `str`.
+
+In typeshed we unofficially call returning `Foo | Any` "the Any trick".
+We tend to use it whenever something can be `None`,
+but requiring users to check for `None` would be more painful than helpful.
+
 ## Submitting Changes
```
Akuli, line 628: Maybe this isn't a good place for the `Any` trick. We should probably explain it towards the end, after more important sections (such as Submitting Changes).

Avasam: Any suggestion? I agree it's preferable to have more important sections first. But also "Incomplete" --> "Any Vs. incomplete" --> "Any trick" (just after "Conventions" and "Naming" seems like a nice coherent grouping of the Any type and its different meanings.

Akuli: I think it is fine here for now.

Akuli (approved):
Looks good to me, but another review would be good IMO.

PR #11118, opened 2023-12-07
Add deprecation classes in importlib.metadata

tungol:

Generally speaking these classes print warnings and then call super().method() without changing anything's type signature, so they don't need much to be filled out here.

### DIFF
--- a/stdlib/importlib/metadata/__init__.pyi
+++ b/stdlib/importlib/metadata/__init__.pyi
@@ -174,9 +174,7 @@ class FileHash:
     def __init__(self, spec: str) -> None: ...
 
 if sys.version_info >= (3, 12):
-    class DeprecatedNonAbstract:
-        def __new__(cls, *args: object, **kwargs: object) -> Self: ...
-
+    class DeprecatedNonAbstract: ...
     _distribution_parent = DeprecatedNonAbstract
 else:
     _distribution_parent = object
--- a/stdlib/importlib/metadata/__init__.pyi
+++ b/stdlib/importlib/metadata/__init__.pyi
@@ -1,15 +1,20 @@
 import abc
 import pathlib
 import sys
+from _collections_abc import dict_keys, dict_values
 from _typeshed import StrPath
-from collections.abc import Iterable, Mapping
+from collections.abc import Iterable, Iterator, Mapping
 from email.message import Message
 from importlib.abc import MetaPathFinder
 from os import PathLike
 from pathlib import Path
 from re import Pattern
-from typing import Any, ClassVar, NamedTuple, TypeVar, overload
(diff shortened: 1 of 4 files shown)


stdlib/importlib/metadata/__init__.pyi
```diff
@@ -69,8 +77,30 @@ class EntryPoint(_EntryPointBase):
     def __hash__(self) -> int: ...
     def __eq__(self, other: object) -> bool: ...
 
-if sys.version_info >= (3, 10):
-    class EntryPoints(list[EntryPoint]):  # use as list is deprecated since 3.10
+if sys.version_info >= (3, 12):
+    class EntryPoints(tuple[EntryPoint]):
```
AlexWaygood, line 81: `tuple[EntryPoint]` means "a `tuple` of length 1, where the sole element is an instance of `EntryPoint`". Is that what you want here, or did you maybe mean

```suggestion
    class EntryPoints(tuple[EntryPoint, ...]):
```

tungol: The latter. I've also got a stubtest failure, so I'll mark this as draft until I sort that out.

stdlib/importlib/metadata/__init__.pyi
```diff
@@ -68,9 +82,33 @@ class EntryPoint(_EntryPointBase):
 
     def __hash__(self) -> int: ...
     def __eq__(self, other: object) -> bool: ...
+    if sys.version_info >= (3, 11):
+        def __lt__(self, other: object) -> bool: ...
+    if sys.version_info < (3, 12):
+        def __iter__(self) -> Iterator[Any]: ...  # result of iter((str, Self)), really
```
srittau, line 88: So why not use `Iterator[tuple[str, Self]]`?

srittau (commented):
Thanks, LGTM, but one question.

stdlib/importlib/metadata/__init__.pyi
tungol: That's an iterator that returns `tuple[str, Self]`. This is an iterator that will always return two items, the first of which is str and the second of which is Self. As far as I know, there's no way to express that in the current system. We could do `Iterator[str | Self]`, but my understanding is that it's better to avoid union return types.

The implementation here is `return iter((self.name, self))`; I meant "iter()` as the runtime function, not a misspelling of the type, but that's a bit confusing because then I switched to the types of the values. Maybe the comment could be better worded, but I'm not sure how without getting too wordy. Might be better to just leave off the comment if it's too confusing. The function was removed in 3.12, so this is going away (eventually) anyway.

srittau (approved):

PR #11120, opened 2023-12-08
Bump various test dependencies

AlexWaygood:
Should hopefully unblock #11015 and #11014, due to the pytype bump!


AlexWaygood:
The stubtest failures on py312 are unrelated (they're due to the recent release of Python 3.12.1), and will be fixed by #11121

PR #11124, opened 2023-12-08
Remove write modes from importlib.resources.abc.Traversable

pganssle:
According to [the documentation for `Traversable`](https://docs.python.org/3/library/importlib.resources.abc.html#importlib.resources.abc.Traversable.open), it seems like `open` only supports read modes:

> mode may be ‘r’ or ‘rb’ to open as text or binary. Return a handle suitable for reading (same as [pathlib.Path.open](https://docs.python.org/3/library/pathlib.html#pathlib.Path.open)).

I am also not clear on whether they explicitly are suggesting that `rt` is not supported, since the documentation says it's the same as `pathlib.Path.open`, which says it's the same as `open` (which in turn supports `rt`). In practice all the `Traversable`s I know of support `rt`.


JelleZijlstra (approved):
Thanks. I wouldn't worry about `t` unless perhaps if someone explicitly asks for it; there may be third-party implementations that don't support it, and the `t` is redundant anyway in modern Python.

PR #11126, opened 2023-12-09
make pointer types in ctypes.wintypes subclasses instead of aliases

tungol:


stdlib/ctypes/wintypes.pyi
```diff
@@ -186,55 +186,53 @@ class WIN32_FIND_DATAW(Structure):
     cFileName: _CField[Array[WCHAR], str, str]
     cAlternateFileName: _CField[Array[WCHAR], str, str]
 
-# These pointer type definitions use _Pointer[...] instead of POINTER(...), to allow them
-# to be used in type annotations.
```
tungol, line 190: I removed the comment because it seemed redundant: POINTER() is used to define these at runtime, but it's a runtime function. That's not appropriate for use in typeshed by definition.

Avasam (approved):
This does look more correct than before:
```py
>>> ctypes.wintypes.PBOOL       
<class 'ctypes.wintypes.LP_c_long'>
>>> type(ctypes.wintypes.PBOOL)
<class '_ctypes.PyCPointerType'>
>>> ctypes.wintypes.PBOOL.__bases__       
(<class '_ctypes._Pointer'>,)
>>> type(ctypes.wintypes.PBOOL.__bases__[0]) 
<class '_ctypes.PyCPointerType'>
```

PR #11128, opened 2023-12-09
uWSGI: Use a `TypeVarTuple` in `uwsgidecorators.thread`

Daverball:
This addresses one of the FIXME comments (and a stubtest_allowlist entry) in types-uWSGI

### DIFF
--- a/stubs/uWSGI/@tests/stubtest_allowlist.txt
+++ b/stubs/uWSGI/@tests/stubtest_allowlist.txt
@@ -5,10 +5,3 @@
 uwsgidecorators.erlang
 # This should really only be internal API, so we don't export it
 uwsgidecorators.harakiri.real_call
-
-# Error: is inconsistent
-# ======================
-# This is a limitation of ParamSpec, we can't specify that the
-# the ParamSpec does have no keyword arguments, but we need the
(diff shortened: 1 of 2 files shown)


Daverball:
Looks like pytype isn't happy with the use of a `TypeVarTuple` in a generic class yet.

JelleZijlstra:
Is there a pytype issue we can refer to that tracks the necessary support?

AlexWaygood:
> Is there a pytype issue we can refer to that tracks the necessary support?

The issue is that pytype doesn't yet support classes being generic over `Unpack[*Ts]`. I don't think pytype has a more specific issue for that than https://github.com/google/pytype/issues/1525.

When we use `Unpack[*Ts]` for a parameter annotation, pytype just falls back to `Any`, but it doesn't yet do that if a class is generic over `Unpack[*Ts]`. I can think of reasons why that might be a little harder to do, but I don't know that it actually _is_ harder.

Daverball:
The reason it's harder is that the generic now needs to be able to accept multiple values instead of just one, as well as some of the other special rules, such as not requiring redundant brackets. So I can see why there is maybe no easy way to add partial support for it.

Avasam:

PR #11129, opened 2023-12-09
gevent: Use `TypeVarTuple` instead of old `ParamSpec` workaround.

Daverball:
This is going to have the same pytype issue as #11128 for `BaseServer`, but it's probably still not worth splitting the `BaseServer` change into its own PR.


AlexWaygood:
> This is going to have the same pytype issue as #11128 for `BaseServer`, but it's probably still not worth splitting the `BaseServer` change into its own PR.

Eh, I disagree. Let's make the changes that are possible now, and defer the changes that aren't possible. Incremental improvements will benefit the users of these stubs more than waiting until we can do everything all at once :)

PR #11132, opened 2023-12-09
libsass: Use `TypeVarTuple` in `sass.SassFunction`

Daverball:
Another place where I previously used `ParamSpec` as a band aid solution.

Requires `TypeVarTuple` in a generic class to stop crashing pytype.


Avasam:

srittau:
@Daverball This now has merge conflicts that I can't resolve in the web editor. Could you have a look?

Daverball:
It looks like the package was removed from typeshed: https://github.com/python/typeshed/issues/15410

So this can be closed

PR #11134, opened 2023-12-10
fix use of importlib.abc.Finder base class

tungol:
importlib.abc.Finder is removed in 3.12, but it's deprecated and no longer used as a base class starting in 3.10.


srittau (approved):
Thanks!

PR #11136, opened 2023-12-10
stdlib: Replace overload hack in `operator.itemgetter` by PEP 646 unpack

hamdanal:


stdlib/_operator.pyi
```diff
@@ -105,22 +106,10 @@ class attrgetter(Generic[_T_co]):
 
 @final
 class itemgetter(Generic[_T_co]):
-    # mypy lacks support for PEP 646 https://github.com/python/mypy/issues/12280
-    # So we have to define all of these overloads to simulate unpacking the arguments
     @overload
-    def __new__(cls, item: _T_co) -> itemgetter[_T_co]: ...
+    def __new__(cls, item: _T_co) -> itemgetter[_T_co]: ...  # type: ignore[overload-overlap]
```
hamdanal, line 110: This overload overlaps with the second with "0-length `*items`" but both mypy and pyright are able to pick this overload correctly in this case.

srittau: A slightly more correct way would be to add a second required item to the second overload, since the tuple variant always requires two items. Also, while we're here:

```suggestion
    def __new__(cls, __item: _T_co) -> itemgetter[_T_co]: ...  # type: ignore[overload-overlap]
```

stdlib/_operator.pyi
```diff
@@ -105,22 +106,10 @@ class attrgetter(Generic[_T_co]):
...
-    def __new__(cls, item: _T_co, __item2: _T_co, __item3: _T_co) -> itemgetter[tuple[_T_co, _T_co, _T_co]]: ...
-    @overload
-    def __new__(
-        cls, item: _T_co, __item2: _T_co, __item3: _T_co, __item4: _T_co
-    ) -> itemgetter[tuple[_T_co, _T_co, _T_co, _T_co]]: ...
-    @overload
-    def __new__(
-        cls, item: _T_co, __item2: _T_co, __item3: _T_co, __item4: _T_co, *items: _T_co
-    ) -> itemgetter[tuple[_T_co, ...]]: ...
+    def __new__(cls, item: _T_co, *items: Unpack[_Ts]) -> itemgetter[tuple[_T_co, Unpack[_Ts]]]: ...
```
srittau, line 112: Like this:

```suggestion
    def __new__(cls, __item1: _T_co, __item2: _T2_co, *items: Unpack[_Ts]) -> itemgetter[tuple[_T_co, _T2_co, Unpack[_Ts]]]: ...
```

stdlib/_operator.pyi
hamdanal, line 110:
```suggestion
    def __new__(cls, __item: _T_co) -> itemgetter[_T_co]: ...
```

stdlib/_operator.pyi
```diff
@@ -2,14 +2,15 @@ import sys
 from _typeshed import SupportsGetItem
 from collections.abc import Callable, Container, Iterable, MutableMapping, MutableSequence, Sequence
 from typing import Any, AnyStr, Generic, Protocol, SupportsAbs, TypeVar, overload
-from typing_extensions import ParamSpec, SupportsIndex, TypeAlias, final
+from typing_extensions import ParamSpec, SupportsIndex, TypeAlias, TypeVarTuple, Unpack, final
 
 _R = TypeVar("_R")
 _T = TypeVar("_T")
 _T_co = TypeVar("_T_co", covariant=True)
```
srittau, line 9:
```suggestion
_T_co = TypeVar("_T_co", covariant=True)
_T2_co = TypeVar("_T2_co", covariant=True)
```

stdlib/_operator.pyi
hamdanal: Indeed item is positional-only in the C implementation but not in the Python implementation. I applied this patch but with the type ignore comment removed

srittau (approved):
Thanks!

stdlib/_operator.pyi
```diff
@@ -105,22 +107,10 @@ class attrgetter(Generic[_T_co]):
...
-    def __new__(cls, item: _T_co, __item2: _T_co, __item3: _T_co) -> itemgetter[tuple[_T_co, _T_co, _T_co]]: ...
-    @overload
-    def __new__(
-        cls, item: _T_co, __item2: _T_co, __item3: _T_co, __item4: _T_co
-    ) -> itemgetter[tuple[_T_co, _T_co, _T_co, _T_co]]: ...
-    @overload
-    def __new__(
-        cls, item: _T_co, __item2: _T_co, __item3: _T_co, __item4: _T_co, *items: _T_co
-    ) -> itemgetter[tuple[_T_co, ...]]: ...
+    def __new__(cls, __item1: _T_co, __item2: _T2_co, *items: Unpack[_Ts]) -> itemgetter[tuple[_T_co, _T2_co, Unpack[_Ts]]]: ...
```
AlexWaygood, line 113: It's unclear to me here whether `_T_co` is scoped to the function or the class here. I.e., if you do `x = itemgetter(1, 2, 3)`, is the type of `x` `itemgetter[int]` (`_T_co` is solved as being scoped to the class, and `__item1` is an argument of type `int`, so therefore `x` is of type `itemgetter[int]`? Or is the type of `x` `itemgetter[tuple[int, int, int]]` (`_T_co` is solved as being scoped to the method).

It would be better to avoid using `_T_co` in this overload at all (use a different TypeVar instead), so that it is unambiguous whether the typevar is scoped to the function or the class.

hamdanal:
`attrgetter` also accepts a keyword `attr` in the stub but the C implementation only take positional arguments
```console
$ python -c 'import operator; operator.attrgetter(attr="a")'
TypeError: attrgetter() takes no keyword arguments
```
Should I send a PR fixing it?

stdlib/_operator.pyi
srittau: Good point, I missed that `_T_co` is also used for the class.

hamdanal: Good call. I guess covariance is meaningless for the typevars in this case so we can have something like:
```python
def __new__(cls, __item1: _T1, __item2: _T2, *items: Unpack[_Ts]) -> itemgetter[tuple[_T1, _T2, Unpack[_Ts]]]: ...
```
Or am I not understanding the situation of variance?

AlexWaygood: Yep, you're correct — variance is only meaningful for class-scoped typevars, not function-scoped typevars. Your solution looks correct!

stdlib/_operator.pyi
```diff
@@ -105,22 +108,10 @@ class attrgetter(Generic[_T_co]):
 
 @final
 class itemgetter(Generic[_T_co]):
-    # mypy lacks support for PEP 646 https://github.com/python/mypy/issues/12280
-    # So we have to define all of these overloads to simulate unpacking the arguments
     @overload
-    def __new__(cls, item: _T_co) -> itemgetter[_T_co]: ...
+    def __new__(cls, __item: _T_co) -> itemgetter[_T_co]: ...
```
AlexWaygood, line 112: Using `_T_co` in this overload has the same issues when it comes to ambiguous scoping -- though it's less clear how it could cause actual _problems_ when it comes to this overload, it would be nice to use a different TypeVar here as well, just to avoid any ambiguity: 

```suggestion
    def __new__(cls, __item: _T) -> itemgetter[_T]: ...
```

AlexWaygood (commented):
Thanks, this looks good -- I left an inline comment about one more nit, though it doesn't really matter that much.

In general, I'm _slightly_ confused about why we're bothering to make this class generic at all, actually. It doesn't seem like it being generic really does much for us; it's not really possible to express the semantics of this class properly right now, with our current type system. But hey.

AlexWaygood:
Maybe the best way to type this class would be to do something like this:

```py
_T = TypeVar("_T")
_R = TypeVar("_R")
_T1 = TypeVar("_T1")
_T2 = TypeVar("_T2")
_T3 = TypeVar("_T3")
_T4 = TypeVar("_T4")
_T5 = TypeVar("_T5")

@final
class itemgetter(Generic[_T, Unpack[_Ts]]):
    def __new__(cls, __item1: _T1, *items: Unpack[_Ts]) -> itemgetter[_T1, Unpack[_Ts]]: ...
    @overload
    def __call__(self: itemgetter[_T1], obj: SupportsGetItem[_T1, _R]) -> _R: ...
    @overload
    def __call__(self: itemgetter[_T1, _T2], obj: SupportsGetItem[_T1 | _T2, _R]) -> tuple[_R, _R]: ...
    @overload
    def __call__(self: itemgetter[_T1, _T2, _T3], obj: SupportsGetItem[_T1 | _T2 | _T3, _R]) -> tuple[_R, _R, _R]: ...
    @overload
    def __call__(
        self: itemgetter[_T1, _T2, _T3, _T4], obj: SupportsGetItem[_T | _T2 | _T3 | _T4, _R]
    ) -> tuple[_R, _R, _R, _R]: ...
    @overload
    def __call__(
        self: itemgetter[_T1, _T2, _T3, _T4, _T5], obj: SupportsGetItem[_T1 | _T2 | _T3 | _T4 | _T5, _R]
    ) -> tuple[_R, _R, _R, _R, _R]: ...
    @overload
    def __call__(self, obj: SupportsGetItem[Any, Any]) -> Any: ...
```

This will crash pytype right now, but I'd actually be curious to see the mypy_primer output, if you fancy trying it!

hamdanal:
> Thanks, this looks good -- I left an inline comment about one more nit, though it doesn't really matter that much.
> 
> In general, I'm _slightly_ confused about why we're bothering to make this class generic at all, actually. It doesn't seem like it being generic really does much for us; it's not really possible to express the semantics of this class properly right now, with our current type system. But hey.

I asked myself the same question and I thought it was either used before or is intended to be used in the future so I didn't bother make the class non-generic because I don't have the context for why it was made generic in the first place.

AlexWaygood (approved):
There's definitely some big TODOs remaining with this class, but this seems like it maintains the status quo for users while simplifying the stub. So I'm happy to merge this now. Thanks!

PR #11137, opened 2023-12-10
Make multiprocessing pipes generic

Avasam:
A while ago I was toying with the idea of using multiprocessing for a long-running task, and got annoyed that I could not specify a type for my interfacing pipes. So I hashed up a quick generic multiprocessing stub for my need.

I didn't end up going that route, but since I already did some typing work, figure I'd publish it rather than throw it away.
Idk if the TypeVar variance is correct, please advise.


stdlib/multiprocessing/connection.pyi
```diff
@@ -1,17 +1,20 @@
...
 from typing_extensions import Self, SupportsIndex, TypeAlias
 
 __all__ = ["Client", "Listener", "Pipe", "wait"]
 
 # https://docs.python.org/3/library/multiprocessing.html#address-formats
 _Address: TypeAlias = str | tuple[str, int]
 
-class _ConnectionBase:
+_T1 = TypeVar("_T1")
+_T2 = TypeVar("_T2")
```
srittau, line 15: Calling these `_S` and `_R` or even `_SendT` and `_ReceiveT` might be better for documentation purposes.

stdlib/multiprocessing/connection.pyi
```diff
@@ -22,54 +25,54 @@ class _ConnectionBase:
...
     def recv_bytes_into(self, buf: Any, offset: int = 0) -> int: ...
-    def recv(self) -> Any: ...
+    def recv(self) -> _T2: ...
     def poll(self, timeout: float | None = 0.0) -> bool: ...
     def __enter__(self) -> Self: ...
     def __exit__(
-        self, exc_type: type[BaseException] | None, exc_value: BaseException | None, exc_tb: types.TracebackType | None
+        self, exc_type: type[BaseException] | None, exc_value: BaseException | None, exc_tb: TracebackType | None
     ) -> None: ...
-    def __del__(self) -> None: ...
```
srittau, line 34: We need to keep this.

stdlib/multiprocessing/connection.pyi
```diff
@@ -22,54 +25,54 @@ class _ConnectionBase:
...
 if sys.platform == "win32":
-    class PipeConnection(_ConnectionBase): ...
+    class PipeConnection(_ConnectionBase[_T1, _T2]): ...
 
 class Listener:
     def __init__(
         self, address: _Address | None = None, family: str | None = None, backlog: int = 1, authkey: bytes | None = None
     ) -> None: ...
-    def accept(self) -> Connection: ...
+    def accept(self) -> Connection[Incomplete, Incomplete]: ...
```
srittau, line 47: This is `Incomplete` instead of `Any` since you'd like to make `Listener` generic as well?

stdlib/multiprocessing/connection.pyi
```diff
@@ -22,54 +25,54 @@ class _ConnectionBase:
...
 
-def answer_challenge(connection: Connection, authkey: bytes) -> None: ...
+def answer_challenge(connection: Connection[Any, Any], authkey: bytes) -> None: ...
 def wait(
-    object_list: Iterable[Connection | socket.socket | int], timeout: float | None = None
-) -> list[Connection | socket.socket | int]: ...
-def Client(address: _Address, family: str | None = None, authkey: bytes | None = None) -> Connection: ...
+    object_list: Iterable[Connection[_T1, _T2] | socket.socket | int], timeout: float | None = None
+) -> list[Connection[_T1, _T2] | socket.socket | int]: ...
+def Client(address: _Address, family: str | None = None, authkey: bytes | None = None) -> Connection[Incomplete, Incomplete]: ...
```
srittau, line 69: I don't think we (currently) have any way to annotate `Connection` correctly, so I think `Connection[Any, Any]` would be correct here.

stdlib/multiprocessing/connection.pyi
```diff
@@ -22,54 +25,54 @@ class _ConnectionBase:
...
 # N.B. Keep this in sync with multiprocessing.context.BaseContext.Pipe.
 # _ConnectionBase is the common base class of Connection and PipeConnection
 # and can be used in cross-platform code.
 if sys.platform != "win32":
-    def Pipe(duplex: bool = True) -> tuple[Connection, Connection]: ...
+    def Pipe(duplex: bool = True) -> tuple[Connection[_T1, _T2], Connection[_T2, _T1]]: ...
 
 else:
-    def Pipe(duplex: bool = True) -> tuple[PipeConnection, PipeConnection]: ...
+    def Pipe(duplex: bool = True) -> tuple[PipeConnection[_T1, _T2], PipeConnection[_T2, _T1]]: ...
```
srittau, line 78: As seen in the primer hit (and the tests), this means that the return types of `Pipe()` need to be annotated now, which is a bit cumbersome.

srittau (commented):
I'm undecided on whether this is a good idea or not. I like the added type safety, but especially the `Pipe` case seems a bit cumbersome. Maybe this should wait for PEP 696?

But some comments below.

stdlib/multiprocessing/connection.pyi
AlexWaygood: I'm not sure it's actually valid to use a TypeVar like this, where it's entirely scoped within a return annotation. IIRC type checkers special-case Callable so that you can do things like `def x() -> Callable[[T], T]`, which comes up a lot. But in general, I don't think there's any supported way of spelling what you're trying to express here, unfortunately :(

stdlib/multiprocessing/connection.pyi
Avasam: Oops. Didn't notice that. Did not meant do delete `__del__`

stdlib/multiprocessing/connection.pyi
Avasam: Yeah this could be improved further, rather than "too complex for the type system" or "you can truly send anything through the connection". Just not something I care to do in the scope of making `Pipe` generic.

stdlib/multiprocessing/connection.pyi
Avasam: The methods `answer_challenge` and `deliver_challenge` in `Client` also don't use `send` and `recv`. So this is completely up to the user's annotations. (ie: `test: Connection[str, str] = Client("")`)

stdlib/multiprocessing/connection.pyi
Avasam: As shown in the tests. Mypy doesn't support this to restrict the generics between each other (but it also doesn't fail). Pyright supports it, but idk if intentionally or not.
I don't see anything about it either in https://peps.python.org/pep-0695/ or https://peps.python.org/pep-0696/
I was hoping `def Pipe[_T1, _T2](duplex: bool = True) -> tuple[PipeConnection[_T1, _T2], PipeConnection[_T2, _T1]]: ...` would eventually be supported syntax.

AlexWaygood: I'd rather not do this for now. It's an underspecified area of the spec, and we can see from the primer hits that mypy can't currently solve the TypeVars if we do things like this, causing false positives in user code

Avasam: I've moved the expectation of inverting the generic parameters, from the type definition, to a comment.

stdlib/multiprocessing/connection.pyi
AlexWaygood, line 1: Is the variance of the TypeVars correct here? It looks like maybe `_SendT` could be contravariant, and `_RecvT` covariant?

Avasam: > Idk if the TypeVar variance is correct, please advise.
🤷 variance with Python's TypeVars is still something I get highly confused about and has not "clicked" at all for me.

AlexWaygood: I think anybody who claims to be able to think about generics in Python without their head hurting sometimes is lying to you :)

In general, a good heuristic is that if the TypeVar only appears in return annotations, it should be covariant; if it only appears in parameter annotations, it should be contravariant; if it appears in both, it has to be invariant

Avasam: Noted and bookmarked. It may not always apply, but heuristics I can follow.

Akuli: I have explained variance on typeshed a few times, and I'm too lazy to find a previous explanation, so here goes again :)

Most of the time you only need invariant (default) and covariant TypeVars. To distinguish them, think about `list`s and `tuple`s with `int` and `float` items. A `list[int]` is not a valid `list[float]`, because we don't want to allow this:

```
foo: list[int] = [1, 2, 3]
bar: list[float] = foo
bar.append(12.34)
```

We communicate this to type checkers by using an invariant TypeVar in `list`. It means that the item types must match exactly for lists to be compatible.

On the other hand, a `tuple[int, ...]` is also a valid `tuple[float, ...]`, so the TypeVar in `tuple` is covariant. A tuple of a more specific type is also a valid tuple of a more general type.

Contravariance comes up less often, but it's basically the opposite of covariant: making `T` more general actually makes the generic type more specific. Consider `Callable[[T], None]` for example. A `Callable[[float], object]` (think `time.sleep`) is also a valid `Callable[[int], object]`.

srittau:

stdlib/concurrent/futures/process.pyi
```diff
@@ -19,8 +19,9 @@ _global_shutdown: bool
 
 class _ThreadWakeup:
     _closed: bool
-    _reader: Connection
-    _writer: Connection
+    # Any: Unused send and recv methods
+    _reader: Connection[Any, Any]
+    _writer: Connection[Any, Any]
```
AlexWaygood, line 24: Maybe these should be `Connection[object, object]` or `Connection[Unused, Unused]`? Otherwise we might get complaints from the (very small) group of people who use mypy's `--disallow-any-expr` optional lint: https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-disallow-any-expr

AlexWaygood (approved):

stdlib/concurrent/futures/process.pyi
Avasam: For the first, `_SendT`, contravariant, used for params argument sure.
Idk about the second, `_RecvT`, covariant, used for return type.

```python
reader1: Connection[int, object] = _ThreadWakeup()._reader  # OK
reader2: Connection[object, int] = _ThreadWakeup()._reader  # "Connection[object, object]" cannot be assigned to declared type "Connection[object, int]"

class MyThreadWakeup(_ThreadWakeup):
    _reader: Connection[int, int]  # OK
```

Arguably this is also a "private" name and I fail to come up with a real-world example

AlexWaygood: One option is to use the stricter non-Any types for now, and see if anybody complains about false positives... But I'm also fine with the PR being merged as-is; I don't have a strong opinion here :-)

PR #11138, opened 2023-12-10
gevent: Use a `TypeVarTuple` in `gevent.baseserver.BaseServer`

Daverball:
Second part of the gevent changes, that cannot be merged yet.

Requires `TypeVarTuple` in a generic class to stop crashing pytype.


Avasam:

srittau:
This has a conflict now.

Daverball:
Not really sure what the new stubtest failure is about, the part it's complaining about hasn't changed.

brianschubert:
The stubtest failures look potentially related to the changes to `_socket.socket.__init__` in https://github.com/python/typeshed/pull/14503, cc @donBarbos

Daverball:
> The stubtest failures look potentially related to the changes to `_socket.socket.__init__` in #14503, cc @donBarbos

That looks to be the case. I think I will add a stubtest ignore for this, with the same justification @donBarbos gave for choosing that default value, the `-1` is essentially an implementation detail, the documented default is `0` and only when you pass a `fileno` without specifying `proto` can you end up with something different.

With gevent stubtest knows about this white lie, since their cooperative wrapper around `_socket.socket` is written in pure Python, rather than C, so it knows about the exact default values in the implementation.

PR #11146, opened 2023-12-11
add email._policybase

tungol:

### DIFF
--- a/stdlib/email/_policybase.pyi
+++ b/stdlib/email/_policybase.pyi
@@ -20,12 +20,12 @@ class Policy(_PolicyBase, metaclass=ABCMeta):
     def __init__(
         self,
         *,
-        max_line_length: int | None = ...,
-        linesep: str = ...,
-        cte_type: str = ...,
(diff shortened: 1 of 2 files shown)


stdlib/email/_policybase.pyi
```diff
@@ -0,0 +1,51 @@
...
+    message_factory: Callable[[Policy], Message] | None
+    def __init__(
+        self,
+        *,
+        max_line_length: int | None = ...,
+        linesep: str = ...,
+        cte_type: str = ...,
+        raise_on_defect: bool = ...,
+        mangle_from_: bool = ...,
+        message_factory: Callable[[Policy], Message] | None = ...,
```
srittau, line 28:
```suggestion
        max_line_length: int | None = 78,
        linesep: str = "\n",
        cte_type: str = "8bit",
        raise_on_defect: bool = False,
        mangle_from_: bool = False,
        message_factory: Callable[[Policy], Message] | None = None,
```

srittau (approved):
LGTM, but we could add defaults.

PR #11147, opened 2023-12-11
add xml.sax._exceptions module

tungol:


stdlib/xml/sax/_exceptions.pyi
```diff
@@ -0,0 +1,19 @@
+from typing import Any, NoReturn
+from xml.sax.xmlreader import Locator
+
+class SAXException(Exception):
+    def __init__(self, msg: str, exception: Exception | None = None) -> None: ...
+    def getMessage(self) -> str: ...
+    def getException(self) -> Exception: ...
+    def __getitem__(self, ix: Any) -> NoReturn: ...
```
srittau, line 8:
```suggestion
    def __getitem__(self, ix: object) -> NoReturn: ...
```

stdlib/xml/sax/_exceptions.pyi
```diff
@@ -0,0 +1,19 @@
+from typing import Any, NoReturn
```
srittau, line 1:
```suggestion
from typing import NoReturn
```

srittau (approved):
LGTM, but while we're here we could update one annotation, see below.

tungol:
Makes sense to me

PR #11148, opened 2023-12-11
follow implementation details of dis.Instruction

tungol:
related (vaguely) to https://github.com/python/typeshed/issues/3968

### DIFF
--- a/stdlib/dis.pyi
+++ b/stdlib/dis.pyi
@@ -48,7 +48,7 @@ if sys.version_info >= (3, 11):
         end_col_offset: int | None = None
 
 if sys.version_info >= (3, 11):
-    class Instruction(NamedTuple):
+    class _Instruction(NamedTuple):
         opname: str
         opcode: int
         arg: int | None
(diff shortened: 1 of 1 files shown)


stdlib/dis.pyi
```diff
@@ -70,6 +70,9 @@ else:
         starts_line: int | None
         is_jump_target: bool
 
+class Instruction(_Instruction):
+    def _disassemble(self, lineno_width: int = ..., mark_as_current: bool = ..., offset_width: int = ...) -> str: ...
```
srittau, line 74:
```suggestion
class Instruction(_Instruction):
    def _disassemble(self, lineno_width: int = 3, mark_as_current: bool = False, offset_width: int = 4) -> str: ...
```

srittau (changes requested):

stdlib/dis.pyi
tungol: Thanks, I wasn't sure about preferred style for that.

PR #11150, opened 2023-12-11
add zipfile._path

tungol:

zipfile.Path is unfortunately duplicated in this MR; the alternative would be to define it in zipfile._path and import it into zipfile for all python versions, not just 3.12.


srittau (approved):
Not pretty, but probably the best we can do.

PR #11151, opened 2023-12-11
add pyclbr._Object

tungol:

### DIFF
--- a/stdlib/pyclbr.pyi
+++ b/stdlib/pyclbr.pyi
@@ -3,7 +3,15 @@ from collections.abc import Sequence
 
 __all__ = ["readmodule", "readmodule_ex", "Class", "Function"]
 
-class Class:
+class _Object:
+    if sys.version_info >= (3, 10):
+        def __init__(
+            self, module: str, name: str, file: str, lineno: int, end_lineno: int | None, parent: _Object | None
+        ) -> None: ...
+    else:
+        def __init__(self, module: str, name: str, file: str, lineno: int, parent: _Object | None) -> None: ...
+
+class Class(_Object):
     module: str
     name: str
     super: list[Class | str] | None
@@ -34,7 +42,7 @@ class Class:
             self, module: str, name: str, super: list[Class | str] | None, file: str, lineno: int, parent: Class | None = None
         ) -> None: ...
 
-class Function:
+class Function(_Object):
     module: str
     name: str
     file: int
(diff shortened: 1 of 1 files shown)


stdlib/pyclbr.pyi
```diff
@@ -1,19 +1,30 @@
...
-    super: list[Class | str] | None
-    methods: dict[str, int]
     file: int
     lineno: int
 
     if sys.version_info >= (3, 10):
         end_lineno: int | None
 
+    parent: _Object | None
+    children: Mapping[str, _Object]
```
tungol, line 16: I'm not totally sure about using Mapping here. Mypy was unhappy with dict because of variance issues, and this does fix that, but normally I'd only reach for Mapping as an argument type

srittau: Sounds like `Mapping` is the best idea in this case. I think we have other, similar examples in typeshed. Could you add comment to the attribute?

stdlib/pyclbr.pyi
```diff
@@ -1,19 +1,30 @@
...
+
+    if sys.version_info >= (3, 10):
+        def __init__(
+            self, module: str, name: str, file: str, lineno: int, end_lineno: int | None, parent: _Object | None
+        ) -> None: ...
+    else:
+        def __init__(self, module: str, name: str, file: str, lineno: int, parent: _Object | None) -> None: ...
+
+class Class(_Object):
```
srittau, line 25: Nit: Could you move this behind `Function` to match the implementation?

stdlib/pyclbr.pyi
```diff
@@ -1,19 +1,30 @@
...
+    super: list[Class | str] | None
```
srittau, line 26: I know that you only moved it here: But this will have similar issues like `children` above, especially in the constructor. (I.e. `super = ["foo"]; Class(..., super, ...)` will fail.) We should probably use `Sequence` here and in the constructor, but no need to fix this in this PR, as it hasn't made problems so far.

srittau (changes requested):
Thanks, two nits (and one comment) below.

stdlib/pyclbr.pyi
tungol: Done

stdlib/pyclbr.pyi
tungol: Done

srittau (approved):

PR #11152, opened 2023-12-11
fix inheritance for urllib.response.addbase

tungol:


srittau (approved):
Much better, thanks!

PR #11155, opened 2023-12-12
add _ssl module

tungol:
Really all I needed for fixing the inheritance was _ssl._SSLContext. But then I needed all the other stuff in _ssl, and if I was doing that I wanted to do a thorough job of it.

Motivation was originally related to https://github.com/python/typeshed/issues/3968 , but we're well beyond that now, really.


stdlib/_ssl.pyi
```diff
@@ -0,0 +1,344 @@
...
+    notBefore: str
+    notAfter: str
+    subjectAltName: NotRequired[tuple[tuple[str, str], ...] | None]
+    OCSP: NotRequired[tuple[str, ...] | None]
+    caIssuers: NotRequired[tuple[str, ...] | None]
+    crlDistributionPoints: NotRequired[tuple[str, ...] | None]
+
+class _EmptyCertInfo(TypedDict): ...
+
+def _test_decode_cert(__path: str) -> _CertInfo: ...
```
AlexWaygood, line 51: In general, we prefer not to add private functions/attributes/classes unless somebody specifically needs them for something. This helps reduce the maintenance burden for us (since these are usually implementation details that often change in patch releases), and means that IDEs are less likely to have their autocomplete suggestions cluttered with these usually-undocumented internals. If a private class is used as a base class for a public class, then we've obviously got no choice except to include the private class in the stub -- but it looks like there's a few instances in this PR (like this function here) where you're adding private things that _maybe_ aren't really necessary :)

AlexWaygood:
> Motivation was originally related to #3968

Thanks so much for this series of PRs, by the way, I'm a big fan! Great to see all this stuff being cleaned up.

stdlib/_ssl.pyi
tungol: That makes sense; I'll do some cleaning up.

stdlib/_ssl.pyi
```diff
@@ -0,0 +1,297 @@
...
+    time: int
+    timeout: int
+    def __eq__(self, other: object) -> bool: ...
+    def __ge__(self, other: object) -> bool: ...
+    def __gt__(self, other: object) -> bool: ...
+    def __le__(self, other: object) -> bool: ...
+    def __lt__(self, other: object) -> bool: ...
+    def __ne__(self, other: object) -> bool: ...
+
+# _ssl.Certificate is weird: it can't be instantiated or subclassed.
```
tungol, line 133: @AlexWaygood I removed _ssl._SSLObject which underlies ssl.SSLObject but isn't a formal base class for it. Doing so leaves _ssl.Certificate a little stranded.

It don't have an underscore prefix, so stubcheck complains if it's absent, but the only possible way to create an instance is via a method of _ssl._SSLSocket. Part of the reason I fleshed out the stubs for _SSLSocket in the first place was that I was trying to figure how to generate an instance of Certificate in order to observe its behavior.

I waffled on leaving the _SSLSocket stubs in, but I figured this comment would have the same effect without (as much) maintenance burden. On the other hand, that's probably outside the goals of typeshed, so I understand if it'd be better to remove the comment and just let Certificate be an oddity with no clear explanation.

stdlib/ssl.pyi
```diff
@@ -4,30 +4,48 @@ import sys
 from _typeshed import ReadableBuffer, StrOrBytesPath, WriteableBuffer
 from collections.abc import Callable, Iterable
 from typing import Any, NamedTuple, overload
-from typing_extensions import Literal, Never, Self, TypeAlias, TypedDict, final
+from typing_extensions import Literal, Never, Self, TypeAlias
+
+from _ssl import (
```
AlexWaygood, line 9: Could you add `"_ssl"` to the table here, so that ruff puts the import in the right place?

https://github.com/python/typeshed/blob/3171cb44aa9d7c7ada31ef78720b42ea28acd24a/pyproject.toml#L50-L93

stdlib/ssl.pyi
```diff
@@ -1,33 +1,50 @@
...
+    SSLSession as SSLSession,
+    _PasswordType as _PasswordType,  # typeshed only, but re-export for other type stubs to use
+    _SSLContext,
+)
 from _typeshed import ReadableBuffer, StrOrBytesPath, WriteableBuffer
 from collections.abc import Callable, Iterable
-from typing import Any, Literal, NamedTuple, TypedDict, final, overload
-from typing_extensions import Never, Self, TypeAlias
+from typing import Any, NamedTuple, TypedDict, final, overload
+from typing_extensions import Literal, Never, Self, TypeAlias
```
JelleZijlstra, line 31:
```suggestion
from typing import Any, Literal, NamedTuple, TypedDict, final, overload
from typing_extensions import Never, Self, TypeAlias
```
Bad merge from me

JelleZijlstra (approved):

PR #11158, opened 2023-12-13
add unittest.case._AssertRaisesBaseContext class

tungol:


stdlib/unittest/case.pyi
```diff
@@ -27,6 +27,23 @@ DIFF_OMITTED: str
...
+    def __init__(
+        self,
+        expected: type[BaseException] | tuple[type[BaseException], ...],
+        test_case: TestCase,
+        expected_regex: str | Pattern[str] | None = None,
+    ) -> None: ...
+
+    # This returns Self if args is the empty list, and None otherwise.
+    # but it's not possible to construct an overload which expresses that
+    def handle(self, name: str, args: list[Any], kwargs: dict[str, Any]) -> Any: ...
```
tungol, line 45: I'm about 90% sure this comment is accurate, but if I'm wrong and there's a way to do this overload, I'd be glad to hear it.

stdlib/unittest/_log.pyi
```diff
@@ -2,15 +2,15 @@ import logging
...
+from unittest.case import TestCase, _AssertRaisesBaseContext
 
 _L = TypeVar("_L", None, _LoggingWatcher)
 
 class _LoggingWatcher(NamedTuple):
     records: list[logging.LogRecord]
     output: list[str]
 
-class _AssertLogsContext(Generic[_L]):
+class _AssertLogsContext(_AssertRaisesBaseContext, Generic[_L]):
```
srittau, line 13: In Python 3.11 and 3.12 at least, this derives directly from `_BaseTestCaseContext`, not `_AssertRaisesBaseContext`.

stdlib/unittest/case.pyi
```diff
@@ -27,6 +27,23 @@ DIFF_OMITTED: str
 class _BaseTestCaseContext:
     def __init__(self, test_case: TestCase) -> None: ...
```
srittau, line 28:
```suggestion
class _BaseTestCaseContext:
    test_case: TestCase
    def __init__(self, test_case: TestCase) -> None: ...
```

stdlib/unittest/case.pyi
```diff
@@ -27,6 +27,23 @@ DIFF_OMITTED: str
...
 
+class _AssertRaisesBaseContext(_BaseTestCaseContext):
+    expected: type[BaseException] | tuple[type[BaseException], ...]
+    expected_regex: str | Pattern[str] | None
```
srittau, line 32: This is compiled in `__init__`:

```suggestion
    expected_regex: Pattern[str] | None
```

srittau (changes requested):
Thanks, comments below.

stdlib/unittest/case.pyi
tungol: good catch - I can remove `test_case: TestCase` from some of the child classes with that too.

stdlib/unittest/_log.pyi
tungol: It's that way on all versions - best I can figure is a copy/paste error of some kind. Fixed now, at any rate.

srittau (approved):

Issue #11141, opened 2023-12-10
What to do when the runtime lies?


tungol:
I've been cleaning up inheritance mismatches. I've found that several modules with a C implementation fudge the module name of their classes. For example:

`ReferenceType` is a class defined in the `_weakref` C implementation, and then imported from there into the `weakref` module. Because it's a C implementation, it's free to declare it's own qualified name, and calls itself `weakref.ReferenceType`: https://github.com/python/cpython/blob/main/Objects/weakrefobject.c#L368C1-L368C1

The intention, I believe, being that this class canonically lives in the `weakref` module, and the fact that it's being defined in `_weakref` is an implementation detail which is hidden at runtime. Typeshed declares the class in `_weakref.pyi`, which means it gets a qualified name of `_weakref.ReferenceType`

I can see two possible arguments:

 - Typeshed should continue to follow the layout of implementation over the declared implementation. e.g. `ReferenceType` is defined in `_weakref` originally, so that's the correct place for it.

OR 

 - Typeshed should follow the declared implementation over the layout. e.g. `ReferenceType` considers itself to live in the `weakref` module, that's where the majority of users will get it from, so that's the correct place for it.

This pattern pops up in a lot of places: in `_ast`, `_decimal`, and `_sqlite3` are some of the others I've noticed. I suspect the current state of typeshed is inconsistent on this issue, since not all modules that have a C implementation currently have  a `_foo.pyi` stub file. 

Is there a consensus among typeshed's maintainers here? I'm a little biased towards option 2 at the moment, since I'm trying to make as many inheritances match as I can, but I think there's a good argument for either side, and this is, in the end, a very minor issue with not much practical effect.

srittau:
Generally we try to follow the actual implementation, until there is a good reason not to. Typing private modules is "better", but it was often neglected, because practically speaking there isn't too much difference, and is slightly more work.

hauntsaninja:
This is a great question. I don't have a strong opinion, but I lean towards:

> Typeshed should follow the declared implementation over the layout. e.g. ReferenceType considers itself to live in the weakref module, that's where the majority of users will get it from, so that's the correct place for it.

I have this preference for two reasons:
- It results in better user facing diagnostics
- It better reflects the truth in alternative Python implementations or cases where C accelerator modules are missing

[...]

PR #11162, opened 2023-12-14
resort _ast and ast modules

tungol:
Well, I got a lukewarm to positive reaction from https://github.com/python/typeshed/issues/11141, about what to do when c-implemented classes claim their __module__ is the non-private one. So here's an MR to re-organize ast and _ast and get the type stubs to match what the implementation claims.

Mostly this is just lift and shift, but I took the opportunity to make the order within the file more closely match the implementation - that seems to be preferred?

related to https://github.com/python/typeshed/issues/3968 and the discussion in https://github.com/python/typeshed/issues/11141


JelleZijlstra:
This is heavily merge conflicted now and it's probably better to restart from scratch. I merged a few similar PRs recently and would be happy to review and merge this one if the conflicts are resolved.

3.8 is also now almost at EOL, so it's less important if we get the module wrong in 3.8.

tungol:
This MR should be ready now as well.

JelleZijlstra (approved):

PR #11164, opened 2023-12-14
add intermediate classes in selectors module

tungol:


stdlib/selectors.pyi
```diff
@@ -31,49 +31,41 @@ class BaseSelector(metaclass=ABCMeta):
...
         def select(self, timeout: float | None = None) -> list[tuple[SelectorKey, _EventMask]]: ...
-        def get_map(self) -> Mapping[FileDescriptorLike, SelectorKey]: ...
 
 # Not a real class at runtime, it is just a conditional alias to other real selectors.
 # The runtime logic is more fine-grained than a `sys.platform` check;
 # not really expressible in the stubs
-class DefaultSelector(BaseSelector):
-    def register(self, fileobj: FileDescriptorLike, events: _EventMask, data: Any = None) -> SelectorKey: ...
-    def unregister(self, fileobj: FileDescriptorLike) -> SelectorKey: ...
+class DefaultSelector(_BaseSelectorImpl):
```
tungol, line 68: Not the point of this MR, but I do wonder if this is really the best approach to the DefaultSelector? I would have thought that if we can't replicate the logic, it would make sense for it to be a union type like `DefaultSelector: SelectSelector | KqueueSelector | etc = ...`

I don't really know enough about this module to say what would be best, but it seems odd to create a new class where it never is a unique class at runtime

Akuli: If I remember correctly, that caused trouble with inheriting from `DefaultSelector`. You could try it (in a new PR) and see what mypy_primer says.

tungol: That makes sense

stdlib/selectors.pyi
```diff
@@ -31,49 +31,41 @@ class BaseSelector(metaclass=ABCMeta):
     def __enter__(self) -> Self: ...
     def __exit__(self, *args: Unused) -> None: ...
 
-class SelectSelector(BaseSelector):
+class _BaseSelectorImpl(BaseSelector, metaclass=ABCMeta):
+    def __init__(self) -> None: ...
```
srittau, line 35: Nit: do we need this? In typeshed we usually skip methods that match the super-class's.

stdlib/selectors.pyi
```diff
@@ -31,49 +31,41 @@ class BaseSelector(metaclass=ABCMeta):
...
     def register(self, fileobj: FileDescriptorLike, events: _EventMask, data: Any = None) -> SelectorKey: ...
     def unregister(self, fileobj: FileDescriptorLike) -> SelectorKey: ...
-    def select(self, timeout: float | None = None) -> list[tuple[SelectorKey, _EventMask]]: ...
+    def modify(self, fileobj: FileDescriptorLike, events: _EventMask, data: Any = None) -> SelectorKey: ...
+    def close(self) -> None: ...
     def get_map(self) -> Mapping[FileDescriptorLike, SelectorKey]: ...
 
+class SelectSelector(_BaseSelectorImpl):
+    def register(self, fileobj: FileDescriptorLike, events: _EventMask, data: Any = None) -> SelectorKey: ...
+    def unregister(self, fileobj: FileDescriptorLike) -> SelectorKey: ...
```
srittau, line 44: Same here.

srittau (commented):
LGTM, just two questions below.

stdlib/selectors.pyi
tungol: good catch; I can drop close() from this class too

srittau (approved):

PR #11166, opened 2023-12-14
use _compression.BaseStream for lzma.LZMAFile

tungol:

### DIFF
--- a/stdlib/lzma.pyi
+++ b/stdlib/lzma.pyi
@@ -1,4 +1,4 @@
-import io
+from _compression import BaseStream
 from _typeshed import ReadableBuffer, StrOrBytesPath
 from collections.abc import Mapping, Sequence
 from typing import IO, Any, TextIO, overload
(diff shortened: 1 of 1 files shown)


AlexWaygood (approved):

stdlib/lzma.pyi
```diff
@@ -104,7 +104,7 @@ class LZMACompressor:
 
 class LZMAError(Exception): ...
 
-class LZMAFile(io.BufferedIOBase, IO[bytes]):  # type: ignore[misc]  # incompatible definitions of writelines in the base classes
+class LZMAFile(BaseStream, IO[bytes]):  # type: ignore[misc]  # incompatible definitions of writelines in the base classes
```
tungol, line 107: Unrelated to this MR:

I stared at this for a while, and I can't quite figure out what the purpose of inheriting from IO[bytes] here is. I found the discussion of typing.IO here <https://github.com/python/typing/discussions/829> which seems to say typeshed's policy is to avoid the use of typing.IO.

I guess the concern is that because typing.IO isn't actually a Protocol, functions typed to accept a typing.IO argument require classes in the standard library to be subclasses of typing.IO to work as expected?

AlexWaygood: Yes, yes, and yes :) Most of our uses of `IO` are legacy uses from years back, before we came up with the policy of mostly using narrower protocols for I/O related things. Nowadays, we recommend that our users don't use `typing.IO`, and we wouldn't usually start using it in a new class we were adding stubs for.

We could possibly think about changing this -- but, as you say, it comes at the risk of breaking user code that's already using `typing.IO`. So, best to consider it separately in another PR, if at all.

PR #11170, opened 2023-12-16
seaborn: Complete the categorical module

hamdanal:


srittau (approved):
Thanks, did not compare in detail, but looks reasonable.

PR #11176, opened 2023-12-17
Type `obj` parameter of `PIL.Image.fromarray`

Avasam:
This supports numpy's NDArray, OpenCV's Mat & MatLike, Pillow's Image, etc.

Fixes a `pyright: reportUnknownMemberType` report in: https://github.com/Toufool/AutoSplit/blob/master/src/compare.py#L101-L102


stubs/Pillow/PIL/Image.pyi
```diff
@@ -36,6 +36,11 @@ _Color: TypeAlias = (
 class _Writeable(SupportsWrite[bytes], Protocol):
     def seek(self, __offset: int) -> Any: ...
 
+# Ref: https://numpy.org/doc/stable/reference/arrays.interface.html#python-side
+class _SupportsArrayInterface(Protocol):
+    @property
+    def __array_interface__(self) -> dict[str, Any]: ...
```
JelleZijlstra, line 42: Should we define the TypedDict instead?

A complication there is that `dict[str, Any]` is not compatible with a TypedDict.

Avasam: I don't think that's possible. otherwise ndarrays themselves wouldn't be compatible.
https://github.com/numpy/numpy/blob/0ebbdb9f24f96fc97b8df1ecb632e23a86b65625/numpy/__init__.pyi#L990C14-L990C14

So with the following definition
```py
from typing import NotRequired, TypedDict

class _ArrayInterfaceTypedDict(TypedDict):
    shape: tuple[int,...]
    typestr: str
    descr: NotRequired[list[tuple[str, str]]]
    data: NotRequired[tuple[int, Incomplete]]
    strides: NotRequired[Incomplete | tuple[int,...] | None]
    mask: NotRequired[_SupportsArrayInterface | None]
    offset: NotRequired[int]
    version: int
    

# Ref: https://numpy.org/doc/stable/reference/arrays.interface.html#python-side
class _SupportsArrayInterface(Protocol):
    @property
    def __array_interface__(self) -> _ArrayInterfaceTypedDict: ...
    
def fromarray(obj: _SupportsArrayInterface, mode: _Mode | None = None) -> Image: ...
```
The following code:
```py
from cv2.typing import MatLike
import imagehash
from PIL import Image

def compare_phash(source: MatLike, capture: MatLike, mask: MatLike | None = None):
    ...
    source_hash = imagehash.phash(Image.fromarray(source))
    capture_hash = imagehash.phash(Image.fromarray(capture))
```

results in the following pyright error:
```py
error: Argument of type "MatLike" cannot be assigned to parameter "obj" of type "_SupportsArrayInterface" 
in function "fromarray"
    Type "MatLike" cannot be assigned to type "_SupportsArrayInterface"
      "NumPyArrayGeneric" is incompatible with protocol "_SupportsArrayInterface"
        "__array_interface__" is an incompatible type
          Type "() -> dict[str, Any]" cannot be assigned to type "() -> _ArrayInterfaceTypedDict"
            Function return type "dict[str, Any]" is incompatible with type "_ArrayInterfaceTypedDict" (reportGeneralTypeIssues)
```
image

PR #11179, opened 2023-12-18
Add types for defusedxml

solkaz:
Initial attempt at adding types for [defusedxml](https://github.com/tiran/defusedxml).

Some notes:
- This library relies on modules that aren't typed; namely:
  - `xml.sax.expatreader` - undocumented
  - `lxml` - no types, and I would prefer to ignore it for now


stubs/defusedxml/defusedxml/cElementTree.pyi
```diff
@@ -0,0 +1,13 @@
+from .ElementTree import (
+    DefusedXMLParser,
+    ParseError as ParseError,
+    fromstring as fromstring,
+    iterparse as iterparse,
+    parse as parse,
+    tostring as tostring,
+)
```
solkaz, line 8: The original source code has the following imports:
```
XMLTreeBuilder as XMLTreeBuilder
XMLParse as XMLParse
XMLParser as XMLParser
XML as XML
```

They were removed as Flake8 complained that below, they were being redefined, or that `XML` could not be redefined because it's a constant. If this isn't right, let me know how I could improve this

srittau: FYI: The current version of this file looks good, although you should probably add the `__all__` statements.

solkaz: @srittau added

srittau (approved):
I'm sorry for the late review. This looks good to me, although I didn't compare it to the implementation.

solkaz:
@srittau thank you for the review! I'll follow up with fixes once I implement this into my personal project

jbethune:
@solkaz Thanks for putting this together!

I noticed that `defusedxml.minidom.parse` and `defusedxml.minidom.parseString` don't have return type annotations. The return type should be `xml.dom.minidom.Document`. See also: [expatbuilder.parse()](https://github.com/python/typeshed/blob/26e77cbf67ee0e224808a85b18c08c1dda63aaec/stubs/defusedxml/defusedxml/expatbuilder.pyi#L25-L31)

I know this PR is already merged. I guess we need to create a new one?

srittau:
> I know this PR is already merged. I guess we need to create a new one?

Indeed. Any PRs to further improve the stubs are welcome!

tyrossel:
Thanks for adding  those types !

I got an error with the find function:

```python
import xml.etree.ElementTree as ET
import defusedxml.ElementTree as DET

el = DET.fromstring(filecontent)
subEl = el.find("Document", ns)
```
Mypy gives the error: `mypy: "Element" has no attribute "find"` (i.e. `xml.dom.minidom.Element`)

I can cast `el` to `ET.Element` to solve this issue, but I had the understanding that defusedxml was a drop-in replacement to those functions.

AlexWaygood:
@tyrossel, please open an issue on our tracker if there's a problem with these stubs; it's hard to keep track of comments on merged PRs :)

PR #11180, opened 2023-12-18
Fix annotation for networkx.subgraph_view

NeilGirdhar:
(Also, adjust to 80 column character width.)


NeilGirdhar:
@rhelmot Thanks for the annotations!  (FYI)

PR #11181, opened 2023-12-18
Add some missing networkx annotations

NeilGirdhar:
Repalced some uses of `Imcomplete` with `Any` where it seems like no elaboration will be possible unless we're planning on making the data type generic?  If so, I suggest creating a type alias for the data type and then it can be made into a generic parameter one day.


srittau (approved):
Thanks!

PR #11187, opened 2023-12-19
add _io module

tungol:
In addition to shuffling defintions between files, these things got tweaked:

 - I moved BufferedIOBase.raw to the child classes which *do* define that attribute on themselves
 - Removed several redundant definitions of \_\_enter\_\_. These seem to be an artifact of using TypeVar instead of Self originally
 - seek and truncate have inconsistent variable names. They're also positional only, so they were in the stubtest allowlist before, but I went ahead and made them explicit in the stubs
 - BytesIO.readlines shouldn't have been on the allowlist in the first place. It differs not only in variable name, but also default value.
 - A big block of functions in TextIOWrapper were commented that mypy needed them, but I don't think it does anymore, unless there's a problem with subclassing TextIOWrapper that doesn't show up in typeshed itself. No indication in the history about that.
 - In the implementation, the concrete classes inherit from the private implementation \_\*IOBase classes, not the classes in io which are actually metaclasses, but they are registered to those metaclasses at runtime. It wasn't technically required for any reason, but I split the difference on this by keeping the \_\*IOBase classes in their base classes directly. I think it's a bit of a reminder of the actual implementation, and means that a stubtest check for inheritance will show that typeshed is adding to the base classes, rather than replacing the base class, and I think that's a little cleaner.

Partially related to https://github.com/python/typeshed/issues/3968


JelleZijlstra:
Unfortunately this has severe merge conflicts now. If you fix them, I will review the PR.

JelleZijlstra:
Thanks for contributing! I'm closing this PR for now, because it still has a merge conflict
after three months of inactivity. If you are still interested, please feel free to open

PR #11188, opened 2023-12-20
move re.error into re.pyi

tungol:
implementation is in sre\_constants / re.\_constants, but it sets it's \_\_module\_\_ to re


srittau (approved):

stdlib/re.pyi
```diff
@@ -47,6 +46,14 @@ if sys.version_info >= (3, 11):
 
 _T = TypeVar("_T")
 
+class error(Exception):
```
srittau, line 49: It's probably best to leave a comment here:

```suggestion
# Defined in module sre_constants in the implementation, but __module__ is overwritten to say re.error.
class error(Exception):
```

Or maybe worded better.

srittau (changes requested):

tungol:
I'm making this a draft pending discussion in https://github.com/python/typeshed/issues/11141

I'm inclined to think using an explicit __module__ attribute (like runtime does) is a better way to handle this discrepancy.

JelleZijlstra (approved):

srittau (approved):

PR #11189, opened 2023-12-20
follow implementation more closely in zoneinfo

tungol:


stdlib/zoneinfo/__init__.pyi
```diff
@@ -1,38 +1,31 @@
-from _typeshed import StrPath
-from collections.abc import Iterable, Sequence
+import sys
+from collections.abc import Iterable
 from datetime import datetime, timedelta, tzinfo
-from typing import Any, Protocol
 from typing_extensions import Self
 
-__all__ = ["ZoneInfo", "reset_tzpath", "available_timezones", "TZPATH", "ZoneInfoNotFoundError", "InvalidTZPathWarning"]
+if sys.version_info >= (3, 9):
```
srittau, line 6: This should not be required, since `stdlib/VERSIONS` says this module is only available on Python 3.9+.

tungol: That's what I thought, but pyright on 3.7 and 3.8 fails if it isn't there:

```
  /home/runner/work/typeshed/typeshed/stdlib/zoneinfo/__init__.pyi:4:6 - error: Stub file not found for "zoneinfo._common" (reportMissingTypeStubs)
  /home/runner/work/typeshed/typeshed/stdlib/zoneinfo/__init__.pyi:5:6 - error: Stub file not found for "zoneinfo._tzpath" (reportMissingTypeStubs)
  /home/runner/work/typeshed/typeshed/stdlib/zoneinfo/__init__.pyi:4:78 - error: "_IOBytes" is unknown import symbol (reportGeneralTypeIssues)
```

I tried adding zoneinfo._common to the VERSIONS file but that didn't change anything. Do you know what other solution there might be?

AlexWaygood: My understanding is that no type checker has special-cased the `VERSIONS` file specifically for our CI.

If a user tries to import something from `zoneinfo` on Python 3.8, the type checker will emit an error, as it will understand (due to the `VERSIONS` file) that the module only exists on Python 3.9+. But that's a different situation to our CI, where we're asking pyright to run with `--python-version 3.8` on files in `zoneinfo` itself.

I.e., pyright is correctly understanding (due to `VERSIONS`) that submodules in `zoneinfo` are unavailable on Python 3.8, and is therefore (amusingly) emitting errors when it analyses other `zoneinfo` submodules with `--python-version 3.8`, and ses them importing from `zoneinfo` submodules.

I think we don't get the same errors with mypy in CI, because we run mypy via a custom script in CI rather than just running it on all of typeshed directly (which is what we do with pyright), and `mypy_test.py` knows about `VERSIONS`. But I'd rather avoid having to run pyright via a custom script in CI, if we can avoid it.

srittau: So if I understand correctly, the problem is that before, we basically run the equivalent of `pyright stdlib/zoneinfo/__init__.pyi`, which works with all Python versions. But now that `__init__.pyi` imports from other modules, pyright looks at VERSIONS and notices that this module shouldn't exist and errors?

In that case, we unfortunately need some kind of solution. I don't think we should change the stubs (and destroy diffs and blames in the process) to work around shortcomings in our testing infrastructure. The stubs live outside this repository after all, and will be inspected by thousands of people.

tungol: That makes sense. What do you think I should do about it here? If we don't want the sys.version_info check, I can put in some `# pyright: ignore` statements. They'll apply to every version, but it's just the imports and probably should be too big a deal.

srittau: That might be good workaround for now. I'll look into modifying our pyright test script in the next few days, if no one beats me to it. It seems to me the easiest way is to create a temp config file from the normal one and add the non-supported files to the exclude list.

This might also help us in the future to get rid of the stricter config and possibly replace it with a METADATA key.

tungol: Looks like it's not that simple; pyright is throwing an error for unnecessary ignore statements on the supported versions with that in place.

I think the best path forward is to add back the version check, but also add a note explaining the situation.

tungol: Actually, I haven't looked at the pyright config file yet; Let me see if that will work without too much fuss

AlexWaygood: Note that we already have "redundant" `sys.version_info` checks in several other stdlib stub files, to work around this exact issue (`importlib.readers`, all the `importlib.resources` submodules, etc.)

tungol: The failing CI test uses `pyrightconfig.stricter.json`, which already has a fair amount of stdlib files in it's `exclude` block. 

I can make it pass by adding zoneinfo/__init__.pyi there, or by keeping the version_info check. The latter effects the actual type file, but it does mean that pyright can still run on the supported versions. So I think that's probably better overall.

PR #11190, opened 2023-12-20
Bump pyright to 1.1.342

sobolevn:
We've missed 3 new versions: https://github.com/microsoft/pyright/releases/tag/1.1.342


srittau (approved):
CI failure is unrelated.

PR #11192, opened 2023-12-20
Add stubs for antlr4

Beakerboy:


Beakerboy:
It looks like I can easily clean up the Test / Test typeshed with pyright failures by referencing and tracing through the source. What's the best way to resolve issues where the source devs initialize variables to None?

AlexWaygood:
> for the Third-party stubtests, what's the best way to resolve issues where the source devs initialize variables to None?

I haven't looked at the `antlr4` source code at all, but if the runtime has a function like this:

```py
def foo(name=None): ...
```

And you've written a stub like this:

```py
def foo(name: str = ...) -> None: ...
```

Then you generally need to change it to something like this:

```py
def foo(name: str | None = None) -> None ...
```

Since the function obviously won't emit an error if `None` is passed (since `None` is the default)

srittau:
Most of the pyright errors seem to be referencing missing generic arguments. The best option is to add the proper generic argument. But if you don't have the bandwidth for that, it's fine to use `Incomplete` (from the `_typeshed` module) for now. This is a marker that proper types need to be added at some point: `set` -> `set[Incomplete]`, `dict` -> `dict[Incomplete, Incomplete]` etc.

I haven't looked at the stubtest problems yet, but this indicated a mismatch between the implementation and the stubs.

Beakerboy:
@srittau I have half the missing generic arguments fixed in a dev branch. I'm looking at the source to make sure `TokenName: list` is actually `TokenName: list[str]`. Thanks for the "Incomplete" hint. If I find one excessively difficult, I'll make sure to use that.

Beakerboy:
`perl -pi -e 's/(def.*)\): \.\.\./$1\) -> Incomplete: \.\.\./g' *.pyi` to the rescue

stubs/antlr4-python3-runtime/antlr4/BufferedTokenStream.pyi
```diff
@@ -25,15 +25,15 @@ class BufferedTokenStream(TokenStream):
...
-    def getSourceName(self): ...
-    def getText(self, start: int = ..., stop: int = ...): ...
+    def setTokenSource(self, tokenSource: Lexer) -> Incomplete: ...
+    def nextTokenOnChannel(self, i: int, channel: int) -> Incomplete: ...
+    def previousTokenOnChannel(self, i: int, channel: int) -> Incomplete: ...
+    def getHiddenTokensToRight(self, tokenIndex: int, channel: int = ...) -> Incomplete: ...
+    def getHiddenTokensToLeft(self, tokenIndex: int, channel: int = ...) -> Incomplete: ...
+    def filterForChannel(self, left: int, right: int, channel: int) -> Incomplete: ...
+    def getSourceName(self) -> Incomplete: ...
+    def getText(self, start: int = ..., stop: int = ...) -> Incomplete: ...
```
AlexWaygood, line 38: Apologies for not saying this earlier -- rather than adding `-> Incomplete` in return types everywhere, we generally prefer to add entries to this list here, so that the directory is excluded from pyright's stricter settings. So for a type that's completely unknown, we'd just omit the annotations (if the stub is excluded from pyright's stricter settings):

```py
def foo(x: int, y): ...
```

But if the type is _partially_ known, we'd generally use `Incomplete`:

```py
def foo(x: int, y: list[Incomplete]) -> set[Incomplete]: ...
```

Beakerboy:
Last errors:

```
error: antlr4.tree.Tree.ParserRuleContext is not present at runtime
Stub: in file /home/runner/work/typeshed/typeshed/stubs/antlr4-python3-runtime/antlr4/tree/Tree.pyi:23
Any
Runtime:
MISSING
```

AlexWaygood:
> Thanks! This is the way stubgen created the file.

Yeah, stubgen has bad default settings. In our convenience script for generating new stubs for typeshed, we always run stubgen with the `--export-less` option: https://github.com/python/typeshed/blob/main/scripts/create_baseline_stubs.py

> Should I remove the “as Foo” from all the imports?

Probably

stubs/antlr4-python3-runtime/METADATA.toml
```diff
@@ -0,0 +1,5 @@
+version = "4.13.*"
+upstream_repository = "https://github.com/antlr/antlr4"
+
+[tool.stubtest]
```
AlexWaygood, line 4: While we obviously prefer stubs to be as complete as possible, it's fine to do this for now if you want stubtest to shut up about things being missing from the stub even though they're present at runtime

```suggestion
[tool.stubtest]
ignore_missing_stub = true
```

Beakerboy:
This pull request has a green check. Let me know if there are any other must-dos before merging it.

srittau:
Could you do a final cleanup pass with the following replacements?

* ` -> Incomplete:` -> `:`
* `: Incomplete)` -> `)`
* `: Incomplete,` -> `,`

AlexWaygood:
I just pushed a bunch of cleanups to your PR branch. I'm not sure exactly what I did that could have triggered new stubtest errors, but stubtest is now complaining that some things in these stubs don't exist at runtime, for some reason. At least some of the errors appear to be correct, however:

```pycon
>>> import antlr4.Recognizer
>>> antlr4.Recognizer.RecognizeException
AttributeError: module 'antlr4.Recognizer' has no attribute 'RecognizeException'
```

(At runtime, `RecognizeException` is set [here](https://github.com/antlr/antlr4/blob/28eb03612f6fff900d240e51b90c251e4357d4e3/runtime/Python3/src/antlr4/Recognizer.py#L11), but then deleted again [here](https://github.com/antlr/antlr4/blob/28eb03612f6fff900d240e51b90c251e4357d4e3/runtime/Python3/src/antlr4/Recognizer.py#L147))

AlexWaygood:
Ah, I see stubtest didn't _run_ on 9f184a1, for some reason, which explains why we didn't see those complaints on the commit immediately prior to the cleanups I just pushed.

stubs/antlr4-python3-runtime/antlr4/CommonTokenFactory.pyi
```diff
@@ -0,0 +1,14 @@
...
+from antlr4.Token import CommonToken as CommonToken
+
+class TokenFactory: ...
+
+class CommonTokenFactory(TokenFactory):
+    DEFAULT: Incomplete
+    copyText: Incomplete
+    def __init__(self, copyText: bool = False) -> None: ...
+    def create(
+        self, source: tuple[Incomplete], type: int, text: str, channel: int, start: int, stop: int, line: int, column: int
```
AlexWaygood, line 12: `tuple[Incomplete]` means "A tuple of length 1, where the sole element in the tuple is of type `Incomplete`". Is that what you meant, or did you want to say "A tuple of arbitrary length, in which all elements are of type `Incomplete`"? If the latter, you should do this instead:

```suggestion
        self, source: tuple[Incomplete, ...], type: int, text: str, channel: int, start: int, stop: int, line: int, column: int
```

Beakerboy:
@AlexWaygood Do you have any advice on resolving the “not present at runtime” issues given the state of the antlr4 source material?

AlexWaygood:
> @AlexWaygood Do you have any advice on resolving the “not present at runtime” issues given the state of the antlr4 source material?

Since these things do indeed seem to be not present at runtime when you actually use the package, I would delete them from the stubs

Beakerboy:
All green and good to go from my end.

AlexWaygood:
> @AlexWaygood what’s the usual wait time from “all good” to “merge”? Anything I need to do?

Sorry for the wait. I'm afraid I'm more busy than usual at the moment, so probably won't be able to take a proper look at this soon. Hopefully somebody else on the team will be able to take a look soon, but I'm afraid we're all volunteers, so I can't make any promises about timescale here :(

Beakerboy:
@AlexWaygood No problem. I noticed there were other mergers and wanted to make sure there wasn’t anything I needed to do.

stubs/antlr4-python3-runtime/antlr4/PredictionContext.pyi
```diff
@@ -0,0 +1,87 @@
...
+    def getReturnState(self, index: int): ...
+    def __eq__(self, other): ...
+    def __hash__(self): ...
+
+def PredictionContextFromRuleContext(atn: ATN, outerContext: RuleContext | None = None): ...
+def merge(
+    a: PredictionContext,
+    b: PredictionContext,
+    rootIsWildcard: bool,
+    mergeCache: dict[tuple[Incomplete], SingletonPredictionContext],
```
srittau, line 64: Looking at the source code, the `None` case is handled explicitly. Also, the tuple seems to have two elements:

```suggestion
    mergeCache: dict[tuple[Incomplete, Incomplete], SingletonPredictionContext] | None,
```

stubs/antlr4-python3-runtime/antlr4/PredictionContext.pyi
```diff
@@ -0,0 +1,87 @@
...
+): ...
+def mergeSingletons(
+    a: SingletonPredictionContext,
+    b: SingletonPredictionContext,
+    rootIsWildcard: bool,
+    mergeCache: dict[tuple[Incomplete], SingletonPredictionContext],
```
srittau, line 70:
```suggestion
    mergeCache: dict[tuple[Incomplete, Incomplete], SingletonPredictionContext] | None,
```

stubs/antlr4-python3-runtime/antlr4/PredictionContext.pyi
```diff
@@ -0,0 +1,87 @@
...
+): ...
+def mergeRoot(a: SingletonPredictionContext, b: SingletonPredictionContext, rootIsWildcard: bool): ...
+def mergeArrays(
+    a: ArrayPredictionContext,
+    b: ArrayPredictionContext,
+    rootIsWildcard: bool,
+    mergeCache: dict[tuple[Incomplete], SingletonPredictionContext],
```
srittau, line 77:
```suggestion
    mergeCache: dict[tuple[Incomplete, Incomplete], SingletonPredictionContext] | None,
```

stubs/antlr4-python3-runtime/antlr4/PredictionContext.pyi
```diff
@@ -0,0 +1,87 @@
...
+): ...
+def combineCommonParents(parents: list[PredictionContext]): ...
```
srittau, line 79: This should work with any sequence:

```suggestion
def combineCommonParents(parents: SupportsLenAndGetItem[PredictionContext]): ...
```

stubs/antlr4-python3-runtime/antlr4/PredictionContext.pyi
```diff
@@ -0,0 +1,87 @@
+from _typeshed import Incomplete
```
srittau, line 1: See above:

```suggestion
from collections.abc import MutableMapping
from _typeshed import Incomplete, SupportsLenAndGetItem
```

stubs/antlr4-python3-runtime/antlr4/PredictionContext.pyi
```diff
@@ -0,0 +1,87 @@
...
+def mergeRoot(a: SingletonPredictionContext, b: SingletonPredictionContext, rootIsWildcard: bool): ...
+def mergeArrays(
+    a: ArrayPredictionContext,
+    b: ArrayPredictionContext,
+    rootIsWildcard: bool,
+    mergeCache: dict[tuple[Incomplete], SingletonPredictionContext],
+): ...
+def combineCommonParents(parents: list[PredictionContext]): ...
+def getCachedPredictionContext(
+    context: PredictionContext, contextCache: PredictionContextCache, visited: dict[PredictionContext, PredictionContext]
```
srittau, line 81:
```suggestion
    context: PredictionContext, contextCache: PredictionContextCache, visited: MutableMapping[PredictionContext, PredictionContext]
```

stubs/antlr4-python3-runtime/antlr4/PredictionContext.pyi
```diff
@@ -0,0 +1,87 @@
...
+): ...
+def getAllContextNodes(
+    context: PredictionContext,
+    nodes: list[Incomplete] | None = None,
+    visited: dict[PredictionContext, PredictionContext] | None = None,
```
srittau, line 86: The implementation uses `visited.put()`, which isn't a `dict` method, so I'm not sure what's going on there. I guess it's fine to keep the annotation as it is for now.

stubs/antlr4-python3-runtime/antlr4/atn/ParserATNSimulator.pyi
```diff
@@ -0,0 +1,133 @@
...
+from antlr4.Token import Token as Token
+from antlr4.Utils import str_list as str_list
+
+class ParserATNSimulator(ATNSimulator):
+    debug: bool
+    trace_atn_sim: bool
+    dfa_debug: bool
+    retry_debug: bool
+    parser: Incomplete
+    decisionToDFA: Incomplete
```
srittau, line 42:
```suggestion
    decisionToDFA: list[DFA]
```

stubs/antlr4-python3-runtime/antlr4/atn/ParserATNSimulator.pyi
```diff
@@ -0,0 +1,133 @@
...
+    def removeAllConfigsNotInRuleStopState(self, configs: ATNConfigSet, lookToEndOfRule: bool): ...
+    def computeStartState(self, p: ATNState, ctx: RuleContext, fullCtx: bool): ...
+    def applyPrecedenceFilter(self, configs: ATNConfigSet): ...
+    def getReachableTarget(self, trans: Transition, ttype: int): ...
+    def getPredsForAmbigAlts(self, ambigAlts: set[int], configs: ATNConfigSet, nalts: int): ...
+    def getPredicatePredictions(self, ambigAlts: set[int], altToPred: list[int]): ...
+    def getSynValidOrSemInvalidAltThatFinishedDecisionEntryRule(self, configs: ATNConfigSet, outerContext: ParserRuleContext): ...
+    def getAltThatFinishedDecisionEntryRule(self, configs: ATNConfigSet): ...
+    def splitAccordingToSemanticValidity(self, configs: ATNConfigSet, outerContext: ParserRuleContext): ...
+    def evalSemanticContext(self, predPredictions: list[Incomplete], outerContext: ParserRuleContext, complete: bool): ...
```
srittau, line 67:
```suggestion
    def evalSemanticContext(self, predPredictions: Iterable[Incomplete], outerContext: ParserRuleContext, complete: bool): ...
```

stubs/antlr4-python3-runtime/antlr4/atn/ParserATNSimulator.pyi
```diff
@@ -0,0 +1,133 @@
+from _typeshed import Incomplete
```
srittau, line 1:
```suggestion
from collections.abc import Iterable
from _typeshed import Incomplete
```

stubs/antlr4-python3-runtime/antlr4/atn/PredictionMode.pyi
```diff
@@ -0,0 +1,40 @@
...
+    @classmethod
+    def hasNonConflictingAltSet(cls, altsets: list[set[int]]): ...
+    @classmethod
+    def hasConflictingAltSet(cls, altsets: list[set[int]]): ...
+    @classmethod
+    def allSubsetsEqual(cls, altsets: list[set[int]]): ...
+    @classmethod
+    def getUniqueAlt(cls, altsets: list[set[int]]): ...
+    @classmethod
+    def getAlts(cls, altsets: list[set[int]]): ...
```
srittau, line 32:
```suggestion
    def resolvesToJustOneViableAlt(cls, altsets: Sequence[set[int]]): ...
    @classmethod
    def allSubsetsConflict(cls, altsets: Sequence[set[int]]): ...
    @classmethod
    def hasNonConflictingAltSet(cls, altsets: Sequence[set[int]]): ...
    @classmethod
    def hasConflictingAltSet(cls, altsets: Sequence[set[int]]): ...
    @classmethod
    def allSubsetsEqual(cls, altsets: Sequence[set[int]]): ...
    @classmethod
    def getUniqueAlt(cls, altsets: Sequence[set[int]]): ...
    @classmethod
    def getAlts(cls, altsets: Sequence[set[int]]): ...
```

stubs/antlr4-python3-runtime/antlr4/atn/PredictionMode.pyi
```diff
@@ -0,0 +1,40 @@
...
+    @classmethod
+    def getConflictingAltSubsets(cls, configs: ATNConfigSet): ...
+    @classmethod
+    def getStateToAltMap(cls, configs: ATNConfigSet): ...
+    @classmethod
+    def hasStateAssociatedWithOneAlt(cls, configs: ATNConfigSet): ...
+    @classmethod
+    def getSingleViableAlt(cls, altsets: list[set[int]]): ...
```
srittau, line 40:
```suggestion
    def getSingleViableAlt(cls, altsets: Sequence[set[int]]): ...
```

stubs/antlr4-python3-runtime/antlr4/atn/PredictionMode.pyi
```diff
@@ -0,0 +1,40 @@
+from enum import Enum
```
srittau, line 1:
```suggestion
from collections.abc import Sequence
from enum import Enum
```

stubs/antlr4-python3-runtime/antlr4/dfa/DFASerializer.pyi
```diff
@@ -0,0 +1,17 @@
+from _typeshed import Incomplete
+
+from antlr4 import DFA as DFA
+from antlr4.dfa.DFAState import DFAState as DFAState
+from antlr4.Utils import str_list as str_list
+
+class DFASerializer:
+    dfa: Incomplete
+    literalNames: Incomplete
+    symbolicNames: Incomplete
```
srittau, line 10:
```suggestion
    literalNames: list[str] | None
    symbolicNames: list[str] | None
```

srittau (changes requested):
Thanks! Looks mostly good to me, although I didn't compare it to the implementation in-depth. A few things I noticed below.

stubs/antlr4-python3-runtime/antlr4/PredictionContext.pyi
Beakerboy: Look like MutableMapping is unused.

stubs/antlr4-python3-runtime/antlr4/atn/ParserATNSimulator.pyi
Beakerboy: Iterable appear to be unused

srittau:
The "Run mypy on the test cases" error is unrelated.

srittau (approved):

Beakerboy:
Thanks so much! I’m curious how long it takes for the types_antlr4 package to show up at PYPI?

srittau:
The types are built once per day, I think during the night UTC time.

PR #11193, opened 2023-12-20
Stdlib modules removed in Python 3.13 (PEP 594)

lancelote:
Set last available version for stdlib modules removed in Python 3.13 (deprecated in PEP 594)

* The [corresponding section](https://docs.python.org/3.13/whatsnew/3.13.html#pep-594-dead-batteries) from [What's New in Python 3.13](https://docs.python.org/3.13/whatsnew/3.13.html)
* [PEP 594 - Removing dead batteries from the standard library](https://peps.python.org/pep-0594/)


AlexWaygood:
Thanks! We _generally_ hold off on making this kind of change until the beta period, however, as things can change a lot during the alpha period (removed modules have often been added back during the alpha period after it was found that their removal caused too much disruption).

Having said that, when we start running stubtest on py313 in CI, stubtest _will_ emit an error if it finds that any of these modules do in fact still exist, so it shouldn't cause any problems if we merge this now.

Not sure, what do other maintainers think?

Avasam:
Given the small, CI testable, and backwards compatible nature of the changes. I don't see much harm in having it ahead of time. If it helps early testing of Python 3.13.

It's a bit more straightforward than added platform checks, which would be more annoying to revert.

So if you're fine knowing some of these might have to be added back once we start testing on Python 3.13 .

---

But that raises the question: are we "officially" moving the line (from beta to alpha) for stdlib removals? Or should we stick to a rigid "no changes until beta"?

hauntsaninja:
typeshed isn't really a rigid rules kind of project. The reason we try to keep to beta is that CPython has a habit of changing things, potentially in ways that stubtest cannot catch. If we're confident a change is not churn-y and stubtest will catch any later changes, like this one, there's no reason not to merge.

JelleZijlstra:
Getting to this late, but we might want to move away from the rule to hold off until the beta, since there's now a lot more overlap between active typeshed and CPython maintainers than before.

hauntsaninja:
Interesting point, but I still think beta is the appropriate time to set about updating things module by module. If you do this too early, you just end up having to do it again later or being okay with missing the things stubtest doesn't catch.

PR #11197, opened 2023-12-22
IOBase.readline should not accept None

danking:
AFAICT this is an issue in mypy & pyright's typesheds as well.

Python 3.9, at least, [does not permit `None`](https://docs.python.org/3.9/library/io.html?highlight=iobase#io.IOBase.readline) for IOBase.readline

> readline(size=-1, /)
> Read and return one line from the stream. If size is specified, at most size bytes will be read.
> 
> The line terminator is always b'\n' for binary files; for text files, the newline argument to open() can be used to select the line terminator(s) recognized.

When I try to subclass IOBase like this:

```python3
from io import IOBase

class Foo(IOBase):
    def readline(self, size: int = -1):
        raise OSError
```
I get this error:

```python3
  /private/tmp/foo.py:4:9 - error: Method "readline" overrides class "IOBase" in an incompatible manner
    Parameter 2 type mismatch: base parameter is type "int | None", override parameter is type "int"
      Type "int | None" cannot be assigned to type "int"
        "None" is incompatible with "int" (reportIncompatibleMethodOverride)
```

Moreover io.pyi appears to be internally inconsistent because [`TextIOBase.readline`](https://github.com/python/typeshed/blob/16933b838eef7be92ee02f66b87aa1a7532cee63/stdlib/io.pyi#L148) does not permit None.


srittau:
The implementation of `IOBase` *does* allow `None`:

```python
from io import IOBase

class Foo(IOBase):
    def __init__(self):
        self.content = b"line 1\nline 2\n"
        self.pos = 0 
    
    def read(self, i):
        response = self.content[self.pos:self.pos+i]
        self.pos += i
        return response

print(Foo().readline(None))  # -> b'line 1\n'
print(Foo().readline(-1))  # -> b'line 1\n'
```

Looking at the implementation this seems deliberate: 

https://github.com/python/cpython/blob/5f665e99e0b8a52415f83c2416eaf28abaacc3ae/Modules/_io/iobase.c#L548

As we tend to following implementation rather than documentation, I think the current annotations in typeshed are correct.

danking:
Huh. So, None is a synonym for -1? And I see now that TextIOBase has an explicit ignore. Alright then.

PR #11199, opened 2023-12-25
have type stubs for cli_support

tngraf:
Have type stubs on typeshed for my library cli_support at https://github.com/sw360/clipython.


AlexWaygood:
Hi, thanks for these stubs! I'm curious, though – if these are for a library you maintain, have you considered bundling these stubs with your runtime library and marking your library as `py.typed`? That way users would get the stubs "for free" when they pip-installed the runtime package, rather than having to install a separate package for their type hints :)

tngraf:
@AlexWaygood Thank you very much for the explanation ... but it seems not to work.  
Yes, I have added a `py.typed` file, but when I run `mypy` on an application that uses the library,   
`mypy` still complains that is does not find type information. This also happens for other - far more often
used - libraries: they contain a `py.typed` file, but `mypy` is not satisfied.  
What am I missing?

Besides all this I like to thank all typeshed maintainers - you do great work and just by reading all your documentation and creating the PR I learned something new.

Beakerboy:
@tngraf 
I just did this for a project of mine. I added the py.typed file in the same directory as the __init__.py file and added this to my pyproject.toml:
```
[tool.setuptools.package-data]
"pkgname" = ["py.typed"]
```

I made sure the pyi files were in the same directory as the corresponding py files and it worked
The project is [antlr-vba](https://pypi.org/project/antlr4-vba/)

hauntsaninja:
It looks like you use poetry, which I believe automatically includes py.typed files.

`cli_support` 2.0.0 works totally fine for me with mypy. After running `python -m pip install cli_support`, running `python -m mypy -c 'import cli_support; reveal_type(cli_support.CLI.CliFile.read_from_file)'` works as expected

tngraf:
Thanks for all the support. I got it now.  
My bad ... too many different virtual environments and too many library versions.  
No cli_support at typeshed needed 😄

PR #11201, opened 2023-12-26
Generalise attribute `__func__` of class `MethodType` from `FunctionType` to `Callable`

### LINKED ISSUE #11200: `MethodType` uses `FunctionLike` types inconsistently
Since version 1.8, Mypy [does not intersect types in isinstance checks if at least one is final](https://github.com/python/mypy/pull/16330).  In my code base, this leads to unexpected unreachable warnings.  I could narrow them down to:

```python
from types import MethodType

m: MethodType
if isinstance(m.__func__, MethodType):  # error: Subclass of "FunctionType" and "MethodType" cannot exist: "FunctionType" is final  [unreachable] \
                                        # error: Subclass of "FunctionType" and "MethodType" cannot exist: "MethodType" is final  [unreachable]
    m = m.__func__  # error: Statement is unreachable  [unreachable]
```

As I implemented the Mypy change, I double-checked that I did not break something.  But it seems the definition of `MethodType` in typeshed is inconsistent.  Its `__init__` method accepts `Callable` (#1499), but its `__func__` attribute returns `FunctionType` (#1383).  As `MethodType` can accept and return other methods at runtime, it seems right to define the return type as `Callable`, too.

tyralla:
Fixes #11200

### DIFF
--- a/stdlib/types.pyi
+++ b/stdlib/types.pyi
@@ -418,18 +418,6 @@ class CoroutineType(Coroutine[_YieldT_co, _SendT_contra, _ReturnT_co]):
     @overload
     def throw(self, __typ: BaseException, __val: None = None, __tb: TracebackType | None = ...) -> _YieldT_co: ...
 
-class _StaticFunctionType:
-    # Fictional type to correct the type of MethodType.__func__.
-    # FunctionType is a descriptor, so mypy follows the descriptor protocol and
-    # converts MethodType.__func__ back to MethodType (the return type of
-    # FunctionType.__get__). But this is actually a special case; MethodType is
-    # implemented in C and its attribute access doesn't go through
-    # __getattribute__.
-    # By wrapping FunctionType in _StaticFunctionType, we get the right result;
-    # similar to wrapping a function in staticmethod() at runtime to prevent it
-    # being bound as a method.
-    def __get__(self, obj: object, type: type | None) -> FunctionType: ...
-
 @final
 class MethodType:
     @property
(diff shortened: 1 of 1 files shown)


srittau (commented):
The comment of `_StaticFunctionType` hints at an old (2017) mypy hack, although it seems the implementation also uses a `member_descriptor` class. Did you try just changing the return type of `__get__` to a `Callable`?

tyralla:
No, I did not because `_StaticFunctionType` seems unnecessary when returning `Callable` instead of `FunctionType`.  However, if you think my change could cause unexpected trouble, we could also follow your suggestion, which leads to the same result in the following example (`MethodType2` vs. `MethodType4`):

```python

from types import FunctionType
from typing import Any, Callable

class MethodType1:
    @property
    def __func__(self) -> FunctionType: ...

class MethodType2:
    @property
    def __func__(self) -> Callable[[...], Any]: ...

class Descriptor3:
    def __get__(self, obj: object, type: type | None) -> FunctionType: ...

class MethodType3:
    @property
    def __func__(self) -> Descriptor3: ...

class Descriptor4:
    def __get__(self, obj: object, type: type | None) -> Callable[[...], Any]: ...

class MethodType4:
    @property
    def __func__(self) -> Descriptor4: ...

reveal_type(MethodType1().__func__)  # note: Revealed type is "types.MethodType"
reveal_type(MethodType2().__func__)  # note: Revealed type is "def (Any) -> Any"
reveal_type(MethodType3().__func__)  # note: Revealed type is "types.FunctionType"
reveal_type(MethodType4().__func__)  # note: Revealed type is "def (Any) -> Any"
```

PR #11202, opened 2023-12-26
Make attributes required on AST type_param classes

jacobtylerwalls:
Refs python/cpython#106224


sobolevn (approved):
Yes, that's right. PR was backported to 3.12 as well.

Thank you!

PR #11206, opened 2023-12-27
Add support for the refresh_token grant type (RefreshTokenGrant) (#11…

### LINKED ISSUE #11205: types-oauthlib missing oauthlib.openid.connect.core.grant_types.refresh_token
I am getting the following error when using mypy in a project based on oauthlib 3.2.0:
```
error: Skipping analyzing "oauthlib.openid.connect.core.grant_types.refresh_token": module is installed, but missing library stubs or py.typed marker
```

It seems the following files are out of sync with the type annotations:
* https://github.com/oauthlib/oauthlib/blob/master/oauthlib/openid/connect/core/grant_types/__init__.py
* https://github.com/oauthlib/oauthlib/blob/master/oauthlib/openid/connect/core/grant_types/refresh_token.py

gfour:
This PR adds support for the refresh_token grant type (RefreshTokenGrant). Fixes #11205 

The changes are based on the following oauthlib files:

* https://github.com/oauthlib/oauthlib/blob/master/oauthlib/openid/connect/core/grant_types/__init__.py

* https://github.com/oauthlib/oauthlib/blob/master/oauthlib/openid/connect/core/grant_types/refresh_token.py

### DIFF
--- a/stubs/oauthlib/oauthlib/openid/connect/core/grant_types/refresh_token.pyi
+++ b/stubs/oauthlib/oauthlib/openid/connect/core/grant_types/refresh_token.pyi
@@ -1,0 +1,10 @@
+from _typeshed import Incomplete
+
+from .base import GrantTypeBase
+
+log: Incomplete
+
+class RefreshTokenGrant(GrantTypeBase):
+    proxy_target: Incomplete
+    def __init__(self, request_validator: Incomplete | None = None, **kwargs) -> None: ...
+    def add_id_token(self, token, token_handler, request): ...
--- a/stubs/oauthlib/oauthlib/openid/connect/core/grant_types/__init__.pyi
+++ b/stubs/oauthlib/oauthlib/openid/connect/core/grant_types/__init__.pyi
@@ -7,3 +7,4 @@ from .dispatchers import (
 )
 from .hybrid import HybridGrant as HybridGrant
 from .implicit import ImplicitGrant as ImplicitGrant
+from .refresh_token import RefreshTokenGrant as RefreshTokenGrant


stubs/oauthlib/oauthlib/openid/connect/core/grant_types/refresh_token.pyi
```diff
@@ -0,0 +1,11 @@
+from _typeshed import Incomplete
+from typing import Any
+
+from .base import GrantTypeBase
+
+log: Any
```
srittau, line 6: Nit:

```suggestion
log: Incomplete
```

(Or even better: `logging.Logger`, but that needs an import.)

srittau (changes requested):
Thanks! One nit below. Also, we should add `from .refresh_token import RefreshTokenGrant as RefreshTokenGrant` to the `__init__.pyi` file in this directory.

stubs/oauthlib/oauthlib/openid/connect/core/grant_types/refresh_token.pyi
gfour: True, I had seen that `python3 scripts/create_baseline_stubs.py oauthlib` generates `Incomplete` instead of `Any` but I used the latter since that was also used in `authorization_code.pyi` (in the same directory). I amended my original commit.

About the proposed `__init__.pyi` change, I had included that line but "[pre-commit.ci]" deleted it.

AlexWaygood: > About the proposed `__init__.pyi` change, I had included that line but "[pre-commit.ci]" deleted it.

if you use `from .refresh_token import RefreshTokenGrant as RefreshTokenGrant` instead of `from .refresh_token import RefreshTokenGrant`, type checkers and linters will recognise it as an explicit re-export, so pre-commit will leave it in for you

gfour: Thanks! I edited the PR.

srittau (approved):

srittau:
Thanks again. For future reference: Please don't use force push when making updates. We squash the commits anyway when merging, but force pushed make reviews harder.

gfour:
> Thanks again. For future reference: Please don't use force push when making updates. We squash the commits anyway when merging, but force pushed make reviews harder.

Noted, thanks!

PR #11207, opened 2023-12-28
requests: Use the `Any` trick in `HTTPError`

Akuli:
See #10875 for context.

Usually, when you get a `HTTPError`, the response will be set. Specifically, it will always be set when the requests library raises the error.

But people also use `HTTPError` as a generic "just show this generic error message" exception that doesn't include any details about what happened. For example, [paasta_tools/mesos_maintenance.py](https://github.com/Yelp/paasta/blob/master/paasta_tools/mesos_maintenance.py#L137) has plenty of code like this (as pointed out by mypy_primer):

```python3
    try:
        status = get_maintenance_status(mesos_config_path).json()
        status = status["get_maintenance_status"]["status"]
    except HTTPError:
        raise HTTPError("Error getting maintenance status.")
```

I think we should get the best of both worlds with the `Any` trick.


Akuli:
The mypy_primer output isn't really comparable to the previous PR, because people have already added redundant casts and `None` checks:
- https://github.com/python-poetry/poetry/pull/8541
- https://github.com/schemathesis/schemathesis/commit/b231ab6076b9276dbafb483506f1669ba405244d
- https://github.com/internetarchive/openlibrary/commit/5dc2dd1ed2e02410ecf2f03bb3c7dca3382aad2f

Also, mypy doesn't warn about unnecessary checks or casts: an `if` statement could make sense with `Response | Any`, and casting from `Response | Any` to `Response` produces a slightly different result than using `Any` trick.

hauntsaninja (approved):

PR #11209, opened 2023-12-28
Fixes #11208: import all events in gdb-stubs init file

### LINKED ISSUE #11208: gdb events are imported incorrectly
Hi,

There is an issue with `gdb/events.pyi` being imported incorrectly. The `gdb-stubs/__init__.pyi` file should do an `from .events import *` because the `gdb` module should see the `gdb.NewObjFileEvent` type.

Without this, for a simple file like this:

```py
# cat x.py
import gdb

x: gdb.NewObjFileEvent = 1
```
We get the following issue:
```
# ./.venv/bin/python3 -m mypy x.py
x.py:3: error: Name "gdb.NewObjFileEvent" is not defined  [name-defined]
    x: gdb.NewObjFileEvent = 1
       ^
```

Which should not be there.

Here are the versions I tested this on:
```
# ./.venv/bin/python3 -m pip freeze | egrep -e 'types-gdb|mypy'
mypy==1.3.0
mypy-extensions==1.0.0
types-gdb==12.1.4.5
```

And here is a proof that this type should be available from the `gdb` module:
```
# gdb --quiet --nx --batch --ex 'pi import gdb; print(gdb.NewObjFileEvent)'
<class 'gdb.NewObjFileEvent'>
```

And here is the GDB version I tested this on:
```
# gdb --version
GNU gdb (Ubuntu 12.1-0ubuntu1~22.04) 12.1
```

disconnect3d:
This commit/PR fixes #11208. It turned out that there is no `gdb.events` module in `gdb` and instead all events types are available in `gdb.*` namespace.

### DIFF
--- a/stubs/gdb/gdb/__init__.pyi
+++ b/stubs/gdb/gdb/__init__.pyi
@@ -11,7 +11,8 @@ from typing_extensions import TypeAlias
 import gdb.types
 
 # The following submodules are automatically imported
-from . import events as events, printing as printing, prompt as prompt, types as types
+from . import printing as printing, prompt as prompt, types as types
+from .events import *
 
 # Basic
 


Akuli:
This doesn't seem right to me:

```
$ gdb --quiet --nx --batch --ex 'pi import code; code.interact()'
Python 3.11.2 (main, Mar 13 2023, 12:18:29) [GCC 12.2.0] on linux
(InteractiveConsole)
>>> import gdb
>>> gdb.VERSION
'13.1'
>>> gdb.events.before_prompt
<gdb.EventRegistry object at 0x7f29fc83af70>
>>> gdb.before_prompt
  File "<console>", line 1, in <module>
AttributeError: module 'gdb' has no attribute 'before_prompt'
```

It seems like `gdb.events` contains global variables (`gdb.events.before_prompt`) whose types are defined in `gdb` (`gdb.EventRegistry`).

disconnect3d:
> This doesn't seem right to me:
> 
> (...)
> 
> It seems like `gdb.events` contains global variables (`gdb.events.before_prompt`) whose types are defined in `gdb` (`gdb.EventRegistry`).

Yeah, you are right. The event types are defined in the global namespace while the specific events to be hooked/connected with in `gdb.events`. 

So indeed, this PR is a bit flawed/wrong.

disconnect3d:
I'll prepare another, better PR

disconnect3d:
@Akuli Hmm what's the best way to implement it? Shall I move all event types to some _events.pyi and then import them from events.pyi to add typing for the names/objects in `gdb.events`? Won't such an import expose those through `gdb.events` as well?

I believe/hope this is the full list fwiw:
```py
from events import ThreadEvent, ContinueEvent, ExitedEvent, StopEvent, BreakpointEvent, NewObjFileEvent, ClearObjFilesEvent, InferiorCallEvent, MemoryChangedEvent, RegisterChangedEvent, NewInferiorEvent, InferiorDeletedEvent, NewThreadEvent
```

Akuli:
Here's how I would do it:
- `__init__.pyi` has `EventRegistry`, `ThreadEvent`, etc
- `events.pyi` imports `EventRegistry` from `__init__.pyi` (using `from gdb import EventRegistry`) and defines `new_thread: EventRegistry`. You can also import anything else you need.

The imports won't expose `EventRegistry` as `gdb.events.EventRegistry`. That only happens when the import uses `as` or `*`.

PR #11210, opened 2023-12-30
stdlib: `base64.b64decode` accepts `str` `altchars`

mayeut:
`altchars` can be passed as an ASCII string to the `base64.b64decode` function.

c.f. https://github.com/python/cpython/blob/88cb9720001295f82c7771ab4ebf20f3cd0b31fb/Lib/base64.py#L65-L88


stdlib/base64.pyi
```diff
@@ -27,7 +27,7 @@ if sys.version_info >= (3, 10):
     __all__ += ["b32hexencode", "b32hexdecode"]
 
 def b64encode(s: ReadableBuffer, altchars: ReadableBuffer | None = None) -> bytes: ...
-def b64decode(s: str | ReadableBuffer, altchars: ReadableBuffer | None = None, validate: bool = False) -> bytes: ...
+def b64decode(s: str | ReadableBuffer, altchars: str | ReadableBuffer | None = None, validate: bool = False) -> bytes: ...
 def standard_b64encode(s: ReadableBuffer) -> bytes: ...
 def standard_b64decode(s: str | ReadableBuffer) -> bytes: ...
 def urlsafe_b64encode(s: ReadableBuffer) -> bytes: ...
```
Akuli, line 33: Various other functions could be fixed similarly, basically anything that uses `_bytes_from_decode_data()`. For example, `urlsafe_b64encode()` does this and should be changed to `ReadableBuffer | str`:

```
def urlsafe_b64decode(s):
    """..."""
    s = _bytes_from_decode_data(s)
    s = s.translate(_urlsafe_decode_translation)
    return b64decode(s)
```

Akuli (commented):
The CI failure is unrelated.

stdlib/base64.pyi
mayeut: `urlsafe_b64encode()` does not do this (as the snippet shows, it's `urlsafe_b64decode` which is already correctly annotated).
I only checked the base64 functions as it's my area of interest & spotted an issue when running mypy.

After checking other functions using `_bytes_from_decode_data()`, `b32decode()` needs to be updated as well. I'll add a commit for this one.

Akuli (approved):
Thanks!

PR #11214, opened 2024-01-01
seaborn: Update and complete the relational module

hamdanal:
This PR completes the relational module similar to the previous ones.
In addition, I updated seaborn to the latest version and fixed the stubtest errors. I also audited the use of `list`, `dict`, and `Mapping` to take into account the variance of keys/values to reduce false positives.

Closes #11217


stubs/seaborn/METADATA.toml
```diff
@@ -1,4 +1,4 @@
-version = "0.13.0"
+version = "0.13.*"
```
AlexWaygood, line 1: We've had our tests broken once before by new things being added in `seaborn` micro releases ( #11212), so I worry that that could happen again. I think I'd rather just bump the micro version here? It's not a massive issue if stubsabot files frequent PRs prompting us to bump the micro version, but stubtest failing every time a seaborn micro version is release (#11212) is more of a pain -- we can't take our time over fixing it like we can when stubsabot tries to bump the version number

stubs/seaborn/seaborn/_core/typing.pyi
```diff
@@ -1,14 +1,15 @@
...
+from typing import Any, Protocol, type_check_only
 from typing_extensions import TypeAlias
 
 from matplotlib.colors import Colormap, Normalize
 from numpy import ndarray
 from pandas import DataFrame, Index, Series, Timedelta, Timestamp
 
-class _SupportsDataFrame(Protocol):
+@type_check_only
+class SupportsDataFrame(Protocol):
```
AlexWaygood, line 12: I'd rather keep the leading underscore here. I might be wrong, but my understanding is that while _stubtest_ supports `@type_check_only`, mypy proper still doesn't assign any special meaning to the decorator. So by removing the leading underscore, we make it look like it's part of the public API, which means users might not think twice before importing it in runtime code if it's what their IDE suggests -- and mypy won't complain, because it still doesn't understand `@type_check_only`. (I have no idea whether other type checkers do or don't support `@type_check_only`.)

AlexWaygood (commented):
Not a full review (sorry!), just two quick comments:

stubs/seaborn/seaborn/_core/typing.pyi
hamdanal: I understand and agree with you that we should take care not to confuse users with exposing non-existing objects. But in this specific case, the class is already private "enough" because it is defined in the private module `_core` and in a file specific to type definitions. I think whoever reaches out to this module will know what they are doing especially that the class is not reexported anywhere public and you need to look into the source code to find it. Honestly all seaborn usages I've seen are with `import seaborn as sns` and `import seaborn.objects as so` so I don't think it is likely that this class will cause problems. Also, I am planning to add this same class in this same module to seaborn upstream because the author left [this comment](https://github.com/mwaskom/seaborn/blob/master/seaborn/_core/typing.py#L20-L24) which makes me believe he'll welcome such an addition.

That said if you feel strongly about it, I can happily change it back :)

hamdanal: Also happy new year :D

AlexWaygood: > But in this specific case, the class is already private "enough" because it is defined in the private module `_core` and in a file specific to type definitions.

Oh nice, I didn't notice that! This should be fine, then – thanks!

AlexWaygood: And, happy new year to you too!

srittau (approved):
Thanks!

PR #11218, opened 2024-01-02
updated type hints for paramiko to 3.4 - fix for CVE 2023-48795 (Terrapin Attack)

manfred-kaiser:
Paramiko 3.4 introduced some fixes against "Terrapin Attack" CVE 2023-48795.

Those fixes added some new attributes to the Transport class, which are added with this pull request.


manfred-kaiser:
This pull request allready includes the changes done by stubsabot in #11183

stubs/paramiko/paramiko/transport.pyi
```diff
@@ -26,6 +26,9 @@ class _KexEngine(Protocol):
 class Transport(Thread, ClosingContextManager):
     active: bool
     hostname: str | None
+    server_extensions: dict[str, str]
```
hauntsaninja, line 29:
```suggestion
    server_extensions: dict[str, bytes]
```
I think this might be a slightly different type based on https://github.com/paramiko/paramiko/blob/main/paramiko/transport.py#L2818

hauntsaninja (commented):
Thanks, looks good, one question

stubs/paramiko/paramiko/transport.pyi
manfred-kaiser: Thanks for the information 👍🏻 
I have pushed the new type.

PR #11219, opened 2024-01-02
socket: rework to match docs more closely, improve accuracy on linux

hauntsaninja:
Match the documentation more closely

Fixes some of #8101 , in particular I no longer see any missing constants on the Linux machines I have access to

### DIFF
--- a/stdlib/_socket.pyi
+++ b/stdlib/_socket.pyi
@@ -515,8 +515,8 @@ if sys.platform == "linux":
     SO_VM_SOCKETS_BUFFER_MIN_SIZE: int
     VM_SOCKETS_INVALID_VERSION: int  # undocumented
 
-if sys.platform != "win32" or sys.version_info >= (3, 9):
-    AF_LINK: int
+if sys.platform == "darwin":
+    AF_LINK: int  # BSD, macOS
 
 has_ipv6: bool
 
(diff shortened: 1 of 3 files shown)


hauntsaninja:
(This PR might be a little tricky to review, but I'm fairly confident. We check socket with stubtest quite tightly on macOS and Windows, and I've manually confirmed we fix stuff without regressions on Linux. Primer is also good)

srittau (approved):
LGTM, when skimming over it and going by your comment and the primer output. Please merge in a day or two, unless someone else wants to do a deep review.

AlexWaygood (approved):
Haven't done a proper review here (and I can't without access to a Linux machine :), but this all looks highly plausible, and I trust your judgement + stubtest!

How many stubtest errors would there be in CI if we removed the blanket allowlist entries for linux now, do you know? Is it low enough that we could remove the [blanket entries](https://github.com/python/typeshed/issues/8101#issuecomment-1789438742) and just enumerate all the errors? (Doesn't need to be done in this PR, I'm just curious!)

hauntsaninja:
Yes, just enumerating all the errors would be great. There's still quite a few errors... this PR only eliminates ones that are certainly errors, where we're missing things in the stubs.

The cases where we're missing things at runtime are trickier, because I don't know whether that's just the Linux machines I have access to or if it's real. So I might try to whittle down what I can, and when I give up I'll just allowlist how many ever are left.

PR #11221, opened 2024-01-02
fix pexpect delaybeforesend_type

PierreMarchand20:
See documentation https://github.com/pexpect/pexpect/blob/master/pexpect/pty_spawn.py#L136-L150


srittau (approved):
Thanks!

PR #11223, opened 2024-01-02
Move ellipsis definition to types

hauntsaninja:


stdlib/builtins.pyi
```diff
@@ -1914,9 +1914,13 @@ def __import__(
...
 
 if sys.version_info >= (3, 10):
-    # In Python 3.10, EllipsisType is exposed publicly in the types module.
-    @final
-    class ellipsis: ...
+    from types import EllipsisType
+
+    # Backwards compatibility hack for folks who relied on the ellipsis type
+    # existing in typeshed in Python 3.9 and earlier.
+    ellipsis = EllipsisType
```
JelleZijlstra, line 1921: I'd favor removing this at some point, but not sure what the fallout would be.

JelleZijlstra (approved):
Thanks, this is an improvement.

stdlib/builtins.pyi
hauntsaninja: Ah I had a better comment that it looks I forgot to commit. Yeah, agreed. One point in time to remove it is when we drop 3.9 support

PR #11229, opened 2024-01-04
Fix passlib.utils.handlers.TruncateMixin.using

### LINKED ISSUE #11104: `passlib.hash.bcrypt.using` doesn't know about parameters
As per the [passlib.hash.bcrypt documentation](https://passlib.readthedocs.io/en/stable/lib/passlib.hash.bcrypt.html) the following code should type check:

```python
from passlib.hash import bcrypt

bcrypt.using(rounds=13).hash("password")
```

but with pyright 1.1.336 and its builtin typeshed it currently fails with:

```
  /tmp/foo.py:3:14 - error: No parameter named "rounds" (reportGeneralTypeIssues)
```

Pyright's "Go to Type Definition" on `using` jumps to the following method definition, which indeed doesn't allow the supported parameters:
https://github.com/python/typeshed/blob/fd7ff195771b59eb1715037c7ed7d102da109def/stubs/passlib/passlib/utils/handlers.pyi#L26-L30

not-my-profile:
Fixes #11104.

### DIFF
--- a/stubs/passlib/@tests/test_cases/check_bcrypt_using_rounds.py
+++ b/stubs/passlib/@tests/test_cases/check_bcrypt_using_rounds.py
@@ -1,0 +1,3 @@
+from passlib.hash import bcrypt
+
+bcrypt.using(rounds=13).hash("password")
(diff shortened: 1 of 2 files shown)


srittau (approved):
Thanks!

Issue #11230, opened 2024-01-04
Some stubs for Lib/bdb.py are missing optional error `str` in return type.


buermarc:
Following `bdb.Bdb` methods optionally return an error `str` besides `None`
                                                                     
- def set_break                                                      
- def clear_break                                                    
- def clear_bpbynumber                                               
- def clear_all_file_breaks                                          
- def clear_all_breaks           

E.g. `set_break`: https://github.com/python/cpython/blob/35ef8cb25917bfd6cbbd7c2bb55dd4f82131c9cf/Lib/bdb.py#L394

PR #11231, opened 2024-01-04
Lib/bdb.py: add some missing optional str in return

buermarc:
Some methods of `bdb.Bdb` return either `None` or `str`.

- Issue: https://github.com/python/typeshed/issues/11230


srittau (approved):
Thanks!

PR #11232, opened 2024-01-04
[Pillow] Update to 10.2.*

srittau:


Avasam (approved):
LGTM. The use of IO here reflects runtime annotations. Looks like type hints are getting added to Pillow 👀

PR #11233, opened 2024-01-04
Update description of `_typeshed.Self`

srittau:
~~This was never marked as "stable" and all uses have been replaced with `typing_extensions.Self`.~~

### DIFF
--- a/stdlib/_typeshed/__init__.pyi
+++ b/stdlib/_typeshed/__init__.pyi
@@ -19,6 +19,10 @@ _T = TypeVar("_T")
 _T_co = TypeVar("_T_co", covariant=True)
 _T_contra = TypeVar("_T_contra", contravariant=True)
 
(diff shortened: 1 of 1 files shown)


stdlib/_typeshed/__init__.pyi
```diff
@@ -19,8 +19,9 @@ _T = TypeVar("_T")
 _T_co = TypeVar("_T_co", covariant=True)
 _T_contra = TypeVar("_T_contra", contravariant=True)
 
-# Use for "self" annotations:
-#   def __enter__(self: Self) -> Self: ...
+# Alternative to `typing_extensions.Self`, for use with `__new__`:
```
AlexWaygood, line 22: I believe the only cases where we still use this are the `__new__` methods of metaclasses, since the Self PEP specifically says that usage of typing.Self is prohibited in metaclasses

AlexWaygood (approved):

Issue #11220, opened 2024-01-02
Flaky mypy error on typeshed test_cases


hauntsaninja:
I've seen this a few times recently. It goes away if you retry:
```
mypy --platform win32 --python-version 3.8 on the test cases for 'requests': FAILURE

Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "/tmp/tmps_hms0ww/.venv/lib/python3.11/site-packages/mypy/__main__.py", line 37, in <module>
    console_entry()
  File "/tmp/tmps_hms0ww/.venv/lib/python3.11/site-packages/mypy/__main__.py", line [15](https://github.com/python/typeshed/actions/runs/7384516395/job/20087616635?pr=11219#step:5:16), in console_entry
    main()
  File "mypy/main.py", line 100, in main
  File "mypy/main.py", line [18](https://github.com/python/typeshed/actions/runs/7384516395/job/20087616635?pr=11219#step:5:19)2, in run_build
  File "mypy/build.py", line [19](https://github.com/python/typeshed/actions/runs/7384516395/job/20087616635?pr=11219#step:5:20)1, in build
  File "mypy/build.py", line 265, in _build
  File "mypy/build.py", line 2943, in dispatch
  File "mypy/build.py", line 3334, in process_graph
  File "mypy/build.py", line 3415, in process_fresh_modules
  File "mypy/build.py", line [21](https://github.com/python/typeshed/actions/runs/7384516395/job/20087616635?pr=11219#step:5:22)11, in fix_cross_refs
  File "mypy/fixup.py", line 52, in fixup_module
  File "mypy/fixup.py", line 1[27](https://github.com/python/typeshed/actions/runs/7384516395/job/20087616635?pr=11219#step:5:28), in visit_symbol_table
  File "mypy/lookup.py", line [31](https://github.com/python/typeshed/actions/runs/7384516395/job/20087616635?pr=11219#step:5:32), in lookup_fully_qualified
AssertionError: Cannot find module for resource.struct_rusage
```
https://github.com/python/typeshed/actions/runs/7384516395/job/20087616635?pr=11219

AlexWaygood:
The traceback always seems to mention struct_rusage, so I'm guessing it must be related to https://github.com/python/typeshed/commit/3edfd9037caeb91e531646da2bd7cd8ede5db23e. But I have no idea:

1. Why that commit would be causing mypy to crash
[...]

PR #11236, opened 2024-01-04
Remove macpath (removed in Python 3.8)

srittau:


AlexWaygood (approved):

AlexWaygood:
The regr_test.py error is unrelated, just another instance of #11220

PR #6749, opened 2021-12-29
`curses` is not available on Windows

AlexWaygood:
Refs #6734

PR #11241, opened 2024-01-04
Remove Windows checks from curses

rmccampbell:
Curses is available on Windows through a patch via the windows-curses package. See discussion in #6749.


JelleZijlstra:
Could you add some comments explaining the unusual situation here? (https://github.com/python/typeshed/pull/6749#issuecomment-1529012238)

rmccampbell:
Sure, I already added a comment in _curses.pyi but I'll add one in the top-level module too.

srittau (dismissed):
Just blocking merging for now, pending the outcome of the discussion in #6749. I'm not convinced that unconditionally making curses available on Windows in the right course of action.

Akuli:
I think we should just fix the conflicts and merge this. After a couple months of no progress, it seems like the "right" thing won't happen any time soon, and it's time to go with the practical thing and just give people the autocompletions they want.

We can always come up with a better solution for specifying platforms later, and add back the if statements if it then makes sense.

TheStef56:
Why is this pr still open and the check for windows in _curses.pyi (if sys.platform != win32) hasn't been removed yet?

Akuli:
This is easiest to review locally by telling git to ignore whitespace:

```
$ git diff -w origin/main HEAD
```

TheStef56:
> Thanks for doing this, and sorry it took us so long :)

Thank you for having listened to us Windows programmers 👍🏻🙏🏻

PR #11244, opened 2024-01-05
Update typing_extensions imports in stdlib

srittau:


AlexWaygood (approved):
Haven't read through the whole thing, but LGTM

AlexWaygood:
> mypy crashed locally, I suspect due to the typing.pyi changes. Will see what happens here.

oof, that's quite the error message from `mypy_test.py` there...

AlexWaygood:
I fixed the mypy crash in 5b6443a. It looks like, for whatever reason, we can't import `Literal` from `typing` in `builtins.pyi`; we have to import it from `typing_extensions`. Also, there's another alias in `zipfile.pyi` that we can't use `TypeAlias` for, for whatever reason.

PR #11250, opened 2024-01-05
Use PEP 570 syntax in stdlib

### LINKED ISSUE #11237: Switch to PEP 570 positional-only argument syntax

hauntsaninja:
Fixes #11237

This is the script I used: https://raw.githubusercontent.com/hauntsaninja/snippets/main/pep570.py

I haven't looked through the entire diff, let's see stubtest


hauntsaninja:
Fixing the stubtest issue...

hauntsaninja:
stubtest PR here: https://github.com/python/mypy/pull/16750

AlexWaygood:
> This is the script I used: https://raw.githubusercontent.com/hauntsaninja/snippets/main/pep570.py

Looks reasonable, though I wonder if it handles classmethods? You could probably special-case parameters named `cls` (and maybe even `mcls`, for metaclass classmethods?) in the same way you special-case parameters named `self`, right?

hauntsaninja:
Ah did I not push that change? Oops, looks like I did not. Good point on metaclasses, will do that as well

AlexWaygood:
Looks like there's only one instance of the old syntax that https://github.com/PyCQA/flake8-pyi/pull/461 detects in the stdlib and this PR doesn't fix:

https://github.com/python/typeshed/blob/1de5830a2703936a96a126248227d5c7db883674/stdlib/builtins.pyi#L204-L205

You could fix that as well by also special-casing parameters named `metacls` in the script (or just fix it manually with an additional commit).

Maybe it'd be best to revert the handful of changes stubtest has issues with for now? We can easily defer them until there's a mypy release with the stubtest fix

hauntsaninja:
It's basically all overloads containing positional-only args. Since this isn't user facing, I'd rather either just wait or use a pre-release mypy for stubtest CI

JelleZijlstra (approved):
Looked at a couple of files, looks great.

PR #11251, opened 2024-01-05
Implement Preset Argument in tarfile.open Function and Add Test Coverage

### LINKED ISSUE #11072: `tarfile.open()` is missing `preset` kwarg
In Python 3.9+:

> For modes `'w:xz'` and `'x:xz'`, [tarfile.open()](https://docs.python.org/3.9/library/tarfile.html#tarfile.open) accepts the keyword argument preset to specify the compression level of the file.

However, the stub seems to be missing that argument and `mypy` is complaining about it.

karimbaidar:
This pull request addresses the and Closes #11072 issue. The preset argument was missing from the `tarfile.open` function in the `tarfile.pyi` stub file. This argument is used to set the compression level when opening a `tarfile` in write mode with the xz method.

**Changes made:**
- Added preset: int | None = ... to the open function in `tarfile.pyi`.
- To verify these changes, a test script was written and placed in the tests directory. This script uses the tarfile.open function with the preset argument in various ways to cover different edge cases.

**How to Run the Test:**
- The test script can be run directly with the Python interpreter:
`mypy --custom-typeshed-dir=<path_to_the_typeshed_dir> tests/test_tarfile.py `

This will execute the script and print the output to the console.

**How the Test Works:**
The test script checks the following scenarios:
- **Valid preset values**: The script opens a tarfile in write mode with the xz method and a preset value ranging from 0 to 9. No errors should be raised in this case.
- **Invalid preset values**: The script attempts to open a tarfile with preset values of -1, 10, “high”, and 5.5. These values are invalid because preset should be an integer between 0 and 9 inclusive. The script checks that a ValueError or TypeError is raised in these cases.
- **Modes that don’t support preset**: The script attempts to open a tarfile in read mode and write mode (without the xz method) with a preset value. These modes do not support the preset argument, so the script checks that a tarfile.TarError is raised.

By covering these scenarios, the test script ensures that the preset argument is correctly recognized and handled by the tarfile.open function.


tests/test_tarfile.py
```diff
@@ -0,0 +1,47 @@
+import tarfile
+
+with tarfile.open("test.tar.xz", "w:xz") as tar:
+    pass
+
+# Test with valid preset values
+for preset in range(10):
+    with tarfile.open("test.tar.xz", "w:xz", preset=preset) as tar:
+        pass
```
srittau, line 9: We just need to verify that this works with `int` values:

```suggestion
tarfile.open("test.tar.xz", "w:xz", preset=0)
tarfile.open("test.tar.xz", "w:xz", preset=5)
tarfile.open("test.tar.xz", "w:xz", preset=9)
```

stdlib/tarfile.pyi
```diff
@@ -119,6 +119,7 @@ def open(
     debug: int | None = ...,
     errorlevel: int | None = ...,
     compresslevel: int | None = ...,
+    preset: int | None = ...,
```
srittau, line 122:
```suggestion
    preset: Literal[0, 1, 2, 3, 4, 5, 6, 7, 8, 9] | None = ...,
```

(`Literal` needs an import.)

tests/test_tarfile.py
```diff
@@ -0,0 +1,47 @@
...
+    with tarfile.open("test.tar.xz", "w:xz", preset=-1) as tar:
+        pass
+except ValueError:
+    pass
+
+try:
+    with tarfile.open("test.tar.xz", "w:xz", preset=10) as tar:
+        pass
+except ValueError:
+    pass
```
srittau, line 22: Similarly, this won't do anything in the tests. We'd need to check it like this:

```suggestion
tarfile.open("test.tar.xz", "w:xz", preset=-1)  # type: ignore
tarfile.open("test.tar.xz", "w:xz", preset=10)  # type: ignore
```

tests/test_tarfile.py
```diff
@@ -0,0 +1,47 @@
...
+    with tarfile.open("test.tar.xz", "r", preset=5) as tar:
+        pass
+except tarfile.TarError:
+    pass
+
+try:
+    with tarfile.open("test.tar.xz", "w", preset=5) as tar:
+        pass
+except tarfile.TarError:
+    pass
```
srittau, line 47: These are good tests, but won't work with our current annotations. We'd need overloads for that. I suggest to skip those tests for now.

srittau (changes requested):
Thanks! Looking at the tarfile module, I think we need to restructure it a bit, although this PR is fine for now.

Regarding the tests: We don't run the tests using a test runner, but just run mypy on them. See below for a few suggested changes.

tests/test_tarfile.py
AlexWaygood, line 1: This test is in the wrong directory: it should be in the `test_cases` directory. We have information in `test_cases/README.md` about how the tests in that directory work

AlexWaygood (changes requested):

tests/test_tarfile.py
karimbaidar: fixed at by adding it to the recommended directory as per the README.md file. Now the the file is renamed to check_tarfile.py and is placed in test_cases/stdlib/check_tarfile.py

AlexWaygood: Thanks!

tests/test_tarfile.py
karimbaidar: As suggested, skipped it for a time being.

tests/test_tarfile.py
karimbaidar: added the suggested changes accordingly

stdlib/tarfile.pyi
karimbaidar: fixed.

tests/test_tarfile.py
karimbaidar: adjusted.

srittau (approved):
Thanks!

Issue #11113, opened 2023-12-06
Can attributes named exactly like their type lead to issues?


Avasam:
From https://github.com/python/typeshed/pull/11085#discussion_r1412902472
> I am not sure this is safe in all type checkers, since some may understand this as a forward reference. I'd prefer to keep the alias.

CC @JelleZijlstra 

Can the following code lead to problems?
```py
class A: ...

class B:
  A: A
```

Or should we do:
```py
from typing_extensions import TypeAlias

class A: ...

_A: TypeAlias = A

class B:
  A: _A
```

If it is deemed safe, then this issue can be closed. Otherwise, it would be a good candidate for flake8-pyi.

Note that this is different than the following case, which is caught by checkers (at least by pyright):
```py
class Foo:
    Bar: type
    def bar(self, Bar: Bar) -> None: ... # Variable not allowed in type expression (reportGeneralTypeIssues)
```

Daverball:
It's worth noting that the ambiguity only exists in pyi files (and in a regular file with a `from __future__ import annotations` import), since in either case the declarations are order-independent and could potentially be inferred by the type checker in any arbitrary order.

Realistically it's not going to be completely arbitrary, but it's also difficult to prove that it's never an issue, especially with potential future revisions to the logic, but it's also probably in the type checker's best interest to detect this case, since it could otherwise result in `A` infinitely waiting on itself to finish inference and if you could deadlock a type checker this easily, then it's not very robust, but it may decide to report an error when it detects this case, rather than look for the symbol in an outer scope.

karimbaidar:
Python’s type hinting system allows you to specify the expected type of a variable or an attribute. In your example, `class B `has an `attribute A` that is expected to be of `type A`. This is perfectly valid Python code.

However, some type checkers might interpret `A: A` as a forward reference, i.e., they might think that `A` is referring to a type that has not been defined yet. This is because Python’s type hinting system supports forward references, where you can refer to a type that is defined later in the code.

The alternative approach you mentioned, using TypeAlias, is a way to avoid this potential confusion. By creating an alias for the type A, you ensure that there’s no ambiguity in the type hint.

While the Python language allows attributes to be named like their types, it can potentially lead to confusion or unexpected behavior with some type checkers. Using a type alias can help avoid these issues. It’s always a good idea to consider the clarity and readability of your code, especially when working with advanced features like type hinting.

Avasam:
[...]

PR #11255, opened 2024-01-07
[docutils] Add remaining node classes and functions

flying-sheep:
I checked that everything is included using this little script:

```py
import ast
from types import FunctionType

import rich.console
from docutils import nodes

DEPRECATED = {"ensure_str", "unescape", "reprunicode"}


def main() -> None:
    con = rich.console.Console()

    with open("stubs/docutils/docutils/nodes.pyi") as f:
        stubs = ast.parse(f.read(), f.name)
    names_stubs = dict.fromkeys(stmt.name for stmt in stubs.body if isinstance(stmt, (ast.ClassDef, ast.FunctionDef))).keys()
    names_real = dict.fromkeys(
        name
        for name, global_ in vars(nodes).items()
        if not name in DEPRECATED
        if isinstance(global_, (type, FunctionType))
        if not global_.__name__.startswith("_")
        if global_.__module__ == "docutils.nodes"
    ).keys()
    for name in names_real:
        is_missing = name not in names_stubs
        if is_missing:
            con.print(name, style="red")
        else:
            con.print(name, style="bold")

    if superfluous := [name for name in names_stubs if not name.startswith("_") if name not in names_real]:
        con.print(f"Superfluous items: {superfluous}", style="bold red")


if __name__ == "__main__":
    main()
```


stubs/docutils/docutils/nodes.pyi
```diff
@@ -77,6 +80,25 @@ class Node:
         ascend: bool = False,
     ) -> Node: ...
 
+# Left out
+# - def ensure_str (deprecated)
+# - def unescape (canonical import from docutils.utils)
```
flying-sheep, line 85: Self-explanatory

stubs/docutils/docutils/nodes.pyi
```diff
@@ -77,6 +80,25 @@ class Node:
...
+
+class Text(Node, str):
```
flying-sheep, line 87: Moved this up so the module layout matches the source file.

This makes it easier to compare

stubs/docutils/docutils/nodes.pyi
```diff
@@ -113,40 +135,231 @@ class Element(Node):
...
+class Labeled: ...
+
+# Root Element
 
 class document(Root, Structural, Element):
     transformer: Transformer
-    def copy(self) -> Self: ...
-    def deepcopy(self) -> Self: ...
-    def pformat(self, indent: str = "    ", level: int = 0) -> str: ...
-    def astext(self) -> str: ...
```
flying-sheep, line 143: as `document` inherits from `Element`, duplicating these is not necessary

flying-sheep:
No idea what the “Test / Run mypy on the test cases” failure is about:

> AssertionError: Cannot find module for resource.struct_rusage

doesn’t look like anything I’ve touched.

AlexWaygood:
> No idea what the “[Test / Run mypy on the test cases](https://github.com/python/typeshed/actions/runs/7438647669/job/20237637537?pr=11255)” failure is about:
> 
> > AssertionError: Cannot find module for resource.struct_rusage
> 
> doesn’t look like anything I’ve touched.

It's a known, unrelated issue:
- #11220

stubs/docutils/docutils/nodes.pyi
```diff
@@ -113,40 +135,234 @@ class Element(Node):
...
+
+node_class_names: list[str]
+
+# necessary to disambiguate type and field in NodeVisitor
+_Document = document
 
 class NodeVisitor:
-    def __init__(self, document: document): ...
+    optional: ClassVar[tuple[str, ...]]
+    document: _Document
```
flying-sheep, line 337: IDK if there’s a better way to do this

Avasam: I don't think it's needed for the field. But AFAIK there's no better way to deal with this for the parameter name (line below). See #11113

stubs/docutils/docutils/nodes.pyi
Avasam: There's a history of leaving out deprecated symbols for third-party stubs, I'd say it's fine.
But *if you want*, you *could* use the `typing_extensions.deprecated` decorator.

stubs/docutils/docutils/nodes.pyi
```diff
@@ -113,40 +135,234 @@ class Element(Node):
...
+    def __init__(self, document: _Document): ...
```
srittau, line 338:
```suggestion
    def __init__(self, document: _Document) -> None: ...
```

stubs/docutils/docutils/nodes.pyi
```diff
@@ -113,40 +135,234 @@ class Element(Node):
...
+
+class GenericNodeVisitor(NodeVisitor):
+    # all the visit_<node_class_name> methods
     def __getattr__(self, __name: str) -> Incomplete: ...
 
-def __getattr__(name: str) -> Incomplete: ...
+class TreeCopyVisitor(GenericNodeVisitor):
+    parent_stack: list[Node]
+    parent: list[Node]
+    def get_tree_copy(self) -> None: ...
```
srittau, line 353:
```suggestion
    def get_tree_copy(self) -> Node: ...
```

srittau (changes requested):
Thanks! A few – mostly nits – below.

stubs/docutils/docutils/nodes.pyi
```diff
@@ -2,17 +2,20 @@ import sys
 import xml.dom.minidom
 from _typeshed import Incomplete
 from abc import abstractmethod
-from collections.abc import Callable, Generator, Iterable, Iterator, Sequence
-from typing import Any, ClassVar, Literal, Protocol, SupportsIndex, TypeVar, overload
+from collections.abc import Callable, Generator, Iterable, Iterator, Mapping, Sequence
+from typing import Any, ClassVar, Literal, Protocol, SupportsIndex, TypeAlias, TypeVar, overload
```
srittau, line 6:
```suggestion
from typing import Any, ClassVar, Literal, Protocol, SupportsIndex, TypeVar, overload
from typing_extensions import TypeAlias
```

srittau (dismissed):

srittau:
I don't like the mypy primer output. One of the problems is that we annotate `Node.parent` using `Node | None`. But Sphinx often does the equivalent of `x = foo.parent` and then assumes that `x` is not `None`. I believe this is usually a valid assumption as – I guess – only the root node of a document has a `None` parent. For example https://github.com/sphinx-doc/sphinx/blob/587e01177deeaf10b4fbe3f567b99ec083238b7e/sphinx/util/nodes.py#L132-L134:

```python
    elif isinstance(node, nodes.classifier) and not node.source:
        # docutils-0.15 fills in rawsource attribute, but not in source.
        node.source = node.parent.source
```

In this case, it's valid to assume that `node` has a parent.

I see two possible solutions:

* The easy solution is to use the "Any trick" and annotate `Node.parent` as `Node | Any` instead of `Node | None`. This can lead to false negatives, but is preferable to the big number of false positives we get now.
* Add a `parent: Node` field to most sub-classes of `Element` (and possibly some `Node` sub-classes as well).

flying-sheep:
In a fully parsed document, that assumption holds, but it’s always possible to create nodes whereever you want and use them. And these don’t have parents.

I don’t like the “Any trick”. I do type checking to accurately handle all cases in my code. Corner cases like this are exactly what I want typing to be correct for: so I get notified that they exist and I can handle them before they become a problem.

flying-sheep:
I invested quite some time into this. Sad to see all the conflicts being introduced.

@adamtheturtle FYI this is a complete PR that adds *all* node classes, and you’re making this harder

Avasam:
> I don’t like the “Any trick”. I do type checking to accurately handle all cases in my code. Corner cases like this are exactly what I want typing to be correct for: so I get notified that they exist and I can handle them before they become a problem.

> Unless anything is incorrect, can this please be merged and nitpicks and design choices deferred to later?

I understand your feeling on this, and I agree, **but**, if you first introduce less disruptive changes (ie, by using the Any trick for example), then other maintainers are likely be more inclined to merge as-is (my understanding is that's the only thing blocking @srittau 's approval)
Then the change to "potentially `None`" can be the only thing that gets debated over in its own PR.

stubs/docutils/docutils/nodes.pyi
flying-sheep: I think in these two cases, it’s not necessary. One is for Python 2 only and therefore completely unnecessary, the other should just be imported from somewhere else, which is a single line change.

Avasam:
Trying to cleanup the "messages only changes" in the last primer output (`[-+](.+?: error:).+?\n([-+]\1.+?\n)+`) gives me the following:
```diff
- sphinx/addnodes.py: note: At top level:
- sphinx/addnodes.py:172: error: Unused "type: ignore" comment  [unused-ignore]
+ sphinx/util/nodes.py: note: In function "apply_source_workaround":
+ sphinx/util/nodes.py:130:21: error: Unsupported operand types for - ("None" and "int")  [operator]
+ sphinx/util/nodes.py:130:21: note: Left operand is of type "Union[int, None, Any]"
+ sphinx/util/nodes.py: note: At top level:
+ sphinx/util/docutils.py: note: In function "register_role":
+ sphinx/util/docutils.py:104:37: error: Argument 2 to "register_local_role" has incompatible type "Callable[[str, str, str, int, Inliner, dict[str, Any], Sequence[str]], tuple[list[Node], list[system_message]]]"; expected "Callable[[str, str, str, int, Inliner, dict[str, Any], list[str]], tuple[list[reference], list[reference]]]"  [arg-type]
- sphinx/util/docutils.py:124: error: Unused "type: ignore" comment  [unused-ignore]
- sphinx/util/docutils.py: note: In member "dispatch_visit" of class "SphinxTranslator":
- sphinx/util/docutils.py:588:13: error: "dispatch_visit" undefined in superclass  [misc]
- sphinx/util/docutils.py: note: In member "dispatch_departure" of class "SphinxTranslator":
- sphinx/util/docutils.py:605:13: error: "dispatch_departure" undefined in superclass  [misc]
- sphinx/util/docutils.py: note: In function "new_document":
- sphinx/util/docutils.py:636:44: error: Argument 2 to "document" has incompatible type "Reporter"; expected "Node"  [arg-type]
- sphinx/writers/html5.py: note: In member "depart_term" of class "HTML5Translator":
+ sphinx/roles.py: note: In function "setup":
+ sphinx/roles.py:445:45: error: Argument 2 to "register_local_role" has incompatible type "Callable[[str, str, str, int, Inliner, dict[str, Any], Sequence[str]], tuple[list[Node], list[system_message]]]"; expected "Callable[[str, str, str, int, Inliner, dict[str, Any], list[str]], tuple[list[reference], list[reference]]]"  [arg-type]
- sphinx/environment/adapters/toctree.py:169: error: Unused "type: ignore" comment  [unused-ignore]
+ sphinx/environment/adapters/toctree.py:422:20: error: Item "Node" of "Union[Node, Any]" has no attribute "get"  [union-attr]
+ sphinx/environment/adapters/toctree.py:426:21: error: Unsupported target for indexed assignment ("Node")  [index]
+ sphinx/transforms/i18n.py: note: In member "update_title_mapping" of class "_NodeUpdater":
+ sphinx/transforms/i18n.py:185:17: error: Incompatible types in assignment (expression has type "Node", variable has type "target")  [assignment]
+ sphinx/transforms/i18n.py: note: In member "update_autofootnote_references" of class "_NodeUpdater":
+ sphinx/transforms/i18n.py:202:14: error: List item 0 has incompatible type "Generator[Node, None, None]"; expected "footnote_reference"  [list-item]
+ sphinx/transforms/i18n.py:204:14: error: List item 0 has incompatible type "Generator[Node, None, None]"; expected "footnote_reference"  [list-item]
+ sphinx/transforms/i18n.py:215:17: error: Item "Node" of "Union[Node, Any]" has no attribute "remove"  [union-attr]
+ sphinx/transforms/i18n.py: note: In member "update_refnamed_references" of class "_NodeUpdater":
+ sphinx/transforms/i18n.py:243:45: error: List item 0 has incompatible type "Generator[Node, None, None]"; expected "reference"  [list-item]
+ sphinx/transforms/i18n.py:244:45: error: List item 0 has incompatible type "Generator[Node, None, None]"; expected "reference"  [list-item]
+ sphinx/transforms/i18n.py: note: In member "update_refnamed_footnote_references" of class "_NodeUpdater":
+ sphinx/transforms/i18n.py:267:59: error: List item 0 has incompatible type "Generator[Node, None, None]"; expected "footnote_reference"  [list-item]
+ sphinx/transforms/i18n.py:269:59: error: List item 0 has incompatible type "Generator[Node, None, None]"; expected "footnote_reference"  [list-item]
+ sphinx/transforms/i18n.py: note: In member "update_citation_references" of class "_NodeUpdater":
+ sphinx/transforms/i18n.py:285:59: error: List item 0 has incompatible type "Generator[Node, None, None]"; expected "citation_reference"  [list-item]
+ sphinx/transforms/i18n.py:286:59: error: List item 0 has incompatible type "Generator[Node, None, None]"; expected "citation_reference"  [list-item]
+ sphinx/transforms/i18n.py:414:59: error: Argument 4 to "publish_msgstr" has incompatible type "Optional[int]"; expected "int"  [arg-type]
+ sphinx/transforms/i18n.py:417:66: error: Argument 5 to "make_glossary_term" has incompatible type "Optional[int]"; expected "int"  [arg-type]
+ sphinx/transforms/i18n.py:480:36: error: Argument 4 to "publish_msgstr" has incompatible type "Optional[int]"; expected "int"  [arg-type]
+ sphinx/transforms/i18n.py:446: error: Unused "type: ignore" comment  [unused-ignore]
+ sphinx/transforms/i18n.py: note: In member "apply" of class "RemoveTranslatableInline":
+ sphinx/transforms/i18n.py:614:13: error: Item "Node" of "Union[Node, Any]" has no attribute "remove"  [union-attr]
+ sphinx/transforms/i18n.py:615:13: error: Unsupported left operand type for + ("Node")  [operator]
+ sphinx/transforms/i18n.py:615:13: note: Left operand is of type "Union[Node, Any]"
+ sphinx/search/__init__.py: note: In member "dispatch_visit" of class "WordCollector":
+ sphinx/search/__init__.py:234:19: error: Value of type "Union[Node, Any]" is not indexable  [index]
+ sphinx/search/__init__.py:237:62: error: Argument 1 to "_is_meta_keywords" has incompatible type "Element"; expected "meta"  [arg-type]
+ sphinx/search/__init__.py: note: At top level:
+ sphinx/search/__init__.py: note: In function "_visit_nodes":
+ sphinx/search/__init__.py:506:23: error: Value of type "Union[Node, Any]" is not indexable  [index]
+ sphinx/builders/_epub_base.py: note: In member "get_refnodes" of class "EpubBuilder":
+ sphinx/builders/_epub_base.py:196:23: error: Item "Node" of "Union[Node, Any]" has no attribute "attributes"  [union-attr]
+ sphinx/builders/_epub_base.py: note: In function "add_visible_links":
+ sphinx/builders/_epub_base.py:329:24: error: Incompatible return value type (got "tuple[Union[Node, Any], Any]", expected "tuple[Element, int]")  [return-value]
+ sphinx/builders/_epub_base.py:329:35: error: Item "Node" of "Union[Node, Any]" has no attribute "index"  [union-attr]
+ sphinx/builders/_epub_base.py:332:28: error: Incompatible return value type (got "tuple[Union[Node, Any], Any]", expected "tuple[Element, int]")  [return-value]
+ sphinx/builders/_epub_base.py:332:41: error: Item "Node" of "Union[Node, Any]" has no attribute "index"  [union-attr]
+ sphinx/builders/_epub_base.py: note: In member "add_visible_links" of class "EpubBuilder":
+ sphinx/builders/_epub_base.py:348:23: error: Item "Node" of "Union[Node, Any]" has no attribute "index"  [union-attr]
+ sphinx/builders/_epub_base.py:353:21: error: Item "Node" of "Union[Node, Any]" has no attribute "insert"  [union-attr]
+ sphinx/builders/_epub_base.py:358:21: error: Item "Node" of "Union[Node, Any]" has no attribute "insert"  [union-attr]
- sphinx/writers/latex.py: note: In member "visit_entry" of class "LaTeXTranslator":
- sphinx/writers/latex.py:1163:23: error: Item "None" of "Optional[Node]" has no attribute "parent"  [union-attr]
+ sphinx/writers/latex.py:1240:41: error: Argument 1 to "get_nested_level" has incompatible type "Union[Node, Any]"; expected "Element"  [arg-type]
+ sphinx/domains/citation.py: note: In member "note_citation" of class "CitationDomain":
+ sphinx/domains/citation.py:72:33: error: Incompatible types in assignment (expression has type "tuple[Any, Any, Optional[int]]", target has type "tuple[str, str, int]")  [assignment]
- sphinx/ext/autosectionlabel.py: note: In function "get_node_depth":
- sphinx/ext/autosectionlabel.py:27:20: error: Incompatible types in assignment (expression has type "Optional[Node]", variable has type "Node")  [assignment]
+ sphinx/writers/texinfo.py: note: In member "collect_node_names" of class "TexinfoTranslator":
+ sphinx/writers/texinfo.py:284:45: error: Value of type variable "_N" of "next_node" of "Node" cannot be "Titular"  [type-var]
- sphinx/builders/latex/transforms.py: note: In member "get_docname_for_node" of class "ShowUrlsTransform":
- sphinx/builders/latex/transforms.py:114:24: error: Incompatible types in assignment (expression has type "Optional[Node]", variable has type "Node")  [assignment]
- sphinx/builders/latex/transforms.py: note: At top level:
```

I think that something really interesting we can see with that primer output, is a lot of assumptions in sphinx code that a value will be a subclass of a `Node` when that can't statically be defined.

Especially for `parent` attribute. A good example is https://github.com/sphinx-doc/sphinx/blob/master/sphinx/environment/adapters/toctree.py#L422 where `subnode` is checked to be a `reference`, but `subnode.parent.parent` is assumed to be an `Element`

Or `# type: nodes.NodeSubclass` comments that could be removed and work with a basic Node type (most of `sphinx/transforms/i18n.py`)

Avasam:
Under the exact same reasoning of using "the Any trick", you could be cheeky and type `parent` as `parent: Self | Any`: Still more accurate/safer than just `Any` whilst avoiding forcing `isinstance` calls everywhere.

Quick test: https://github.com/Avasam/typeshed/pull/38#issuecomment-1913298633

flying-sheep:
Most parents are `Elements`, so I guess `Element | document | Any` might be an improvement?

`Self` is very wrong, a node’s parent is more often of a different type than the node itself, unless I misunderstood something, I think that would be a bad choice.

Avasam (approved):
Tentatively approving. The mypy_primer output now looks great.
With the exception of `List item 0 has incompatible type "Generator[Node, None, None]";` new positives which are nonsensical to me looking at the source code.

srittau (approved):
My previous points were addressed and the primer output looks good. Thanks, @flying-sheep and @Avasam for ironing out the remaining issues!

PR #11257, opened 2024-01-07
Fix "state" type annotations for tkinter "Canvas.create_*" methods

### LINKED ISSUE #11256: "Canvas.create_*" methods have invalid type hints for the "state" parameter
# Description

The type annotations of the `state` parameters of the "Canvas.create_*" methods are incorrect: they indicate that "active" is a valid value, when it is not (See https://www.tcl.tk/man/tcl8.6/TkCmd/canvas.html). Also, the "hidden" value is missing in the `Literal`:

Actual:
```python
state: Literal["normal", "active", "disabled"] = ...,
```

Expected:
```python
state: Literal["normal", "hidden", "disabled"] = ...,
```

# Steps to reproduce

## Environment

- OS: Windows
- Python: 3.11.7
- Tk: 8.6
- *mypy*: 1.8.0

## False negative
```python
import tkinter as tk
tk.Canvas().create_rectangle(0, 0, 1, 1, state="active")
```

This crashes (as expected) with the following error:
> _tkinter.TclError: bad state value "active": must be normal, hidden, or disabled

*mypy* passes

## False positive
```python
import tkinter as tk
tk.Canvas().create_rectangle(0, 0, 1, 1, state="hidden")
```

This works (as expected), but *mypy* (1.8.0) does not pass:
> error: No overload variant of "create_rectangle" of "Canvas" matches argument types "int", "int", "int", "int", "str"  [call-overload]

rdesparbes:
Fixes #11256


Akuli (approved):
Thanks! The annotations probably ended up this way, because many widgets have a `state` option that accepts `"active"`.

Issue #11254, opened 2024-01-07
types-protobuf fails pre-commit cache


Itay4:
when configuring pre-commit mypy to run with `types-protobuf`, for example:
```
  -   repo: https://github.com/pre-commit/mirrors-mypy
      rev: v1.8.0
      hooks:
        -   id: mypy
            additional_dependencies:
              - types-protobuf
```

it fails with the following:

```
mypy.....................................................................Failed
- hook id: mypy
- exit code: 2

mypy: can't read file '/home/runner/.cache/pre-commit/repoem44ysms/py_env-python3.10/lib/python3.10/site-packages//google': No such file or directory
```

i suspect its related to https://github.com/python/typeshed/pull/11249, but not sure - @AlexWaygood @hauntsaninja i would appreciate your insights here

andylamp:
Encountering the same issue as well, this was previously working....

srittau:
Neither #11249, not #11245, the two PRs included in the latest release, *should* have an influence here. I'm not sure what exactly is going on here, but types-protobuf doesn't add a `google` directory, but a `google-stubs` directory, so it's no wonder that mypy can't find it. Does downgrading to types-protobuf 4.24.0.4 help?

andylamp:
Hi @srittau, as I mentioned previously I do know a way to specific specific version within mypy in a pre-commit environment.

Using the default though, exhibits the behavior mentioned by the original author.

If you know a way to "pin" an additional depency version, let me know and I can test it.

srittau:
This problem was introduced by typeshed-internal/stub_uploader#109. Previously, the `google-stubs` directory was missing the `py.typed` file that marked the stubs as partial. I think this was an error according to the [typing spec](https://typing.readthedocs.io/en/latest/spec/distributing.html):

> For namespace packages (see [PEP 420](https://www.python.org/dev/peps/pep-0420)), stub-only packages should use the -stubs suffix on only the root namespace package. All stub-only namespace packages should omit `__init__.pyi` files. py.typed marker files are not necessary for stub-only packages, but similarly to packages with inline types, if used, they should be in submodules of the namespace to avoid conflicts and for clarity.

Cc @Avasam

Avasam:
If I understand correctly, "google" is a namespace package and the `py.typed` marker should be found in `typeshed/stubs/protobuf/google
/protobuf/` rather than `typeshed/stubs/protobuf/google
/` ?

[...]

PR #11261, opened 2024-01-08
protobuf: Prevent namespace package google from having a py.typed file at root

Avasam:
Short-term fix for #11254 (see discussion in the issue)

Long story short, assuming I got everything right: Spec says that namespace stubs-only packages should either not have a `py.typed` marker (which is already do for complete stubs) or that it should be in the submodules, not at root (we add a `py.typed` file at root for partial stubs).

mypy gets tripped up by the presence of the file. And it is not mypy's fault as it's within spec. This could be improved in stub_uploader. But the short term fix is simply to not generate a partial `py.typed` marker file for this stub at all.


srittau:
Let's hold off for now as stub_uploader should be fairly easy to fix.

Avasam:
Good to hear.
(for reference: https://github.com/typeshed-internal/stub_uploader/pull/119 )

srittau:
typeshed-internal/stub_uploader#122 should fix this issue.

Avasam:
@srittau Since you quickly worked on a proper solution, I'll just close this.

PR #11269, opened 2024-01-12
Convert AbstractContextManager from a protocol to an ABC

NeilGirdhar:
AbstractContextManager should not be a protocol or else the exit method, which is abstract, cannot be determined by type checkers to
be implemented.  This prevents users from calling super.  For details, see https://github.com/microsoft/pyright/issues/6965#issuecomment-1889704569 and https://github.com/microsoft/pyright/issues/6965

### DIFF
--- a/stdlib/contextlib.pyi
+++ b/stdlib/contextlib.pyi
@@ -4,7 +4,7 @@ from _typeshed import FileDescriptorOrPath, Unused
 from abc import abstractmethod
 from collections.abc import AsyncGenerator, AsyncIterator, Awaitable, Callable, Generator, Iterator
 from types import TracebackType
-from typing import IO, Any, Generic, Protocol, TypeVar, overload, runtime_checkable
+from typing import IO, Any, Generic, Protocol, TypeVar, overload
 from typing_extensions import ParamSpec, Self, TypeAlias
 
 __all__ = [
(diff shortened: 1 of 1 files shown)


NeilGirdhar:
This decorator also caused a lot of confusion here: https://github.com/pylint-dev/pylint/issues/1594

NeilGirdhar:
See https://discuss.python.org/t/can-we-make-abstractcontextmanager-exit-concrete/43025

srittau:
Apart from the necessary stubtest changes (and comment there), could you also add a comment to the stub to explain the situation for posterity?

NeilGirdhar:
Happy to do that, but there is some debate as to whether this is the right change, so I'll just wait until it's resolved.

stdlib/contextlib.pyi
```diff
@@ -40,7 +40,9 @@ _CM_EF = TypeVar("_CM_EF", bound=AbstractContextManager[Any] | _ExitFunc)
 @runtime_checkable
 class AbstractContextManager(Protocol[_T_co]):
     def __enter__(self) -> _T_co: ...
-    @abstractmethod
+    # While __exit__ is decorated with @abstractmethod in the runtime, it is also implemented.  Having the same decoration in the
```
JelleZijlstra, line 43: Whatever the outcome, we should make the same change to `AbstractAsyncContextManager.__aexit__`.

NeilGirdhar:
Making this a draft.  In an ideal world, we would have `Self` be a supported generic parameter.

JelleZijlstra:
Closing this as it currently breaks more than it fixes and the PR has been open for a long time.

PR #11273, opened 2024-01-14
`POEntry.msgstr_plural` is mistyped in polib

### LINKED ISSUE #11274: Incorrect type for `POEntry.msgstr_plural` in polib
See https://github.com/izimobil/polib/blob/master/polib.py#L845 that it's a `dict`, and https://github.com/izimobil/polib/blob/master/polib.py#L1674 and https://github.com/izimobil/polib/blob/master/polib.py#L1780 that keys are `int`s. 

This is likely done in `polib` because it's possible to have gaps in the `msgstr_plural` translations.

blaise-io:
Closes #11274

### DIFF
--- a/stubs/polib/polib.pyi
+++ b/stubs/polib/polib.pyi
@@ -70,7 +70,7 @@ class _BaseEntry:
     msgid: str
     msgstr: str
     msgid_plural: str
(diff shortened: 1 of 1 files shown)


srittau (approved):
Thanks!

PR #11276, opened 2024-01-15
Add `aware_datetime` option to plistlib functions

aisk:
Since https://github.com/python/cpython/pull/113363 and https://github.com/python/cpython/pull/113582 have been merged, the `plistlib` module in upcoming Python3.13 release, will have a new optional parameter `aware_datetime` in most of it's functions.

In addition, the `plistlib.loads` also accepts `str` type as inputs value.

### DIFF
--- a/stdlib/plistlib.pyi
+++ b/stdlib/plistlib.pyi
@@ -16,8 +16,23 @@ class PlistFormat(Enum):
 
 FMT_XML = PlistFormat.FMT_XML
 FMT_BINARY = PlistFormat.FMT_BINARY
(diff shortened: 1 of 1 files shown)


aisk:
The failed CI check https://github.com/python/typeshed/actions/runs/7528137651/job/20489713547?pr=11276 should not related to this change.

srittau (approved):
Thanks! Ideally, we'd have overloads to allow a `str` value only if `fmt` is `FMT_XML`, but this will do for now.

srittau:
(The CI failure is unrelated.)

PR #11277, opened 2024-01-15
Disable regression tests temporarily

srittau:
They fail intermittently, see #11220 for details.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -109,6 +109,7 @@ jobs:
   regression-tests:
     name: Run mypy on the test cases
     runs-on: ubuntu-latest
+    if: false  # Temporarily disabled, see #11220
     steps:
       - uses: actions/checkout@v4
       - uses: actions/setup-python@v4


AlexWaygood:
my worry is that the intermittent test failure indicates some kind of import cycle between the `os` and `resource` modules that may cause mypy to crash in "real life" outside of our test suite. But it is odd that it only seems to be happening with `regr_test.py`.

I also worry that these tests will quickly start to fail with mypy if we do this.

AlexWaygood:
If we need a stopgap to fix the intermittent failures, I'd honestly prefer to revert https://github.com/python/typeshed/commit/3edfd9037caeb91e531646da2bd7cd8ede5db23e. All the failures mention "Cannot find module for resource.struct_rusage", and the test failures started immediately after that commit, so I think it must be the cause (although I cannot explain it -- though I haven't done any in-depth investigation yet).

Avasam:
I have no idea either. But I'm down to revert it if investigation requires more time.

PR #11283, opened 2024-01-17
psycopg2: Add ReplicationCursor.consume_stream

fovc9:
[Here](https://www.psycopg.org/docs/extras.html#psycopg2.extras.ReplicationCursor.consume_stream) are the docs for this missing annotation

> consume_stream(_consume_, _keepalive\_interval_=_None_)
> 
> Parameters:
> * **consume** – a callable object with signature consume(msg)
> * **keepalive\_interval** – interval (in seconds) to send keepalive messages to the server
>
> ...
> The msg object passed to consume() is an instance of `ReplicationMessage` class
> ...
> This method also sends feedback messages to the server every `keepalive_interval` (in seconds). The value of this parameter must be set to at least 1 second, but it can have a fractional part. If the `keepalive_interval` is not specified, the value of status_interval specified in the `start_replication()` or `start_replication_expert()` will be used.


fovc9:
I'm not sure what the failed check is testing, but I don't think it's related to this PR

srittau (changes requested):
Thanks! One nit: As the return value of the callable is ignored, it should return `object`, instead of `Any`.

srittau:
CI failure is unrelated.

srittau (approved):

PR #11284, opened 2024-01-17
Added some typings to peewee stubs

mikeziminio:


AlexWaygood:
Don't worry about the "run mypy on the test_cases" error. It's flaky at the moment and we're not sure why: https://github.com/python/typeshed/issues/11220

srittau (approved):
Thanks! I didn't compare this to the implementation, but the annotations all look reasonable.

PR #11285, opened 2024-01-17
Make concurrent.futures.Future covariant in its type variable

li-dan:
It's convenient to have futures act covariantly. A toy example of code that relies on it to type check:

```python
from concurrent.futures import as_completed, Future, ThreadPoolExecutor
import time

class Parent:
    pass


class Child(Parent):
    pass


with ThreadPoolExecutor() as executor:
    future1 = executor.submit(lambda: Parent())
    future2 = executor.submit(lambda: Child())
    for future in as_completed([future1, future2]):
        reveal_type(future.result())
```

---

Most users of `Future` interact with the type via `Future.result()`, and `Future[Subclass].result()` is a safe substitute for `Future[Superclass].result()`.

It's not safe to substitute `Future[Subclass].set_result(obj)` for `Future[Superclass].set_result(obj)`, as `obj` may be an instance of `Superclass` but not `Subclass`. However, I think code calling `set_result()` is much less common as it's mainly used by `Executor` implementations.

Similarly, `DoneAndNotDoneFutures` is technically mutable, but I think it would be unusual for mutation to be used in practice, so `DoneAndNotDoneFutures[Subclass]` should be a safe substitute for `DoneAndNotDoneFutures[Superclass]`.


hauntsaninja:
Thank you, but it's often convenient for things to work covariantly ;-) Unfortunately for convenience, and unlike TypeScript, Python's type system has chosen not to have this form of unsoundness. I agree `set_result` is much less common, but `set_result` is not niche, so I think we probably cannot accept this PR. I'm not seeing any clever tricks that would make things work better in your case either. I'll leave the PR open for a bit in case another maintainer does, or has additional opinions.

AlexWaygood:
I agree with @hauntsaninja here; I don't think we can accept this, unfortunately :/

hauntsaninja:
I came up with a thing that could help in https://github.com/python/typeshed/pull/11291

li-dan:
Okay, thanks both for taking a look. I hope the approach in #11291 works.

PR #11286, opened 2024-01-17
Bump pyright-action to v2

jakebailey:
I just bumped the action's major version to upgrade it to node20; no changes to the API were made (but it's standard to bump major if the runtime used changes).

Not sure if you use a bot to upgrade these, so here's a PR 😄


AlexWaygood:
> Not sure if you use a bot to upgrade these, so here's a PR 😄

We don't — this is welcome, thanks!

AlexWaygood (approved):

PR #11288, opened 2024-01-18
WebOb: CacheControl descriptors are always allowed to accept `None`

Daverball:
Fixes a minor bug in the WebOb stubs I encountered recently.

Something like  the following is perfectly legal:

```python
response.cache_control.max_age = None
```

It will result in the same as `del` or setting to `False` for `exists_property` descriptors and the same as `del` for `value_property` descriptors if their default value is `None` and the same as setting it to its `none` value (or `True`) otherwise.

So in the above example `cache_control.max_age` would return `-1` after having been set to `None`.


Daverball:
Just dropping some source code references to aid with review:

https://github.com/Pylons/webob/blob/7bf29f1f9a0e90e1fee3cc4110c1a37d942101f2/src/webob/cachecontrol.py#L79-L90

https://github.com/Pylons/webob/blob/7bf29f1f9a0e90e1fee3cc4110c1a37d942101f2/src/webob/cachecontrol.py#L120-L132

hauntsaninja (approved):
Thanks!

PR #11291, opened 2024-01-18
concurrent.futures: allow as_completed to act covariantly

hauntsaninja:
Fixes #11285. I think this works? Not quite sure the best way to describe this, feel free to edit PR title.

### DIFF
--- a/stdlib/concurrent/futures/_base.pyi
+++ b/stdlib/concurrent/futures/_base.pyi
@@ -4,7 +4,7 @@ from _typeshed import Unused
 from collections.abc import Callable, Iterable, Iterator
 from logging import Logger
 from types import TracebackType
-from typing import Any, Generic, Literal, NamedTuple, TypeVar
+from typing import Any, Generic, Literal, NamedTuple, Protocol, TypeVar
 from typing_extensions import ParamSpec, Self
 
 if sys.version_info >= (3, 9):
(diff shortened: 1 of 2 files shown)


stdlib/concurrent/futures/_base.pyi
```diff
@@ -69,7 +75,17 @@ class Executor:
         self, exc_type: type[BaseException] | None, exc_val: BaseException | None, exc_tb: TracebackType | None
     ) -> bool | None: ...
 
-def as_completed(fs: Iterable[Future[_T]], timeout: float | None = None) -> Iterator[Future[_T]]: ...
+class _AsCompletedFuture(Protocol[_T_co]):
+    # as_completed only mutates non-generic aspects of passed Futures and does not do any nominal
+    # checks. Therefore, we can use a Protocol here to allow as_completed to act covariantly.
+    # See check_concurrent_futures.py
```
Akuli, line 81: File names tend to get outdated when refactoring.
```suggestion
    # There are tests to ensure that this works.
```

test_cases/stdlib/check_concurrent_futures.py
```diff
@@ -0,0 +1,23 @@
...
+
+class Parent:
+    ...
+
+
+class Child(Parent):
+    ...
+
+
+def check_as_completed_covariance() -> None:
```
Akuli, line 16: As far as I can tell, the original PR that just makes `Future` covariant would pass this test. The test basically checks that `Future` acts covariantly, and that's all it does.

Can you add a test that the previous PR fails and this PR passes? If you can't, or if the test becomes ends up so "niche" it doesn't do anything practical, I think it's a clear sign that we should just make the thing covariant. My instinct is that it's probably what we should do anyway, because it is practical and simple, but a good test case could convince me otherwise.

Akuli:
> feel free to edit PR title.

IMO your title is actually pretty good :)

hauntsaninja:
Thanks, I addressed feedback. Let me know what you think of the new test case. I've used `set_result` in real world code before, happy to describe use cases.

Akuli (approved):

Issue #11294, opened 2024-01-20
Add tuple key type for memoryview.__getitem__ and __setitem__


shz42:
`memoryview` should accept a length `ndim` tuple as the key of `__getitem__` and `__setitem__` according to [memoryview](https://docs.python.org/3/library/stdtypes.html?highlight=tuple#memoryview)

Maybe change
```diff
- def __getitem__(self, __key: SupportsIndex) -> int: ...
+ def __getitem__(self, __key: SupportsIndex | tuple[SupportsIndex, ...] ) -> int: ...

- def __setitem__(self, __key: SupportsIndex, __value: SupportsIndex) -> None: ...
+ def __setitem__(self, __key: SupportsIndex | tuple[SupportsIndex, ...], __value: SupportsIndex) -> None: ...
```

Example:
```Python
>>> class Foo:
...     def __index__(self):
...         return 0
>>> view = memoryview(bytearray(range(4)))  # [0, 1, 2, 3], ndim=1
>>> print(view[Foo(),])                     # getting with length 1 tuple
0
>>> view[Foo(),] = 4                        # setting with length 1 tuple
>>> print(view.tolist())
[4, 1, 2, 3]
>>> view = view.cast("b", (2, 2))           # [[4, 1], [2, 3]], ndim=2
>>> print(view[Foo(), 1])                   # getting with length 2 tuple
1
>>> view[Foo(), 1] = 5
>>> print(view.tolist())                    # setting with length 2 tuple
[[4, 5], [2, 3]]
```

srittau:
Makes sense to me. PR welcome!

PR #11296, opened 2024-01-21
Add tuple key type for memoryview.__getitem__ and __setitem__

shz42:
Fix https://github.com/python/typeshed/issues/11294


srittau (approved):
Thanks!

PR #11299, opened 2024-01-22
Add precise values for enum members where possible

erictraut:
…d [here](https://discuss.python.org/t/draft-of-typing-spec-chapter-for-enums/43496/9) and [here](https://github.com/python/typing-council/issues/11).

This is an experiment to understand the potential impact of this change: DO NOT MERGE


stubs/fpdf2/fpdf/enums.pyi
```diff
@@ -21,62 +19,62 @@ class CoerciveIntFlag(IntFlag):
     def coerce(cls, value: Self | str | int) -> Self: ...
 
 class WrapMode(CoerciveEnum):
-    WORD: str
-    CHAR: str
+    WORD = ...
```
AlexWaygood, line 22: You may have answered this already elsewhere, in which case sorry, but how are type checkers supposed to be able to infer the type of `WrapMode.WORD.value` with something like this? I think mypy is currently able to infer a `str` here, but I don't think it would be able to with this change.

My preference would be to use the runtime values wherever possible (as you're already doing elsewhere in this PR) but, where it's not possible, to introduce linting rules that enforce annotations for this kind of pattern, e.g.

```suggestion
    WORD: str = ...
```

AlexWaygood (commented):
Overall this looks great, thank you!

stubs/fpdf2/fpdf/enums.pyi
AlexWaygood: Reading the discuss.python.org thread, I see you propose using `_value_: str` for cases where it's hard for us to provide the exact runtime value in the stub for each member. That will work in most cases, but it won't work for enums with heterogenous data types for their member values, e.g.

```py
class Foo(Enum):
    BAR = 99999999999999999999999999999999999999999999999999999999
    BAZ = "baaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaz"
```

You might say that that's an antipattern, and I might agree with you, but we should still have a way to express that in the stubs if that's what the runtime is doing. The most important thing in a stub is for us to be accurate w.r.t. to the runtime, even if the runtime is making use of a bad antipattern that we wouldn't necessarily endorse when we were writing our own code.

erictraut: @AlexWaygood, if you think that the `_value_` proposal is too constraining, please respond to the discuss.python.org thread and make your case there. Most of the participants in that broader thread are probably not following the detailed PR comments, so they won't see your above concerns.

JelleZijlstra:
I took a brief look at the remaining errors. The third-party stubtest run still produces a lot of errors, but most look relatively easy to fix. The stdlib stubtest run just complains about a single enum in `ssl` that may be  doing something unusual.

erictraut:
I think this PR is now ready for review.

FWIW, this change is blocking the consideration of a [change to the typing spec](https://github.com/python/typing-council/issues/11).

AlexWaygood:
This PR is unfortunately still unmergeable for now, due to the `pre-commit.ci` failures: there are errors in the Ruff and flake8-pyi checks.

The Ruff errors seem to be an obvious bug where Ruff isn't distinguishing between different scopes properly in `.pyi` files; I'll file an issue with Ruff momentarily.

The flake8-pyi errors are also a bug, but might be hard for us to fix without some ugly hacks at flake8-pyi. (flake8-pyi is maintained by the typeshed team.) Currently, flake8-pyi disallows assignments without annotations in stub files, unless it detects that the assignment is in a class scope in an enum class. But [the logic that it uses to detect whether a class is an enum class is pretty crude](https://github.com/PyCQA/flake8-pyi/blob/5d0c5ebf0156e17a5bdaac0c8ac20fb19d72d86c/pyi.py#L700-L702), due to the fact that flake8-pyi doesn't have access to a semantic model of any kind (it works via very simple analysis of the AST). I can think of some workarounds here for flake8-pyi, but they won't be pretty.

srittau:
I'm fine with noqa'ing the warnings in the `CoersiveEnum`s. Maybe even do so for the whole file.

AlexWaygood:
> I'm fine with noqa'ing the warnings in the `CoersiveEnum`s. Maybe even do so for the whole file.

Good point, we could easily add an entry to `per-file-ignores` in our `.flake8` file here, rather than making changes over at flake8-pyi. It's an edge case that pretty rarely comes up, after all.

stdlib/ssl.pyi
```diff
@@ -316,8 +316,8 @@ class _ASN1Object(_ASN1ObjectBase):
     def fromname(cls, name: str) -> Self: ...
 
 class Purpose(_ASN1Object, enum.Enum):
-    SERVER_AUTH: _ASN1Object
-    CLIENT_AUTH: _ASN1Object
+    SERVER_AUTH = (129, "serverAuth", "TLS Web Server Authentication", "1.3.6.1.5.5.7.3.2")  # pyright: ignore[reportCallIssue]
+    CLIENT_AUTH = (130, "clientAuth", "TLS Web Client Authentication", "1.3.6.1.5.5.7.3.1")  # pyright: ignore[reportCallIssue]
```
AlexWaygood, line 320: What was pyright complaining about here, out of curiosity?

stubs/pynput/pynput/mouse/_base.pyi
```diff
@@ -8,41 +8,41 @@ from typing_extensions import Self
...
-        button26: int
-        button27: int
-        button28: int
-        button29: int
-        button30: int
-        scroll_down: int
-        scroll_left: int
-        scroll_right: int
-        scroll_up: int
+        button8 = ...
```
AlexWaygood, line 16: This change leads to a regression in how mypy infers the type of the value of this enum member. On `main`:
```
% MYPYPATH=stubs/pynput mypy -c "import pynput.mouse._base; reveal_type(pynput.mouse._base.Button.button8.value)" --platform linux
<string>:1: note: Revealed type is "builtins.int"
```

With this PR:
```
MYPYPATH=stubs/pynput mypy -c "import pynput.mouse._base; reveal_type(pynput.mouse._base.Button.button8.value)" --platform linux
<string>:1: note: Revealed type is "types.EllipsisType"
```

AlexWaygood (commented):
All these changes look good to me other than https://github.com/python/typeshed/pull/11299/files#r1560806163, which seems like a clear regression in user experience. The last remaining CI error will be fixed by https://github.com/PyCQA/flake8-pyi/pull/479; we can do a hotfix release for that after it's merged, to get CI passing here.

.flake8
```diff
@@ -10,6 +10,10 @@ per-file-ignores =
   *.py: NQA
   # TODO: Remove bare "Incomplete"s from stubs (Y065)
   *.pyi: Y065
+  # Ignore Y052 in this file: there are loads of false positives
+  # due to the fact that flake8-pyi doesn't understand subclasses of `CoerciveEnum`
+  # as being enum classes
+  stubs/fpdf2/fpdf/enums.pyi: Y052, Y065
```
srittau, line 16: It's annoying that flake8 doesn't provide a way to do this in the file itself.

stdlib/ssl.pyi
erictraut: When an enum instance is constructed and a tuple value is provided (like in this example), the tuple is expanded and passed to the constructor as individual positional arguments. In this case, the `__new__` and `__init__` methods should accept four arguments. The `__new__` and `__init__` methods for the `_ASN1ObjectBase` (which are auto-generated by the NamedTuple mechanism) do accept four arguments. But the `_ASN1Object.__new__` method accepts only a single positional argument called "oid". Pyright is noting this problem. This looks like a bug in the type stub because this code will not run as it's currently defined in the stub. Try pasting the following code into a ".py" file and running it.

```python
from enum import Enum
from typing import NamedTuple, Self

class _ASN1ObjectBase(NamedTuple):
    nid: int
    shortname: str
    longname: str
    oid: str

class _ASN1Object(_ASN1ObjectBase):
    def __new__(cls, oid: str) -> Self: ...
    @classmethod
    def fromnid(cls, nid: int) -> Self: ...
    @classmethod
    def fromname(cls, name: str) -> Self: ...

class Purpose(_ASN1Object, Enum):
    SERVER_AUTH = (
        129,
        "serverAuth",
        "TLS Web Server Authentication",
        "1.3.6.1.5.5.7.3.2",
    )
    CLIENT_AUTH = (
        130,
        "clientAuth",
        "TLS Web Client Authentication",
        "1.3.6.1.5.5.7.3.1",
    )
```

AlexWaygood: Interesting -- thanks for the explanation. I don't think there's a pre-existing bug in the stubs here -- the signatures look accurate to me. The runtime code is here, and is pretty readable: https://github.com/python/cpython/blob/c06be6bbb8d138dde50c0a07cbd64496bee537c5/Lib/ssl.py#L394-L419

Rather, the issue is that what the runtime is doing is pretty dynamic, and hard to represent statically in a a stub file. The runtime is assigning a single string value to each enum member in the class scope, but that string value is then passed to `_ASN1Object.__new__` as part of the creation of the enum class. `_ASN1Object.__new__` "expands" the single string value into a four-item tuple, which then becomes the value of the final enum member at runtime.

stubs/pynput/pynput/mouse/_base.pyi
erictraut: I'd prefer not to wait for a bug fix in mypy 1.10. That will take another several weeks. 

Do you know where we can find the values for these enum members? IIRC, it wasn't obvious from the sources. I suppose we could use introspection to determine their values, but I'd need help from someone on a Linux and Windows machine to do this.

Akuli: Here are the values on linux with `pynput==1.7.6`:

```
>>> from pynput.mouse import Button
>>> list(Button)
[<Button.unknown: None>, <Button.left: 1>, <Button.middle: 2>, <Button.right: 3>, <Button.scroll_up: 4>, <Button.scroll_down: 5>, <Button.scroll_left: 6>, <Button.scroll_right: 7>, <Button.button8: 8>, <Button.button9: 9>, <Button.button10: 10>, <Button.button11: 11>, <Button.button12: 12>, <Button.button13: 13>, <Button.button14: 14>, <Button.button15: 15>, <Button.button16: 16>, <Button.button17: 17>, <Button.button18: 18>, <Button.button19: 19>, <Button.button20: 20>, <Button.button21: 21>, <Button.button22: 22>, <Button.button23: 23>, <Button.button24: 24>, <Button.button25: 25>, <Button.button26: 26>, <Button.button27: 27>, <Button.button28: 28>, <Button.button29: 29>, <Button.button30: 30>]
```

The stub is a bit misleading. These are actually defined in `pynput.mouse._xorg`, not in `pynput.mouse._base`. There is another, different `Button` enum in `pynput.mouse._base`, but that one lacks most of the buttons:

```
>>> Button.__module__
'pynput.mouse._xorg'
>>> from pynput.mouse._base import Button
>>> Button.__module__
'pynput.mouse._base'
>>> list(Button)
[<Button.unknown: 0>, <Button.left: 1>, <Button.middle: 2>, <Button.right: 3>]
>>> Button.button8
  File "/usr/lib/python3.11/enum.py", line 789, in __getattr__
    raise AttributeError(name) from None
AttributeError: button8
```

JelleZijlstra: I don't see these in the source code at all (https://github.com/moses-palmer/pynput/blob/12acf84dc0f721d91a957da65311497acb664933/lib/pynput/mouse/_base.py#L33), are they added dynamically somewhere?

I tried installing pynput on a Linux server but it fails to import; apparently it requires a running X server. I don't have time right now to try to cajole it, but I might take another look later.

AlexWaygood: > I'd prefer not to wait for a bug fix in mypy 1.10. That will take another several weeks.

Somewhat adjacent to the point you're making, but what does pyright infer for the value of the `button8` member here @erictraut? I'm not sure this does qualify as a mypy bug really; I don't know how it could be expected to know that the value of the member really is an `int` here

erictraut: The `...` token means `Any` in this context, so that's what pyright assumes. Mypy should do the same.

AlexWaygood: I see. That would still be a regression, though, since currently mypy is able to infer `int`.

AlexWaygood: Following https://github.com/python/typeshed/pull/11299/commits/bd170fab90094ce6f711af36db556299234a4c58, here's what mypy now infers:

```
% MYPYPATH=stubs/pynput mypy -c "import pynput.mouse._base; reveal_type(pynput.mouse._base.Button.button8.value)" --platform linux
<string>:1: note: Revealed type is "Literal[0]?"
```

According to @Akuli's comment in https://github.com/python/typeshed/pull/11299#discussion_r1561336972, that's also incorrect, since the value of the `button8` member is `8`. Why don't you just use `8` for this member, and `13` for the `button13` member, etc., according to the values Akuli listed in https://github.com/python/typeshed/pull/11299#discussion_r1561336972?

erictraut: Ah, I missed @Akuli's post above. I added the linux-specific values. There are still two windows-specific values where the values are still unknown.

Avasam: They are tuples under Windows. The x1 and x2 values simply don't exist in the `_base.Button` enum, but does in `_win32.Button`:
https://github.com/moses-palmer/pynput/blob/12acf84dc0f721d91a957da65311497acb664933/lib/pynput/mouse/_win32.py#L48-L56

JelleZijlstra: Yes, we might want to do a round of review on these stubs to make them follow the implementation more closely, that would make it easier to verify correctness. That's out of scope for this PR, though.

erictraut: Yeah, I appreciate the desire to make these values perfectly match the implementation, but I'd appreciate it if we could unblock this PR (which is itself blocking a bunch of things) and then do a separate pass to refine these types. Are you OK with that?

stubs/pynput/pynput/keyboard/_base.pyi
```diff
@@ -25,71 +25,71 @@ class KeyCode:
...
-    f12: int
-    f13: int
-    f14: int
-    f15: int
-    f16: int
-    f17: int
-    f18: int
-    f19: int
-    f20: int
+    alt = 0
```
JelleZijlstra, line 28: It appears these values are actually KeyCode object, not ints. I'll try to push a fix.

JelleZijlstra (approved):

stubs/pynput/pynput/keyboard/_base.pyi
JelleZijlstra: Oh wait, that wasn't released. In the last release all the values are really 0.

JelleZijlstra: For reference: https://github.com/moses-palmer/pynput/commit/4ee72f900d92df1ae5c5dd1388a6553ef3f3eaa6. The last release was in January 2022.

JelleZijlstra:
I would like to merge this, regardless of what happens with the proposed spec change, since the change adds strictly more information to the stubs and as far as I can tell no longer regresses anything for mypy users.

cc @AlexWaygood since you added the "DO NOT MERGE" label.

AlexWaygood:
> cc @AlexWaygood since you added the "DO NOT MERGE" label.

I added that ages ago because Eric's PR description seemed to indicate that that was what he wanted. Happy for it to be removed now :-)

I agree that overall this PR improves things greatly for users of these stubs, and we should land a version of it regardless of whether the spec changes land. I'm still unsure about one or two of the specific changes — I'll add inline comments shortly

stubs/fpdf2/fpdf/enums.pyi
```diff
@@ -89,173 +87,173 @@ class RenderStyle(CoerciveEnum):
...
-    TWO_COLUMN_LEFT: Name
-    TWO_COLUMN_RIGHT: Name
-    TWO_PAGE_LEFT: Name
-    TWO_PAGE_RIGHT: Name
+    SINGLE_PAGE = "SinglePage"
+    ONE_COLUMN = "OneColumn"
+    TWO_COLUMN_LEFT = "TwoColumnLeft"
+    TWO_COLUMN_RIGHT = "TwoColumnRight"
+    TWO_PAGE_LEFT = "TwoPageLeft"
+    TWO_PAGE_RIGHT = "TwoPageRight"
```
AlexWaygood, line 132: At runtime, these look like strings, but they're actually `fpdf.syntax.Name` instances, so this isn't correct:

```pycon
Python 3.11.8 (main, Feb 15 2024, 19:32:18) [Clang 15.0.0 (clang-1500.1.0.2.5)] on darwin
>>> import fpdf.enums
>>> fpdf.enums.PageLayout.SINGLE_PAGE.value
'SinglePage'
>>> type(fpdf.enums.PageLayout.SINGLE_PAGE.value)
<class 'fpdf.syntax.Name'>
```

The same goes for several other enums in this file. I don't know about pyright, but this leads to mypy not correctly understanding the values of these enum members.

On `main`:
```
% MYPYPATH=stubs/fpdf2:stubs/Pillow mypy -c "import fpdf.enums; reveal_type(fpdf.enums.PageLayout.SINGLE_PAGE.value)"
<string>:1: note: Revealed type is "fpdf.syntax.Name"
```

With your PR branch:
```
% MYPYPATH=stubs/fpdf2:stubs/Pillow mypy -c "import fpdf.enums; reveal_type(fpdf.enums.PageLayout.SINGLE_PAGE.value)"
<string>:1: note: Revealed type is "Literal['SinglePage']?"
```

stdlib/ssl.pyi
AlexWaygood, line 320: I can't point to a concrete regression here, as this enum seems to be just too complex for mypy both before and after -- it infers the value of the members as `Any` both on `main` and on your PR branch. But I can't say that I _love_ this, since at runtime the values for these enums are actually `_ASN1Object` instances (where the members can be accessed by name as well as index), not tuples:

```pycon
>>> import ssl
>>> ssl.Purpose.SERVER_AUTH.value.nid
129
>>> ssl.Purpose.SERVER_AUTH.value.shortname
'serverAuth
```

But I get that the runtime is being somewhat dynamic here and that this is really difficult to represent statically in a stub, so I don't think there's a good solution here. This comment is non-blocking.

AlexWaygood (approved):

PR #11305, opened 2024-01-24
defusedxml.ElementTree: get `Element` from ElementTree instead of minidom

tyrossel:
When using defusedxml as a drop-in replacement of `xml.etree.ElementTree`, I got an error using the `fromstring` function: 

```python
import xml.etree.ElementTree as ET
import defusedxml.ElementTree as DET

el = DET.fromstring(filecontent)
subEl = el.find("Document", ns)
```

This PR fixes this problem, as I believe `xml.dom.minidom.Element` should only be used in `defusedxml/minidom.pyi`


stubs/defusedxml/defusedxml/ElementTree.pyi
```diff
@@ -1,8 +1,13 @@
 from _typeshed import Incomplete
 from collections.abc import Iterator, Sequence
 from typing import Any
-from xml.dom.minidom import Element
-from xml.etree.ElementTree import ElementTree, ParseError as ParseError, XMLParser as _XMLParser, tostring as tostring
+from xml.etree.ElementTree import (
+    Element as Element,
```
srittau, line 5: `Element` is no re-exported in the implementation, which is why our tests fail:

```suggestion
    Element,
```

srittau (changes requested):
Thanks! One fix below.

srittau (approved):

PR #11306, opened 2024-01-24
Add TensorFlow stubs

hoel-bagard:
This PR adds some functions and classes to the TensorFlow stubs (maybe most notably `tensorflow.keras.models.Model`).

This PR is stacked on top of https://github.com/python/typeshed/pull/11292.


hmc-cs-mdrissi:
Ping me when you think this is ready for a review pass. Mainly cross reference it with stubs I shared as those tend to have broader types.

PR #11308, opened 2024-01-24
Add type stubs for docutils.nodes.literal_block and docutils.nodes.Fi…

adamtheturtle:
…xedTextElement

These stubs closely match the implementation:

* `literal_block`: https://github.com/docutils/docutils/blob/b768e2626088711dec257b0847b563d02700a712/docutils/docutils/nodes.py#L1721
* `FixedTextElement`: https://github.com/docutils/docutils/blob/b768e2626088711dec257b0847b563d02700a712/docutils/docutils/nodes.py#L1117


adamtheturtle:
I'm not sure what to do given the above diff. Advice is welcome.

srittau (approved):
Yes, confusingly in primer output usually red is good (an error message less) and green is bad (an additional error message).

PR #11310, opened 2024-01-24
`posixpath.commonpath` supports `Iterable` parameter

Avasam:
`paths` is immediately passed to a `map`, an iterable should work. Fixes https://github.com/pypa/setuptools/pull/3979/files#diff-6c888d2f35e4149ad56292334032e9a216d1ab2b32a4df3c791d864c42dfad04R603


srittau (approved):
One slight issue in the implementation: It's doing `if not paths` to check for an empty sequence, which won't work with iterables. I'll send a PR to CPython as the fix is trivial.

PR #11315, opened 2024-01-25
Correct `_waiters` type

### LINKED ISSUE #11314: Add `_waiters` type hints to asyncio.locks
There's no public vars can be used to check Lock has waiters or not...
But it has `_waiters`

But my editor, used `typedshed` and says: `Unresolved attribute reference '_waiters' for class 'Lock'`
Screenshot 2024-01-25 at 16 44 31

Also Semaphore already has `_waiters` type hint... 
https://github.com/python/typeshed/blob/26452a7c68c6ad512be00b553f7a03f869f64be9/stdlib/asyncio/locks.pyi#L84-L86

But it has **wrong** type. It should be `Optional`, cause it's `None` on init:
Screenshot 2024-01-25 at 16 55 37

Olegt0rr:
Closes #11314

### DIFF
--- a/stdlib/asyncio/locks.pyi
+++ b/stdlib/asyncio/locks.pyi
@@ -47,6 +47,7 @@ else:
         ) -> None: ...
 
 class Lock(_ContextManagerMixin, _LoopBoundMixin):
+    _waiters: deque[Future[Any]] | None
     if sys.version_info >= (3, 10):
         def __init__(self) -> None: ...
     else:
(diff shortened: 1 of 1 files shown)


Olegt0rr:
> ```diff
> src/anyio/_backends/_asyncio.py:1654
> ```

https://github.com/agronholm/anyio/blob/1e60219d95bb6a3c1ddf0618634f8ee3eb566037/src/anyio/_backends/_asyncio.py#L1652-L1655

srittau (approved):
Thanks!

PR #11321, opened 2024-01-26
pywin32: Use PEP 570 positional-only argument syntax and fix arguments

Avasam:
The `pywin32` part of #11237

Split into multiple commits for ease of review:
1. The first commit assumes all c-extension modules are positional-only
1. The second commit removes redundant `__` prefixes
1. For the third commit I searched the source code for all instances of `PyArg_ParseTupleAndKeywords` and updated all methods in the stubs that should support kwargs, and tested which ones accepted default arguments (except for methods in like 2-3 credential-related classes in `_win32typing.pyi` that I could figure out how to instantiate). This includes a bunch of fixes to argument order and return types I noticed along the way.


Avasam:
That's an interesting error (at the very least the "remove leading double underscore" message should be updated):
```
error: win32.lib.pywintypes.error.__init__ is inconsistent, stub argument "self" should be positional or keyword (remove leading double underscore)
Stub: in file E:\Users\Avasam\Documents\Git\typeshed\stubs\pywin32\win32\lib\pywintypes.pyi:14
def (win32.lib.pywintypes.error, builtins.int, builtins.str, builtins.str)
Runtime: in file <string>
def (self, *args, **kw)
```

srittau (approved):

PR #11324, opened 2024-01-27
`tensorflow`: Add (and rename) aliases

hoel-bagard:
This PR adds aliases (taken from [here](https://github.com/hmc-cs-mdrissi/tensorflow_stubs/blob/main/stubs/tensorflow/_aliases.pyi)) so that they can be used in the future. It also renames aliases that do not have a runtime equivalent (by adding an `_`).

@hmc-cs-mdrissi [The previous PR](https://github.com/python/typeshed/pull/11306) keeps getting bigger because `tf.keras.Model` relies on many other types, so I will make smaller PRs adding those types before adding `tf.keras.Model`.


hoel-bagard:
@hmc-cs-mdrissi Should the type aliases declared in `_aliases.pyi` not start with an underscore since they are already in a private module ? It made more sense to me to have all the type aliases that do not have a runtime equivalent start with an `_`, but it seems that flake8 does not agree :thinking:

hmc-cs-mdrissi:
Yup. The convention is anything in _aliases.py is type checking only and has no equivalent at runtime. Partly motivated by that flake8 rule so no need to have underscore start in alias name.

stubs/tensorflow/tensorflow/_aliases.pyi
```diff
@@ -1,30 +1,55 @@
...
+_Gradients: TypeAlias = tf.Tensor | tf.IndexedSlices
 
 class _KerasSerializable1(Protocol):
     def get_config(self) -> dict[str, Any]: ...
 
 class _KerasSerializable2(Protocol):
     __name__: str
 
-KerasSerializable: TypeAlias = _KerasSerializable1 | _KerasSerializable2
+_KerasSerializable: TypeAlias = _KerasSerializable1 | _KerasSerializable2
```
hmc-cs-mdrissi, line 28: I think since you've unified aliases to all live here consistently they can all lose underscore prefix.

Except for that the PR looks good to me.

hoel-bagard: Done in c7d65fd

hmc-cs-mdrissi (approved):

JelleZijlstra:
This looks good but I suspect it conflicts with some of the other PRs I just merged, so I updated the branch and might push some changes to the branch to get it to pass.

PR #11325, opened 2024-01-27
`tensorflow`: Add some functions from the config module.

hoel-bagard:
Add some functions related to `PhysicalDevice`s.


hoel-bagard:

stubs/tensorflow/tensorflow/config/__init__.pyi
```diff
@@ -0,0 +1,11 @@
...
+
+from tensorflow.config import experimental as experimental
+
+class PhysicalDevice(NamedTuple):
+    name: str
+    device_type: str
+
+def list_physical_devices(device_type: None | str = None) -> list[PhysicalDevice]: ...
+def get_visible_devices(device_type: None | str = None) -> list[PhysicalDevice]: ...
+def set_visible_devices(devices: list[PhysicalDevice], device_type: None | str = None) -> None: ...
```
hmc-cs-mdrissi, line 11:
```suggestion
def set_visible_devices(devices: list[PhysicalDevice] | PhysicalDevice, device_type: None | str = None) -> None: ...
```

Reviewing implementation it looks like [here](https://github.com/tensorflow/tensorflow/blob/0ad2ed4a1fb54e9f3a53937b5c7abbfb16fc4703/tensorflow/python/eager/context.py#L1730) the code also supports single PhysicalDevice. Mostly checked as list as input type hint is usually yellow flag (variance), but there's isinstance list in that code so tuple/other Sequence would misbehave.

stubs/tensorflow/tensorflow/config/__init__.pyi
```diff
@@ -0,0 +1,11 @@
+from typing import NamedTuple
```
hmc-cs-mdrissi, line 1: For files that are incomplete stubs you should always add,

```python
def __getattr__(name: str) -> Incomplete: ...
```

The stubs I shared do not do this since codebase I work on I tend to prefer untyped methods being false positive vs false negative. But for typeshed/general users we don't want incomplete stubs to create false positives. So until module stubs has every publicly documented member it should always have  `__getattr__` Incomplete. Same comment for the other file.

stubs/tensorflow/tensorflow/config/experimental.pyi
```diff
@@ -0,0 +1,13 @@
...
+
+from tensorflow.config import PhysicalDevice
+
+class _MemoryInfo(TypedDict):
+    current: int
+    peak: int
+
+def get_memory_info(device: str) -> _MemoryInfo: ...
+def reset_memory_stats(device: str) -> None: ...
+def get_memory_usage(device: PhysicalDevice) -> bytes: ...
```
hmc-cs-mdrissi, line 11: I think this returns an int. The documentation says `Total memory usage in bytes.` but means total is int like 5000 bytes.

stubs/tensorflow/tensorflow/config/__init__.pyi
hoel-bagard: Yes, I did not try using a single PhysicalDevice, but I tried using a tuple and it does not work (it raises `ValueError: Unrecognized device`).

stubs/tensorflow/tensorflow/config/experimental.pyi
hoel-bagard: Fixed in 45ad7496c

stubs/tensorflow/tensorflow/config/__init__.pyi
hoel-bagard: Fixed in 927b79f6c

stubs/tensorflow/tensorflow/config/__init__.pyi
hoel-bagard: Fixed in 13afbe415

hmc-cs-mdrissi (approved):

stubs/tensorflow/tensorflow/config/experimental.pyi
```diff
@@ -0,0 +1,15 @@
...
+
+from tensorflow.config import PhysicalDevice
+
+class _MemoryInfo(TypedDict):
+    current: int
+    peak: int
+
+def get_memory_info(device: str) -> _MemoryInfo: ...
+def reset_memory_stats(device: str) -> None: ...
+def get_memory_usage(device: PhysicalDevice) -> int: ...
```
JelleZijlstra, line 12: Looks like this is deprecated:

```
@deprecation.deprecated(
    None,
    "Use tf.config.experimental.get_memory_info(device)['current'] instead.")
```

Consider marking it with `@typing_extensions.deprecated`.

JelleZijlstra (commented):
Thanks, one optional comment

stubs/tensorflow/tensorflow/config/experimental.pyi
hoel-bagard: Done in 9898dc763

PR #11326, opened 2024-01-27
`tensorflow`: add some tensorflow functions

hoel-bagard:
The functions have been checked against/fixed using the stubs available [here](https://github.com/hmc-cs-mdrissi/tensorflow_stubs/blob/main/stubs/tensorflow/__init__.pyi).

I have not included the overloads for `expand_dims` and `concat` as they do not pass the mypy CI.


hoel-bagard:

stubs/tensorflow/tensorflow/__init__.pyi
```diff
@@ -412,3 +412,24 @@ def convert_to_tensor(
     dtype_hint: _DTypeLike | None = None,
     name: str | None = None,
 ) -> Tensor: ...
+def expand_dims(input: _TensorCompatible, axis: int, name: str | None = None) -> Tensor: ...
+def concat(values: Sequence[_TensorCompatible] | _TensorCompatible, axis: int, name: str | None = "concat") -> Tensor: ...
```
hmc-cs-mdrissi, line 416: I see issue with concat. I'm not sure what issue with expand_dims though is? I don't notice a mypy error

```python
@overload
def expand_dims(input: _TensorCompatible, axis: int, name: str | None = None) -> Tensor: ...
@overload
def expand_dims(input: RaggedTensor, axis: int, name: str | None = None) -> RaggedTensor: ...
```

For concat I'm fine leaving that one out. The overload overlap is due to handling case of concat list of one or more ragged tensors + normal tensors, but Sequence[RaggedTensor | _TensorCompatible] overlaps with just Sequence[_TensorCompatible]

hmc-cs-mdrissi:
Looking at mypy primer the first two errors are correct (they are redefinitions of variable's type). The 3rd error looked wrong and then I learned Sequence is not actually protocol. So all 3 errors are correct.

stubs/tensorflow/tensorflow/__init__.pyi
hoel-bagard: I've re-added the overloads [here](https://github.com/python/typeshed/pull/11326/commits/11cbf77e97615f8da7c4996263cd93205f2ee75d) by making the `concat` version for `RaggedTensor` more specific ( `Sequence[RaggedTensor]` instead of `Sequence[RaggedTensor | _TensorCompatible]`).  It's more restrictive than it needs to be, but I suppose it's better than not having it at all.

hmc-cs-mdrissi (approved):

PR #11327, opened 2024-01-27
`tensorflow`: add `tf.train.CheckpointOptions` and other `tf.train` members.

hoel-bagard:
This adds `tf.train.CheckpointOptions` and the other `tf.train` members from [here](https://github.com/hmc-cs-mdrissi/tensorflow_stubs/blob/main/stubs/tensorflow/train/__init__.pyi).


hoel-bagard:
@hmc-cs-mdrissi I don't know how to fix the metaclass issue since `google._upb._message.MessageMeta` does seem not exists at runtime. Do you know how this should be handled ?

stubs/tensorflow/tensorflow/train/__init__.pyi
```diff
@@ -0,0 +1,100 @@
...
+        experimental_enable_async_checkpoint: bool = False,
+        # experimental_write_callbacks: None | list[Callable[[str], Any] | Callable[[], Any]] = None,
+        enable_async: bool = False,
+        # experimental_skip_slot_variables: bool = False,
+        # experimental_sharding_callback: tf.train.experimental.ShardingCallback | None = None,
+    ) -> None: ...
+    # def __copy__(self) -> Self: ...
+
+@final
+class Example(GeneratedProtocolMessageType):
```
hmc-cs-mdrissi, line 31: These types are protobuf generated so while there is tf.train.Example, it's not defined here. It's actually defined [here](https://github.com/python/typeshed/blob/main/stubs/tensorflow/tensorflow/core/example/example_pb2.pyi). You should re-export it in this file by doing,

```from tensorflow.core.example.example_pb2 import Example as Example```

Same for rest. mypy-protobuf is responsible for generating protobuf stubs and we have a script here to [build them](https://github.com/python/typeshed/blob/main/scripts/sync_tensorflow_protobuf_stubs.sh). You should not need `GeneratedProtocolMessageType` and can drop cpp_message.pyi

**Comment by hmc-cs-mdrissi** [comment]
If you re-export protobuf types from mypy-protobuf generated files that should fix your stubtest metaclass issues.

File: stubs/tensorflow/tensorflow/train/__init__.pyi (hunk 1 above)
Comment (hoel-bagard) [reply stubs/tensorflow/tensorflow/train/__init__.pyi]: @hmc-cs-mdrissi I've tried to re-export the protobuf generated types, but that did not fix the metaclass issues.

Comment (hmc-cs-mdrissi) [reply stubs/tensorflow/tensorflow/train/__init__.pyi]: Hmm looks like we are hitting this [issue](https://github.com/nipunn1313/mypy-protobuf/blame/7f4a558c00faf8fac0cd6d7a6d1332d1643cc08c/stubtest_allowlist.txt#L106). Well if mypy-protobuf produces inconsistent metaclasses I do not thing protobuf stubs should be hand modified. So we should disable those stubtest errors.

There is an allowlist file [here](https://github.com/python/typeshed/blob/main/stubs/tensorflow/%40tests/stubtest_allowlist.txt). You can add `tensorflow.train.BytesList`/along with rest of protobuf types to allowlist and leave a comment that stubtest does not pass for protobuf generated stubs.

Comment (hoel-bagard) [reply stubs/tensorflow/tensorflow/train/__init__.pyi]: Done in https://github.com/python/typeshed/pull/11327/commits/794e76935ddf3b5bb6957b99095f432e0706aca4

File: stubs/tensorflow/tensorflow/train/experimental.pyi [hunk 2]
```diff
@@ -0,0 +1,23 @@
+from _typeshed import Incomplete
+
+# @dataclasses.dataclass(frozen=True)
+# class ShardableTensor:
```
hmc-cs-mdrissi, line 4: For this one I am mainly waiting on either commented out stubs to be finished and not comment or from them to be removed and left for future pr.

hoel-bagard: This class was [added to the tensorflow repo last month](https://github.com/tensorflow/tensorflow/blame/6dddfa1b666fab4d070052311a312f62f2739605/tensorflow/python/checkpoint/sharding/sharding_util.py#L39), so it's not in the doc yet (nor in the latest runtime version). I realized this afterwards and left them comment out to be able to easily add them in the future (I'm assuming the classes will make their way into the next version of tensorflow).
Should I add a comment explaining this, or is it better to remove them for now and add them in a future PR once they have been added to a released version of tensorflow ?

hmc-cs-mdrissi: I recommend removing them. After next tensorflow release they can be added then.

hoel-bagard: Done in [080b155](https://github.com/python/typeshed/pull/11327/commits/080b1558d13f7f6a8b30a2522a3b5b31088fbe0f)

hmc-cs-mdrissi (approved):

stubs/tensorflow/tensorflow/train/__init__.pyi
```diff
@@ -0,0 +1,77 @@
...
+    Int64List as Int64List,
+)
+from tensorflow.core.protobuf.cluster_pb2 import ClusterDef as ClusterDef
+from tensorflow.core.protobuf.tensorflow_server_pb2 import ServerDef as ServerDef
+from tensorflow.python.trackable.base import Trackable
+
+class CheckpointOptions:
+    experimental_io_device: None | str
+    experimental_enable_async_checkpoint: bool
+    # experimental_write_callbacks: None | list[Callable[[str], Any] | Callable[[], Any]]
```
JelleZijlstra, line 23: Are these also due to be added in a future release? I'm OK with leaving them commented out for now, but please put a comment somewhere in the stub about it, so we don't forget why there is this commented out code here.

hoel-bagard: Yes, looking at the [2.12 doc](https://www.tensorflow.org/versions/r2.12/api_docs/python/tf/train/CheckpointOptions) they are not present.
One of them is already in 2.15, two are not present in a release yet but are present in [the tensorflow code](https://github.com/tensorflow/tensorflow/blame/8f1e353db44fb5d9b2ddfd762b475aefa23ac49f/tensorflow/python/checkpoint/checkpoint_options.py#L49).

I've removed the ones not present in 2.15, and added comments for the rest in 2ad4be9a2.

PR #11328, opened 2024-01-27
`tensorflow`: add `tensorflow.math.reduce_variance`

hoel-bagard:


hmc-cs-mdrissi (approved):
Looks good and confirmed locally ragged input always reduces to dense tensor.

PR #11330, opened 2024-01-27
`tensorflow` fix `tensorflow.VariableSynchronization`

hoel-bagard:
See documentation [here](https://www.tensorflow.org/api_docs/python/tf/VariableAggregation)

@hmc-cs-mdrissi

### DIFF
--- a/stubs/tensorflow/tensorflow/__init__.pyi
+++ b/stubs/tensorflow/tensorflow/__init__.pyi
@@ -137,10 +137,10 @@ class VariableSynchronization(Enum):
     ON_READ = 3
 
 class VariableAggregation(Enum):
-    AUTO = 0
-    NONE = 1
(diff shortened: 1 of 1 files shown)


hmc-cs-mdrissi (approved):
Good catch.

hmc-cs-mdrissi:
General typeshed maintainer on tf prs I've marked approved. Those are ones I think look good for final review and being landed.

I'll review other tf prs after a few get merged.

JelleZijlstra:
Is it worth investigating why stubtest didn't catch this?

JelleZijlstra:
> Is it worth investigating why stubtest didn't catch this?

Possibly just because tensorflow defines this enum in some weird dynamic way. I looked at it though and can confirm these values are right.

PR #11332, opened 2024-01-28
`tensorflow`: Add `tensorflow.keras.callbacks` module

hoel-bagard:
The stub is originally taken from [here](https://github.com/hmc-cs-mdrissi/tensorflow_stubs/blob/main/stubs/tensorflow/keras/callbacks.pyi).


hoel-bagard:
@hmc-cs-mdrissi How can I mark the `Model` class as `Incomplete` so that mypy does not check it against the runtime version ?

stubs/tensorflow/tensorflow/keras/models.pyi
```diff
@@ -0,0 +1,6 @@
+from _typeshed import Incomplete
+
+import tensorflow as tf
+from tensorflow.keras.layers import Layer, _InputT, _OutputT
+
+class Model(Layer[_InputT, _OutputT], tf.Module, Incomplete): ...
```
hmc-cs-mdrissi, line 6: I'd discourage subclassing from Incompete. If your goal is to have Model class defined for callbacks but to not fill it's stub yet (for a future pr), I'd recommend

`Model = Incomplete`

Incomplete is just an alias for Any used for readability. Use Any when there's no good way to type hint it, incomplete when it is not yet done.

hmc-cs-mdrissi: Future note Model I'd recommend doing `tensorflow.keras.Model.__new__`, `tensorflow.keras.Model.build`, `tensorflow.keras.Model.call` in stubtest allowlist as exceptions for same reason those 3 methods are exceptions for layers [here](https://github.com/python/typeshed/blob/6d9f28a7cae8cdb2721ed18a77e384b8fee4eefd/stubs/tensorflow/%40tests/stubtest_allowlist.txt#L45). Like any other stubtest allowlist exception brief comment is helpful.

stubs/tensorflow/tensorflow/keras/callbacks.pyi
```diff
@@ -0,0 +1,146 @@
...
+class LearningRateScheduler(Callback):
+    def __init__(self, schedule: LearningRateSchedule, verbose: Literal[0, 1] = 0) -> None: ...
+
+class ProgbarLogger(Callback):
+    use_steps: bool
+    def __init__(
+        self, count_mode: Literal["steps", "samples"] = "samples", stateful_metrics: Iterable[str] | None = None
+    ) -> None: ...
+
+class TerminateOnNaN(Callback): ...
```
hmc-cs-mdrissi, line 146: This file needs getattr incomplete (same comment for other files) as there are a few more callbacks still missing. Or just add the rest. There's only 4 missing callbacks.

stubs/tensorflow/tensorflow/keras/models.pyi
hoel-bagard: Understood, thanks.

stubs/tensorflow/tensorflow/keras/callbacks.pyi
```diff
@@ -0,0 +1,146 @@
+from collections.abc import Iterable, Mapping, Sequence
+from typing import Any, Literal
+from typing_extensions import TypeAlias
+
+import tensorflow as tf
+from tensorflow.distribute.experimental.coordinator import RemoteValue
+from tensorflow.keras import Model
+from tensorflow.keras.optimizers.schedules import LearningRateSchedule
+
+_Logs: TypeAlias = Mapping[str, object] | RemoteValue[Any] | None
```
hmc-cs-mdrissi, line 10: Hmm, while this type hint is correct it might be too strict. A lot of codebases probably never handle `RemoteValue[Any]` case as it depends on distributed strategy being used and common local training will never see it.

If mypy primer is happy I'm fine leaving it like this. A more lax type hint would `Mapping[str, object] | None | Any` known as Any trick where Any means some other types may happen but not mentioned.

hmc-cs-mdrissi: Mapping[str, object] may also be better relaxed to Mapping[str, Any]. Usually object type you know what to expect based on your model/metrics.

stubs/tensorflow/tensorflow/keras/callbacks.pyi
hoel-bagard: I've added the missing callbacks in c533ddc1c

stubs/tensorflow/tensorflow/keras/callbacks.pyi
hoel-bagard: I've not used this type yet, so I don't know what would be the most helpful in practice. If the strict typing might cause somewhat obscure errors in many codebases, then I suppose it might be better to go with the lax option.

hoel-bagard:
Should I remove `tensorflow/tensorflow/distribute/experimental/coordinator.pyi` ? It's not needed anymore since I relaxed the type of `_Logs`, and while it is present in my runtime, it seems to be absent from the CI's runtime :thinking:

stubs/tensorflow/tensorflow/keras/callbacks.pyi
```diff
@@ -0,0 +1,188 @@
...
+        add_history: bool = False,
+        add_progbar: bool = False,
+        # model: Model[Any, object] | None = None,
+        model: Model | None = None,
+        **params: Any,
+    ) -> None: ...
+    # def set_model(self, model: Model[Any, object]) -> None: ...
+    def set_model(self, model: Model) -> None: ...
+    def set_params(self, params: dict[str, Any]) -> None: ...
+    def on_batch_begin(self, batch: int, logs: Mapping[str, object] | None = None) -> None: ...
```
hmc-cs-mdrissi, line 50: These should also be using _Logs

stubs/tensorflow/tensorflow/keras/callbacks.pyi
```diff
@@ -0,0 +1,188 @@
...
+    def __init__(
+        self, count_mode: Literal["steps", "samples"] = "samples", stateful_metrics: Iterable[str] | None = None
+    ) -> None: ...
+
+class TerminateOnNaN(Callback): ...
+
+class LambdaCallback(Callback):
+    def __init__(
+        self,
+        on_epoch_begin: Callable[[int, Incomplete], Incomplete] | None = None,
```
hmc-cs-mdrissi, line 179: The incomplete for second argument of callable is _Logs. The return type for all the callables is `object` (the return value is ignored.

For the last two callables they should be `Callable[[_Logs], object]`

stubs/tensorflow/tensorflow/keras/callbacks.pyi
```diff
@@ -0,0 +1,188 @@
...
+        on_epoch_end: Callable[[int, Incomplete], Incomplete] | None = None,
+        on_batch_begin: Callable[[int, Incomplete], Incomplete] | None = None,
+        on_batch_end: Callable[[int, Incomplete], Incomplete] | None = None,
+        on_train_begin: Callable[[Incomplete], Incomplete] | None = None,
+        on_train_end: Callable[[Incomplete], Incomplete] | None = None,
+        **kwargs: Any,
+    ) -> None: ...
+
+def __getattr__(name: str) -> Incomplete: ...
```
hmc-cs-mdrissi, line 188: As you have added missing callbacks `getattr` should be dropped now.

stubs/tensorflow/tensorflow/keras/callbacks.pyi
```diff
@@ -0,0 +1,188 @@
...
+        patience: float = 0,
+        verbose: Literal[0, 1] = 0,
+        mode: Literal["auto", "min", "max"] = "auto",
+        baseline: float | None = None,
+        restore_best_weights: bool = False,
+        start_from_epoch: int = 0,
+    ) -> None: ...
+
+class LearningRateScheduler(Callback):
+    def __init__(self, schedule: LearningRateSchedule, verbose: Literal[0, 1] = 0) -> None: ...
```
hmc-cs-mdrissi, line 166: Amusingly this does not require a LearningRateSchedule. Instead `Callable[[int], float | tf.Tensor] | Callable[[int, float], float | tf.Tensor]`  is [enough](https://github.com/keras-team/keras/blob/f9336cc5114b4a9429a242deb264b707379646b7/keras/callbacks.py#L2241). A `LearningRateSchedule` is callable, but any callable with those signatures also works fine.

stubs/tensorflow/tensorflow/keras/callbacks.pyi
```diff
@@ -0,0 +1,188 @@
...
+        save_best_only: bool = False,
+        save_weights_only: bool = False,
+        mode: Literal["auto", "min", "max"] = "auto",
+        save_freq: str | int = "epoch",
+        options: tf.train.CheckpointOptions | tf.saved_model.SaveOptions | None = None,
+        initial_value_threshold: float | None = None,
+    ) -> None: ...
+    def _save_model(self, epoch: int, batch: int | None, logs: _Logs) -> None: ...
+
+class TensorBoard(Callback):
```
hmc-cs-mdrissi, line 88: Minor Nit: Can we have defined callbacks in alphabetical order?

stubs/tensorflow/tensorflow/keras/callbacks.pyi
```diff
@@ -0,0 +1,188 @@
...
+class CSVLogger(Callback):
+    def __init__(self, filename: str, separator: str = ",", append: bool = False) -> None: ...
+
+class EarlyStopping(Callback):
+    monitor_op: Any
+    def __init__(
+        self,
+        monitor: str = "val_loss",
+        min_delta: float = 0,
+        patience: float = 0,
```
hmc-cs-mdrissi, line 157: I think patience makes more sense as an int. It's number of epochs and unsure 0.5 makes sense.

stubs/tensorflow/tensorflow/keras/callbacks.pyi
```diff
@@ -0,0 +1,188 @@
...
+        self, backup_dir: str, save_freq: str = "epoch", delete_checkpoint: bool = True, save_before_preemption: bool = False
+    ) -> None: ...
+
+class RemoteMonitor(Callback):
+    def __init__(
+        self,
+        root: str = "http://localhost:9000",
+        path: str = "/publish/epoch/end/",
+        field: str = "data",
+        headers: dict[Incomplete, Incomplete] | None = None,
```
hmc-cs-mdrissi, line 126: https://github.com/python/typeshed/blob/284afde9cc9f0b68805a1e0fa987fec4bad3974e/stubs/requests/requests/api.pyi#L8 is type for it as this is just passed to request. Mapping[str, str | bytes]. Or just import type alias from requests _HeadersMapping directly.

stubs/tensorflow/tensorflow/keras/callbacks.pyi
```diff
@@ -0,0 +1,188 @@
...
+        self,
+        log_dir: str = "logs",
+        histogram_freq: int = 0,
+        write_graph: bool = True,
+        write_images: bool = False,
+        write_steps_per_second: bool = False,
+        update_freq: int | Literal["epoch"] = "epoch",
+        profile_batch: int | tuple[int, int] = 0,
+        embeddings_freq: int = 0,
+        embeddings_metadata: dict[str, None] | None = None,
```
hmc-cs-mdrissi, line 110: Missing **kwargs? Although at same time I think all kwargs are always either ignored, deprecated, or an error in the end.

stubs/tensorflow/tensorflow/keras/callbacks.pyi
```diff
@@ -0,0 +1,188 @@
...
+    def __init__(
+        self,
+        callbacks: Sequence[Callback] | None = None,
+        add_history: bool = False,
+        add_progbar: bool = False,
+        # model: Model[Any, object] | None = None,
+        model: Model | None = None,
+        **params: Any,
+    ) -> None: ...
+    # def set_model(self, model: Model[Any, object]) -> None: ...
```
hmc-cs-mdrissi, line 47: I'd drop commented out stub and when Model is added later as generic ci checks should notice.

hmc-cs-mdrissi (approved):
A couple small comments, but overall looks good.

stubs/tensorflow/tensorflow/keras/callbacks.pyi
hoel-bagard: Done in b7438aa00

stubs/tensorflow/tensorflow/keras/callbacks.pyi
hoel-bagard: Done in 8fef1b215

stubs/tensorflow/tensorflow/keras/callbacks.pyi
hoel-bagard: Done in a2348bb34

stubs/tensorflow/tensorflow/keras/callbacks.pyi
hoel-bagard: Done in 18e72a6cc.

stubs/tensorflow/tensorflow/keras/callbacks.pyi
hoel-bagard: Done in 7a77bd3cc

stubs/tensorflow/tensorflow/keras/callbacks.pyi
hoel-bagard: Done in 2c86770bd

stubs/tensorflow/tensorflow/keras/callbacks.pyi
hoel-bagard: kwargs added in c219b9065

stubs/tensorflow/tensorflow/keras/callbacks.pyi
hoel-bagard: Done in 3b965af2a

stubs/tensorflow/tensorflow/keras/callbacks.pyi
```diff
@@ -0,0 +1,190 @@
...
+    def on_test_batch_begin(self, batch: int, logs: _Logs = None) -> None: ...
+    def on_test_batch_end(self, batch: int, logs: _Logs = None) -> None: ...
+    def on_test_begin(self, logs: _Logs = None) -> None: ...
+    def on_test_end(self, logs: _Logs = None) -> None: ...
+    def on_train_batch_begin(self, batch: int, logs: _Logs = None) -> None: ...
+    def on_train_batch_end(self, batch: int, logs: _Logs = None) -> None: ...
+    def on_train_begin(self, logs: _Logs = None) -> None: ...
+    def on_train_end(self, logs: _Logs = None) -> None: ...
+
+# A CallbackList has exact same api as a callback,
```
JelleZijlstra, line 35: But the stub doesn't have the `model` and `params` attribute that exist on `Callback`. Should these be added?

stubs/tensorflow/tensorflow/keras/callbacks.pyi
```diff
@@ -0,0 +1,190 @@
...
+        self,
+        root: str = "http://localhost:9000",
+        path: str = "/publish/epoch/end/",
+        field: str = "data",
+        headers: _HeadersMapping | None = None,
+        send_as_json: bool = False,
+    ) -> None: ...
+
+class TensorBoard(Callback):
+    _start_batch: int
```
JelleZijlstra, line 163: Any particular reason to include these private attributes? We generally aim to add private attributes only if someone explicitly asks for them.

hmc-cs-mdrissi: These probably come from my stubs and are attributes I've accessed before.

I'd expect most of these are rarely used and are fine just type ignoring them on my end, so neutral to dropping them.

hoel-bagard: Ok, I'll remove the private attributes then. We can always re-add them later if requested.

stubs/tensorflow/tensorflow/keras/callbacks.pyi
hoel-bagard: They are not documented, but they do exist, so I added them in 75773a5b9.

JelleZijlstra:
There are some CI failures. Haven't looked in detail but might need to edit the metadata file for the package.

PR #11329, opened 2024-01-27
`tensorflow`: Add members from `tensorflow.keras.metrics`

hoel-bagard:
These subs are taken from: https://github.com/hmc-cs-mdrissi/tensorflow_stubs/blob/main/stubs/tensorflow/keras/metrics.pyi

PR #11333, opened 2024-01-28
`tensorflow`: Add missing members to the `tensorflow.keras.layers` module.

hoel-bagard:
Some of the added stubs are taken from [here](https://github.com/hmc-cs-mdrissi/tensorflow_stubs/blob/main/stubs/tensorflow/keras/layers.pyi).


hoel-bagard:
@hmc-cs-mdrissi I've tried to add the `MultiHeadAttention`'s `__call__` to the `stubtest_allowlist.txt`, but it is still detected as an error. Did I misunderstand how to add functions to the stubtest allowlist ?

JelleZijlstra:
FYI this has a merge conflict now.

hoel-bagard:
@hmc-cs-mdrissi Quite a few things from your stubs are only present in tensorflow 2.15, would it be ok to bump the stubs' tensorflow version ? (in another PR, since it breaks a few things)

JelleZijlstra:
> Quite a few things from your stubs are only present in tensorflow 2.15, would it be ok to bump the stubs' tensorflow version ?

Yes, we generally aim for the latest version. And agree it should be done in a separate PR :)

hmc-cs-mdrissi:
Yes feel free to bump the tf version in it's own pr. Stubtest will fail on new arguments and need some methods fixed. Usually filling in new arguments for stubtest is main work needed for upgrading version.

hoel-bagard:
I'll do that then!

hmc-cs-mdrissi:
This has been a busy work week. I’ll take a pass over all your open prs that I haven’t already approved tomorrow.

hoel-bagard:
Thanks!
Make sure to take time to rest.

hoel-bagard:
@JelleZijlstra I don't understand why the CI is failing. It was passing after your merge commit, so I tried to revert the test commit I did earlier. This caused the CI to fail again, so I removed it, and now the CI is failing on the merge commit.

JelleZijlstra:
Some weird pytype issue, I'm unfortunately not sure how to address that. Seems like it also failed on the merge commit, though?

hoel-bagard:
I did a rebase, which made the CI re-run on the merge commit. It was passing on the first run.

JelleZijlstra:
Interestingly it also started to fail on #11416, which definitely has nothing to do with tensorflow.

hoel-bagard:
The same (or at least similar) issue happened on other tensorflow PRs before, but it got resolved after [this](https://github.com/python/typeshed/pull/11329#issuecomment-1936827549).

hoel-bagard:
So it's most likely an issue with pytype, and not with the PR itself. But the bug should've been resolved, so I'm wondering if the failing results somehow got cached :thinking:

JelleZijlstra:
Yes, it must have gotten introduced after #11329. It seems like we never upgraded pytype after Rebecca deployed a fix. I saw that CI was green and assumed it was fixed, but something must have been off.

I opened #11434 to upgrade pytype.

stubs/tensorflow/tensorflow/keras/layers/__init__.pyi
```diff
@@ -127,7 +129,7 @@ class Dense(Layer[tf.Tensor, tf.Tensor]):
         kernel_constraint: _Constraint = None,
         bias_constraint: _Constraint = None,
         trainable: bool = True,
-        dtype: _LayerDtype = None,
```
hmc-cs-mdrissi, line 130: One optional improvement, there's comment near here that some of these signatures can be simplified with Unpack TypedDict ([pep 692](https://peps.python.org/pep-0692/)). The upstream stubs actually do use that technique which saves on repetition of dtype/trainable/name/dynamic/etc. I see that [mypy](https://github.com/python/typeshed/issues/9710) is checked in this issue. 

Can we start using 692 in typeshed now (has it been used yet?) @JelleZijlstra

stubs/tensorflow/tensorflow/keras/layers/__init__.pyi
```diff
@@ -189,6 +191,133 @@ class Embedding(Layer[tf.Tensor, tf.Tensor]):
...
+        dtype: _LayerDtype | None = None,
+        dynamic: bool = False,
+        name: str | None = None,
+    ) -> None: ...
+
+class Conv2D(Layer[tf.Tensor, tf.Tensor]):
+    def __init__(
+        self,
+        filters: int,
+        kernel_size: int | tuple[int, int],
```
hmc-cs-mdrissi, line 203: Iterable[int]. The actual implementation is more lax then strict tuple and I suspect think list like [3, 3] is pretty common. Same comment for rest of tuple[int, int], usually runtime normalizes iterable first.

stubs/tensorflow/tensorflow/keras/layers/experimental/preprocessing.pyi
```diff
@@ -0,0 +1,6 @@
+import abc
+
+from tensorflow._aliases import TensorLike
+from tensorflow.keras.layers import Layer
+
+class PreprocessingLayer(Layer[TensorLike, TensorLike], metaclass=abc.ABCMeta): ...
```
hmc-cs-mdrissi, line 6: Hmm, while Layer[TensorLike, TensorLike] is true I suspect it'll be a bit noisy and overloads for calls would be helpful here. I'd expect that a lot/most code that uses any of IndexLookup layers will be passing specific tensor type (like tf.Tensor) and returning TensorLike will make a bunch of false positives. Specifically something like,

```
    @overload
    def __call__(self, inputs: tf.Tensor) -> tf.Tensor: ...
    @overload
    def __call__(self, inputs: tf.SparseTensor) -> tf.SparseTensor: ...
    @overload
    def __call__(self, inputs: tf.RaggedTensor) -> tf.RaggedTensor: ...  # type: ignore
```

Other thing is major aspect of this class that differs it from normal layer is adapt/compile methods which I'd include [here](https://www.tensorflow.org/api_docs/python/tf/keras/layers/experimental/preprocessing/PreprocessingLayer#adapt)

stubs/tensorflow/tensorflow/keras/layers/preprocessing/index_lookup.pyi
```diff
@@ -0,0 +1,9 @@
+from _typeshed import Incomplete
+
+import tensorflow as tf
+from tensorflow.keras.layers.experimental.preprocessing import PreprocessingLayer
+
+class IndexLookup(PreprocessingLayer):
```
hmc-cs-mdrissi, line 6: This class exists at runtime and is helpful for IntegerLookup/StringLookup, but it's undocumented internal class. So I'd lean to underscore prefix it here.

hmc-cs-mdrissi (commented):
Finished full pass. 

My main comment is avoid TensorLike being an output type for any layer with default call because it is Union and most of time (maybe almost always) overloads would be preferred in that situation. The layers that do support ragged/sparse tensors generally if a user passes a dense tensor will return dense and code after that is unlikely to work with all 3 types.

Otherwise thank you for continued work on stubs and my break/delay reviewing these.

stubs/tensorflow/tensorflow/keras/layers/__init__.pyi
hoel-bagard: I checked an both example below are valid.

```python
tf.keras.layers.Conv2D(2, [3,3])
tf.keras.layers.Conv2D(2, iter([3,3]))
```

Fixed in 16c5775fd.

stubs/tensorflow/tensorflow/keras/layers/experimental/preprocessing.pyi
hoel-bagard: Done in 37c668ba1 and 4228893b9.

For the data arg, the doc says it must be a Tensor or a numpy array, but in the example they give it's a list, so I added `DataSequence` as an acceptable type.

stubs/tensorflow/tensorflow/keras/layers/preprocessing/index_lookup.pyi
hoel-bagard: Done in 31134a1b1.

hoel-bagard:
Thank you for the review.

Having `TensorLike` as an output type would be inconvenient indeed, I'll try to keep that in mind in the future.

hmc-cs-mdrissi (approved):

hoel-bagard:
@JelleZijlstra It seems like pytype is causing CI to fail again. Is it happening in other PRs ?

stubs/tensorflow/tensorflow/keras/layers/__init__.pyi
JelleZijlstra: Still waiting for pytype as you can see in #9710

JelleZijlstra:
This failure looks different than before. Presumably it's triggered by commit https://github.com/python/typeshed/pull/11333/commits/37c668ba1943d116099bedb75d05274a7fdd31b1, since the one before that is green. However, the failures on that commit (https://github.com/python/typeshed/actions/runs/7941246262/job/21683390603) also look different from those on the most recent commit (https://github.com/python/typeshed/actions/runs/7941516630/job/21683958032?pr=11333). @rchen152 I'm afraid I have to ping you again.

hoel-bagard:
It's happening on #11444  too: https://github.com/python/typeshed/actions/runs/7949448698/job/21700633925?pr=11444

rchen152:
> This failure looks different than before. Presumably it's triggered by commit [37c668b](https://github.com/python/typeshed/commit/37c668ba1943d116099bedb75d05274a7fdd31b1), since the one before that is green. However, the failures on that commit (https://github.com/python/typeshed/actions/runs/7941246262/job/21683390603) also look different from those on the most recent commit (https://github.com/python/typeshed/actions/runs/7941516630/job/21683958032?pr=11333). @rchen152 I'm afraid I have to ping you again.

Looking. Sorry about all the weird pytype errors!

rchen152:
I have a pytype fix out for review, but I'm out the rest of this week; I'll try to do a release first thing on Monday.

rchen152:
Ok, I think pytype is finally happy.

JelleZijlstra:
But now there is a merge conflict. If you fix that, I can merge.

hoel-bagard:
@JelleZijlstra @rchen152  I merged main into the branch, but this caused pytype to fail again.

JelleZijlstra:
```
  File "/home/runner/work/typeshed/typeshed/./tests/pytype_test.py", line 88, in run_pytype
  File "/opt/hostedtoolcache/Python/3.11.8/x64/lib/python3.11/site-packages/pytype/load_pytd.py", line 641, in finish_and_verify_ast
    self._resolver.verify(mod_ast)
  File "/opt/hostedtoolcache/Python/3.11.8/x64/lib/python3.11/site-packages/pytype/load_pytd.py", line 314, in verify
    raise BadDependencyError(str(e), name) from e
pytype.load_pytd.BadDependencyError: Unreplaced NamedType: 'tf.DType', referenced from 'tensorflow'
```

hoel-bagard:
@JelleZijlstra Is it because there's something wrong with the PR ?

The PR doesn't use `tensorflow.DType`, and CI was passing before merging with main, so I don't know what's the cause of it failing now. I tried to look up what the error means in the pytype docs but couldn't find it.

Looking at `tensorflow/__init__.pyi`, there is a comment about `DType` causing a crash in pytype in the past. Could this be related ?

```main
# Explicit import of DType is covered by the wildcard, but
# is necessary to avoid a crash in pytype.
from tensorflow.dtypes import *
from tensorflow.dtypes import DType as DType
```

JelleZijlstra:
Pytype errors can be somewhat opaque. I wouldn't be surprised if this PR affects DType in some indirect way. To check this further I tried to merge main into some of your other open PRs about tensorflow, but seems like GitHub is having an outage at the moment.

rchen152:
Try doing this (remove the extra DType import in `__init__.pyi`, import `numpy` and `DType` differently in `_aliases.pyi` for the definition of `DTypeLike`):
```
diff --git a/stubs/tensorflow/tensorflow/__init__.pyi b/stubs/tensorflow/tensorflow/__init__.pyi
index 308690fe7..db10b0756 100644
--- a/stubs/tensorflow/tensorflow/__init__.pyi
+++ b/stubs/tensorflow/tensorflow/__init__.pyi
@@ -22,10 +22,7 @@ from tensorflow._aliases import AnyArray, DTypeLike, ShapeLike, Slice, TensorCom
 from tensorflow.autodiff import GradientTape as GradientTape
 from tensorflow.core.protobuf import struct_pb2
 
-# Explicit import of DType is covered by the wildcard, but
-# is necessary to avoid a crash in pytype.
 from tensorflow.dtypes import *
-from tensorflow.dtypes import DType as DType
 from tensorflow.experimental.dtensor import Layout
 from tensorflow.keras import losses as losses
 from tensorflow.linalg import eye as eye
diff --git a/stubs/tensorflow/tensorflow/_aliases.pyi b/stubs/tensorflow/tensorflow/_aliases.pyi
index 5566b8069..e26c80880 100644
--- a/stubs/tensorflow/tensorflow/_aliases.pyi
+++ b/stubs/tensorflow/tensorflow/_aliases.pyi
@@ -5,9 +5,11 @@ from collections.abc import Iterable, Mapping, Sequence
 from typing import Any, Protocol, TypeVar
 
+import numpy  # pytype needs the unaliased import to resolve DTypeLike
 import numpy as np
 import numpy.typing as npt
 import tensorflow as tf
+from tensorflow.dtypes import DType
 from tensorflow.keras.layers import InputSpec
 
 _T = TypeVar("_T")
@@ -52,7 +54,7 @@ SparseTensorCompatible: TypeAlias = TensorCompatible | tf.SparseTensor
 TensorOrArray: TypeAlias = tf.Tensor | AnyArray
 
 ShapeLike: TypeAlias = tf.TensorShape | Iterable[ScalarTensorCompatible | None] | int | tf.Tensor
-DTypeLike: TypeAlias = tf.DType | str | np.dtype[Any] | int
+DTypeLike: TypeAlias = DType | str | numpy.dtype[Any] | int
 
 ContainerTensors: TypeAlias = ContainerGeneric[tf.Tensor]
 ContainerTensorsLike: TypeAlias = ContainerGeneric[TensorLike]
```

Right now, pytype handles circular dependencies among pyi files in a rather ugly way, by just resolving names multiple times until everything is sorted. So tweaking the dependency structure can help with name resolution issues.

PR #10972, opened 2023-11-03
Various improvements to `Markdown` stubs

oprypin:

PR #11335, opened 2024-01-28
markdown: Restore wrongly deleted DoubleTagPattern

andersk:
This should not have been deleted in #10972, since it’s still there upstream.


oprypin (approved):

Avasam (approved):
Thanks!

PR #11344, opened 2024-01-30
Limit regression test workers to 1

srittau:
This should address #11220.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -109,15 +109,19 @@ jobs:
   regression-tests:
     name: Run mypy on the test cases
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


srittau:
It's fairly slow now. I wonder whether we could reduce the amount of tests run. For example, run only on Python 3.8 and 3.12. In the long run it could also make sense to separate out tests that need to run on multiple platforms or Python versions as currently we are running lots of redundant tests.

AlexWaygood:
Or we can shard it.

I'm experimenting locally with whether adding the `--no-incremental` flag fixes things, since the race condition (if it is a race condition) is likely due to different mypys concurrently trying to read the same cache.

PR #11349, opened 2024-01-31
`tensorflow`: add `tf.distribute.experimental.coordinator.RemoteValue`

hoel-bagard:
This class is only available in `tensorflow>=2.15` (cf [documentation](https://www.tensorflow.org/api_docs/python/tf/distribute/experimental/coordinator/RemoteValue)).


hoel-bagard:
This PR is getting the following import error, but I don't understand why:
```
error: tensorflow.distribute.experimental.coordinator failed to import. Maybe install the runtime package or alter PYTHONPATH? ModuleNotFoundError: No module named 'tensorflow.distribute.experimental.coordinator'
```


Probably unrelated, but in a previous PR I had to make [this commit](https://github.com/python/typeshed/pull/11332/commits/1fc8fcab49ecb5cf6086b74fbc6725be8d1e0185
) to change an import in order for CI to pass. I understand that TensorFlow's imports are a bit cursed, but I was a bit surprised to have to do that in the stubs.

hmc-cs-mdrissi:
https://github.com/python/typeshed/blob/7a6a749449e15ec591fe9b3b17a3ffe312236cf2/stubs/tensorflow/%40tests/stubtest_allowlist.txt#L40 is a similar example. For some documented types importing it from module vs accessing it by attribute so,

```python
from tensorflow.distribute.experimental.coordinator import RemoveValue
```

vs

```python
import tensorflow as tf

tf.distribute.experimental.coordinator.RemoteValue
```

one works, other crashes. I don't know any way to describe this behavior in stubs. It's undocumented behavior and side effect of weird import magic tensorflow does. Maybe future version/issue will fix this and make both ways work. I recommend stubtest allowlist with a comment of `cursed import magic`.

hmc-cs-mdrissi (approved):

PR #11352, opened 2024-02-01
`tensorflow` bump version to 2.15

hoel-bagard:
While doing this, I noticed that quite a few things that were in 2.12 (both in runtime and documentation) disappeared. [The TensorFlow documentation](https://www.tensorflow.org/guide/versions) says that they are following semver, but what they did between 2.12 and 2.15 doesn't seem backwards compatible to me. Did I miss something ?

For example see [`Tensor.consumers`](https://www.tensorflow.org/versions/r2.12/api_docs/python/tf/Tensor#consumers), it isn't marked as private or experimental, but it disappeared in 2.15.


hmc-cs-mdrissi:
```
says that they are following semver, but what they did between 2.12 and 2.15 doesn't seem backwards compatible to me. Did I miss something ?

```

What is consumers? Consumers is a concept for symbolic tensors and lacks meaning for eager/concrete tensors. tf.Tensor/tf.Variable hierarchy is a bit of mess as at runtime Variables do not subclass Tensor, but they dynamically mock most of their methods. The stubs claim Variable(Tensor), but that's technically a lie.  

In [2.14 release notes](https://github.com/tensorflow/tensorflow/releases/tag/v2.14.0) one of breaking changes was,

```
tf.Tensor

The class hierarchy for tf.Tensor has changed, and there are now explicit EagerTensor and SymbolicTensor classes for eager and tf.function respectively. Users who relied on the exact type of Tensor (e.g. type(t) == tf.Tensor) will need to update their code to use isinstance(t, tf.Tensor). The tf.is_symbolic_tensor helper added in 2.13 may be used when it is necessary to determine if a value is specifically a symbolic tensor.
```

and so consumers which only makes sense for symbolic tensor went down to it.

Also as a side effect some of stubtest complaints on Tensor/Variable make sense to ignore. When you make a tensor in tensorflow you do not actually make tf.Tensor in 2.14+. You make either SymbolicTensor or EagerTensor. But neither of those classes are documented/public. Those classes do have consumers defined (although for EagerTensor it always raises) Similarly variables/layers often have special `__new__`/metaclass tricks so that exact type at runtime is not simple one you would guess. 

So `[tensorflow.Tensor.consumers](error: tensorflow.Tensor.consumers is not present at runtime)` is technically true, but practically false and should just be stubtest ignored.

tensorflow.Tensor.graph/tensorflow.Tensor.op I believe are same issue. These are not defined on "tf.Tensor", but are defined on internal subclasses which is what you really get in practice. All 3 of those errors I'd allowlist.

hoel-bagard:
Thanks a lot for the detailed explanation!
I'll put back `Tensor.consumers` and add it (and the others) to the allowlist.

stubs/tensorflow/tensorflow/train/__init__.pyi
```diff
@@ -20,16 +20,14 @@ from tensorflow.python.trackable.base import Trackable
...
+    experimental_write_callbacks: None | list[Callable[[str], Any] | Callable[[], Any]]
     enable_async: bool
 
     def __init__(
         self,
         experimental_io_device: None | str = None,
         experimental_enable_async_checkpoint: bool = False,
-        # Uncomment when the stubs' TF version is updated to 2.15
-        # experimental_write_callbacks: None | list[Callable[[str], Any] | Callable[[], Any]] = None,
+        experimental_write_callbacks: None | list[Callable[[str], Any] | Callable[[], Any]] = None,
```
hmc-cs-mdrissi, line 30: If return value does not matter for a callable prefer object over Any. Return argument is covariant so in practice it behaves same. Just that conventionally object as a return means anything is fine, Any when correct type is difficult to describe.

stubs/tensorflow/tensorflow/train/__init__.pyi
```diff
@@ -51,7 +49,7 @@ class Checkpoint:
     def read(self, save_path: str, options: CheckpointOptions | None = None) -> _CheckpointLoadStatus: ...
     def restore(self, save_path: str, options: CheckpointOptions | None = None) -> _CheckpointLoadStatus: ...
     def save(self, file_prefix: str, options: CheckpointOptions | None = None) -> str: ...
-    # def sync(self) -> None: ...  # Uncomment when the stubs' TF version is updated to 2.15
+    def sync(self) -> None: ...  # Uncomment when the stubs' TF version is updated to 2.15
```
hmc-cs-mdrissi, line 52: Remove the stale comment

stubs/tensorflow/tensorflow/data/__init__.pyi
```diff
@@ -223,8 +223,17 @@ class Dataset(ABC, Generic[_T1]):
         name: str | None = None,
     ) -> Dataset[Dataset[_T1]]: ...
     def with_options(self, options: Options, name: str | None = None) -> Dataset[_T1]: ...
+    @overload
+    @staticmethod
+    def zip(*args: Collection[Dataset[Any]], name: str | None = None) -> Dataset[tuple[Any, ...]]: ...
```
hmc-cs-mdrissi, line 228: The *args types here seem weird. In particular *args: tuple is rarely right choice. Even though *args is a tuple the type for it actually is type for values of tuple. A function that takes *args of one or more ints is *args: int not *args: tuple[int, …]

You could maybe use PEP 646 Unpack operator to specify a specific tuple like here.

Something like *args: Unpack[tuple[Dataset[T1], Dataset[T2]]]

hmc-cs-mdrissi, line 228: In addition to collection I think it should be ContainerGeneric[Dataset[Any]] here. The documentation allows arbitrarily nested types holding dataset. The return type though can’t be described at that point beyond Any.

hmc-cs-mdrissi (approved):
Generally looks good. Most of my comments are on zip method and it’s high amount of expressiveness.

stubs/tensorflow/tensorflow/train/__init__.pyi
hoel-bagard: Understood, done in 4dfca8278.

stubs/tensorflow/tensorflow/train/__init__.pyi
hoel-bagard: Sorry, done in cd76d225d

stubs/tensorflow/tensorflow/data/__init__.pyi
hoel-bagard: Done in a7f33d175

hoel-bagard: Yes, I copy/pasted it from the datasets variant without thinking :disappointed: 
Done in 8325466c6.

PR #11353, opened 2024-02-01
Add more return types to the `numbers` module

AlexWaygood:
This PR represents a compromise between what the `numbers` module was supposed to be, and the reality of where we are today with Python's typing system.

This module is _supposed_ to provide a framework of ABCs that would allow for third-party implementations of Complex, Rational, Real and Integral numbers. As such, the annotations we _should_ be giving to the `Real` class should be something like this:

```py
class Real:
    @abstractmethod
    def __mod__(self, other: Real) -> Real: ...
    @abstractmethod
    def __round__(self, ndigits: int) -> Real: ...
    @abstractmethod
    def __divmod__(self, other: Real) -> tuple[Real, Real]: ...
```

Unfortunately, that doesn't work! The reason is that `fractions.Fraction` inherits from `numbers.Real`, and it overrides these abstract methods to provide concrete implementations. The concrete implementations return concrete types such as `int`, which type checkers don't recognise as subtypes of `Real` (because the `numbers` module has never been supported by static type checkers, and never will be):

https://github.com/python/typeshed/blob/547cbc748f48367cda36cf7410c85d612573a072/stdlib/fractions.pyi#L105-L106

As such, for every method I tried adding types to according to the above strategy, mypy emitted an error warning me that `fractions.Fraction` incompatibly overrode `numbers.Real`. (There were also similar errors for a `numbers.Real` subclass in our `Pillow` stubs.)

Instead of returning `Complex` from methods that really _should_ be returning `Complex`, therefore, this PR annotates those methods as returning `SupportsComplex`. `SupportsComplex` works pretty well here:
- It's a supertype of `numbers.Complex`, so it's not _incorrect_
- It's a good deal more precise than `Any`
- It's vague enough that it will permit third-party subclassers to return their own implementations of `numbers.Complex` from these methods, which is what the module was designed for anyway.

In a similar way, I've annotated methods that _should_ be returning `numbers.Integral` as returning `SupportsIndex`, and I've annotated methods that _should_ be returning `numbers.Real` as returning `SupportsFloat`.

I haven't attempted to add any new argument annotations as part of this PR. The return annotations were hard enough.


AlexWaygood:

That looks like the kind of diff I'd expect, given that type checkers don't support these ABCs at all. I think it's a net win:
- We introduce 3 new errors in third-party code
- But we get rid of 6
- 1 existing error gets a different, more verbose error message

stdlib/numbers.pyi
```diff
@@ -1,130 +1,154 @@
...
+    def __abs__(self) -> SupportsFloat: ...
     @abstractmethod
-    def conjugate(self): ...
+    def conjugate(self) -> _SupportsComplex: ...
     @abstractmethod
     def __eq__(self, other: object) -> bool: ...
 
+# See comment at the top of the file
+# for why some of these return types are purposefully vague
 class Real(Complex, SupportsFloat):
```
JelleZijlstra, line 73: I think Real needs to override some methods from Complex to say they support SupportsIndex rather than _SupportsComplex. For example, saying that `__sub__` should return SupportsFloat rather than SupportsComplex may make this error in ibis go away: https://github.com/ibis-project/ibis/blob/0db47322595f93065997c990920b3b89475a1f9e/ibis/util.py#L168

JelleZijlstra (approved):
Thanks, this is a good step forward.

However, we can probably do a bit better by using protocols that support a few more common unary dunder methods that exist on common number classes: `__abs__` and others.

Also, I don't think we can realistically annotate the parameters to binary dunder methods on these classes. You'd think that say `Integral.__add__` should accept any other Integral, but that's impossible to implement in a concrete class (how would it know to add itself to any other Integral that might exist somewhere?) and inherently an LSP violation, because Integral is a subclass of Complex, and `Complex.__add__` should allow any complex numbers. So I endorse keeping parameter annotations out of this PR.

stdlib/numbers.pyi
AlexWaygood: Unfortunately that wouldn't be sound, since it's perfectly legal to subtract a `Complex` from a ` Real`, and the result won't necessarily be a `Real`:

```pycon
>>> isinstance(1.2, numbers.Real)
True
>>> isinstance(1.2 - 1j, numbers.Real)
False
```

JelleZijlstra: Good point. Maybe an overload that takes a Real argument and returns SupportsFloat would help, but that sounds like a rabbit hole that might cause more problems.

AlexWaygood: Yeah, definitely not something for this PR

AlexWaygood:
> However, we can probably do a bit better by using protocols that support a few more common unary dunder methods that exist on common number classes: `__abs__` and others.

Okay, I've had a go! Let's see what primer says...

AlexWaygood:
That last commit made the typing a lot more complex, didn't change the primer output at all, and broke our test cases for `unittest.assertAlmostEqual`. I'm inclined to just revert it, and merge the PR as it was before that commit?

AlexWaygood:
> That last commit made the typing a lot more complex, didn't change the primer output at all, and broke our test cases for `unittest.assertAlmostEqual`. I'm inclined to just revert it, and merge the PR as it was before that commit?

Everything I tried locally seemed to break either our test cases for `pow` or our test cases for `assertAlmostEqual`. I've gone ahead and reverted the last commit for now, but I'd be happy if somebody managed to find a way to improve the types even more after this PR :)

PR #11358, opened 2024-02-04
`tensorflow` Add `tensorflow.summary` module 

hoel-bagard:
Part of the stubs are taken from [here](https://github.com/hmc-cs-mdrissi/tensorflow_stubs/blob/main/stubs/tensorflow/summary.pyi)


stubs/tensorflow/tensorflow/_aliases.pyi
```diff
@@ -29,6 +29,7 @@ KerasSerializable: TypeAlias = KerasSerializable1 | KerasSerializable2
 
 Slice: TypeAlias = int | slice | None
 FloatDataSequence: TypeAlias = Sequence[float] | Sequence[FloatDataSequence]
+IntDataSequence: TypeAlias = Sequence[int] | Sequence[FloatDataSequence]
```
hmc-cs-mdrissi, line 32: Probably want Sequence[IntDataSequence] on right side?

stubs/tensorflow/tensorflow/experimental/dtensor.pyi
```diff
@@ -0,0 +1,19 @@
...
+
+import numpy as np
+import numpy.typing as npt
+from tensorflow._aliases import IntDataSequence
+
+class Mesh:
+    def __init__(
+        self,
+        dim_names: list[str],
+        global_device_ids: npt.NDArray[np.uint] | IntDataSequence,
```
hmc-cs-mdrissi, line 11: Hmm would np.int also work? Would be a little surprised if it forced unsigned int.

stubs/tensorflow/tensorflow/summary.pyi
```diff
@@ -0,0 +1,61 @@
...
+) -> SummaryWriter: ...
+def create_noop_writer() -> SummaryWriter: ...
+def flush(writer: SummaryWriter | None = None, name: str | None = None) -> tf.Operation: ...
+def graph(graph_data: tf.Graph | tf.compat.v1.GraphDef) -> bool: ...
+def histogram(
+    name: str, data: tf.Tensor, step: int | None = None, buckets: int | None = None, description: str | None = None
+) -> bool: ...
+def image(
+    name: str,
+    data: tf.Tensor | npt.NDArray[np.uint8 | np.number[Any]],
```
hmc-cs-mdrissi, line 46: Is there any difference in `np.uint8 | np.number[Any]` vs  `np.number[Any]`?

hmc-cs-mdrissi: I'd also lean to have any npt.NDArray types with a specific data type to be defined in _aliases so we can have IntNDArray, FloatNDArray, AnyNDArray, etc and be consistent on our definitions.

hmc-cs-mdrissi (approved):
A few small comments. I'm neutral to specifying numpy array data type when so far we tend to avoid it (Tensor is not generic at moment over data type), but as long as we have reasonable number of type aliases it's fine to include them piece meal.

hoel-bagard:
> Tensor is not generic at moment over data type

Do you think it would be possible to make Tensor into a generic the same way numpy arrays are ? I'm guessing a lot of functions/methods would require overloads, and it might mean making other classes into generics. I'm not sure how practical/realistic this would be.

It would definitely require a decent amount of effort, but if it's to be done (or at least attempted), then it would be better to do it before adding other modules.

stubs/tensorflow/tensorflow/_aliases.pyi
hoel-bagard: Yes, fixed in fe656689c. Thanks for pointing it out.

stubs/tensorflow/tensorflow/experimental/dtensor.pyi
hoel-bagard: I suppose it would work as long as the values are not negative. The doc says:

> Values in this array should increment sequentially from 0.

Which to me implied no negative values, hence np.uint (looking at [the code](https://github.com/tensorflow/tensorflow/blob/2e86693cc76a2222750f7d807ed76b5672d1fc5c/tensorflow/dtensor/python/layout.py#L136), negative values would indeed not work).

hoel-bagard: I would have leaned towards keeping a strict uint, but `IntDataSequence` already allows negative number anyways, so I'll add np.int to be consistent.

stubs/tensorflow/tensorflow/summary.pyi
hoel-bagard: I've modified this in a2e5fcaac, I'm not sure if that's what you meant.

stubs/tensorflow/tensorflow/experimental/dtensor.pyi
hoel-bagard: Done in ebe2af975.

hmc-cs-mdrissi:
It is possible but I would prefer to not do it until pep 696 (type var defaults) is accepted and ready to use in typeshed so that users do not get an error if they leave out data types. Additionally I think complexity of stubs would increase dramatically with data types that plan was to leave them until much later. After tensorflow stubs are complete enough then start gradually adding data types. I’d rather have more stubs without data types then invest a lot in some stubs with data types.

PR #11361, opened 2024-02-04
Update test dependencies

AlexWaygood:
- Bump various test dependencies
- Add a few more ruff autofixes to CI
- Fix deprecation warnings with `ruff==0.2.0`:
  - The `tool.ruff.select` configuration now needs to be `tool.ruff.lint.select`
  - The `tool.ruff.isort` configuration now needs to be `tool.ruff.lint.isort`

Note that you don't see the deprecation warnings if you run ruff via pre-commit (only if you run ruff directly, e.g. via `ruff .`), since pre-commit [swallows all output from a tool if the tool exits with code 0](https://stackoverflow.com/questions/72895720/pre-commit-hook-does-not-echo-on-terminal/72898524#72898524).


AlexWaygood:
(I'm deliberately leaving black to its own PR)

JelleZijlstra (approved):

PR #11371, opened 2024-02-06
Further improve return types in the `numbers` module

AlexWaygood:
Followup to #11353 (fixes 2/3 new errors in https://github.com/python/typeshed/pull/11353#issuecomment-1921903363)


stdlib/numbers.pyi
```diff
@@ -10,18 +10,56 @@
...
+
+# _RealLike is a structural-typing approximation
+# of the `Real` ABC, which is not (and cannot be) a protocol
+#
+# NOTE: _RealLike can't inherit from _SupportsComplex,
+# because not all builtin types that we want to be understood
+# as subtypes of _RealLike have a `__complex__` method
+# (e.g. `builtins.int.__complex__` does not exist)
+class _RealLike(Protocol):
+    def __neg__(self) -> _ComplexLike: ...  # TODO: ideally would be more precise
```
JelleZijlstra, line 44: Why don't you make these two return RealLike?

AlexWaygood: I tried doing this:

```diff
--- a/stdlib/numbers.pyi
+++ b/stdlib/numbers.pyi
@@ -41,8 +41,8 @@ else:
 # as subtypes of _RealLike have a `__complex__` method
 # (e.g. `builtins.int.__complex__` does not exist)
 class _RealLike(Protocol):
-    def __neg__(self) -> _ComplexLike: ...  # TODO: ideally would be more precise
-    def __pos__(self) -> _ComplexLike: ...  # TODO: ideally would be more precise
+    def __neg__(self) -> _RealLike: ...
+    def __pos__(self) -> _RealLike: ...
     def __abs__(self) -> _RealLike: ...
     def __float__(self) -> float: ...
     def __trunc__(self) -> _IntegralLike: ...
@@ -143,6 +143,12 @@ class Real(Complex):
     @property
     def imag(self) -> Literal[0]: ...
     def conjugate(self) -> _RealLike: ...  # type: ignore[override]
+    # Not actually overridden at runtime,
+    # but we override these in the stub to give them more precise return types:
+    @abstractmethod
+    def __pos__(self) -> _RealLike: ...
+    @abstractmethod
+    def __neg__(self) -> _RealLike: ...
```

But on 3.12, mypy complains:

```
stdlib\numbers.pyi:149: error: Return type "_RealLike" of "__pos__" incompatible with return type "_SupportsComplex" in supertype "Complex"  [override]
stdlib\numbers.pyi:151: error: Return type "_RealLike" of "__neg__" incompatible with return type "_SupportsComplex" in supertype "Complex"  [override]
```

And on 3.8, mypy complains with:

```
stdlib\numbers.pyi:149: error: Return type "_RealLike" of "__pos__" incompatible with return type "Union[_SupportsComplex, complex]" in supertype "Complex"  [override]
stdlib\numbers.pyi:151: error: Return type "_RealLike" of "__neg__" incompatible with return type "Union[_SupportsComplex, complex]" in supertype "Complex"  [override]
```

I think this is because `_RealLike` does not have a `__complex__` method, so isn't seen as a subtype of `_SupportsComplex`. (It _can't_ have a `__complex__` method, because `int` doesn't have a `__complex__` method, and we need `int` to be understood as a subtype of `RealLike`.)

This is yet another way in which these ABCs are subtly broken, unfortunately. Another approach might be to return `SupportsAbs[_RealLike] | complex` (instead of `_SupportsComplex | complex`) from most of the `numbers.Complex` methods.

AlexWaygood:
Closing in favour of  #11375

PR #11374, opened 2024-02-07
Add stubs for pygit2

xen0n:
The upstream library is very tricky to type (likely requires nontrivial refactoring), and only contains partial type information, but stubs are a lot easier because only the public signatures are involved this way, so I plan to first make the library usable in typed projects by making stubs available here, then gradually work my way upstream.

The stubs are auto-generated then completed with fully manual inspection of every Python source file. The `_pygit2.pyi` comes from upstream and is mostly untouched except for required style changes, the signature of `options()`, and `FilterSource` which is missing from upstream.


stubs/pygit2/pygit2/__init__.pyi
```diff
@@ -0,0 +1,423 @@
+from _typeshed import AnyStr_co, SupportsAllComparisons
+from collections.abc import Callable
+from os import PathLike
+
+from . import enums
+from ._build import __version__ as __version__
+from ._pygit2 import (
```
JelleZijlstra, line 7: The runtime just does `from ._pygit2 import *`, why not do that here?

stubs/pygit2/pygit2/__init__.pyi
```diff
@@ -0,0 +1,423 @@
...
+from .blob import BlobIO as BlobIO
+from .callbacks import (
+    CheckoutCallbacks as CheckoutCallbacks,
+    Payload as Payload,
+    RemoteCallbacks,
+    StashApplyCallbacks as StashApplyCallbacks,
+    get_credentials as get_credentials,
+)
+from .config import Config as Config
+from .credentials import (
```
JelleZijlstra, line 317: This one is also `import *`ed.

stubs/pygit2/pygit2/__init__.pyi
```diff
@@ -0,0 +1,423 @@
...
+    Keypair as Keypair,
+    KeypairFromAgent as KeypairFromAgent,
+    KeypairFromMemory as KeypairFromMemory,
+    Username as Username,
+    UserPass as UserPass,
+)
+from .errors import Passthrough as Passthrough
+from .filter import Filter as Filter
+from .index import Index as Index, IndexEntry as IndexEntry
+from .legacyenums import (
```
JelleZijlstra, line 327: And this one

stubs/pygit2/pygit2/__init__.pyi
```diff
@@ -0,0 +1,423 @@
...
+from .remotes import Remote as Remote
+from .repository import Repository
+from .settings import Settings
+from .submodules import Submodule as Submodule
+
+features: enums.Feature
+LIBGIT2_VER: tuple[int, int, int]
+
+def init_repository(
+    path: str | bytes | PathLike[AnyStr_co] | None,
```
JelleZijlstra, line 401: A TypeVar doesn't make sense here since there is only one occurrence in the function. In any case, covariance is meaningless in a function.

You can instead use the `StrOrBytesPath` from `_typeshed`.

stubs/pygit2/pygit2/_build.pyi
```diff
@@ -0,0 +1,6 @@
+from pathlib import Path
+from typing import Any
+
+__version__: str
+
+def get_libgit2_paths() -> tuple[Path, dict[str, Any]]: ...
```
JelleZijlstra, line 6:
```suggestion
def get_libgit2_paths() -> tuple[Path, dict[str, list[str]]]: ...
```

stubs/pygit2/pygit2/_pygit2.pyi
```diff
@@ -0,0 +1,773 @@
...
+    ObjectType,
+    Option,
+    ReferenceFilter,
+    ReferenceType,
+    ResetMode,
+    SortMode,
+)
+from .filter import Filter
+
+GIT_APPLY_LOCATION_BOTH: int
```
JelleZijlstra, line 27: Optionally, you can set the actual values here. (To be clear, this isn't required for merging, especially as there is a ginormous number of constants in this library.)

```suggestion
GIT_APPLY_LOCATION_BOTH: Final = 2
```

stubs/pygit2/pygit2/_pygit2.pyi
```diff
@@ -0,0 +1,773 @@
...
+class Blob(Object):
+    data: bytes
+    is_binary: bool
+    size: int
+    def diff(self, blob: Blob = ..., flag: int = ..., old_as_path: str = ..., new_as_path: str = ...) -> Patch: ...
+    def diff_to_buffer(
+        self, buffer: bytes | None = None, flag: DiffOption = ..., old_as_path: str = ..., buffer_as_path: str = ...
+    ) -> Patch: ...
+
+# This is not a real subclassing. Just ensuring type-checkers sees this type as compatible with _CDataBase
```
JelleZijlstra, line 360: I don't understand this comment, how is `_CDataBase` involved?

stubs/pygit2/pygit2/blame.pyi
```diff
@@ -0,0 +1,34 @@
...
+    @property
+    def final_commit_id(self) -> Oid: ...
+    @property
+    def orig_start_line_number(self) -> int: ...
+    @property
+    def orig_committer(self) -> Signature: ...
+    @property
+    def orig_commit_id(self) -> Oid: ...
+    @property
+    def orig_path(self) -> str: ...
```
JelleZijlstra, line 27:
```suggestion
    def orig_path(self) -> str | None: ...
```
The code has a path that returns None.

stubs/pygit2/pygit2/blob.pyi
```diff
@@ -0,0 +1,25 @@
...
+
+from ._pygit2 import Blob, Oid
+from .enums import BlobFilter
+
+class _BlobIO(io.RawIOBase):
+    def __init__(self, blob: Blob, as_path: str | None = None, flags: BlobFilter = ..., commit_id: Oid | None = None) -> None: ...
+    def __exit__(
+        self, exc_type: type[BaseException] | None, exc_value: BaseException | None, traceback: types.TracebackType | None
+    ) -> None: ...
+    def isatty() -> bool: ...  # type: ignore[misc]
```
JelleZijlstra, line 14: Seems like a runtime bug

stubs/pygit2/pygit2/branches.pyi
```diff
@@ -0,0 +1,17 @@
...
+from ._pygit2 import Branch, Commit, Oid
+from .enums import BranchType
+from .repository import BaseRepository
+
+class Branches:
+    local: Branches
+    remote: Branches
+    def __init__(self, repository: BaseRepository, flag: BranchType = ..., commit: Commit | Oid | None = None) -> None: ...
+    def __getitem__(self, name: str) -> Branch: ...
+    def get(self, key: str) -> Branch: ...
```
JelleZijlstra, line 12:
```suggestion
    def get(self, key: str) -> Branch | None: ...
```

stubs/pygit2/pygit2/callbacks.pyi
```diff
@@ -0,0 +1,79 @@
...
+
+from _cffi_backend import _CDataBase
+
+from ._pygit2 import DiffFile, Oid
+from .enums import CheckoutNotify, CheckoutStrategy, CredentialType, StashApplyProgress
+from .remotes import TransferProgress
+from .utils import _IntoStrArray
+
+class Payload:
+    def __init__(self, **kw: dict[str, object]) -> None: ...
```
JelleZijlstra, line 16:
```suggestion
    def __init__(self, **kw: object) -> None: ...
```
Note that the type for `**kwargs` is the type of individual kwargs, not of the whole kwargs dict.

stubs/pygit2/pygit2/callbacks.pyi
```diff
@@ -0,0 +1,79 @@
...
+def libgit2_callback_void(f: Callable[_P, None]) -> Callable[_P, None]: ...
+
+_CredentialsFn: TypeAlias = Callable[[str | None, str | None, CredentialType], _Credentials]
+
+def get_credentials(fn: _CredentialsFn, url: _CDataBase, username: _CDataBase, allowed: CredentialType) -> _CDataBase: ...
+def git_checkout_options(
+    callbacks: CheckoutCallbacks | None = None,
+    strategy: CheckoutStrategy | None = None,
+    directory: PathLike[AnyStr_co] | bytes | str | None = None,
+    paths: _IntoStrArray[AnyStr_co] = None,
```
JelleZijlstra, line 71: It doesn't look like this function really needs to be generic. These types imply that `directory` and `paths` must be of the same type, which doesn't seem to be the case.

stubs/pygit2/pygit2/callbacks.pyi
```diff
@@ -0,0 +1,79 @@
...
+) -> AbstractContextManager[Payload]: ...
+def git_stash_apply_options(
+    callbacks: StashApplyCallbacks | None = None,
+    reinstate_index: bool = False,
+    strategy: CheckoutStrategy | None = None,
+    directory: PathLike[AnyStr_co] | bytes | str | None = None,
```
JelleZijlstra, line 77: Same here

stubs/pygit2/pygit2/config.pyi
```diff
@@ -0,0 +1,60 @@
+from _typeshed import AnyStr_co
+from os import PathLike
+from typing_extensions import Self
+
+from _cffi_backend import _CDataBase
+
+def str_to_bytes(value: object, name: str) -> bytes: ...
```
JelleZijlstra, line 7:
```suggestion
def str_to_bytes(value: str, name: object) -> bytes: ...
```
It throws a TypeError if it's not a str

stubs/pygit2/pygit2/config.pyi
```diff
@@ -0,0 +1,60 @@
...
+    def __next__(self) -> str: ...  # type: ignore[override]
+
+class Config:
+    def __init__(self, path: str | None = None) -> None: ...
+    @classmethod
+    def from_c(cls, repo: _CDataBase, ptr: _CDataBase) -> Config: ...
+    def __del__(self) -> None: ...
+    def __contains__(self, key: str) -> bool: ...
+    def __getitem__(self, key: str) -> str: ...
+    def __setitem__(self, key: str, value: bool | int | _CDataBase | PathLike[AnyStr_co] | bytes | str | None) -> None: ...
```
JelleZijlstra, line 26: Again, don't think AnyStr_co makes sense here.

JelleZijlstra (commented):
Thank you. I looked through the code up to `config.pyi`, lightly comparing the stubs against the implementation. Mostly it looks good, but I see quite a few incorrect uses of `AnyStr_co`.

stubs/pygit2/pygit2/__init__.pyi
xen0n: Originally I kept the star imports, but then mypy started to complain about wrong types of `settings` (it seems unable to distinguish the module `settings` and the global variable `settings` for example), and stubtest also wasn't happy with all those re-exported names either, so I had to spell out the names to basically help the checkers properly resolve the names.

stubs/pygit2/pygit2/__init__.pyi
xen0n: Thanks for the suggestion; I only recently returned to Python development so I'm still getting familiar with the new typing scheme. I'll push an updated version later.

stubs/pygit2/pygit2/blob.pyi
xen0n: Yes this is an actual bug in the upstream library, and fixing it would be another separate issue to file on their repo.

stubs/pygit2/pygit2/config.pyi
xen0n: That must be a thinko. Fixed in the next push.

stubs/pygit2/pygit2/_pygit2.pyi
xen0n: This is mostly copied from `stubs/JACK-Client/jack/__init__.pyi`, as the type checker apparently has some issues with the CFFI types. What's the current best practice of typing CFFI bindings though?

xen0n:
I've just removed all `AnyStr_co` occurrences and fixed the rest of signature mistakes pointed out in the review comments.

xen0n:
It seems simply ignoring the type check error on the `Repository` re-import doesn't work. When I was working on a project with the content of this PR integrated into its `stubs/` I see Pylance failing to infer type for `pygit2.Repository` (saying the name doesn't exist).

JelleZijlstra (approved):

PR #11376, opened 2024-02-07
Pillow: Allow tuple[int, int] for cutoff arg of ImageOps.autocontrast

pa-tna:


radarhere:
I've recreated this as https://github.com/python/typeshed/pull/11408

PR #11377, opened 2024-02-07
Update type of Formatter.convert_field (missing `|None`)

Carreau:
From the source it expects that the second parameter (conversion), can be None:

```python
   def convert_field(self, value, conversion):
        # do any conversion on the resulting object
        if conversion is None:
            return value
   ...
```

Since at least 17 years according to git blame, as None is refered as early as 11529195cae2438a3ac003babcb1b11af67c4037
And still present in main branch:

https://github.com/python/cpython/blob/ef3ceab09d2d0959c343c662461123d5b0e0b64b/Lib/string.py


AlexWaygood (approved):

Carreau:
Thanks !

hauntsaninja:
Thank *you*! :-)

PR #11386, opened 2024-02-09
`tensorflow`: add `tf.linalg` module

hoel-bagard:
Add incomplete stubs taken from [here](https://github.com/hmc-cs-mdrissi/tensorflow_stubs/blob/main/stubs/tensorflow/linalg.pyi).

### DIFF
--- a/stubs/tensorflow/tensorflow/linalg.pyi
+++ b/stubs/tensorflow/tensorflow/linalg.pyi
@@ -42,7 +42,7 @@ def set_diag(
 def eye(
     num_rows: ScalarTensorCompatible,
     num_columns: ScalarTensorCompatible | None = None,
-    batch_shape: int | tuple[int, ...] | None = None,
+    batch_shape: int | list[int] | tuple[int, ...] | None = None,
     dtype: DTypeLike = ...,
     name: str | None = None,
 ) -> Tensor: ...
--- a/stubs/tensorflow/tensorflow/linalg.pyi
+++ b/stubs/tensorflow/tensorflow/linalg.pyi
@@ -1,0 +1,50 @@
+from _typeshed import Incomplete
+from builtins import bool as _bool
+from typing import Literal, overload
+
+from tensorflow import RaggedTensor, Tensor, norm as norm
+from tensorflow._aliases import DTypeLike, ScalarTensorCompatible, TensorCompatible
+from tensorflow.math import l2_normalize as l2_normalize
+
+@overload
+def matmul(
+    a: TensorCompatible,
+    b: TensorCompatible,
+    transpose_a: _bool = False,
+    transpose_b: _bool = False,
+    adjoint_a: _bool = False,
+    adjoint_b: _bool = False,
+    a_is_sparse: _bool = False,
+    b_is_sparse: _bool = False,
+    output_type: DTypeLike | None = None,
+    name: str | None = None,
+) -> Tensor: ...
+@overload
+def matmul(
+    a: RaggedTensor,
+    b: RaggedTensor,
+    transpose_a: _bool = False,
+    transpose_b: _bool = False,
+    adjoint_a: _bool = False,
+    adjoint_b: _bool = False,
+    a_is_sparse: _bool = False,
+    b_is_sparse: _bool = False,
+    output_type: DTypeLike | None = None,
+    name: str | None = None,
+) -> RaggedTensor: ...
+def set_diag(
+    input: TensorCompatible,
+    diagonal: TensorCompatible,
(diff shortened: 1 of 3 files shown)


stubs/tensorflow/tensorflow/linalg.pyi
```diff
@@ -0,0 +1,50 @@
...
+    input: TensorCompatible,
+    diagonal: TensorCompatible,
+    name: str | None = "set_diag",
+    k: int = 0,
+    align: Literal["RIGHT_LEFT", "RIGHT_RIGHT", "LEFT_LEFT", "LEFT_RIGHT"] = "RIGHT_LEFT",
+) -> Tensor: ...
+def eye(
+    num_rows: ScalarTensorCompatible,
+    num_columns: ScalarTensorCompatible | None = None,
+    batch_shape: int | tuple[int, ...] | None = None,
```
JelleZijlstra, line 45:
```suggestion
    batch_shape: int | list[int] | tuple[int, ...] | None = None,
```
According to the docstring

stubs/tensorflow/tensorflow/linalg.pyi
```diff
@@ -0,0 +1,50 @@
...
+    input: TensorCompatible,
+    diagonal: TensorCompatible,
+    name: str | None = "set_diag",
+    k: int = 0,
+    align: Literal["RIGHT_LEFT", "RIGHT_RIGHT", "LEFT_LEFT", "LEFT_RIGHT"] = "RIGHT_LEFT",
+) -> Tensor: ...
+def eye(
+    num_rows: ScalarTensorCompatible,
+    num_columns: ScalarTensorCompatible | None = None,
+    batch_shape: int | list[int] | tuple[int, ...] | None = None,
```
hmc-cs-mdrissi, line 45: Minor nit: I'd lean to relax to TensorCompatible. batch_shape is also compatible with tensor, `A list or tuple of Python integers or a 1-D int32 Tensor`. Looking over implementation I think `Iterable[int] | TensorCompatible` is about as broad as it can be. It's very rare for list[int] to be compatible but not int tensor.

stubs/tensorflow/tensorflow/linalg.pyi
```diff
@@ -0,0 +1,50 @@
...
+    dtype: DTypeLike = ...,
+    name: str | None = None,
+) -> Tensor: ...
+def band_part(input: TensorCompatible, num_lower: int, num_upper: int, name: str | None = None) -> Tensor: ...
```
hmc-cs-mdrissi, line 49: A scalar (0d like `tf.constant(10)`) int tensor will also work for num_lower/num_upper. This is generally true of most tensor functions that take an int, although I'm unsure how common it is and higher rank tensors will fail. I'd weakly lean to prefer ScalarTensorCompatible over int here and in similar cases.

hmc-cs-mdrissi (approved):
Only two small comments on broadening two hints. My stubs have tendency for tensor functions to just have size: int instead of size: ScalarTensorCompatible. Usually functions in linalg/tf.math/tf.nn if it expects a scalar int will accept tensor scalar too. The former choice of just int will have some false positives, while latter ScalarTensorCompatible will have some false negatives as we don't distinguish rank of tensor (scalar vs vector vs matrix). I'd lean for typeshed to prefer latter.

stubs/tensorflow/tensorflow/linalg.pyi
hoel-bagard: `TensorCompatible` includes invalid types, like `str`, `float` and array of strings/float. I feel like `TensorCompatible` is so broad that it is rarely the best match.

`Iterable[int] | IntArray | tf.Tensor` seems better to me, so I went with that in 6e2d07a41.

stubs/tensorflow/tensorflow/linalg.pyi
hoel-bagard: I also noticed that in #11333 and there I added an `Integer` type for that purpose. It seems common enough in TensorFlow to warrant an alias.

```python
Integer: TypeAlias = tf.Tensor | int | IntArray | np.number[Any]
```

hoel-bagard: I've done the same here in ece768045.

hoel-bagard:
> I'd lean for typeshed to prefer latter.

Same here. While false negatives aren't great, they probably won't occur as much as the false positives would. And they will probably quickly be found out and fixed since they cause errors at runtime. I wish the third-party stubtests would return more examples, that would help find edge cases and evaluate the impact.

PR #11387, opened 2024-02-09
`tensorflow`: Add some `tf.raw_ops` members.

hoel-bagard:
Stubs taken from [here](https://github.com/hmc-cs-mdrissi/tensorflow_stubs/blob/main/stubs/tensorflow/raw_ops.pyi).

### DIFF
--- a/stubs/tensorflow/tensorflow/raw_ops.pyi
+++ b/stubs/tensorflow/tensorflow/raw_ops.pyi
@@ -1,7 +1,7 @@
 from _typeshed import Incomplete
 from typing import Literal
 
-from tensorflow import Tensor
+from tensorflow import Operation, Tensor
 from tensorflow._aliases import TensorCompatible
 
 def Fingerprint(data: TensorCompatible, method: Literal["farmhash64"], name: str | None = None) -> Tensor: ...
@@ -15,7 +15,7 @@ def ResourceApplyAdagradV2(
     use_locking: bool = False,
     update_slots: bool = True,
     name: str | None = None,
-) -> None: ...
+) -> Operation: ...
 def ResourceSparseApplyAdagradV2(
     var: Tensor,
     accum: Tensor,
@@ -26,7 +26,7 @@ def ResourceSparseApplyAdagradV2(
     use_locking: bool = False,
     update_slots: bool = True,
     name: str | None = None,
-) -> None: ...
+) -> Operation: ...
 def ResourceApplyAdam(
     var: Tensor,
     m: Tensor,
@@ -41,5 +41,5 @@ def ResourceApplyAdam(
     use_locking: bool = False,
     use_nesterov: bool = False,
     name: str | None = None,
-) -> None: ...
+) -> Operation: ...
 def __getattr__(name: str) -> Incomplete: ...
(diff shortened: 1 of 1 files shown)


stubs/tensorflow/tensorflow/raw_ops.pyi
```diff
@@ -0,0 +1,45 @@
...
+def ResourceApplyAdagradV2(
+    var: Tensor,
+    accum: Tensor,
+    lr: TensorCompatible,
+    epsilon: TensorCompatible,
+    grad: TensorCompatible,
+    use_locking: bool = False,
+    update_slots: bool = True,
+    name: str | None = None,
+) -> None: ...
```
JelleZijlstra, line 18: This returns "The created Operation" according to its docstring.

stubs/tensorflow/tensorflow/raw_ops.pyi
```diff
@@ -0,0 +1,45 @@
...
+def ResourceSparseApplyAdagradV2(
+    var: Tensor,
+    accum: Tensor,
+    lr: TensorCompatible,
+    epsilon: TensorCompatible,
```
JelleZijlstra, line 23: These two must be scalars, so perhaps we can give narrower type (optional)

stubs/tensorflow/tensorflow/raw_ops.pyi
```diff
@@ -0,0 +1,45 @@
...
+    grad: TensorCompatible,
+    indices: TensorCompatible,
+    use_locking: bool = False,
+    update_slots: bool = True,
+    name: str | None = None,
+) -> None: ...
```
JelleZijlstra, line 29: Also returns an Operation

stubs/tensorflow/tensorflow/raw_ops.pyi
```diff
@@ -0,0 +1,45 @@
...
+    beta2_power: TensorCompatible,
+    lr: TensorCompatible,
+    beta1: TensorCompatible,
+    beta2: TensorCompatible,
+    epsilon: TensorCompatible,
+    grad: TensorCompatible,
+    use_locking: bool = False,
+    use_nesterov: bool = False,
+    name: str | None = None,
+) -> None: ...
```
JelleZijlstra, line 44: This one too

hoel-bagard: Fixed in b392c8a8e.

stubs/tensorflow/tensorflow/raw_ops.pyi
hoel-bagard: Fixed in b392c8a8e.

stubs/tensorflow/tensorflow/raw_ops.pyi
hoel-bagard: Fixed in b392c8a8e.

stubs/tensorflow/tensorflow/raw_ops.pyi
hoel-bagard: I don't have code to run this function, so I can't check, but very often the TensorFlow doc says that the input must be a scalar, but anything that can be converted to a scalar will do. So for example it will say that a `float32` is required but `"1.2"` would work. However that behavior isn't consistent, so we can't know without testing.

For example:
```python
import tensorflow as tf
tf.keras.optimizers.Adam(learning_rate="1.0")   # Raises ValueError
tf.keras.optimizers.Adam(epsilon="1.0")   # Works
```

So far I think most of the typing for TensorFlow is using overly lax typing for the input in order to avoid false positives, so I would leave `TensorCompatible`.

PR #11388, opened 2024-02-09
`tensorflow`: add partial `tf.nn` module.

hoel-bagard:
Stacked on top of https://github.com/python/typeshed/pull/11358

Some of the stubs are taken from [here](https://github.com/hmc-cs-mdrissi/tensorflow_stubs/blob/main/stubs/tensorflow/nn.pyi).


stubs/tensorflow/tensorflow/nn.pyi
```diff
@@ -0,0 +1,195 @@
...
+@overload
+def embedding_lookup(
+    params: TensorCompatible, ids: RaggedTensor, max_norm: float | None = None, name: str | None = None
+) -> RaggedTensor: ...
+def leaky_relu(features: TensorCompatible, alpha: float = 0.2, name: str | None = None) -> Tensor: ...
+def log_poisson_loss(
+    targets: TensorCompatible, log_input: TensorCompatible, compute_full_loss: bool = False, name: str | None = None
+) -> Tensor: ...
+
+# tf.nn.moments's shit is not used in the current implementation.
```
JelleZijlstra, line 173:
```suggestion
# tf.nn.moments's shift is not used in the current implementation.
```
Not sure if this is profanity or a typo. In either case, let's keep it out of the repo :)

Issue #10947, opened 2023-10-29
Default values in tkinter stubs


Akuli:
Tkinter is very dynamic. Most keyword arguments are passed to Tcl, and their behavior is described best in manual pages written for Tcl users. This means that when we ran stubdefaulter to figure out default values for most functions in typeshed, it didn't do much for tkinter, and most of tkinter's default values are still `= ...`. 

I (or someone else) should go through the `= ...` occurrences and fill in whatever can be filled. Often the default values are platform-specific, and hard-coding something into the stubs doesn't make sense.

Most defaults in tkinter are widget options (`__init__` and `configure()` methods). To see their default values for a given platform, you can instantiate the widget and convert it to a dict. For example:

```
>>> import pprint
>>> import tkinter
>>> pprint.pprint(dict(tkinter.Text()))
{'autoseparators': 1,
 'background': '#3b3b3e',
 'bd': 1,
 'bg': '#3b3b3e',
 'blockcursor': 0,
 'borderwidth': 1,
 'cursor': 'xterm',
 'endline': '',
 'exportselection': 1,
 'fg': '#ffffff',
 'font': 'TkFixedFont',
 'foreground': '#ffffff',
 'height': 24,
 'highlightbackground': '#3b3b3e',
 'highlightcolor': '#ffffff',
 'highlightthickness': 1,
 'inactiveselectbackground': '#c3c3c3',
 'insertbackground': '#000000',
 'insertborderwidth': 0,
 'insertofftime': 300,
 'insertontime': 600,
 'insertunfocussed': 'none',
 'insertwidth': 2,
 'maxundo': 0,
 'padx': 1,
 'pady': 1,
 'relief': 'sunken',
 'selectbackground': '#7d7d82',
 'selectborderwidth': <pixel object: '0'>,
 'selectforeground': '#ffffff',
 'setgrid': 0,
 'spacing1': 0,
 'spacing2': 0,
 'spacing3': 0,
 'startline': '',
 'state': 'normal',
 'tabs': '',
 'tabstyle': 'tabular',
 'takefocus': '',
 'undo': 0,
 'width': 80,
 'wrap': 'char',
 'xscrollcommand': '',
 'yscrollcommand': ''}
```

TheCodingAyush:
Hi, I’m Ayush, a 2nd year CS student. 
I would like to work on adding default values for Tkinter __init__ methods. 
Could you please assign this issue to me? 
Thank you!

Akuli:
[...]

PR #4363, opened 2020-07-25
Add options to tkinter widgets

Akuli:
Tkinter options now have type hints that describe which values are valid when setting the options. For example, `tkinter.Label(relief='sunkenn')` now creates a mypy error. The correct types for the options come from Tk's manual pages as described in `__init__.pyi`. This affects `__init__`, `config`, `configure`, `cget`, `__setitem__` and `__getitem__` methods because those are the methods that deal with widget options.

Looking up the value of an option always returns `Any`, because tkinter doesn't make much effort to ensure that the return values have consistent type. In practice, they can be pretty much anything in funny corner cases, and figuring out which of the possible types actually come up frequently enough for type hints would be a lot of work.

In tkinter, `widget['foo'] = bar` and `widget.config(foo=bar)` do the same thing, but they type-check differently: the `widget['foo'] = bar` syntax allows `'foo'` to be any string (e.g. a variable, not necessarily a `Literal`) and `bar` to be any object, while `widget.config(foo=bar)` checks the existence of the option and the type of `bar`. Similarly, `cget` takes a `Literal` argument but `__getitem__` takes a string. This is done because:
- It halves the amount of required copy/pasta in stubs files.
- The type of a variable defined with `foo = 'sunken'` is `str` rather than `Literal['sunken']`, so fixing code that does `widget.config(relief=foo)` can be done without a `# type: ignore` comment by changing it to `widget['relief'] = foo`.
- Similarly, setting `bar = 'relief'` and doing `widget[bar] = 'sunken'` works.

~This also includes a script that checks whether the types of widget options actually work at runtime (stubtest doesn't work well with the generous `**kwargs` usage in tkinter) and the travis configuration to make that part of the CI build. Usually it takes about 22 seconds to run, which is a lot quicker than any of the other CI things. The script creates tkinter widgets, parses a small subset of type hint syntax with `ast`, creates possible values of the options based on those type hints and tries to set the options of the widgets.~

~Some classes in `tkinter/ttk.py` inherit from other classes and break the Liskov substitution principle by not having all the widget options of the base class. In those cases, I made the stubs not inherit and instead used `method = BaseClass.method`.~

~I added a `--verbose` option to `tests/pytype_test.py` for finding pytype issues more easily, but that's disabled in the travis config.~

[...]

PR #11391, opened 2024-02-09
Add defaults for tkinter `__init__` methods

MarcellPerger1:
Add defaults for `tkinter`'s `__init__` methods, see #10947 

The defaults I filled in are consistent across:
- My local Windows 11 machine
- <details><summary>All available github actions runners: </summary>
  <ul>
  <li> <code>ubuntu-20.04</code>, <code>ubuntu-22.04</code> </li>
  <li> <code>windows-2019</code>, <code>windows-2022</code> </li>
  <li> <code>macos-11</code>, <code>macos-12</code>, <code>macos-13</code> </li>
  </ul></details>
I used a script to generate the defaults into a JSON file (source code at https://github.com/MarcellPerger1/typeshed-tkinter-defaults), then I filled them into the arguments by hand (I also checked them with the tcl.tk documentation to ensure the defaults make sense).

I tried to fill the defaults in a way that keeps the defaults value consistent with the argument type (while still keeping it close to the types returned by `.cget`).

Some notes about specific arguments:
- `variable` and `textvariable`: The default value (using `dict`) is consistently `""` but this doesn't match the annotated `Variable` type so I didn't fill these in
- `Message(padx, pady)`: The default value is `-1` which is not allowed [according to the documentation](https://tcl.tk/man/tcl8.6/TkCmd/options.htm#M-padx#:~:text=Command%2DLine%20Name%3A%20%2Dpadx,Specifies%20a%20non%2Dnegative%20value) so I didn't fill this in
- `Toplevel(use)`: The default value is always `""` but this doesn't match the annotated `int` type so I didn't fill this in

P.S. Sorry for the huge diff.


stdlib/tkinter/__init__.pyi
```diff
@@ -949,28 +949,28 @@ class Toplevel(BaseWidget, Wm):
...
+        highlightthickness: _ScreenUnits = 0,
         menu: Menu = ...,
         name: str = ...,
-        padx: _ScreenUnits = ...,
-        pady: _ScreenUnits = ...,
-        relief: _Relief = ...,
-        screen: str = ...,  # can't be changed after creating widget
-        takefocus: _TakeFocusValue = ...,
+        padx: _ScreenUnits = "0",
+        pady: _ScreenUnits = "0",
```
Akuli, line 967: These appear to be `_tkinter.Tcl_Obj` instances by default, so the string `"0"` is no more accurate than an integer `0`:

```
>>> top['padx']
<pixel object: '0'>
>>> top['pady']
<pixel object: '0'>
```

IMO we should just use the integer `0` when there is no padding or border or similar, as `0`, `"0"` and `<pixel object: '0'>` are equivalent anyway. An integer is probably the most useful thing for IDE autocompletions and such.

Akuli (commented):
Thanks for doing this! For now I went through `Toplevel` and `Button` options manually.

Akuli:
Are you going to make a separate PR that does this for `tkinter.ttk` widgets?

I mostly use the ttk widgets in my projects, because I can apply [an awesome theme](https://github.com/rdbende/Sun-Valley-ttk-theme/) to them. I think a separate PR for ttk widgets would be easier to review than adding their defaults to this PR. This way, we won't get confused about `tkinter.Button` vs `tkinter.ttk.Button`, for example.

> P.S. Sorry for the huge diff.

This is small compared to [my biggest autogenerated tkinter PR](https://github.com/python/typeshed/pull/4363) :)

stdlib/tkinter/__init__.pyi
MarcellPerger1: I agree that an integer is probably the most useful.

(Although on Windows, `padx` and `pady` are sometimes `<string object: '0'>` and sometimes just `'0'` - this inconsistency is due to python/cpython#101830):
```
>>> import tkinter
>>> root = tkinter.Tk()
>>> top = tkinter.Toplevel(root)
>>> top['padx']
<string object: '0'>
>>> top['padx']
'0'
>>> top['pady']
<string object: '0'>
>>> top['pady']
'0'
```

MarcellPerger1:
> Are you going to make a separate PR that does this for `tkinter.ttk` widgets?

Yes

stdlib/tkinter/__init__.pyi
Akuli: > For the rest of the integer-in-strings, the default value is consistently a string on all platforms,
so I'm not sure if I should replace those with a plain integer (this would probably be more useful in IDEs but less accurate)

IMO we should just use ints. If we sometimes use `"0"` and sometimes `0`, it will probably make people wonder what their difference is, even though there is no practical difference.

stdlib/tkinter/__init__.pyi
```diff
@@ -1775,24 +1775,24 @@ class Checkbutton(Widget):
...
-        tristateimage: _ImageSpec = ...,
-        tristatevalue: Any = ...,
-        underline: int = ...,
-        variable: Variable | Literal[""] = ...,
-        width: _ScreenUnits = ...,
-        wraplength: _ScreenUnits = ...,
+        tristateimage: _ImageSpec = "",
+        tristatevalue: Any = "",
+        underline: int = -1,
+        variable: Variable | Literal[""] = "",
```
Akuli, line 1793: This doesn't seem correct. By default, there is a dummy variable with the same name as the widget itself:

```
>>> import tkinter
>>> tkinter.Checkbutton()['variable']
<parsedVarName object: '!checkbutton1'>
```

(I spotted this with a quick and dirty script that reads the diff of this PR, instantiates widgets, and grabs their default values.)

stdlib/tkinter/__init__.pyi
```diff
@@ -3504,32 +3504,32 @@ class LabelFrame(Widget):
         cnf: dict[str, Any] | None = {},
         *,
         background: str = ...,
-        bd: _ScreenUnits = ...,
+        bd: _ScreenUnits = 2,
         bg: str = ...,
         border: _ScreenUnits = ...,
```
Akuli, line 3509: `border` is an alias for `borderwidth`, so it should have the same default.

The way short names work is a bit surprising. Anything like `borderwi` would be a valid alias for `borderwidth`. I added `border` to the stubs because I have some code that passes it with that name, probably because I copied it from somewhere.

Akuli (approved):
Thanks! In separate PRs, we still need to add:
- the same defaults for `configure()` methods
- defaults for ttk widgets.

Akuli:
>the same defaults for `configure()` methods

This is not actually quite right; the default in `.configure()` is simply the current value of the option, and there's no better way than `...` to express that. So the status quo is good, and nothing needs to be done.

PR #4765, opened 2020-11-13
Change some tkinter function parameter types from bool to int.

rchen152:
I'm in the process of pulling a newer version of typeshed into Google
and seeing a lot of new errors due to code that passes 0 or 1 into
functions that are annotated as expecting bools.

Based on the fact that tk.{YES,NO,TRUE,FALSE} are defined as ints
(https://github.com/python/typeshed/blob/20a847218ac085ad691ea8d99155ee8ac7005479/stdlib/3/tkinter/constants.pyi#L3-L6)
and that the usage example in the source code also uses an int as a bool
(https://github.com/python/cpython/blob/cc75ab791dd5ae2cb9f6e0c3c5f734a6ae1eb2a9/Lib/tkinter/__init__.py#L25),
I think the annotations should be expanded.

Issue #4669, opened 2020-10-14
tkinter.FALSE doesn't work


Akuli:
In #4658 I saw [this commit](https://github.com/EDCD/EDMarketConnector/commit/958f277fde0ba625bf3b87985c7c906a05b8f3a4), which replaces `tkinter.FALSE` usages with `False`. This is because typeshed currently expects many things to be `bool` even though `tkinter.FALSE` has type `Literal[0]`.  The same applies to `NO`, `YES`, `TRUE`, `ON` and `OFF` defined in `tkinter/constants.py`.

I think the correct fix is to change the types from `Literal[0]` and `Literal[1]` to `bool`. In the past, @srittau has mentioned that it's fine to "cheat" and just use `bool` when the value is known to be `0` or `1`.

srittau:
While I was mostly talking about argument types, I agree that just using `bool` here would be the most pragmatic solution. That said, maybe it would make sense for type checkers to accept `Literal[0]` and `Literal[1]` where `bool` is expected (because `0 == False` and `1 == True`).

PR #11394, opened 2024-02-10
Revert "Change some tkinter function parameter types from bool to int (#4765)"

### LINKED ISSUE #11389: Inconsistent `exportselection` type in tkinter
The types for the `exportselection` argument in `tkinter` methods are inconsistent.  
In [7 out of the 8 occurrences](https://github.com/search?q=repo%3Apython%2Ftypeshed+exportselection+path%3Astdlib%2Ftkinter%2F__init__.pyi&type=code), the type is `bool`, but in the 1 other occurrence (in `Listbox.__init__`) the type is `int`:
https://github.com/python/typeshed/blob/9d90eeb29646f88937bcee38e4dd954a6a769ed2/stdlib/tkinter/__init__.pyi#L2113
It seems that some code passes `0` or `1` instead of a `bool` (see #4765) so it would make sense for all of these types to be `int`. Or `bool | Literal[0, 1]` is probably better.

Akuli:
Reverts #4765, fixes #11389. Typings of `tkinter/constants.pyi` were later changed in #4669.

### DIFF
--- a/stdlib/tkinter/__init__.pyi
+++ b/stdlib/tkinter/__init__.pyi
@@ -184,7 +184,7 @@ _Relief: TypeAlias = Literal["raised", "sunken", "flat", "ridge", "solid", "groo
 _ScreenUnits: TypeAlias = str | float  # Often the right type instead of int. Manual page: Tk_GetPixels
 # -xscrollcommand and -yscrollcommand in 'options' manual page
 _XYScrollCommand: TypeAlias = str | Callable[[float, float], object]
-_TakeFocusValue: TypeAlias = int | Literal[""] | Callable[[str], bool | None]  # -takefocus in manual page named 'options'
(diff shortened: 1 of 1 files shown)


JelleZijlstra:
cc @rchen152 who made the original PR. Personally I'm not sure here. `bool` makes more semantic sense, but `int` does work and apparently appears in the docs. Maybe allowing `Literal[0, 1]` would be a good middle ground.

I'd support a CPython PR to remove any remaining mention of `int` types in the docs and in defaults.

srittau:
As pointed out in the original PR, there are some boolean constants in `tkinter.constants`, which equal `0` and `1`. Amending Jelle's comment: Maybe we could introduce a `_TkBool` type aliases to `bool | Literal[0, 1]` and document why it's necessary.

rchen152:
+1 for `bool | Literal[0, 1]`. I took a look at the tkinter documentation, and in [this section](https://docs.python.org/3/library/tkinter.html#packer-options), it explicitly says that the `expand` option to pack() can be "Boolean, `0`, or `1`."

PR #11402, opened 2024-02-11
Add annotations for `tkinter.ttk.LabeledScale`

MarcellPerger1:
Add annotations for `tkinter.ttk.LabeledScale`. This [doesn't seem the be documented](https://docs.python.org/3/search.html?q=labeledscale) in the official Python documentation so I just took the types from [the source code](https://github.com/python/cpython/blob/b1043607884d774acabd255ecdcebb159f76a2fb/Lib/tkinter/ttk.py#L1492-L1527).

### DIFF
--- a/stdlib/tkinter/ttk.pyi
+++ b/stdlib/tkinter/ttk.pyi
@@ -334,6 +334,8 @@ class Combobox(Entry):
     def set(self, value: Any) -> None: ...
 
 class Frame(Widget):
+    # This should be kept in sync with tkinter.ttk.LabeledScale.__init__()
+    # (all of these keyword-only arguments are also present there)
     def __init__(
         self,
         master: tkinter.Misc | None = None,
@@ -1160,6 +1162,8 @@ class Treeview(Widget, tkinter.XView, tkinter.YView):
 class LabeledScale(Frame):
     label: Label
     scale: Scale
+    # This should be kept in sync with tkinter.ttk.Frame.__init__()
+    # (all the keyword-only args except compound are from there)
     def __init__(
         self,
         master: tkinter.Misc | None = None,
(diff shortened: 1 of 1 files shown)


stdlib/tkinter/ttk.pyi
```diff
@@ -1158,18 +1158,27 @@ class Treeview(Widget, tkinter.XView, tkinter.YView):
...
         self,
         master: tkinter.Misc | None = None,
         variable: tkinter.IntVar | tkinter.DoubleVar | None = None,
         from_: float = 0,
         to: float = 10,
         *,
+        border: tkinter._ScreenUnits = ...,
+        borderwidth: tkinter._ScreenUnits = ...,
+        class_: str = ...,
         compound: Literal["top", "bottom"] = ...,
```
Akuli, line 1173: While you're here, you could add a default value for `compound`.

stdlib/tkinter/ttk.pyi
```diff
@@ -1158,18 +1158,27 @@ class Treeview(Widget, tkinter.XView, tkinter.YView):
     def tag_has(self, tagname: str, item: str | int) -> bool: ...
 
 class LabeledScale(Frame):
-    label: Incomplete
-    scale: Incomplete
-    # TODO: don't any-type **kw. That goes to Frame.__init__.
```
Akuli, line 1163: Can you add comments, both here and to `Frame.__init__`, that the arguments are same as for `Frame.__init__` and must be kept in sync? I'm worried that this will get outdated when `Frame.__init__` is changed.

Akuli (commented):
Thanks!

Akuli (approved):
Thanks!

PR #11403, opened 2024-02-11
Improve annotations for various set methods

### LINKED ISSUE #7121: set[str].discard(str | None)
Consider the following:

```python3
names: set[str] = {"a", "b"}
name: str | None = None
names.discard(name)
```

It would be nice if this wasn't a type checker error. Unlike `.remove()`, the `.discard()` method does nothing if the item is not in the set, so I don't see a good reason to ban `None` values.

### LINKED ISSUE #9004: Annotations for `set.__sub__` and related methods are inconsistent
According to the documentation `s1 - s2` returns the set of all elements that are in `s1` but not in `s2`. So what's the point of constraining `s2` elements type? I believe the correct annotation for the argument of `set.__sub__` would be `AbstractSet[object]` rather than `AbstractSet[T | None]`.

https://github.com/python/typeshed/blob/cf9bdc2d9861cb61e76b512988d19afe232a958f/stdlib/builtins.pyi#L1100

AlexWaygood:
- Fixes some regressions introduced by https://github.com/python/typeshed/pull/7161 (see https://github.com/python/typeshed/issues/9004#issuecomment-1295312743).
- Fixes #9004
- Fixes #7121
- Fixes https://github.com/python/mypy/issues/13192 (cc. @finite-state-machine).
  - I wasn't initially sure that `set[str | None] - set[None]` came up often enough to be worth special-casing, but the activity on that issue has persuaded me otherwise.

It's a bit unfortunate how many `type: ignore`s we have to introduce here, but I can't see a good way round them. I've added some test cases to make sure everything's working correctly.


AlexWaygood:
Primer analysis:
- `ibis` is just an error message changing slightly
- `werkzeug` would just need to change their annotations on their subclass to match our new annotation for `discard()`
- The three `pydantic` hits are all new false positives caused by making `set.__sub__` stricter.

  (You _could_ argue that the third one is, strictly speaking, a true positive, in that it's working the way we "want" -- mypy is warning that there's a possibility `AbstractSet[int]` might be being subtracted from `AbstractSet[str]`, which would be a no-op. But it doesn't seem like a "helpful" true positive. You'd have to do some more work to narrow the type before doing the subtraction, but it wouldn't really make your code any more type-safe.)

  These seem pretty unfortunate to me, and make me think that the better way of fixing https://github.com/python/typeshed/issues/9004 is to make it so that `__sub__`, `__isub__`, `difference` and `difference_update` all accept `AbstractSet[object]` rather than `AbstractSet[_T | None]`. @Akuli, thoughts on that?
  - https://github.com/pydantic/pydantic/blob/832225b90672c68e2d4067bd0ffecf834d62b48b/pydantic/v1/main.py#L901
  - https://github.com/pydantic/pydantic/blob/832225b90672c68e2d4067bd0ffecf834d62b48b/pydantic/main.py#L1257
  - https://github.com/pydantic/pydantic/blob/832225b90672c68e2d4067bd0ffecf834d62b48b/pydantic/deprecated/copy_internals.py#L222
- The new `jax` hit looks like a false positive, probably due to the fact that `set.__sub__` is now an overloaded method. https://github.com/google/jax/blob/9352a00dab3c465a4e2c651cc6a0fa91e50ae7a2/jax/_src/lax/parallel.py#L512

srittau:
This has conflicts now.

stdlib/builtins.pyi
```diff
@@ -1104,9 +1104,9 @@ class set(MutableSet[_T]):
     def __init__(self, iterable: Iterable[_T], /) -> None: ...
     def add(self, element: _T, /) -> None: ...
     def copy(self) -> set[_T]: ...
-    def difference(self, *s: Iterable[Any]) -> set[_T]: ...
-    def difference_update(self, *s: Iterable[Any]) -> None: ...
-    def discard(self, element: _T, /) -> None: ...
+    def difference(self, *s: Iterable[_T | None]) -> set[_T]: ...
+    def difference_update(self, *s: Iterable[_T | None]) -> None: ...
+    def discard(self, element: _T | None, /) -> None: ...
```
Bruntaz, line 1109: I believe the correct signature for this should be `def discard(self, element: Hashable, /) -> None: ...`.

`set.discard` will only raise an exception if an unhashable type is passed into it. Any other input is valid (just if it doesn't match `_T`, it will do nothing instead of raising like `set.remove()`:

```python
>>> a = {1, 2, 3}
>>> a.discard(1)
>>> a
{2, 3}
>>> a.discard("Hello")
>>> a
{2, 3}
>>> class Foo:
...     def __eq__(self): return True
...
>>> a.discard(Foo())
TypeError: unhashable type: 'Foo'
```

srittau:
While I'm not terribly worried about the new LSP warnings, this PR doesn't seem to fix any actual problems, only introduce new ones – at least not according to primer. I'm -0 on this solution (at least for now).

ferdnyc:
The number of primer changes/additions where the expected type has demonstrably worsened is concerning. 

### schemathesis

Take the 'schemathesis' diff:

#### Before
`error: Argument 1 to "set" has incompatible type "list[schemathesis.cli.commands.run.hypothesis.Phase]"; expected "Iterable[Optional[hypothesis._settings.Phase]]"  [arg-type]`

#### After
`error: Argument 1 to "set" has incompatible type "list[Phase]"; expected "Iterable[None]"  [arg-type]`

Something's just plain wrong, if the expected type is `Iterable[None]`.

### pydantic
Same thing in the first new 'pydantic' message:

`error: Set comprehension has incompatible type Set[int | str]; expected Set[None]  [misc]`

Not sure what kind of set comprehension is expected to contain only `None` values, but that feels less-than-useful in a type-checking context, and presumably has little-to-no relevance to the runtime type situation.

randolf-scholz:
> These seem pretty unfortunate to me, and make me think that the better way of fixing https://github.com/python/typeshed/issues/9004 is to make it so that __sub__, __isub__, difference and difference_update all accept AbstractSet[object] or Iterable[Any] rather than AbstractSet[_T | None]. 

@AlexWaygood  I strongly agree with that and opened a smaller PR #15160 to address exactly this issue. The primer impact is much smaller. 

There are cases where the current behavior really hurts, for example `set[<Literals>] - set[str]` gets flagged as not allowed.

I believe the case @Akuli worried about in https://github.com/python/typeshed/issues/9004#issuecomment-1294962317, `set[bytes] - set[str]`, is a job for a type-informed linter, but not for typeshed. A linter can decide to flag `set[A] - set[B]` as redundant when `A & B = Never`.

srittau:
@AlexWaygood What are your plans with this PR? At the moment there are merge conflicts, since we already merged some of the changes.

PR #11404, opened 2024-02-11
Add annotations and defaults for `tkinter.ttk.OptionMenu`

MarcellPerger1:
Add annotations and defaults for `tkinter.ttk.OptionMenu`.


Akuli (approved):
Thanks :)

PR #11409, opened 2024-02-12
Pillow: Allow float(s) for cutoff param of autocontrast

pa-tna:
This is an extension of https://github.com/python/typeshed/pull/11376 to permit float values (and 2-tuple values containing 1 or 2 floats) as the value of `cutoff` for the `autocontrast` function in the `ImageOps` module.

The previous pull request only permits `int`s, but the documentation doesn't stipulate `int`s (only "number"), and passing `float`s could be desirable if a more specific cutoff is required. Using a `float` type also permits `int` (see [&ldquo;The numeric tower&rdquo; in PEP 484](https://peps.python.org/pep-0484/#the-numeric-tower)).

### DIFF
--- a/stubs/Pillow/PIL/ImageOps.pyi
+++ b/stubs/Pillow/PIL/ImageOps.pyi
@@ -13,7 +13,7 @@ class _Deformer(Protocol):
 
 def autocontrast(
     image: Image,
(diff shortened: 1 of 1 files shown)


srittau (approved):
Thanks!

PR #11411, opened 2024-02-12
Additions to the html5lib stubs to handle type safety warnings in Beautiful Soup's html5lib code.

leonardr:
I'm adding type hints to my project Beautiful Soup, and part of this work required minor changes to the typeshed stubs for html5lib. This PR contributes those changes back to typeshed.

I've also noticed that the typeshed stub for `HTMLParser.parse()` is wrong: it lists the second argument as `scripting:bool`, but the second argument is actually `treebuilder:str`; `scripting` is one of the keyword arguments. But I'm too new to Python type hints to know what should be done with `scripting`, and I also don't know what downstream effects a change might have on other users, so I'm not trying to fix that in this branch.


stubs/html5lib/html5lib/_inputstream.pyi
```diff
@@ -36,7 +36,7 @@ class HTMLUnicodeInputStream:
     prevNumCols: int
     def reset(self) -> None: ...
     def openStream(self, source): ...
-    def position(self): ...
+    def position(self) -> tuple[int, int]: ...
```
leonardr, line 39: [See here.](https://github.com/html5lib/html5lib-python/blob/master/html5lib/_inputstream.py#L229)

stubs/html5lib/html5lib/html5parser.pyi
```diff
@@ -32,6 +32,7 @@ class HTMLParser:
     lastPhase: Any
     beforeRCDataPhase: Any
     framesetOK: bool
+    tokenizer: Any
```
leonardr, line 35: This is set by [HTMLParser._parse](https://github.com/html5lib/html5lib-python/blob/master/html5lib/html5parser.py#L127), not by the constructor, so it's not always guaranteed to be there, but the same is true of `lastPhase`, `beforeRCDataPhase`, `framesetOK`, and so on.

JelleZijlstra (approved):
Thanks! I'll look into the other issue you mentioned in your description.

JelleZijlstra:
> I've also noticed that the typeshed stub for HTMLParser.parse() is wrong: it lists the second argument as scripting:bool, but the second argument is actually treebuilder:str; scripting is one of the keyword arguments.

I think the stub is actually right. The parse method (not to be confused with the separate module-level function named `parse`) is [here](https://github.com/html5lib/html5lib-python/blob/82c2599585a6119e5afd26e58e754972c79f6734/html5lib/html5parser.py#L265): it calls `_parse` with arguments `stream, False, None, *args, **kwargs`. Now `_parse` [takes](https://github.com/html5lib/html5lib-python/blob/82c2599585a6119e5afd26e58e754972c79f6734/html5lib/html5parser.py#L127) arguments `stream, innerHTML=False, container="div", scripting=False, **kwargs`. `innerHTML` and `container` are always False and None, so the `*args` and `**kwargs` accepted by `parse` have to map to this `scripting` arg and the `**kwargs`.

leonardr:
> > I've also noticed that the typeshed stub for HTMLParser.parse() is wrong: it lists the second argument as scripting:bool, but the second argument is actually treebuilder:str; scripting is one of the keyword arguments.
> 
> I think the stub is actually right. The parse method (not to be confused with the separate module-level function named `parse`) is [here](https://github.com/html5lib/html5lib-python/blob/82c2599585a6119e5afd26e58e754972c79f6734/html5lib/html5parser.py#L265): it calls `_parse` with arguments `stream, False, None, *args, **kwargs`. Now `_parse` [takes](https://github.com/html5lib/html5lib-python/blob/82c2599585a6119e5afd26e58e754972c79f6734/html5lib/html5parser.py#L127) arguments `stream, innerHTML=False, container="div", scripting=False, **kwargs`. `innerHTML` and `container` are always False and None, so the `*args` and `**kwargs` accepted by `parse` have to map to this `scripting` arg and the `**kwargs`.

This makes sense, thanks for the explanation.

PR #11413, opened 2024-02-12
Fix return type of `calendar.HTMLCalendar.formatyearpage`

ShineyDev:
This PR changes the return type of `HTMLCalendar.formatyearpage` from str to bytes.

You can find the stub [here](https://github.com/python/typeshed/blob/main/stdlib/calendar.pyi#L125-L127) and the source [here](https://github.com/python/cpython/blame/main/Lib/calendar.py#L575).


AlexWaygood (approved):
Thanks!

JelleZijlstra (approved):
Thanks! Agree that function returns bytes.

PR #11417, opened 2024-02-13
oauthlib: Fix annotations for `oauthlib.oauth2.RequestValidator`

Daverball:
A couple of small issues I found in the oauth2 `RequestValidator`:

- `dict[str, int | str]` for token on `save_token`/`save_bearer_token` does not seem very helpful, either it should be `dict[str, Any]` since technically the oauth server can put anything in the token dictionary it wants, or a typed dictionary with the approprate bearer token keys, `dict[str, int | str]` is the worst of both worlds and at best inaccurate, the docs state:
  >      The Bearer token dict may hold a number of items:
  >         {
  >             'token_type': 'Bearer',
  >             'access_token': 'askfjh234as9sd8',
  >             'expires_in': 3600,
  >             'scope': 'string of space separated authorized scopes',
  >             'refresh_token': '23sdf876234',  # if issued
  >             'state': 'given_by_client',  # if supplied by client (implicit ONLY)
  >        }

- `str` return type on `save_bearer_token` is consistent with the docs but then `save_token` shouldn't return `None`, nothing in the code actually uses the return value of `save_token` or `save_bearer_token`, so it seems better to make it return `object` and let the actual implementation decide whether it should return something or not.

- `code` on `save_authorization_code` is a dictionary and not a `str`, the docs state:
  > The ``code`` argument is actually a dictionary  containing at least a
        ``code`` key with the actual authorization code: [...]
  > :param code: A dict of the authorization code grant and, optionally, state.


stubs/oauthlib/oauthlib/oauth2/rfc6749/request_validator.pyi
```diff
@@ -1,8 +1,21 @@
...
+    token_type: Literal["Bearer"]
+    access_token: str
+    expires_in: int
+    scope: str
+    refresh_token: NotRequired[str]
+    state: NotRequired[str]
+
+class _AuthorizationCode(TypedDict):
+    code: str
+    state: NotRequired[str]
```
JelleZijlstra, line 17: The docs suggest there is also an optional `nonce` key: https://oauthlib.readthedocs.io/en/latest/oauth2/validator.html?highlight=save_authorization_code#oauthlib.oauth2.RequestValidator.save_authorization_code

stubs/oauthlib/oauthlib/oauth2/rfc6749/request_validator.pyi
```diff
@@ -24,9 +37,9 @@ class RequestValidator:
     def invalidate_authorization_code(self, client_id: str, code: str, request: Request, *args, **kwargs) -> None: ...
     def revoke_token(self, token: str, token_type_hint: str, request: Request, *args, **kwargs) -> None: ...
     def rotate_refresh_token(self, request: Request) -> bool: ...
-    def save_authorization_code(self, client_id: str, code: str, request: Request, *args, **kwargs) -> None: ...
-    def save_token(self, token: dict[str, int | str], request: Request, *args, **kwargs) -> None: ...
-    def save_bearer_token(self, token: dict[str, int | str], request: Request, *args, **kwargs) -> str: ...
+    def save_authorization_code(self, client_id: str, code: _AuthorizationCode, request: Request, *args, **kwargs) -> None: ...
+    def save_token(self, token: _BearerToken | dict[str, Any], request: Request, *args, **kwargs) -> object: ...
```
JelleZijlstra, line 41: Why do you have the `| dict[str, Any]` here but not in the next method?

I'm of two minds about this. On the one hand, invariance means that it may be difficult for users to type their code correctly if we require a specific TypedDict with a non-exported type. On the other hand, giving `dict[str, Any]` as an option means that essentially any dict is accepted, so we lose any type safety from the TypedDict. On balance, I'd prefer to stick with just the TypedDict for now.

stubs/oauthlib/oauthlib/oauth2/rfc6749/request_validator.pyi
Daverball: Good catch

stubs/oauthlib/oauthlib/oauth2/rfc6749/request_validator.pyi
Daverball: Yeah, I am not totally happy about this either, I think the idea is that `save_token` accepts any token type and does the disambiguation and dispatch to the other `save_X_token` methods. Of course there's only the `Bearer` token right now in the RFC, so the function does not do any of that and just immediately calls `save_bearer_token`, but some implementations may have additional token types, so if they want to subclass this for that specific implementation they will have to ignore override errors.

That being said, `dict[str, Any]` is probably not the right way to go about this either, maybe it should just be `Mapping[str, object]` and then implementations can decide what to narrow it to, if anything. Or we just stick with the most common use-case and recognize that `save_token` doesn't have any safeguards, and eat the potential `override` errors.

Daverball: I went with `Mapping[str, object]`. I think this makes the most sense, since `save_token` should generally not be called manually unless you are implementing a non-standard OAuth workflow and in that case you may need to override `save_token` anyways, usually you will only override `save_bearer_token`.

I think what's being passed in is usually a `oauthlib.oauth2.rfc6749.tokens.Token` (although not always, sometimes it's just a plain `dict`) which is a subclass of `dict` and generated using `oauthlib.oauth2.rfc6749.tokens.BearerToken`. The problem is, that `Token` is generic so it can have any specific keys, but `BearerToken` does generate a dictionary with specific keys, so without intersections we can't really type this properly, so typing `save_bearer_token` strict and `save_token` lax feels like the best compromise for now.

JelleZijlstra: Thanks, we'll go with this for now and we can revisit it if users complain.

PR #11419, opened 2024-02-14
ftplib: Correct timeout option to `float | None`

heavywatal:
It is more consistent with the official document.
<https://docs.python.org/3/library/ftplib.html>

### DIFF
--- a/stdlib/ftplib.pyi
+++ b/stdlib/ftplib.pyi
@@ -31,7 +31,7 @@ class FTP:
     sock: socket | None
     welcome: str | None
     passiveserver: int
-    timeout: int
+    timeout: int | None
     af: int
     lastresp: str
     file: TextIO | None
@@ -48,7 +48,7 @@ class FTP:
             user: str = "",
             passwd: str = "",
             acct: str = "",
-            timeout: float = ...,
+            timeout: int | None = ...,
             source_address: tuple[str, int] | None = None,
             *,
             encoding: str = "utf-8",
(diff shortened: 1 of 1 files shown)


AlexWaygood:
Thanks! From skimming the source code of the runtime version of this module, I think you may be correct that `None` is acceptable. However, I think we should use `float | None` instead of `int | None`. Type annotations were only added to the official docs a few weeks ago, and it looks like they didn't mean to imply that passing a `float` is incorrect. See the conversation at https://github.com/python/cpython/pull/114359#discussion_r1489190455

heavywatal:
Thank you for the prompt feedback! Modified them to `float | None` accordingly.

stdlib/ftplib.pyi
```diff
@@ -31,7 +31,7 @@ class FTP:
     sock: socket | None
     welcome: str | None
     passiveserver: int
-    timeout: int
+    timeout: float | None
```
AlexWaygood, line 34: Agree that this could be `None` -- the implementation specifically checks to see whether or not it's `None`: https://github.com/python/cpython/blob/46245b0d831b9f5b15b4a0483c785ea71bffef12/Lib/ftplib.py#L153-L154

AlexWaygood: The `timeout` value is also eventually passed to `socket.create_connection` here: https://github.com/python/cpython/blob/46245b0d831b9f5b15b4a0483c785ea71bffef12/Lib/ftplib.py#L158-L159

And our annotations for `create_connection` state that `None` is accepted:

https://github.com/python/typeshed/blob/41245698e07a9959a4643122ddc814dff1ad3482/stdlib/socket.pyi#L803-L815

AlexWaygood (approved):
Thanks!

Issue #11418, opened 2024-02-13
Bug(?): `GzipFile` typing incompatible with `io.TextIOWrapper`


jamesbraza:
Please see the below code with Python 3.12:

```python
import gzip
import io


def gzip_to_list(file: gzip.GzipFile) -> list[str]:
    with io.TextIOWrapper(buffer=file) as f:
        return list(f)
```

Running `mypy==1.8.0`, I get this error message:

```none
a.py:6:34: error: Argument "buffer" to "TextIOWrapper" has incompatible type "GzipFile"; expected "IO[bytes]"  [arg-type]
```

However, running this code, it actually works fine.

I am thinking this is a typeshed bug, but I am not 100% positive.  https://github.com/python/typeshed/issues/5491 and https://github.com/python/typeshed/issues/6061 are possibly related

srittau:
I/O classes are a bit of a mess, since they predate protocols. `TextIOWrapper` should take a protocol as `buffer` argument. Someone needs to investigate the implementation of `TextIOWrapper` (found here: [https://github.com/python/cpython/blob/main/Modules/_io/textio.c]) to determine, which methods and attributes of `buffer` are actually accessed and then we could use such a protocol.

In fact, `TextIOWrapper` should probably be generic over the buffer type, but we can't do that before PEP 696 lands.

srittau:
Here is what I found:

* `seekable() -> bool`
* `readable() -> bool`
* `writable() -> bool`
* `tell() -> int`
* `write(b: bytes, /) -> object`
* `flush() -> object`
* `read(size: int = ...) -> bytes`
* `seek(offset: Literal[0], whence: Literal[2]) -> int`
* `truncate(size: int, /) -> int`
* `fileno() -> int`
* `isatty() -> int`
* `close() -> object`
* `name: str`
* `closed: bool`
* Optional: `read1(size: int, /) -> bytes` (strictly speaking `Buffer` would do as return type, but `bytes` is safer)

Some of these are only forwarded, so strictly speaking they are not necessary if the method is not called on the wrapper, but not requiring them would be unsafe.

It also checks for `buffer.raw` to exist and be a `FileIO` type under certain conditions, but ignores it if it isn't, so this is not part of the protocol.

AlexWaygood:
> Here is what I found

I found the same.

<details>
<summary>Links to the source code</summary>

[...]

PR #11420, opened 2024-02-14
Turn TextIOWrapper(buffer) into a protocol

srittau:
Closes: #11418


AlexWaygood:
Why do we need all those empty `__init__.py` files in the `test_cases` directory? They appear to be confusing our `regr_test.py` script a bit

srittau:
> Why do we need all those empty `__init__.py` files in the `test_cases` directory? They appear to be causing some confusion for our `regr_test.py` script

Because otherwise mypy gets confused about `check_io.pyi` and `typing/check_io.pyi`. But renaming the latter should help as well.

AlexWaygood:
> > Why do we need all those empty `__init__.py` files in the `test_cases` directory? They appear to be causing some confusion for our `regr_test.py` script
> 
> Because otherwise mypy gets confused about `check_io.pyi` and `typing/check_io.pyi`. But renaming the latter should help as well.

Ugh. Yeah, renaming seems the easiest option there for now... thanks!

stdlib/io.pyi
```diff
@@ -146,10 +149,33 @@ class TextIOBase(IOBase):
     def readlines(self, __hint: int = -1) -> list[str]: ...  # type: ignore[override]
     def read(self, __size: int | None = ...) -> str: ...
 
+@type_check_only
+class _WrappedBuffer(Protocol):
+    @property
+    def name(self) -> str: ...
+    @property
+    def closed(self) -> bool: ...
+    def read(self, size: int = ...) -> bytes: ...
```
AlexWaygood, line 158: Are there any instances of `size` being passed as a keyword argument? If not, the parameter in the protocol should be pos-only:

```suggestion
    def read(self, size: int = ..., /) -> bytes: ...
```

stdlib/io.pyi
```diff
@@ -146,10 +149,33 @@ class TextIOBase(IOBase):
...
+    # Optional: def read1(size: int, /) -> bytes: ...
```
AlexWaygood, line 159:
```suggestion
    # Optional: def read1(self, size: int, /) -> bytes: ...
```

stdlib/io.pyi
```diff
@@ -94,7 +94,10 @@ class BufferedIOBase(IOBase):
 
 class FileIO(RawIOBase, BinaryIO):  # type: ignore[misc]  # incompatible definitions of writelines in the base classes
     mode: str
-    name: FileDescriptorOrPath
+    # The type of "name" equals the argument passed in to the constructor,
+    # but that can make FileIO incompatible with other I/O types that assume
+    # "name" is a str. In the future, making FileIO generic might help.
+    name: Any
```
AlexWaygood, line 100: Are you sure it's `FileIO.name` that has the incorrect annotation, and not `_WrappedBuffer.name`? Does `_WrappedBuffer.name` _have_ to be a `str`?

Note that `TextIOWrapper.name` just returns the `name` attribute of the underlying buffer, and our annotations for `TextIOWrapper` currently have it just inheriting the `name` property from `typing.TextIO`, meaning that type checkers currently infer `TextIOWrapper.name` to be a property return `str | Any`:

https://github.com/python/typeshed/blob/a3d356c04c15afc22b254afae2e6e7adbbe031a5/stdlib/typing.pyi#L708-L711

So it's a bit weird for us to say "the underlying buffer object has to have a `name` attribute that's a `str`, but the `TextIOWrapper` object could return anything from its `name` property"

srittau: The I/O classes are just hilariously incoherent. I'm fine with just using `Any` in the protocol and/or restoring the `FileIO.name` annotation, although leaving the latter as `Any` for now is probably not a bad idea either.

AlexWaygood: My weak preference would be to use `Any` (or possibly `Incomplete`) for the new protocol for now, as it seems least likely to lead to unexpected and unintended false positives

srittau:
Hrm, cc @JelleZijlstra. Since this PR seem to fix the issues without adding to the primer output. I suggest to merge this first, and then have another look at Jelle's old PR.

AlexWaygood:
> Hrm, cc @JelleZijlstra. Since this PR seem to fix the issues without adding to the primer output. I suggest to merge this first, and then have another look at Jelle's old PR.

That sounds reasonable to me, though if it would be more accurate to have the `read()` method return `ReadableBuffer`, I think it would be safe to make that change. That change should _reduce_ the number of false-positives in user code, if anything.

stdlib/io.pyi
```diff
@@ -146,16 +149,45 @@ class TextIOBase(IOBase):
...
+@type_check_only
+class _WrappedBuffer(Protocol):
+    # "name" is wrapped by TextIOWrapper. Its type is inconsistent between
+    # the various I/O types, see the comments on TextIOWrapper.name and
+    # TextIO.name.
+    @property
+    def name(self) -> Any: ...
+    @property
+    def closed(self) -> bool: ...
+    def read(self, size: int = ..., /) -> ReadableBuffer: ...
```
JelleZijlstra, line 161:
```suggestion
    def read(self, size: int, /) -> ReadableBuffer: ...
```
It always [calls read with an argument](https://github.com/python/cpython/blob/81e140d10b77f0a41a5581412e3f3471cc77981f/Modules/_io/textio.c#L1891). Note this change makes the protocol more permissive.

stdlib/io.pyi
```diff
@@ -146,16 +149,45 @@ class TextIOBase(IOBase):
...
+        buffer: _WrappedBuffer,
         encoding: str | None = ...,
         errors: str | None = ...,
         newline: str | None = ...,
         line_buffering: bool = ...,
         write_through: bool = ...,
     ) -> None: ...
+    # Equals the "buffer" argument passed in to the constructor.
+    # TODO: Make TextIOWrapper generic over the buffer argument once
+    # PEP 696 get accepted.
```
JelleZijlstra, line 190:
```suggestion
    # PEP 696 gets accepted.
```
Probably need a similar comment for `detach()`.

JelleZijlstra (commented):
Thanks for picking this up. The major remaining difference from my PR seems to be that you're not overriding `buffer` and `detach`. This helps to get rid of some of the primer hits.

stdlib/io.pyi
srittau: Unfortunately that's not true: https://github.com/python/cpython/blob/81e140d10b77f0a41a5581412e3f3471cc77981f/Modules/_io/textio.c#L1977.

AlexWaygood (approved):
LGTM. https://github.com/python/typeshed/pull/11420#discussion_r1489781107 is optional.

PR #11426, opened 2024-02-15
Document `ignore_missing_stub` in CONTRIBUTING.md

srittau:
Also, make it reference `partial_stub` rather than vice versa

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -202,13 +202,15 @@ supported:
 * `partial_stub` (optional): This field marks the type stub package as
   [partial](https://peps.python.org/pep-0561/#partial-stub-packages). This is for
   3rd-party stubs that don't cover the entirety of the package's public API.
-  In most cases, this field is identical to `ignore_missing_stub`.
 
 In addition, we specify configuration for stubtest in the `tool.stubtest` table.
 This has the following keys:
 * `skip` (default: `false`): Whether stubtest should be run against this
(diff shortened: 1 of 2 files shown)


hauntsaninja (commented):
I think stubtest's flag is about any definition missing, not just entire modules.

CONTRIBUTING.md
```diff
@@ -202,13 +202,16 @@ supported:
...
 
 In addition, we specify configuration for stubtest in the `tool.stubtest` table.
 This has the following keys:
 * `skip` (default: `false`): Whether stubtest should be run against this
   package. Please avoid setting this to `true`, and add a comment if you have
   to.
+* `ignore_missing_stub`: When set to `true`, this will add the
+  `--ignore_missing_stub` option to the stubtest call. See
+  [tests/README.md](./tests/README.md) for more information. In most cases,
+  this field is identical to `partial_stub`.
```
Avasam, line 214: I know this is the same description you just moved. But it feels misleading to me. It sounds like either options can be used interchangeably in most cases.

When in truth, it's that they should be used together in most cases.

JelleZijlstra:
```suggestion
  [tests/README.md](./tests/README.md) for more information. In most cases,
  this field should be identical to `partial_stub`.
```
I think this makes it clearer

JelleZijlstra (approved):

PR #11427, opened 2024-02-15
Argparse: Action.__call__ allows arbitrary value type

abelcheung:
For example, value can be `int` if using `parser.add_argument(type='int')`.

`Sequence[Any]` part is untouched since people may not be aware of it if barebone `Any` is used.

EDIT: The explanation in commit message might be misleading with `action='store_true'`. Didn't have daily coffee dose yet.


abelcheung:
Sorry for the brain fart, should have pull myself together before sending PR.

PR #11429, opened 2024-02-16
html5lib: Add various types

JelleZijlstra:
I started out investigating comments in #11411 and ended up adding a few other
types that were reasonably obvious from the source code. For reference:

https://github.com/html5lib/html5lib-python/tree/master/html5lib

cc @leonardr


leonardr:
I only have experience with a small part of the html5lib API, but these changes make sense to me.

hauntsaninja (approved):
It's weird that the encodings can be bytes, but I guess that's Python 2 for you

PR #11434, opened 2024-02-17
deps: upgrade pytype

JelleZijlstra:
(And Ruff because why not.)

This should fix CI on main (#11333).


JelleZijlstra:
The remaining failure is #11428, will address that separately.

PR #11436, opened 2024-02-17
importlib.metadata: Improve and test SimplePath protocol

JelleZijlstra:
Split out from #9217. See the runtime implementation at https://github.com/python/cpython/blob/main/Lib/importlib/metadata/_meta.py.

I don't really understand why this was made generic at runtime. I don't think it should be.

This leaves out changes to the rest of `importlib.metadata` for simplicity. We can deal with those in a subsequent PR.


JelleZijlstra:
The remaining pyright failure can be reduced to [this](https://pyright-play.net/?pythonVersion=3.12&code=GYJw9gtgBA%2BjwFcAuCQFM5QJYQA5hCSgEMA7UsJYpLMUgZwFgAoUSKJAT1y1IHNseAkQCCpTgBooABXBIwAYzAAbFiwXLi9ejLlgA2gBUAugApZlRSoCUALhZRHUACZpgUAFZheuagAtTejRlYClKPzQQWyh6JBAoAB8oQykAemsoAFoAPmTogDpChydXdzg4hDRnLAA3OEDg0KhwyOjY%2BKSUqHSs3MMCouZ1TW0oAFlOaX8ARntmJxc3T29SXyQAoJCw9daYuLSMnPHJmYH84sdS2BgKqtr6zaaWqL2QA97jqfXZqELzoeYGi0OgmXz8ACY5gsrl4fP4GltmjsXu13kcxJwzhdFmUbiBKtU6jAEU9kW19t1DrlQf5Ib9BiwrlQANZoegwXB6Uy4aIWeT6DHGOz0-4sFlsjlcmnfUzWaxi4is9mcyymaUQ2XWIA):

```python
from typing import Any, Protocol

class Proto[T](Protocol):
    def joinpath(self, other: str | T, /) -> T: ...
    def __truediv__(self, other: str | T, /) -> T: ...

class MyPath1:
    def joinpath(self, other: str, /) -> MyPath1: ...
    def __truediv__(self, other: str, /) -> MyPath1: ...

class MyPath2:
    def joinpath(self, other: str, /) -> Any: ...
    def __truediv__(self, other: str, /) -> MyPath2: ...

def takes_proto(p: Proto[Any]): ...

takes_proto(MyPath1())  # error
takes_proto(MyPath2())  # ok
```

```
Argument of type "MyPath1" cannot be assigned to parameter "p" of type "Proto[Any]" in function "takes_proto"
  "MyPath1" is incompatible with protocol "Proto[Any]"
    "__truediv__" is an incompatible type
      Type "(other: str, /) -> MyPath1" cannot be assigned to type "(other: str | T@Proto, /) -> T@Proto"
        Parameter 1: type "str | T@Proto" cannot be assigned to type "str"
          Type "str | MyPath1" cannot be assigned to type "str"  (reportArgumentType)
```

Pyright is at least arguably right there: for MyPath1 to be compatible with `Proto`, we need to solve `T` to `MyPath1`, and that means the parameters are incompatible. I think to fix this I need to adjust the definition of `SimplePath` some more.

AlexWaygood:
Is it useful to have mypy complaining at us about empty bodies in the test cases? Maybe we should add `--disable-error-code=empty-body` to the flags in `regr_test.py`

hauntsaninja (approved):

PR #11440, opened 2024-02-18
`tensorflow`: add `tensorflow.bitwise`

hoel-bagard:


stubs/tensorflow/tensorflow/bitwise.pyi
```diff
@@ -0,0 +1,33 @@
+from typing import Any, overload
+from typing_extensions import TypeAlias
+
+import numpy as np
+import tensorflow as tf
+from tensorflow._aliases import FloatArray, IntArray
+
+_BitwiseCompatible: TypeAlias = tf.Tensor | bool | int | FloatArray | IntArray | np.number[Any]
```
JelleZijlstra, line 8: Are you sure about bool? I tried `tensorflow.bitwise.bitwise_and(True, False)` and it throws. Though `tensorflow.bitwise.bitwise_and(0, False)` goes through.

It's redundant anyway because bool is a subtype of int.
```suggestion
_BitwiseCompatible: TypeAlias = tf.Tensor | int | FloatArray | IntArray | np.number[Any]
```

hoel-bagard: After a few test, I got the following results (the same holds for `or`, `xor`, etc...)

```python
tensorflow.bitwise.bitwise_and(True, False)   # Throws 
tensorflow.bitwise.bitwise_and(True, 0)       # Throws
tensorflow.bitwise.bitwise_and(1, False)      # Ok
tensorflow.bitwise.bitwise_and(1, 0)          # Ok
```

I suppose this could be typed using overloads, do you think it is worth doing ?

JelleZijlstra: Up to you, I don't know much about how Tensorflow is used. I would imagine in practice people are unlikely to use these operators with literal bools or integers (they'd use tensors, or why are they using tensorflow?), so maybe we wouldn't help a lot of people by adding this complexity.

hoel-bagard: That's also what I'm thinking. I'll add a comment next to the type alias just in case someone wonders about it in the future.

JelleZijlstra (approved):

PR #11442, opened 2024-02-18
`tensorflow`: add `tensorflow.autodiff`

hoel-bagard:


stubs/tensorflow/tensorflow/autodiff.pyi
```diff
@@ -0,0 +1,61 @@
...
+from typing import overload
+from typing_extensions import Self
+
+import tensorflow as tf
+from tensorflow import Tensor, UnconnectedGradients, Variable
+from tensorflow._aliases import ContainerGradients, ContainerTensors, ContainerTensorsLike, Gradients, TensorLike
+
+class ForwardAccumulator:
+    def __init__(self, primals: Tensor, tangents: Tensor) -> None: ...
+    def jvp(
```
JelleZijlstra, line 15: This also has `__enter__` and `__exit__` methods, are those worth adding?

hoel-bagard: Since the class is meant to be used as a context manager, and the methods are present for `GradientTape`, I've added them in f3f3cc9b0.

PR #11443, opened 2024-02-18
`tensorflow`: add `tensorflow.autograph`

hoel-bagard:


stubs/tensorflow/tensorflow/autograph/__init__.pyi
```diff
@@ -0,0 +1,27 @@
...
+
+from tensorflow.autograph.experimental import Feature
+
+_Param = ParamSpec("_Param")
+_RetType = TypeVar("_RetType")
+_Type = TypeVar("_Type")
+
+def set_verbosity(level: int, alsologtostdout: bool = False) -> None: ...
+def to_code(
+    entity: Callable[_Param, Any] | type,
```
JelleZijlstra, line 13:
```suggestion
    entity: Callable[..., Any],
```
A ParamSpec doesn't make sense here as it's the only use in the function, and a type is automatically a callable, so the second union member is redundant.

stubs/tensorflow/tensorflow/autograph/__init__.pyi
```diff
@@ -0,0 +1,27 @@
...
+    recursive: bool = True,
+    experimental_optional_features: None | Feature | tuple[Feature, ...] = None,
+) -> str: ...
+@overload
```
JelleZijlstra, line 17: I don't see what these overloads add, as the second overload accepts everything the first one does.

hoel-bagard: Fixed in baf73245d. The first overload was useless indeed...

PR #11444, opened 2024-02-18
`tensorflow`: add `tensorflow.keras.activations` members

hoel-bagard:


stubs/tensorflow/tensorflow/keras/activations.pyi
```diff
@@ -1,12 +1,37 @@
...
+def linear(x: _ActivationInput) -> Tensor: ...
+def mish(x: _ActivationInput) -> Tensor: ...
+def relu(
+    x: _ActivationInput,
+    alpha: FloatTensorCompatible = 0.0,
+    max_value: FloatTensorCompatible | FloatDataSequence | None = None,
+    threshold: FloatTensorCompatible | FloatDataSequence = 0.0,
+) -> Tensor: ...
+def selu(x: _ActivationInput) -> Tensor: ...
+def serialize(activation: Callable[..., Any], use_legacy_format: bool = False) -> str: ...
```
JelleZijlstra, line 31: It can also return a dict:

```
In [15]: tensorflow.keras.activations.serialize(len)
Out[15]: 
{'module': 'builtins',
 'class_name': 'builtin_function_or_method',
 'config': 'len',
 'registered_name': 'builtin_function_or_method'}

In [16]: tensorflow.keras.activations.serialize(lambda: 42)
/Users/jelle/py/venvs/py311/lib/python3.11/site-packages/keras/src/activations.py:549: UserWarning: The object being serialized includes a `lambda`. This is unsafe. In order to reload the object, you will have to pass `safe_mode=False` to the loading function. Please avoid using `lambda` in the future, and use named Python functions instead. This is the `lambda` being serialized: tensorflow.keras.activations.serialize(lambda: 42)

  fn_config = serialization_lib.serialize_keras_object(activation)
Out[16]: 
{'value': ('4wAAAAAAAAAAAAAAAAEAAAADAAAA8wYAAACXAGQBUwApAk7pKgAAAKkAcgMAAADzAAAAAPofPGlw\neXRob24taW5wdXQtMTYtYjFiYzU1YjgzOGViPvoIPGxhbWJkYT5yBgAAAAEAAABzBgAAAIAAqHKA\nAHIEAAAA\n',
  None,
  None)}

In [17]: tensorflow.keras.activations.serialize(os.path.exists)
Out[17]: 
{'module': 'builtins',
 'class_name': 'function',
 'config': 'exists',
 'registered_name': 'function'}
```

hoel-bagard: @JelleZijlstra Thanks for pointing it out, I'll fix that. Do you have a method / tool to find this kind of errors ?
For this particular function, looking at the source code would have made it clear that the return type could be a dict. Is it how you found out about it ?

When doing the TensorFlow stubs, I usually try a few inputs, but I can't test everything, so I often rely on what the docs say :disappointed:

JelleZijlstra: Just by trying it out. I didn't do that for all functions, I think at first I was mostly curious what this function would do with a builtin.

stubs/tensorflow/tensorflow/keras/activations.pyi
```diff
@@ -1,12 +1,37 @@
...
+def linear(x: _ActivationInput) -> Tensor: ...
+def mish(x: _ActivationInput) -> Tensor: ...
+def relu(
+    x: _ActivationInput,
+    alpha: FloatTensorCompatible = 0.0,
+    max_value: FloatTensorCompatible | FloatDataSequence | None = None,
+    threshold: FloatTensorCompatible | FloatDataSequence = 0.0,
+) -> Tensor: ...
+def selu(x: _ActivationInput) -> Tensor: ...
+def serialize(activation: Callable[..., Any], use_legacy_format: bool = False) -> str | dict[str, str]: ...
```
JelleZijlstra, line 31: The lambda returned a dict with a single tuple value. The exact format probably isn't to be relied upon.

```suggestion
def serialize(activation: Callable[..., Any], use_legacy_format: bool = False) -> str | dict[str, Any]: ...
```

PR #11445, opened 2024-02-18
importlib.metadata: Use the SimplePath protocol

JelleZijlstra:
Closes #9217. Followup from #11436.

This mostly makes a simpler set of changes than #9217; in particular
it does not make PathDistribution generic. I think this is in line with
how the protocol is supposed to be used, though not always in line with
the runtime type annotations (which don't all make sense).

### DIFF
--- a/stdlib/importlib/metadata/__init__.pyi
+++ b/stdlib/importlib/metadata/__init__.pyi
@@ -9,7 +9,7 @@ from importlib.abc import MetaPathFinder
 from os import PathLike
 from pathlib import Path
 from re import Pattern
-from typing import Any, ClassVar, Generic, NamedTuple, TypeVar, overload
+from typing import Any, ClassVar, Generic, NamedTuple, TypeAlias, TypeVar, overload
 from typing_extensions import Self
 
 _T = TypeVar("_T")
@@ -33,8 +33,12 @@ if sys.version_info >= (3, 10):
     __all__ += ["PackageMetadata", "packages_distributions"]
 
 if sys.version_info >= (3, 10):
-    from importlib.metadata._meta import PackageMetadata as PackageMetadata
+    from importlib.metadata._meta import PackageMetadata as PackageMetadata, SimplePath
     def packages_distributions() -> Mapping[str, list[str]]: ...
+
+    _SimplePath: TypeAlias = SimplePath
+else:
+    _SimplePath: TypeAlias = Path
 
 class PackageNotFoundError(ModuleNotFoundError):
     @property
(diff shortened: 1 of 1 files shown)


JelleZijlstra:
The primer hit seems real, e.g. https://github.com/python-poetry/poetry/blob/e015ee534eb09c7d0add6a8440c944ff132ff6f4/src/poetry/utils/env/site_packages.py#L124. Probably the best fix would be for `SimplePath` to add the `__fspath__` method.

JelleZijlstra:
No, that would be wrong, since `zipfile.Path` does not have a `__fspath__` method and can't be converted to a `pathlib.Path`. So maybe the new errors are true positives.

stdlib/importlib/metadata/__init__.pyi
```diff
@@ -233,14 +241,14 @@ class MetadataPathFinder(DistributionFinder):
...
         def invalidate_caches(cls) -> None: ...
 
 class PathDistribution(Distribution):
-    _path: Path
-    def __init__(self, path: Path) -> None: ...
-    def read_text(self, filename: StrPath) -> str: ...
-    def locate_file(self, path: StrPath) -> PathLike[str]: ...
+    _path: _SimplePath
+    def __init__(self, path: _SimplePath) -> None: ...
+    def read_text(self, filename: StrPath) -> str | None: ...
```
srittau, line 250: Considering that neither `SimplePath.read_text()` nor `pathlib.Path.read_text()` can return `None`, I think just `str` is the better annotation here.

stdlib/importlib/metadata/__init__.pyi
```diff
@@ -233,14 +241,14 @@ class MetadataPathFinder(DistributionFinder):
     @classmethod
     def find_distributions(cls, context: DistributionFinder.Context = ...) -> Iterable[PathDistribution]: ...
     if sys.version_info >= (3, 10):
-        # Yes, this is an instance method that has argumend named "cls"
+        # Yes, this is an instance method that has an argument named "cls"
```
srittau, line 244: While we're here:

```suggestion
        # Yes, this is an instance method that has a parameter named "cls"
```

srittau (changes requested):

stdlib/importlib/metadata/__init__.pyi
JelleZijlstra: It returns None if accessing the path throws an exception: https://github.com/python/cpython/blob/53d5e67804227d541ed2f9e8efea8de5d70cb1ec/Lib/importlib/metadata/__init__.py#L812

srittau: You're right. I was looking at the 3.11 branch, where it doesn't have the explicit `return None`, but it still returns `None` implicitly.

srittau (approved):

PR #11446, opened 2024-02-18
Improve typing of `start_new_thread` when called without kwargs

hamdanal:


AlexWaygood (approved):
Nice, thank you! This seems reasonable unless/until we get some kind of TypeVarDict thing to express functions like this properly (which is a long way off, if it ever even happens)

AlexWaygood:
Though I guess the question is whether we're therefore okay with making _all_ functions like this in typeshed overloaded functions? Not sure how many there are, but I know there's more than just this one. I guess I'll leave this open for a day or two in case any other maintainers have objections :)

JelleZijlstra (approved):
Seems fine to me

PR #11451, opened 2024-02-19
`shlex.split`: allow `TextIO` and deprecate `None`

Avasam:
Whilst working on a setuptools stubs update (namely in `setuptools.command.easy_install.CommandSpec.from_param`), I noticed that `shlex.split` didn't allow `TextIO` like `shlex.shlex.__init__` does. And that passing `None` can be allowed (but deprecated) until Python 3.12.

Unless `s` being anything else than a `str` is an accidental implementation detail that should be hidden away.

### DIFF
--- a/stdlib/shlex.pyi
+++ b/stdlib/shlex.pyi
@@ -1,10 +1,20 @@
+import sys
 from collections.abc import Iterable
-from typing import TextIO
-from typing_extensions import Self
+from typing import TextIO, overload
+from typing_extensions import Self, deprecated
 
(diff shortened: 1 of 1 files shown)


JelleZijlstra (approved):

srittau:
Just a general note: `TextIO` isn't an ideal annotation for parameters (similar to `BinaryIO` and `IO`). Usually, a protocol is a better option. That said, as this is only forwarded, this is better than leaving it out.

srittau (approved):

PR #11454, opened 2024-02-19
Improve typing of `sysconfig.get_config_var(s)`

Avasam:
`distutils.sysconfig.get_config_vars` update is relevant for setuptools (https://github.com/pypa/setuptools/pull/4228). I've updated `get_config_var` and all duplicates of these method as well.


stdlib/sysconfig.pyi
```diff
@@ -15,11 +16,15 @@ __all__ = [
...
+@deprecated("SO is deprecated, use EXT_SUFFIX")
+def get_config_var(name: Literal["SO"]) -> Any: ...
+@overload
 def get_config_var(name: str) -> Any: ...
 @overload
-def get_config_vars() -> dict[str, Any]: ...
+def get_config_vars() -> dict[str, Any]: ...  # type: ignore[overload-overlap]
 @overload
-def get_config_vars(arg: str, *args: str) -> list[Any]: ...
+def get_config_vars(*args: str) -> list[Any]: ...
```
Avasam, line 27: mypy says overloads overlap, but also gives the right return type

I could instead do
```suggestion
def get_config_vars(arg: str, /, *args: str) -> list[Any]: ...
```

stdlib/distutils/sysconfig.pyi
```diff
@@ -10,8 +12,15 @@ project_base: str
...
 
 def expand_makefile_vars(s: str, vars: Mapping[str, str]) -> str: ...
+@overload
+@deprecated("SO is deprecated, use EXT_SUFFIX")
+def get_config_var(name: Literal["SO"]) -> int | str | None: ...
+@overload
 def get_config_var(name: str) -> int | str | None: ...
-def get_config_vars(*args: str) -> Mapping[str, int | str]: ...
+@overload
+def get_config_vars() -> dict[str, str | int]: ...  # type: ignore[overload-overlap]
```
JelleZijlstra, line 21: You can probably get rid of this ignore by making the other overload have an extra positional-only argument, like `def get_config_vars(arg: str, /, *args: str)`.

Avasam: Alright so same as my comment above, I'll do that then, thanks for your input

stdlib/sysconfig.pyi
```diff
@@ -15,11 +16,15 @@ __all__ = [
     "parse_config_h",
 ]
 
+@overload
+@deprecated("SO is deprecated, use EXT_SUFFIX")
```
JelleZijlstra, line 20: sysconfig.py doesn't mention this, can you link to where it's indicated that "SO" is deprecated?

Avasam: Looks like support has been removed in Python 3.11
https://github.com/python/cpython/pull/91671/files#diff-d593bd299ba58e440ba411ffa0640ccd9d20d518b0cf2644ed4bdb75a82a3e70L670-L673

I'll add this to the deprecation comment now that I found this.

JelleZijlstra: Thanks, I only looked at the main branch.

JelleZijlstra (approved):

PR #11459, opened 2024-02-21
Add `wurlitzer` stubs

hoel-bagard:
This PR adds stubs for the [wurlitzer](https://github.com/minrk/wurlitzer) package.


stubs/wurlitzer/wurlitzer/__init__.pyi
AlexWaygood, line 1: I think this file could probably be just `stubs/wurlitzer/wurlitzer.pyi`, rather than `stubs/wurlitzer/wurlitzer/__init__.pyi`

stubs/wurlitzer/wurlitzer/__init__.pyi
```diff
@@ -0,0 +1,41 @@
...
+def dup2(a: int, b: int, timeout: int = 3) -> int: ...
+def sys_pipes(encoding: str = ..., bufsize: int | None = None) -> contextlib._GeneratorContextManager[None]: ...
+@contextmanager
+def pipes(
+    stdout: _PIPE | TextIO = 3, stderr: _STDOUT | _PIPE | TextIO = 3, encoding: str = ..., bufsize: int | None = None
+) -> Iterator[tuple[TextIO | io.BytesIO | io.StringIO, TextIO | io.BytesIO | io.StringIO | None]]: ...
+def sys_pipes_forever(encoding: str = ..., bufsize: int | None = None): ...
+def stop_sys_pipes() -> None: ...
+def load_ipython_extension(ip: Incomplete) -> None: ...  # IPython.core.interactiveshell.InteractiveShell
+def unload_ipython_extension(ip: Incomplete) -> None: ...  # IPython.core.interactiveshell.InteractiveShell
```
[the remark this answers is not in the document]
hoel-bagard: Done in 6bc4a65b5.

Those functions are not documented and are not int the `__all__` export, so I'm not sure if they should be included in the stubs. They are not private, so I added them for completeness.

stubs/wurlitzer/wurlitzer/__init__.pyi
```diff
@@ -0,0 +1,43 @@
...
+from contextlib import contextmanager
+from types import TracebackType
+from typing import Any, Final, Literal, TextIO
+from typing_extensions import TypeAlias
+
+STDOUT: Final = 2
+PIPE: Final = 3
+_STDOUT: TypeAlias = Literal[2]
+_PIPE: TypeAlias = Literal[3]
+_InteractiveShell: TypeAlias = Any
```
AlexWaygood, line 13:
```suggestion

# Alias for IPython.core.interactiveshell.InteractiveShell.
# N.B. Even if we added ipython to the stub-uploader allowlist,
# we wouldn't be able to declare a dependency on ipython here,
# since `wurlitzer` does not declare a dependency on `ipython` at runtime
_InteractiveShell: TypeAlias = Any
```

stubs/wurlitzer/wurlitzer/__init__.pyi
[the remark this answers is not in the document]
AlexWaygood: Hmm. I'll leave it up to you. Our CI won't complain if they're deleted from the stubs, since they're not included in `__all__`. But some people could possibly find them useful, and having them in the stubs won't be a big maintenance burden for us. Overall I have no strong opinion either way :)

stubs/wurlitzer/wurlitzer/__init__.pyi
hoel-bagard: Done in 378b2d01f.

hoel-bagard:
Thanks for the (very fast :open_mouth: ) review, I'll look into using protocols.

hoel-bagard:
@AlexWaygood Do you know what in the overloads is causing the CI to fail ?

Akuli:
The error looks like a bug in stubtest, but you can probably work around it by not using overloaded context managers. Instead of `@contextmanager` together with `-> Iterator[Foo]`, you could try `-> AbstractContextManager[Foo]` or `-> _GeneratorContextManager[Foo]`. You need to import from `contextlib`.

AlexWaygood:
> The error looks like a bug in stubtest, but you can probably work around it by not using overloaded context managers. Instead of `@contextmanager` together with `-> Iterator[Foo]`, you could try `-> AbstractContextManager[Foo]` or `-> _GeneratorContextManager[Foo]`. You need to import from `contextlib`.

Yeah, @Akuli is right. Stubtest in general isn't great at dealing with several decorators on a single function -- see https://github.com/python/mypy/issues/14950 (which is now fixed, but was only fixed by special-casing `@final`).

hoel-bagard:
@AlexWaygood If you have time to review the changes.

stubs/wurlitzer/wurlitzer.pyi
```diff
@@ -0,0 +1,97 @@
...
+def pipes(
+    stdout: _PIPE, stderr: _STDOUT, encoding: str, bufsize: int | None = None
+) -> _GeneratorContextManager[tuple[io.StringIO, None]]: ...
+@overload
+def pipes(
+    stdout: _PIPE, stderr: _PIPE, encoding: str, bufsize: int | None = None
+) -> _GeneratorContextManager[tuple[io.StringIO, io.StringIO]]: ...
+@overload
+def pipes(
+    stdout: _PIPE, stderr: _StreamErrT, encoding: str, bufsize: int | None = None
```
Akuli, line 67: The encoding is a string by default, so `encoding: str` overloads should instead be `encoding: str = ...`. This way they are used when no encoding is given, and wurlitzer uses `"utf-8"` or whatever the default happens to be.

```suggestion
    stdout: _PIPE, stderr: _STDOUT, encoding: str = ..., bufsize: int | None = None
) -> _GeneratorContextManager[tuple[io.StringIO, None]]: ...
@overload
def pipes(
    stdout: _PIPE, stderr: _PIPE, encoding: str = ..., bufsize: int | None = None
) -> _GeneratorContextManager[tuple[io.StringIO, io.StringIO]]: ...
@overload
def pipes(
    stdout: _PIPE, stderr: _StreamErrT, encoding: str = ..., bufsize: int | None = None
```

stubs/wurlitzer/wurlitzer.pyi
```diff
@@ -0,0 +1,97 @@
...
+) -> _GeneratorContextManager[tuple[io.StringIO, _StreamErrT]]: ...
+@overload
+def pipes(
+    stdout: _StreamOutT, stderr: _STDOUT, encoding: str | None, bufsize: int | None = None
```
Akuli, line 71: This too should be matched when the default encoding is used:

```suggestion
    stdout: _StreamOutT, stderr: _STDOUT, encoding: str | None = ..., bufsize: int | None = None
```

stubs/wurlitzer/wurlitzer.pyi
```diff
@@ -0,0 +1,97 @@
...
+) -> _GeneratorContextManager[tuple[_StreamOutT, None]]: ...
+@overload
+def pipes(
+    stdout: _StreamOutT, stderr: _PIPE, encoding: None, bufsize: int | None = None
+) -> _GeneratorContextManager[tuple[_StreamOutT, io.BytesIO]]: ...
+@overload
+def pipes(
+    stdout: _StreamOutT, stderr: _PIPE, encoding: str, bufsize: int | None = None
```
Akuli, line 79:
```suggestion
    stdout: _StreamOutT, stderr: _PIPE, encoding: str = ..., bufsize: int | None = None
```

stubs/wurlitzer/wurlitzer.pyi
```diff
@@ -0,0 +1,97 @@
...
+) -> _GeneratorContextManager[tuple[_StreamOutT, io.StringIO]]: ...
+@overload
+def pipes(
+    stdout: _StreamOutT, stderr: _StreamErrT, encoding: str | None, bufsize: int | None = None
```
Akuli, line 83:
```suggestion
    stdout: _StreamOutT, stderr: _StreamErrT, encoding: str | None = ..., bufsize: int | None = None
```

stubs/wurlitzer/wurlitzer.pyi
```diff
@@ -0,0 +1,97 @@
...
+) -> _GeneratorContextManager[tuple[_StreamOutT, _StreamErrT]]: ...
+@overload
+def pipes(
+    stdout: _PIPE | _StreamOutT = 3,
+    stderr: _STDOUT | _PIPE | _StreamErrT = 3,
+    encoding: str | None = ...,
+    bufsize: int | None = None,
+) -> _GeneratorContextManager[tuple[io.BytesIO | io.StringIO | _StreamOutT, io.BytesIO | io.StringIO | _StreamErrT | None]]: ...
+def sys_pipes_forever(encoding: str = ..., bufsize: int | None = None): ...
```
Akuli, line 92:
```suggestion
def sys_pipes_forever(encoding: str = ..., bufsize: int | None = None) -> None: ...
```

stubs/wurlitzer/wurlitzer.pyi
```diff
@@ -0,0 +1,97 @@
...
+def stop_sys_pipes() -> None: ...
+def load_ipython_extension(ip: _InteractiveShell) -> None: ...
+def unload_ipython_extension(ip: _InteractiveShell) -> None: ...
+
+__all__ = ["STDOUT", "PIPE", "Wurlitzer", "pipes", "sys_pipes", "sys_pipes_forever", "stop_sys_pipes"]
```
Akuli, line 97: Nit: It would be better to place `__all__` near the top of the file, similarly to how it is in wurlitzer's source code.

Akuli (commented):
Thanks for working on this! Typing is hard :)

stubs/wurlitzer/wurlitzer.pyi
hoel-bagard: Done in c34be49.

hoel-bagard: Done in c34be49.

hoel-bagard:
@Akuli Thanks for the improvements/fixes!

Akuli (approved):
Thanks!

PR #11460, opened 2024-02-21
WebOb: Complete the stubs and activate stricter pyright config

Daverball:
This adds the final little bibs and bobs that were still missing.

I'll readily admit I went a little bit insane writing the stubs for `webob.dec.wsgify`, so reviewing this one might be a little tough, I did add some tests though to increase confidence, source code can be found here:
https://github.com/Pylons/webob/blob/1.8.7/src/webob/dec.py


Daverball:
I'll have a look at what pyright is unhappy with in the test cases later.

stubs/WebOb/@tests/stubtest_allowlist.txt
```diff
@@ -137,6 +137,17 @@ webob.request.AdhocAttrMixin.__setattr__
 # make sense to annotate them and pretend they're part of the API.
 webob.request.BaseRequest.__init__
 
+# We needed to add a dummy *_: _P.args in order to support ParamSpec
+webob.dec.wsgify.middleware
+webob.dec._MiddlewareFactory.__call__
+
+# We renamed some of the arguments in positional only overloads for greater
+# clarity about what the arguments mean, stubtest should probably be a bit
```
JelleZijlstra, line 145: Maybe open a mypy issue with a precise example.

stubs/WebOb/webob/dec.pyi
```diff
@@ -1,49 +1,196 @@
...
+    _MiddlewareCallable[_RequestT_contra, _AppT_contra, _P] | _MiddlewareMethod[_RequestT_contra, _AppT_contra, _P]
+)
+
+class wsgify(Generic[_RequestT_contra, _P]):
+    RequestClass: type[Request]
+    func: _RequestHandler[_RequestT_contra, _P] | None
+    args: tuple[Any, ...]
+    kwargs: dict[str, Any]
+    middleware_wraps: WSGIApplication | None
+    # NOTE: We disallow passing args/kwargs using this direct API, because
```
JelleZijlstra, line 33: Why can't we have an overload where args/kwargs are nonempty and the rest of the types reduce to Any? That way it's only nonsafe if the user explicitly passed args/kwargs.

stubs/WebOb/webob/dec.pyi
```diff
@@ -1,49 +1,196 @@
...
+    @overload
+    def __call__(self, req: _RequestT_contra, *args: _P.args, **kw: _P.kwargs) -> _AnyResponse: ...
+    def get(self, url: str, **kw: Any) -> _AnyResponse: ...
+    def post(
+        self, url: str, POST: str | bytes | Mapping[Any, Any] | Mapping[Any, list[Any] | tuple[Any, ...]] | None = None, **kw: Any
+    ) -> _AnyResponse: ...
+    def request(self, url: str, **kw: Any) -> _AnyResponse: ...
+    def call_func(self, req: _RequestT_contra, *args: _P.args, **kwargs: _P.kwargs) -> _AnyResponse: ...
+    # technically this could bind different type vars, but we disallow it for safety
+    def clone(self, func: _RequestHandler[_RequestT_contra, _P] | None = None, **kw: Never) -> Self: ...
```
JelleZijlstra, line 114: Why not omit `**kw`? That probably produces better errors if users try to pass args.

stubs/WebOb/webob/dec.pyi
```diff
@@ -1,49 +1,196 @@
...
+        cls, middle_func: _MiddlewareMethod[_RequestT, _AppT, _P2], app: _AppT
+    ) -> type[wsgify[_RequestT, Concatenate[_AppT, _P2]]]: ...
+    @overload
+    @classmethod
+    def middleware(
+        cls, middle_func: _MiddlewareMethod[_RequestT, _AppT, _P2], app: _AppT, *_: _P2.args, **kw: _P2.kwargs
+    ) -> type[wsgify[_RequestT, Concatenate[_AppT, _P2]]]: ...
+
+class _unbound_wsgify(wsgify[_RequestT_contra, _P], Generic[_RequestT_contra, _P, _S]):
+    @overload  # type:ignore[override]
```
JelleZijlstra, line 149:
```suggestion
    @overload  # type: ignore[override]
```

JelleZijlstra (commented):
I'm no expert on WSGI but this seems to make sense. A few questions though.

Do you have a minimal repro of the pyright issue that caused you to comment out some cases? I tried a simpler version in the pyright playground but couldn't reproduce it.

stubs/WebOb/webob/dec.pyi
Daverball: I tried that (with an absolutely absurd number of overloads), unfortunately, when you do that it will complain if you use it to decorate a callable that doesn't also itself take arbitrary `args` and `kwargs`. The types of the arguments don't matter in this case, but the number of arguments unfortunately does matter.

i.e. stuff like this won't work:
```
@wsgify(args=(1.0,))  # type error: wsgify[Request, ...].__call__ no overloads match
def foo(req: Request, timeout: float) -> None: ...
```

I could maybe get around this by adding two more subclasses, which are generic in `_F` instead of `_P`, so they can decorate arbitrary callables and adding a `__new__` to return them, but it doesn't really seem worth the effort, since it won't help with subclasses of wsgify.

The alternative is giving up an parameter safety altogether, or try to only use a `ParamSpec` for the `middleware` call.

stubs/WebOb/webob/dec.pyi
Daverball: That's fair, I think I mostly did this to avoid the stubtest error and `Never` in parameters produces far better error messages than in return types.

Daverball:
> Do you have a minimal repro of the pyright issue that caused you to comment out some cases? I tried a simpler version in the pyright playground but couldn't reproduce it.

Unfortunately I cannot reproduce it with a simpler version on pyright playground either

Issue #11259, opened 2024-01-07
`os.EX_NOTFOUND` was... not found


qexat:
In the module `os`, `EX_NOTFOUND` is exposed on platforms that are neither `win32` nor `darwin`. ([source](https://github.com/python/typeshed/blob/main/stdlib/os/__init__.pyi#L314-L315))

Using **Python 3.11.6** on **EndeavourOS** 64-bit (Linux 6.6.10-arch1-1), raises at runtime:

```py
AttributeError: module 'os' has no attribute 'EX_NOTFOUND'
```

I can also reproduce it with:
* Python 3.9.18
* Python 3.8.18
* Python 3.13.0 (built from source from the `main` branch)

srittau:
According to https://stackoverflow.com/questions/51322536/python-on-unix-value-of-os-ex-notfound it exists on e.g. Solaris. We should add an `and sys.platform != "linux"` to the condition. PR welcome!

hauntsaninja:
@qexat Thanks for the issue! :-) I am kind of curious — what were you doing that led you to noticing this?

(In general, determining platform availability of constants has been a little tricky for Linux, since Linux / Python builds on Linux is not a monolith. So I'm quite curious about use cases that depend on this. As per usual, typeshed prefers false negatives over false positives, so we likely overstate constant availability on Linux anywhere we have a regex in stubtest allowlists)

qexat:
> @qexat Thanks for the issue! :-) I am kind of curious — what were you doing that led you to noticing this?

For my snippets repository, I was writing a little dependency checker (very basic and in a few lines, really) that exits early with a code and I wanted to try using `os` constants — the module supposedly makes programs more portable :)

PR #11461, opened 2024-02-22
Update `os. EX_NOTFOUND` not available in Linux

aminalaee:
Closes https://github.com/python/typeshed/issues/11259

### DIFF
--- a/stdlib/os/__init__.pyi
+++ b/stdlib/os/__init__.pyi
@@ -308,7 +308,7 @@ if sys.platform != "win32":
     EX_NOPERM: int
     EX_CONFIG: int
 
-if sys.platform != "win32" and sys.platform != "darwin":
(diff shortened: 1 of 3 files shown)


stdlib/os/__init__.pyi
```diff
@@ -308,7 +308,7 @@ if sys.platform != "win32":
     EX_NOPERM: int
     EX_CONFIG: int
 
-if sys.platform != "win32" and sys.platform != "darwin":
+if sys.platform != "win32" and sys.platform != "darwin" and sys.platform != "linux":
```
srittau, line 311:
```suggestion
# Exists on some Unix platforms, e.g. Solaris.
if sys.platform != "win32" and sys.platform != "darwin" and sys.platform != "linux":
```

srittau (changes requested):
Thanks! Could you add a comment like my suggestion?

srittau (approved):

PR #11463, opened 2024-02-23
python-dateutil: respect inheritance on relativedelta.__rsub__ (#11462)

### LINKED ISSUE #11462: Relativedelta `__rsub__` doesn't respect subclasses of datetime
Given this sample case:

```python
from datetime import datetime
from dateutil.relativedelta import relativedelta

class MyDateTime(datetime):
    pass

d = MyDateTime.now()
x = d - relativedelta(days=1)
print(type(x))
# reveal_type(x)
```
results in Python:

```
<class '__main__.MyDateTime'>
```

and mypy reports:

```
test.py:10: note: Revealed type is "datetime.datetime"
```

However, I'd expect `MyDateTime` as type here. I'm guessing the generic isn't [correct](https://github.com/python/typeshed/blob/a2095002e446bd0e20f8a469a14e271cd6cc6ad9/stubs/python-dateutil/dateutil/relativedelta.pyi#L8C1-L8C43) here?
With understanding of `TypeVar` and it's effects on inheritance. Should it be adjusted to the following?

```python
_DateT = TypeVar("_DateT", bound=date)
```
Given that `datetime` inherits from `date` already?

MaicoTimmerman:
Support inheritance for `relativedelta.__rsub__` by setting the `TypeVar` upper bound to `datetime.date` (parent of `datetime.datetime`), instead of having both those as constraint.

Resolved #11462


srittau:
C implementation here: https://github.com/python/cpython/blob/015b97d19a24a169cc3c0939119e1228791e4253/Modules/_datetimemodule.c#L3220-L3245

srittau:
Please ignore my comments about `date.__sub__`. The stubs, the C, and Python implementation all override it in `datetime.__sub__`. Our stubs shouldn't match:

```python
class datetime(date):
    # ...
    def __sub__(self, __value: Self) -> timedelta: ...
    @overload
    def __sub__(self, __value: timedelta) -> Self: ...
```

MaicoTimmerman:
> The reason the tests fail is that date.__sub__ has an overload def __sub__(self, __value: Self) -> timedelta: ..., which of course matches the tested line and therefore returns a timedelta.

`__value: Self` would match if the RHS is a (subclass of) `date`, right? But in this case it's `dateutil.relativedelta.relativedelta(days=1)`, which doesn't match. So I think there is some other reason here.

Oddly enough, I wasn't able to reproduce the issue locally using pyright==1.1.342 yet.

srittau:
I'm not sure what's going on. Maybe @erictraut has an idea as it seems that only pyright fails while mypy accepts the tests.

erictraut:
I'm not able to repro this issue with the latest version of pyright (1.1.351). To try to repro, I manually applied the change in `relativedelta.pyi` and then verified that the test code in this PR passes without any type errors.

I also tried to repro it with 1.1.342, which is the version of pyright that typeshed is currently pinned to. So I'm not sure what's happening here.

MaicoTimmerman:
@srittau anything I can do to push this forward? The failures persist with the re-run. It almost seems like the test is running without the updated stubs? That was exactly the case I'm trying to fix in this MR.

srittau:
I'm still stumped what the problem here is. I have the same results as @erictraut when testing this PR locally. It really behaves as if the changes to the python-dateutil stubs were ignored.

PR #11465, opened 2024-02-23
Improve JSON typing in requests.requests.models

AdemOdza:
Added a dict[str, Any] type to the Response json() method and Request.

The Request class expects a serializable object (https://github.com/psf/requests/blob/main/src/requests/models.py#L510)

The Response.json() method returns a deserialized JSON string (https://github.com/psf/requests/blob/main/src/requests/models.py#L958)

This was causing issues with my type checker (MyPy) so I thought I would add it in.
```python
def test(response: Response) -> dict[str, Any]:
  res = response.json()
  return res
  # MyPy complains because this function returns an Any object rather than a python 
  # JSON object
```

### DIFF
--- a/stubs/requests/requests/models.pyi
+++ b/stubs/requests/requests/models.pyi
@@ -61,7 +61,7 @@ class Request(RequestHooksMixin):
     headers: Any
     files: Any
     data: Any
-    json: Any
+    json: dict[str, Any]
     params: Any
     auth: Any
     cookies: Any
@@ -157,7 +157,7 @@ class Response:
         parse_constant: Callable[[str], Any] | None = ...,
         object_pairs_hook: Callable[[list[tuple[Any, Any]]], Any] | None = ...,
         **kwds: Any,
-    ) -> Any: ...
+    ) -> dict[str, Any]: ...
     @property
     def links(self) -> dict[Any, Any]: ...
     def raise_for_status(self) -> None: ...


Akuli:
The `.json()` method doesn't necessarily return a dict:

```python
>>> import requests
>>> requests.get('https://httpbin.org/base64/WzEsMiwzXQ==', data='hey').json()
[1, 2, 3]
```

AdemOdza:
> The `.json()` method doesn't necessarily return a dict:
> 
> ```python
> >>> import requests
> >>> requests.get('https://httpbin.org/base64/WzEsMiwzXQ==').json()
> [1, 2, 3]
> ```

Oh right, of course. Would there be a better way to type this other than Any?

Akuli:
Unfortunately, `Any` is the best we can do now. If we return `dict[str, Any] | something`, mypy will *still* complain when you try to use the return value as a list, because it thinks the return value could be a `dict[str, Any]` and you're not handling that. Ideally we would have something like [AnyOf](https://github.com/python/typing/issues/566) to work around this.

For your use case, you can `return cast(dict[str, Any], response.json())`.

Akuli:
Maybe adding `assert isinstance(res, dict)` is better than a cast:

```python
from requests import Response
from typing import Any

def test(response: Response) -> dict[str, Any]:
    res = response.json()
    assert isinstance(res, dict)
    return res
```

AdemOdza:
Thanks for the suggestions! I'll go ahead and close this PR

PR #11468, opened 2024-02-25
improve type annotations in 'docutils.node.document'

danieleades:
this is based on the type annotations in https://github.com/tk0miya/docutils-stubs, with a few tweaks


stubs/docutils/docutils/nodes.pyi
```diff
@@ -178,11 +179,66 @@ class Labeled: ...
...
 
+_Document: TypeAlias = document
+_Decoration: TypeAlias = decoration
+
 class document(Root, Structural, Element):
-    parent: None
-    transformer: Transformer
-    def __init__(self, settings, reporter: Reporter, rawsource: str = "", *children: Node, **attributes) -> None: ...
-    def __getattr__(self, __name: str) -> Incomplete: ...
+    current_source: str = ...
```
srittau, line 186: Not a full review, but please drop the `= ...` from class attributes. Our tests didn't catch that due to PyCQA/flake8-pyi#98.

danieleades: i've removed the `= ...` from class attributes. Looks like each of the method arguments also has a similar pattern. Are these required? (i've largely lifted this from another source so i'm just cargo-culting these changes)

srittau: `= ...` is required for method arguments that have a default that we can't represent in the stubs. If it is representable (e.g. if the default is a simple value like `None`), we use that instead.

danieleades: ah perfect. have updated

stubs/docutils/docutils/nodes.pyi
```diff
@@ -178,11 +179,66 @@ class Labeled: ...
...
+    citation_refs: dict[str, list[citation_reference]]
+    autofootnotes: list[footnote]
+    autofootnote_refs: list[footnote_reference]
+    symbol_footnotes: list[footnote]
+    symbol_footnote_refs: list[footnote_reference]
+    footnotes: list[footnote]
+    citations: list[citation]
+    autofootnote_start: int
+    symbol_footnote_start: int
+    id_start: int
```
srittau, line 208: In the current docutils source I see `id_counter`, but not `id_start`: https://github.com/docutils/docutils/blob/b768e2626088711dec257b0847b563d02700a712/docutils/docutils/nodes.py#L1301

stubs/docutils/docutils/nodes.pyi
```diff
@@ -178,11 +179,66 @@ class Labeled: ...
...
+    parse_messages: list[system_message]
+    transform_messages: list[system_message]
     transformer: Transformer
-    def __init__(self, settings, reporter: Reporter, rawsource: str = "", *children: Node, **attributes) -> None: ...
-    def __getattr__(self, __name: str) -> Incomplete: ...
+    decoration: decoration
```
srittau, line 212: Also initialized to `None`:

```suggestion
    decoration: decoration | None
```

stubs/docutils/docutils/nodes.pyi
```diff
@@ -178,11 +179,66 @@ class Labeled: ...
...
+    document: document
```
srittau, line 213: I'm not sure whether this works, but could you try using `Self` here?

stubs/docutils/docutils/nodes.pyi
```diff
@@ -178,11 +179,66 @@ class Labeled: ...
...
+    def __init__(self, settings: Values, reporter: Reporter, *args: Node, **kwargs: Any) -> None: ...
+    def asdom(self, dom: Any = None) -> Any: ...
```
srittau, line 215: Functionally the same, but since `None` is explicitly checked it might be nice to document that `None` is a valid argument:

```suggestion
    def asdom(self, dom: Any | None = None) -> Any: ...
```

stubs/docutils/docutils/nodes.pyi
```diff
@@ -178,11 +179,66 @@ class Labeled: ...
...
+    def note_footnote_ref(self, ref: footnote_reference) -> None: ...
+    def note_citation(self, citation: citation) -> None: ...
+    def note_citation_ref(self, ref: citation_reference) -> None: ...
+    def note_substitution_def(self, subdef: substitution_definition, def_name: str, msgnode: Element | None = None) -> None: ...
+    def note_substitution_ref(self, subref: substitution_reference, refname: str) -> None: ...
+    def note_pending(self, pending: pending, priority: int | None = None) -> None: ...
+    def note_parse_message(self, message: system_message) -> None: ...
+    def note_transform_message(self, message: system_message) -> None: ...
+    def note_source(self, source: str, offset: int) -> None: ...
+    def copy(self) -> _Document: ...
```
srittau, line 240:
```suggestion
    def copy(self) -> Self: ...
```

srittau (changes requested):
Thanks! This is a big improvement as the primer output shows. A few notes, though.

stubs/docutils/docutils/nodes.pyi
srittau, line 213:
```suggestion
    document: Self
```

danieleades: have added 'Self'. let's see what happens.

Does that impact compatibility for older interpreters?

srittau: This should work fine on all current type checkers, no matter the Python version.

srittau (approved):

PR #11469, opened 2024-02-26
improve type annotations in 'docutils.statemachine'

danieleades:
adds type annotations for docutils.statemachine.

I'm not super confident about the use of the generic `_Context` for the state machine since the only code examples I can find using this class are in the `sphinx` repo, and there no context is used (ever).


danieleades:
@srittau I could really use your eyes on this one! I think i've made good progress, but could use some assistance to get this over the line, if you have the time

danieleades:
i tried to simplify some of these type signatures using type aliases, but couldn't get this to work

```python
_Context = TypeVar("_Context")
_TransitionResult: TypeAlias = tuple[_Context, str | None, list[str]]
_TransitionMethod: TypeAlias = Callable[[Match[str], _Context, str], _TransitionResult[_Context]]
```

mypy and pyright complain about the generics being unbound

JelleZijlstra:
> i tried to simplify some of these type signatures using type aliases, but couldn't get this to work
> 
> ```python
> _Context = TypeVar("_Context")
> _TransitionResult: TypeAlias = tuple[_Context, str | None, list[str]]
> _TransitionMethod: TypeAlias = Callable[[Match[str], _Context, str], _TransitionResult[_Context]]
> ```
> 
> mypy and pyright complain about the generics being unbound

That looks right, but you should use the alias as `_TransitionResult[_Context]` in this case.

danieleades:
> That looks right, but you should use the alias as `_TransitionResult[_Context]` in this case.

i've reinstated the aliases to show what i mean. Neither mypy nor pyright is happy about them

JelleZijlstra:
```
stubs/docutils/docutils/statemachine.pyi:3: error: Module "typing" has no attribute "TypeAlias"  [attr-defined]
stubs/docutils/docutils/statemachine.pyi:3: note: Use `from typing_extensions import TypeAlias` instead
```

This probably has something to do with the other errors.

danieleades:
@JelleZijlstra anything else needed to get this one merged? (it's blocking a couple of other PRs)

danieleades:
Note i've requested reviews from some Sphinx maintainers since this 'context' generic type will propagate through a great many docutils types and have a significant impact on downstream users of these stubs

JelleZijlstra:
I haven't had a chance to look at this in much detail. The generics are weird to me because each class just seems to use the TypeVar as a type argument to other generics, until you get to `_TransitionResult`. You say that:

> I'm not super confident about the use of the generic `_Context` for the state machine since the only code examples I can find using this class are in the `sphinx` repo, and there no context is used (ever).

Assuming that Sphinx is a typical user of docutils, that might mean making these classes generic is going to be more of an annoyance than a help to many users. Therefore maybe we shouldn't be making these generic, at least until we're able to use PEP-696. (Which might not be far away, as the PEP has been accepted and the upcoming mypy release adds some support.)

danieleades:
> I haven't had a chance to look at this in much detail. The generics are weird to me because each class just seems to use the TypeVar as a type argument to other generics, until you get to `_TransitionResult`. You say that:
> 
> > I'm not super confident about the use of the generic `_Context` for the state machine since the only code examples I can find using this class are in the `sphinx` repo, and there no context is used (ever).
> 
> Assuming that Sphinx is a typical user of docutils, that might mean making these classes generic is going to be more of an annoyance than a help to many users. Therefore maybe we shouldn't be making these generic, at least until we're able to use PEP-696. (Which might not be far away, as the PEP has been accepted and the upcoming mypy release adds some support.)

I was looking at this for this PR actually.

Thing is, since i've not been able to find any uses of the 'context', i'm not sure what the default type should be...

JelleZijlstra:
> Thing is, since i've not been able to find any uses of the 'context', i'm not sure what the default type should be...

The default could be `object`, so users who don't use it just get an opaque object.

danieleades:
> > Thing is, since i've not been able to find any uses of the 'context', i'm not sure what the default type should be...
> 
> The default could be `object`, so users who don't use it just get an opaque object.

i've elected to remove the generics for now to unblock this, and will revisit in a separate PR

danieleades:
replaced `Any` with `list[str]` for context based on this comment - https://github.com/sphinx-doc/sphinx/issues/12007#issuecomment-1975377720

stubs/docutils/docutils/statemachine.pyi
```diff
@@ -1,3 +1,194 @@
...
+_TransitionMethod: TypeAlias = Callable[[Match[str], list[str], str], _TransitionResult]
+
+class StateMachine:
+    input_lines: StringList | None
+    input_offset: int
+    line: str | None
+    line_offset: int
+    debug: bool
+    initial_state: str
+    current_state: list[State]
```
picnixz, line 16:
```suggestion
    current_state: str
```

stubs/docutils/docutils/statemachine.pyi
```diff
@@ -1,3 +1,194 @@
-from _typeshed import Incomplete
+from collections.abc import Callable, Generator, Iterator, Sequence
+from re import Match, Pattern
+from typing import Any, Generic, TypeVar, overload
+from typing_extensions import Self, TypeAlias
 
-def __getattr__(name: str) -> Incomplete: ...
+_TransitionResult: TypeAlias = tuple[list[str], str | None, list[str]]
```
picnixz, line 6: I think it should be fine to have an alias for the StateName so that we know that `str | None` is a state name or None.

stubs/docutils/docutils/statemachine.pyi
```diff
@@ -1,3 +1,194 @@
...
+    def next_line(self, n: int = 1) -> str: ...
+    def is_next_line_blank(self) -> bool: ...
+    def at_eof(self) -> bool: ...
+    def at_bof(self) -> bool: ...
+    def previous_line(self, n: int = 1) -> str: ...
+    def goto_line(self, line_offset: int) -> None: ...
+    def get_source(self, line_offset: int) -> str: ...
+    def abs_line_offset(self) -> int: ...
+    def abs_line_number(self) -> int: ...
+    def get_source_and_line(self, lineno: int | None = None) -> tuple[str, int]: ...
```
picnixz, line 39:
```suggestion
    def get_source_and_line(self, lineno: int | None = None) -> tuple[str, int] | tuple[None, None]: ...
```

danieleades:
> I'll review more when I'm again in another train but here are some comments based on the docutils HEAD implementation. Can you also check whether they are consistent with the stable version?
> 
> Btw, sorry for the order of my comments but I tend to review things in a random order, and comment when I have an idea (I would have thought that gh would have ordered the comments from top to bottom but it appears not).

Thank you, I appreciate the detailed review. Will go through it when I get time

stubs/docutils/docutils/statemachine.pyi
danieleades: do you mean this?

```python
_StateName: TypeAlias = str | None
_TransitionResult: TypeAlias = tuple[list[str], _StateName, list[str]]
_TransitionMethod: TypeAlias = Callable[[Match[str], list[str], str], _TransitionResult]
```

stubs/docutils/docutils/statemachine.pyi
```diff
@@ -1,3 +1,194 @@
...
+class StateMachine:
+    input_lines: StringList | None
+    input_offset: int
+    line: str | None
+    line_offset: int
+    debug: bool
+    initial_state: str
+    current_state: list[State]
+    states: dict[str, State]
+    observers: list[Callable[[Self], None]]
```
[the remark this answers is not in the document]
danieleades: oh confusing.

so based on your description it's not actually `Callable[[str | None, int | None], None]`

it's

`Callable[[str, int], None]` or
`Callable[[str, None], None]` or
`Callable[[None, None], None]` but *not*
`Callable[[None, int], None]`

@JelleZijlstra is there a compact way to annotate that?

stubs/docutils/docutils/statemachine.pyi
```diff
@@ -1,3 +1,194 @@
...
+    def abs_line_number(self) -> int: ...
+    def get_source_and_line(self, lineno: int | None = None) -> tuple[str, int]: ...
+    def insert_input(self, input_lines: list[str] | StringList, source: str) -> None: ...
+    def get_text_block(self, flush_left: bool = False) -> StringList: ...
+    def check_line(self, context: list[str], state: State, transitions: list[str] | None = ...) -> _TransitionResult: ...
+    def add_state(self, state_class: type[State]) -> None: ...
+    def add_states(self, state_classes: Sequence[type[State]]) -> None: ...
+    def runtime_init(self) -> None: ...
+    def error(self) -> None: ...
+    def attach_observer(self, observer) -> None: ...
```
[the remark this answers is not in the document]
danieleades: makes sense. I'll wait for some agreement on https://github.com/python/typeshed/pull/11469/files#r1511283606

stubs/docutils/docutils/statemachine.pyi
```diff
@@ -1,3 +1,194 @@
...
+    parent_offset: int
+    def __init__(
+        self,
+        initlist: ViewList[_Data] | list[_Data] | None = None,
+        source: str | None = None,
+        items: list[tuple[str, int]] | None = None,
+        parent: ViewList[_Data] | None = None,
+        parent_offset: int | None = None,
+    ) -> None: ...
+    def __lt__(self, other: Any) -> bool: ...
```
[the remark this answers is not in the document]
danieleades: > I think you should use __other so that it indicates (without enforcing) that you expect positional only arguments. I don't know whether you can use positional-only arguments syntax or not.

i haven't noticed that pattern elsewhere in typeshed. I'd like to follow convention here. @JelleZijlstra?

stubs/docutils/docutils/statemachine.pyi
```diff
@@ -1,3 +1,194 @@
...
+    def __le__(self, other: Any) -> bool: ...
+    def __eq__(self, other: object) -> bool: ...
+    def __ne__(self, other: object) -> bool: ...
+    def __gt__(self, other: Any) -> bool: ...
+    def __ge__(self, other: Any) -> bool: ...
+    def __contains__(self, item: ViewList[_Data]) -> bool: ...
```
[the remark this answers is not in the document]
danieleades: this is the implementation:

```python
    def __lt__(self, other): return self.data < self.__cast(other)   # noqa
    def __le__(self, other): return self.data <= self.__cast(other)  # noqa
    def __eq__(self, other): return self.data == self.__cast(other)  # noqa
    def __ne__(self, other): return self.data != self.__cast(other)  # noqa
    def __gt__(self, other): return self.data > self.__cast(other)   # noqa
    def __ge__(self, other): return self.data >= self.__cast(other)  # noqa
```

stubs/docutils/docutils/statemachine.pyi
```diff
@@ -1,3 +1,194 @@
...
+    @overload
+    def __getitem__(self, i: int) -> _Data: ...
+    @overload
+    def __setitem__(self, i: slice, item: ViewList[_Data]) -> None: ...
+    @overload
+    def __setitem__(self, i: int, item: _Data) -> None: ...
+    def __delitem__(self, i: int) -> None: ...
+    def __add__(self, other: ViewList[_Data]) -> None: ...
+    def __radd__(self, other: ViewList[_Data]) -> None: ...
+    def __iadd__(self, other: Self) -> Self: ...
```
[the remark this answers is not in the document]
danieleades: certainly appears to:

```python
    def __mul__(self, n):
        return self.__class__(self.data * n, items=(self.items * n))
```

stubs/docutils/docutils/statemachine.pyi
```diff
@@ -1,3 +1,194 @@
...
+    def sort(self, *args: tuple[_Data, tuple[str, int]]) -> None: ...
+    def info(self, i: int) -> tuple[str, int]: ...
+    def source(self, i: int) -> str: ...
+    def offset(self, i: int) -> int: ...
+    def disconnect(self) -> None: ...
+    def xitems(self) -> Generator[tuple[str, int, str], None, None]: ...
+    def pprint(self) -> None: ...
+
+    # dummy atribute to indicate to mypy that ViewList is Iterable[str]
+    def __iter__(self) -> Iterator[str]: ...
```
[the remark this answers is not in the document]
danieleades: see https://github.com/python/mypy/issues/2220

stubs/docutils/docutils/statemachine.pyi
danieleades: done

stubs/docutils/docutils/statemachine.pyi
```diff
@@ -1,3 +1,194 @@
...
+        input_source: str | None = None,
+        initial_state: str | None = None,
+    ) -> list[str]: ...
+    def get_state(self, next_state: str | None = None) -> State: ...
+    def next_line(self, n: int = 1) -> str: ...
+    def is_next_line_blank(self) -> bool: ...
+    def at_eof(self) -> bool: ...
+    def at_bof(self) -> bool: ...
+    def previous_line(self, n: int = 1) -> str: ...
+    def goto_line(self, line_offset: int) -> None: ...
```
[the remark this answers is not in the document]
danieleades: done

stubs/docutils/docutils/statemachine.pyi
```diff
@@ -1,3 +1,194 @@
...
+        context: list[str] | None = None,
+        input_source: str | None = None,
+        initial_state: str | None = None,
+    ) -> list[str]: ...
+    def get_state(self, next_state: str | None = None) -> State: ...
+    def next_line(self, n: int = 1) -> str: ...
+    def is_next_line_blank(self) -> bool: ...
+    def at_eof(self) -> bool: ...
+    def at_bof(self) -> bool: ...
+    def previous_line(self, n: int = 1) -> str: ...
```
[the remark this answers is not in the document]
danieleades: done

stubs/docutils/docutils/statemachine.pyi
danieleades: done

stubs/docutils/docutils/statemachine.pyi
picnixz: Ah yes, it might be redundant but it might help people I think. At least you know what this str is

stubs/docutils/docutils/statemachine.pyi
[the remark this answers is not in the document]
picnixz: IIRC the `__cast` function is the identity on non-ViewList objects and returns other.data otherwise I think. 

So the expected argument is either a ViewList or a list of data (I think).

stubs/docutils/docutils/statemachine.pyi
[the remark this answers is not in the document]
picnixz: Actually your callable should support those kind of inputs except for the combination (None, int), but I don't think we can type it like that unless we have an intersection type. So we could technically allow `Callable[[str | None, int | None], None]` even though it would never be called with (None, int) :/

stubs/docutils/docutils/statemachine.pyi
```diff
@@ -1,3 +1,194 @@
...
+    @overload
+    def __setitem__(self, i: slice, item: ViewList[_Data]) -> None: ...
+    @overload
+    def __setitem__(self, i: int, item: _Data) -> None: ...
+    def __delitem__(self, i: int) -> None: ...
+    def __add__(self, other: ViewList[_Data]) -> None: ...
+    def __radd__(self, other: ViewList[_Data]) -> None: ...
+    def __iadd__(self, other: Self) -> Self: ...
+    def __mul__(self, n: int) -> ViewList[_Data]: ...
+    __rmul__: Any
```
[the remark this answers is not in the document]
danieleades: good catch

stubs/docutils/docutils/statemachine.pyi
```diff
@@ -1,3 +1,194 @@
...
+
+class _SearchOverride:
+    def match(self, pattern: Pattern[str]) -> Match[str]: ...
+
+class SearchStateMachine(_SearchOverride, StateMachine): ...
+class SearchStateMachineWS(_SearchOverride, StateMachineWS): ...
+
+_Data = TypeVar("_Data")
+
+class ViewList(Generic[_Data]):
```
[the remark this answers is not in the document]
danieleades: feels weird to inherit from a base class that adds mixin methods in a typestub

stubs/docutils/docutils/statemachine.pyi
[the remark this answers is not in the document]
picnixz: It's actually incorrect because it means that a function that only accept None and None is allowed but then it cannot process the normal inputs being str and int.

danieleades:
@srittau have you got any bandwidth to help me get this one over the line? A few of the other PRs are dependent on this one.

the state machine classes actually appear to be generic over the 'context' parameter, however i've only been able to find subclasses in actual use where the context is `list[str]`, so making the base class generic is possibly of limited value. Personally i'm actually in favour of making the base class generic but removing the generic is probably the more pragmatic solution. as @JelleZijlstra pointed out, a default generic would be a perfect solution.

stubs/docutils/docutils/statemachine.pyi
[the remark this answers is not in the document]
JelleZijlstra: We should use PEP 570 pos-only syntax in new code, not double underscores. (We had some technical reasons why we didn't migrate all existing code yet.) Here I would make the parameter positional-only if it's also positional-only at runtime. If the class is implemented in Python, it probably isn't.

stubs/docutils/docutils/statemachine.pyi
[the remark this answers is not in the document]
JelleZijlstra: No need to add the inheritance unless the runtime also inherits from these classes.

stubs/docutils/docutils/statemachine.pyi
```diff
@@ -1,3 +1,194 @@
...
+    input_lines: StringList | None
+    input_offset: int
+    line: str | None
+    line_offset: int
+    debug: bool
+    initial_state: str
+    current_state: str
+    states: dict[str, State]
+    observers: list[Callable[[Self], None]]
+    def __init__(self, state_classes: Sequence[type[State]], initial_state: str, debug: bool = False) -> None: ...
```
srittau, line 19:
```suggestion
    def __init__(self, state_classes: Iterable[type[State]], initial_state: str, debug: bool = False) -> None: ...
```

stubs/docutils/docutils/statemachine.pyi
```diff
@@ -1,3 +1,194 @@
...
+    def unlink(self) -> None: ...
+    def run(
+        self,
+        input_lines: list[str] | StringList,
```
srittau, line 23:
```suggestion
        input_lines: Sequence[str] | StringList,
```

stubs/docutils/docutils/statemachine.pyi
```diff
@@ -1,3 +1,194 @@
...
+        input_offset: int = 0,
+        context: list[str] | None = None,
```
srittau, line 25: It says in the docstring that context is a "application-specific storage", and I only see it passed to states. Would it make more sense to use `Any` or even a type alias `_Context: TypeAlias = Any`?

stubs/docutils/docutils/statemachine.pyi
```diff
@@ -1,3 +1,194 @@
...
+    def goto_line(self, line_offset: int) -> str | None: ...
+    def get_source(self, line_offset: int) -> str: ...
+    def abs_line_offset(self) -> int: ...
+    def abs_line_number(self) -> int: ...
+    def get_source_and_line(self, lineno: int | None = None) -> tuple[str, int] | tuple[None, None]: ...
+    def insert_input(self, input_lines: list[str] | StringList, source: str) -> None: ...
+    def get_text_block(self, flush_left: bool = False) -> StringList: ...
+    def check_line(self, context: list[str], state: State, transitions: list[str] | None = ...) -> _TransitionResult: ...
+    def add_state(self, state_class: type[State]) -> None: ...
+    def add_states(self, state_classes: Sequence[type[State]]) -> None: ...
```
srittau, line 44:
```suggestion
    def add_states(self, state_classes: Iterable[type[State]]) -> None: ...
```

stubs/docutils/docutils/statemachine.pyi
```diff
@@ -1,3 +1,194 @@
...
+    def error(self) -> None: ...
+    def attach_observer(self, observer) -> None: ...
+    def detach_observer(self, observer) -> None: ...
+    def notify_observers(self) -> None: ...
+
+class State:
+    patterns: dict[str, str | Pattern[str]] | None
+    initial_transitions: Sequence[str] | Sequence[tuple[str, str]] | None
+    nested_sm: type[StateMachine] | None
+    nested_sm_kwargs: dict[str, Any] | None
```
srittau, line 55: Considering that these are always filled with not-None values in `__init__`, I'd skip the `| None` part. Might need a stubtest override.

stubs/docutils/docutils/statemachine.pyi
```diff
@@ -1,3 +1,194 @@
...
+    def add_states(self, state_classes: Sequence[type[State]]) -> None: ...
+    def runtime_init(self) -> None: ...
+    def error(self) -> None: ...
+    def attach_observer(self, observer) -> None: ...
+    def detach_observer(self, observer) -> None: ...
+    def notify_observers(self) -> None: ...
+
+class State:
+    patterns: dict[str, str | Pattern[str]] | None
+    initial_transitions: Sequence[str] | Sequence[tuple[str, str]] | None
```
srittau, line 53: According to the docstring:

```suggestion
    patterns: ClassVar[dict[str, str | Pattern[str]] | None]
    initial_transitions: ClassVar[Sequence[str] | Sequence[tuple[str, str]] | None]
```

stubs/docutils/docutils/statemachine.pyi
```diff
@@ -1,3 +1,194 @@
...
+    nested_sm_kwargs: dict[str, Any] | None
+    transition_order: list[str]
+    transitions: dict[str, tuple[Pattern[str], Callable[[], None], str]]
+    state_machine: StateMachine
+    debug: bool
+    def __init__(self, state_machine: StateMachine, debug: bool = False) -> None: ...
+    def runtime_init(self) -> None: ...
+    def unlink(self) -> None: ...
+    def add_initial_transitions(self) -> None: ...
+    def add_transitions(self, names: list[str], transitions) -> None: ...
```
srittau, line 64:
```suggestion
    def add_transitions(self, names: Iterable[str], transitions) -> None: ...
```

stubs/docutils/docutils/statemachine.pyi
```diff
@@ -1,3 +1,194 @@
...
+    def detach_observer(self, observer) -> None: ...
+    def notify_observers(self) -> None: ...
+
+class State:
+    patterns: dict[str, str | Pattern[str]] | None
+    initial_transitions: Sequence[str] | Sequence[tuple[str, str]] | None
+    nested_sm: type[StateMachine] | None
+    nested_sm_kwargs: dict[str, Any] | None
+    transition_order: list[str]
+    transitions: dict[str, tuple[Pattern[str], Callable[[], None], str]]
```
srittau, line 57: Considering that a transition is a fairly complex type and it's used multiple times in this class, I'd consider creating a type alias.

stubs/docutils/docutils/statemachine.pyi
```diff
@@ -1,3 +1,194 @@
...
+    state_machine: StateMachine
+    debug: bool
+    def __init__(self, state_machine: StateMachine, debug: bool = False) -> None: ...
+    def runtime_init(self) -> None: ...
+    def unlink(self) -> None: ...
+    def add_initial_transitions(self) -> None: ...
+    def add_transitions(self, names: list[str], transitions) -> None: ...
+    def add_transition(self, name: str, transition: tuple[Pattern[str], str, str]) -> None: ...
```
srittau, line 65: Shouldn't the transition type match the others in this class?

stubs/docutils/docutils/statemachine.pyi
```diff
@@ -1,3 +1,194 @@
...
+    def nop(self, match: Match[str], context: list[str], next_state: str) -> _TransitionResult: ...
+
+class StateMachineWS(StateMachine):
+    def get_indented(self, until_blank: bool = False, strip_indent: bool = True) -> tuple[StringList, int, int, bool]: ...
+    def get_known_indented(
+        self, indent: int, until_blank: bool = False, strip_indent: bool = True
+    ) -> tuple[list[str], int, bool]: ...
+    def get_first_known_indented(
+        self, indent: int, until_blank: bool = False, strip_indent: bool = True, strip_top: bool = True
+    ) -> tuple[list[str], int, int, bool]: ...
```
srittau, line 85: Shouldn't the first return type in all of these be consistently either `list[str]` or `StringList`?

stubs/docutils/docutils/statemachine.pyi
```diff
@@ -1,3 +1,194 @@
...
+
+class StateWS(State):
+    indent_sm: type[StateMachine] | None
+    indent_sm_kwargs: dict[str, Any] | None
+    known_indent_sm: type[StateMachine] | None
+    known_indent_sm_kwargs: dict[str, Any] | None
```
srittau, line 91: As above:

```suggestion
    indent_sm: type[StateMachine]
    indent_sm_kwargs: dict[str, Any]
    known_indent_sm: type[StateMachine]
    known_indent_sm_kwargs: dict[str, Any]
```

stubs/docutils/docutils/statemachine.pyi
```diff
@@ -1,3 +1,194 @@
...
+    ws_patterns: dict[str, str]
```
srittau, line 92:
```suggestion
    ws_patterns: dict[str, Pattern[str]]
```

stubs/docutils/docutils/statemachine.pyi
```diff
@@ -1,3 +1,194 @@
...
+    def known_indent(self, match: Match[str], context: list[str], next_state: str) -> _TransitionResult: ...
+    def first_known_indent(self, match: Match[str], context: list[str], next_state: str) -> _TransitionResult: ...
+
+class _SearchOverride:
+    def match(self, pattern: Pattern[str]) -> Match[str]: ...
+
+class SearchStateMachine(_SearchOverride, StateMachine): ...
+class SearchStateMachineWS(_SearchOverride, StateMachineWS): ...
+
+_Data = TypeVar("_Data")
```
srittau, line 107: Nit: I'd use `_T` as name, similar to most collection types in typeshed are typed. An move the type var to the top of the file, seeing as how generic it is.

stubs/docutils/docutils/statemachine.pyi
```diff
@@ -1,3 +1,194 @@
...
+
+class ViewList(Generic[_Data]):
+    data: list[_Data]
+    items: list[tuple[str, int]]
+    parent: ViewList[_Data]
+    parent_offset: int
+    def __init__(
+        self,
+        initlist: ViewList[_Data] | list[_Data] | None = None,
```
srittau, line 116:
```suggestion
        initlist: ViewList[_Data] | Sequence[_Data] | None = None,
```

stubs/docutils/docutils/statemachine.pyi
```diff
@@ -1,3 +1,194 @@
...
+        parent: ViewList[_Data] | None = None,
+        parent_offset: int | None = None,
+    ) -> None: ...
+    def __lt__(self, other: Any) -> bool: ...
+    def __le__(self, other: Any) -> bool: ...
+    def __eq__(self, other: object) -> bool: ...
+    def __ne__(self, other: object) -> bool: ...
+    def __gt__(self, other: Any) -> bool: ...
+    def __ge__(self, other: Any) -> bool: ...
+    def __contains__(self, item: ViewList[_Data]) -> bool: ...
```
srittau, line 128: Either:

```suggestion
    def __contains__(self, item: _Data) -> bool: ...
```

or

```suggestion
    def __contains__(self, item: object) -> bool: ...
```

(I think that's what @picnixz meant.)

stubs/docutils/docutils/statemachine.pyi
```diff
@@ -1,3 +1,194 @@
...
+    def __len__(self) -> int: ...
+    @overload
+    def __getitem__(self, i: slice) -> ViewList[_Data]: ...
```
srittau, line 131:
```suggestion
    def __getitem__(self, i: slice) -> Self[_Data]: ...
```

stubs/docutils/docutils/statemachine.pyi
```diff
@@ -1,3 +1,194 @@
...
+    @overload
+    def __getitem__(self, i: int) -> _Data: ...
```
srittau, line 133: Per @picnixz:

```suggestion
    def __getitem__(self, i: SupportsIndex) -> _Data: ...
```

stubs/docutils/docutils/statemachine.pyi
```diff
@@ -1,3 +1,194 @@
...
+    @overload
+    def __setitem__(self, i: slice, item: ViewList[_Data]) -> None: ...
+    @overload
+    def __setitem__(self, i: int, item: _Data) -> None: ...
+    def __delitem__(self, i: int) -> None: ...
```
srittau, line 138:
```suggestion
    def __setitem__(self, i: SupportsIndex, item: _Data) -> None: ...
    def __delitem__(self, i: SupportsIndex) -> None: ...
```

stubs/docutils/docutils/statemachine.pyi
```diff
@@ -1,3 +1,194 @@
...
+    def __add__(self, other: ViewList[_Data]) -> None: ...
+    def __radd__(self, other: ViewList[_Data]) -> None: ...
+    def __iadd__(self, other: Self) -> Self: ...
+    def __mul__(self, n: int) -> ViewList[_Data]: ...
+    def __rmul__(self, n: int) -> ViewList[_Data]: ...
+    def __imul__(self, n: int) -> Self: ...
```
srittau, line 144:
```suggestion
    def __add__(self, other: ViewList[_Data]) -> Self[_Data]: ...
    def __radd__(self, other: ViewList[_Data]) -> Self[_Data]: ...
    def __iadd__(self, other: ViewList[_Data]) -> Self[_Data]: ...
    def __mul__(self, n: int) -> Self[_Data]: ...
    __rmul__ = __mul__
    def __imul__(self, n: int) -> Self[_Data]: ...
```

srittau (changes requested):
Thanks, remarks below.

stubs/docutils/docutils/statemachine.pyi
AlexWaygood: `Self` can't be parameterised

srittau: Good point, I fixed it in the comment.

stubs/docutils/docutils/statemachine.pyi
danieleades: Seems reasonable, though there's a possibility it might be set to None in a subclass and not initialised in __init__

stubs/docutils/docutils/statemachine.pyi
danieleades: i did have it set up this way at one point. see https://github.com/python/typeshed/pull/11469#issuecomment-1975938313 and https://github.com/python/typeshed/pull/11469#issuecomment-1986788144

i could reinstate the generics

stubs/docutils/docutils/statemachine.pyi
srittau: I don't know the intricacies of docutils. If that's a possibility, we should keep it as is.

stubs/docutils/docutils/statemachine.pyi
danieleades: it isn't used the same way twice at the moment...

`initial_transitions: ClassVar[Sequence[str] | Sequence[tuple[str, str]] | None]`
`transitions: dict[str, tuple[Pattern[str], Callable[[], None], str]]`
`make_transition(...) -> tuple[Pattern[str], _TransitionMethod[_Context], str]`

stubs/docutils/docutils/statemachine.pyi
srittau: Ok, sounds good then. You might still consider a type var (to `list[str]`) for documentation purposes, but I leave this decision to you.

stubs/docutils/docutils/statemachine.pyi
```diff
@@ -1,3 +1,199 @@
...
+        parent: Self | None = None,
+        parent_offset: int | None = None,
+    ) -> None: ...
+    def __lt__(self, other: Any) -> bool: ...
+    def __le__(self, other: Any) -> bool: ...
+    def __eq__(self, other: object) -> bool: ...
+    def __ne__(self, other: object) -> bool: ...
+    def __gt__(self, other: Any) -> bool: ...
+    def __ge__(self, other: Any) -> bool: ...
+    def __contains__(self, item: Self) -> bool: ...
```
srittau, line 133: `item` should either be `_T` or `object`. The implementation does just `item in self.data`, where `self.data` is `list[_T]`.

srittau (changes requested):
I spotted one thing. But looks good to go apart from that.

srittau (approved):

PR #11471, opened 2024-02-27
improve type annotations in 'docutils.transforms'

danieleades:


JelleZijlstra:
Could you comment on whether the new errors shown in mypy-primer are true positives?

danieleades:
> Could you comment on whether the new errors shown in mypy-primer are true positives?

i'll need to revisit this after https://github.com/sphinx-doc/sphinx/pull/12034 is merged.

the `NodeMatcher` class used by Sphinx confounds the type checking of the `Node.find_all` method, which results in a lot (though not all) of the errors seen here

danieleades:
> ```diff
> + sphinx/transforms/i18n.py:102:43: error: Only concrete class can be given where "type[translatable]" is expected  [type-abstract]
> ```

i don't really understand this one. the snippet in sphinx is

```python
        for node in self.document.findall(addnodes.translatable):
            node.preserve_original_messages()
```

which seems ok, no? `addnodes.translatable` is indeed a class

danieleades:
> 
> ```diff
> - sphinx/transforms/__init__.py: note: In member "apply_transforms" of class "SphinxTransformer":
> + sphinx/transforms/__init__.py: note: In member "apply" of class "ExtraTranslatableNodes":
> - sphinx/transforms/__init__.py:83:13: error: "apply_transforms" undefined in superclass  [misc]

should be fixed in https://github.com/python/typeshed/pull/11471

> + sphinx/transforms/__init__.py:273:9: error: Incompatible types in assignment (expression has type "Node", variable has type "Element")  [assignment]

this will have to be suppressed in sphinx. the implementation is not friendly to type checking

> + sphinx/transforms/i18n.py: note: In member "apply" of class "PreserveTranslatableMessages":
> + sphinx/transforms/i18n.py:102:43: error: Only concrete class can be given where "type[translatable]" is expected  [type-abstract]

i don't understand why this is causing an error

> - sphinx/builders/latex/transforms.py: note: In member "run" of class "BibliographyTransform":
> - sphinx/builders/latex/transforms.py:525:13: error: "BibliographyTransform" has no attribute "document"  [attr-defined]

should be fixed in https://github.com/python/typeshed/pull/11471

> 
> alectryon (https://github.com/cpitclaudel/alectryon)
> + alectryon/docutils.py:206: error: "Transform" has no attribute "is_post_transform"  [attr-defined]

should be fixed in https://github.com/python/typeshed/pull/11471

> ```

danieleades:
marking as a draft until https://github.com/python/typeshed/pull/11471 is merged

stubs/docutils/docutils/transforms/__init__.pyi
```diff
@@ -1,14 +1,31 @@
...
+    sorted: bool
+    components: dict[str, Component]
+    serialno: int
+    def __init__(self, document: nodes.document) -> None: ...
     def add_transform(self, transform_class: type[Transform], priority: int | None = None, **kwargs) -> None: ...
-    def __getattr__(self, __name: str) -> Incomplete: ...
-
-def __getattr__(name: str) -> Incomplete: ...
+    def add_transforms(self, transform_list: list[type[Transform]]) -> None: ...
+    def add_pending(self, pending: pending, priority: int | None = None) -> None: ...
```
picnixz, line 28: I would suggest using `nodes.NODE_TYPE` because it's clearer rather than just `NODE_TYPE`.

stubs/docutils/docutils/transforms/__init__.pyi
```diff
@@ -1,14 +1,31 @@
...
+    applied: list[tuple[int, type[Transform], pending, Any]]
+    sorted: bool
+    components: dict[str, Component]
+    serialno: int
+    def __init__(self, document: nodes.document) -> None: ...
     def add_transform(self, transform_class: type[Transform], priority: int | None = None, **kwargs) -> None: ...
-    def __getattr__(self, __name: str) -> Incomplete: ...
-
-def __getattr__(name: str) -> Incomplete: ...
+    def add_transforms(self, transform_list: list[type[Transform]]) -> None: ...
```
picnixz, line 27: Is a list require or is an iterable sufficient?

stubs/docutils/docutils/transforms/__init__.pyi
```diff
@@ -1,14 +1,31 @@
...
+    language: ModuleType
+    def __init__(self, document: nodes.document, startnode: Node | None = None) -> None: ...
+    def apply(self, **kwargs: Any) -> None: ...
 
-class Transformer:
-    def __init__(self, document: document): ...
+class Transformer(TransformSpec):
+    transforms: list[tuple[str, type[Transform], Node | None]]
+    document: nodes.document
+    applied: list[tuple[int, type[Transform], pending, Any]]
```
picnixz, line 21: Same comment as above

picnixz (commented):
A few comments based on what I remember from memory

stubs/docutils/docutils/transforms/__init__.pyi
danieleades: have updated, though i've made an exception for `Node` because i think that's quite clear, and `nodes.Node` feels tautological

danieleades:
> 
> ```diff
> - sphinx/transforms/__init__.py: note: In member "apply_transforms" of class "SphinxTransformer":
> + sphinx/transforms/__init__.py: note: In member "apply" of class "ExtraTranslatableNodes":
> - sphinx/transforms/__init__.py:83:13: error: "apply_transforms" undefined in superclass  [misc]
> + sphinx/transforms/__init__.py:273:9: error: Incompatible types in assignment (expression has type "Node", variable has type "Element")  [assignment]
> + sphinx/transforms/i18n.py: note: In member "apply" of class "PreserveTranslatableMessages":
> + sphinx/transforms/i18n.py:102:43: error: Only concrete class can be given where "type[translatable]" is expected  [type-abstract]

I don't understand this error

> - sphinx/builders/latex/transforms.py: note: In member "run" of class "BibliographyTransform":
> - sphinx/builders/latex/transforms.py:525:13: error: "BibliographyTransform" has no attribute "document"  [attr-defined]
> 
> alectryon (https://github.com/cpitclaudel/alectryon)
> + alectryon/docutils.py:206: error: "Transform" has no attribute "is_post_transform"  [attr-defined]
> ```

picnixz:
The alceytron error is probably because of the new attribute being set though I don't know why it would raise (maybe because it's untyped?).

As for the error on the Sphinx side, I don't know why it would raise an error for that transform but not for others (maybe is it because of the `+=` that we use?)

stubs/docutils/docutils/transforms/__init__.pyi
```diff
@@ -1,14 +1,31 @@
...
+
+class TransformError(ApplicationError): ...
 
 class Transform:
-    def __init__(self, document: document, startnode: Node | None = None): ...
-    def __getattr__(self, name: str, /) -> Incomplete: ...
+    default_priority: int | None
+    document: nodes.document
+    startnode: Node | None
+    language: ModuleType
```
picnixz, line 14: It's more than just a module, it's a module with specific attributes. It's more important to have the protocol rather than the module type IMO.

The protocol is:

```
class _LanguageModule(Protocol):
    labels: dict[str, str]
    author_separators: list[str]
    bibliographic_fields: list[str]
```

stubs/docutils/docutils/transforms/__init__.pyi
```diff
@@ -1,14 +1,31 @@
...
+    def __init__(self, document: nodes.document, startnode: Node | None = None) -> None: ...
+    def apply(self, **kwargs: Any) -> None: ...
```
picnixz, line 16:
```suggestion
	@abc.abstractmethod
    def apply(self, **kwargs: Any) -> None: ...
```

stubs/docutils/docutils/transforms/__init__.pyi
```diff
@@ -1,14 +1,31 @@
...
+from types import ModuleType
+from typing import Any
 
-from docutils.nodes import Node, document
+from docutils import ApplicationError, Component, TransformSpec, nodes
+from docutils.nodes import Node
+
+class TransformError(ApplicationError): ...
 
 class Transform:
```
picnixz, line 10:
```suggestion
class Transform(ABC):
```

I'm not sure whether this should really be done here because the implementation does not explicitly inherit from ABC. However, `apply` raise `NotImplementedError`, so it *is* an abstract class per se. However, I'm not sure if (runtime) subclasses would possibly be marked as conflicting if they have a different metaclass (that is not compatible with `ABCMeta`, e.g., it does not directly inherit from it (at runtime, again)).

stubs/docutils/docutils/transforms/__init__.pyi
```diff
@@ -1,14 +1,31 @@
-from _typeshed import Incomplete
+from collections.abc import Iterable
```
picnixz, line 1:
```suggestion
from abc import ABC
from collections.abc import Iterable
```

stubs/docutils/docutils/transforms/__init__.pyi
```diff
@@ -1,14 +1,31 @@
...
+    document: nodes.document
+    startnode: Node | None
+    language: ModuleType
+    def __init__(self, document: nodes.document, startnode: Node | None = None) -> None: ...
+    def apply(self, **kwargs: Any) -> None: ...
 
-class Transformer:
-    def __init__(self, document: document): ...
+class Transformer(TransformSpec):
+    transforms: list[tuple[str, type[Transform], Node | None]]
```
picnixz, line 19:
```suggestion
    transforms: list[tuple[str, type[Transform], nodes.pending | None]]
```

stubs/docutils/docutils/transforms/__init__.pyi
```diff
@@ -1,14 +1,31 @@
...
+    document: nodes.document
+    applied: list[tuple[int, type[Transform], nodes.pending, Any]]
```
picnixz, line 21:
```suggestion
    applied: list[tuple[int, type[Transform], nodes.pending | None, Any]]
```

stubs/docutils/docutils/transforms/__init__.pyi
```diff
@@ -1,14 +1,31 @@
...
+    components: dict[str, Component]
+    serialno: int
+    def __init__(self, document: nodes.document) -> None: ...
     def add_transform(self, transform_class: type[Transform], priority: int | None = None, **kwargs) -> None: ...
-    def __getattr__(self, name: str, /) -> Incomplete: ...
-
-def __getattr__(name: str) -> Incomplete: ...
+    def add_transforms(self, transform_list: Iterable[type[Transform]]) -> None: ...
+    def add_pending(self, pending: nodes.pending, priority: int | None = None) -> None: ...
+    def get_priority_string(self, priority: int) -> str: ...
```
picnixz, line 29: At runtime, you *should* not have priority = None but with the implementation you could have it. The reason is that `default_priority` of a transformation class must be overridden in subclasses but it is *not* enforced.

You somehow enforce the fact that the transformation class lacks an explicit `default_priority` value as follows:

- Use `add_trasnform(t)` without *priority*.
- The implementation uses ``p = t.default_priority``. It is `None` but directly calls `get_priority_string(p)`.
- Format ``'%03d-%03d' % (priority, self.serialno)``. Since None is not '%d'-valid, you get an exception. 

Forcing `int` as the argument type would make docutils fail on that. However, using `int | None` to make it pass could make people assume that a None priority is still allowed. Would it make sense to have something like:

```suggestion
    @overload
    def get_priority_string(self, priority: None) -> NoReturn: ...
    @overload
    def get_priority_string(self, priority: int) -> str: ...
```
?

stubs/docutils/docutils/transforms/__init__.pyi
```diff
@@ -1,14 +1,31 @@
...
-    def __init__(self, document: document): ...
+class Transformer(TransformSpec):
+    transforms: list[tuple[str, type[Transform], Node | None]]
+    document: nodes.document
+    applied: list[tuple[int, type[Transform], nodes.pending, Any]]
+    sorted: bool
+    components: dict[str, Component]
+    serialno: int
+    def __init__(self, document: nodes.document) -> None: ...
     def add_transform(self, transform_class: type[Transform], priority: int | None = None, **kwargs) -> None: ...
```
picnixz, line 26:
```suggestion
    def add_transform(self, transform_class: type[Transform], priority: int | None = None, **kwargs: Any) -> None: ...
```

stubs/docutils/docutils/__init__.pyi
```diff
@@ -34,7 +37,7 @@ class SettingsSpec:
 class TransformSpec:
     def get_transforms(self) -> list[Any]: ...
     default_transforms: ClassVar[tuple[Any, ...]]
-    unknown_reference_resolvers: ClassVar[list[Any]]
+    unknown_reference_resolvers: ClassVar[Sequence[Callable[[Node], bool]]]
 
 class Component(SettingsSpec, TransformSpec):
     component_type: ClassVar[str | None]
```
picnixz, line 43:
```suggestion
    component_type: ClassVar[ComponentType | None]
```

stubs/docutils/docutils/transforms/__init__.pyi
```diff
@@ -1,14 +1,31 @@
...
+    def apply(self, **kwargs: Any) -> None: ...
 
-class Transformer:
-    def __init__(self, document: document): ...
+class Transformer(TransformSpec):
+    transforms: list[tuple[str, type[Transform], Node | None]]
+    document: nodes.document
+    applied: list[tuple[int, type[Transform], nodes.pending, Any]]
+    sorted: bool
+    components: dict[str, Component]
```
picnixz, line 23:
```suggestion
    components: dict[ComponentType, Component]
```

stubs/docutils/docutils/__init__.pyi
```diff
@@ -34,7 +37,7 @@ class SettingsSpec:
 class TransformSpec:
     def get_transforms(self) -> list[Any]: ...
     default_transforms: ClassVar[tuple[Any, ...]]
-    unknown_reference_resolvers: ClassVar[list[Any]]
+    unknown_reference_resolvers: ClassVar[Sequence[Callable[[Node], bool]]]
```
picnixz, line 41:
```suggestion

ComponentType = Literal['reader', 'parser', 'writer']

```

picnixz (commented):
Sorry for having completely forgotten about that PR.

Here are some comments. There are one thing I am worried about is whether we should make the stubs good for docutils or for downstream repos instead. What I mean is that there are a lot of attributes in abstract classes that are initialized to `None` and where concrete subclasses are expected to override them with a non-None value.

For instance, `Component.component_type` should be a string at runtime and is used as `self.components[component.component_type] = ...` in `docutils.transforms.Transformer.populate_from_components`. If we type `components` as a `dict[str, Component]` instead of `dict[str | None, Component]`, then the above assignment would fail the type-check.

However, at runtime, it is expected to have non-None values so... I could alternatively suggest that we only make abstract classes "bad" classes in docutils, namely, even if you have a `None` default value, ignore it and force, for instance, `Component.component_type` to be a string.

stubs/docutils/docutils/transforms/__init__.pyi
danieleades: @JelleZijlstra? what's the correct idiom here?

Akuli: We could use `@abstractmethod` without inheriting from ABC, so that at least mypy will complain if it isn't overridden in a subclass. This is documented in [these mypy docs](https://mypy.readthedocs.io/en/stable/class_basics.html#abstract-base-classes-and-multiple-inheritance).

stubs/docutils/docutils/transforms/__init__.pyi
danieleades: i wish we could just refactor docutils...

stubs/docutils/docutils/transforms/__init__.pyi
picnixz: Oh I wasn't aware of that. Let's go for that then !

stubs/docutils/docutils/transforms/__init__.pyi
picnixz: Since we can safely use abstractmethod without inheriting from ABC, we should mark the method as abstract (it raises a NotImplementedError at runtime).

stubs/docutils/docutils/transforms/__init__.pyi
picnixz: Here, you really want to allow None nodes (that's their "default" values when adding a new transformation).

stubs/docutils/docutils/transforms/__init__.pyi
picnixz: That's unfortunately a separate issue. Maybe a -1 would have been a better default priority.

picnixz (commented):
I think I reviewed but it appears that what I said was treated as comments.

stubs/docutils/docutils/transforms/__init__.pyi
danieleades: could we mark it as an `@abstractproperty` and then suppress the stubcheck error?

```python
    @property
    @classmethod
    @abstractmethod
    def default_priority(cls) -> int: ...
```

trouble is, a get a lint warning that `classmethod` is deprecated in python 3.11 and removed in python 3.13. not sure what the alternative is?

JelleZijlstra:
I got CI to pass but some of @picnixz's review has not been addressed.

JelleZijlstra:
@danieleades you moved this back to draft, anything that still needs to get done?

danieleades:
> @danieleades you moved this back to draft, anything that still needs to get done?

still a few mypy primer errors to resolve, though i'm not sure how to make progress

JelleZijlstra:
I took a quick look at a few of them, but as you're more familiar with Sphinx/Docutils, you may be able to make more progress. It's OK if this introduces some new errors in Sphinx if they're correct, but the Sphinx errors may tell us if some of the new stubs you're adding have bugs.

> + sphinx/transforms/__init__.py:94: error: Unused "type: ignore" comment  [unused-ignore]

Seems like Sphinx got an error here because the transforms module was incompletely typed, so good that the error went away.

> + sphinx/transforms/__init__.py:296:13: error: Unsupported target for indexed assignment ("Node")  [index]

This is in a subclass of Transformer that does:

```
        for node in self.document.findall(is_translatable_node):
            node['translatable'] = True
```

Seems like either some stub is wrong, or that Sphinx code is broken. I haven't looked further to figure out which of the two it is.

> + sphinx/transforms/i18n.py:110:43: error: Only concrete class can be given where "type[translatable]" is expected  [type-abstract]

This is a questionable mypy error. The stub is probably right, sphinx will just have to ignore the error.

danieleades:
abandoning as out of date

PR #11479, opened 2024-02-27
Add stubs for lzstring

Akuli:
Adds stubs for the [lzstring](https://pypi.org/project/lzstring/) package.

LZString is a compression format supported by many programming languages. I'm using a Python LZString implementation to interact with a project written in JavaScript.

The python `lzstring` package hasn't been updated in a long time, but:
- it is the first result when searching for a Python LZString implementation
- it works
- it is small (implementation is 430 lines, API is tiny)
- it probably won't need any updating any time soon, because LZString compression format is not new or changing.


stubs/lzstring/lzstring/__init__.pyi
```diff
@@ -0,0 +1,17 @@
...
+    @staticmethod
+    def compressToEncodedURIComponent(uncompressed: str) -> str: ...
+    @staticmethod
+    def decompress(compressed: str) -> str: ...
+    @staticmethod
+    def decompressFromUTF16(compressed: str) -> str: ...
+    @staticmethod
+    def decompressFromBase64(compressed: str) -> str: ...
+    @staticmethod
+    def decompressFromEncodedURIComponent(compressed: str) -> str: ...
```
AlexWaygood, line 17: The runtime implementations for all of these seem to explicitly check to see if `None` was passed in; maybe we should allow that? https://github.com/gkovacs/lz-string-python/blob/916183026bdc80c33913f97b4927bd20df109987/lzstring/__init__.py#L371-L430

```suggestion
    def compress(uncompressed: str | None) -> str: ...
    @staticmethod
    def compressToUTF16(uncompressed: str | None) -> str: ...
    @staticmethod
    def compressToBase64(uncompressed: str | None) -> str: ...
    @staticmethod
    def compressToEncodedURIComponent(uncompressed: str | None) -> str: ...
    @staticmethod
    def decompress(compressed: str | None) -> str: ...
    @staticmethod
    def decompressFromUTF16(compressed: str | None) -> str: ...
    @staticmethod
    def decompressFromBase64(compressed: str | None) -> str: ...
    @staticmethod
    def decompressFromEncodedURIComponent(compressed: str | None) -> str: ...
```

Akuli: The decompress methods can also return None:

```
>>> LZString.decompress('foo bar') is None
True
```

AlexWaygood (approved):

PR #11481, opened 2024-02-27
improve type annotations in 'docutils.parsers.rst.roles'

danieleades:


stubs/docutils/docutils/parsers/rst/roles.pyi
```diff
@@ -1,20 +1,131 @@
...
 ]
 
+def register_canonical_role(name: str, role_fn: _RoleFn) -> None: ...
 def register_local_role(name: str, role_fn: _RoleFn) -> None: ...
 def role(
     role_name: str, language_module: _LanguageModule, lineno: int, reporter: Reporter
 ) -> tuple[_RoleFn | None, list[SystemMessage]]: ...
-def __getattr__(name: str) -> Incomplete: ...
+def set_implicit_options(role_fn: _RoleFn) -> None: ...
+def register_generic_role(canonical_name: str, node_class: Node) -> None: ...
```
JelleZijlstra, line 25:
```suggestion
def register_generic_role(canonical_name: str, node_class: type[Node]) -> None: ...
```
The name suggests this takes a class, not an instance.

stubs/docutils/docutils/parsers/rst/roles.pyi
```diff
@@ -1,20 +1,131 @@
...
+
+class GenericRole:
+    name: str
+    node_class: Node
```
JelleZijlstra, line 29:
```suggestion
    node_class: type[Node]
```

stubs/docutils/docutils/parsers/rst/roles.pyi
```diff
@@ -1,20 +1,131 @@
...
+    def __init__(self, role_name: str, node_class: type[TextElement]) -> None: ...
```
JelleZijlstra, line 30:
```suggestion
    def __init__(self, role_name: str, node_class: type[Node]) -> None: ...
```
Or change to TextElement above; it should be consistent.

stubs/docutils/docutils/parsers/rst/roles.pyi
```diff
@@ -1,20 +1,131 @@
...
+    def __call__(
+        self,
+        role: str,
+        rawtext: str,
+        text: str,
+        lineno: int,
+        inliner: Inliner,
+        options: dict[str, Any] | None = None,
```
JelleZijlstra, line 38: Prefer using `Mapping` if possible. Similarly, instead of `list`, it's probably better to use `Sequence` for many of these.

danieleades: i've replaced most of them. I kept a couple in return value position, and in 'normalized_role_options' since it returns a dict, but may forward the input options directly

stubs/docutils/docutils/parsers/rst/roles.pyi
danieleades: done. I've gone with `Node`, although looking at the code i suspect `TextElement` may be more accurate. thoughts?

stubs/docutils/docutils/parsers/rst/roles.pyi
JelleZijlstra: Oh I see, sorry. Argument annotations in callbacks are one place where it's good to be as narrow as possible, because that means more callbacks are accepted (since they behave contravariantly). I'm guessing the options dict gets passed to the RoleFn callbacks? It might be best to stick with the dict/list types to make sphinx's life simpler.

PR #11486, opened 2024-02-28
Add missing constants to pyOpenSSL

### LINKED ISSUE #11456: pyOpenSSL stubs are missing types for DTLS constants
Pretty straight forward.

https://github.com/pyca/pyopenssl/blob/a972ea4a5df700d60a5b59f7990952baf0bb3974/src/OpenSSL/SSL.py#L162-L164

They're there in the actual code, but missing from the stubs.

Sachaa-Thanasius:
This adds the `DTLS_METHOD`, `DTLS_SERVER_METHOD`, and `DTLS_CLIENT_METHOD` int constants to the pyOpenSSL stubs.

Constants in the actual code: [here](https://github.com/pyca/pyopenssl/blob/a972ea4a5df700d60a5b59f7990952baf0bb3974/src/OpenSSL/SSL.py#L162-L164)

Fixes #11456.


JelleZijlstra:
That's some nice mypy-primer output! Looks good, I can merge it if you take it out of draft.

Sachaa-Thanasius:
Sure thing. That first error isn't a problem?

JelleZijlstra:
That line means your change removed an error; i.e., mypy showed this (false positive) error without your change, but no longer shows it with your change.

PR #11487, opened 2024-02-28
[bs4] Make `Tag.attrs` more permissive

Sachaa-Thanasius:
`Tag.attrs` is a dict can technically have `list[str]` as a valid value in the event one queries it for "class" — i.e. it should be more than just `str`. Making the value type a union with `Any` allows `list[str]` to be valid in a type checker's eyes while still favoring `str`.

Based on discussion in, and fixes the issue mentioned in, #8755.


JelleZijlstra:
Thanks!

PR #11488, opened 2024-02-28
Add `@deprecated` to `abstractclass/staticmethod/property`

Viicos:
There might be a reason it wasn't added yet, but couldn't find anything when searching.


AlexWaygood (approved):
Thanks!

PR #11489, opened 2024-02-28
stdlib: Add Shutdown error and shutdown method to queue module

aisk:
Added in https://github.com/python/cpython/pull/104750


stdlib/queue.pyi
```diff
@@ -29,6 +34,9 @@ class Queue(Generic[_T]):
     def full(self) -> bool: ...
     def get(self, block: bool = True, timeout: float | None = None) -> _T: ...
     def get_nowait(self) -> _T: ...
+    if sys.version_info >= (3, 13):
+        def shutdown(self, immediate: bool = False): ...
```
JelleZijlstra, line 38:
```suggestion
        def shutdown(self, immediate: bool = False) -> None: ...
```

JelleZijlstra (approved):

Issue #11484, opened 2024-02-27
Set up automation for updating pytype and pyright versions


JelleZijlstra:
Both pytype and pyright release pretty frequently and it's useful for us to make sure we test against the latest version. It would be nice to have a GitHub action that automatically updates them for us so we don't have to remember to make a PR. Example PRs: #11483, #11190.

I suppose other dependencies could also be auto-updated but the need isn't as great.

AlexWaygood:
We could probably do this fairly easily with dependabot. We wouldn't need to do it with all our dependencies if we don't want to; but also, dependabot now supports setting dependency groups that are all updated together, so it wouldn't be especially noisy to do it for all our dependencies

JelleZijlstra:
I think I'd probably want new versions of mypy, pyright, and pytype to trigger an immediate (daily) update PR, and everything else can get updated every month.

Avasam:
I don't know if dependabot can detect our deduplicated pyright version though.
If it's automatically handled, it would be fine to have it duplicated again, but even then I'm not sure if the version given as an argument to `pyright-action` will be picked up.

JelleZijlstra:
Sounds great, thank you! We'll review your PR soon.

AlexWaygood:
We now have four potential solutions prototyped:
1. Carry on pinning dependencies the way we do now, and use a custom script to auto-update our dependencies: https://github.com/python/typeshed/pull/11491
2. Move our pyright pin to a `package.json` file and use dependabot to auto-update our dependencies: https://github.com/python/typeshed/pull/11564
3. Move our pyright pin to `requirements.txt` and use dependabot to auto-update our dependencies: https://github.com/python/typeshed/pull/11575
4. Carry on pinning dependencies the way we do now, and use renovate to auto-update our dependencies: https://github.com/python/typeshed/pull/11565

I think all four proposals have their pros and cons. Of the four, I like (4) best, then (3). I like (4) because:
- The configuration for renovate wasn't as complex as I feared it might be
- It looks like somebody successfully authorised renovate to run on this repository, which I was worried might be a blocker (https://github.com/python/typeshed/pull/11568)
- We can use quarterly schedules for dependencies we don't care about updating that regularly; the longest period dependabot allows is monthly
- We can have different schedules for different groups of pip dependencies; dependabot only allows us to set one schedule per ecosystem.

[...]

PR #11491, opened 2024-02-28
Set up automation for updating pytype and pyright versions.

Mr-Sunglasses:
fix: #11484


scripts/check_for_update_dependency.py
```diff
@@ -0,0 +1,63 @@
...
+    with open(file_path, "r", encoding="utf-8") as file:
+        config = toml.load(file)
+
+    latest_version = get_latest_version("pyright")
+
+    if latest_version:
+        config["tool"]["typeshed"]["pyright_version"] = latest_version
+
+        with open(file_path, "w", encoding="utf-8") as file:
+            toml.dump(config, file)
```
Akuli, line 52: This will delete all comments from the file and change its formatting a lot. Maybe just search for a line that starts with `"pyright_version = "` and replace the rest of it?

.github/workflows/update_daily_ dependency.yml
```diff
@@ -0,0 +1,34 @@
+name: Check Dependencies
```
Akuli, line 1: - The name of this `.yml` file has a space in it.
- Maybe instead call it something like `Update mypy and pyright` (and rename the `.yml` file accordingly)?

.github/workflows/update_daily_ dependency.yml
```diff
@@ -0,0 +1,34 @@
...
+                  python-version: "3.11"
+            - name: Install dependencies
+              run: |
+                  pip install requests
+                  pip install toml
+            - name: Run script
+              run: python scripts/check_for_update_dependency.py
+            - uses: peter-evans/create-pull-request@v3
+              with:
+                token: ${{ secrets.PAT }}
```
Akuli, line 28: Maybe:

```suggestion
                token: ${{ secrets.GITHUB_TOKEN }}
```

.github/workflows/update_daily_ dependency.yml
```diff
@@ -0,0 +1,34 @@
...
+                title: "Update dpendency Daily"
+                commit-message: "Updated Depenceny after the daily check"
+                body: |
+                    There appear to be some dependency updates in ${{ github.sha }}. This pull request
+                    uses the script to find and update these dependency.
+                base: ${{ github.head_ref }}
```
Akuli, line 34: This will need a few things that our other scripts have:
- `create-issue-on-failure` part at the end
- `permissions` sections

Feel free to copy/paste from `stubsabot.yml`.

scripts/check_for_update_dependency.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+def get_latest_version(package_name: str) -> Optional[str]:
+    """Fetch the latest version of a package from PyPI."""
+    response = requests.get(f"https://pypi.org/pypi/{package_name}/json")
+    if response.status_code == 200:
+        data = response.json()
+        return str(data["info"]["version"])
+    else:
+        return None
```
Akuli, line 16: Why should we fail silently? IMO it would be better to fail with an error, so that the script will get fixed. You can use `response.raise_for_status()` or `assert response.status_code == 200`, for example.

Akuli (commented):
Thanks! Before merging this, I will figure out why the CI is failing on this pull request, and test this in my fork.

.github/workflows/update_daily_ dependency.yml
```diff
@@ -0,0 +1,34 @@
...
+        steps:
+            - uses: actions/checkout@v2
+            - name: Set up Python
+              uses: actions/setup-python@v4
+              with:
+                  python-version: "3.11"
+            - name: Install dependencies
+              run: |
+                  pip install requests
+                  pip install toml
```
AlexWaygood, line 23: We should use `tomli`, which is better maintained, rather than `toml`

scripts/check_for_update_dependency.py
AlexWaygood: `tomlkit` is a style-preserving and comment-preserving library for writing TOML that we could use as an alternative to `toml` here. We use it in scripts/stubsabot.py. But I'd also be okay with a simple find-and-replace solution like @Akuli suggests.

.github/workflows/update_daily_ dependency.yml
```diff
@@ -0,0 +1,34 @@
...
+            - name: Run script
+              run: python scripts/check_for_update_dependency.py
+            - uses: peter-evans/create-pull-request@v3
+              with:
+                token: ${{ secrets.PAT }}
+                title: "Update dpendency Daily"
+                commit-message: "Updated Depenceny after the daily check"
+                body: |
+                    There appear to be some dependency updates in ${{ github.sha }}. This pull request
+                    uses the script to find and update these dependency.
```
AlexWaygood, line 33:
```suggestion
                title: "Daily dependencies update"
                commit-message: "Updated dependencies after the daily check"
                body: |
                    There appear to be some dependency updates in ${{ github.sha }}. This pull request
                    uses the script to find and update these dependencies.
```

JelleZijlstra:
Did you consider setting up a Dependabot configuration instead? That would be preferable since it means we can rely on a well-tested, widely used tool instead of having to write the script ourselves.

AlexWaygood:
> Did you consider setting up a Dependabot configuration instead? That would be preferable since it means we can rely on a well-tested, widely used tool instead of having to write the script ourselves.

We can't do that for pyright because of the unique way we pin pyright in pyproject.toml. Perhaps we could do a handrolled solution for pyright and do dependabot for everything else, though

Avasam:
About dependabot, I'll repost this conversation: https://github.com/jakebailey/pyright-action/issues/9#issuecomment-1629781059

Renovate could possibly do it. Otherwise it looks like it might be possible to [Contributing new ecosystems](https://github.com/dependabot/dependabot-core/blob/main/CONTRIBUTING.md#contributing-new-ecosystems)

There's also the option of using https://pypi.org/project/pyright/ and updating our scripts to get the version from `requirements.txt` rather than `pyproject.toml` (just like we do for mypy). That way dependabot can pick it up.

The sole reason the version is stored in `pyproject.toml` currently is to deduplicate it.

jakebailey:
You can pin pyright in a `package.json` and dependabot will update that for you; the work to pull it out of the JSON file in CI is about the same as doing TOML:

```
jq '.devDependencies.pyright' < ./path.to/package.json
```

AlexWaygood:
> Renovate could possibly do it. Otherwise it looks like it might be possible to [Contributing new ecosystems](https://github.com/dependabot/dependabot-core/blob/main/CONTRIBUTING.md#contributing-new-ecosystems)

I use renovate over at https://github.com/AlexWaygood/typeshed-stats and it works fine, though it's a bit of a pain to configure it correctly. I'd prefer dependabot if possible; @jakebailey's suggestion of reading the pyright version from a json file sounds great if it'll work with dependabot.

jakebailey:
For a more concrete example of how to do this in GHA: https://github.com/microsoft/TypeChat/blob/main/.github/workflows/ci.python.yml#L49

But if you find this too cumbersome, I feel like I should probably add something to the action that lets you give it a package.json to read.

scripts/check_for_update_dependency.py
Mr-Sunglasses: @Akuli @AlexWaygood I've followed the @Akuli sugesstion with a simple find-and-replace solution.

AlexWaygood (commented):
I still think a solution that uses dependabot would be simpler, easier to maintain, and less prone to bugs (since dependabot is widely used and well tested)

PR #11496, opened 2024-02-29
Replace Flake8 checks with Ruff (except for flake8-pyi)

Avasam:
I got curious to see if we could replace the rest of Flake8 (but keeping flake8-pyi and with it flake8-noqa) with Ruff.
I've also been less selective about `UP` and `RUF` rules.

This is the result.
The bigger challenges and possible roadbump I'd say are https://github.com/astral-sh/ruff/issues/3011 and https://github.com/plinss/flake8-noqa/issues/22 / https://github.com/plinss/flake8-noqa/pull/30

There's a couple things here that can be immediately useful that I'll split up into smaller PRs.


stubs/six/six/moves/builtins.pyi
```diff
@@ -1 +1,4 @@
-from builtins import *
+# flake8: noqa: NQA102 # https://github.com/plinss/flake8-noqa/issues/22
+# six explicitely re-exports builtins. Normally this is something we'd want to avoid.
```
Avasam, line 2:
```suggestion
# six explicitly re-exports builtins. Normally this is something we'd want to avoid.
```

stdlib/builtins.pyi
```diff
@@ -100,7 +100,7 @@ class object:
     def __class__(self) -> type[Self]: ...
     # Ignore errors about type mismatch between property getter and setter
```
Akuli, line 101: Maybe the comment can be deleted?

danieleades:
this seems pretty close

> warning: Invalid rule code provided to `# ruff: noqa` at stubs/six/six/moves/builtins.pyi:1: NQA102

https://docs.astral.sh/ruff/settings/#lint_external

Avasam:
> > warning: Invalid rule code provided to `# ruff: noqa` at stubs/six/six/moves/builtins.pyi:1: NQA102
> 
> [docs.astral.sh/ruff/settings/#lint_external](https://docs.astral.sh/ruff/settings/#lint_external)

NQA is already marked as external. I opened https://github.com/astral-sh/ruff/issues/10202

Avasam:
This will conflict with #11522 and some Ruff config changes don't make sense w/o the pre-commit changes, so we can't have this PR work standalone whilst avoiding conflicts. Keeping as draft for now for that reason.

pyproject.toml
```diff
@@ -32,13 +32,15 @@ exclude = [
 # are invoked via separate runs of ruff in pre-commit:
 # see our .pre-commit-config.yaml file for details
 exclude = ["**/test_cases/**/*.py"]
-select = [
+external = ["F821", "NQA", "Y"]
```
AlexWaygood, line 35: I think you might have dropped a version of this comment accidentally when merging in `main`:

```suggestion
# We still use flake8-pyi and flake8-noqa to check these (see .flake8 config file);
# tell ruff not to flag these as e.g. "unused noqa comments"
external = ["F821", "NQA", "Y"]
```

pyproject.toml
```diff
@@ -32,13 +32,15 @@ exclude = [
...
+extend-select = [
```
AlexWaygood, line 36: What effect does switching from `select` to `extend-select` have here?

pyproject.toml
```diff
@@ -51,15 +53,33 @@ select = [
...
     # B033 could be slightly useful but Ruff doesn't have per-file select
     "B", # flake8-bugbear
+    # Rules that are out of the control of stub authors:
+    "E501", # Line too long
+    "E741", # ambiguous variable name
+    "F403", # `from . import *` used; unable to detect undefined names
+    # False positives in stubs
+    "F821", # Undefined name: https://github.com/astral-sh/ruff/issues/3011
+    # FIXME: Fix these in stubs, then remove these entries
+    "F405", # may be undefined, or defined from star imports
```
Avasam, line 82: These appear in `stdlib\_pydecimal.pyi` and `stdlib\ast.pyi`
Is that a pattern we'd like to change as my comment implies here? Or disabled specifically in those two files. Or is using a name from a star import something we wanna continue doing in typeshed (I feel like explicit would be better than implicit, even if we're talking entire module re-exports: do the `*` import + explicitly import symbols used in the stub)

AlexWaygood (commented):
Very nice! One comment suggestion, and one question:

pyproject.toml
Avasam: > What effect does switching from `select` to `extend-select` have here?

https://docs.astral.sh/ruff/rules/
> By default, Ruff enables Flake8's F rules, along with a subset of the E rules, omitting any stylistic rules that overlap with the use of a formatter, like ruff format or [Black](https://github.com/psf/black).

https://docs.astral.sh/ruff/formatter/#conflicting-lint-rules
> None of the above are included in Ruff's default configuration. However, if you've enabled any of these rules or their parent categories (like Q), we recommend disabling them via the linter's [lint.ignore](https://docs.astral.sh/ruff/settings/#lint_ignore) setting.

AlexWaygood: That makes sense. Is `extend-select` still completely overridden if you specify `--select` on the command line, like we do in our pre-commit config?

Avasam: I wanted to say yes, but reading https://docs.astral.sh/ruff/linter/#rule-selection makes it sounds like `extend-select` is always applied after even a "highest priority `select`" (which is the CLI), then the `ignore`.

> To resolve the enabled rule set, Ruff may need to reconcile [lint.select](https://docs.astral.sh/ruff/settings/#lint_select) and [lint.ignore](https://docs.astral.sh/ruff/settings/#lint_ignore) from a variety of sources, including the current pyproject.toml, any inherited pyproject.toml files, and the CLI (e.g., [--select](https://docs.astral.sh/ruff/settings/#lint_select)).

> 
> In those scenarios, Ruff uses the "highest-priority" [select](https://docs.astral.sh/ruff/settings/#lint_select) as the basis for the rule set, and then applies [extend-select](https://docs.astral.sh/ruff/settings/#lint_extend-select) and [ignore](https://docs.astral.sh/ruff/settings/#lint_ignore) adjustments. CLI options are given higher priority than pyproject.toml options, and the current pyproject.toml file is given higher priority than any inherited pyproject.toml files.

There's also less formatter-conflicting rules than I remembered (at least when it comes to `F` and `E`. So I might just do that instead.

pyproject.toml
JelleZijlstra: I don't think F405 is useful in stubs, type checkers will catch anything it catches.

pyproject.toml
```diff
@@ -51,15 +59,42 @@ select = [
...
+    # see https://github.com/astral-sh/ruff/issues/6465
+    "E721", # Do not compare types, use `isinstance()`
+    ###
+    # False-positives, but already checked by type-checkers
+    ###
+    # Ruff doesn't support multi-file analysis yet: https://github.com/astral-sh/ruff/issues/5295
+    "RUF013", # PEP 484 prohibits implicit `Optional`
 ]
-
 [tool.ruff.lint.per-file-ignores]
```
Avasam, line 84:
```suggestion

[tool.ruff.lint.per-file-ignores]
```

AlexWaygood (approved):
LGTM!

AlexWaygood:
If we _did_ want to use `extend-select` rather than `select` in `pyproject.toml`, we could possibly do something like this:

```diff
diff --git a/.pre-commit-config.yaml b/.pre-commit-config.yaml
index dc4bc138b..9612c5417 100644
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -20,12 +20,12 @@ repos:
         # Run this separately because we don't really want
         # to use --unsafe-fixes for all rules
         name: Remove unnecessary `sys.version_info` blocks
-        args: ["--exit-non-zero-on-fix", "--select=UP036", "--unsafe-fixes"]
+        args: ["--isolated", "--target-version=py38", "--exit-non-zero-on-fix", "--select=UP036", "--fix", "--unsafe-fixes"]
       - id: ruff
         # Very few rules are useful to run on our test cases;
         # we explicitly enumerate them here:
         name: Run ruff on the test cases
-        args: ["--exit-non-zero-on-fix", "--select=FA,I", "--no-force-exclude", "--unsafe-fixes"]
+        args: ["--isolated", "--target-version=py38", "--exit-non-zero-on-fix", "--select=FA,I", "--no-force-exclude", "--fix", "--unsafe-fixes"]
         files: '.*test_cases/.+\.py$'
```

But I personally find this PR's current solution cleaner!

PR #11497, opened 2024-02-29
Fix a few invalid noqa comments

Avasam:
Extracted from #11496


stubs/WebOb/webob/exc.pyi
```diff
@@ -41,7 +41,7 @@ class WSGIHTTPException(Response, HTTPException):
     def generate_response(self, environ: WSGIEnvironment, start_response: StartResponse) -> Iterable[bytes]: ...
     @property
     def wsgi_response(self) -> Self: ...  # type:ignore[override]
```
JelleZijlstra, line 43:
```suggestion
    def wsgi_response(self) -> Self: ...  # type: ignore[override]
```

stubs/WebOb/webob/exc.pyi
```diff
@@ -41,7 +41,7 @@ class WSGIHTTPException(Response, HTTPException):
...
-    def __str__(self) -> str: ...  # type:ignore[override]  # noqaY029
+    def __str__(self) -> str: ...  # type:ignore[override]  # noqa: Y029
```
JelleZijlstra, line 44:
```suggestion
    def __str__(self) -> str: ...  # type: ignore[override]  # noqa: Y029
```

JelleZijlstra (commented):
Thanks, suggested two other changes while we're here

stubs/WebOb/webob/exc.pyi
Avasam: I've seen (and done) that often. Idk if it's just outside a formatter's purview (Ruff formatter/black). Maybe Ruff linting and/or flake8-pyi could check for it.

JelleZijlstra: Yes, this could be a good lint rule.

Avasam: There's 40 instances of `type:ignore` in the code. I could fix them all at once here if you'd like.

JelleZijlstra: Sure!

PR #11499, opened 2024-02-29
Enable nearly all pyupgrade rules (except on test cases)

Avasam:
There's more rules that we want than rules we don't want. And with Ruff having a distinction for "unsafe fixes" (not applied by default) the comment "Only enable rules that have safe autofixes" holds less weight.

Rules relevant to code changes:
https://docs.astral.sh/ruff/rules/lru-cache-without-parameters/
https://docs.astral.sh/ruff/rules/f-string/

Even the protobuf-specific ignore could be activated and autofixes run, as they have a redundant (for typeshed at least) Python >= 3.8 check. 

Extracted from #11496 , overlaps with #11498

xref https://github.com/plinss/flake8-noqa/issues/22 as with it there would be 1 less noqa


stubs/six/six/moves/builtins.pyi
```diff
@@ -1 +1,4 @@
-from builtins import *
+# flake8: noqa: NQA102 # https://github.com/plinss/flake8-noqa/issues/22
+# six explicitely re-exports builtins. Normally this is something we'd want to avoid.
```
Avasam, line 2:
```suggestion
# six explicitly re-exports builtins. Normally this is something we'd want to avoid.
```

pyproject.toml
```diff
@@ -39,14 +40,16 @@ select = [
...
-    "UP006", # PEP-585 autofixes
-    "UP007", # PEP-604 autofixes
-    "UP013", # Class-based syntax for TypedDicts
-    "UP014", # Class-based syntax for NamedTuples
-    "UP019", # Use str over typing.Text
-    "UP035", # import from typing, not typing_extensions, wherever possible
-    "UP039", # don't use parens after a class definition with no bases
+]
+ignore = [
+    # We like explicit open modes
```
Avasam, line 45: Is that even true? I know I do, because of the different open modes in different languages, one's expectation could be different.
The only place this is flagged in typeshed's codebase is in a stub test
https://github.com/python/typeshed/blob/5e9589dd75f8a5f39a38a2ed45f098ff62babc7f/stubs/requests/%40tests/test_cases/check_post.py#L46

AlexWaygood: I think it might be a good idea to exclude all test cases from these rules. Test cases often deliberately contain code that might not be considered idiomatic or modern; I think it would be confusing (especially to new contributors) to have that incorrectly autofixed

AlexWaygood:
Please also ignore UP038 due to https://github.com/astral-sh/ruff/issues/7871. It doesn't matter for us now, since it only triggers if the target version is set to py310 or higher, but I don't want a nasty pre-commit surprise when we drop support for Python 3.9 :)

Avasam:
> Please also ignore UP038 everywhere [...]

Forgot about that one, I even have it in my own personal projects' ignore

pyproject.toml
```diff
@@ -39,14 +40,21 @@ select = [
...
-    "UP035", # import from typing, not typing_extensions, wherever possible
-    "UP039", # don't use parens after a class definition with no bases
+]
+ignore = [
+    # Slower and more verbose https://github.com/astral-sh/ruff/issues/7871
+    "UP038", # Use `X | Y` in `isinstance` call instead of `(X, Y)`
+]
+
+[tool.ruff.lint.per-file-ignores]
+# Disable "mordernization" rules with autofixes from test cases as they often
```
AlexWaygood, line 50:
```suggestion
# Disable "modernization" rules with autofixes from test cases as they often
```

AlexWaygood (approved):
Thanks!

pyproject.toml
Avasam: I think I'm seriously starting to consider in-editor spellchecking

PR #11500, opened 2024-02-29
Replace flake8-bugbear with Ruff

Avasam:
Replaces flake8-bugbear with https://docs.astral.sh/ruff/rules/#flake8-bugbear-b
Extracted from #11496

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -17,7 +17,6 @@
 #     E701   Multiple statements on one line (colon) -- disallows "..." on the same line
 
 # Some rules are considered irrelevant to stub files:
-#     B      All flake8-bugbear rules are .py-specific
 #     F401   imported but unused -- does not recognize re-exports
 #            https://github.com/PyCQA/pyflakes/issues/474
 
@@ -30,15 +29,15 @@
 extend-ignore = A, D, N8, SIM, RST, TYP, E301, E302, E305, E501, E701
 per-file-ignores =
   *.py: E203
-  *.pyi: B, E741, F401, F403, F405
+  *.pyi: E741, F401, F403, F405
   # Since typing.pyi defines "overload" this is not recognized by Flake8 as typing.overload.
   # Unfortunately, Flake8 does not allow to "noqa" just a specific error inside the file itself.
   # https://github.com/PyCQA/flake8/issues/1079
   #     F811 redefinition of unused '...'
-  stdlib/typing.pyi: B, E741, F401, F403, F405, F811
+  stdlib/typing.pyi: E741, F401, F403, F405, F811
   # Generated protobuf files include docstrings,
   # and import some things from typing_extensions that could be imported from typing
-  *_pb2.pyi: B, E741, F401, F403, F405, Y021, Y023, Y026, Y053, Y054
+  *_pb2.pyi: E741, F401, F403, F405, Y021, Y023, Y026, Y053, Y054
 
 exclude = .venv*,.git
 noqa_require_code = true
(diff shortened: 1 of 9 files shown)


.flake8
```diff
@@ -30,15 +29,15 @@
 extend-ignore = A, D, N8, SIM, RST, TYP, E301, E302, E305, E501, E701
```
AlexWaygood, line 29: According to the principle you articulated in #9715 ;)

```suggestion
extend-ignore = A, B, D, N8, SIM, RST, TYP, E301, E302, E305, E501, E701
```

Avasam: Yes thank you!
Even more so devs will already have it installed in their venv

pyproject.toml
```diff
@@ -49,6 +50,12 @@ select = [
     "UP039", # don't use parens after a class definition with no bases
 ]
 
+[tool.ruff.lint.extend-per-file-ignores]
```
AlexWaygood, line 53: I'd prefer if we just ignored all flake8-bugbear rules on .pyi files, like we do currently. We know none of them are really relevant to stubs. Ideally, ruff wouldn't emit any false positives on stub files if any of these rules are enabled globally, but that's bug-prone, and I don't really see a good reason for us to keep any of these enabled for .pyi files. (Unless you have some specifically in mind that you think would be useful on stubs? :)

Avasam: Since Ruff does often care about differentiating `.py` from `.pyi` and special cases some rules, I was also trying to see if we could simplify the config and just turn it on w/o exception

As far as rules that can apply to stubs, I found a handful, but nothing that applies to typeshed, except maybe B0033

There's B018 [useless-expression](https://docs.astral.sh/ruff/rules/useless-expression/) . But it's already handled by pyright [reportInvalidStubStatement](https://github.com/microsoft/pyright/blob/main/docs/configuration.md#reportInvalidStubStatement) & [reportUnusedExpression](https://github.com/microsoft/pyright/blob/main/docs/configuration.md#reportUnusedExpression) both trigger

B021 [f-string-docstring](https://docs.astral.sh/ruff/rules/f-string-docstring/) but Typeshed doesn't include docstrings

The following can be valuable in stubs to catch actual issues in source, but there's nothing typeshed can do about it (so doesn't concern us as a third-party stubs repo)
B006 [mutable-argument-default](https://docs.astral.sh/ruff/rules/mutable-argument-default/)
B019 [cached-instance-method](https://docs.astral.sh/ruff/rules/cached-instance-method/)
B024 [abstract-base-class-without-abstract-method](https://docs.astral.sh/ruff/rules/abstract-base-class-without-abstract-method/)

B033 [duplicate-value](https://docs.astral.sh/ruff/rules/duplicate-value/) is technically not caught by any other rule. But also is a useless statement in stubs.
```py
TEST: Final = {1, 2, 3, 1} # kindof useless anyway since the type is `set[int]`, could be caught by flake8-PYI

def foo(bar: set[int] = {1, 2, 3, 1}): ... # Actually kindof somewhat useful for default values? Ignoring B006 this currently passes checks.
```

AlexWaygood: Alright, I guess I'd be open to enabling B033! That seems useful. But I'd still prefer to keep the others disabled on our `.pyi` files for now; we'll have less chance of running into bugs like https://github.com/astral-sh/ruff/issues/10083

Avasam: > people writing rules primarily for .py files won't necessarily remember that .pyi files also exist :)

Yeah I had to raise a few of those issues ^^

Unless you know of a clean way to only enable `B033` for pyi, I'll just drop that one too.

AlexWaygood (approved):
Thanks!

Avasam:
Sorry for the accidental commit in the wrong branch

PR #11501, opened 2024-02-29
Update `pyright` version to 1.1.350

sobolevn:
There are multiple releases after the current one: https://github.com/microsoft/pyright/releases/

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -100,5 +100,5 @@ extra-standard-library = [
 known-first-party = ["parse_metadata", "utils"]
 
 [tool.typeshed]
-pyright_version = "1.1.342"
+pyright_version = "1.1.352"
 oldest_supported_python = "3.8"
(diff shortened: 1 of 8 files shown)


sobolevn:
Looks like it reports quite a lot of new problems, for example:    
```
/home/runner/work/typeshed/typeshed/stdlib/builtins.pyi:450:29 - error: Position-only parameter not allowed after parameter that is not position-only (reportGeneralTypeIssues)
```

Avasam:
1.1.351 added checks for positional-only params after pos-or-kw params
Some version before that made a few `reportGeneralTypeIssues` more specific (so the report name changed)

sobolevn:
I will first update it to 350, then to newer versions 👍

sobolevn:
Working on test cases now 👍

Avasam:
No longer at my computer right now, but those openpyxl test failures look scary 😶
Could be that either pyright had a regression, or it stopped supporting something I was doing to make openpyxl descriptors work.

AlexWaygood (approved):

AlexWaygood:
Thanks!

PR #11504, opened 2024-02-29
Fix `_operator.attrgetter` pos-only params

sobolevn:
There cannot be any keywords:

```c
static PyObject *
attrgetter_new(PyTypeObject *type, PyObject *args, PyObject *kwds)
{
    attrgetterobject *ag;
    PyObject *attr;
    Py_ssize_t nattrs, idx, char_idx;

    if (!_PyArg_NoKeywords("attrgetter", kwds))
        return NULL;

// ...
```


stdlib/_operator.pyi
```diff
@@ -95,15 +95,15 @@ def length_hint(__obj: object, __default: int = 0) -> int: ...
...
     @overload
-    def __new__(cls, attr: str, __attr2: str, __attr3: str) -> attrgetter[tuple[Any, Any, Any]]: ...
+    def __new__(cls, __attr: str, __attr2: str, __attr3: str) -> attrgetter[tuple[Any, Any, Any]]: ...
     @overload
-    def __new__(cls, attr: str, __attr2: str, __attr3: str, __attr4: str) -> attrgetter[tuple[Any, Any, Any, Any]]: ...
+    def __new__(cls, __attr: str, __attr2: str, __attr3: str, __attr4: str) -> attrgetter[tuple[Any, Any, Any, Any]]: ...
     @overload
-    def __new__(cls, attr: str, *attrs: str) -> attrgetter[tuple[Any, ...]]: ...
+    def __new__(cls, __attr: str, *attrs: str) -> attrgetter[tuple[Any, ...]]: ...
     def __call__(self, obj: Any) -> _T_co: ...
```
JelleZijlstra, line 107: While we're here:

```suggestion
    def __call__(self, obj: Any, /) -> _T_co: ...
```


```
In [33]: operator.attrgetter("count")(obj="x")
Cell In[33], line 1
----> 1 operator.attrgetter("count")(obj="x")

TypeError: attrgetter() takes no keyword arguments

In [34]: operator.attrgetter("count")("x")
Out[34]: <function str.count>
```

(Also, we can use PEP 570 pos-only syntax now.)

sobolevn: Nice! 👍

AlexWaygood (approved):

PR #11505, opened 2024-02-29
Fix pos-only params in `os` module

sobolevn:
Problem:

```
/home/runner/work/typeshed/typeshed/stdlib/os/__init__.pyi
  /home/runner/work/typeshed/typeshed/stdlib/os/__init__.pyi:872:33 - error: Position-only parameter not allowed after parameter that is not position-only (reportGeneralTypeIssues)
  /home/runner/work/typeshed/typeshed/stdlib/os/__init__.pyi:873:34 - error: Position-only parameter not allowed after parameter that is not position-only (reportGeneralTypeIssues)
```

It is the same notation for `__arg1: X, *args: X`
Link: https://mypy-play.net/?mypy=latest&python=3.12&gist=b35729b668bf88a0d98a78a8d37c5be1


stdlib/os/__init__.pyi
```diff
@@ -869,8 +869,8 @@ if sys.platform != "win32":
 def abort() -> NoReturn: ...
 
 # These are defined as execl(file, *args) but the first *arg is mandatory.
-def execl(file: StrOrBytesPath, __arg0: StrOrBytesPath, *args: StrOrBytesPath) -> NoReturn: ...
-def execlp(file: StrOrBytesPath, __arg0: StrOrBytesPath, *args: StrOrBytesPath) -> NoReturn: ...
+def execl(file: StrOrBytesPath, *args: Unpack[tuple[StrOrBytesPath, Unpack[tuple[StrOrBytesPath, ...]]]]) -> NoReturn: ...
+def execlp(file: StrOrBytesPath, *args: Unpack[tuple[StrOrBytesPath, Unpack[tuple[StrOrBytesPath, ...]]]]) -> NoReturn: ...
```
AlexWaygood, line 873: Using PEP646 feels unnecessarily complicated here (and e.g. pytype still doesn't support PEP646 -- it just falls back to `Any` when it sees `Unpack`, I believe). Can't we just do this?

```suggestion
def execl(file: StrOrBytesPath, arg0: StrOrBytesPath, /, *args: StrOrBytesPath) -> NoReturn: ...
def execlp(file: StrOrBytesPath, arg0: StrOrBytesPath, /, *args: StrOrBytesPath) -> NoReturn: ...
```

sobolevn: Yes, looks easier!

AlexWaygood:
Oh interesting. Stubtest is correct; you really can pass `file` as a keyword argument:

```pycon
>>> import os
>>> os.execl(file='foo')
  File "<frozen os>", line 548, in execl
ValueError: execv() arg 2 must not be empty
```

sobolevn:
Hm, this way we have a stubtest problem:

```
error: os.execl is inconsistent, stub argument "file" should be positional or keyword (remove leading double underscore)
Stub: in file stdlib/os/__init__.pyi:872
def (Union[builtins.str, builtins.bytes, os.PathLike[builtins.str], os.PathLike[builtins.bytes]], Union[builtins.str, builtins.bytes, os.PathLike[builtins.str], os.PathLike[builtins.bytes]], *args: Union[builtins.str, builtins.bytes, os.PathLike[builtins.str], os.PathLike[builtins.bytes]]) -> Never
Runtime: in file /opt/hostedtoolcache/Python/3.10.13/x64/lib/python3.10/os.py:538
def (file, *args)

error: os.execlp is inconsistent, stub argument "file" should be positional or keyword (remove leading double underscore)
Stub: in file stdlib/os/__init__.pyi:873
def (Union[builtins.str, builtins.bytes, os.PathLike[builtins.str], os.PathLike[builtins.bytes]], Union[builtins.str, builtins.bytes, os.PathLike[builtins.str], os.PathLike[builtins.bytes]], *args: Union[builtins.str, builtins.bytes, os.PathLike[builtins.str], os.PathLike[builtins.bytes]]) -> Never
Runtime: in file /opt/hostedtoolcache/Python/3.10.13/x64/lib/python3.10/os.py:553
def (file, *args)
```

AlexWaygood:
I guess your PEP646 way is the best way, in that case :)

AlexWaygood (approved):

PR #11506, opened 2024-02-29
Fix pos-only params in `tkinter` module

sobolevn:
Same as https://github.com/python/typeshed/pull/11505


AlexWaygood:
Same comment as https://github.com/python/typeshed/pull/11505#discussion_r1507197242 ;)

AlexWaygood (approved):

Akuli:
This is less readable. What problem does this solve?

AlexWaygood:
> This is less readable. What problem does this solve?

See conversation in https://github.com/python/typeshed/pull/11505#discussion_r1507197242

Akuli:
I still don't see a practical problem being solved. Type checkers will generally tolerate small issues in stubs as long as stubs are somewhat understandable. On the other hand, IDE users will not be happy trying to decipher the new `Unpack` contraption.

AlexWaygood:
> > it's valid at runtime to pass default using a keyword argument
> 
> It's not. This method cannot be called so that you pass a keyword argument for `default`. If you try to do it, python will complain that you're passing two values for it:

Got it -- my mistake. In that case, we should be able to simply update the signature to the much-simpler

```py
def wm_iconphoto(self, default: bool, image1: _PhotoImageLike | str, /, *args: _PhotoImageLike | str) -> None: ...
```

Akuli:
> FWIW, I did test locally, but obviously not thoroughly enough :)

To be honest, the whole `wm_iconphoto()` method is a mess. A positional-only boolean parameter is just a bad idea for readability, but it's too late to change it now: that would break every tkinter program that sets a custom icon for the app.

PR #11507, opened 2024-02-29
tkinter: Simplify annotations for `wm_iconphoto`

AlexWaygood:
It's not valid to pass `default` as a keyword parameter: https://github.com/python/typeshed/pull/11506#issuecomment-1970923202


Akuli:
Lol, I created #11508 that does exactly the same thing :)

I think I will merge mine, because it has a longer and self-contained description.

AlexWaygood:
I got there first though so I still win the race ;)

PR #11508, opened 2024-02-29
tkinter: Simplify `wm_iconphoto()`

Akuli:
The whole method is just weird. The `default` argument is a boolean cannot be passed as a keyword argument:

```
>>> import tkinter
>>> root = tkinter.Tk()
>>> root.wm_iconphoto('lol', default=True)
TypeError: Wm.wm_iconphoto() got multiple values for argument 'default'
```

Also, the `default` argument **must** be given. So calling this method looks like `window.wm_iconphoto(False, some_image)` or  `window.wm_iconphoto(False, image1, image2, image3)`.

Having a required positional-only boolean parameter is dumb, but tkinter does it, and it can't be changed without breaking backwards compatibility.

See also the discussion in #11506.


AlexWaygood (approved):
Thanks!

AlexWaygood:
> We should do the same thing for `os.execl` and `os.execlp`, which were changed in #11505. `file` is pos-or-keyword according to `inspect.signature`, so we'll need to add a stubtest allowlist. But you can't _actually_ pass it as a keyword parameter, same as for `wm_iconphoto`:
> 
> ```
> >>> import os
> >>> os.execl(file='foo')
> Traceback (most recent call last):
>   File "<stdin>", line 1, in <module>
>   File "<frozen os>", line 548, in execl
> ValueError: execv() arg 2 must not be empty
> >>> os.execl('bar', file='foo')
> Traceback (most recent call last):
>   File "<stdin>", line 1, in <module>
> TypeError: execl() got multiple values for argument 'file'
> ```

Cc. @sobolevn

sobolevn:
I think the proper solution is to use `*args: Unpack`, actually. It is harder to read, but at the same time it is more correct semantically: our goal is to specify the `*args` structure and `Unpack` does just that. But, it is a bit of a mess, I agree :(

AlexWaygood:
> I think the proper solution is to use `*args: Unpack`, actually. It is harder to read, but at the same time it is more correct semantically: our goal is to specify the `*args` structure and `Unpack` does just that. But, it is a bit of a mess, I agree :(

The current signature after https://github.com/python/typeshed/pull/11505 specifies that `file` can be passed as a keyword argument to those `os` functions. Look at the REPL snippet I posted above: it cannot be passed as a keyword argument, or you get a runtime error.

sobolevn:
@AlexWaygood mypy correctly handles this case: https://mypy-play.net/?mypy=latest&python=3.12&gist=6133cabb92b165e44e240d341383fe31

It is a general rule: you cannot pass keyword args if you also pass `*args` args.

AlexWaygood:
> @AlexWaygood mypy correctly handles this case: https://mypy-play.net/?mypy=latest&python=3.12&gist=6133cabb92b165e44e240d341383fe31
> 
> It is a general rule: you cannot pass keyword args if you also pass `*args` args.

Hmm, fair enough. I still feel like the signature for the `os` functions is now unnecessarily ugly though, which isn't great for users of IDEs, who'll now have to decipher `Unpack` in autocomplete suggestions etc

sobolevn:
I always forget about IDE users, since I don't use one :)

PR #11509, opened 2024-02-29
Add pyupgrade check outdated-version-block (UP036)

Avasam:
Added the Ruff/pyupgrade [outdated-version-block (UP036)](https://docs.astral.sh/ruff/rules/outdated-version-block/) check

Updated `scripts/generate_proto_stubs.sh` and `scripts/sync_tensorflow_protobuf_stubs.sh` to fix `UP036`
This will also simplify https://github.com/python/typeshed/pull/11499/files#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711R50-R54


scripts/generate_proto_stubs.sh
```diff
@@ -73,7 +73,8 @@ PROTO_FILES=$(grep "GenProto.*google" $PYTHON_PROTOBUF_DIR/python/setup.py | \
 # shellcheck disable=SC2086
 protoc_install/bin/protoc --proto_path="$PYTHON_PROTOBUF_DIR/src" --mypy_out="relax_strict_optional_primitives:$REPO_ROOT/stubs/protobuf" $PROTO_FILES
 
-ruff check "$REPO_ROOT/stubs/protobuf" --fix-only
+# --unsafe-fixes to apply outdated-version-block (UP036)
+ruff check "$REPO_ROOT/stubs/protobuf" --fix-only --unsafe-fixes
```
AlexWaygood, line 77: I feel slightly queasy about enabling `--unsafe-fixes` here, as we have a lot of rules enabled in our `pyproject.toml`, and I haven't audited all the ones that have unsafe fixes to see just how unsafe they are. Also:
- We might add more rules to `pyproject.toml` in the future; we'll probably forget to audit those rules for unsafe fixes, given that we don't have `--unsafe-fixes` enabled in our CI elsewhere
- It feels a little weird to have a flag in this script that we don't use for ruff in CI; that might be unexpected for people running this script
- With this script, the autofix isn't even shown as a separate commit on the resulting PR; it's just "invisibly" applied before the PR is even made. So I worry that people might not even notice that ruff made changes

Avasam: Would it be fine if I added this line instead?
`ruff check "$REPO_ROOT/stubs/protobuf" --select=UP036 --fix-only --unsafe-fixes`
With that I wouldn't even need config change in this PR

AlexWaygood: That SGTM!

stubs/s2clientprotocol/s2clientprotocol/query_pb2.pyi
```diff
@@ -4,19 +4,14 @@ isort:skip_file
 """
 import builtins
 import collections.abc
-import sys
+import typing as typing_extensions
```
Avasam, line 7: I can't find a generation script for `s2clientprotocol`. I don't feel too bad about this "manual" change because it's checked by a linter, but still

scripts/sync_tensorflow_protobuf_stubs.sh
```diff
@@ -62,7 +62,8 @@ rm tensorflow/compiler/xla/service/hlo_execution_profile_data_pb2.pyi \
    tensorflow/core/util/example_proto_fast_parsing_test_pb2.pyi
 
 # --unsafe-fixes to apply outdated-version-block (UP036)
```
AlexWaygood, line 64:
```suggestion
```

stubs/s2clientprotocol/s2clientprotocol/query_pb2.pyi
AlexWaygood: yeah, we don't have an update script for this stubs package... I guess we'll need to manually run mypy-protobuf when the time comes?

AlexWaygood (approved):

PR #11510, opened 2024-02-29
make pyproject.toml valid TOML again

AlexWaygood:
all CI on `main` is failing


AlexWaygood:
merge race between #11499 and #11500

PR #11455, opened 2024-02-20
`pkg_resources`: Updates from upstream typing merge

Avasam:
So why even complete `pkg_resources` in the first place, if it's considered deprecated in favor of `importlib.resources`, `importlib.metadata` and their backports?
Because I've started merging the stubs directly into the setuptools repo, with the goal of being able to take it out of typeshed. And in doing so, found a lot of improvements to the typeshed stubs. At that point I was a few changes away from completing it, so here it is.

This also allowed me to cross-validate a handful of annotations, and I'll use the typeshed stubs as the reference for an upcoming PR that finishes typing `pkg_resources` upstream.

PR #11511, opened 2024-02-29
`pkg_resources`: Types from assignments and aliases

Avasam:
Extracted from #11455
Updated types that are either simplified or improved by using an assignement


stubs/setuptools/pkg_resources/__init__.pyi
```diff
@@ -349,11 +427,12 @@ def safe_version(version: str) -> str: ...
...
 def to_filename(name: str) -> str: ...
 def get_build_platform() -> str: ...
-def get_platform() -> str: ...
+
+get_platform = get_build_platform
+
 def get_supported_platform() -> str: ...
 def compatible_platforms(provided: str | None, required: str | None) -> bool: ...
 def get_default_cache() -> str: ...
-def get_importer(path_item: str) -> _Importer: ...
```
JelleZijlstra, line 356: You also removed this on another PR, not sure why.

Otherwise this PR looks good.

Avasam: It's a re-import, (moved all the way to the top of the file).

Both PRs touched it because this change was relevant to both (it's now an import alias and it removes the _Importer class)

JelleZijlstra: Ah right, I missed that.

PR #11512, opened 2024-02-29
`pkg_resources`: Remove type-only `_Importer` class

Avasam:
Extracted from  #11455

As I was attempting to merge stubs directly into `pkg_resources` (setuptools repo), I couldn't figure out what `_Importer` was actually supposed to be.

After working through the `register_*` methods, I think I've figured out the correct typing


stubs/setuptools/pkg_resources/__init__.pyi
```diff
@@ -18,8 +17,10 @@ _InstallerType: TypeAlias = Callable[[Requirement], Distribution | None]
 _EPDistType: TypeAlias = Distribution | Requirement | str
 _MetadataType: TypeAlias = IResourceProvider | None
 _PkgReqType: TypeAlias = str | Requirement
-_DistFinderType: TypeAlias = Callable[[_Importer, str, bool], Generator[Distribution, None, None]]
-_NSHandlerType: TypeAlias = Callable[[_Importer, str, str, types.ModuleType], str]
+_ModuleLike: TypeAlias = object | types.ModuleType  # Any object that optionally has __loader__ or __file__, usually a module
+_ProviderFactoryType: TypeAlias = Callable[[_ModuleLike], IResourceProvider]
+_DistFinderType: TypeAlias = Callable[[_T, str, bool], Iterable[Distribution]]
```
Avasam, line 22: If it wasn't for the `find_nothing` method that returns an empty tuple, this could correctly return a Generator.

stubs/setuptools/pkg_resources/__init__.pyi
```diff
@@ -353,7 +352,6 @@ def get_platform() -> str: ...
 def get_supported_platform() -> str: ...
 def compatible_platforms(provided: str | None, required: str | None) -> bool: ...
 def get_default_cache() -> str: ...
-def get_importer(path_item: str) -> _Importer: ...
```
JelleZijlstra, line 356: Why did you remove this function? Seems to exist at runtime.

JelleZijlstra: Answered on another PR, it's imported now.

PR #11514, opened 2024-02-29
Override `pkg_resources.ZipProvider.loader` type

Avasam:
This is the last change to `stubs/setuptools/@tests/stubtest_allowlist.txt` extracted from #11455


stubs/setuptools/pkg_resources/__init__.pyi
```diff
@@ -317,8 +322,11 @@ class PathMetadata(DefaultProvider):
 class ZipProvider(EggProvider):
     eagers: list[str] | None
     zip_pre: str
+    # ZipProvider's loader should always be a zipimporter
+    loader: zipimport.zipimporter
+    def __init__(self, module: _ZipLoaderModule): ...
```
Avasam, line 327:
```suggestion
    def __init__(self, module: _ZipLoaderModule) -> None: ...
```

PR #11515, opened 2024-02-29
Use `uv` for installing dependencies on all Linux-only CI jobs

AlexWaygood:
Using [`uv`](https://github.com/astral-sh/uv) rather than pip means our requirements are installed almost instantly in CI, compared to around 10s to install our `requirements.txt` file currently on each job.

I've only changed the jobs that are Linux-only in this PR. `uv` works fine on Windows, but it's slightly fiddly to install `uv` in a workflow that runs on both Windows and Linux without using pip (which is slow, defeating the point of the exercise). It's [possible](https://github.com/AlexWaygood/typeshed-stats/blob/f55c5bee175a97a345209df3fdb86bb611956168/.github/workflows/test.yml#L59-L64C14) to install uv in a cross-platform workflow without using pip, but I wasn't sure it was worth the added complexity for us here.

### DIFF
--- a/.github/workflows/stubsabot.yml
+++ b/.github/workflows/stubsabot.yml
@@ -27,14 +27,13 @@ jobs:
       - uses: actions/setup-python@v4
         with:
           python-version: "3.11"
-          cache: pip
-          cache-dependency-path: requirements-tests.txt
+      - run: curl -LsSf https://astral.sh/uv/install.sh | sh
       - name: git config
         run: |
           git config --global user.name stubsabot
           git config --global user.email '<>'
       - name: Install dependencies
-        run: pip install -r requirements-tests.txt
+        run: uv pip install -r requirements-tests.txt --system
       - name: Run stubsabot
         run: GITHUB_TOKEN=${{ secrets.GITHUB_TOKEN }} python scripts/stubsabot.py --action-level everything
 
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -30,9 +30,8 @@ jobs:
       - uses: actions/setup-python@v4
         with:
           python-version: "3.11"
-          cache: pip
-          cache-dependency-path: requirements-tests.txt
-      - run: pip install -r requirements-tests.txt
+      - run: curl -LsSf https://astral.sh/uv/install.sh | sh
+      - run: uv pip install -r requirements-tests.txt --system
       - run: python ./tests/check_consistent.py
 
   new-syntax:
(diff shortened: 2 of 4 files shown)


.github/workflows/stubsabot.yml
```diff
@@ -27,14 +27,13 @@ jobs:
       - uses: actions/setup-python@v4
         with:
           python-version: "3.11"
-          cache: pip
```
JelleZijlstra, line 30: Does losing this cache have a significant cost?

.github/workflows/tests.yml
```diff
@@ -132,25 +128,25 @@ jobs:
...
             source .venv/bin/activate
             echo "Installing packages: $DEPENDENCIES"
-            pip install $DEPENDENCIES
+            uv pip install $DEPENDENCIES --system
           fi
       - name: Activate the isolated venv for the rest of the job
         run: echo "$PWD/.venv/bin" >> $GITHUB_PATH
       - name: List 3rd-party stub dependencies installed
-        run: pip freeze --all
+        run: uv pip freeze
```
JelleZijlstra, line 149: Does uv not do the thing where setuptools etc. are only listed if you pass `--all`?

JelleZijlstra (commented):
I'm OK with doing this, thanks! A few questions.

.github/workflows/stubsabot.yml
AlexWaygood: With pip + the cache it took 10s. With uv without the cache it takes 1s. So I think we just don't need to worry about caching installs anymore with uv?

.github/workflows/tests.yml
AlexWaygood: uv doesn't install setuptools/pip/wheel into environments the way the stdlib venv module does, unless you explicitly ask for them with `--seed` (because, the theory goes, you don't need any of them if you have `uv`). So even if uv had a `--all` flag for `uv pip freeze` (it doesn't currently), we wouldn't need to use it

JelleZijlstra (approved):

.github/workflows/tests.yml
AlexWaygood: And I assume that it does, by default, list setuptools/pip etc if they are installed into an environment. Though I haven't checked.

hauntsaninja (approved):
Hooray! It'll be great when we use uv in mypy_primer too

.github/workflows/stubsabot.yml
Avasam: It only cached the download, not the install anyway. It probably helped avoid connection issues with PyPI more than anything

PR #11517, opened 2024-03-01
Use `uv` for installing dynamic dependencies in `mypy_test.py` and `regr_test.py`

AlexWaygood:
The execution time of `mypy_test.py` goes down from around 2m10s to around 1m30s in CI

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -141,7 +141,7 @@ jobs:
           if [ -n "$DEPENDENCIES" ]; then
             source .venv/bin/activate
             echo "Installing packages: $DEPENDENCIES"
-            uv pip install $DEPENDENCIES --system
+            uv pip install $DEPENDENCIES
           fi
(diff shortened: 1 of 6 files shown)


.github/workflows/tests.yml
```diff
@@ -141,7 +141,7 @@ jobs:
           if [ -n "$DEPENDENCIES" ]; then
             source .venv/bin/activate
             echo "Installing packages: $DEPENDENCIES"
-            uv pip install $DEPENDENCIES --system
+            uv pip install $DEPENDENCIES
```
AlexWaygood, line 144: This just fixes a slight oversight from #11515. We've already created a virtual environment here (we were doing so before we switched to `uv`), so there's no need to use the `--system` flag here.

AlexWaygood:
Interesting! I guess we can't do `stubtest_third_party.py` for now...

JelleZijlstra (approved):

PR #11521, opened 2024-03-02
improve type annotations in 'docutils.utils.Reporter'

danieleades:


danieleades:
I think this one's good to go @JelleZijlstra!

stubs/docutils/docutils/utils/__init__.pyi
```diff
@@ -20,16 +19,46 @@ class DependencyList:
...
+        report_level: int,
+        halt_level: int,
+        stream: IO[str] | str | bool | None = None,
+        debug: bool = False,
+        encoding: str | None = None,
+        error_handler: str = "backslashreplace",
+    ) -> None: ...
+
     source: str
-    report_level: _SystemMessageLevel
```
picnixz, line 30: Why this change?

stubs/docutils/docutils/utils/__init__.pyi
```diff
@@ -20,16 +19,46 @@ class DependencyList:
...
     WARNING_LEVEL: Literal[2]
     ERROR_LEVEL: Literal[3]
     SEVERE_LEVEL: Literal[4]
 
+    def __init__(
+        self,
+        source: str,
+        report_level: int,
+        halt_level: int,
+        stream: IO[str] | str | bool | None = None,
```
srittau, line 35: Looking at the annotations for `ErrorOutput`, this should probably be:

```suggestion
        stream: SupportsWrite[str] | SupportsWrite[bytes] | str | bool | None = None,
```

`SupportsWrite` needs to be imported from `_typeshed`.

stubs/docutils/docutils/utils/__init__.pyi
```diff
@@ -20,16 +19,46 @@ class DependencyList:
...
+    error_handler: str
+    debug_flag: bool
     report_level: _SystemMessageLevel
-    halt_level: _SystemMessageLevel
-    def __getattr__(self, __name: str) -> Incomplete: ...
+    halt_level: int
+    stream: ErrorOutput
+    encoding: str
+    observers: list[Callable[[nodes.system_message], None]]
+    max_level: int
```
srittau, line 49: Nit: Could you move the attributes above the `__init__` method?

stubs/docutils/docutils/utils/__init__.pyi
```diff
@@ -20,16 +19,46 @@ class DependencyList:
...
     source: str
+    error_handler: str
+    debug_flag: bool
     report_level: _SystemMessageLevel
-    halt_level: _SystemMessageLevel
-    def __getattr__(self, __name: str) -> Incomplete: ...
+    halt_level: int
+    stream: ErrorOutput
+    encoding: str
+    observers: list[Callable[[nodes.system_message], None]]
```
srittau, line 48: As the return value of the observers is ignored:

```suggestion
    observers: list[Callable[[nodes.system_message], object]]
```

stubs/docutils/docutils/utils/__init__.pyi
```diff
@@ -20,16 +19,46 @@ class DependencyList:
...
+    max_level: int
+    def set_conditions(
+        self, category: Any, report_level: int, halt_level: int, stream: IO[str] | None = None, debug: bool = False
```
srittau, line 51:
```suggestion
        self, category: Unused, report_level: int, halt_level: int, stream: SupportsWrite[str] | SupportsWrite[bytes] | None = None, debug: bool = False
```

`Unused` needs to be imported from `_typeshed`.

stubs/docutils/docutils/utils/__init__.pyi
```diff
@@ -20,16 +19,46 @@ class DependencyList:
...
+    ) -> None: ...
+    def attach_observer(self, observer: Callable[[nodes.system_message], None]) -> None: ...
+    def detach_observer(self, observer: Callable[[nodes.system_message], None]) -> None: ...
```
srittau, line 54: See above for the observer return types. Maybe it would make sense to add an `_Observer` type alias?

stubs/docutils/docutils/utils/__init__.pyi
```diff
@@ -20,16 +19,46 @@ class DependencyList:
...
+    def notify_observers(self, message: nodes.system_message) -> None: ...
+    def system_message(self, level: int, message: str, *children: nodes.Node, **kwargs: Any) -> nodes.system_message: ...
```
srittau, line 56:
```suggestion
    def system_message(self, level: int, message: str | Exception, *children: nodes.Node, base_node = ..., **kwargs: Unused) -> nodes.system_message: ...
```

* `message` can also be an `Exception`.
* `base_node` is one potential kwarg.
* The other kwargs are ignored.

stubs/docutils/docutils/utils/__init__.pyi
```diff
@@ -20,16 +19,46 @@ class DependencyList:
...
+    def debug(self, *args: Any, **kwargs: Any) -> nodes.system_message: ...
+    def info(self, *args: Any, **kwargs: Any) -> nodes.system_message: ...
+    def warning(self, *args: Any, **kwargs: Any) -> nodes.system_message: ...
+    def error(self, *args: Any, **kwargs: Any) -> nodes.system_message: ...
+    def severe(self, *args: Any, **kwargs: Any) -> nodes.system_message: ...
```
srittau, line 61: `args` should be `nodes.Node` per `system_message`'s annotations. See above for `kwargs` and `base_node`.

srittau (changes requested):
Thanks, notes below.

stubs/docutils/docutils/utils/__init__.pyi
srittau: In reality it doesn't matter in stubs. But when using a callable annotation at runtime, it prevents accidentally using the return type:

```python
def foo(c: Callable[[], Any):
    do_stuff(c())  # this should be an error, but isn't if Any is used
```

In typeshed, using `object` is a sign to the user that the return type isn't used. `Any` would indicate that some kind of return type is probably required, but the type system is not powerful enough to express it. Functionally, it's the same.

stubs/docutils/docutils/utils/__init__.pyi
danieleades: > base_node is one potential kwarg.
> The other kwargs are ignored.

where are you getting this from? looks to me like a couple of different kwarg keys are used, no?

srittau: I was looking at https://sourceforge.net/p/docutils/code/HEAD/tree/trunk/docutils/docutils/utils/__init__.py#l156, but I see that I missed the `attributes = kwargs.copy()` line. Still, it's better to either add the kwargs that are actually used or use `_typeshed.Incomplete` for now as a marker to add those later.

stubs/docutils/docutils/utils/__init__.pyi
danieleades:
```
+ sphinx/roles.py: note: In member "run" of class "PEP":
+ sphinx/roles.py:193:47: error: Argument 1 to "error" of "Reporter" has incompatible type "str"; expected "Node"  [arg-type]
+ sphinx/roles.py: note: In member "run" of class "RFC":
+ sphinx/roles.py:227:47: error: Argument 1 to "error" of "Reporter" has incompatible type "str"; expected "Node"  [arg-type]
+ sphinx/ext/ifconfig.py: note: In function "process_ifconfig_nodes":
+ sphinx/ext/ifconfig.py:66:46: error: Argument 1 to "error" of "Reporter" has incompatible type "str"; expected "Node"  [arg-type]
+ sphinx/ext/inheritance_diagram.py:382:52: error: Argument 1 to "warning" of "Reporter" has incompatible type "InheritanceException"; expected "Node"  [arg-type]
```

seems this is often called with a str, rather than a node

stubs/docutils/docutils/utils/__init__.pyi
```diff
@@ -43,24 +49,33 @@ class Reporter:
...
+        *children: nodes.Node,
+        base_node: nodes.Node,
+        source: str,
+        **kwargs: Incomplete,
+    ) -> nodes.system_message: ...
+    def debug(self, *args: nodes.Node, **kwargs: Any) -> nodes.system_message: ...
+    def info(self, *args: nodes.Node, **kwargs: Any) -> nodes.system_message: ...
+    def warning(self, *args: nodes.Node, **kwargs: Any) -> nodes.system_message: ...
+    def error(self, *args: nodes.Node, **kwargs: Any) -> nodes.system_message: ...
+    def severe(self, *args: nodes.Node, **kwargs: Any) -> nodes.system_message: ...
```
danieleades, line 78: these are not correct

danieleades:
> 
> ```diff
> + sphinx/roles.py: note: In member "run" of class "PEP":
> + sphinx/roles.py:193:19: error: Missing named argument "base_node" for "error" of "Reporter"  [call-arg]
> + sphinx/roles.py: note: In member "run" of class "RFC":
> + sphinx/roles.py:227:19: error: Missing named argument "base_node" for "error" of "Reporter"  [call-arg]
> + sphinx/ext/inheritance_diagram.py:382:21: error: Missing named argument "base_node" for "warning" of "Reporter"  [call-arg]
> ```

@srittau how do i provide a type annotation for a kwarg that doesn't have to be provided?

i have:

```python
    def system_message(
        self,
        level: int,
        message: str | Exception,
        *children: nodes.Node,
        base_node: nodes.Node,
        source: str,
        **kwargs: Incomplete,
    ) -> nodes.system_message: ...
```

but this causes mypy errors downstream if `system_message` is called without an explicit `base_node` kwarg

srittau (approved):

PR #11522, opened 2024-03-02
Rework our linting setup

AlexWaygood:
- Do all our linting via pre-commit, rather than doing ruff and flake8 via GitHub Actions. This has several advantages:
  - It's less code/configuration for us to worry about
  - We only have to have our `ruff` version pin in 2 places, rather than 3
  - pre-commit.ci is faster due to better caching
  - It's easier to do multiple runs of ruff with slightly different CLI arguments, which is useful if we only want to use `--unsafe-fixes` for one specific rule
  - It means we can simplify and reduce duplication in our protobuf-generation scripts
- Bump ruff to the latest version
- Add a new pre-commit hook `check-hooks-apply`. This will error out if a regex passed to the `files` parameter for any individual pre-commit hook matches 0 files
- Move ruff before black in our pre-commit config file. It's reasonably common for ruff to produce code in its autofixes that will cause black violations.

I didn't succeed in running the protobuf-stub-generation scripts locally on my MacBook (something about `realpath` not accepting an `-L1` argument on the MacOS version?), so please review the changes to those files carefully :)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -11,7 +11,7 @@ repos:
         args: [--fix=lf]
       - id: check-case-conflict
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.3.0 # must match requirements-tests.txt and tests.yml
+    rev: v0.3.0 # must match requirements-tests.txt
     hooks:
       - id: ruff
         name: Run ruff on stubs, tests and scripts
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -9,16 +9,28 @@ repos:
       - id: check-merge-conflict
       - id: mixed-line-ending
       - id: check-case-conflict
+  - repo: https://github.com/astral-sh/ruff-pre-commit
+    rev: v0.3.0 # must match requirements-tests.txt and tests.yml
+    hooks:
+      - id: ruff
+        name: Run ruff on stubs, tests and scripts
+        args: ["--exit-non-zero-on-fix"]
+      - id: ruff
+        # Run this separately because we don't really want
+        # to use --unsafe-fixes for all rules
+        name: Remove unnecessary `sys.version_info` blocks
+        args: ["--exit-non-zero-on-fix", "--select=UP036", "--unsafe-fixes"]
+      - id: ruff
+        # This is the only rule
+        # that it's really useful to run on the test cases:
+        name: Check __future__ annotations in test cases
+        args: ["--exit-non-zero-on-fix", "--select=FA", "--no-force-exclude", "--unsafe-fixes"]
+        files: '.*test_cases/.+\.py$'
   - repo: https://github.com/psf/black-pre-commit-mirror
     rev: 24.1.1 # must match requirements-tests.txt
     hooks:
       - id: black
         language_version: python3.10
-  - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.2.1 # must match requirements-tests.txt and tests.yml
-    hooks:
-      - id: ruff
-        args: [--exit-non-zero-on-fix, --fix-only]
   - repo: https://github.com/pycqa/flake8
     rev: 7.0.0 # must match requirements-tests.txt
     hooks:
@@ -28,11 +40,13 @@ repos:
           - "flake8-pyi==24.1.0" # must match requirements-tests.txt
         types: [file]
         types_or: [python, pyi]
+  - repo: meta
+    hooks:
+      - id: check-hooks-apply
 
 ci:
   autofix_commit_msg: "[pre-commit.ci] auto fixes from pre-commit.com hooks"
   autofix_prs: true
   autoupdate_commit_msg: "[pre-commit.ci] pre-commit autoupdate"
   autoupdate_schedule: quarterly
-  skip: [flake8]
   submodules: false
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -21,10 +21,10 @@ repos:
         name: Remove unnecessary `sys.version_info` blocks
         args: ["--exit-non-zero-on-fix", "--select=UP036", "--unsafe-fixes"]
       - id: ruff
-        # This is the only rule
-        # that it's really useful to run on the test cases:
-        name: Check __future__ annotations in test cases
-        args: ["--exit-non-zero-on-fix", "--select=FA", "--no-force-exclude", "--unsafe-fixes"]
+        # Very few rules are useful to run on our test cases;
+        # we explicitly enumerate them here:
+        name: Run ruff on the test cases
+        args: ["--exit-non-zero-on-fix", "--select=FA,I", "--no-force-exclude", "--unsafe-fixes"]
         files: '.*test_cases/.+\.py$'
   - repo: https://github.com/psf/black-pre-commit-mirror
     rev: 24.1.1 # must match requirements-tests.txt
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,7 +26,11 @@ exclude = [
 [tool.ruff.lint]
 # Disable all rules on test cases by default:
 # test cases often deliberately contain code
-# that might not be considered idiomatic or modern
+# that might not be considered idiomatic or modern.
+#
+# Note: some rules that are specifically useful to the test cases
+# are invoked via separate runs of ruff in pre-commit:
+# see our .pre-commit-config.yaml file for details
 exclude = ["**/test_cases/**/*.py"]
 select = [
     "B", # flake8-bugbear
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -24,6 +24,10 @@ exclude = [
 ]
 
 [tool.ruff.lint]
+# Disable all rules on test cases by default:
+# test cases often deliberately contain code
+# that might not be considered idiomatic or modern
+exclude = ["**/test_cases/**/*.py"]
 select = [
     "B", # flake8-bugbear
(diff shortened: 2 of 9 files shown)


scripts/generate_proto_stubs.sh
```diff
@@ -45,7 +45,7 @@ PYTHON_PROTOBUF_DIR="protobuf-$PYTHON_PROTOBUF_VERSION"
 VENV=venv
 python3 -m venv "$VENV"
 source "$VENV/bin/activate"
-pip install -r "$REPO_ROOT/requirements-tests.txt"  # for Black and Ruff
+pip install pre-commit
 
 # Install mypy-protobuf
 pip install mypy-protobuf=="$MYPY_PROTOBUF_VERSION"
```
Avasam, line 51: If you wanna be more consistant between scripts
```suggestion
pip install pre-commit mypy-protobuf=="$MYPY_PROTOBUF_VERSION"
```

pyproject.toml
```diff
@@ -24,6 +24,10 @@ exclude = [
 ]
 
 [tool.ruff.lint]
+# Disable all rules on test cases by default:
+# test cases often deliberately contain code
+# that might not be considered idiomatic or modern
+exclude = ["**/test_cases/**/*.py"]
```
Avasam, line 30: Are you sure about disabling flake8-future-annotations-fa, isort, probably eventually pycodestyles and possibly more eventually?

AlexWaygood: - Definitely sure about disabling pycodestyle -- I think basically any antipatterns in test cases that black won't autofix are likely to be false positive that we're deliberately using in the test cases
- Agree that the FA codes are useful -- I added some stuff to `.pre-commit-config.yaml` to ensure that we still run those on the test cases, even though they're disabled in our `pyproject.toml` file by default

AlexWaygood: I'll add a comment to `pyproject.toml` to make this clearer!

AlexWaygood: I suppose isort might also be useful on these files -- that's a fair point.

tests/check_consistent.py
AlexWaygood, line 1: The changes to this file were necessary because the test currently assumes that all repos listed in `.pre-commit-config.yaml` will have a `rev` key in them, and that's not true for the "meta" repo that I'm adding as part of this PR

pyproject.toml
Avasam: It feels a bit odd te me that local runs will differ and editors won't autofix (either from a manual call or on-save) or even visually indicate rule violations (yellow squiggle) that will be raised by CI.

As far as local runs go, maybe instead of running Ruff check, black, flake8, etc, the recommendation would be to run `pre-commit`, especially since the default command only runs on changed files. A lot of `scripts/runtests.py` could be replaced with a `pre-commit` call too.

But I'm still putoff a bit by having my editor stop showing me some issues. Granted it's only in test_cases, which already often differ in local runs with type-checkers anyway 😅

AlexWaygood: > It feels a bit odd te me that local runs will differ and editors won't autofix (either from a manual call or on-save) or even visually indicate rule violations (yellow squiggle) that will be raised by CI.

I see your point... but the only rules we'll be enabling on the test cases are isort and flake8-annotations -- and they'll be autofixed in CI anyway? So I feel like it's not a _massive_ issue... typeshed is already kind-of a hopeless case when it comes to editors understanding our directory structure anyway, in my experience 😆

> That makes me think, as far as local runs go, maybe instead of running Ruff check, black, flake8, etc, the recommendation would be to run `pre-commit`, especially since the default command only runs on changed files. A lot of `scripts/runtests.py` could be replaced with a `pre-commit` call too.

Goood point -- I pushed some updates along those lines!

Avasam: At least now local runs will match the CI, which was a bigger concern for me. And yeah not a big deal if it only affects `test_cases`.

> typeshed is already kind-of a hopeless case when it comes to editors understanding our directory structure anyway, in my experience 😆

Honestly other than `test_cases`, stubs I migrated from https://github.com/microsoft/python-type-stubs, it hasn't been too bad. Some mypy/pyright discrepancies due to strictness level here and there.

Avasam (commented):
A suggestion and a question. Everything else looks good to me

.pre-commit-config.yaml
```diff
@@ -9,16 +9,28 @@ repos:
       - id: check-merge-conflict
       - id: mixed-line-ending
```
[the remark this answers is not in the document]
AlexWaygood: Good call

scripts/runtests.py
```diff
@@ -203,7 +194,7 @@ def main() -> None:
         print(colored("\n\n--- TEST SUMMARY: One or more tests failed. See above for details. ---\n", "red"))
     else:
         print(colored("\n\n--- TEST SUMMARY: All tests passed! ---\n", "green"))
-    print("Flake8:", _SUCCESS if flake8_result.returncode == 0 else _FAILED)
+    print("pre-commit:", _SUCCESS if pre_commit_result.returncode == 0 else _FAILED)
```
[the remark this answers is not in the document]
AlexWaygood: Good point. I think it's useful for black and ruff to give nonzero exit codes when they do autofixes, because it's sometimes important to manually review the output of the autofixes before making the PR. What do you think of 6f079ee96c7aa7c39eaf8d56692203085605d5d9?

> idk how this should interact with git pre-commit hooks if `pre-commit` is installed as one, given I don't use 'em.

`¯\_(ツ)_/¯` (I don't use them either 😄)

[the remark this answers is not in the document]
AlexWaygood: Example output from runtests.py if pre-commit fails, following https://github.com/python/typeshed/commit/6f079ee96c7aa7c39eaf8d56692203085605d5d9:

image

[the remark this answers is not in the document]
Avasam: > I think it's useful for black and ruff to give nonzero exit codes when they do autofixes, because it's sometimes important to manually review the output of the autofixes before making the PR. What do you think of [6f079ee](https://github.com/python/typeshed/commit/6f079ee96c7aa7c39eaf8d56692203085605d5d9)?

I think that's valid and something we already could've done by showing a warning instead of passing, for example.
Now with pre-commit, the next best thing is probably failing with a message like you did.


One tiny nitpick is that `Check the output of pre-commit for more details.` might be overly verbose given ` One or more tests failed. See above for details.` is written two lines above. But that won't block my approval

Avasam (approved):
This all LGTM. I've used `pre-commit` as *the* CLI command for linting/formatting in some project, and I don't see an issue using it more that way in typeshed. 

All my concerns have been addressed though I may want to hear from another maintainer. Otherwise happy to merge.

scripts/runtests.py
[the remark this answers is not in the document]
JelleZijlstra: Alex finding out the hard way that Terminal is transparent on Macs...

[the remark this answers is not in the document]
AlexWaygood: WHY CAN'T THERE BE A BUTTON ON THE KEYBOARD FOR PRINT-SCREENING LIKE ON WINDOWS

scripts/generate_proto_stubs.sh
```diff
@@ -73,9 +70,7 @@ PROTO_FILES=$(grep "GenProto.*google" $PYTHON_PROTOBUF_DIR/python/setup.py | \
 # shellcheck disable=SC2086
 protoc_install/bin/protoc --proto_path="$PYTHON_PROTOBUF_DIR/src" --mypy_out="relax_strict_optional_primitives:$REPO_ROOT/stubs/protobuf" $PROTO_FILES
 
-ruff check "$REPO_ROOT/stubs/protobuf" --fix-only
-ruff check "$REPO_ROOT/stubs/protobuf" --fix-only --select=UP036 --unsafe-fixes
-black "$REPO_ROOT/stubs/protobuf"
+pre-commit run --files "$REPO_ROOT/stubs/protobuf" || true
```
JelleZijlstra, line 73: Might be worth commenting on why we do `|| true`.

JelleZijlstra (approved):

scripts/generate_proto_stubs.sh
AlexWaygood:
```suggestion
# use `|| true` so the script still continues even if a pre-commit hook
# applies autofixes (which will result in a nonzero exit code)
pre-commit run --files "$REPO_ROOT/stubs/protobuf" || true
```

scripts/sync_tensorflow_protobuf_stubs.sh
```diff
@@ -61,9 +61,7 @@ rm tensorflow/compiler/xla/service/hlo_execution_profile_data_pb2.pyi \
    tensorflow/core/protobuf/worker_service_pb2.pyi \
    tensorflow/core/util/example_proto_fast_parsing_test_pb2.pyi
 
-ruff check "$REPO_ROOT/stubs/tensorflow/tensorflow" --fix-only
-ruff check "$REPO_ROOT/stubs/protobuf" --fix-only --select=UP036 --unsafe-fixes
-black "$REPO_ROOT/stubs/tensorflow/tensorflow"
+pre-commit run --files "$REPO_ROOT/stubs/tensorflow/tensorflow" || true
```
AlexWaygood, line 64:
```suggestion
# use `|| true` so the script still continues even if a pre-commit hook
# applies autofixes (which will result in a nonzero exit code)
pre-commit run --files "$REPO_ROOT/stubs/tensorflow/tensorflow" || true
```

AlexWaygood:
Thanks, both!

PR #11523, opened 2024-03-02
improve type annotations in 'docutils.parsers.rst'

danieleades:


danieleades:
> 
> ```diff
> + sphinx/directives/other.py: note: In member "run" of class "Include":
> + sphinx/directives/other.py:421:20: error: Incompatible return value type (got "Sequence[Node]", expected "list[Node]")  [return-value]
> + sphinx/directives/other.py:425:16: error: Incompatible return value type (got "Sequence[Node]", expected "list[Node]")  [return-value]
> ```

these are downstream errors. Sphinx is assuming incorrectly that the return sequence is a list

stubs/docutils/docutils/parsers/rst/__init__.pyi
```diff
@@ -1,34 +1,66 @@
...
 
-from docutils import parsers
-from docutils.parsers.rst import states
+from docutils import nodes, parsers
+from docutils.parsers.rst.states import Inliner, RSTState, RSTStateMachine
+from docutils.statemachine import StringList
+from docutils.transforms import Transform
 
 class Parser(parsers.Parser):
+    settings_spec: ClassVar[tuple[str, str | None, tuple[str, list[str], dict[str, Any]]]]
```
danieleades, line 11: Tempted to simplify this and defer figuring out the actual type to another PR

danieleades:
@JelleZijlstra can you see the issue with the failing test? i'm a bit stumped

stubs/docutils/docutils/parsers/rst/__init__.pyi
```diff
@@ -1,34 +1,66 @@
...
 
-from docutils import parsers
-from docutils.parsers.rst import states
+from docutils import nodes, parsers
+from docutils.parsers.rst.states import Inliner, RSTState, RSTStateMachine
+from docutils.statemachine import StringList
+from docutils.transforms import Transform
 
 class Parser(parsers.Parser):
+    settings_spec: ClassVar[tuple[str, str | None, tuple[tuple[str, list[str], dict[str, Any]], ...]]]
```
danieleades, line 11: @picnixz any ideas here?

danieleades:
> 
> ```diff
> + sphinx/directives/other.py: note: In member "run" of class "Include":
> + sphinx/directives/other.py:421:20: error: Incompatible return value type (got "Sequence[Node]", expected "list[Node]")  [return-value]
> + sphinx/directives/other.py:425:16: error: Incompatible return value type (got "Sequence[Node]", expected "list[Node]")  [return-value]
> ```

see https://github.com/sphinx-doc/sphinx/pull/12042

JelleZijlstra:
> @JelleZijlstra can you see the issue with the failing test? i'm a bit stumped

Wow, that's quite a type. To understand it I copied the big type to a file and ran Black on it, which made the structure more clear:

```
tuple[
    Literal["Generic Parser Options"],
    None,
    tuple[
        tuple[
            Literal[
                'Disable directives that insert the contents of an external file; replaced with a "warning" system message.'
            ],
            builtins.list[Any],
            builtins.dict[Any, Any],
        ],
        tuple[
            Literal[
                "Enable directives that insert the contents of an external file. (default)"
            ],
            builtins.list[Any],
            builtins.dict[Any, Any],
        ],
        tuple[
            Literal[
                'Disable the "raw" directive; replaced with a "warning" system message.'
            ],
            builtins.list[Any],
            builtins.dict[Any, Any],
        ],
        tuple[
            Literal['Enable the "raw" directive. (default)'],
            builtins.list[Any],
            builtins.dict[Any, Any],
        ],
        tuple[
            Literal["Maximal number of characters in an input line. Default 10 000."],
            builtins.list[Any],
            builtins.dict[Any, Any],
        ],
    ],
    Literal["reStructuredText Parser Options"],
    None,
    tuple[
        tuple[
            Literal[
                'Recognize and link to standalone PEP references (like "PEP 258").'
            ],
            builtins.list[Any],
            builtins.dict[Any, Any],
        ],
        tuple[
            Literal[
                'Base URL for PEP references (default "https://peps.python.org/").'
            ],
            builtins.list[Any],
            builtins.dict[Any, Any],
        ],
        tuple[
            Literal['Template for PEP file part of URL. (default "pep-%04d")'],
            builtins.list[Any],
            builtins.dict[Any, Any],
        ],
        tuple[
            Literal[
                'Recognize and link to standalone RFC references (like "RFC 822").'
            ],
            builtins.list[Any],
            builtins.dict[Any, Any],
        ],
        tuple[
            Literal[
                'Base URL for RFC references (default "https://tools.ietf.org/html/").'
            ],
            builtins.list[Any],
            builtins.dict[Any, Any],
        ],
        tuple[
            Literal["Set number of spaces for tab expansion (default 8)."],
            builtins.list[Any],
            builtins.dict[Any, Any],
        ],
        tuple[
            Literal["Remove spaces before footnote references."],
            builtins.list[Any],
            builtins.dict[Any, Any],
        ],
        tuple[
            Literal["Leave spaces before footnote references."],
            builtins.list[Any],
            builtins.dict[Any, Any],
        ],
        tuple[
            Literal[
                'Token name set for parsing code with Pygments: one of "long", "short", or "none" (no parsing). Default is "long".'
            ],
            builtins.list[Any],
            builtins.dict[Any, Any],
        ],
        tuple[
            Literal[
                'Change straight quotation marks to typographic form: one of "yes", "no", "alt[ernative]" (default "no").'
            ],
            builtins.list[Any],
            builtins.dict[Any, Any],
        ],
        tuple[
            Literal['Characters to use as "smart quotes" for <language>. '],
            builtins.list[Any],
            builtins.dict[Any, Any],
        ],
        tuple[
            Literal[
                'Inline markup recognized at word boundaries only (adjacent to punctuation or whitespace). Force character-level inline markup recognition with "\\ " (backslash + space). Default.'
            ],
            builtins.list[Any],
            builtins.dict[Any, Any],
        ],
        tuple[
            Literal[
                "Inline markup recognized anywhere, regardless of surrounding characters. Backslash-escapes must be used to avoid unwanted markup recognition. Useful for East Asian languages. Experimental."
            ],
            builtins.list[Any],
            builtins.dict[Any, Any],
        ],
    ],
]
```

The type you wrote is `tuple[builtins.str, Union[builtins.str, None], builtins.tuple[tuple[builtins.str, builtins.list[builtins.str], builtins.dict[builtins.str, Any]], ...]]`. So the runtime type is essentially this type that you wrote, twice. You could fix it by adding more elements to the tuple, but is this type really going to be useful for anyone? It seems so complex that it's hard to follow.

stubs/docutils/docutils/parsers/rst/__init__.pyi
picnixz: See `docutils/__init__.py` which gives you an explanation about that object. It essentially tells you the configuration of the argument parser (which options, what to do, what they do, etc). Best idea is just to ignore literal values since it can easily change and/or you write a mypy plugin for type checking this (but honestly, it'll be hard to maintain).

danieleades: so putting aside the literal strings, what is the signature?
I can match the runtime signature exactly (see https://github.com/python/typeshed/pull/11523#issuecomment-1975190781) but that's just one instance of this spec- what is the actual structure?

danieleades:
what do you think @JelleZijlstra? Type it as `Any` and call it a day?

JelleZijlstra:
> what do you think @JelleZijlstra? Type it as `Any` and call it a day?

 I'm fine with that.

danieleades:
> > what do you think @JelleZijlstra? Type it as `Any` and call it a day?
> 
> I'm fine with that.

cool. I've used `Incomplete` rather than `Any`. is that appropriate?

picnixz:
For me, "Incomplete" essentially tells you "in the future, we might give you a better type but for now we gave up" (like a TODO) whereas "Any" is more for "I more or less know the type and can write it in Typescript probably but I cannot do it in Python for now :(" (e.g., higher-kinded types) (though, I personally think that there are some places where Any is used more as a "oh I don't know anymore, let's call it a day").

I think "Incomplete" is fine in this case but I don't think we will come back to it later. I'll try to see upon my return if there are possible instances of a different type, other than the one mentioned by Jelle.

stubs/docutils/docutils/parsers/rst/__init__.pyi
```diff
@@ -1,34 +1,67 @@
...
+    def debug(self, message: str) -> DirectiveError: ...
+    def info(self, message: str) -> DirectiveError: ...
+    def warning(self, message: str) -> DirectiveError: ...
+    def error(self, message: str) -> DirectiveError: ...
+    def severe(self, message: str) -> DirectiveError: ...
+    def assert_has_content(self) -> None: ...
+    def add_name(self, node: nodes.Node) -> None: ...
 
-def convert_directive_function(directive_fn): ...
+_DirectiveFn: TypeAlias = Callable[
```
picnixz, line 63: Should it be generic in its return type?

danieleades: actually i think `convert_directive_function` should return a `type[Directive]`, no?

stubs/docutils/docutils/parsers/rst/__init__.pyi
```diff
@@ -1,34 +1,67 @@
...
         lineno: int,
         content_offset: int,
         block_text: str,
-        state: states.RSTState,
-        state_machine: states.RSTStateMachine,
+        state: RSTState,
+        state_machine: RSTStateMachine,
     ) -> None: ...
-    def __getattr__(self, name: str) -> Incomplete: ...
+    def run(self) -> Sequence[nodes.Node]: ...
```
picnixz, line 53: Here, we could keep the `list` because we usually return a new list, so the function should be pure. With a list, we can write something like

```
ret = super().run()
ret.extend(...)
return ret
```

in a subclass and this pattern might be quite common. Or we can use `insert` for instance. As such, keeping a mutable sequence of the correct type should be fine, although we lose the covariance of sequences (but I don't think the covariance is needed and mutability is more interesting).

stubs/docutils/docutils/parsers/rst/__init__.pyi
picnixz: More precisely, I think:

```
(directive_fn: DirectiveFn[T]) -> type[T]
```

should be fine (though I am not sure of the variance here since I am still sleepy)

stubs/docutils/docutils/parsers/rst/__init__.pyi
danieleades: i'm unsure of the best approach here.
I think the covariance property is more important since if it's typed as `list[Node]` and you return `[nodes.Subclass, nodes.OtherSubclass]` that's going to be an error, right?

Also it's been typed as `Sequence` in docutils-stubs for years

and finally, if you really need a list you can always write:

```python
ret = list(super().run())
ret.extend(...)
return ret
```

I'm open to being convinced though

stubs/docutils/docutils/parsers/rst/__init__.pyi
danieleades: here's the implementation:

```python
def convert_directive_function(directive_fn):
    """
    Define & return a directive class generated from `directive_fn`.

    `directive_fn` uses the old-style, functional interface.
    """

    class FunctionalDirective(Directive):

        option_spec = getattr(directive_fn, 'options', None)
        has_content = getattr(directive_fn, 'content', False)
        _argument_spec = getattr(directive_fn, 'arguments', (0, 0, False))
        required_arguments, optional_arguments, final_argument_whitespace \
            = _argument_spec

        def run(self):
            return directive_fn(
                self.name, self.arguments, self.options, self.content,
                self.lineno, self.content_offset, self.block_text,
                self.state, self.state_machine)

    # Return new-style directive.
    return FunctionalDirective
```

so it actually returns `type[FunctionalDirective]`, but since the definition of the class is local to the function, this becomes `type[Directive]` in the parent scope

stubs/docutils/docutils/parsers/rst/__init__.pyi
picnixz: Mmh. Ok, the covariance is actually more important now that you think about it. 

Let's keep it like that.

srittau (approved):

PR #11524, opened 2024-03-03
improve type annotations in 'docutils.parsers.rst.directives.misc'

danieleades:
@JelleZijlstra i wasn't sure the best way to annotate the class variables in this one, since they are literal values (not changed at runtime) but can in principle be overridden by subclasses. What's the usual approach?


danieleades:
pyright doesn't like the untyped base class

danieleades:
this is blocked on https://github.com/python/typeshed/pull/11525, which is in turn blocked on https://github.com/python/typeshed/pull/11469

JelleZijlstra:
> @JelleZijlstra i wasn't sure the best way to annotate the class variables in this one, since they are literal values (not changed at runtime) but can in principle be overridden by subclasses. What's the usual approach?

If a subclass can override them to any value of type `T` (where `T` is e.g. str or int), I'd allocate them as `ClassVar[T]`.

stubs/docutils/docutils/parsers/rst/directives/misc.pyi
```diff
@@ -1,3 +1,35 @@
...
+
+class Class(Directive): ...
+
+class Role(Directive):
+    argument_pattern: Pattern[str]
+
+class DefaultRole(Directive): ...
+class Title(Directive): ...
+
+class MetaBody(SpecializedBody):
```
JelleZijlstra, line 28:
```suggestion
# SpecializedBody has not yet been stubbed
class MetaBody(SpecializedBody):  # pyright: ignore[reportUntypedBaseClass]
```

PR #11525, opened 2024-03-03
[WIP] improve type annotations in 'docutils.parsers.rst.states'

danieleades:
note to reviewers: i don't love the amount of `type: ignore` statements I'm having to put in here to match the implementation in docutils. Would welcome input if there's a better approach


danieleades:
> error: Module "docutils.parsers.rst" has no attribute "tableparser"

I'm deeply confused by this. It does indeed appear to be missing.

This is the imports block at the top of the `docutils.parsers.rst` module:

```python
import docutils.parsers
import docutils.statemachine
from docutils.parsers.rst import roles, states
from docutils import frontend, nodes
from docutils.transforms import universal
```

not only is `tableparser` not defined or imported in that file, it appears to be importing items from itself:

`from docutils.parsers.rst import roles, states`

???

danieleades:
@JelleZijlstra I think this PR should not me merged before https://github.com/python/typeshed/pull/11469 because the addition of generics in `statemachine` will impact the type annotations in this PR

danieleades:
this rebase is beyond me. will close it and reopen a new PR

PR #11526, opened 2024-03-03
improve type annotations in 'docutils.utils'

danieleades:


danieleades:
I think this one's good to go @JelleZijlstra!

stubs/docutils/docutils/utils/__init__.pyi
```diff
@@ -20,21 +20,92 @@ class DependencyList:
...
+
+class NameValueError(DataError): ...
+
+def decode_path(path: str) -> str: ...
+def extract_name_value(line: str) -> list[tuple[str, str]]: ...
+def clean_rcs_keywords(paragraph: nodes.paragraph, keyword_substitutions: list[tuple[Pattern[str], str]]) -> None: ...
+def relative_path(source: str, target: str) -> str: ...
+def get_stylesheet_reference(settings: Values, relative_to: str | None = None) -> str: ...
+def get_stylesheet_list(settings: Values) -> list[str]: ...
+def find_file_in_dirs(path: str, dirs: list[str]) -> str: ...
```
picnixz, line 90: For the path arguments in general, is it `str` only or `str | os.Pathlike[str]` ?

danieleades: in this case:

```python
def find_file_in_dirs(path, dirs):
    """
    Search for `path` in the list of directories `dirs`.

    Return the first expansion that matches an existing file.
    """
    path = Path(path)
    if path.is_absolute():
        return path.as_posix()
    for d in dirs:
        f = Path(d).expanduser() / path
        if f.exists():
            return f.as_posix()
    return path.as_posix()
```

so possibly we want `_typeshed.StrPath`

picnixz: StrPath is, I think, an alias of `str | os.PathLike[str]` so I think it's fine

stubs/docutils/docutils/utils/__init__.pyi
```diff
@@ -1,16 +1,17 @@
...
-from docutils import ApplicationError
-from docutils.io import FileOutput
+from docutils import ApplicationError, DataError, nodes
+from docutils.frontend import Values
+from docutils.io import ErrorOutput, FileOutput
 from docutils.nodes import document
 
 class DependencyList:
-    list: _list[str]
+    list: list[str]
```
JelleZijlstra, line 14: I personally feel this shadowing shouldn't be allowed but I suppose type checkers allow it.

stubs/docutils/docutils/utils/__init__.pyi
```diff
@@ -20,21 +21,92 @@ class DependencyList:
...
+def extract_options(field_list: nodes.field_list) -> list[tuple[str, str]]: ...
+def assemble_option_dict(
+    option_list: Iterable[tuple[str, str]], options_spec: Mapping[str, Callable[[str], Any]]
+) -> dict[str, Any]: ...
+
+class NameValueError(DataError): ...
+
+def decode_path(path: str) -> str: ...
+def extract_name_value(line: str) -> list[tuple[str, str]]: ...
+def clean_rcs_keywords(paragraph: nodes.paragraph, keyword_substitutions: list[tuple[Pattern[str], str]]) -> None: ...
```
JelleZijlstra, line 87:
```suggestion
def clean_rcs_keywords(paragraph: nodes.paragraph, keyword_substitutions: Iterable[tuple[Pattern[str], str]]) -> None: ...
```

PR #11527, opened 2024-03-04
`pkg_resources`: Make `_InstallerType` stricter and generic

Avasam:
Extracted from, and improved over #11455 to complete `installer` parameters

I'm not set on the new name(s), could be `_GenericInstallerType`, `_NoneableInstallerType`, etc.
I had to swap the order of some overloads. Some new overloads triggers `overload-overlap` from mypy, but playing around with `reveal_type` the return type seemed correct. If you're uncertain I can add some tests.


JelleZijlstra:
This one now has a merge conflict.

srittau:
And another conflict now. :)

JelleZijlstra (approved):
There was yet another conflict but I resolved it.

PR #11529, opened 2024-03-04
Add VERSION to gdb stubs

gsingh93:


gsingh93:
@pmderodat, a project I contribute to uses `gdb.VERSION`, so I wanted to add this type upstream to fix the type checking errors:
```
$ gdb -nx -ex 'pi gdb.VERSION' -ex quit
'13.1'
```

However, I can't find this documented in the GDB documentation. I'm not sure if this is an oversight in the docs, or whether this shouldn't be part of the API (and so shouldn't have type stubs either).

srittau (approved):
LGTM. Even if it's not officially part of the API, we add fields that exist on request.

PR #11531, opened 2024-03-04
Use `uv` for installing requirements dynamically in `stubtest_third_party.py`

AlexWaygood:
Follow-up to #11517

### DIFF
--- a/requirements-tests.txt
+++ b/requirements-tests.txt
@@ -21,7 +21,7 @@ termcolor>=2.3
 tomli==2.0.1
 tomlkit==0.12.3
 typing_extensions>=4.9.0rc1
-uv
+uv>=0.1.14
 
 # Type stubs used to type check our scripts.
 types-pyyaml>=6.0.12.7
--- a/tests/stubtest_third_party.py
+++ b/tests/stubtest_third_party.py
@@ -45,7 +45,15 @@ def run_stubtest(
         venv_dir = Path(tmp)
 
         try:
-            subprocess.run(["uv", "venv", venv_dir], capture_output=True, check=True)
+            # Use the --seed flag to automatically include pip in the environment,
+            # since loads of packages have undeclared dependencies on pip...
+            subprocess.run(["uv", "venv", venv_dir, "--seed"], capture_output=True, check=True)
+            if sys.version_info >= (3, 12):
+                # ...Same goes for setuptools/wheel!
+                # On Python <3.12, these are also installed if you specify --seed.
+                # On newer Pythons, however, uv does *not* install them as part of --seed
+                # (matching the behaviour of the stdlib venv module, which stopped doing this in py312).
+                subprocess.run(["uv", "pip", "install", "setuptools", "wheel"], capture_output=True, check=True)
         except subprocess.CalledProcessError as e:
             print_command_failure("Failed to setup a virtual environment", e)
             return False
--- a/tests/stubtest_third_party.py
+++ b/tests/stubtest_third_party.py
@@ -13,7 +13,7 @@ from textwrap import dedent
 from typing import NoReturn
 
 from parse_metadata import NoSuchStubError, get_recursive_requirements, read_metadata
-from utils import PYTHON_VERSION, colored, get_mypy_req, print_error, print_success_msg
+from utils import PYTHON_VERSION, colored, get_mypy_req, print_error, print_success_msg, venv_python
 
 
 def run_stubtest(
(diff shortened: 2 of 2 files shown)


JelleZijlstra:
Seems like `passlib` has an undeclared runtime dependency on `pkg_resources`, fun.

AlexWaygood:
> Seems like `passlib` has an undeclared runtime dependency on `pkg_resources`, fun.

Ouch, surprised they haven't fixed that yet since Python 3.12 doesn't install setuptools into venvs by default anymore?

AlexWaygood:
the `playsound` error looks much more fun:

```pytb
Failed to install
error: Failed to download and build: playsound==1.3.0
  Caused by: Failed to build: playsound==1.3.0
  Caused by: Build backend failed to determine extra requires with `build_wheel()`:
--- stdout:

--- stderr:
  File "<string>", line 14, in <module>
  File "/Users/alexw/Library/Caches/uv/.tmpRKpAih/.venv/lib/python3.12/site-packages/setuptools/build_meta.py", line 325, in get_requires_for_build_wheel
    return self._get_build_requires(config_settings, requirements=['wheel'])
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/alexw/Library/Caches/uv/.tmpRKpAih/.venv/lib/python3.12/site-packages/setuptools/build_meta.py", line 295, in _get_build_requires
    self.run_setup()
  File "/Users/alexw/Library/Caches/uv/.tmpRKpAih/.venv/lib/python3.12/site-packages/setuptools/build_meta.py", line 487, in run_setup
    super().run_setup(setup_script=setup_script)
  File "/Users/alexw/Library/Caches/uv/.tmpRKpAih/.venv/lib/python3.12/site-packages/setuptools/build_meta.py", line 311, in run_setup
    exec(code, locals())
  File "<string>", line 6, in <module>
  File "/Library/Frameworks/Python.framework/Versions/3.12/lib/python3.12/inspect.py", line 1282, in getsource
    lines, lnum = getsourcelines(object)
                  ^^^^^^^^^^^^^^^^^^^^^^
  File "/Library/Frameworks/Python.framework/Versions/3.12/lib/python3.12/inspect.py", line 1264, in getsourcelines
    lines, lnum = findsource(object)
                  ^^^^^^^^^^^^^^^^^^
  File "/Library/Frameworks/Python.framework/Versions/3.12/lib/python3.12/inspect.py", line 1093, in findsource
    raise OSError('could not get source code')
OSError: could not get source code
```

AlexWaygood:
## Roundup of the failures

Build-time failures (not solvable without changes to the upstream packages and/or uv)
- `playsound` [attempts to call `inspect.getsource()` on a lambda function defined inside `setup.py`](https://github.com/TaylorSMarks/playsound/blob/9cf4af20caa5ae8586f88b65659681b24f0c4e69/setup.py#L6). This doesn't work when `uv` invokes `setuptools` on the `setup.py` file, but does work when `pip` invokes `setuptools` on the `setup.py` file. Haven't looked into _exactly_ why yet.
- `dateparser`:
  - Stubtest attempts to install `dateparser` with the `fasttext` extra: https://github.com/python/typeshed/blob/e80ad6b2bce7ef6b2a20aec2d79a672859b31864/stubs/dateparser/METADATA.toml#L5
  - The `fasttext` extra adds an (obvious) dependency on `fasttext`: https://github.com/scrapinghub/dateparser/blob/29b1f94cd87c073ffbb93b44ecfeb5bbff9969e3/setup.py#L45
  - `fasttext` has an undeclared build dependency on `pip`: https://github.com/facebookresearch/fastText/blob/ae1fe80e9fbbe97d16214647310e2523327cecab/setup.py#L41

Runtime failures (these can be solved by installing setuptools into the environment before each package's stubtest run, like I did with my latest commit):
- `libsass`:
  - The `sassutils` package [imports `setuptools` and `pkg_resources`](https://github.com/sass/libsass-python/blob/368a16626e060c1e5999447bccb6e009e14c1e3c/sassutils/distutils.py#L76-L77) (which is distributed with `setuptools`) without declaring a dependency on `setuptools`. `setuptools` is automatically installed into a virtual environment on Python <3.12 with the stdlib venv module, and you can get similar behaviour with uv if you specify `--seed` (but uv, like the stdlib venv module, doesn't include setuptools by default on py312+, even if you specify `--seed`).
  - I wondered whether `sassutils` was meant to be public API or not, since the main package distributed with `libsass` is the `sass` package. But `sassutils` [does seem to be documented](https://sass.github.io/libsass-python/sassutils.html).
- `passlib` [imports from `pkg_resources`](https://foss.heptapod.net/python-libs/passlib/-/blob/branch/stable/passlib/pwd.py?ref_type=heads#L16) without declaring a dependency on `setuptools`.
- `google-cloud-ndb`:
  - Currently we have `google-cloud-ndb` pinned to v2.2.* for stubtest: see https://github.com/python/typeshed/pull/11520/files
  - V2.2.* has an undeclared dependency on `pkg_resources`. V2.3.* does not. The solution there is just to update our stubs to v2.3.*.

srittau:
While the performance improvements would obviously be nice, this doesn't sound as if uv was really production ready, at least not for our purposes. I'd prefer not to add this additional headache at this point.

AlexWaygood:
> While the performance improvements would obviously be nice, this doesn't sound as if uv was really production ready, at least not for our purposes. I'd prefer not to add this additional headache at this point.

Strictly speaking, these are all errors in third-party packages rather than `uv`, _per se_. But, fully agreed that this would cause unnecessary complication for our purposes right now.

Avasam:
If anything, it at least revealed some issues in the ecosystem that can be raised upstream. I can't blame `fasttext` for assuming that `pip` is always installed, for example, but Python packaging is changing and some assumptions can't be made anymore.

srittau:
One thing we could do is start using `pip --use-pep517` as soon as the existing packages are fixed and start rejecting third-party package that are not installable that way. Then we can switch to uv at some later point as soon as the remaining issues are ironed out.

PR #11537, opened 2024-03-07
Fix type of `concurrent.futures.wait` for Python <= 3.8

### LINKED ISSUE #11533: `concurrent.futures.wait` requires a `Collection` before Python 3.9
https://github.com/python/cpython/pull/30168 which converts the input to a set was not backported to Python < 3.9.  Currently, `concurrent.futures.wait` is typed to take an `Iterable`, but that will error on Python 3.8 and below.

layday:
Closes #11533.


srittau (approved):
Thanks!

PR #11538, opened 2024-03-08
Use strict pyright configs for `pkg_resources`

Avasam:
Use strict pyright configs for `pkg_resources`
Extracted from #11455


srittau:
Has a conflict now.

srittau (approved):

